PR #467, opened 2014-12-29
Point Homebrew users to a GCC 4.5 formula that works on 10.10

lethosor:


lethosor:
`make package` uses DF's libstdc++ (4.5) by default, so a DFHack build compiled with a newer GCC won't be distributable without a change to CMakeLists.txt, as in #463.

PR #472, opened 2015-01-01
New tweaks

lethosor:
* `eggs-fertile`: Displays an egg fertility indicator on nestboxes
* `max-wheelbarrow`: Allows assigning more than 3 wheelbarrows to a stockpile

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,9 @@ DFHack Future
     Fixes
     New Plugins
     New Scripts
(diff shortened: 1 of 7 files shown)


PeridexisErrant:
Could you also add these to `dfhack.init-example`?

lethosor:
I consider `eggs-fertile` to be a form of cheating, but I'll add `max-wheelbarrow`.

PR #476, opened 2015-01-05
Fix error in full-heal.lua when unit not given.

cdombroski:


lethosor:
Unless I'm missing something, this was already fixed by #443

cdombroski:
That might be true. I was just looking at another potential issue with full-heal.lua and noticed that it didn't work if you didn't specify the unit in the command. Looks like we've got the same code so I'll just close this.

Issue #479, opened 2015-01-06
Replace `needs_porting/copypaste.cpp` with a Quickfort blueprints generator


PeridexisErrant:
The original plugin was designed to copy a section of the fortress to another section.  

I believe that generating blueprints is a better way to go though; as it's less 'cheaty', allows multi-fort usage and sharing, and the blueprints can be implemented with other tools (eg digfort and fastdwarf).

The command would be something like `make-blueprint <volume> <plan_name> [dig [build [place [query]]]]` (see [QF syntax](http://www.joelpt.net/quickfort/#editing-blueprints)), and the script would output the blueprint for the corresponding phase or phases.

The dig phase would read the shape of rooms to carve out, and the blueprint could be implemented by `digfort.rb`.  The build phase would place all buildings, furniture, and constructions, and the blueprint might be implemented by the new `fortplan` plugin.  The place phase records stockpiles, and would have to be implemented by Quickfort.  The query phase records created rooms and is likewise not yet possible with DFHack.

These could be implemented in stages, and would be a giant leap for fortress construction techniques - a way to eliminate the complex and tedious process of writing a blueprint.  Instead you could just decide to record something you like!

cdombroski:
I've made a start on this, any feedback welcome. cdombroski@8193273

cdombroski:
[Completed dig mode](cdombroski@b627240)
Ran: blueprint 10 10 -5 test
At this location:
![screenshot 2015-01-08 12 07 12](https://cloud.githubusercontent.com/assets/244909/5666815/77f826ca-9730-11e4-8946-6e19a1cb82ee.png)
![screenshot 2015-01-08 12 08 44](https://cloud.githubusercontent.com/assets/244909/5666821/7e064934-9730-11e4-9afb-f665808cb357.png)
![screenshot 2015-01-08 12 08 48](https://cloud.githubusercontent.com/assets/244909/5666822/80abe194-9730-11e4-8238-40a29cbb2e28.png)
![screenshot 2015-01-08 12 08 51](https://cloud.githubusercontent.com/assets/244909/5666824/829ece08-9730-11e4-97e6-7b94c54e1bd8.png)
![screenshot 2015-01-08 12 08 53](https://cloud.githubusercontent.com/assets/244909/5666825/8432d7b4-9730-11e4-8b92-0175f33fa1b5.png)
and got [this output](https://gist.github.com/cdombroski/8e91cf9676e13dbf136f)

[...]

PR #483, opened 2015-01-08
Finish off needs_porting - issues raised and hotkeys ported

PeridexisErrant:
`copypaste` needs to be redone, not ported (see issue #479).

`creaturemanager` is obsolete due to the manipulator plugin, vjeck's scripts for the cheaty bits, and changes to DF itself.

`hotkey-notes.lua` is now working nicely, and will only report on hotkeys that have been set.


NEWS
```diff
@@ -1,4 +1,8 @@
 DFHack Future
+    New Scripts
+		hotkey-notes:  print key, name, and jump position of hotkeys
+    Removed
+		needs_porting/*
```
lethosor: Could you use spaces here? Tabs and spaces generally don't mix very well, as evidenced here, and the rest of this file uses spaces.

scripts/hotkey-notes.lua
```diff
@@ -0,0 +1,12 @@
+-- prints info on assigned hotkeys to the console
+
+local h_list = {'F1', 'F2', 'F3', 'F4', 'F5', 'F6', 'F7', 'F8',
+                'Shift+F1', 'Shift+F2', 'Shift+F3', 'Shift+F4',
+				'Shift+F5', 'Shift+F6', 'Shift+F7', 'Shift+F8'}
+
```
lethosor: It would be more correct to use something like
```lua
dfhack.screen.getKeyDisplay(df.interface_key.D_HOTKEY1)
```
here, to account for possible changes to keybindings. In fact, since the binding IDs are in order, you could get rid of `h_list` entirely and replace `h_list[i]` with this:
```lua
dfhack.screen.getKeyDisplay(df.interface_key.D_HOTKEY1 + i - 1)
```

scripts/hotkey-notes.lua
```diff
@@ -0,0 +1,12 @@
...
+for i=1, #df.global.ui.main.hotkeys do
+    local hk = df.global.ui.main.hotkeys[i-1]
+	if hk.cmd ~= -1 then
+		print(h_list[i]..': '..hk.name..':  x='..hk.x..'  y='..hk.y..'  z='..hk.z)
+	end
+end
```
lethosor: Same whitespace issues here.

PR #485, opened 2015-01-11
Expose key modifier state to C++/Lua

lethosor:


library/Core.cpp
```diff
@@ -1589,31 +1590,27 @@ int Core::DFH_SDL_Event(SDL::Event* ev)
...
-        }
-        else
-        if(ke->state == SDL::BTN_PRESSED && !hotkey_states[ke->ksym.sym])
+        if (ke->ksym.sym == SDL::K_LSHIFT || ke->ksym.sym == SDL::K_RSHIFT)
+            modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_SHIFT : modstate & ~MOD_SHIFT;
+        else if (ke->ksym.sym == SDL::K_LCTRL || ke->ksym.sym == SDL::K_RCTRL)
+            modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_CTRL : modstate & ~MOD_CTRL;
+        else if (ke->ksym.sym == SDL::K_LALT || ke->ksym.sym == SDL::K_RALT)
+            modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_ALT : modstate & ~MOD_ALT;
+        else if(ke->state == SDL::BTN_PRESSED && !hotkey_states[ke->ksym.sym])
```
crowder: This looks like it wants to be a switch...

lethosor: What, this? I was using the existing structure, but I suppose this works as well.
```cpp
switch (ke->ksym.sym)
{
    case SDL::K_LSHIFT:
    case SDL::K_RSHIFT:
        modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_SHIFT : modstate & ~MOD_SHIFT;
        break;
    case SDL::K_LCTRL:
    case SDL::K_RCTRL:
        modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_CTRL : modstate & ~MOD_CTRL;
        break;
    case SDL::K_LALT:
    case SDL::K_RALT:
        modstate = (ev->type == SDL::ET_KEYDOWN) ? modstate | MOD_ALT : modstate & ~MOD_ALT;
        break;
    default:
        if(ke->state == SDL::BTN_PRESSED && !hotkey_states[ke->ksym.sym])
        {
            hotkey_states[ke->ksym.sym] = true;

            // Use unicode so Windows gives the correct value for the
            // user's Input Language
            if((ke->ksym.unicode & 0xff80) == 0)
            {
                int key = UnicodeAwareSym(*ke);
                SelectHotkey(key, modstate);
            }
            else
            {
                // Pretend non-ascii characters don't happen:
                SelectHotkey(ke->ksym.sym, modstate);
            }
        }
        else if(ke->state == SDL::BTN_RELEASED)
        {
            hotkey_states[ke->ksym.sym] = false;
        }
}
```

PR #489, opened 2015-01-13
New Plugin:blueprint

### LINKED ISSUE #479: Replace `needs_porting/copypaste.cpp` with a Quickfort blueprints generator

cdombroski:
Fixes #479 

The base functionality is here, but I'll probably do a few more commits to clean up how the phase list in the parameters is handled and how buildings and stockpiles are written out.


PeridexisErrant:
This is very exciting - I'm watching with great interest.

cdombroski:
Unless somebody breaks it or has another idea, I've finished working on this now.

expwnent:
Merged manually.

1. fixTexts should be run once before releases. It keeps the repo smaller that way.
2. Please do PRs targetted at DFHack:develop instead in the future.

expwnent:
I had to fix [this compilation bug](https://github.com/DFHack/dfhack/commit/53a8032f8af30ae668bcf27e44474b7e6667180d) to make it compile on linux. Be aware of that for the future.

PR #490, opened 2015-01-13
Fix digexp vs expdig typo

tdhsmith:
Also ran the rst2html script, which readded some components that seem to have gone missing.


expwnent:
rst2html should generally be run once just before a release to keep the repo smaller, and please do pull requests to DFHack:develop instead of DFHack:master.

PR #495, opened 2015-01-18
fix skill-change so it would use insert_sorted

warmist:
This is needed because the skills vector is actually sorted. Adding skill with id<last one would mess things up very badly.


warmist:
Also not sure how adding xp works?
Does it autolevel if you exceed the xp for next level? If not we might need to add that to.

PR #505, opened 2015-01-26
Allow user to set additional entries in LD_PRELOAD on linux

miffedmap:
Some 64bit distros fail with errors about libpng when running df becaue of multilib issues. This
can be fixed by setting LD_PRELOAD properly, but dfhack doesn't let you dod that without manually
editing the script. 

This patch includes the user specified PRELOAD_LIB if it exists in the final LD_PRELOAD value used.

### DIFF
--- a/package/linux/dfhack
+++ b/package/linux/dfhack
@@ -38,12 +38,12 @@ old_tty_settings=$(stty -g)
 
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:"./hack/libs":"./hack"
 
-PRELOAD_LIB=./hack/libdfhack.so
+PRELOAD_LIB="$PRELOAD_LIB ./hack/libdfhack.so"
 
 case "$1" in
   -g | --gdb)
     shift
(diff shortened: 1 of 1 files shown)


package/linux/dfhack
```diff
@@ -65,7 +65,7 @@ case "$1" in
     ret=$?
     ;;
   *)
-    setarch i386 -R env LD_PRELOAD=$PRELOAD_LIB ./libs/Dwarf_Fortress "$@"
+    setarch i386 -R env LD_PRELOAD="$PRELOAD_LIB" ./libs/Dwarf_Fortress "$@"
```
lethosor: You should add quotes around `$PRELOAD_LIB` to the lines above as well (otherwise this will prevent DFHack from being loaded when using a debugger).

miffedmap: I thought I did, what line do you mean?

lethosor: 54, 59, and 64

miffedmap: Fixed. Thanks for noticing.

miffedmap:
redirecting to develop branch.

PR #509, opened 2015-01-27
Updated add-thought for proper stress change.

Putnam3145:
Apparently, just adding the thought doesn't cut it.

Also added thought strength to the whole thing.


expwnent:
Merged into develop.

PR #519, opened 2015-01-29
Fs expansion

expwnent:
Makes it so we can detect file modification times, creation times, etc.


expwnent:
Thanks for doing this, by the way.

Issue #276, opened 2014-08-04
Trying to use the wrong libz?


smickles:
On fedora, in order to run vanilla DF, I have to `export LD_PRELOAD=/usr/lib/libz.so.1`. However, it seems the way dfhack's launch script is, I cannot do this. Also, even though I specified this library during build at the advanced section of ccmake, the issue persists.

The error presents itself in the exact same manner in which it does on a vanilla install. This is a popup box titled "tileset not found" and the content of "Not found: data/art/curses_640x300.png".

Switching to .bmp does not fix this, nor does using a graphics set.

Please advise.

lethosor:
This should be easier to fix now with PRELOAD_LIB working more reliably.

PR #535, opened 2015-02-07
LD_PRELOAD should use colon rather then space as delimiter

miffedmap:
#508

### DIFF
--- a/package/linux/dfhack
+++ b/package/linux/dfhack
@@ -38,7 +38,7 @@ old_tty_settings=$(stty -g)
 
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:"./hack/libs":"./hack"
 
-PRELOAD_LIB="$PRELOAD_LIB ./hack/libdfhack.so"
+PRELOAD_LIB="$PRELOAD_LIB:./hack/libdfhack.so"
 
 case "$1" in
   -g | --gdb)


lethosor:
According to #276 and several online manpages ([this](http://manpages.ubuntu.com/manpages/utopic/man8/ld.so.8.html) and [this](http://lmbench.sourceforge.net/cgi-bin/man?keyword=ld.so&section=8), to name a couple), spaces shouldn't be a problem.

package/linux/dfhack
```diff
@@ -38,7 +38,7 @@ old_tty_settings=$(stty -g)
 
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:"./hack/libs":"./hack"
 
-PRELOAD_LIB="$PRELOAD_LIB ./hack/libdfhack.so"
+PRELOAD_LIB="$PRELOAD_LIB:/hack/libdfhack.so"
```
lethosor: Should still be `./hack/libdfhack.so`.

miffedmap: all thumbs here.

miffedmap:
Yeah, but I haven't tested it with multiple entries having dirnames containing spaces and I'm guessing that wouldn't work properly.

Pushed the missing period -- you have sharp eyes.

PR #539, opened 2015-02-08
Adds new console methods and commands to hide/show console. 

warmist:
Only on windows.


lethosor:
Yeah, it's not possible on OS X and Linux, due to the fact that DFHack can be run from any terminal.

PR #542, opened 2015-02-08
fix-armory: Substitute removed jobtypes with working ones

scamtank:
Replaced StoreItemInChest with StoreItemInHospital and StoreItemInCabinet with StoreOwnedItem. Seems to work on a basic level with waterskins and uniform cloaks, provided the revised weaponrack-unassign patch is applied.


lethosor:
You'll also want to re-enable fix-armory in plugins/CMakeLists.txt. Looks good otherwise.

scamtank:
Heh, oops.

PR #543, opened 2015-02-08
Fix fix-armory

lethosor:
Patch submitted by scamtank on the forums:
http://www.bay12forums.com/smf/index.php?topic=139553.msg6021113#msg6021113


lethosor:
See #542

PR #549, opened 2015-02-11
exportlegends -  more data in legends_plus.xml

PeridexisErrant:
Originally written by Mason11987, adapted and included with permission.

The 'info' option now exports more data than is possible in vanilla, to a "region-date-legends_plus.xml" file developed to extend the World Viewer utility and potentially compatible with others.

### DIFF
--- a/scripts/exportlegends.lua
+++ b/scripts/exportlegends.lua
@@ -26,7 +26,7 @@ local MAPS = {
     "Diplomacy",
 }
 
-getItemSubTypeName = function (itemType, subType)
+function getItemSubTypeName(itemType, subType)
     if (dfhack.items.getSubtypeCount(itemType)) <= 0 then
         return tostring(-1)
     end
@@ -38,7 +38,7 @@ getItemSubTypeName = function (itemType, subType)
     end
 end
 
-findEntity = function (id)
+function findEntity(id)
     for k,v in ipairs(df.global.world.entities.all) do
         if (v.id == id) then
             return v
--- a/scripts/exportlegends.lua
+++ b/scripts/exportlegends.lua
@@ -26,12 +26,508 @@ local MAPS = {
     "Diplomacy",
 }
 
+getItemSubTypeName = function (itemType, subType)
+    if (dfhack.items.getSubtypeCount(itemType)) <= 0 then
+        return tostring(-1)
+    end
+    local subtypename = dfhack.items.getSubtypeDef(itemType, subType)
+    if (subtypename == nil) then
+        return tostring(-1)
+    else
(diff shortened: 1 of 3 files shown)


lethosor:
Lines 163, 164, 165, 166, 430, 431, and 433 contain tabs.

scripts/exportlegends.lua
```diff
@@ -26,12 +26,508 @@ local MAPS = {
     "Diplomacy",
 }
 
+getItemSubTypeName = function (itemType, subType)
```
lethosor: @mason11987 Not that it makes a difference, but is there a reason for using `getItemSubTypeName = function` instead of `function getItemSubTypeName` that I'm not aware of?

PeridexisErrant:
Things that happen without git... this is down to Mason writing most of the code, me tweaking it, him tweaking it further, and then me making the pull (with permission).

I assume that switching `foo = function (x)` is identical to `function foo(x)`?

Mason11987:
I assume there is no difference, I'm not an expert in lua, this was my
quick-hack solution while teaching myself lua.  So if the other way is
better style, go for it :).

On Wed, Feb 11, 2015 at 11:44 PM, Peridexis Errant <notifications@github.com

scripts/exportlegends.lua
```diff
@@ -26,12 +26,501 @@ local MAPS = {
...
+function export_more_legends_xml()
+    local julian_day = math.floor(df.global.cur_year_tick / 1200) + 1
+    local month = math.floor(julian_day / 28) + 1 --days and months are 1-indexed
+    local day = julian_day % 28 + 1
+    local year_str = string.format('%0'..math.max(5, string.len(''..df.global.cur_year))..'d', df.global.cur_year)
+    local date_str = year_str..string.format('-%02d-%02d', month, day)
+
+    io.output(tostring(df.global.world.cur_savegame.save_dir).."-"..date_str.."-legends_plus.xml")
+
+    io.write ("<?xml version=\"1.0\" encoding='UTF-8'?>".."\n")
```
lethosor: @BenLubar pointed out that this is not correct, as dfhack.TranslateName() does not return UTF-8 output. You should use `dfhack.df2utf` to convert it appropriately.

Mason11987: Good to know, I'll see if I can't tweak the script to incorporate that.  so
just use dfhack.df2utf(dfhack.translatename(NAME)) instead?

lethosor: Already fixed in https://github.com/DFHack/dfhack/commit/d14a8b4308799bba63d27527da756081b63fdb3e (just keep it in mind for future changes)

PR #557, opened 2015-02-17
Add current job as a view mode in manipulator

jlogsdon:
This replaces the Profession/Squad toggle with an enum of 3 different modes: Profession, Squad or Job. Sorting and the like works with it properly (Idle has the lowest "priority" when sorting, not sure if that will be desired or not but its easy enough to rip out). Eventually it may be nice to set the color of the job to the profession it belongs to, but I'm not currently sure of the best way to handle that.

This is my first time writing any C++ code ever, so there may be better ways to do the mode switching that I am unaware of.

### DIFF
--- a/Contributors.rst
+++ b/Contributors.rst
@@ -75,6 +75,7 @@ Ben Lubar               BenLubar
 miffedmap               miffedmap
 scamtank                scamtank
 Mason11987              Mason11987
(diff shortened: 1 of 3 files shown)


lethosor:
Closing and reopening this to trigger Travis.

PR #558, opened 2015-02-19
Pass RTLD_LOCAL to dlopen()

### LINKED ISSUE #205: add-spatter plugin crashes the game on OSX
When loading an existing savegame or generating a new game with raws that use the add-spatter plugin the game crashes. After deleting the plugin the crash disappears. Vanilla does not crash.
Tested with Masterwork mod v.3d and DfHack r3 Darwin - compiled on 2013. 03. 15 running on OSX 10.7.5.

Crashlog: http://pastebin.com/Ppnz7z49

lethosor:
May have fixed #205

### DIFF
--- a/library/PlugLoad-posix.cpp
+++ b/library/PlugLoad-posix.cpp
@@ -26,7 +26,7 @@ namespace DFHack
     DFLibrary * OpenPlugin (const char * filename)
     {
         dlerror();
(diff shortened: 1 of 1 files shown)


lethosor:
Confirmed to fix #205 (removing the AS prefix from classes in add-spatter.cpp crashes except with this change).

PR #559, opened 2015-02-19
Move DFHack version to a separate (static) library

lethosor:
This allows the DFHack version to be changed without recompiling everything - only DFHackVersion.cpp (a two-line file) needs to be recompiled.
(Any structure/include changes will still cause source files that depend on them to be rebuilt.)


expwnent:
Will make/cmake correctly detect that it needs to update DFHackVersion even though the file itself hasn't changed when the version macro has?

lethosor:
Yes. Since DFHACK_VERSION is a preprocessor definition, CMake will rebuild any files that it is defined for when it changes. (This is why changing it previously required rebuilding _every_ file, since DFHACK_VERSION was a project-wide definition.)

library/CMakeLists.txt
```diff
@@ -249,6 +250,9 @@ ELSE(WIN32)
   SET(PROJECT_LIBS psapi dfhack-tinyxml dfhack-tinythread)
 ENDIF()
 
+ADD_LIBRARY(dfhack-version STATIC DFHackVersion.cpp)
+SET_TARGET_PROPERTIES(dfhack-version PROPERTIES COMPILE_DEFINITIONS DFHACK_VERSION="${DFHACK_VERSION}")
+
```
lethosor: This should work on Windows, but it would be nice if somebody could test it before merging.

expwnent:
Seems to work.

PR #570, opened 2015-03-05
Add view-item-info to extend item viewscreens

PeridexisErrant:
A script to extend the item or unit viewscreen with additional information including properties such as material info, weapon and attack properties, armor thickness and coverage, and more.  Adapted from nb_item_info by Raidau, which has a lot of special cases unique to his unreleased mod.

History is visible in [this branch](https://github.com/PeridexisErrant/dfhack/tree/add-view-item-info), in more commits than are needed here.  There's a couple of magic numbers I couldn't track down, but not many.  And I haven't found any bugs in a while, but it's not like I'm any good with lua so that doesn't guarantee everything is working...


lethosor:
I haven't been following development of this closely. Have the issues with blocking certain screens been resolved? Aside from that, it looks good, although you may want to move the script to the gui/ subdirectory.

PeridexisErrant:
This is a substantially different version - I've removed a lot of the custom stuff, including the image popups (which were blocking) and custom descriptions.  

Due to the changes it doesn't actually paint any new viewscreens, but simply inserts text into an existing one (I'm pretty sure...).  It seemed to me that was less a `gui/` thing, but happy to move if you think it's appropriate.

scripts/item-descriptions.lua
```diff
@@ -0,0 +1,75 @@
+-- Holds custom descriptions for view-item-info
+-- By PeridexisErrant
+
+print("This is a content library; calling it does nothing.")
```
lethosor: This will be run when `dfhack.script_environment` is called as well, which probably isn't what you intended. Since this file isn't intended to be used as a script at all, it should probably go in `hack/lua` (`library/lua` in the source tree) or a subdirectory of it (`items`?), which can then be used with `require()`.

lethosor: Oh, your idea is that this script can be overridden on a per-save basis. I'd recommend implementing `desc_of_item` in a Lua library and loading item descriptions from other Lua files/scripts, which could simply contain a table of item descriptions:
```lua
{ANVIL =  {"An essential component of the forge."}}
```

PeridexisErrant: The idea is that there are two ways for mods to override stuff on a per-save basis:  they can add or replace individual descriptions with `more-item-descriptions.lua`, or discard the whole thing and start over with `item-descriptions.lua`.  That may well be overcomplicated.

Re: implementation detail - reworking this is how I learnt lua, so I have no idea what specific structure you're suggesting but it's probably a good idea.

lethosor: [Here](https://github.com/lethosor/dfhack-scripts/tree/master/example) is an example. (I don't know for sure if save-specific scripts in subdirectories are supported, but it should be easy enough to change the example paths.)

expwnent:
```Lua
if not moduleMode then
 print("scripts/item-descriptions.lua is a content library; calling it does nothing.")
end
```

That will correct the ```script_environment``` issue and still print a helpful warning if called by the user. I also recommend saying the name of the file just in case it's triggered by something else. I'm tempted to suggest throwing an error so they see a stack trace but that's persnickety even by my standards.

PeridexisErrant:
Closing due to the history getting convoluted and (locally) too conflicted to work with.  Will reopen from a clean base soon.

PR #572, opened 2015-03-06
Autolabor: negative pool size uses most unskilled dwarves.

melkor217:
subj

PR #574, opened 2015-03-08
Negative autolabor pool size uses most unskilled dwarves.

lethosor:
From #572


lethosor:
Looks good. @melkor217 do you have a preferred name for Contributors.rst?

melkor217:
Nope :(

PR #578, opened 2015-03-15
Stockflow repairs

eswald:
* A potential memory leak has been closed.
* An error message had been reported in arena mode, as reported by Propman and scamtank in the forum thread.
* The DF version is no longer used for deciding which reactions to exclude from the list.
* Ballista arrow heads had been getting ordered incorrectly.


PeridexisErrant:
I think the linter is just complaining about the trailing spaces on L#158, which is otherwise empty.

Issue #548, opened 2015-02-10
Support built-in enable/disable commands for lua scripts


lethosor:
Currently, scripts that can be enabled or disabled can only accept commands of the form `script-name enable|disable`. It would be nice to support `enable|disable script-name`, for consistency with plugins. (A method similar to the way lua scripts are run would be simple, although it could be difficult to list scripts that support these commands when invoking `enable` or `disable`.)

Issue #568, opened 2015-03-03
Attempting to run a function from a non-existent script causes an error


PeridexisErrant:
While I can see that this makes sense, it also means it's impossible to attempt to call an external script which may or may not be available.  If there's a native way to check in advance whether the second script exists, I don't know it. 

This is a problem for me at the moment because I'm working on a script which stores data (descriptions of items, creatures, and materials) as a function.  If attempting to access a script, which for a mods would be in `/raw/scripts`, failed non-fatally I could include all the vanilla config in the same file.  As is, I'll need to supply a second script with the vanilla info, and allow that to be overridden by a copy in the raws.  At best this will require some inelegant workarounds, and at least a placeholder script seems essential.

lethosor:
If you're referring to `script_environment` from bf5e491647da9a16126a8bda9c533b4e839e2f4c, wouldn't `pcall` work?

PeridexisErrant:
I didn't know of `pcall` or `dfhack.findScript`, but the latter works perfectly by returning a bool.  Thanks.

PR #580, opened 2015-03-19
Allow Lua scripts to implement enable/disable commands and specify other behavior (e.g. "module mode") that they implement

lethosor:
See #548
This currently works by scanning comments for lines of the form `--@ enable = true`, for example (`--@ enable = false`  also works, if script authors want to be explicit). Feel free to suggest a better way of retrieving data from a script without actually executing it.

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -399,9 +399,13 @@ function Script:get_flags()
         local f = io.open(self.path)
         local contents = f:read('*all')
         f:close()
-        for line in contents:gmatch('--@([^\n]+)') do
+        for line in contents:gmatch('%-%-@([^\n]+)') do
             local chunk = load(line, self.path, 't', self._flags)
-            if chunk then chunk() end
+            if chunk then
+                chunk()
+            else
+                dfhack.printerr('Parse error: ' .. line)
+            end
         end
     end
     return self._flags
@@ -434,7 +438,7 @@ function dfhack.findScript(name)
 end
 
 local valid_script_flags = {
-    enable = {required = true},
+    enable = {required = true, error = 'Does not recognize enable/disable commands'},
     enable_state = {required = false},
     module = {required = true, error = 'Cannot be used as a module'},
 }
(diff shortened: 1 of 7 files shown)


lethosor:
@PeridexisErrant As this stands, this would require a small change to all scripts that could be used as modules (since many scripts currently don't handle being used as modules safely). You mentioned using scripts as modules in #568 - which (if any) in this repo were you referring to?

PeridexisErrant:
The only one I know of is `item-descriptions.lua`, in pull #570 (which isn't in yet).  Presumably some mod-focussed scripts (Roses' scripts?) would also be affected, but I guess they're not in the report.

lethosor:
Looks like expwnent implemented module mode in add-thought.lua in bf5e491647da9a16126a8bda9c533b4e839e2f4c

library/lua/dfhack.lua
```diff
@@ -380,15 +380,33 @@ end
...
-internal.scriptMtime = internal.scriptMtime or {}
-internal.scriptRun = internal.scriptRun or {}
+Script = defclass(Script)
+function Script:init(path)
+    self.path = path
+    self.mtime = dfhack.filesystem.mtime(path)
+    self._flags = {}
+end
+function Script:get_flags()
+    if self.flags_mtime ~= dfhack.filesystem.mtime(self.path) then
```
expwnent: Does this correctly deal with the case where the user loads or unloads a save which overrides a DFHack script with one modified at the same time? If their filesystem is confused and they installed a mod and DF and DFHack all at once all the files might have the same mtime.

lethosor: It should, since internal.scripts uses script paths as keys (so scripts in different locations should be independent from each other).

lethosor:
I've added an alternative to script_environment() that requires scripts to specify that they support module mode, so the way script_environment() works shouldn't change.

PeridexisErrant:
Does this close issue #548?

lethosor:
Forgot about that issue, thanks!

PR #581, opened 2015-03-21
Add a C++ and Lua JSON library

lethosor:


expwnent:
I'm not opposed to this, but do you have a particular use in mind?

lethosor:
I was thinking it would be a more human-friendly configuration format (for things that should be human-editable, such as plugin/script configuration). The text protobuf format requires the full protobuf runtime library, which we don't build due to issues with reloading plugins that use it, not to mention that it's not as widely-known. I haven't looked at using JSON in any supported plugins (or core), but I was using this in another plugin and it works well.

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "depends/clsocket"]
 	path = depends/clsocket
 	url = git://github.com/DFHack/clsocket.git
+[submodule "depends/jsonxx"]
+	path = depends/jsonxx
+	url = git://github.com/lethosor/jsonxx.git
```
lethosor: Would you want to move this fork over to the DFHack organization?

PR #583, opened 2015-03-22
Plugin exports

lethosor:
This allows plugins to make certain functions available to an object. It currently works much like the Lua side by returning a single object with all functions that a plugin exports available as methods.

### DIFF
--- a/library/PluginManager.cpp
+++ b/library/PluginManager.cpp
@@ -268,6 +268,7 @@ bool Plugin::load(color_ostream &con)
     plugin_enable = (command_result (*)(color_ostream &,bool)) LookupPlugin(plug, "plugin_enable");
     plugin_is_enabled = (bool*) LookupPlugin(plug, "plugin_is_enabled");
     plugin_eval_ruby = (command_result (*)(color_ostream &, const char*)) LookupPlugin(plug, "plugin_eval_ruby");
+    plugin_get_exports = (void* (*)(void)) LookupPlugin(plug, "plugin_get_exports");
     index_lua(plug);
     this->name = *plug_name;
     plugin_lib = plug;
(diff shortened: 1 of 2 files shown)


expwnent:
I'm uncomfortable with the notion of plugins having direct dependencies. I'm concerned about loading, unloading, and reloading plugins. As written, you could grant access to a void function with no arguments, someone could copy down the address of that function in another plugin, and if the first plugin is unloaded that would cause crashes at best.

lethosor:
Right, plugin developers should be careful to avoid that. At the moment, GET_PLUGIN_EXPORTS returns a null pointer if a plugin is not loaded, which should be easy to detect, but I'm unsure how to handle cases where a plugin is unloaded in another thread. Perhaps something similar to CoreSuspender could be used (by preventing a plugin from being unloaded until all references to its "exports" instance are deleted)?
In Lua, calls to functions in unloaded plugins are prevented, but I'm unsure of a way to do this in C++. (For all I know, it could run into the same possible race condition, although it should be rare in both circumstances.)

expwnent:
If something is useful for more than one plugin it should really be moved to a module. For Lua it's ok to be a little more sloppy and let scripters organize things however they want with ``script_environment`` but C++ should be cleaner. I could be convinced but it seems like it would make things more confusing overall.

lethosor:
What inspired this was the fact that TwbT distributes several non-TwbT-related plugins modified to work properly. The changes required (e.g. telling TwbT to draw a specific tile to the map) are simple enough to integrate, and there have been some additional features added to TwbT's versions of those plugins that would be nice to merge as well.
Since there's no guarantee that TwbT will be available, I'm not sure of a cleaner way to accomplish this (a module certainly wouldn't work), but I'd be happy to make it cleaner.

expwnent:
What exactly is the problem with those plugins? Is it that they both want to draw to the screen in the same place?

lethosor:
Drawing to the normal screen buffer when TwbT is drawing to its map buffer breaks rendering (in some cases, no map tiles are visible at all).
In addition to that, there are a couple places in mousequery where the z-coordinate is adjusted to handle multilevel view, and [this](https://github.com/mifki/df-twbt/blob/master/plugins/mousequery.cpp#L648-L658), which I believe corrects the grid coordinates of mouse clicks.

PR #593, opened 2015-03-31
Add view-item-info.lua, content, and documentation

PeridexisErrant:
See readme diff for more information.  A follow-up pull will eventually
come to complete the description coverage.


PeridexisErrant:
It took me a while to work out what the problem was (trailing whitespace), partly because Travis gives no feedback beyond the exit code and `lint.py` only runs under Python 2.  Can either of those issues be fixed?

I've obviously addressed the whitespace problem now, and improved some descriptions while I was there.

lethosor:
Travis does give feedback (click the "[Details](https://travis-ci.org/DFHack/dfhack/builds/57304804)" link). I'm not sure why it's broken on python 3, but I'll try to fix it.

lethosor:
lint.py should work on Python 3 now.

PeridexisErrant:
Is there anything else I need to do before this can be merged?

scripts/view-item-info.lua
```diff
@@ -0,0 +1,380 @@
...
+end
+
+function get_custom_item_desc (item)
+    local desc
+    local ID = get_textid (item)
+    if ID and dfhack.items.getSubtypeCount(df.item_type[ID]) ~= -1 then
+        ID = item.subtype.id end
+    if not ID then return nil end
+    if dfhack.findScript("item-descriptions") then
+        local desc = dfhack.script_environment("item-descriptions").descriptions.ID
```
lethosor: Out of curiosity, did you actually test this? I had to fix a couple of issues with this section for it to work, namely:
* This should be using `descriptions[ID]` instead. `descriptions.ID` is equivalent to `descriptions["ID"]`.
* Declaring `desc` as a local variable here scopes it to the `if` block only - `desc` will be `nil` anywhere besides this line.

Issue #480, opened 2015-01-06
Autolabor could be re-optimized following the labor changes


PeridexisErrant:
Autolabor still works, but runs by a heuristic designed for the old labor system.  I'll let the documentation speak for itself:

>The idea behind this module is to constantly adjust labors so that the right dwarves
are assigned to new tasks. The key is that, for almost all labors, once a dwarf begins
a job it will finish that job even if the associated labor is removed. Thus the
strategy is to frequently decide, for each labor, which dwarves should possibly take
a new job for that labor if it comes in and which shouldn't, and then set the labors
appropriately. The updating should happen as often as can be reasonably done without
causing lag.

>The obvious thing to do is to just set each labor on a single idle dwarf who is best
suited to doing new jobs of that labor. This works in a way, but it leads to a lot
of idle dwarves since only one dwarf will be dispatched for each labor in an update
cycle, and dwarves that finish tasks will wait for the next update before being
dispatched. An improvement is to also set some labors on dwarves that are currently
doing a job, so that they will immediately take a new job when they finish. The
details of which dwarves should have labors set is mostly a heuristic.

>A complication to the above simple scheme is labors that have associated equipment.
Enabling/disabling these labors causes dwarves to change equipment, and disabling
them in the middle of a job may cause the job to be abandoned. Those labors
(mining, hunting, and woodcutting) need to be handled carefully to minimize churn.

The recent changes to labors and job priorities suggest to me that the heuristic should be updated, since dwarves are much smarter about doing the right jobs when several labors are enabled.  I suspect that efficiency could be improved by having more dwarves with a given labor enabled, a less clear distinction between crafting and hauling dwarves, and keeping each dwarves higher-skill labors active (including maybe labors where a dwarf could acquire the skill).  Note however that the specifics are speculation and  should be backed up by some !!science!! before implementation.

PeridexisErrant:
Similar changes are probably appropriate for the `autohauler` plugin, or even integrating them.

PeridexisErrant:
I think this was closed by 684ff433, which only made it into the changelog more recently.

Issue #598, opened 2015-04-10
Is autolabor working in 0.40.24?


FredZeppelin:
I have autolabor enabled, and mining designated for a bunch of squares, but autolabor never turns mining on for my miners. In fact, if I manually enable it, autolabor seems to turn it off...

FredZeppelin:
Unfortunately this makes autolabor pretty much unusable at the current time for obvious reasons. :(

PeridexisErrant:
Closed by #610

PR #610, opened 2015-05-08
Autolabor changes

RossM:
Fix bug in autolabor that caused DISABLED skills to be modified in some circumstances.

Try harder to develop specific skills on dwarves.

Change handling of jobs that use tools.


PeridexisErrant:
This probably closes issue #480, though bigger changes would also be fine if you wanted to address it in detail.  It may be related to #598.  Nice work, either way!

Issue #524, opened 2015-02-01
Unable to set workflow quantity of quarry bush leaves


drakesiard:
Setup: Vanilla df 0.40.24 and dfhack 0.40.24-r1 
To reproduce:
1. Create a "Process plant to bag" task at a Farmer's Workshop.
2. Use the workshop-job plugin to change the material to "quarry bush plant"
3. Try to set a range constraint using the workflow plugin
Note that quarry bush leaves are not in the list of possible outputs.

I assume this was broken by the plant part rewrite.

cdombroski:
This is because workflow does not understand GET_ITEM_DATA_FROM_REAGENT token used in the reaction. No idea if it's fixable or not, but changing the reaction provides a workaround (assuming you only want to process quarry bushes to bags and not any other future processable plants):

    [REACTION:PROCESS_PLANT_TO_BAG]
    	[NAME:process plant to bag]
    	[BUILDING:FARMER:CUSTOM_B]
    	[REAGENT:plant:1:PLANT:NONE:PLANT_MAT:BUSH_QUARRY:STRUCTURAL]
    		[UNROTTEN]
    	[REAGENT:bag:1:BOX:NONE:NONE:NONE]
    		[EMPTY]
    		[BAG]
    		[PRESERVE_REAGENT]
    		[DOES_NOT_DETERMINE_PRODUCT_AMOUNT]
    	[PRODUCT:100:5:PLANT_GROWTH:LEAVES:PLANT_MAT:BUSH_QUARRY:LEAF]
    		[PRODUCT_TO_CONTAINER:bag]
    	[PRODUCT:100:1:SEEDS:NONE:GET_MATERIAL_FROM_REAGENT:plant:SEED_MAT]
    	[SKILL:PROCESSPLANTS]

KlonZK:
more specifically, the item type selection dialog use by workflow is implemented in the materials plugin, which uses dfhack.items.getSubtypeCount() to find the number of subtypes for a given type of item. however the function returns -1 for the type PLANT_GROWTH.

afaik to find the actual subtypes for the plant_growths one has to browse through the complete list of plant raws. if the plant material is known finding the growth is simple, but the item type dialog doesnt know about materials.

Issue #487, opened 2015-01-12
workflow: some resumed jobs are ignored by dwarves


kimitsu:
There seem to be a situation for some jobs: a job that gets resumed (after item constraint check) is ignored by dwarfes - it shows up as resumed in workshop, but never gets active. Manual suspending/resuming does not resolve the issue, using "now!" or moving jobs around in the list does nothing too. Marking workshop for removal and stopping removal, after which jobs get restored in suspended state, and then resuming them manually doesn't seem to reliably solve the issue. Canceling repeat, cancelling the job and then adding the job back manually solves the problem for time being.
Example save (dfhack 0.40.23-r1): http://kimitsudesu.net/files/df/region1.zip
On level -4 there is a clothier's workshop with bunch of resumed jobs that are ignored even though most of the dwarfs have that labor active. Note that there is another clothier's workshop to the left of it. It had the same set of jobs and they were supposed to work in parallel, but in fact only one of them worked and the other got stuck. To isolate the issue I cleared all jobs from the one workshop that worked.

acwatkins:
Also having this issue on dfhack develop branch, with 0.40.24, all over the place.  Removing repeat, canceling, re-adding, and re-repeating the top stuck item seems to fix it temporarily for the entire list.  Just canceling and letting repeat readd it doesn't work.  Removing some other item doesn't fix the issue.  So something seems to be up with the job that would be active itself.

I initially wondered if it was the newish priority code, somehow the unsuspended job has the lowest priority or something  Disabling all other labors and hauling, leaving the dwarf with nothing to do doesn't help.

melkor217:
Setting job field unk_v4020_1 to -1 fixes the job. Or causes game crash sometimes.

acwatkins:
Odd since most items are -1 and that is its init value.  I've investigated some, and the items that are stuck have different values for the unknown 40.20 field.

Below are two jobs that are stuck, note their odd unk_v4020_1 value.

Job 15876: MakeArmor (repeat)
    material: any (cloth)
    item: vest ()
  Input Item 1: cloth; quantity=10000; min_dimension=10000
    flags2: plant
unk_v4020_1: 6
  Constraint ARMOR:ITEM_ARMOR_VEST amount 10 (gap 5)
    items: amount 0; 0 stacks available, 10 in use.
Job 16490: MakeArmor (repeat)
    material: any (yarn)
    item: dress ()
  Input Item 1: cloth; quantity=10000; min_dimension=10000
[...]

Issue #367, opened 2014-10-29
Workflow vs powders


HrabiaVulpes:
DF 0.40.13
dfhack 0.40.13-r1

Because gypsum plaster, sand, and dimple dye are for basic workflow the same thing (and are counted together) I tried to use advanced mode, but it doesn't work. After choosing 'update' nothing happens, no changes are applied. Quern or mill is still counting sand and dimple dye together, and entering advanced parameters shows that they reset.

PeridexisErrant:
Closed by pull #611

Issue #366, opened 2014-10-22
Workflow and plant cloth


MIC132:
When I try to set Range/Limit for my looms using workflow for "cloth of any plant" it shows the amount in stock as 0, no matter how much I have in stock (this causes them to work as on repeat, cause they always think there is not enough). When set to "any cloth" it shows total cloth properly.

dorsey:
I am also having this issue in 0.40.24-r3.

PeridexisErrant:
Closed by pull #611

PR #611, opened 2015-05-08
Workflow changes for plant jobs

RossM:
Changes workflow's handling of built-in plant processing jobs to understand that the results are always plants even if no specific material is set, and that they also produce seeds.


PeridexisErrant:
This and the autolabor fixes are really impressive - and long overdue!

Related issues:   #524, can't process quarry plants to a bag;  #366, workflow doesn't count plant cloth correctly; and possibly #367 and #487.  Would you expect to fix any of these?

RossM:
Issues #366 and #367 should be fixed by this change. #524 isn't because it's a reaction and I didn't manage to fix reactions.

PR #617, opened 2015-05-23
Improved issue with workflow job being resumed but never being worked

### LINKED ISSUE #487: workflow: some resumed jobs are ignored by dwarves

acwatkins:
Fixes #487

* This doesn't fix existing stuck jobs, in order to fix, remove repeat, cancel, add, repeat
* Most workshops worked great after this, however, I noticed my bone bolts and wood bolts still got stuck, not sure if it is the same issue

* The unk_v4020_1 field was not being reset to -1 when resuming the job.
* Updated to be reset only when the job is being resumed
  * Setting it to -1 without checking sets this field on all workflow jobs, which causes a crash
* Made other calls to suspend call set_resumed rather than setting the suspend field

This is the behavior I saw for the unk_v4020_1 field:
Suspended jobs: -1
Jobs not in the top of the list but not suspended: -1
Jobs at the top of the list to work next, not suspended: A positive integer (priority of job?)

### DIFF
--- a/plugins/workflow.cpp
+++ b/plugins/workflow.cpp
@@ -254,7 +254,7 @@ public:
             if (!resume_time)
                 want_resumed = false;
             else if (world->frame_counter >= resume_time)
-                actual_job->flags.bits.suspend = false;
+                set_resumed(true);
         }
     }
 
@@ -262,7 +262,7 @@ public:
     {
         actual_job = job;
         job->flags.bits.repeat = true;
-        job->flags.bits.suspend = true;
+        set_resumed(false);
 
         resume_delay = std::min(DAY_TICKS*MONTH_DAYS, 5*resume_delay/3);
         resume_time = world->frame_counter + resume_delay;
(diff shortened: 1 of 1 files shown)


nsteenv:
A la fdesdça k pour

Issue #615, opened 2015-05-18
Add TwbT structures research, TwbT


PeridexisErrant:
Adding the structures used by Mifki's *Text will be Text* plugin would improve compatibility and ease of updating.  [Here's how to find them.](https://github.com/mifki/df-twbt/blob/master/PATCHES.md)

Ultimately, this could lead to importing TwbT as a submodule, or outright addign the code to DFHack.  Mifki has stated that having to find the structures manually is the only barrier to merging, and he's fine with it.

lethosor:
These won't compile at all since `renderer_twbt.h` is not included and SDL_GetKeyState isn't exposed by DFHack portably (and possibly other issues as well).

Given that these plugins' method of detecting TwbT reads from uninitialized memory if TwbT isn't active, and there's no guarantee that the structure of `renderer_twbt` will remain the same, I'd prefer to see some sort of C++ API exposed by TwbT (see #583, mifki/df-twbt#21).

PeridexisErrant:
Made less urgent by #719, but I'll still open an issue over at df-structures.

PR #618, opened 2015-05-26
Twbt compatible plugins

PeridexisErrant:
TwbT has for some time broken the standard versions of the `automaterial`, `mousequery`, and `resume` plugins.  It thus comes with replacement versions - which I've been using for some time.  Since they also work as normal when TwbT is not present, and don't need to be changed with TwbT versions, it seems crazy to maintain multiple copies of the same code.

For `automaterial` and `resume`, the TwbT-related changes are the only changes - I went through and fixed trailing whitespace, which was the only commit since they were last synchronized.  The TwbT version of `mousequery` additionally supports map dragging, which is a nice feature.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,11 +25,13 @@ DFHack Future
         tweak fps-min fixed
         workflow: Fixed some issues with stuck jobs
             - Note: Existing stuck jobs must be cancelled and re-added
+        automaterial, mousequery, and resume plugins are now compatible with TwbT by default
     Misc Improvements
         dwarfmonitor date format can be modified (see Readme)
         "keybinding list" accepts a context
         nyan: Can now be stopped with dfhack-run
         quicksave: Restricted to fortress mode
+        mousequery:  drag command enables dragging the map with configurable mouse buttons
     Removed
 
 DFHack 0.40.24-r3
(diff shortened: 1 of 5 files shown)


PeridexisErrant:
This is about as much as I can do towards issue #615 - but hopefully making a start will inspire someone to at least get the structures research into standard DFHack.

lethosor:
These won't compile at all since `renderer_twbt.h` is not included and SDL_GetKeyState isn't exposed by DFHack portably (and possibly other issues as well).

Given that these plugins' method of detecting TwbT reads from uninitialized memory if TwbT isn't active, and there's no guarantee that the structure of `renderer_twbt` will remain the same, I'd prefer to see some sort of C++ API exposed by TwbT (see #583, mifki/df-twbt#21).

(Moved from #615)

lethosor:
I'm looking at implementing a hook API (similar to vmethod hooks) instead, which would allow TwbT (and theoretically other plugins) to handle calls to functions that it needs to without modifying everything else that uses them. Since this can't be compiled or merged, I'm going to close it, but I'll look into merging drag support in mousequery (I'm confused as to why it doesn't just use enabler->mouse_lbut and mouse_rbut).

PR #633, opened 2015-06-11
Rebuilt missing half of plant product descriptions

scamtank:
Turns out that *get_plant_reaction_products* is tuned to look for the specific material reaction products in the Natural Balance mod that this was originally designed around instead of looking at the basic item types.
 
The calls for liquid pressing, rendering, soapmaking and cheesemaking work as intended (being handled with MRPs by default), but the brewing stuff overlapped with the similar set of instructions in the next function and the milling/threshing/sugaring(?)/extracting/plastermaking were looking for tokens that just aren't there.
So what I did was comment out the addition of any DRINK_MAT material reaction products into the description and extended the mechanism at the ass-end of *GetFoodPropertiesStringList* to handle basic products other than the present booze, thread and powder, using verbal descriptions with more flair than the same "made into" for everything.

Seeds still need to have their product yields looked up and stuff like quarry bushes don't mention they can be threshed for growths at all, but at least it works without coming up blank or barfing on the user's lap most of the time. It's a start.

### DIFF
--- a/scripts/view-item-info.lua
+++ b/scripts/view-item-info.lua
@@ -249,7 +249,6 @@ end
 
 function get_plant_reaction_products (mat)
     local list = {}
---  add_react_prod (list, mat, "DRINK_MAT", "Used to brew ") -- redundant with the jury-rig in the next function
     add_react_prod (list, mat, "GROWTH_JUICE_PROD", "Pressed into ")
     add_react_prod (list, mat, "PRESS_LIQUID_MAT", "Pressed into ")
     add_react_prod (list, mat, "LIQUID_EXTRACTABLE", "Extractable product: ")
--- a/scripts/view-item-info.lua
+++ b/scripts/view-item-info.lua
@@ -249,7 +249,7 @@ end
 
 function get_plant_reaction_products (mat)
     local list = {}
-    add_react_prod (list, mat, "DRINK_MAT", "Used to brew ")
+--  add_react_prod (list, mat, "DRINK_MAT", "Used to brew ") -- redundant with the jury-rig in the next function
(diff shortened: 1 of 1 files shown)


scripts/view-item-info.lua
```diff
@@ -249,7 +249,7 @@ end
 
 function get_plant_reaction_products (mat)
     local list = {}
-    add_react_prod (list, mat, "DRINK_MAT", "Used to brew ")
+--  add_react_prod (list, mat, "DRINK_MAT", "Used to brew ") -- redundant with the jury-rig in the next function
```
lethosor: Might as well remove this.

scripts/view-item-info.lua
```diff
@@ -290,15 +290,26 @@ function GetFoodPropertiesStringList (item)
...
-                    or string.find (k,"type_seed") or string.find (k,"type_tree") then
+            if v ~= -1 and string.find (k,"type_") and not (string.find (k,"type_basic")
+                    or string.find (k,"type_seed") or string.find (k,"type_tree")) then
                 local targetmat = dfhack.matinfo.decode (v,
                     plant.material_defs["idx_"..string.match (k,"type_(.+)")])
                 local state = "Liquid"
-                if string.find (k,"type_mill") then state = "Powder"
-                elseif string.find (k,"type_thread") then state = "Solid" end
+                local describe = "Made into "
+                if string.find (k,"type_mill")
```
lethosor: `k:find("type_mill")` might be more readable.

scamtank: I trust your judgment. I am certain that there's a better way, but my only priority was to get it going again without breaking something else.

scamtank: I did what as you suggested - handles everything Steamygalleys can throw at it and looks nicer, too.

Issue #630, opened 2015-06-07
Add search functionality to the items shown on tiles inspected by 'k' - would be very helpful for QSPs


pineapplemachine:

lethosor:
By "items" are you suggesting only what DF considers items or everything listed in the [k] menu? The main issue here is that there are many types of objects that can be displayed in this menu:
``` c
// from generated df/ui_look_list.h
  struct DFHACK_EXPORT ui_look_list {
    struct DFHACK_EXPORT T_items {
      enum T_type : int16_t {
        Item,
        Floor,
        Unit,
        Building,
        Vermin,
        Flow,
        Campfire,
        Spatter,
        BuildingItem,
        Fire,
        Water,
        Magma,
        Spoor
      };
    };
  };
```
Units, items, and possibly vermin and buildings are about the only things for which names are easily obtainable, unless DFHack exposes ways to obtain others that I'm not aware of. I'm thinking of either filtering out types of objects that can't be handled or always displaying them (since there are usually few of the others on a single tile) - thoughts?

pineapplemachine:
I was thinking of everything in the menu, though a profusion of literal items was what prompted me to open the issue. Perhaps when filtering, the ones that are difficult to filter might appear indiscriminately at the end of the list?

lethosor:
Added in #640

pineapplemachine:
Awesome. Thanks!

PR #640, opened 2015-06-17
Add look menu search

lethosor:
See #630

### DIFF
--- a/plugins/search.cpp
+++ b/plugins/search.cpp
@@ -5,6 +5,7 @@
 
 #include <VTableInterpose.h>
 
+#include "df/ui_look_list.h"
 #include "df/viewscreen_announcelistst.h"
 #include "df/viewscreen_petst.h"
 #include "df/viewscreen_storesst.h"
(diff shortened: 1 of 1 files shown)


plugins/search.cpp
```diff
@@ -1618,6 +1644,113 @@ IMPLEMENT_HOOKS(df::viewscreen_joblistst, joblist_search);
...
+{
+    typedef df::ui_look_list::T_items::T_type elt_type;
+public:
+    bool can_init(df::viewscreen_dwarfmodest *screen)
+    {
+        if (ui->main.mode == df::ui_sidebar_mode::LookAround)
+        {
+            return look_menu_search_base::can_init(screen);
+        }
+
```
maxthyme: Could that same trick work with adventurer look menu or is that a different can of worms?

lethosor: It would require another hook (this one hooks into `df::viewscreen_dwarfmodest`), but it would probably work. I don't think the search plugin implements anything in adventure mode currently, though, so it's possible that there are other issues.

PR #647, opened 2015-06-25
Add git commit information to libdfhack-version

lethosor:
This runs `git describe --tags` to produce a more descriptive version identifier, like `0.40.24-r3-155-g0fa5570` (commit 0fa5570, 155 commits after 0.40.24-r3). Currently, plugins can be loaded as long as their DFHack version matches core's, but a warning will be displayed if their commits don't match.


library/git-describe.cmake
```diff
@@ -0,0 +1,9 @@
+execute_process(COMMAND git describe --tags
```
lethosor: I'm not entirely sure if git can be located from MSVC in this way.

warmist: cmake might have find_package(git)? maybe?

lethosor: Yep, thanks.

PR #652, opened 2015-07-15
Extended lever.rb

MaienM:
Added names to the lever list, to make the list actually useful if you've given your levers custom names.

`lever pull` without any arguments now pulls the lever under the cursor, so it can be used in keybindings.

### DIFF
--- a/scripts/lever.rb
+++ b/scripts/lever.rb
@@ -32,7 +32,7 @@ def lever_descr(bld, idx=nil)
     # lever description
     descr = ''
     descr << "#{idx}: " if idx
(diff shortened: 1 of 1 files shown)


lethosor:
Thanks!
The Travis build failed because you opened this against the `master` branch, by the way - we keep that pointing to the last release and work towards the next release on the `develop` branch. (It's not possible to default pull requests to that branch, unfortunately - the closest thing we've done is adding a note to Contributing.rst, which causes a banner to be displayed when opening an issue or PR.) I'll merge this commit into the develop branch.

PR #658, opened 2015-07-25
Create gm-unit.lua

warmist:
A gui util to (eventually) user friendly edit any unit.

### DIFF
--- a/scripts/gui/gm-unit.lua
+++ b/scripts/gui/gm-unit.lua
@@ -1,0 +1,170 @@
+-- Interface powered (somewhat user friendly) unit editor.
+local gui = require 'gui'
+local dialog = require 'gui.dialogs'
+local widgets =require 'gui.widgets'
+local guiScript = require 'gui.script'
+local utils = require 'utils'
+local args={...}
+
+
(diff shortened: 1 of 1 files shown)


lethosor:
@warmist are you looking to include this in r4? I'm not sure if you consider it finished enough yet.

warmist:
It's always finished enough. Always in progress. Maybe only wound editor needs to be finished or disabled.
The idea is that it's meant to be modular. So in the long term it would have a module for each of editable things in unit. But even thinking about some stuff makes my head hurt (e.g. appearances?) and will probably never be complete (e.g. changing caste of unit).

Originally i wanted for this to be crowdsourced. But now i see that there is too little people that are interested and/or proficient at this.

maxthyme:
Oh, I'm still intending to figure out how to hack in the appearance stuff, it's one of those things I do regularly enough that I'm going to end up scripting it, it's basically what I started screwing with them for in the first place.

I'm pretty sure the current skills template can be used for an attribute editor actually. Like I said, I've got the idea in my head, just a matter of getting the bug to figure it out.

PR #667, opened 2015-08-07
Add a framework for including 3rd-party script repos

lethosor:
Repos need to include a CMakeLists.txt file with calls to the DFHACK_SCRIPTS macro, which functions similarly to DFHACK_PLUGIN. The `open-legends` script from lethosor/dfhack-scripts is included as an example.

Examples:
```cmake
DFHACK_SCRIPTS(foo.lua)
DFHACK_SCRIPTS(gui/foo.lua SUBDIRECTORY gui)
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "depends/clsocket"]
 	path = depends/clsocket
 	url = git://github.com/DFHack/clsocket.git
+[submodule "scripts/3rdparty/lethosor"]
+	path = scripts/3rdparty/lethosor
+	url = https://github.com/lethosor/dfhack-scripts
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,8 @@ IF(BUILD_PLUGINS)
     add_subdirectory (plugins)
 endif()
 
+add_subdirectory(scripts)
+
 # Packaging with CPack!
 IF(UNIX)
     if(APPLE)
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -377,6 +377,7 @@ install(DIRECTORY ${dfhack_SOURCE_DIR}/scripts
         DESTINATION ${DFHACK_DATA_DESTINATION}
         FILES_MATCHING PATTERN "*.lua"
                        PATTERN "*.rb"
+                       PATTERN "3rdparty" EXCLUDE
         )
 
 install(DIRECTORY ${dfhack_SOURCE_DIR}/patches
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -1,0 +1,2 @@
+include(Scripts.cmake)
+DFHACK_3RDPARTY_SCRIPT_REPO(lethosor)
--- a/scripts/Scripts.cmake
+++ b/scripts/Scripts.cmake
@@ -1,0 +1,19 @@
+include(../plugins/Plugins.cmake)
+
+MACRO(DFHACK_SCRIPTS)
+  PARSE_ARGUMENTS(SCRIPT
+    "SUBDIRECTORY"
+    "SOME_OPT"
+    ${ARGN}
+  )
+  CAR(SCRIPT_SUBDIRECTORY ${SCRIPT_SUBDIRECTORY})
+  install(FILES ${SCRIPT_DEFAULT_ARGS}
+    DESTINATION ${DFHACK_DATA_DESTINATION}/scripts/${SCRIPT_SUBDIRECTORY})
+ENDMACRO()
+
+MACRO(DFHACK_3RDPARTY_SCRIPT_REPO repo_path)
+  if(NOT EXISTS ${dfhack_SOURCE_DIR}/scripts/3rdparty/${repo_path}/CMakeLists.txt)
+    MESSAGE(FATAL_ERROR "Script submodule scripts/3rdparty/${repo_path} does not exist - run `git submodule update`.")
+  endif()
+  add_subdirectory(3rdparty/${repo_path})
+ENDMACRO()


lethosor:
cc @PeridexisErrant 

The only other "dfhack-scripts" repo I'm aware of is https://github.com/dscorbett/dfhack-scripts, which contains one script, but let me know of any others. Gists are harder to work with - it's technically possible to create submodules for a gist, but that's a lot of overhead for a single script.

PeridexisErrant:
The only ones I see come up regularly that *aren't* in the standard distribution - because I've added most of the others - are yours, [Vjek's (on the wiki)](http://dwarffortresswiki.org/index.php/User:Vjek), [Rose's huge collection](http://www.bay12forums.com/smf/index.php?topic=135597), and the [misc. scripts thread](http://www.bay12forums.com/smf/index.php?topic=135506).  Of these, most won't work with this system - there are a variety of way to share scripts, and your approach is pretty rare.  Why not simply add the scripts manually?

lethosor:
I felt that this would make it easier to update multiple scripts at once, but it also makes it harder to avoid including specific versions of specific scripts - I'm not sure which one is more important. Either way, scripts that aren't tracked by git are harder for us to include, since there's no easy way for us to contribute changes upstream (see dfstatus - the number of whitespace and code changes makes it very difficult to merge changes).

PeridexisErrant:
Fair enough.  I assume that in the distributed version all 3rd party scripts behave and are located as if they were included normally?

And as for a plan, how about merging this, contacting Rose and Vjek about setting up a compatible repo (and in the latter case, someone else might make one as a placeholder), and then later merging other stuff manually without worrying about pushing work back upstream?

lethosor:
Yeah, they're installed to the `hack/scripts` folder (or optionally a subdirectory) in CMakeLists.txt in the root of each repo:
```
DFHACK_SCRIPTS(foo.lua)
DFHACK_SCRIPTS(gui/foo.lua SUBDIRECTORY gui)
```
It looks like Roses' scripts are at https://github.com/Pheosics/v24-r3_Scripts, but all of them are in hack/scripts, which makes things a little more complicated:
* Each filename passed to DFHACK_SCRIPTS can be prefixed by hack/scripts
* There can be another CMakeLists.txt in hack/scripts, using filenames in that folder, and the root CMakeLists.txt can call `add_subdirectory(hack/scripts)`
* I can add another option to the DFHACK_SCRIPTS macro (say, `PREFIX`), that can be used to specify a common prefix in the script repo for all of the scripts being added

Vjek's scripts don't seem to be on GitHub, so I'll post a message on their wiki talk page.

PeridexisErrant:
I think option 2 is most elegant - a little more config for odd repos, but still simple for everyone else.  Agreed that authors maintaining would be best, but it often doesn't happen.  Maybe they could just be given push access to the common repo, and encouraged to start their own?

expwnent:
Relevant forum thread: http://www.bay12forums.com/smf/index.php?topic=152920.0

expwnent:
Continuing discussion from #679 

Another issue is that we'd have to fork every repo if we want to make changes to the scripts, and that could run into issues if more than one person calls their script repo "scripts". Is it possible to fork and rename with github?

lethosor:
Yeah, that probably won't work - you could fork and set up remotes with different repo names manually with git, but GitHub wouldn't recognize our repos as forks, which would prevent pull requests.

An alternative could be to create repos for individual script authors under this organization (e.g. dfhack/lethosor or dfhack/lethosor-scripts), which would allow authors and DFHack maintainers to work on them more easily (and avoid the need for frequent pull requests). I don't know if it's currently possible to grant individuals access to specific repos without them being on a team, but it might be with the new organization permissions system.

Of course, there's still the possibility of manually including scripts, which is easy for us to work with but harder to contribute changes back to non-git upstreams.

expwnent:
I think renaming repos to username-scripts is a good way of doing it, at least for now. We can post issues on github as requests for manual pulls. It's not perfect but it'll do.

Eventually if one user has multiple sets of unrelated scripts and it makes sense to have them in separate repos we can make an exception and name the repos something like bob-cheat-scripts and bob-civilization-scripts.

We can manually include individual scripts from the forum when they prove useful enough to be included. Forum scripts usually have an initial version then maybe a few bugfixes, then end up stable or abandoned, so it makes sense to just manually include them instead of setting up a whole repo for one script.

expwnent:
One more possible issue: I'd like to have an automatic way of testing whether 3rd party script names conflict with either each other or scripts in the main repo. Can travis test things across a submodule boundary?

lethosor:
Yeah, Travis initializes submodules (that's why I had to fix whitespace in one of my scripts to get that test to pass above).

expwnent:
I'm not sure if I understand the CMake file: does this support script subfolders or does it only install to the main script directory?

lethosor:
Yeah, it supports subdirectories with the SUBDIRECTORY argument - added examples to the first post. It currently requires you to specify the source and destination directory since that could vary between the core and individual repos (e.g. scripts in the root folder of a repo might be better suited to the `modtools` folder here), although I could simplify that.

PeridexisErrant:
I still feel that this is a technical solution to a policy problem, and still doesn't address some common issues - particularly documentation.

As an example of an alternative structure, we could for example move *all* scripts to a single `DFHack/DFHack-scripts` repo, and allow write access to anyone making a substantial contribution.  That then becomes the standard place to put scripts, there's just one separate but substantial readme document, and the whole scripts folder can be a submodule.

>  [this is] a convenient approach for me and others (Max, Roses, others?) that store scripts on GitHub, but less so for people that store them on the forums or DFFD. Of course, it would be good to encourage people to use git so they can incorporate changes made by others (especially in this repo) - dfstatus, for example, has gone through a lot of cleanup and had some features added, which I don't think the original author has picked up on.

For those not using Github, and some like myself who do, there often isn't any meaningful upstream to pull changes back to.  Development of an unstable prototype (or addition of a feature) happens wherever, and then when it's stable I put it in the big DFHack repo and that's the canonical version.  Any code on the forums etc is just a development artefact.

I actually think the solution in this pull is a good one, but it only applies to script maintainers active on github - and that's a minority among the authors of scripts not yet included, even if they tend to be more prolific.  The other side probably just requires someone with more experience to just add scripts when they come up, as I've been doing for a while.

expwnent:
I definitely agree documentation is a separate and important issue. I'm not sure what the right answer is for documentation of 3rd party scripts.

I don't think it's a good idea to have widely open write access. Most scripts are one-offs highly specific to a certain problem. Those certainly don't belong in the main repo. Scripts in the main repo should be reusable and have clear, accurate, and literal names. All widely inclusive threads for organizing user scripts that I've seen end up having so many scripts that it's hard to find the one you're looking for, and most of the scripts they contain will not be useful to most users. I don't want to clutter the main repo with so many scripts that it's hard for users to find the one they need.

I somewhat like the idea of a "supplemental" script library which has a wider range, but who would organize it, and more importantly, maintain it? A lot of the time people will write one script and then more or less forget about it, and if a new major DF version comes along and breaks it they'll be nowhere to be found, and the number of users of highly-specific scripts is small so it's unlikely to get reported to someone who can fix it and it makes it unclear to users whether the script is broken or whether they're just using it incorrectly. Heck, even I've abandoned scripts before and exactly that happened. Point is, there are a lot more people with enough time to write a script than there are people who have enough time to write it and maintain it for future versions.

PR #668, opened 2015-08-07
New plugin: fix-unit-occupancy

lethosor:
Fixes [bug 3499](http://bay12games.com/dwarves/mantisbt/view.php?id=3499)


lethosor:
Not sure if this ought to be enabled by default, but I'll merge it since it currently isn't.

Issue #412, opened 2014-11-30
ls does not list scripts outside of hack/scripts


expwnent:
It should do that and maybe also note when that script overrides a script in hack/scripts. The fix will involve Core.cpp if someone gets to it before I do.

PeridexisErrant:
Just to make the implementation a little clearer:  I'd propose a new section at the end of `ls` output, for scripts in `raw/scripts` (and other searched locations).  Then any script in `hack/scripts` which is overridden could be listed in red text.  I think that would be enough information - there's not really any room for timestamps or diffs, so those could be a separate command.

PR #673, opened 2015-08-14
PluginManager improvements

### LINKED ISSUE #666: Plugin system-related issues
- [x] `plug` only lists plugins with commands
- [x] The name passed to `DFHACK_PLUGIN` does not have to match the plugin filename (RemoteFortressReader vs. remotefortressreader, for example), allowing for possible name collisions
- [x] Plugins that don't exist on startup can't be loaded later
- [x] Plugins that are deleted after startup produce an unhelpful error message when reloaded (different from the message for other nonexistent plugins)
- [x] `ls` gives no feedback for unloaded plugins
- [x] `load`/`unload`/`reload` give no feedback for broken plugins

lethosor:
* load/unload/reload are no longer restricted to plugins that exist
  on startup
* Names passed to DFHACK_PLUGIN must match the plugin's filename
  (see remotefortressreader vs RemoteFortressReader, counters vs probe)
* `plug` output lists all plugins and state/command information
* Deleted plugins can be reloaded again if they are replaced
* load/unload/reload don't fail silently with broken plugins
* Built-in commands are recognized internally (e.g. "help help"
  does not display "help is not a recognized command"), although help
  for them is not yet implemented
* New command: `type` (bash-like) - shows where/how a command is
  implemented
* `plug` can accept multiple plugin names
* `ls` displays more information about unloaded/unrecognized plugins
* `load all` changed to `load -all` (or `load --all`, `load -a`, ...)

Closes #666

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -386,12 +386,12 @@ static bool try_autocomplete(color_ostream &con, const std::string &first, std::
     std::vector<std::string> possible;
 
     auto plug_mgr = Core::getInstance().getPluginManager();
-    for(size_t i = 0; i < plug_mgr->size(); i++)
+    for (auto it = plug_mgr->begin(); it != plug_mgr->end(); ++it)
     {
-        const Plugin * plug = (plug_mgr->operator[](i));
+        const Plugin * plug = it->second;
         for (size_t j = 0; j < plug->size(); j++)
         {
-            const PluginCommand &pcmd = plug->operator[](j);
+            const PluginCommand &pcmd = (*plug)[j];
             if (pcmd.isHotkeyCommand())
                 continue;
             if (pcmd.name.substr(0, first.size()) == first)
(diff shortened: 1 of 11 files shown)


PeridexisErrant:
Does this make fixing  #412 easier?  It sounds somewhat similar, to me.

lethosor:
This really only affects plugins, and I didn't change how `ls` with no arguments works (just the `ls PLUGIN` form).

library/Core.cpp
```diff
@@ -485,6 +485,43 @@ static std::string sc_event_name (state_change_event id) {
...
+    if (cmd == "ls" ||
+        cmd == "help" ||
+        cmd == "type" ||
+        cmd == "load" ||
+        cmd == "unload" ||
+        cmd == "reload" ||
+        cmd == "enable" ||
+        cmd == "disable" ||
+        cmd == "plug" ||
+        cmd == "type" ||
```
PeridexisErrant: "type" is duplicated here.

PR #674, opened 2015-08-15
Documention fixes

PeridexisErrant:
This ensures that all scripts in the general and `fix/*` directories are documented in the readme, and the existence of `devel/*` is noted.

GUI scripts are less clear, and several scripts could be moved to `gui/*`.  `dfstatus` appears to be duplicated by `gui/dfstatus`; I left this alone.  I haven't looked at plugins or builtin commands.  Some scripts that I documented also needed an initial comment, which I added in cases I noticed but didn't check for.

`exportlegends` is now compatible with `open-legends`.

### DIFF
--- a/Readme.rst
+++ b/Readme.rst
@@ -645,6 +645,39 @@ Options:
 :show X:       Marks the selected map feature as discovered.
 :hide X:       Marks the selected map feature as undiscovered.
 
+fortplan
+--------
+Usage: fortplan [filename]
+
+Designates furniture for building according to a .csv file with
+quickfort-style syntax. Companion to digfort.
+
+The first line of the file must contain the following::
+
+   #build start(X; Y; <start location description>)
+
+...where X and Y are the offset from the top-left corner of the file's area
+where the in-game cursor should be located, and <start location description>
+is an optional description of where that is. You may also leave a description
+of the contents of the file itself following the closing parenthesis on the
+same line.
+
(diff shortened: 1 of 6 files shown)


Readme.rst
```diff
@@ -2370,6 +2449,16 @@ Examples::
...
     create-items bar adamantine
 
+deathcause
+==========
+Focus a body part ingame, and this script will display the cause of death of
+the creature.
+Also works when selecting units from the (``u``) unitlist viewscreen.
+
+dfstatus
+========
```
lethosor: There isn't a `dfstatus` script in the scripts folder in the master or develop branches (only `gui/dfstatus`).

scripts/unsuspend.rb
```diff
@@ -1,3 +1,5 @@
+# un-suspend all jobs, one time only
```
lethosor: Only building construction jobs, actually.

lethosor:
Is this ready to be merged?

PeridexisErrant:
**Yes, it's ready to merge**, though it doesn't completely resolve the associated issue - `gui/*` scripts, plugins, and builtin commands still need to be checked.

I'd be happy to handle the gui scripts too, but that should probably be a separate pull request as it would involve substantial changes to the structure of the readme and I'd want some more review.  Plugin documentation is similarly scattered, and could be (mostly, I'm not fluent in cpp) handled at the same time.  Someone else would have to handle builtin commands.

PR #675, opened 2015-08-15
Luasocket

warmist:
Add a way to use sockets in lua.


plugins/lua/luasocket.lua
```diff
@@ -0,0 +1,67 @@
...
+end
+
+
+local server=defclass(server,socket)
+function server:accept()
+    local id=_funcs.lua_server_accept(self.server_id,false)
+    if id~=nil then
+        return client{server_id=self.server_id,client_id=id}
+    else
+        return
```
lethosor: Travis is complaining about the space after `return` here.

Issue #670, opened 2015-08-12
Improve project structure; ensure documentation is clear and complete


PeridexisErrant:
~~The `startdwarf` script is completely undocumented, as is the `hide` command.~~  *Fixed by #671*

Where there's two, there's probably more - it would be good if someone checked that each script, plugin, or command is at least mentioned in the readme to not it's existence, and preferably include some basic information.

This issue has evolved into part of a larger discussion of DFHack structure.  See the thread of comments and forum links in #667 for more information.

lethosor:
A lot of internal commands aren't listed, so I'd like to keep this open.

PeridexisErrant:
It occurs to me that #667, including 3rd party scripts, is going to cause some documentation headaches too.  We probably want to include a link to their documentation in the main readme, if not the full description of each script.

PeridexisErrant:
...counting scripts shows 34 of 54 documented.  At this point, I'm thinking we might want to generate the documentation from the scripts (and make missing docs in the script a linting error).

The least-case would be to require a readme entry and current NEWS entry for any plugin or script modified by a pull request.

(I still plan to make another pull request with scripts and plugins documented, FWIW)

PeridexisErrant:
Follow #674, all scripts in the general and fix/* directories are documented in the readme, and the existence of `devel/*` is noted.  It doesn't completely resolve the associated issue - `gui/*` scripts, plugins, and builtin commands still need to be checked.  Someone should also decide to what extent the `modtools/*` should be documented.

I'd be happy to handle the gui scripts in a separate pull request as it would involve substantial changes to the structure of the readme and I'd want some more review. Plugin documentation is similarly scattered, and could be (mostly, I'm not fluent in cpp) handled at the same time. Someone else would have to handle builtin commands.

expwnent:
Manually checking each and every script/plugin would take a long time and it's monotonous enough that we're likely to make mistakes if we do it all at once and if we don't do it all at once it's easy to miss things. Maybe we should include a temporary "documentation confirmation todo list" file and manually check at least a handful of them every so often and cross them off as we do so it's easier to make and share progress.

The current system is basically just waiting for someone to complain about a hole in the docs then filling it, which is a bad way of doing things.

PeridexisErrant:
[...]

PR #679, opened 2015-09-04
Import Vjek's scripts from the wiki

PeridexisErrant:
[*Source on the wiki*](http://dwarffortresswiki.org/index.php/User:Vjek)

Vjek's scripts are pretty much classic at this point, and there's a reasonably wide user base.  However installation can be tricky - at the very least, more so than coming included - and not being on Github Lethosor's proposed inclusion system won't work.  Given that upstream changes are very slow now the scripts are mature, I suggest merging them entirely.

Currently (two commits) all I've done is import the .lua files and added appropriate documentation.  It is likely that style or substantive changes might be needed (and I've heard of fourth-party-updates), so this pull request is for discussion rather than immediate merging.  Thoughts?

### DIFF
--- a/Contributors.rst
+++ b/Contributors.rst
@@ -81,6 +81,7 @@ acwatkins               acwatkins
 Wes Malone              wesQ3
 Michon van Dooren       MaienM
 Seth Woodworth          sethwoodworth
+Vjek
 ======================= ====================    ===========================
 
 And these are the cool people who made **Stonesense**.
--- a/NEWS
+++ b/NEWS
@@ -28,6 +28,8 @@ DFHack Future
         view-item-info:  adds information and customisable descriptions to item viewscreens
         warn-starving:  check for starving, thirsty, or very drowsy units and pause with warning if any are found
         points:  set number of points available at embark screen
+        Vjek's script collection: armoks-blessing, brainwash, elevate-mental,
+            elevate-physical, make-legendary, pref0adjust, rejuvenate
     New tweaks
         embark-profile-name: Allows the use of lowercase letters when saving embark profiles
         kitchen-keys: Fixes DF kitchen meal keybindings
--- a/Readme.rst
+++ b/Readme.rst
@@ -2400,7 +2400,7 @@ or the gui invoked by running ``add-thought gui`` with a unit selected.
 
 armoks-blessing
 ===============
-Runs the equivalent of rejuvenate, elevate_physical, elevate_mental, and brainwash
+Runs the equivalent of rejuvenate, elevate-physical, elevate-mental, and brainwash
 on all dwarves currently on the map.  This is an extreme change, which sets every
 stat to an ideal - legendary skills, great traits, and easy-to-satisfy preferences.
 Use in moderation; it's as 'cheaty' as DFHack gets.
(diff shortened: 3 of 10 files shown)


expwnent:
I'll just ramble off my thoughts.

My biggest hesitation is naming conventions. There are old scripts in the repo that don't follow this convention I haven't gotten around to fixing, but in general scripts should be named as simply and descriptively as possible. elevate_physical is fine, but someone who hasn't seen it before would have no idea what armoks_blessing does and if they happen to be looking for a script which does what that script does, the name armoks_blessing isn't going to be what they're looking for if they browse the scripts folder. At some point we should really try to standardize camelCase vs underscores_between_words vs dashes-between-words but that's a much bigger project because df-structures are pretty chaotic that way.

Vjek's scripts are intended for users more than modders so I suppose they should go in the main scripts folder so people don't have to type out long subfolders every time (modders can just copy+paste so it's less of an issue), but we don't want to clutter the main folder too much and make it hard to find things but then again that's sort of a pre-existing problem but then again it's something we/I should really fix at some point.

At some point we'll have to start managing scripts and subfolders in a smarter way as the number of scripts increases. There's certainly a fundamental difference between scripts intended for modders and scripts intended for players. Players shouldn't be expected to have as high of a DFHack skill level as modders so they should be easier to use. "Cheat" scripts like these (and I use the term in a non-judgmental way) are more for players but can also be useful for testing purposes.

While we're at it, I also support including Roses' script system as soon as he gives the OK and adds one or two features I already discussed with him, and I'm not sure why @lethosor hasn't merged in his custom script repo. Are there any other script libraries we should consider including?

Backwards compatibility is relevant. I strongly support standardizing scripts, especially the way they process arguments. Not all of the scripts already in the repo use the new processArgs system (but they should). The benefit of standardization is that it's easier for players and modders to learn / memorize how to use new scripts. The drawback is breaking backwards compatibility for 3rd party scripts. In general I think backwards compatibility is good, but if there's a good reason then we should do it the better way. It would be inconvenient for modders if we incorporated a 3rd party script library and changed the syntax to match the new standard but in the long run it's better for things to be standardized.

It's not as big of a problem as it seems because modders can "override" DFHack scripts by putting scripts in the raw/scripts folder, but still.

expwnent:
Relevant thread: #667

expwnent:
Relevant forum thread: http://www.bay12forums.com/smf/index.php?topic=152920.0

lethosor:
PeridexisErrant actually renamed all of the scripts to use dashes instead of underscores - there's one place in the readme that still refers to elevate_physical, though.

I was holding off on #667 until there was more of a consensus on it - it's a convenient approach for me and others (Max, Roses, others?) that store scripts on GitHub, but less so for people that store them on the forums or DFFD. Of course, it would be good to encourage people to use git so they can incorporate changes made by others (especially in this repo) - dfstatus, for example, has gone through a lot of cleanup and had some features added, which I don't think the original author has picked up on.

PeridexisErrant:
I'm happy to change the names of the scripts if you think that's important - and as a general principle I agree (and changed to dash-words for this reason.  However in this case I think maintaining a clear link back upstream justifies the somewhat odd choice.

That said, I think this is ready to merge and we can address the bigger issues as part of a longer-term response to #670.

PR #683, opened 2015-09-09
Advfort

warmist:
Fixes and improvements to advfort.
~~That bot will probably complain, but i'll fix it (probably)...~~


warmist:
Build fails with: ./plugins/proto/RemoteFortressReader.proto: Contains tabs: line 185
My pull request does not touch it so not my problem, :PLZ PULL K THX BYE EMOTICON:

PR #693, opened 2015-09-22
Experiments with Sphinx for generating documentation

PeridexisErrant:
[Sphinx](http://sphinx-doc.org) is a set of documentation tools first designed for the Python language documentation, and later generalised.  Concrete benefits of switching right now include better HTML generation, and a cleaner main directory - but it's not done yet.

In the near future, this will enable significant cleanup and restructuring of the documentation.  For example, it should be possible to arrange documentation by type (eg scripts, plugins, builtins), and then pull parts of each onto a page for a particular kind of use (eg bugfixes, adv. mode, etc).

Eventually (ie in a future pull), the autodoc functionality may pull documentation for each script or plugin from it's source code.  This helps with a number of issues from stale documentation, to ensuring completeness (via linting scripts / travis), to scripts being pulled in from third-party repositories.

This may close #670, is related to #677, and closes #597.

To build the docs, you need Python and Sphinx (`pip install sphinx`); then just run `make html` in the main DFHack directory and the docs will be generated.  You can read them by opening `./README.html` in your preferred browser.

PR #697, opened 2015-09-24
Documentation formatting, eg internal links

PeridexisErrant:
This is a follow-on pull from #693, which overhauled the organisation and build system for DFHack documentation.

This one:

* Adds internal hyperlinks wherever I found a reference; there's certainly more but this is a good start
* Fixes line length in many places for readability as raw text
* Adds basic formatting to LICENSE , without changing the overall structure or any substantive content (ie just markup)
* Added draft standards for documentation
* Started standardising markup for eg plugin arguments, monospace font
* Adds df-structures documentation
* Adds 3rdparty script documentation (via globbing, so it's automatic if more repos are added)

It's a smaller changeset than the last one, and represents the low-hanging polish that Sphinx lets us add to the docs.  Further work will take some more effort, but there's an example now too.

The changes are again visible at http://peridexiserrant.neocities.org


PeridexisErrant:
* Added many more internal links
* Checked for undocumented plugins, and added them
* Formatting is reasonably consistent
* Generated HTML uses the DFHack icon, because what else is it for?

Things for someone else to do:

* @expwnent is sorting out the build process and getting DFHack onto readthedocs.org (sorry if this pull makes your life harder), which will make everything viewable online.  We should put a link to that in the root README.rst for users who turn up to GitHub.

* We need some coherent way of categorising plugins.  This should probably be looked at by someone fluent in cpp, at about the same time as Sphinx autodoc is considered - nobody wants to do that work twice.  Tying togther the help system, `ls`, and autodoc is probably a good idea to reduce duplication.

* The lack of autodoc support for ruby and lua [can be worked around](https://github.com/PeridexisErrant/dfhack/commit/8306fc07e776848ca602306f796c9e65bfd24dab).  However we probably want to tie this into the builtin``help commandname``, possibly by having it print a multiline string in the file with a special name (between the identifying points, etc).  I'm happy to do the grunt work on this if someone sets up the general system.

PeridexisErrant:
Rebased onto develop to get the newer build system and third party scripts etc.

I then discovered that my perl install had broken, fixed that, discovered other problems, fixed the VS2010 install, registered, product key, still not working.  I have no idea what's going wrong, but it shouldn't be this difficult.

A `docs/build-docs.[sh|bat]` script might be useful for many users, who want to rebuild the documentation but don't need to compile everything.  It's pretty common for eg script authors, and removes a couple of big dependencies for them.

expwnent:
Is there a reason you rebased instead of merged?

PeridexisErrant:
Nothing beyond recently having had to rebase branches for work - not a good reason, I just ran ``git pull --rebase`` without thinking much about it.  I could probably reconstruct the history in some other way if you'd prefer.

expwnent:
I would prefer it without the rebase if it's not too much work.

PeridexisErrant:
Yep, can do.  Other than that I'm done with this pull request.

expwnent:
Sorry if I misunderstood: are you going to go back and do it all without the rebase?

expwnent:
I think you can start with a fresh branch that's pointing at dfhack/develop and rebase onto that. It should just be a few commands if you do it right.

PeridexisErrant:
Alright, everything's clean and ready at the fresh [documentation-updates](https://github.com/PeridexisErrant/dfhack/tree/documentation-updates) branch.

The only tiny niggle is that in the movement of `conf.py` (or something related), Sphinx for some reason can't find `dfhack-icon.ico` to use as the html favicon.  Simply adjusting the path in `conf.py.in` didn't work, so I'll leave that to you.

expwnent:
Thanks!

PR #704, opened 2015-09-27
Develop

expwnent:


expwnent:
The check is wrong: it is included.

PR #708, opened 2015-10-07
Fix typographical error(s)

orthographic-pedant:
@DFHack, I've corrected a typographical error in the documentation of the [dfhack](https://github.com/DFHack/dfhack) project. Specifically, I've changed confortable to comfortable. You should be able to merge this pull request automatically. However, if this was intentional or if you enjoy living in linguistic squalor, please let me know and [create an issue](https://github.com/thoppe/orthographic-pedant/issues/new) on my home repository.

### DIFF
--- a/Readme.html
+++ b/Readme.html
@@ -3610,7 +3610,7 @@ It re-uses the keybindings from the track stop building interface:</p>
 <div class="section" id="gui-advfort">
 <h2><a class="toc-backref" href="#id173">gui/advfort</a></h2>
 <p>This script allows to perform jobs in adventure mode. For more complete help
-press '?' while script is running. It's most confortable to use this as a
+press '?' while script is running. It's most comfortable to use this as a
 keybinding. (e.g. keybinding set Ctrl-T gui/advfort). Possible arguments:</p>
 <ul class="simple">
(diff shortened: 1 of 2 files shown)


thoppe:
I'm honored my semi-autonomous bot was able to contribute in some small way to such an important project. DFHack and other tools like it helped me understand what Fun really is. 

Out of curiosity, why not just merge the change directly?

lethosor:
We keep the current release in the `master` branch and active development in `develop`, and the documentation setup has changed enough there that merging this directly isn't possible (I forgot to attribute the change like I did in #707, so sorry about that).

I'm not sure of a good way to convey this information to tools like yours, though - if we make the develop branch default, PRs would default to it, but so would all URLs without an explicit branch (e.g. https://github.com/DFHack/dfhack), and that's what we've always referred people to for the latest release's information.

thoppe:
Hmmmm, it seems like this is hard problem to generalize to other projects. My logic in picking the default branch was to correct the documentation most _users_ would see -- even if develop branches are active, they might not be update to date, or they might be in flux. Case in point, [this](https://github.com/DFHack/dfhack/blob/develop/README.rst) looks like it isn't something that should be shown to the world.

Let me know if you have any ideas though!

PR #710, opened 2015-10-12
Fixing assume identities since it causes a crash.

Mason11987:
Fixing assume identities since it causes a crash.


PeridexisErrant:
The build is failing because there are tabs in the file; replace them with 4 spaces and it'll be fine.

It would be great if you added the fix for open-legends compatibility too.

Mason11987:
Thanks, here's the latest file which should be good.

https://github.com/Mason11987/dfhack/blob/patch-1/scripts/exportlegends.lua

PR #715, opened 2015-10-26
Add 'undisturbed' flag to job item checks

DoctorVanGogh:
This change allows a _somewhat_ more manageable filtering for 'Collect Web' material selections. Without it, every body part of every creature is available, with this addition in place it's a manageable three pages, all of actually valid items.

### DIFF
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -307,7 +307,7 @@ bool ItemTypeInfo::matches(const df::job_item &item, MaterialInfo *mat, bool ski
     RQ(1,extract_bearing_plant); RQ(1,extract_bearing_fish); RQ(1,extract_bearing_vermin);
     RQ(1,processable_to_vial); RQ(1,processable_to_bag); RQ(1,processable_to_barrel);
     RQ(1,solid); RQ(1,tameable_vermin); RQ(1,sand_bearing); RQ(1,milk); RQ(1,milkable);
-    RQ(1,not_bin); RQ(1,lye_bearing);
+    RQ(1,not_bin); RQ(1,lye_bearing); RQ(1, undisturbed);
 
     RQ(2,dye); RQ(2,dyeable); RQ(2,dyed); RQ(2,glass_making); RQ(2,screw);
     RQ(2,building_material); RQ(2,fire_safe); RQ(2,magma_safe);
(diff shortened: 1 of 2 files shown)


DoctorVanGogh:
It shouldn't ;)

From my understanding the [modified] code is only invoked on material filtering - so if there is any error it would immediately show in the dialogs. I just checked all 'weave' tasks, and they all show the expected lists.

Addendum:
I ran the game with a debugger attached, none of my breakpoints (inside the modified methods) were hit during job execution, only the CTRL-A material selection dialog hit the beakpoints.
Things _might_ behave different on a non windows machine (can only check this on my local machine), but I wouldn't expect it.


I have to say though: _The whole job item to material matching seems fishy to me_ :confused:
That might entirely be because I don't fully grasp the matching logic yet - but somehow somthing is missing in there. I can see tons of 'is this really milkable?' or similar checks for a multitude of flags, but I can't see any basic
> Job needs item of type X - Is this material of type X/does it have a reaction product of type X

check.

I think this is one of the reasons for the mile long selection list for silk or plant threads...

-
Back my fix: My commit is somewhat 'hacky' :flushed: - basically I'm equating the 'undisturbed' flag with 'is this silk?'. This might told true for now (since webs can only be made of silk, and only they can be undisturbed), but may change in the future, so it's no 100% clean solution. Apart from that I think it's good.

warmist:
This code (to my understanding) is not called by vanilla "weaving job". I mean if someone would ask dfhack to match this job that requires silk it will fail (e.g. gui/advfort uses this to some degree, don't remember how it would react to this.)
Imho it would be better to add the hack to the workshop gui or correctly implement this by adding a job type (e.g. if you want collect silk job you mean undisturbed silk items).

lethosor:
What list are you referring to? Something in gui/workflow?

DoctorVanGogh:
Query workshop with jobs, select job <kbd>CTRL</kbd>-<kbd>a</kbd>, <kbd>m</kbd>

If you do this for weave jobs and drill down to 'creatures' the 'default' implementation offers _all parts_ for _all creatures_.
With this change at least the 'collect webs' list does down to a manageable 3 pages of types of silk or so.

And from my understanding it's 'gui/workflow', which uses 'gui/materials', which uses the changed c++ method ;)

lethosor:
We're planning a release soon (hopefully!), so I'll merge this - I'm not too familiar with how DFHack works with materials, but it sounds like the improved methods you're suggesting would take a while to implement (which is fine, but it would be better to have this change in the next release than nothing).

PR #713, opened 2015-10-19
More docs improvements

PeridexisErrant:
Now ready to merge!  Closes #670.  Closes #703.  Includes docs-in-scripts, linting, improved organisation (including automatic sorting of entries), and much more.

You can see the [docs here](http://peridexiserrant.neocities.org), and I've got  https://dfhack.rtfd.org ready to start serving the same as soon as this pull is merged - with per-branch, per-tag docs.  If someone [enables the webhook](https://read-the-docs.readthedocs.org/en/latest/webhooks.html) (takes one minute) it'll automatically update when commits are merged.

Also after this pull is merged, I'll make pulls against the 3rdparty scripts repos to put docs in those scripts.  The globbing and inclusion of their README files is a perfectly good solution for the time being though.

PR #717, opened 2015-10-28
State of decay - Onmiclasm's FPS boosting scripts

PeridexisErrant:
[Bay12 forums thread](http://www.bay12forums.com/smf/index.php?topic=150782).  The `deteriorate*.rb` scripts could probably be merged, but I'm terrible with Ruby.

### DIFF
--- a/scripts/deteriorateclothes.rb
+++ b/scripts/deteriorateclothes.rb
@@ -1,0 +1,81 @@
+# Increase the rate at which clothes wear out
+=begin
+
+deteriorateclothes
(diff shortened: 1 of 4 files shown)


lethosor:
I assume this builds with #713, yes?

PeridexisErrant:
Yep, all documented in the scripts.

lethosor:
These should be added to NEWS, but doing that here will conflict with #713. Can you add it to NEWS.rst on that branch?

PR #719, opened 2015-10-31
Allow loading (external) plugins compiled against r3

DoctorVanGogh:
Commit 4fc6cb6f175aebc96bf7ec4629a90cd05b71e8ee broke binary compatibility with (third party) plugins compiled against r3 (or any develoment build from before this commit).

Specifically the renaming of `pluginmanager.h`'s `DFHACK_PLUGIN_AUX` fields `name` and `version` to `plugin_name` & `plugin_version` in https://github.com/DFHack/dfhack/commit/4fc6cb6f175aebc96bf7ec4629a90cd05b71e8ee#diff-bd71300c004474c4053e54ada7483b31L290 and corresponding changes to  `pluginmanager.cpp`'s `Plugin::load`at https://github.com/DFHack/dfhack/commit/4fc6cb6f175aebc96bf7ec4629a90cd05b71e8ee#diff-bcb993821b53968bc228a308c6193963L229 cause errors on load for 'old' plugins. 

One **popular** example would be *twbt* from LNP ;). Last available version is compiled against r3, which means it only has fields _without_ the new `plugin_` prefix, which means it won't load on anything compiled after 4fc6cb6f175aebc96bf7ec4629a90cd05b71e8ee :(

My change would allow a (silent) fallback to the old field names to allow using those plugins. I've been running this for serveral days and have not seen any adverse effects.

A _good_ implementation would probably put the whole 'allow r3 plugins' switch inside some compilation (symbol) check, a _great_ one would put it inside some configuration setting. I don't trust my C++ enough to write the corresponding macros for the first, and I don't know the config system well enough - i.e. at all -  for the second. And since you might even intentionally force people to only use 'matching' binaries, I'm not going to try this myself.

But in case you want to keep backwards compatibility, this should do it.

### DIFF
--- a/library/PluginManager.cpp
+++ b/library/PluginManager.cpp
@@ -280,14 +280,14 @@ bool Plugin::load(color_ostream &con)
             return false; \
         }
 
-    plugin_check_symbols("plugin_name", "name")					// allow r3 plugins		
-    plugin_check_symbols("plugin_version", "version")			// allow r3 plugins	
+    plugin_check_symbols("plugin_name", "name")                 // allow r3 plugins		
+    plugin_check_symbols("plugin_version", "version")           // allow r3 plugins	
     plugin_check_symbol("plugin_self")
     plugin_check_symbol("plugin_init")
     plugin_check_symbol("plugin_globals")
     const char ** plug_name =(const char ** ) LookupPlugin(plug, "plugin_name");
-	if (!plug_name)												// allow r3 plugin naming	
-		plug_name = (const char ** )LookupPlugin(plug, "name");
+    if (!plug_name)                                            // allow r3 plugin naming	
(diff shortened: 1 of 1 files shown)


lethosor:
Travis builds are failing currently (because some scripts have been added without being added to the readme, due to #713), but these messages from lint.py are relevant:
```
./library/PluginManager.cpp: Contains trailing whitespace: lines 283-284, 289
./library/PluginManager.cpp: Contains tabs: lines 283-284, 289
```

Anyway, I would prefer to have this set up in a configuration file, but DFHack doesn't have a "core" configuration file yet (there are a few poorly-documented environment variables like `DFHACK_NO_CONSOLE` and `DFHACK_NO_DEV_PLUGINS`, but this is something that should be more persistent). This'll work for now.

Issue #721, opened 2015-11-02
Automaterials: Will build constructed floor over existing constructed floor with 'box select'


DoctorVanGogh:
I noticed a bug in automaterials:

In 'default' mode the game does not allow constructing floor over already constructed floor - you get an 'invalid' indicator and a 'Construction present' error:

![floor-on-floor-default](https://cloud.githubusercontent.com/assets/324067/10872193/918c3462-80fc-11e5-88c3-69618ab7f3ea.png)

If you use automaterials' "box select" setting, it will happily create construction jobs over the already existing constructed floor:
![floor-on-floor-box-select](https://cloud.githubusercontent.com/assets/324067/10872202/b753ad56-80fc-11e5-901e-9daa997c0f0a.png) ![floor-on-floor-box-select-final](https://cloud.githubusercontent.com/assets/324067/10872203/ba9f92d6-80fc-11e5-8ad9-2c4ab6d026ad.png)

Automaterials *should* exclude existing construction (as it does with walls for example).

This bug stems from the fact, that automaterials 
1. Does not check the base tiletypes
2. Only checks if the layer below is a wall - *not* if it is a __constructed__ wall
3. And even if a check for a __constructed wall__ in the layer below is made, we'd still be able to rebuild a constructed floor on top of that ad infinitum

I've hacked together a small fix for 1 (and 2) in 092979f3624d9029c14d77688a85746aa1a245ad (also I renamed some variables to avoid confusion due to variable name reuse). With this I think checking for 2 becomes unnecessary since the 'constructed floor over natural floor' situation gets caught by it.

![floor-on-floor-partialfix](https://cloud.githubusercontent.com/assets/324067/10873105/5e55b4fc-810f-11e5-9a2a-500759391d2e.png)
It's mostly some cannibalized code from `probe` - no idea if there are better ways to query the base tiletype than using MapCache.

---

(3) is a **really really** nasty can of worms. I don't see any _simple_ way to prevent building an infinite number of constructed floors above a constructed wall, because the tile types & shapes are completely identical if it is just a wall, or a wall with a floor on top. 

Currently the __only__ differentiating feature I see is in `d`, `n` ("Designate", "Remove Construction"): Tiles where automaterials _should_ allow floor construction are colored red, tiles where it should disallow it are colored green. I don't know if there's a way to query that information, or if we could try designating tiles, see what takes, and work with that, or if the game happily takes invalid designations.

lethosor:
[...]

Issue #613, opened 2015-05-09
automaterial: Constructions can be built on floors constructed on natural floors


lethosor:
Reported by [Jordan~](http://www.bay12forums.com/smf/index.php?topic=119369.msg6219482#msg6219482) on the forums:
![screenshot](http://i.imgur.com/uwM6uhJ.png)

PR #727, opened 2015-11-03
Automaterials: Partial fix for constructing floor over existing floor

DoctorVanGogh:
Partial fix for issue #721.

### DIFF
--- a/plugins/automaterial.cpp
+++ b/plugins/automaterial.cpp
@@ -456,35 +456,17 @@ static bool is_valid_building_site(building_site &site, bool orthogonal_check, b
         }
         else
         {
-            if (shape_basic != tiletype_shape_basic::Floor)
+            if (shape != tiletype_shape::STAIR_DOWN && shape_basic != tiletype_shape_basic::Floor)
                 return false;
 
-            if (material == tiletype_material::CONSTRUCTION)
+            // Can build on top of a wall, but not on other construction
+            auto construction = Constructions::findAtTile(site.pos);
+            if (construction)
             {
-                // Check this is not a  'constructed floor over X':
-                // those have something else than open space as *basic* tile types: stonefloor, smoothstonefloor, ...
-                MapExtras::MapCache mc;
-                auto btt = mc.baseTiletypeAt(site.pos);
-                auto bshape = tileShape(btt);
-                if (bshape != tiletype_shape_basic::Open)
-                    return false;
-
-                // Can build on top of a wall, but not on a constructed floor
-                df::coord pos_below = site.pos;
-                pos_below.z--;
-                if (!Maps::isValidTilePos(pos_below))
-                    return false;
-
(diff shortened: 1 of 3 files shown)


PeridexisErrant:
./plugins/automaterial.cpp: Contains trailing whitespace: line 472

That's a real build failure; the others are docs being broken because they require pull #713.

DoctorVanGogh:
My bad - you really need to fix the broken builds - I'm just ignoring the checks right now....

DoctorVanGogh:
But I have to add.... complaining about a blank space at the end of a comment line is bordering on... obsessive ;)

PeridexisErrant:
> My bad - you really need to fix the broken builds - I'm just ignoring the checks right now....

I agree!  But it's breaking because my huge pull hasn't bee merged yet, so I'm not pushing too hard ;)

> But I have to add.... complaining about a blank space at the end of a comment line is bordering on... obsessive ;)

Maybe, but with hundreds of people contributing if you're not anal about the little things it's way too easy to end up with a real mess.  Basically we try to break the build on literally any little issue we can detect, and hope that makes people more careful with the other stuff too ;)

lethosor:
Also, some editors remove trailing whitespace automatically, so keeping it from being introduced prevents clutter when it gets removed later.

plugins/automaterial.cpp
```diff
@@ -460,19 +461,27 @@ static bool is_valid_building_site(building_site &site, bool orthogonal_check, b
 
             if (material == tiletype_material::CONSTRUCTION)
             {
+                // Check this is not a  'constructed floor over X':
+                // those have something else than open space as *basic* tile types: stonefloor, smoothstonefloor, ...
+                MapExtras::MapCache mc;
+                auto btt = mc.baseTiletypeAt(site.pos);
+                auto bshape = tileShape(btt);
+                if (bshape != tiletype_shape_basic::Open)
```
lethosor: From GCC: 
```
warning: comparison between 'enum df::enums::tiletype_shape::tiletype_shape' and 'enum df::enums::tiletype_shape_basic::tiletype_shape_basic'
```

DoctorVanGogh: There's my general ignorance of df structures & types showing ;)

DoctorVanGogh:
Fixes #721. Fixes #734.

No more floor over floor, and you can only construct *once* on top of a wall. Oh, and you can now build on top of down stairs.


Okay - this could probably be improved performance wise, but the basic functionality is in there. Theoretical worst case right now is iterating through all world construction for each and every designated tile. 
Some Caching? Doing this before checking individual tiles for validity en block like "Get all constructions inside these n candidate positions"?.

lethosor:
I haven't gotten around to testing this yet - if you have time, does this fix the issue in #613?

lethosor:
Looks good!

PR #729, opened 2015-11-03
Add furnace to fillsidebarmenu hook

warmist:
Can someone test if it's working? If i'll find time to do it myself i'll change this message...


lethosor:
You forgot to include df/furnace_type.h, so this explodes (with GCC, at least).

lethosor:
Not sure why that build log was replaced with the older (shorter) one where RFR was broken, but I'll fix this.

lethosor:
Another issue: GCC complains about passing a df::building_furnacest instance to onWorkshopFillSidebarMenu and postWorkshopFillSidebarMenu because building_furnacest doesn't inherit from building_workshopst.

lethosor:
@warmist should I change that event to handle a generic ``df::buildingst`` instead?

Issue #722, opened 2015-11-02
Some plugins could be converted to scripts


PeridexisErrant:
Scripts do have several advantages over plugins, when the functionality could be performed by either: a clear relationship between command and filename, ease of documentation and de-duplication, limited scope for crashes on error, and better portability across DF versions.

This is not an urgent task, since things are not actually broken at the moment.  I don't expect that all of the plugins listed below will (or even should) be replaced by scripts, but it seems like a good place to start:

- [x] catsplosion
- [x] colonies - see #810
- [x] drybuckets - see also #248
- [x] feature
- [x] fixpositions - to separate fix/diplomats and fix/merchants scripts
- [x] initflags - to separate twaterlvl and tidlers scripts
- [x] weather

The criteria for this list:

- plugins provide one-use commands only (or repeat them regularly, replaceable with the repeat tool)
- reasonably simple to re-implement
- no persistent interaction or GUI elements
- not too much indexing (which is way faster in C++)

warmist:
Things that do a lot of indexing (e.g. lair) need to be in C++ for performance reasons.
Other things like this:
* lair
* fixveins
* fix-unit-occupancy (imho it should not be encorouged to be run auto-repeat, commands like these are FIXES, better to track down stuff that messes stuff up and leave this as sort-of last resort)
* deramp
* cleanedowned (this probably would be very slow in lua)
* regrass

Generally if it is already slow imagine it being 10x slower.

PeridexisErrant:
Hmm, that makes sense - I'll take them off the list.  I'd usually want to measure perf issues, but given that the existing implementation *works* we can just leave well enough alone.

Can I take it you agree in principle with the rest?

warmist:
I agree with the principle. Looking them over and seeing if it's possible (and/or hard) to port is very needed.

I would say we would also gain performance, but in last dfhack versions something takes long even in scripts (do they all get run at startup?)

Putnam3145:
Catsplosion shouldn't be hard for me to remake; it's not the first time I've made a script that forces pregnancy. Heck, I could make any male cats switch to female just so that they can give birth, too.

lethosor:
I agree with most of the current list, except for a few:

* dwarfexport hasn't been supported since 0.40.01, maybe earlier, although it doesn't appear performance-dependent
* cleanconst would be slow in large fortresses if it were written as a script (and since it's most useful in those fortresses, I'm not sure rewriting it is worth it)
[...]

PR #732, opened 2015-11-05
Clarify and restructure some docs, improve off-travis linting

PeridexisErrant:
Now ready to merge.

There's a variety of stuff in this pull, but most of the commits are logical changesets instead of reflecting the messy process, and the messages should explain them well.  The others are mostly in response to feedback, which you can read below.

PR #733, opened 2015-11-06
Remove deprecated and inactive plugins

PeridexisErrant:
`createitem` and `forceequip` have been deprecated for some time in favor of the modtools scripts.  The other plugins removed have not been built since before 40.01, and don't have much chance of that changing.  `treefarm` and `advtools` may be reactivated, so I simply moved their docs from the built documentation to their source files to properly hibernate them.

Early work on #722.


lethosor:
I wasn't aware of `createitem` being deprecated, aside from the note you just added in https://github.com/DFHack/dfhack/commit/a8ab595428fc9e31e4432010507fe37442460f5e#diff-274fd4ea3385873ecc4f0ee3ca7a7c50R2130 a few days ago, and it's still fairly widely-used (there's even an entire page with examples on the wiki). I don't think it should be removed unless it's replaced with a wrapper for modtools/create-item (like spawn-unit) and modtools/create-item offers the same functionality (it doesn't support item quantities yet). I haven't checked to see how the situation with forceequip compares.
I think some of the others (particularly misery, and maybe dwarfexport/mapexport/df2mc) could be fixed instead of removing them entirely, although that would obviously take some time. We also don't want to nuke working plugins right before a release, so I'll assign this to the "r5" milestone for now (which is a temporary name, but will hopefully be released faster than r4, whatever its name ends up being).

PeridexisErrant:
Sure.  I'll leave this for discussion then, but replicate the moved docs in #732 so Plugins.rst reflects what actually exists in someone's install.

maxthyme:
Yeah, definitely favor the wrapper method, as I hadn't even checked the modtools/create-item route at all, but the createitem SOMETHING:ITEM_SOMETHING_NEEDED WHATEVER:MATERIAL:NECESSARY 15 is way too useful for testing mods, which, yeah, fits modtools and such but just tossing in the extra - would annoy the hell out of me because I'd keep missing it.

PR #736, opened 2015-11-07
Replace initflags plugin with scripts

PeridexisErrant:
Progress towards #722 

Really it's trivial, but IMO that's all the more reason not to use a plugin for it.


scripts/tidlers.lua
```diff
@@ -0,0 +1,17 @@
+--Toggle idlers count
+--[[=begin
+
+tidlers
+=======
+Toggle between all possible positions where the idlers count can be placed.
+
+=end]]
+-- see also the enum: df.d_init_idlers
```
lethosor: It would be nice to use enum names here (and even better to use/implement something like ENUM_NEXT_ITEM in lua).

lethosor: Done in 3fb2628

PR #740, opened 2015-11-09
restrictliquid is actually restrictliquids

Enkrod:
Tried it out in the game. DFHack actually says: "restrictliquid is not recognized. Did you mean restrictliquids?"

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -1494,7 +1494,7 @@ The only undo option is to restore your save from backup.
 alltraffic
 ==========
 Set traffic designations for every single tile of the map - useful for resetting
-traffic designations.  See also `filltraffic`, `restrictice`, and `restrictliquid`.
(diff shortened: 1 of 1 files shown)


docs/Plugins.rst
```diff
@@ -1998,11 +1998,11 @@ Regrows all the grass. Not much to it ;)
...
 ===========
 Restrict traffic on all tiles on top of visible ice.
-See also `alltraffic`, `filltraffic`, and `restrictliquid`.
+See also `alltraffic`, `filltraffic`, and `restrictliquids`.
 
-.. _restrictliquid:
+.. _restrictliquids:
 
-restrictliquid
+restrictliquids
```
lethosor: Updating the line below would be good too (although I'm not sure if sphinx cares about that).

PeridexisErrant: It does - the underline must be at least as long as the title, and *may* be longer (but that looks messy).

Issue #723, opened 2015-11-02
Alternative way to organise DFHack script development


PeridexisErrant:
For script developers, the current system is awkward enough to use that most simply don't.  Many maintain git repos for their scripts with no shared history, and more still don't use git at all.

I think it self-evident that script authors working on a fork would be better than our new submodule system, and that encouraging more people to use git for DFHack scripts would be a good thing.  I further assume that we would like to know about as much of the script development being done as possible, and prefer to have the option of adopting that work into the standard distribution.

So the proposal is actually pretty simple:

1. move the scripts directory to a new repository, and make it a submodule in it's current location
2. each author forks the repo, and works in /devel/$name/ if they're not comfortable putting stuff in the main folders
3. when ready for publication, scripts are moved to the main directories and pull-requests are made

What are the advantages for users (ie script authors)?

- it's trivial to work in a functioning environment:  just grab a DF+DFHack install, `cd hack && rm -rf scripts && git clone ...`
- the same technique makes getting new stuff easy
- all the linting and testing tools are available
- encourages collaboration, since you fork one repo not many and can fetch from other repos
- no baggage:  it doesn't include any of the scary C++/Perl/Lisp/XML/Cmake/whatever that scares people off

Advantages for DFHack core developers?

- **shared history**
- more people use git for DFHack scripts, meaning better history and more content
- the build systems and merge patterns are more sensible
- management of the scripts *could* (but need not) be handed off, and just checked when it's time to update the submodule for a release
- less duplication, as a higher proportion of the code is widely known among script authors
- +many above, eg script authors linting is a Good Thing

Thoughts?  I know this is a big step, but the DFHack project is getting big enough that it's hard to manage and this seems like a natural and helpful split to make.

dscorbett:
If each author forks the repo, what is the purpose of working in devel/`$name`?

PeridexisErrant:
@warmist asked in [the Bay12 thread](http://www.bay12forums.com/smf/index.php?topic=139553.msg6586916#msg6586916) (and approved of this pull in principle)

It's simply less of a change from the way people are currently working in their own repos, and would be exempt from all the linters and style-checking (eg for presence of documentation) that the standard scripts are subject to.

I would hope that most people just work in branches if they need that, but this idea is mainly targeted at people who just don't get git at all yet, and often don't use it.

PeridexisErrant:
(oops, wrong button)

lethosor:
[...]

PR #742, opened 2015-11-10
Ls cleanup

### LINKED ISSUE #739: Standardise command descriptions for output of "ls"
At the moment, some are missing, others are too long (wrapping is pretty disruptive), and many are uninformative or verbose.

We can fix all of these, and for scripts the first two can be added to the docs linter to ensure they stay fixed.

This requirement should also be noted in the `Contributing` doc.

PeridexisErrant:
Closes #739 

>At the moment, some are missing, others are too long (wrapping is pretty disruptive), and many are uninformative or verbose.

>We can fix all of these, and for scripts the first two can be added to the docs linter to ensure they stay fixed.  This requirement should also be noted in the Contributing doc.

Done.  It turns out that some existing help wasn't being picked up.  I fixed that, and also where I saw it added useful help instead of just the script name.

### DIFF
--- a/plugins/devel/memview.cpp
+++ b/plugins/devel/memview.cpp
@@ -36,7 +36,7 @@ DFHACK_PLUGIN("memview");
 
 DFhackCExport command_result plugin_init (color_ostream &out, std::vector <PluginCommand> &commands)
 {
-    commands.push_back(PluginCommand("memview","Shows DF memory in real time.",memview));
+    commands.push_back(PluginCommand("memview","Shows DF memory in real time.",memview,false,"Shows memory in real time.\nParams: adrr length refresh_rate. If addr==0 then stop viewing."));
     memdata.state=STATE_OFF;
     mymutex=new tthread::mutex;
     return CR_OK;
--- a/plugins/devel/memview.cpp
+++ b/plugins/devel/memview.cpp
@@ -36,7 +36,7 @@ DFHACK_PLUGIN("memview");
 
 DFhackCExport command_result plugin_init (color_ostream &out, std::vector <PluginCommand> &commands)
 {
-    commands.push_back(PluginCommand("memview","Shows memory in real time. Params: adrr length refresh_rate. If addr==0 then stop viewing",memview));
+    commands.push_back(PluginCommand("memview","Shows DF memory in real time.",memview));
     memdata.state=STATE_OFF;
     mymutex=new tthread::mutex;
     return CR_OK;
--- a/plugins/reveal.cpp
+++ b/plugins/reveal.cpp
@@ -77,7 +77,7 @@ command_result nopause(color_ostream &out, vector<string> & params);
 
 DFhackCExport command_result plugin_init ( color_ostream &out, vector <PluginCommand> &commands)
 {
-    commands.push_back(PluginCommand("reveal","Reveal the map. Args 'hell', 'demon' are risky.",reveal,false,
+    commands.push_back(PluginCommand("reveal","Reveal the map.",reveal,false,
         "Reveals the map, by default ignoring hell.\n"
         "Options:\n"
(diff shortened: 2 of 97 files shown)


plugins/reveal.cpp
```diff
@@ -77,7 +77,7 @@ command_result nopause(color_ostream &out, vector<string> & params);
 
 DFhackCExport command_result plugin_init ( color_ostream &out, vector <PluginCommand> &commands)
 {
-    commands.push_back(PluginCommand("reveal","Reveal the map. 'reveal hell' will also reveal hell. 'reveal demon' won't pause.",reveal,false,
+    commands.push_back(PluginCommand("reveal","Reveal the map. Args 'hell', 'demon' are risky.",reveal,false,
```
lethosor: You probably don't need to list arguments here since they're listed in the help text below.

lethosor:
What's a script plugin?
It would be better to stop `ls` from requiring a space (and the travis check as well), since there's really no reason to.

plugins/devel/memview.cpp
```diff
@@ -36,7 +36,7 @@ DFHACK_PLUGIN("memview");
 
 DFhackCExport command_result plugin_init (color_ostream &out, std::vector <PluginCommand> &commands)
 {
-    commands.push_back(PluginCommand("memview","Shows memory in real time. Params: adrr length refresh_rate. If addr==0 then stop viewing",memview));
+    commands.push_back(PluginCommand("memview","Shows DF memory in real time.",memview));
```
lethosor: It would be nice to add extended help text here.

scripts/autofarm.rb
```diff
@@ -1,4 +1,4 @@
-
+# Select cropt to plant based on current stocks
```
dscorbett: “cropt” → “crops”

lethosor:
2882422 should fix the space issue.

PeridexisErrant:
"script plugin" -> I have no idea why I typed that.  Your patch to `ls` is clearly better; I'm just read-only in CPP and don't know the DFHack code that well.  Sounds like it's done now?

expwnent:
Yeah, looks good. For `ls` the description should be short as long as there's a longer more detailed explanation available. If we include the docs offline (issue #801)  then we should consider a way of automatically opening them to the relevant part with the help command. Maybe we should do that anyway except directly open the online version in the default browser. I don't know how to to do that but it would be neat.

PeridexisErrant:
Note:  it will be far more difficult to merge this if it's left for after #723 (sorry, @lethosor, I keep finding more dependent issues!)

> If we include the docs offline (issue #801) then we should consider a way of automatically opening them to the relevant part with the help command. Maybe we should do that anyway except directly open the online version in the default browser.

There's some similar discussion in #869; I proposed `search.lua` which would just open the docs search page with any arguments as the url query parameter (properly escaped, of course).  Eg `search plugin script` would open a browser to https://dfhack.readthedocs.org/en/stable/search.html?q=plugin+script - or the local copy is available.

expwnent:
I can't conveniently merge it right now so I just marked it for the next release milestone.

PeridexisErrant:
@lethosor @expwnent - could this be merged before r1?  If it would help, I'm happy to rebase for easier merging.

scripts/make-monarch.lua
```diff
@@ -1,4 +1,4 @@
---set target unit as king/queen
+-- set target unit as king/queen
```
lethosor: Isn't necessary anymore

lethosor:
I'll probably be able to tomorrow. Any idea how bad the merge conflict is?

PeridexisErrant:
Pull rebased on develop, and dropped unneeded fixes.  Should be ready and easy to merge!

Also a nice education in the power of `git rebase -i`, wow.  Took a few attempts, but it'll be a lot of fun if I need that again :smile:

PR #746, opened 2015-11-14
Support for hooking into GUI-related functions

lethosor:
This allows plugins like TwbT to modify the behavior of functions like `Screen::paintString()` or `Gui:: getDwarfmodeViewDims()` (mifki currently distributes separate versions of a few plugins with hardcoded calls to different TwbT functions when TwbT is active).
As it stands, this only supports hooking into all functions that use the screen-drawing API, and no API functions support the "draw to map" parameter suggested by mifki yet (although the hook itself does). An example plugin is included that colors all DFHack-drawn text differently:
screen shot 2015-11-13 at 8 19 10 pm

@mifki, @warmist: Should address mifki/df-twbt#21


lethosor:
@pronvit: This now includes hooks for getDwarfmodeViewDims() and getDepthAt(). A couple notes:
* getDepthAt() takes map x/y coordinates, rather than window coordinates, which is more consistent with other functions in the Gui namespace, so you'll need to convert them back.
* Your current `renderer_cool::map_dims()` only returns the four map-related dimensions, so you'll need to return useful values for the rest. It's pretty easy to call the default function (or the next installed hook), though - something like this ought to work:
```c++
Gui::DwarfmodeDims map_dims();
GUI_HOOK_CALLBACK(Gui::Hooks::dwarfmode_view_dims, map_dims_hook, map_dims);
Gui::DwarfmodeDims map_dims()
{
    auto dims = map_dims_hook.next()();
    if (is_twbt())
    {
        dims.map_x1 = 1;
        dims.map_x2 = r->gdimx;
        dims.y1 = 1;
        dims.y2 = r->gdimy;
    }
    return dims;
}
```

pronvit:
There should be separate y1/y2 in DwarfmodeDims then, for map and menu.

library/include/modules/Gui.h
```diff
@@ -150,6 +152,10 @@ namespace DFHack
         DFHACK_EXPORT bool setDesignationCoords (const int32_t x, const int32_t y, const int32_t z);
 
         DFHACK_EXPORT bool getMousePos (int32_t & x, int32_t & y);
+
+        // The distance from the z-level of the tile at map coordinates (x, y) to the closest ground z-level below
```
pronvit: ...to the closest rendered level, ground, branches, or whatever it is

warmist:
I wonder what is the performance hit? We will be calling virtual function in a very tight loop, although it was never a bottleneck (except for e.g. manipulator) i would be nice to check it.

lethosor:
It doesn't seem to impact performance much on my end (at least, it increases CPU time by less than DF fluctuates when paused), but the same is true for manipulator. Are you having performance issues with the manipulator plugin (or this) on your machine?

lethosor:
@pronvit Are the three plugins you distribute the only plugins that use the map y dimensions? I may just add separate `map_y1` and `map_y2` fields without renaming `y1` and `y2`, but I want to make sure that those fields are taken into account everywhere they need to be.

pronvit:
I'm not sure, I personally don't use many plugins, so these three are what I noticed and/or other users reported. Most of the plugins I've found using getDwarfmodeViewDims draw in menu area anyway, but I may have missed some. Another thing, there are some plugins, e.g. siege-engine that don't work with TWBT for other reason, need to check them at some point.

lethosor:
The main issue with gui/siege-engine, from what I understand, is that it draws to the map (with the DFHack drawing functions). Once I get around to adding a `map` argument to those and modifying that script to use it, you should be able to intercept those drawing calls appropriately.

pronvit:
Maybe. Btw, some plugins use drawing functions from uicommon.h (that will need the `map` argument), some have the same functions included in the source, it's a bit messy.

lethosor:
I've been meaning to clean up the duplicate OutputString() functions in various plugins, so that'll be good to handle at the same time. The Lua API only exposes the paintTile() family, though, which means that all Lua scripts that draw to the screen can be handled pretty easily, and I think the plugins you distribute also use those functions.

pronvit:
Those three use OutputString() (from uicommon or their own) to output single X on the map, probably that should be changed to use lower-lever function.

lethosor:
Yeah, something like that would be easier with paintTile() (and all versions of OutputString() should use that internally; the only problem is not supporting a `map` parameter).

lethosor:
@pronvit 2a2ab00 improves the situation with Output*(), although I haven't gotten around to removing individual functions in plugins or changing lua scripts yet. Have you had any luck with implementing hooks in twbt?

PR #750, opened 2015-11-16
Add Felton's menu-mouse plugin

PeridexisErrant:
menu-mouse is not perfect, and can have some strange interactions with other GUI-altering tools.  However it does work and provides a very valuable function for some players who would be willing to live with the downsides.

Given that Felton seems to have gone offline for good, I think it should be in the DFhack repo where further development might take place.


PeridexisErrant:
Welp, looks like someone who knows more C++ than me will need to fix this before it can be merged.  I'll leave it here in the hope that someone does so!

lethosor:
There have been several reports of this crashing on embark.

Issue #735, opened 2015-11-07
Reduce duplication of tools, remove obsolete tools


PeridexisErrant:
There are three lists:  things to remove, things to update, and tools to de-duplicate.

On the deprecation end, there's a bunch of stuff hanging around that hasn't been built or used since 0.40.01:

- [x] `dwarfexport`: disabled for some time, unused by modern tools
- [x] `mapexport`: disabled for some time, unused by modern tools
- [x] `df2mc`: hasn't been compatible with minecraft for years
- [x] `DFusion`: very old, all functions replaced save multirace embark parties
- [ ] `modtools/reaction-trigger-transition`: it's been two years, hopefully the transition is over and the script can be removed!

Some plugins are still incompatible with 40.xx, but would be great to update.  Alternatively they might be removed if it's judged not worth the bother:

- [ ] `advtools`
- [ ] `misery`
- [ ] `treefarm`

Tools with duplicated functionality, or things that could be merged: (related to #722)

- [ ] item creation: `createitem` plugin, `modtools/create-item`, `gui/create-item`.  Could drop the plugin, upgrade modtools and use it as a backend for the gui (with script_environment or similar)
- [ ] `forceequip` plugin is mostly replaced by `modtools/equip-item`
- [ ] `fastdwarf`, `teledwarf`, `superdwarf` - needs sorting out somehow.  See #664, #665.
- More to come.  Adding links between similar commands in the documentation often uncovers possibilities.

maxthyme:
Is this also a point for doubled stuff?

Like, dfusion allows you to heal up, but it doesn't like to double-up on a unit, and full-heal is faster.

create-items duplicates a lot of gui/hack-wish, is almost the same name as modtools/create-item, and far more limited than createitem, plus the digv/digtype/etc being official now.

PeridexisErrant:
I see deduplicating the standard library is a huge job - even more so because we probably want to have an official version of anything commonly used.  No issue for that because it's not clear what would be required to close it!

Ideally there would be "one obvious way to do it", whatever "it" is, but that's probably out of reach for the foreseeable future.  For now, I'm happy to just include everything and work on standardising can happen where it happens - because there's far more duplication of useful bits when DFHack doesn't supply a standard.

Putnam3145:
Yeah, it's a pretty interesting problem. Does create-items do anything that createitem doesn't? For that matter, does createitem do anything that gui/create-item does?

actually why are there so many item creation scripts around

[...]

PR #751, opened 2015-11-17
Remove dfusion, df2mc, mapexport, dwarfexport

PeridexisErrant:
See #735 - these plugins have not been built since 40.01 (or earlier), and nobody missed them.  DFusion *has* been built, but @warmist notes that almost all functions have been implemented by more recent tools (and multirace embarks is on the issue tracker).

This is mostly cleaning out unused code, with a side of making the project structure easier to understand for users and developers.

Only dfusion is noted as removed in NEWS.rst, as the others have already been removed from the release packages.


BenLubar:
Doesn't dfusion control the `lua` command? Or is that dead code?

lethosor:
No, that's `scripts/lua.lua`. `scripts/dfusion.lua` is the current dfusion front-end.

JapaMala:
mapexport is used for fortress overseer, but I don't know if anybody still
uses it.

lethosor:
Right now, I'm leaning towards removing df2mc and dfusion (waiting to hear back from @warmist on the latter). dwarfexport and mapexport might still be useful, at least until everything they implement is also implemented by RemoteFortressReader (which @JapaMala would know better than I would), so I'm leaning towards repairing them if that's possible.

warmist:
Just to have same info here as in irc. Dfusion is mostly outdated. Most important thing that i could think of is the site creation that should probably be moved to mod-tools.

NEWS.rst
```diff
@@ -36,6 +36,10 @@ Fixes
 -----
 - confirm haul-delete: Fixed issue preventing deletion of stop conditions or using "x" in route names
 
+Removed
+-------
+- DFusion: legacy script system, obsolete or replaced by better alternatives
```
lethosor: Wrong section

lethosor:
Okay, so it turns out that the NEWS.rst entries here and in #754 wound up in the wrong section because the commits were made before 0.40.24-r5. Sorry about that.
Maybe rebasing old PRs in the future would help, but at the very least we should verify that changelog entries end up in the right place when merging.

JapaMala:
I don't know anything about dwarfexport, but mapexport is used by Fortress Overseer, which can't use anything else. But since the overseer project is abandoned, as far as I can see, we might be able to get rid of it unless somebody feels like updating it.

PeridexisErrant:
(Pull already merged, but here's some rationale again)

[fortress overseer](http://dffd.bay12games.com/file.php?id=3882) hasn't worked since 34.11 - it can't handle eg multitile trees - and was last updated in 2012.  Plus the plugin hasn't compiled for ages.

`dwarfexport` was only used by [Runesmith](http://dffd.bay12games.com/file.php?id=2538), a Therapist-precursor with hints of `gm-editor`.  Outdated since 31.25!

`df2mc` only exported valid levels for Minecraft Alpha, and again hadn't built in ages.

`dfusion` is the most questionable case, but is clearly legacy code.  Having been mostly replaced by better things, it's time to take it out.

PR #752, opened 2015-11-17
Replace drybuckets and feature plugins with scripts

### LINKED ISSUE #248: drybuckets - only target appropriate buckets
Currently, the drybuckets command removes water from every bucket in your entire fortress. In order to prevent problems, it should probably skip buckets which are part of a building or assigned to a job.

PeridexisErrant:
Progress towards #722.  Also closes #248, by checking job and building flags.


scripts/fix/dry-buckets.lua
```diff
@@ -0,0 +1,22 @@
...
+to be used for making lye.  Skips buckets in buildings (eg a well),
+being carried, or currently used by a job.
+
+=end]]
+
+local emptied = 0
+local water_type = dfhack.matinfo.find('WATER').type
+
+for _,item in ipairs(df.global.world.items.all) do
+    if item:getMaterial() == water_type and not (item.flags.in_job or item.flags.in_building) then
```
lethosor: I don't see a corresponding `(item->getType() == item_type::LIQUID_MISC)` check here. The in_building/job checks are a good idea, though. You could also check to see if the item is contained in a bucket - probably in `general_refs` or `specific_refs`.

PeridexisErrant: I can add a check that `dfhack.items.getContainer(item) ~= nil`, but checking `item:getType()` causes an error since some items don't have a `.mat_type` attribute.

warmist: item:getType() is not a material type "get". It's fundamental item class type "get" so it should be okay for all the items.

PR #754, opened 2015-11-18
Merge gamelog scripts to modtools/extra-gamelog

PeridexisErrant:
Combining `log-region` and `soundsense-season` with most of the actual `soundsense` script.  Keeping this in one place makes sense to me, and makes life easier for users of other tools such as Soundsense.


scripts/modtools/extra-gamelog.lua
```diff
@@ -0,0 +1,189 @@
...
+    log_buildings()
+    dfhack.timeout(50, 'ticks', event_loop)
+end
+
+local args = {...}
+if args[1] == 'disable' then
+    dfhack.onStateChange[_ENV] = nil
+elseif args[1] == 'enable' then
+    dfhack.onStateChange[_ENV] = log_on_load
+    event_loop()
```
lethosor: This will keep running after the script is disabled. (I usually have a global `enabled` variable that I set here and check before scheduling the timeout again.)

scripts/modtools/extra-gamelog.lua
```diff
@@ -0,0 +1,192 @@
...
+    "Kiln",
+    "Magma Smelter",
+    "Magma Glass Furnace",
+    "Magma Kiln",
+    "Custom Furnace",
+    }
+
+buildStates = {}
+
+function log_buildings()
```
lethosor: This stuff looks new and ought to be mentioned.

lethosor:
I'm not entirely sure if "all of these scripts write to gamelog.txt" is enough of a reason to combine them, although it makes more sense now that I realize you added new features at the same time. It would be nice to still have the ability to enable/disable features separately, although I can work on that.

scripts/modtools/extra-gamelog.lua
```diff
@@ -0,0 +1,192 @@
...
+    if siege ~= old_siege and siege then
+        msg("A vile force of darkness has arrived!")
+    elseif siege ~= old_siege and not siege then
+        msg("Siege was broken.")
+    end
+end
+
+
+local workshopTypes = {
+    [0]="Carpenters Workshop",
```
lethosor: Ideally these would be implemented as enum attributes in df-structures (e.g. `df.building_type.attrs[0].caption`). I'll probably change that later when I get a chance.

PeridexisErrant: This (and the leadership logging above) is taken from the `soundsense` script Zwei distriubutes - I added most of the logging it does, somewhat more structured and without the spammy bits.

expwnent:
Looks good to me.

NEWS.rst
```diff
@@ -36,6 +36,10 @@ Fixes
 -----
 - confirm haul-delete: Fixed issue preventing deletion of stop conditions or using "x" in route names
 
+New Scripts
+-----------
+- `modtools/extra-gamelog`: replaces ``log-region``, ``soundsense-season``, and ``soundsense``
```
lethosor: Also the wrong section

PR #758, opened 2015-11-24
Re-enable GCC 5 compilation for OSX and Linux; updated build-osx.sh script

TheBloke:
Apologies that these are A) on Develop, and B) two unrelated commits in a single Pull.  I'm still having git issues - this time, it merged my local branches into origin/develop when I pushed, for no reason I can yet understand.

Commit 1:

Re-enable GCC 5 compilation for OSX and Linux, passing -D_GLIBCXX_USE_CXX11_ABI=0 to disable new C++11 name mangling and remain compatible with DF.

Commit 2:
OSX build script (build-osx.sh) - multiple improvements:
  Can specify -gcc X to change GCC version from default of 4.5.
  Can specify -par Y to give a -jY parallel build flag to make.
  Various error checks: existence of executables, output dir, etc.
  Improved Bash syntax, especially for Lua patching (now a valid .sh file).
  Times the overall build time and outputs this to stdout.
  If GCC version is not default, print reminder about installing libstdc++.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,13 @@ project(dfhack)
 macro(CHECK_GCC COMPILER_PATH)
   execute_process(COMMAND ${COMPILER_PATH} -dumpversion OUTPUT_VARIABLE GCC_VERSION_OUT)
   string(STRIP "${GCC_VERSION_OUT}" GCC_VERSION_OUT)
-  if (${GCC_VERSION_OUT} VERSION_LESS "4.5" OR ${GCC_VERSION_OUT} VERSION_GREATER "4.9.9")
-    message(SEND_ERROR "${COMPILER_PATH} version ${GCC_VERSION_OUT} cannot be used - use GCC 4.5 through 4.9")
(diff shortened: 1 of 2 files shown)


lethosor:
I think the lua patch is no longer necessary on 10.6 (check to see if there's a strnlen implementation in LuaTypes.cpp).

TheBloke:
Ah yes, looks like it!

```C++
#ifdef _DARWIN
#if MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_6
size_t strnlen (const char *str, size_t max)
{
    const char *end = (const char*)memchr(str, 0, max);
    return end ? (size_t)(end - str) : max;
}
#endif
#endif
```


So lua patching definitely not required because of the above?

TheBloke:
I think I'll convert the script to build-nix.sh and make it applicable for Linux too, as there's no script for Linux and it could benefit from some of this as well.

lethosor:
Yeah, the patch shouldn't be necessary. The issue was that strnlen() didn't exist on 10.6, so the change to LuaTypes.cpp added a replacement if 10.6 is targeted.

PR #765, opened 2015-12-04
Implement "weather" as a script

PeridexisErrant:
Another item from #722 

Behavior is identical except for helpstring and messages on change (notably, makes the same changes and prints an identical map.


lethosor:
Are you sure this actually changes the weather?

PeridexisErrant:
...whoops, typo.  `ReadCurrentWeather` *is* a valid function, but `SetCurrentWeather` is the one I actually wanted.  Fixed.

lethosor:
Merged manually in 81b055e.

Issue #755, opened 2015-11-20
Command line option to automatically load a save game on DF launch, bypassing main menu, and load a different game into current DF


TheBloke:
I don't know if any of this is possible - perhaps not, else it's the sort of thing that would already exist!  But just in case:  

* Add a command line option that will load DF straight into the named save game, bypassing (or at least instantly auto-navigating) the main menu
  * Will open DF and cause *data/save/xyz/* to be loaded; no main menu is shown; the user first sees the normal game loading screens and once these are completed is at the map.
* Would be somewhat useful for all users, who most likely use the main menu 95% of the time only to load their current save and might thus appreciate the ability to set up a shortcut that takes them straight into it.
* Also useful for speeding DFHack testing in situations where a new DF instance is required because load/unload is insufficient
  * Could also enable some basic automated testing: create a script that launches DF into a given game, execute a defined lists of commands then exit at the end.
* Other potentially useful options:
  * Load the most recent save
  * Load a random save from a given list
  * Load save 1, run comands XYZ, then load save 2, etc
  * If necessary all could be achieved from the launcher script rather than the plugin itself

**Example invocation:**
* Windows:  DwarfFortress.exe -load region1   *or*,  DFLoaderScript.bat -load region1
* Linux/OSX:  ./dfhack -load region1
  * If DFHack can't read DF's argc/argv, then perhaps this could be facilitated by a launcher script/batch file (a new one for Windows, existing ```dfhack``` for *nix)
    * The launcher script writes the name of the game to load to a text file which is then checked by a plugin; if the file is found and the game name within it is valid, the plugin loads this game on DF launch.  The plugin then deletes the text file.
    * Or the script uses the API to pass a "load game" command to the DFHack plugin - assuming the plugin can load fast enough to accept the command in a timely manner from the same script that launched it

**And the ability to switch saves in a loaded DF:**
* DFHack plugin the can be executed from console/Lua UI: ```loadgame xyz```
  * When executed, kills the current game (like ```die```) but does not exit DF; instead loads a different game into existing DF
  * Equivalent to quitting back to the main menu (without deleting all saves) and loading a different game.

lethosor:
DFHack's lack of access to argc/argv is the main obstacle here, but maybe a simple way to pass arguments to DFHack and have it handle them would be possible - I think parsing them in three different launcher scripts (in two different languages) would be harder to maintain.
[...]

PR #767, opened 2015-12-13
vim-style +args

BenLubar:
For example, `./dfhack +echo foo` would be equivalent to running `./dfhack` and then typing `echo foo` into the console. uses shell argument splitting, so `./dfhack +somecommand "foo bar" baz` does the right thing.

See DFHack#755.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -62,6 +62,7 @@ using namespace std;
 using namespace DFHack;
 
 #include "df/ui.h"
+#include "df/ui_sidebar_menus.h"
 #include "df/world.h"
 #include "df/world_data.h"
 #include "df/interfacest.h"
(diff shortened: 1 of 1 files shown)


expwnent:
What's wrong with putting it in dfhack.init?

BenLubar:
It's easier to run a command once by putting it on the command line than it is to edit dfhack.init and then start the game and edit dfhack.init again. vim has .vimrc and +foo for the same reason.

The other part of #755 is that a plugin would have to define a `load region1` command, and then you could run `./dfhack +load region1` to skip the entire menu system and go straight to the loading screen for the world. Something like that would generally be a one-time use, not something you'd want to run every time you started the game.

library/Core.cpp
```diff
@@ -1549,6 +1550,39 @@ bool Core::Init()
     if (!server->listen(RemoteClient::GetDefaultPort()))
         cerr << "TCP listen failed.\n";
 
+    vector<string *> & args = df::global::ui_sidebar_menus->unk.anon_2;
```
lethosor: This should check for the goal being non-null (and the fields should probably be renamed too).

BenLubar:
Ok, so DF does some weird stuff with command line arguments.

My new code should handle everything, but the way DF parses `./dfhack +foo "-bar -baz"` means that `foo` gets two arguments, each starting with a hyphen. There's no way to differentiate between that and `./dfhack +foo -bar -baz`, unfortunately.

However, I don't think there are any dfhack commands that need ` -` inside any of their arguments, so this shouldn't be a problem.

PR #768, opened 2015-12-15
Compile.rst: major updates to Windows and OSX sections;  

TheBloke:
minor to Linux; new section re compiling documentation; many small fixes.

Introduction.rst: fix typo and remove erroneous end version for OSX.

All changes have been tested as compiled HTML.

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -203,23 +203,23 @@ following environment variable::
 * Additional notes for El Capitan (OSX 10.11) and XCode 7.x users
 
   * You will probably find that gcc45 will fail to install on OSX 10.11,
-     or any older OSX that is using XCode 7.
+    or any older OSX that is using XCode 7.
   * There are two workarounds:
 
     * Install GCC 5.x instead (``brew install gcc5``), and then after compile
-       replace ``hack/libstdc++.6.dylib`` with a symlink to GCC 5's i386
-       version of this file::
+      replace ``hack/libstdc++.6.dylib`` with a symlink to GCC 5's i386
+      version of this file::
 
         cd <path to df>/hack && mv libstdc++.6.dylib libstdc++.6.dylib.orig &&
         ln -s /usr/local/Cellar/gcc5/5.2.0/lib/gcc/5/i386/libstdc++.6.dylib .
 
     * Install XCode 6, which is available as a free download from the Apple
-       Developer Center.
+      Developer Center.
 
       * Either install this as your only XCode, or install it additionally
-         to XCode 7 and then switch between them using ``xcode-select``
+        to XCode 7 and then switch between them using ``xcode-select``
       * Ensure XCode 6 is active before attempting to install GCC 4.5 and
-         whenever you are compiling DFHack with GCC 4.5.
+        whenever you are compiling DFHack with GCC 4.5.
 
 * Install Perl dependencies
 
@@ -260,7 +260,6 @@ following environment variable::
     and it is also immune to changes from OS updates.
 
 * Get the DFHack source as per section `compile-how-to-get-the-code`, above.
-
 * Set environment variables
 
   Homebrew (if installed elsewhere, replace /usr/local with ``$(brew --prefix)``)::
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -464,25 +464,27 @@ Building the documentation
 DFHack documentation, like the file you are reading now, is created as .rst files,
 which are in `reStructuredText (reST) <http://sphinx-doc.org/rest.html>`_ format.
 This is a documenation format that has come from the Python community. It is very
-similar in concept - and in syntax - to Markdown, as found on GitHub and many other places.
-However it is more advanced than Markdown, and can be compiled to sophisticated HTML files
-with tables of contents, cross-linking, references and more.
+similar in concept - and in syntax - to Markdown, as found on GitHub and many other
+places. However it is more advanced than Markdown, with more features available when
+compiled to HTML, such as automatic tables of contents, cross-linking, special
+external links (forum, wiki, etc) and more. The documentation is compiled by a
+Python tool, `Sphinx <http://sphinx-doc.org>`_.
 
-The documentation can be built during the standard DFHack compilation procedure,
-but this has been disabled by default.  You only need to build the docs if you're changing
-them, or perhaps if you want a local HTML copy; otherwise, read them easily online at
+The DFHack build process will compile the documentation but this has been disabled
+by default. You only need to build the docs if you're changing them, or perhaps
+if you want a local HTML copy; otherwise, read them easily online at
 `ReadTheDoc's DFHack hosted documentation <https://dfhack.readthedocs.org>`_.
 
-(Note that even if you do want a local copy, it is certainly not necesesary to compile the
-documentation in order to read it. Like Markdown, reST documents are designed to be just as
-readable in a plain-text editor as they are in HTML format. The main thing you lose in plain
-text format is links.)
+(Note that even if you do want a local copy, it is certainly not necesesary to
+compile the documentation in order to read it. Like Markdown, reST documents are
+designed to be just as readable in a plain-text editor as they are in HTML format.
+The main thing you lose in plain text format is hyperlinking.)
 
 
 Enabling documentation building
 ===============================
-First, make sure you have followed all the necessary steps for your platform as outlined
-in the rest of this document.
+First, make sure you have followed all the necessary steps for your platform as
+outlined in the rest of this document.
 
 To compile documentation with DFHack, add the following flag to your ``cmake`` command::
 
@@ -492,12 +494,12 @@ For example::
 
   cmake .. -DCMAKE_BUILD_TYPE:string=Release -DBUILD_DOCS:bool=ON -DCMAKE_INSTALL_PREFIX=<path to DF>
 
-Alternatively you can use the CMake GUI which allows options to be easily changed.
+Alternatively you can use the CMake GUI which allows options to be changed easily.
 
-On Windows you should either use ``generate-msvc-gui.bat`` and set the option through the
-GUI, or else if you want to use an alternate file, such as ``generate-msvc-all.bat``, you
-will need to edit it to add the flag. Or you could just run ``cmake`` on the command line
-like in other platforms.
+On Windows you should either use ``generate-msvc-gui.bat`` and set the option
+through the GUI, or else if you want to use an alternate file, such as
+``generate-msvc-all.bat``, you will need to edit it to add the flag.
+Or you could just run ``cmake`` on the command line like in other platforms.
 
 Required dependencies
 =====================
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -10,6 +10,7 @@ and `install the latest release instead <installing>`.
 .. contents::
    :depth: 2
 
+.. _compile-how-to-get-the-code:
 
 How to get the code
 ===================
@@ -17,7 +18,7 @@ DFHack doesn't have any kind of system of code snapshots in place, so you will h
 get code from the github repository using git.  How to get git is described under
 the instructions for each platform.
 
-To get the code::
+To get the latest release code (master branch)::
 
     git clone --recursive https://github.com/DFHack/dfhack
     cd dfhack
@@ -25,9 +26,26 @@ To get the code::
 If your version of git does not support the ``--recursive`` flag, you will need to omit it and run
 ``git submodule update --init`` after entering the dfhack directory.
 
+To get the latest development code (develop branch)::
+
+  git clone --recursive https://github.com/DFHack/dfhack
+  cd dfhack
+  git checkout develop
+  git submodule update
+
+**Important note on submodule update**:
+
+It is necessary to run ``git submodule update`` every time you change Git branch,
+for example when switching between master and develop branches and back.
+
+
+Contributing to DFHack
+======================
 If you want to get involved with the development, create an account on
 Github, make a clone there and then use that as your remote repository instead.
-We'd love that; join us on IRC (#dfhack channel on freenode) if you need help.
+
+We'd love that; join us on IRC (#dfhack channel on freenode) for discussion,
+and whenever you need help.
 
 
 Build types
@@ -40,7 +58,8 @@ Without specifying a build type or 'None', cmake uses the
 ``CMAKE_CXX_FLAGS`` variable for building.
 
 Valid and useful build types include 'Release', 'Debug' and
-'RelWithDebInfo'. 'Debug' is not available on Windows.
+'RelWithDebInfo'.
+'Debug' is not available on Windows.
 
 
 Linux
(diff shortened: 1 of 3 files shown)


docs/Compile.rst
```diff
@@ -25,9 +26,26 @@ To get the code::
...
+To get the latest development code (develop branch)::
+
+  git clone --recursive https://github.com/DFHack/dfhack
+  cd dfhack
+  git checkout develop
+  git submodule update
+
+**Important note on submodule update**:
+
+It is necessary to run ``git submodule update`` every time you change Git branch,
```
lethosor: Grammar?

docs/Compile.rst
```diff
@@ -240,12 +392,123 @@ Next, run one of the scripts with ``generate`` prefix. These create the MSVC sol
...
+In order to build the documentation, you must have Python with Sphinx
+version 1.3.1 or later. Both Python 2.x and 3.x are supported.
+
+When installing Sphinx from OS package managers, be aware that there is
+another program called Sphinx, completely unrelated to documentation management.
+Be sure you are installing the right Sphinx!
+
+
+Linux
+-----PTION(BUILD_DOCS "Choose whether to build the documentation (requires python and Sphinx)." ON)
```
lethosor: s/-----/O/

docs/Compile.rst
```diff
@@ -240,12 +392,123 @@ Next, run one of the scripts with ``generate`` prefix. These create the MSVC sol
...
+
+
+Required dependencies
+=====================
+In order to build the documentation, you must have Python with Sphinx
+version 1.3.1 or later. Both Python 2.x and 3.x are supported.
+
+When installing Sphinx from OS package managers, be aware that there is
+another program called Sphinx, completely unrelated to documentation management.
+Be sure you are installing the right Sphinx!
```
lethosor: Often python-sphinx, I think

docs/Compile.rst
```diff
@@ -25,9 +26,26 @@ To get the code::
 If your version of git does not support the ``--recursive`` flag, you will need to omit it and run
 ``git submodule update --init`` after entering the dfhack directory.
 
+To get the latest development code (develop branch)::
+
+  git clone --recursive https://github.com/DFHack/dfhack
+  cd dfhack
```
lethosor: I'm not entirely sure if repeating these first two lines is necessary.

TheBloke:
OK check again

TheBloke:
Or rather do now that I've actually committed the fixes

TheBloke:
Few more changes

docs/Compile.rst
```diff
@@ -10,14 +10,15 @@ and `install the latest release instead <installing>`.
 .. contents::
    :depth: 2
 
+.. _compile-how-to-get-the-code:
 
 How to get the code
 ===================
 DFHack doesn't have any kind of system of code snapshots in place, so you will have to
-get code from the github repository using git.  How to get git is described under
+get code from the Github repository using git.  How to get git is described under
```
dscorbett: `s/Github/GitHub/; s/git/Git/g`

TheBloke: Thanks, changed.

expwnent:
Are you sure Git is supposed to be capitalized when not a command? You don't capitalize ls, by comparison.

dscorbett:
Yes, Git is supposed to be capitalized. See http://git-scm.com/.

TheBloke:
I'd originally thought not, but after dscrobett's comment I Googled and Git as a product/package is capitalised.

TheBloke:
OK I'm definitely done now, closing my Vim window so I stop tweaking.  Let me know if there's any more issues or desirable changes.

docs/Compile.rst
```diff
@@ -51,15 +71,19 @@ Dependencies
...
-work due to ABI changes (the entire plugin loading system won't work, for
-example). On 64-bit distributions, you'll need the multilib development tools
-and libraries (``gcc-multilib`` or ``gcc-4.x-multilib`` on Debian). Note that
-installing a 32-bit GCC on 64-bit systems (e.g. ``gcc:i386`` on Debian) will
+(see below), but any version from 4.5 onwards (including 5.x) will work.
+
+On 64-bit distributions, you'll need the multilib development tools and libraries:
+
+* ``gcc-multilib`` and ``g++-multilib``
+* On Debian: ``gcc-4.x-multilib`` and ``g++-4.x-multilib``
```
lethosor: No, it's still `gcc-multilib` and `g++-multilib` (I'm pretty sure gcc-4.x isn't a package). The 4.x part is necessary if you're using GCC 4.x (e.g. use gcc-4.9-multilib and g++-4.9-multilib if you're compiling with GCC 4.9).

docs/Compile.rst
```diff
@@ -121,7 +156,9 @@ DF, which causes DF to use your system libstdc++ instead::
 Note that distributing binaries compiled with newer GCC versions requires end-
 users to delete libstdc++ themselves and have a libstdc++ on their system from
 the same GCC version or newer. For this reason, distributing anything compiled
-with GCC versions newer than 4.8 is discouraged.
```
lethosor: The first version we had issues with was 4.9, since Arch adopted it before Debian, so the 4.8 was intentional because it should work on pretty much any reasonably up-to-date system. However, if you have to install a different version of gcc for compatibility, an older version (>=4.5) would probably be better, although support for 4.5-4.7 (maybe 4.8 too) ought to be equally widely available on different distros.

docs/Compile.rst
```diff
@@ -134,12 +171,16 @@ following environment variable::
 
     export MACOSX_DEPLOYMENT_TARGET=10.9
 
-#. Download and unpack a copy of the latest DF
-#. Install Xcode from Mac App Store
-#. Open Xcode, go to Preferences > Downloads, and install the Command Line Tools.
-#. Install dependencies
+* Download and unpack a copy of the latest DF
+* Install Xcode from Mac App Store
```
lethosor: Might want to mention 10.11 issues sooner (here or above?)

docs/Compile.rst
```diff
@@ -200,39 +280,141 @@ On Windows, DFHack replaces the SDL library distributed with DF.
...
+and check for and install the SP1 update. If no update is found, check that
+your Windows Update settings include "Updates from all Microsoft products".
+
+You can confirm whether you have SP1 by opening the Visual Studio 2010 IDE
+and selecting About from the Help menu.  If you have SP1 it will have *SP1Rel*
+at the end of the version number, for example: *Version 10.0.40219.1 SP1Rel*
+
+It is vital that you use SP1 as, while building with the original release
+of Visual Studio 2010 (RTM) may succeed, it will result in non-working DFHack
+binaries that crash when connecting to Dwarf Fortress.
```
lethosor: I thought missing SP1 just led to compilation issues but the ABIs were compatible. Did SP1 also break ABI compatibility?

("interacting with" might be better than "connecting to", since DFHack runs in-process.)

docs/Compile.rst
```diff
@@ -200,39 +280,141 @@ On Windows, DFHack replaces the SDL library distributed with DF.
...
+
+Additional dependencies: installing with the Chocolatey Package Manager
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The remainder of dependencies - Git, CMake and StrawberryPerl - can be most
+easily installed using the Chocolatey Package Manger. Chocolatey is a
+\*nix-style package manager for Windows. Think "``apt-get`` for Windows."
+
+Chocolatey is the recommended way of installing the required dependencies
```
lethosor: I'm a bit hesitant to refer to this as "the" recommended way, unless there was a discussion that led to that which I'm not aware of.

docs/Compile.rst
TheBloke: OK, so just remove those Debian lines?  That's not anything I've edited, just brought over from the original version of the doc (with updated formatting.)

docs/Compile.rst
TheBloke: Ah, I thought it was ABI issues on the basis that it *does* compile on RTM (non-SP1).   At least someone in #dfhack got it compiled the other day on RTM.  He did have a bunch of issues but ultimately a lot of that may have been his own system issues.

If you've heard it's not ABI issues, and I've heard it's not compilation issues, then maybe it's no issue at all.

I can try it out I guess, but it's so easy for anyone to upgrade to SP1 I'll probably just tone down the above message to something like "Using SP1 may cause issues, don't do it." rather than stating what they are.

docs/Compile.rst
TheBloke: No discussion has occurred, it's merely my own strong recommendation.

Chocolatey installs the exact same files as one would manually, just it sets up PATHs etc correctly, installs any needed dependencies, provides ongoing application management (upgrades etc), and is a useful system to have in place for any more complex requirements eg when I reference it under Docs as a one-line install for Python/Pip for Sphinx.

Chocolatey has been around 4.5 years and has 3000+ unique packages with a pretty active community, so it's pretty stable.  And we've had known issues in #dfhack with people not getting PATHs set up right, so known-good config is a strong benefit in my view.  So I'd have to say it's little different to recommending apt-get as the standard for Ubuntu. 

But I can tone down the nature of the recommendation if you think it's necessary.

docs/Compile.rst
TheBloke: OK I'm a bit confused about what needs to be said here.

My understanding was that:
* If we're distributing, then currently only 4.5 should be used until such time as we look at libstdc++ distribution
* If it's for personal, anything >= 4.5 is fine, right up to current 5.2.1 and, according to GCC's stated ABI compat, beyond for the foreseeable future
  *  and there's a strong benefit to using a later version on modern distros, simply in that it's installed by default and one doesn't need to spend the time working out how to install an older version, mess with the alternatives system, etc.

I'm not sure where a 4.7 or 4.8 or anything-other-than-4.5 fits in?  If you have to install an old version of GCC, surely it's only for distribution purposes and therefore must be 4.5?  And if you're not distributing, you can use any later version and it's easier for you if you do so.  

So if a user happens to be on a 4.7, 4.8 or 4.9 system then that's fine, just as it is if they're on 5.x, but there's no reason anyone would want to install 4.x unless it's 4.5 for distributing?  So is there a need to mention any version besides 4.5 still?

(As an aside, it just occurred to me that it might be useful to have a new "Note on using GCC 5.x" section which explicitly mentions the ABI compatibility flag that cmake is applying.  This relates to potential new contributors, who must be aware that they need to compile their own plugins with that flag if they're testing on GCC 5 and plan to work with DFHack.  That might belong in Contributing.rst, instead of or as well as this Compile.rst,)

docs/Compile.rst
TheBloke: I've made it vague and general in the next commit, merely stating that SP1 "has been reported to cause issues" and is "not supported by DFHack".

docs/Compile.rst
TheBloke: OK done in next commit

docs/Compile.rst
TheBloke: Just removed the extra "On Debian" line in next commit

docs/Compile.rst
lethosor: Sorry, I was primarily talking about distribution there (on Linux only). Pretty much any linux distro at this point should default to GCC 4.8 or newer, so compiling with GCC 4.5-4.8 should result in binaries that work on a majority of distros. We ran into issues in the past where Arch moved to GCC 4.9 but Debian stayed at 4.8, for example, and in that case GCC 4.9 wasn't even available in apt, so a couple DFHack builds broke completely on a lot of distros.
In short:
* GCC 4.5 produces builds that should work for everyone, but probably needs to be compiled from source or obtained from a third-party package repo
* GCC 4.6-4.8 (and maybe 4.9 by now?) produce builds that work for nearly everyone (with a new enough libstdc++), but users have to get rid of DF's libstdc++
I haven't investigated the availability of GCC 5 yet, but I'd say sticking with 4.5-4.8 (or 4.9) is best for releases for now.

docs/Compile.rst
lethosor: I don't know how much time/space it takes to install, though - we've also had people complain about the number/size of dependencies on Windows, so I don't want to make it sound like additional ones are required when they technically aren't. (It sounds useful, though, so do encourage it.)

docs/Compile.rst
lethosor: (Of course, the system GCC is fine for personal builds as long as you get rid of DF's libstdc++.)

TheBloke: OK great.  I'm pretty sure what I have covers that already, but let me know if it isn't clear or needs more explanation.

Or rather, what I have basically says "for now only 4.5 for distribution, anything >=4.5 for personal"

So I could clarify that actually 4.6, 4.7 and 4.8 might be OK for distribution, but I'm not sure that's necessary given distribution only needs to be available for a few people, all of whom have 4.5 available anyway.

docs/Compile.rst
TheBloke: Chocolatey itself is using 18.5MB of disk space on my system with 20 or so packages installed.

I'd guess it takes maybe five minutes to install including opening/closing the cmd.exe windows etc.  Definitely significantly less time than it takes to install the three individual dependencies, factoring in time to go to three URLs, find the download link, download them, double click on them, click Next through the various installer wizards, then manually add them to PATH where not done already (Perl definitely needs manually adding, CMake doesn't, not sure about Git.)   

I'd say it's maybe 5 minutes total time for the Chocolatey + all dependencies install, versus at least 10 for the manual method, maybe 15.

I'd say it has about the same impact in terms of space and time as Homebrew or MacPorts on OSX - really not any at all.

But I'll reword it to highlight it's not actually required. Maybe 'preferred' instead of recommended.

TheBloke: OK I was curious so I timed it on a separate PC.

3:43 to install Chocolatey + and three dependencies.  Of that, 45 seconds was the time to install Chocolatey itself, and about two minutes was the time it took to download the 75MB StrawberryPerl package.

Basically it's just as fast and easy as you'd expect any binary-based package manager to be.

TheBloke: Incidentally, if we're ever able to move to MSVC >2010, there are Chocolatey packages to silently install the free versions of Visual Studio versions 2012->2015.

I could probably make and submit a Chocolatey formula to do that for 2010 as well, however the version we found was distributed as an ISO and I'm not sure if Windows 7 can auto-mount that (I haven't yet checked to see if Chocolatey has other ISO-based formula and therefore if they have a way to deal with that.)

But definitely if we ever get off 2010, Windows can have all its dependencies handled in a single command line same as Linux and OSX.

TheBloke: OK updated it in the next commit, now it is "a preferred way of installing the required dependencies".  And I put in a brief note on the very small disk requirements (I checked on the Windows 7 install in which I installed Chocolatey + just the 3 DFHack dependencies, and disk usage is 7.5MB there.)  And similarly removed the strong recommendation for Chocolatey under the Manually Installing Dependencies section.

PR #772, opened 2015-12-19
Some simplification to eventful

warmist:
Fixed eventful and other places where it was broken.

### DIFF
--- a/plugins/eventful.cpp
+++ b/plugins/eventful.cpp
@@ -92,8 +92,6 @@ static bool is_lua_hook(const std::string &name)
 /*
  * Hooks
  */
-static void handle_fillsidebar(color_ostream &out,df::building_actual*,bool *call_native){};
-static void handle_postfillsidebar(color_ostream &out,df::building_actual*){};
 
 DEFINE_LUA_EVENT_NH_2(onWorkshopFillSidebarMenu, df::building_actual*, bool*);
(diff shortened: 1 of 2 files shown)


lethosor:
This has some conflicts. Also, can you rebase and update the xml submodule somehow? I think the produce() issues are fixed on develop now.

plugins/eventful.cpp
```diff
@@ -306,11 +287,11 @@ IMPLEMENT_VMETHOD_INTERPOSE(furnace_hook, fillSidebarMenu);
 
 struct product_hook : item_product {
     typedef item_product interpose_base;
-
+
```
lethosor: whitespace

plugins/eventful.cpp
```diff
@@ -306,11 +287,11 @@ IMPLEMENT_VMETHOD_INTERPOSE(furnace_hook, fillSidebarMenu);
 
 struct product_hook : item_product {
     typedef item_product interpose_base;
-
+    
     DEFINE_VMETHOD_INTERPOSE(
         void, produce,
         (df::unit *unit,
-         std::vector<void*> *unk,
+         std::vector<df::reaction_product*>* out_products,
```
lethosor: Would be nice to match whitespace style here (and below)

warmist:
Oh this is still unmerged? I'll look into it.

warmist:
This is no longer about the vmethod change but about general eventful simplification.

PR #776, opened 2015-12-23
Update Compile.rst windows section

moversti:
Need to install cmake.portable instead of cmake and also the XML::LibXML perl module.

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -361,9 +361,10 @@ To install Chocolatey and the required dependencies:
 
 * Paste in the Chocolatey install command and hit enter
 * Close this ``cmd.exe`` window and open another Admin ``cmd.exe`` in the same way
-* Run the following command::
+* Run the following commands::
 
-    choco install git cmake strawberryperl -y
+    choco install git cmake.portable strawberryperl -y
+    cpanm XML::LibXML
 
 * Close the Admin ``cmd.exe`` window; you're done!
 


docs/Compile.rst
```diff
@@ -361,9 +361,10 @@ To install Chocolatey and the required dependencies:
 
 * Paste in the Chocolatey install command and hit enter
 * Close this ``cmd.exe`` window and open another Admin ``cmd.exe`` in the same way
-* Run the following command::
+* Run the following commands::
 
-    choco install git cmake strawberryperl -y
+    choco install git cmake.portable strawberryperl -y
+    cpanm XML::LibXML
```
TheBloke: Can you explain what this is about?

What's the benefit to installing cmake.portable instead of cmake?

Also the CPAN line is definitely not necessary - the reason to install Strawberry Perl vs any other kind of Perl is that it provides both XML::LibXML and XML::LibXSLT already.  The build will work fine without the extra cpan command.

TheBloke: Also, the cpan command won't work until a new cmd.exe is opened because PATH won't be updated:

```
C:\Users\Media>choco install strawberryperl -y
Installing the following packages:
strawberryperl
By installing you accept licenses for the packages.

StrawberryPerl v5.20.1.1
 Installing StrawberryPerl...
 StrawberryPerl has been installed.
 The install of strawberryperl was successful.

Chocolatey installed 1/1 package(s). 0 package(s) failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

C:\Users\Media>cpanm XML::LibXML
'cpanm' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\Media>
```

TheBloke: OK I understand now re cmake.portable - Chocolatey doesn't properly install the PATH for CMake. I didn't notice on my original test system because I'd had an old version of CMake installed and it as still in path.

So yeah the line changing cmake to cmake.portable is good.

But the cpanm line is definitely not needed - the whole reason to install StrawberryPerl is that it provides LibXML already, and running cpanm will just update the package by a small amount; not necessary.

Could you update the PR to just have the cmake -> cmake.portable edit.

TheBloke: @moversti I have just done a compile on a fresh Windows 7 installation, just to be 100% certain of what I was saying on cpan.

I can confirm that for certain, cpanm step is not required.  StrawberryPerl provides XML::LibXML and XML::LibXSLT already, and the provided versions are fine to run the Perl needed for DFHack compilation.

I am wondering if maybe you have a second Perl install on your system, which is higher in PATH than StrawberryPerl?  So maybe you thought you had to run cpanm because you are in fact not using StrawberryPerl, and your other Perl install did not have XML::LibXML installed.

Could you let me know if you are going to re-commit to this PR to remove that.  If not, I will open a new PR to change cmake -> cmake.portable which definitely is required (thanks for raising that!)

Also I am going to update the Chocolatey instructions slightly, to mention that it may not be necessary to re-open a new cmd.exe window. I found in my Windows 7 testing just now that I did not have to, despite Chocolatey's warnings saying that I almost certainly would have to.  Maybe this warning applies only to Windows XP or other, older Windows.

moversti:
Didn't compile without cpan on my new install of windows 10 with StrawberryPerl. I noticed something curious during the cpanm XML::LibXML as it actually downgraded the module some minur version number and after that it compiled fine.

TheBloke:
OK not sure what's going on there - I've done it now on three systems, one Windows 10 and two Windows 7, and all have been fine.  I did a fresh install this evening so I definitely got the latest StrawberryPerl.

It's also very strange if you're seeing an XML package downgrade, as I see an upgrade from 2.0116 to 2.0123 if I run cpanm on a StrawberryPerl install.

So you installed StrawberryPerl through Chocolatey and definitely don't have any other Perls installed?  

I'll keep an eye on it, see if anyone else reports issues.

PR #784, opened 2016-01-02
clean up a lot of warnings

BenLubar:
~~md5: changed code that casted back and forth between uint32_t* and
unsigned char* to use unions.~~

protobuf: ignored warnings generated by compile-time assertions that
succeeded.

ignored warnings in DataStaticsFields.cpp caused by generated code

in several places, fixed things being declared as static in headers.
static things are not shared between compilation units.

put several initializers into the same order as their fields.


depends/md5/md5.h
```diff
@@ -23,11 +23,20 @@
...
+        unsigned char bits[8];
+        unsigned char in[64];
+    } byte;
+    struct
+    {
+        uint32_t buf[4];
+        uint32_t bits[2];
+        uint32_t in[16];
+    } word;
 };
```
lethosor: This is Undefined Behavior™ (I think - there's some rule against accessing data in an instance of a union through more than one field, e.g. accessing byte.buf and word.buf in the same instance, although it's interesting that GCC would warn about conversions between a `char*` and `uint32_t*` but not that).

library/CMakeLists.txt
```diff
@@ -233,6 +233,10 @@ IF(UNIX)
   # Don't produce debug info for generated stubs
   SET_SOURCE_FILES_PROPERTIES(DataStatics.cpp DataStaticsCtor.cpp DataStaticsFields.cpp
                               PROPERTIES COMPILE_FLAGS "-g0 -O1")
+
+  # Don't emit thousands of warnings
+  SET_SOURCE_FILES_PROPERTIES(DataStaticsFields.cpp PROPERTIES COMPILE_FLAGS
+    "-Wno-missing-field-initializers -Wno-delete-non-virtual-dtor")
```
lethosor: Have you checked (e.g. with `make VERBOSE=1`) to make sure this doesn't overwrite the `-g0 -O1` from above?

library/Core.cpp
```diff
@@ -1733,7 +1723,9 @@ void Core::Suspend()
 
 void Core::Resume()
 {
+#ifndef NDEBUG
     auto tid = this_thread::get_id();
+#endif
     lock_guard<mutex> lock(d->AccessMutex);
 
     assert(d->df_suspend_depth > 0 && d->df_suspend_thread == tid);
```
lethosor: `assert(d->df_suspend_depth > 0 && d->df_suspend_thread == this_thread::get_id());` might be cleaner than the NDEBUG guard above (also around line 1767).

depends/md5/md5.h
BenLubar: Ok, I'll skip this one.

library/CMakeLists.txt
BenLubar: It seems to pick one at random. I'll put both things for DataStaticsFields.cpp in the same SET_SOURCE_FILES_PROPERTIES.

library/Core.cpp
BenLubar: I agree.

BenLubar:
Alright, I'm going to redo this one from scratch. You can ignore the commits I put here until I'm all done.

PR #785, opened 2016-01-06
blueprint: fix compilation (i.e. undefined find)

abbradar:
Without this, `std::find` is not found, leading to the same problem as in https://stackoverflow.com/questions/27996423/how-do-i-pass-a-constant-string-vector-to-a-function-for-string-matching .

### DIFF
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -2,6 +2,8 @@
 //By cdombroski
 //Translates a region of tiles specified by the cursor and arguments/prompts into a series of blueprint files suitable for digfort/buildingplan/quickfort
 
+#include <algorithm>
+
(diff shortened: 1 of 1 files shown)


lethosor:
What compiler are you using? This has been compiling for several releases now.

abbradar:
GCC 4.9 on NixOS -- I'm remaking a `dfhack` package. I've found the problem -- it was caused by me replacing some of your bundled dependencies (`protobuf`, `tinyxml` and `jsoncpp`) with system libraries. I believe newer version of one of them removed `<algorithm>` from its headers...

This patch then would be the right thing to do anyway, but feel free to ignore it -- this is my own changes kicking in, after all.

abbradar:
Thanks!

PR #789, opened 2016-01-09
New autogems plugin, automatically cutting rough gems.

eswald:
This has been requested a few times, partly because it's tedious to set up with Workflow and even more tedious with Stockflow.


plugins/autogems.cpp
```diff
@@ -0,0 +1,293 @@
...
+            INTERPOSE_NEXT(feed)(input);
+        }
+    }
+    
+    DEFINE_VMETHOD_INTERPOSE(void, render, ()) {
+        INTERPOSE_NEXT(render)();
+        if (in_menu()) {
+            auto dims = Gui::getDwarfmodeViewDims();
+            int x = dims.menu_x1 + 1;
+            int y = dims.y1 + 12;
```
lethosor: There are a couple standing orders that don't show up unless there are reactions that use them  - "auto smelter" and "auto other", I believe. Does this avoid overwriting them?

eswald: No, it doesn't.  Good catch.

lethosor: Just curious - do you know of an easy way to get those options to show up (for testing purposes)?

eswald: I added [AUTOMATIC] to a bunch of reactions in the raws of an existing saved game.

lethosor: It turns out that this works well in-game, for the record:
```lua
for _,r in pairs(df.global.world.raws.reactions) do r.flags.AUTOMATIC=true end
```

eswald: (I wish the gem reactions were as simple to make automatic.  Unfortunately, they're still hardcoded.)

PR #808, opened 2016-01-29
Fix floodgate allocation in C++/Lua and well and pressure plate allocation in Ruby.

BenLubar:


library/modules/Buildings.cpp
```diff
@@ -369,6 +370,12 @@ df::building *Buildings::allocInstance(df::coord pos, df::building_type type, in
                 obj->ready_timeout = 500;
             break;
         }
+    case building_type::Floodgate:
+        {
+            auto obj = (df::building_floodgatest*)bld;
```
lethosor: virtual_cast would be better here.

BenLubar: Should I change the other cases as well?

lethosor:
Turns out this was the issue that kept floodgates from working in buildingplan as well.

library/modules/Buildings.cpp
warmist: Isn't the static_cast correct way of casting in this case?

PR #810, opened 2016-02-01
Colonies as a script, + place feature by Warmist

PeridexisErrant:
See #722 

I have no idea where the index and merge commits come into it, so feel free to cherry-pick a3d9507


scripts/colonies.lua
```diff
@@ -0,0 +1,82 @@
...
+    verm.pos:assign(pos)
+    df.global.world.vermin.colonies:insert("#", verm)
+    df.global.world.vermin.all:insert("#", verm)
+end
+
+local args = {...}
+local target_verm = args[2] or "HONEY_BEE"
+
+if args[1] == 'help' or args[1] == '?' then
+    print(help)
```
lethosor: Dumping `=begin` and `=end` to the console isn't great behavior. A couple scripts strip them manually, but is it possible to avoid using those markers in lua scripts? (Lua does support variations on the `[[` and `]]` string markers, like `[=[` and `]=]`, if that helps.)

PeridexisErrant: Yep; it's just a matter of adding an alternative template to [conf.py](https://github.com/DFHack/dfhack/blob/develop/conf.py#L59) and having it filled in for `.lua` files and updating every script to use the new markers.

I'd be happy to do this, but I'm not seeing much urgency when I've got four pull requests waiting to be merged - all of them touch on the documentation somewhat, and I'd rather not have yet another active branch editing so many of the same files.  I'll open an issue for this though, so it doesn't get lost.

PR #813, opened 2016-02-02
uicommon cleanup

BenLubar:
This should get rid of a lot of the warning messages.

### DIFF
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -555,7 +555,7 @@ public:
     {
         watchedBurrows.clear();
         vector<df::burrow *> v = burrows_column.getSelectedElems();
-        for_each_<df::burrow *>(v, [] (df::burrow *b) { watchedBurrows.add(b->id); });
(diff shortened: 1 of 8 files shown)


lethosor:
I'm not sure what the benefit of getting rid of for_each_ and transform_ is - they're more concise than the standard library functions they wrap.

BenLubar:
We probably don't even need to use `for_each` or `transform` - all of the places they're used are simple 1-line callback functions that could just be part of a `for` loop.

lethosor:
Sure, but for_each and transform make it clearer what types are being iterated over and get rid of some cruft, like referencing the same container multiple times, (*it)->foo, etc.
If there are warnings related to these, fixing them is fine, but I don't think using for_each and transform is incorrect.

expwnent:
I agree with @lethosor. I like it better with `for_each` and `transform` unless they're causing warnings. The static -> inline stuff would be good though.

PR #818, opened 2016-02-11
Manipulator update and fix

sv-esk:
histfig.entity_links vector is one of things that determines ability to set unit's labors.
If it contains histfig_entity_link_memberst with entity_id equals to ui.group_id - player will be able to set labors in vanilla v-p-l menu otherwise menu shows "No labors available" text(assuming all other requirements are met).
histfig.entity_links vector also contains other types of entity links including histfig_entity_link_former_memberst. So type should be checked.
No more need in isOwnRace check since you can have not own race citizens.
flags2.bits.visitor and  antity links should be checked instead.
flags3.bits.ghostly should be checked - ghosts cant be edited in vanilla way, and they are not flags1.bits.dead.
Also added new 0.42.x skills to table.
Rebase fixupped one error, thanks warmist for pointing me out.

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -888,6 +888,22 @@ bool Units::isOwnCiv(df::unit* unit)
     return unit->civ_id == ui->civ_id;
 }
 
+// check if creature belongs to the player's group
+bool Units::isOwnGroup(df::unit* unit)
+{
+    CHECK_NULL_POINTER(unit);
+    auto histfig = df::historical_figure::find(unit->hist_figure_id);
+    if (!histfig)
+        return false;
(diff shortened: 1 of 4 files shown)


library/modules/Units.cpp
```diff
@@ -888,6 +888,22 @@ bool Units::isOwnCiv(df::unit* unit)
     return unit->civ_id == ui->civ_id;
 }
 
+// check if creature belongs to the player's group
+bool Units::isOwnGroup(df::unit* unit)
```
lethosor: This should be added to LuaApi.cpp too.

sv-esk: WRAPM(Units, isOwnGroup)? Done.

PR #826, opened 2016-02-24
Updating Stockflow for version 42.06

eswald:
The manager can now specify each type of craft individually, as well as in general.
There may be a better justification for not allowing pearl and shell scepters, though.

### DIFF
--- a/plugins/lua/stockflow.lua
+++ b/plugins/lua/stockflow.lua
@@ -272,6 +272,8 @@ function collect_reactions()
     materials.wood.adjective = "wooden"
     materials.tooth.adjective = "ivory/tooth"
     materials.leather.clothing_flag = "LEATHER"
+    materials.shell.short = true
+    materials.pearl.short = true
 
     -- Collection and Entrapment
     reaction_entry(result, job_types.CollectWebs)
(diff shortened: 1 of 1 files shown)


plugins/lua/stockflow.lua
```diff
@@ -527,6 +530,14 @@ function collect_reactions()
...
                     {job_types.MakeChain, "Forge", "Chain"},
                     {job_types.MakeCrafts, "Make", "Crafts"},
+                    {job_types.MakeFigurine, "Make", "Figurine"},
+                    {job_types.MakeAmulet, "Make", "Amulet"},
+                    {job_types.MakeScepter, "Make", "Scepter"},
+                    {job_types.MakeCrown, "Make", "Crown"},
+                    {job_types.MakeRing, "Make", "Ring"},
+                    {job_types.MakeEarring, "Make", "Earring"},
+                    {job_types.MakeBracelet, "Make", "Bracelet"},
+                    {job_types.MakeGem, "Make Large", "Gem"},
```
lethosor: Would "Make", "Large Gem" make more sense here? I'm not entirely sure what these two fields are (on mobile), but the item's name is "large gem".

eswald: The split is where the material's name goes.  In the manager, the line is "Make Large iron Gem" or "Make Large wooden Gem" and I'm replicating that here.  Technically, the "Large" corresponds to the adjective field in itemdefs, as seen on trap components and large pots, but crafts seem to be hard-coded.

lethosor: Ah, that makes sense.

Issue #835, opened 2016-03-05
teledwarf leaves babies lying on ground


PeridexisErrant:
http://www.bay12forums.com/smf/index.php?topic=139553.msg6852653#msg6852653

>>> sorry for referring to 0.40.26 still but tele-dwarves kinda cause babies to be abandoned, random babies lying around most of the time.  either they don't hunger/thirst or my absolute laziness waiting for the right phase of game to "be normal" on one of my worlds keeping teledwarf on is having dwarves provide... but they're perfectly fine and even grow up.

>> Considering the way teledwarf works (by checking every unit on every tick to see if they're pathing, then teleporting them to the path destination should there be one), that doesn't surprise me in the slightest. "Teleporting" here refers to a process where all occupancy flags at the beginning and ending locations are set up properly for the new positions, the teleporting unit's laying-down flag is set as appropriate for if there's a standing creature at the destination, and the teleporting creature's position is set to the destination. Note that none of that accounts for babies, which I have no idea how the game keeps track of babies being held (some sort of general ref?)

>babies are counted as riding the mothers, I believe.

This is probably fixable by checking for creatures riding a teleporting unit, and if found teleporting them to the same location (and ensuring they're still riding).

sv-esk:
> babies are counted as riding the mothers, I believe.

Indeed. 
Mother has `flags1.bits.ridden` 
Baby has `flags1.bits.rider` and `relations.rider_mount_id`
And it looks like all riders are stored in `world->units.other[units_other_id::ANY_RIDER]`

So I propose check unit's `flags1.bits.ridden` and teleport units from `world->units.other[units_other_id::ANY_RIDER]` with matching `relations.rider_mount_id` to its pos.

expwnent:
Closed by #839.

PR #839, opened 2016-03-10
fix teledwarf bug (leaves babies lying on ground)

sv-esk:
fix https://github.com/DFHack/dfhack/issues/835
Mother has `flags1.bits.ridden` 
Baby has `flags1.bits.rider` and `relations.rider_mount_id`
And it looks like all riders are stored in `world->units.other[units_other_id::ANY_RIDER]`

So I propose check unit's `flags1.bits.ridden` and teleport units from `world->units.other[units_other_id::ANY_RIDER]` with matching `relations.rider_mount_id` to its pos.

### DIFF
--- a/plugins/fastdwarf.cpp
+++ b/plugins/fastdwarf.cpp
@@ -10,6 +10,7 @@
 #include "df/unit.h"
 #include "df/unit_action.h"
 #include "df/map_block.h"
+#include "df/units_other_id.h"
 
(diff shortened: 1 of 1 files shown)


PeridexisErrant:
@sv-esk - for future reference, it would be good to add a note to `changelog.rst` :)

lethosor:
`NEWS.rst`

PR #840, opened 2016-03-10
Fix persistent histfig crashbug.

expwnent:


NEWS.rst
```diff
@@ -89,6 +89,8 @@ Fixes
 
 - `showmood`: Fixed name display on OS X/Linux
 
+- Fixed a crash bug caused by the historical figures DFHack uses to store persistent data.
```
lethosor: Most of the non-tool-specific entries are at the top of this section. (Also, it looks like there's a blank line above this one.)

expwnent: Fixed.

PR #887, opened 2016-04-09
actual information

PF4Public:
For completeness sake


lethosor:
Looks like you got author information from the master branch instead of develop.
Honestly, I'm not sure that we need author information on that page anymore. We don't really use it for anything (although we don't want to break old links that might be pointing to it), so we could just replace the list with a link to dfhack.readthedocs.org.

PF4Public:
Quite reasonable. One more commit then.

PR #902, opened 2016-04-19
item-trigger.lua fix

AtomicChicken:
This should allow item-trigger to correctly handle items lacking a subtype.


scripts/modtools/item-trigger.lua
```diff
@@ -64,10 +64,18 @@ function processTrigger(command)
  dfhack.run_command(table.unpack(command2))
 end
 
+function getitemType(item)
```
lethosor: This stuff in this function should be indented.

PR #903, opened 2016-04-20
Request for adding exportmaps plugin to DFHack

ragundo:
For approval and, hopefully, addition to DFHack.

The plugin has a github repo in https://github.com/ragundo/exportmaps with the corresponding doc.

For the moment, Windows and Linux only.
Tested in Ubuntu 15.10 64 bits
Windows XP 32 bits
Windows 7

Added support for OSX but untested


PeridexisErrant:
Nice, I really like this idea - you're absolutely right that even `exportlegends` takes ages!

I'll leave code-related comments to someone qualified, but regarding documentation and compatibility with utilities which use exported maps:

- It would be really really nice if maps identical to the vanilla maps had the vanilla names.  This ensures that Isoworld, all the various legends viewers, Dorven Realms, and any others I've forgotten will still work - it would be a real shame to thrown that away.

- Adding more exports (eg the true-heightmaps) is great; making sure the complete set of vanilla maps is available would also be good (see above)

- Your documentation is great!  It would be really nice if it could use `.rst`, ReStructuredText markup - it could then be easily incorporated into the online (or offline) DFHack docs.  Look at how Stonesense does it for an example, and if you get stuck I'm happy to fix any problems - it's just more elegant to fix *before* merging.

ragundo:
OK. I changed the names to mimic the DF original ones (but the extension will be .PNG)

About the heightmaps, the DF ones are already true heightmaps, only elevation and elevation respecting water aren't correct, so, I generated them.

About the doc, I'll keep in my mind for the future.

Thanks for your constructive comments PeridexisErrant!!

lethosor:
Following up on what I mentioned on IRC: 
I made a fork of your plugin at https://github.com/DFHack/exportmaps. It looks like the commits you've made here aren't in your repo, though. Could you add them there?

ragundo:
The code is the same at both repositories.
As I'm very bad at git, I keep one copy of the code in /dfhack/plugins/exportmaps (my fork of dfhack)
and another in a differente folder, which is the one that have it's own repository.
I commited in the dfhack fork (for updating the pull request) and, after that I copy the files to the external folder and commit again to my github repository for the plugin
Not very elegant and for sure that it can be done more efficiently, but if the plugin is linked as a submodule, I'll only need to commit to my repository.

lethosor:
From IRC, this only supports 32-bit DF right now. If anyone is motivated to work on this besides Ragundo, there's a DFHack fork of the plugin (DFHack/exportmaps, linked above); otherwise, this is probably on hold for now.

ragundo:
I close this PR as it would not work in 64 bits DF

PR #904, opened 2016-04-20
Allow arbitrary syndromes for add-thought

Putnam3145:
I mean, that seemed like something obvious to do.

### DIFF
--- a/scripts/add-thought.lua
+++ b/scripts/add-thought.lua
@@ -14,13 +14,26 @@ local utils=require('utils')
 
 function addEmotionToUnit(unit,thought,emotion,severity,strength,subthought)
     local emotions=unit.status.current_soul.personality.emotions
-    if not (type(emotion)=='number') then emotion=df.emotion_type[emotion] end
-    if not (type(thought)=='number') then thought=df.unit_thought_type[thought] end
+    if not (type(emotion)=='number') then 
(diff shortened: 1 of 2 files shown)


scripts/add-thought.lua
```diff
@@ -14,13 +14,26 @@ local utils=require('utils')
...
+        emotion=df.emotion_type[emotion] 
+    end
+    local properThought=nil
+    if not (type(thought)=='number') then 
+        properThought=df.unit_thought_type[thought] 
+        if not properThought then
+            for k,syn in ipairs(df.global.world.raws.syndromes.all) do
+                if syn.syn_name==thought then
+                    properThought=df.unit_thought_type['Syndrome']
+                    subthought=syn.id
```
warmist: Seems like it's missing break here.

lethosor: @Putnam3145?

Putnam3145: Yeah, that should probably have a break there. I didn't really consider it at the time since the performance gain is negligible (only whenever add-thought is run, which shouldn't be necessary except directly from DFHack stuff since the proper creature effect was added), but I can't in particular think of any issues it could cause.

lethosor:
Merged manually into dfhack/scripts

PR #912, opened 2016-04-30
Add a way to insert primitive types in gui/gm-editor

warmist:
Allow to insert ints and other simple types. Also allow inserting of same base type as the container.
Also simplify it a bit.

### DIFF
--- a/scripts/gui/gm-editor.lua
+++ b/scripts/gui/gm-editor.lua
@@ -211,22 +211,26 @@ function GmEditorUi:find_id()
 end
 function GmEditorUi:insertNew(typename)
     local tp=typename
-    if typename== nil then
-        dialog.showInputPrompt("Class type","Input class type:",COLOR_WHITE,"",self:callback("insertNew"))
-        return
-    end
-    local ntype=df[tp]
-    if ntype== nil then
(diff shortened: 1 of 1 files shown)


scripts/gui/gm-editor.lua
```diff
@@ -211,22 +211,26 @@ function GmEditorUi:find_id()
...
     local tp=typename
-    if typename== nil then
-        dialog.showInputPrompt("Class type","Input class type:",COLOR_WHITE,"",self:callback("insertNew"))
-        return
-    end
-    local ntype=df[tp]
-    if ntype== nil then
-        dialog.showMessage("Error!","Type '"..tp.." not found",COLOR_LIGHTRED)
+    if typename == nil then
+        dialog.showInputPrompt("Class type","Input class type or leave empty for auto and '*' for new ptr type auto:",COLOR_WHITE,"",self:callback("insertNew"))
```
warmist: This message seems a bit long, has anyone got any ideas how to simplify it?

lethosor: Yeah, it's 72 characters, which is close to the minimum screen width. You could put in a few newlines, I guess. (That would have the advantage of allowing you to explain things more thoroughly too - "new ptr type auto" took me a while to figure out.)

PR #901, opened 2016-04-19
Replace fixpositions plugin with scripts

PeridexisErrant:
See issue #722 - this is an experimental pull for comment, do not merge!

PR #917, opened 2016-05-08
Replace fixpositions plugin with scripts

warmist:
Building on PeridexisErrant work. Few bugs fixed. Not sure about the ESTABLISH_COLONY_TRADE_AGREEMENTS but it must be either nowhere or in both places or else it fixes the entity each time


PeridexisErrant:
Closes #901, when this is merged.

PR #921, opened 2016-05-10
improve windows build scripts

BenLubar:
They all pass the exit status on to the parent. build-\* now pauses for both debug and release, and new build-*-nopause scripts have been added for use by automated builders.

Additionally, the generate-MSVC-\* scripts no longer use removed flags and documentation has been added to the release build.

I didn't touch the breakfast script because I'm not sure what it's for.


lethosor:
There was some discussion regarding whether it's better to have separate -nopause scripts or a `nopause` argument to the existing scripts. Not sure what other devs think - @expwnent, @warmist?

warmist:
I don't use these as it's usually easier to have everything explicit for online build tools. Also when i use them i start a new console (with correct paths etc...) and run cmake in console and then use IDE. So not sure how other people are using this.

lethosor:
I'll try to merge these in manually, since we're handling build system changes anyway.

Regarding the -nopause variants: I just noticed that there are only two, build-debug and build-release, and build-release (but not build-debug) already pauses at the end. I don't have a problem with adding pause/nopause variants, then - I must have thought it applied to more of the generate-* scripts at first glance.

expwnent:
Having variants with and without nopause saves a few seconds closing the terminal if the build passes. If it paused when there's an error so you can see the message that would cover most of what I use it for, but I might want to see warnings without rejiggering cmake variables. I think the simplest solution is to keep the pause / nopause variants in.

PR #924, opened 2016-05-15
Updating Stockflow for v0.43

eswald:
New manager order flags, new manager interface behavior, and no limit on order number.

### DIFF
--- a/plugins/lua/stockflow.lua
+++ b/plugins/lua/stockflow.lua
@@ -798,6 +798,11 @@ screen = gui.FramedScreen {
 function screen:onRenderBody(dc)
     -- Emulates the built-in manager screen.
     
+    if not (self.page_size == self.frame_rect.height - ExtraLines) then
+        -- The screen size has changed.
+        self:refilter()
+    end
+    
     -- Top instruction line.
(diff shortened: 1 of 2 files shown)


plugins/lua/stockflow.lua
```diff
@@ -797,10 +797,30 @@ screen = gui.FramedScreen {
...
     dc:string(" to abort.", COLOR_WHITE)
-    dc:seek(1, PageSize + 5):string(self.search_string, COLOR_LIGHTCYAN)
+    
+    -- Search term, if any.
+    dc:seek(1, FirstRow + self.page_size + 1):string(self.search_string, COLOR_LIGHTCYAN)
+    
+    -- Bottom instruction line.
+    dc:seek(1, FirstRow + self.page_size + 2)
+    dc:string(gui.getKeyDisplay("STANDARDSCROLL_UP"), COLOR_LIGHTGREEN)
+    dc:string(gui.getKeyDisplay("STANDARDSCROLL_DOWN"), COLOR_LIGHTGREEN)
```
lethosor: dc:key() might be simpler here.

eswald: True; thanks for pointing it out.

PR #927, opened 2016-05-17
Persistent JSON

expwnent:
Replace histfig-based persistent data with json-based persistent data in a way that's efficient and automatically converts old histfig data to json for old saves


expwnent:
@lethosor Please check that this works on OSX.

depends/jsoncpp/CMakeLists.txt
```diff
@@ -1,2 +1,9 @@
 PROJECT(jsoncpp)
+
+SET(JSONCPP_HEADERS
+    jsoncpp.h
+    jsoncpp-ex.h
+)
+SET_SOURCE_FILES_PROPERTIES(${JSONCPP_HEADERS} PROPERTIES HEADER_FILE_ONLY TRUE)
+
```
lethosor: Sure this is shared?

library/modules/EventManager.cpp
```diff
@@ -250,6 +283,19 @@ void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event
...
+        DFHack::Persistent::clear(); //just in case
+        DFHack::Persistent::readFromFile(persistent_file);
+        /*savetime = DFHack::Filesystem::mtime(savegame);
+        std::string persistent_file = save_dir + "/dfhack_persistent_data.dat";
+        if ( DFHack::Filesystem::exists(persistent_file) ) {
+            boost::property_tree::json_parser::read_json(persistent_file, persistent_tree);
+        } else {
+            persistent_tree = pt::ptree();
+        }*/
+
```
eswald: You seem to be adding a few chunks of commented-out code near here.  Are they ideals to work toward in the future, or should they be deleted?

plugins/autogems.cpp
```diff
@@ -146,6 +190,9 @@ void create_jobs() {
             continue;
         }
 
+        if (workshop->isForbidden())
+            continue;
+
```
eswald: This appears to be an entirely unrelated change.  Meanwhile, how often do buildings get forbidden on purpose, instead of by accident?  How likely is this to surprise people?

plugins/autogems.cpp
```diff
@@ -281,14 +328,12 @@ IMPLEMENT_VMETHOD_INTERPOSE(autogem_hook, render);
...
-        if (enabled && World::isFortressMode()) {
-            // Determine whether auto gem cutting has been disabled for this fort.
-            auto config = World::GetPersistentData(CONFIG_KEY);
-            running = config.isValid() && !config.ival(0);
-            last_frame_count = world->frame_counter;
-        }
+        running = enabled && World::isFortressMode();
+        //if ( !World::isFortressMode() )
+        //    return CR_OK;
+        load_config(out);
```
eswald: Did you intend to change the behavior here?  Is this still safe in arena mode?  Did you mean to have the plugin up and enable itself if it had been active when the newly loaded fortress was last saved?

library/modules/EventManager.cpp
```diff
@@ -53,11 +58,29 @@ using namespace df::enums;
...
+        if ( !already_wrote_to_file ) {
+            already_wrote_to_file = true;
+            //out << "write_persistent" << endl;
+            write_persistent(out);
+        }
+        INTERPOSE_NEXT(write_file)(file);
+    }
+};
+IMPLEMENT_VMETHOD_INTERPOSE(nemesis_hook, write_file);
+
```
eswald: Clever.  It looks like this will write the file on seasonal autosaves too, right?  I'm less sure about whether it writes when saving the world after a paused autosave, in case configuration has been changed without unpausing first, but manageEvents should get called in between somewhere, right?

plugins/autogems.cpp
```diff
@@ -51,6 +57,44 @@ const char *usage = (
...
+        //load from property_tree
+        running = p["running"].isBool() ? p["running"].asBool() : false;
+    } else {
+        std::cerr << __FILE__ << ":" << __LINE__ << ": unknown version: " << version << endl;
+        exit(1);
+    }
+    running = running && World::isFortressMode();
+    if ( running )
+        plugin_self->plugin_enable(out,true);
+    Persistent::erase("autogems");
```
eswald: This erasure surprises me, though I guess it could clear out obsolete flags, and probably saves a bit of memory until the next seasonal autosave.  On the other hand, leaving the config in the Persistent tree would mean the whole on_presave_callback() dance could be omitted.  It's not like that flag gets toggled all that often.

plugins/autogems.cpp
```diff
@@ -51,6 +57,44 @@ const char *usage = (
...
+        PersistentDataItem config = World::GetPersistentData(CONFIG_KEY);
+        running = config.isValid() && !config.ival(0);
+        last_frame_count = world->frame_counter;
+        World::DeletePersistentData(config);
+    } else if ( version == 1 ) {
+        //load from property_tree
+        running = p["running"].isBool() ? p["running"].asBool() : false;
+    } else {
+        std::cerr << __FILE__ << ":" << __LINE__ << ": unknown version: " << version << endl;
+        exit(1);
```
eswald: Killing the whole Dwarf Fortress program over a corrupt autogems configuration seems extreme.

plugins/lua/eventful.lua
```diff
@@ -164,6 +164,7 @@ eventType=invertTable{
     "UNIT_ATTACK",
     "UNLOAD",
     "INTERACTION",
+    "PRESAVE",
     "EVENT_MAX"
```
eswald: Is there a reason this still uses the old EVENT_MAX name, instead of EVENT_MAX_EM?  For that matter, is there a reason for that rename?

plugins/autochop.cpp
```diff
@@ -170,38 +189,73 @@ static int string_to_int(string s, int default_ = 0)
...
-        max_logs = config_autochop.ival(2);
-        wait_for_threshold = config_autochop.ival(3);
-    }
-    else
-    {
-        config_autochop = World::AddPersistentData("autochop/config");
-        if (config_autochop.isValid())
-            save_config();
+    Json::Value& persist_autochop = Persistent::get("autochop");
+    int32_t version = persist_autochop["version"].isInt() ? persist_autochop["version"].asInt() : 0;
```
lethosor: This is what jsoncpp-ex.h is for - this could be replaced with ```cpp
int32_t version = Json::get<int>(persist_autochop, "version", 0);
```

File: plugins/autogems.cpp (hunk 7 above)
Comment (lethosor) [inline plugins/autogems.cpp]: Same for the other 12 instances. Also, you should be logging this to `out` (`out.printerr()` will print to the console and stderr).

**Comment by lethosor** [comment]
There's a bunch of the braces-on-the-same-line-as-statements style here too - I don't mind it myself (I tend to use it for other things), but we do have something about that in Contributing.rst...

**Comment by eswald** [comment]
Yes: "Opening and closing braces on their own lines or opening brace at the end of the previous line."  So either style is explicitly allowed.  Granted, my eye twitches at the braceless if statements, but that's not mentioned at all.

Overall, my worst objects are to the behavioral changes in the plugins.  Migrating persistent data to the new system could have been done without making half of them auto-enable themselves.  In that vein, I'm wondering if we can salvage this by splitting it into multiple commits:
https://github.com/eswald/dfhack/commit/72d881d75f2f78904f85b237318793b88d4bc8e0

**Comment by lethosor** [comment]
Oh, sure enough. I must have missed that change (359d739c8d624bc6a0a4145d5a55785ae95b6ed6).

File: plugins/autochop.cpp [hunk 10]
```diff
@@ -170,38 +189,73 @@ static int string_to_int(string s, int default_ = 0)
...
-static void initialize()
+static void load_config(color_ostream& out)
 {
+    //call this on_load, regardless of whether the plugin is enabled or not
+    //do NOT call this except on_load
     watchedBurrows.clear();
-    autochop_enabled = false;
+    //default values here
+    //autochop_enabled = false;
+    plugin_self->plugin_enable(out,false);
```
lethosor: A couple things:
- why not just use `plugin_enable(out, false)`?
- does it really need to be disabled?

plugins/autogems.cpp
lethosor: Do jobs work at forbidden workshops?
Agreed that this is unrelated, though.

plugins/autochop.cpp
eswald: I can see some justification for merging the is_enabled and autochop_enabled variables, but that kind of change shouldn't be mingled with the persistent data change.

lethosor: If the idea is that the plugin enables itself in worlds where it needs to, I think it makes sense (and I can see a reason for it if Expwnent's rewriting the code that checks whether it should be enabled).

depends/jsoncpp/CMakeLists.txt
expwnent: It's not shared. I gave up on that.

library/modules/EventManager.cpp
expwnent: They should be deleted. I missed that one.

plugins/autogems.cpp
expwnent: I'd have to check the original. I think the check was originally done somewhere else.

plugins/autogems.cpp
expwnent: I moved some stuff to load_config if I remember correctly.

library/modules/EventManager.cpp
expwnent: I think that happens in write_persistent.

plugins/autogems.cpp
expwnent: I wanted to do it the same way for different plugins for instructional purposes. If you have a lot of persistent memory you should really be keeping it in binary format instead of text / json.

plugins/autogems.cpp
expwnent: That was supposed to be temporary, thanks.

plugins/lua/eventful.lua
expwnent: The compiler was giving me a hard time because some other header conflicted with it. It could probably be reverted. I'll try it when I get a chance.

plugins/autochop.cpp
expwnent: Sounds good. I'll replace that everywhere.

expwnent:
It's not supposed to change behavior of persistent enabling. I merged the two variables because it was confusing to have two.

expwnent:
It is likely I will never finish this. If there's another one that does the same thing, such as https://github.com/DFHack/dfhack/pull/1402 or https://github.com/DFHack/dfhack/pull/955, and it is otherwise ready for merging, I would not be at all offended if that one were used and this one were closed. I don't remember how much work was left to be done before this one would be ready. If the same functionality can be achieved with less effort than completing this version of the feature, then that's a better way to do it.

lethosor:
Hi! Thanks for checking in. My plan is to get #1402 in for now, since it's fairly compact and addresses the issues of keeping data in fake histfigs (which can't handle binary data and is truncated at 16KB, among other things). It also keeps the existing API more or less intact for users of it, from what I can tell, which makes it less painful of a transition. Hopefully we can come up with a new JSON-based storage/config API that fits all of our use cases at some point.
This PR and #955 do have useful things in them, though. I'll see if I can manage to get some of those in once I've dealt with this.

expwnent:
I remember the hardest part was making lua interface well with this, replacing a certain lua file with an implementation in C++, so that would be a good candidate of what to take from this PR. I remember I did convert ALL plugins to use the new system, even though it was only a small optimization in size of DF saves, mainly just to illustrate how it's done for new plugins. There may have been some minor changes in plugin behavior but those were unintended.

It was also nice to have operator[] stuff for persistent data, but I think that was done for me by the json stuff we already had? I forget.

I thought I remembered doing a little extra stuff that I hadn't yet pushed, so I checked. It looks like I just deleted a comment in EventManager.cpp and uncommented `if ( !World::isFortressMode() ) return CR_OK` in autogems.cpp. I'm sure that can safely be thrown out.

That's all the summary information I can remember that might be useful for salvaging part of this.

myk002:
I tagged this PR in an issue so we can keep a reference to the cpp-based persist-table, but since it is not likely to happen anytime soon, I'll close this PR

PR #930, opened 2016-05-22
tinythread memory leak

BenLubar:
1. update to tinythread version 1.1, which provides a detach method
2. move the thread disassociation to the join function, which allows
attached threads to be joined even after they finish (this was the main
leak)

Also update RemoteServer.cpp to detach threads and delete the thread
objects instead of leaking them (although they are much smaller than the
actual threads).

### DIFF
--- a/depends/tthread/tinythread.cpp
+++ b/depends/tthread/tinythread.cpp
@@ -1,5 +1,5 @@
-/*
-Copyright (c) 2010 Marcus Geelnard
+/* -*- mode: c++; tab-width: 2; indent-tabs-mode: nil; -*-
+Copyright (c) 2010-2012 Marcus Geelnard
 
 This software is provided 'as-is', without any express or implied
 warranty. In no event will the authors be held liable for any damages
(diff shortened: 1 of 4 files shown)


BenLubar:
To explain this a bit further, you can reproduce the problem by creating a lot of connections to DFHack's RPC server:

```
for i in {1..10000}; do echo $i; ./dfhack-run ls > /dev/null; done
```

Without the patch, DFHack will run out of memory and accept the connection but not be able to create a thread, causing the connection to hang until Dwarf Fortress is restarted. With the patch, the loop eventually terminates because each connection handler thread gets cleaned up soon after it exits.

depends/tthread/tinythread.cpp
```diff
@@ -171,13 +171,17 @@ void * thread::wrapper_function(void * aArg)
   catch(...)
   {
     // Uncaught exceptions will terminate the application (default behavior
-    // according to the C++0x draft)
+    // according to C++11)
     std::terminate();
   }
 
+#if 0
```
lethosor: I might've missed a discussion about this, but is this something that should be fixed upstream, or is it DFHack-specific for some reason?

BenLubar: It should be fixed upstream.

PR #931, opened 2016-05-22
Advfort2

warmist:
Mostly fixes to advfort. There also a bug with workshops.lua that it's outdated. E.g. still is no longer a hardcoded building. Somebody should look the full list other :P

### DIFF
--- a/library/lua/dfhack/workshops.lua
+++ b/library/lua/dfhack/workshops.lua
@@ -126,19 +126,6 @@ jobs_workshop={
             job_fields={job_type=df.job_type.CatchLiveFish}
         }, -- no items?
     },
-    [df.workshop_type.Still]={
-        {
-            name="brew drink",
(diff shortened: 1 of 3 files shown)


library/lua/dfhack/workshops.lua
```diff
@@ -126,19 +126,6 @@ jobs_workshop={
             job_fields={job_type=df.job_type.CatchLiveFish}
         }, -- no items?
     },
-    [df.workshop_type.Still]={
```
lethosor: Does workshop_type need to be updated?

warmist: Nope this file (workshops.lua) is out of date. Though not much depends on it but there are some jobs that are no longer hardcoded.

PeridexisErrant:
A related issue is that [advtools](https://github.com/DFHack/dfhack/blob/master/plugins/advtools.cpp) needs updating (since 40.01).  Does any of your work here suggest how to fix it, or that it should perhaps be deleted for good?  (see also #735)

PR #932, opened 2016-05-22
Include math.h in 3dveins.cpp

carbohydratesn:
3dveins.cpp won't build for me without math.h as a dependency - errors about 'fabsf' and other math functions

arch linux, gcc-multilib 6.1.1-1


lethosor:
Merged manually into develop (9be9285)

PR #938, opened 2016-06-07
Convert catsplosion plugin to a script

PeridexisErrant:
Thanks to Putnam.  Closes issue #722, at last!


lethosor:
This looks like it does something similar to the plugin, but in a different way and lacking some features (e.g. only cats, no listing races, etc.)

scripts/catsplosion.lua
```diff
@@ -0,0 +1,58 @@
...
+    end
+    local ngenes = unit.relations.pregnancy_genes
+    local rng=dfhack.random.new()
+    local timer=rng:random(100)+1
+    unit.relations.pregnancy_timer=timer
+    unit.relations.pregnancy_caste=1
+    if unit.sex==1 then
+        local normal_caste=unit.enemy.normal_caste
+        unit.enemy.normal_caste=getFemaleCasteWithSameMaxAge(unit)
+        script.sleep(timer+1,'ticks')
```
lethosor: What's this for?

lethosor:
Okay, for reference: http://www.bay12forums.com/smf/index.php?topic=139553.msg7036221#msg7036221
It looks like this is an adaptation of Putnam's `impregnate` script that gets applied only to cats, which is great, but not quite the same as catsplosion. It appears to have some of the caste-replacement code as well, although I haven't investigated to see whether this applies to male cats as well.

lethosor:
Okay, I took care of this. It would be great to merge Putnam's script in if it's useful.

PR #937, opened 2016-06-05
Adding the dwarfvet plugin

NCommander:
dwarfvet is a plugin to fix and enable the built in animal caretaker behavior in Dwarf Fortress. I worked out the mechanics about a year ago, and started work on a plugin, but real life and personal issues got int he way. Now that I'm back in DF, I decided to dust off this project, update it to work against alpha, and fix most of the remaining "alpha" items I had such as crashes on Windows.

As of right now, it works, and works fairly well, though there are edgecases I need more eyes to look at (I can't tell if an animal will ever generate a traction request), and I'd like to know what the possibility is of actually getting this into DFHack officially.

I can add a "pet drop" fort somewhere to help speed up any testing. Here's the original thread on bay12. http://www.bay12forums.com/smf/index.php?topic=150251.0

PR #941, opened 2016-06-08
Dwarfvet: third time is the charm

NCommander:
Actually got the PR right this time.

Implements dwarfvet which allows animal caretakers to work.  Original PR is here https://github.com/DFHack/dfhack/pull/937  before I moved it to its own branch.


NEWS.rst
```diff
@@ -40,6 +40,7 @@ Lua
 New Features
 ------------
 - `gui/gm-editor` it's now possible to insert default types to containers. For primitive types leave the type entry empty, for references use ``*``.
+- `dwarfvet` enables animal caretaking.
```
lethosor: Should be under "new plugins" (which probably doesn't exist, but see the initial comment or previous releases for where it goes).

docs/Plugins.rst
```diff
@@ -1109,6 +1109,27 @@ Example:
     This will confiscate rotten and dropped food, garbage on the floors and any
     worn items with 'X' damage and above.
 
+.. _dwarfvet:
```
lethosor: I think this file is sorted alphabetically, so this should be after dwarfmonitor.

plugins/CMakeLists.txt
```diff
@@ -93,6 +93,7 @@ if (BUILD_SUPPORTED)
     DFHACK_PLUGIN(dig dig.cpp)
     DFHACK_PLUGIN(digFlood digFlood.cpp)
     add_subdirectory(diggingInvaders)
+	DFHACK_PLUGIN(dwarfvet dwarfvet.cpp)
```
lethosor: Indentation doesn't match - probably change this to spaces.

docs/Plugins.rst
NCommander: It's not at least according to the index on docs, but I reordered this.

PR #950, opened 2016-06-27
Support linking against an external tinyxml if EXTERNAL_TINYXML is set

TC01:
I've been working on [providing pre-built packages of dfhack for Fedora](https://www.acm.jhu.edu/~bjr/pages/dwarf-fortress-for-fedora.html). Fedora (and most other Linux distributions) tend to have a policy frowning on packaged software bundling dependencies, with various degrees of strictness. Fedora specifically these days simply asks that package maintainers publicly contact upstream about unbundling libraries when possible, which is what I am doing here.

Some of the projects in the [depends/](https://github.com/DFHack/dfhack/tree/develop/depends) tree seem to have been modified since they were included. As best as I can tell, the copy of tinyxml dfhack uses is unmodified from whenever it was first bundled, or at least since it was moved into the depends/ tree.

This commit adds an option to CMake, EXTERNAL_TINYXML, that if set to ON, will attempt to link against a system tinyxml instead of using the dfhack-bundled one.  It defaults to OFF, so there is no change in default behavior. The DFHACK_TINYXML variable is then set to either "tinyxml" or "dfhack-tinyxml" so the library (and any plugins that need updating) can link against one or the other. (stonensense would need updating if this patch were to be merged; I'm not sure if other out-of-tree plugins would too).

Note that if EXTERNAL_TINYXML is set and there is *not* a copy of tinyxml available on the system, the build will fail instead of falling bundled back to the bundled tinyxml. This could be changed, but I think it makes more sense to fail in this case so the user (who presumably knows what they're doing if they are setting this option) do not link against the wrong platform.

The FindTinyXML.cmake script was taken from https://github.com/ros/cmake_modules (licensed under the 3-clause BSD license).

This may very well be something no one is interested in, or indeed something that is actively *not* wanted, in case there's a desire to fork the bundled copy of tinyxml in the future. That would be fine; I am providing a patch here mainly because I don't want it to seem as if I'm simply here to complain without being willing to help out, too. :) Personally, I think it would be preferential to send patches upstream rather than forking things downstream, but I certainly recognize that;s not always possible (upstream might not be interested in carrying said patches, etc). Apologies in advance if I am inadvertently re-hashing an old argument.

(The name of the branch is "linux-system-tinyxml"; I have only tested that dfhack builds this way on Linux. Currently though I've not configured cmake to only listen to ExTERNAL_TINYXML on Linux, it can be set on any platform).


lethosor:
This looks fine to me. I think the main reason for putting it in depends/ was that it was fairly small and an easier way to distribute it at the time (although that choice was made years before I showed up).

It looks to me like Stonesense is the only "official" plugin using tinyxml at the moment. There's also exportmaps, which we're in the process of merging in. Anyway, if maintaining source compatibility is really important, something like this could work:
```cmake
add_library(dfhack-tinyxml ALIAS tinyxml)
```
This was implemented in CMake 2.8.12, though. It might be possible to fall back to implementing a static dfhack-tinyxml library that pulls in the external one somehow, but that's probably too much work, honestly.

I would be more comfortable actually including the CMake script's license somewhere - in the script itself is probably fine, since it's only used in the build process.

PeridexisErrant:
[LICENSE.rst](https://github.com/DFHack/dfhack/blob/master/LICENSE.rst) is the place for that; we've already got some BSD-licensed components so you can just follow the template.

lethosor:
My point was that the file is only included in source distributions, so the license doesn't really need to be included in binary distributions too.

TC01:
> This looks fine to me. I think the main reason for putting it in depends/ was that it was fairly small and an easier way to distribute it at the time (although that choice was made years before I showed up).

Great, good to know.

> It looks to me like Stonesense is the only "official" plugin using tinyxml at the moment. There's also exportmaps, which we're in the process of merging in. Anyway, if maintaining source compatibility is really important, something like this could work:

> This was implemented in CMake 2.8.12, though.

Hm. It looks like CMake 2.8.12 came out in October 2013? Do you think that's too new to take advantage of at the moment? It would probably be a cleaner solution than what I'm currently doing, but if only stonesense and exportmaps would need to be updated anyway, that's not so bad.

> I would be more comfortable actually including the CMake script's license somewhere - in the script itself is probably fine, since it's only used in the build process.

I have updated the commit including the full text of the license in the CMake script; I can add it to LICENSE.rst as well if that's desired.

PR #953, opened 2016-06-27
Update of autolabor2 for 42.06/43.30

ab9rf:
I dragged out my old autolabor2 plugin and updated it to 43.03 and renamed it to labormanager. There are still unmapped labors added since 34.11, but those only make the plugin behave less than optimally, rather than actually crash, so this still passes muster as a devel plugin.

Also, I fixed the flapping behavior that was seen in the 34.11 version so it's now Much Improved, too!

Most of the commits are just noise from me dragging my repo forward to modern times. There's actually only two files changed (the source and the makefile)


PeridexisErrant:
@ab9rf - please use `git rebase develop` to clean out the spurious commits :smile:

lethosor:
There are actually a bunch of commits here that somehow didn't get merged - e.g. a125b2f01dcc2be8ca5d69c988cc97be00a57ab3

ab9rf:
@PeridexisErrant rebasing generates enough errors and conflicts that it would take me a month to resolve them all. I'm not at all an expert at git. i'll just withdraw this until i can batter my repo into better shape, i guess;. sorry to have bothered you all.

PR #954, opened 2016-06-27
Rename autolabor2 to labormanager and bring up to date with current

ab9rf:
I dragged out my old autolabor2 plugin and updated it to 43.03 and renamed it to labormanager. There are still unmapped labors added since 34.11, but those only make the plugin behave less than optimally, rather than actually crash, so this still passes muster as a devel plugin.

Also, I fixed the flapping behavior that was seen in the 34.11 version so it's now Much Improved, too!

Redone this time with a lobotomized repo.


ab9rf:
I've been testing this for the last week or so and it works fairly well with 43.03, certainly as well as it ever did with 34.11.

lethosor:
Should it be moved to the stable plugin list, then? Or do you want to leave it where it is and move it later?

ab9rf:
Please leave it as a devel plugin for now. I'm still playtesting it, and that's a timeconsuming process. Having it in mainline devel makes it slightly easier for me to recruit playtesters in the community though.

lethosor:
Sounds good. If you want to make a build for playtesters, I can build for OS X and Linux (I've been doing OS X builds for df-ai too, so it's not a problem).

PR #956, opened 2016-06-29
Add protection from spaces in path and add warning in documentation

mlindner:


lethosor:
Merged into develop.

I also managed to get rid of the restrictions in eef66b0, which is what the Linux launcher script was already doing.

PR #969, opened 2016-08-01
Minor fixes/updates for Visual Studio 2015 Win64

Qartar:


CMakeLists.txt
```diff
@@ -67,7 +67,17 @@ add_definitions ( "/D_CRT_NONSTDC_NO_WARNINGS")
...
-SET(DFHACK_BUILD_ARCH "32" CACHE STRING "Architecture to build ('32' or '64')")
+# Automatically detect architecture based on Visual Studio generator
+IF(MSVC AND NOT DEFINED DFHACK_BUILD_ARCH)
+    IF(${CMAKE_GENERATOR} MATCHES "Win64")
+        SET(DFHACK_BUILD_ARCH "64")
+    ELSE()
+        SET(DFHACK_BUILD_ARCH "32")
+    ENDIF()
+ELSE()
+    SET(DFHACK_BUILD_ARCH "32" CACHE STRING "Architecture to build ('32' or '64')")
```
lethosor: The options here (`CACHE STRING "Architecture to build ('32' or '64')"`) are fairly important - they make this option show up in CMakeCache.txt and various GUIs (ccmake, cmake-gui, etc.). Does this change break that? I would leave the original line as-is, except that would break the part that detects whether DFHACK_BUILD_ARCH is already defined. Maybe something like this would work:

```cmake
SET(DFHACK_BUILD_ARCH "" CACHE STRING "Architecture to build ('32' or '64')")
IF(MSVC AND "${DFHACK_BUILD_ARCH}" STREQUAL "")
    IF(${CMAKE_GENERATOR} MATCHES "Win64")
        SET(DFHACK_BUILD_ARCH "64")
    ELSE()
        SET(DFHACK_BUILD_ARCH "32")
    ENDIF()
ELSE()
    SET(DFHACK_BUILD_ARCH "32")
ENDIF()
```

depends/protobuf/CMakeLists.txt
```diff
@@ -202,6 +202,9 @@ LIST(APPEND LIBPROTOBUF_FULL_SRCS ${LIBPROTOBUF_LITE_SRCS})
 IF(CMAKE_COMPILER_IS_GNUCC)
     SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -Wno-sign-compare")
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-result")
+ELSEIF(MSVC)
+    # Disable warnings for integer conversion to smaller type
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4267")
```
lethosor: We should fix these warnings instead - any specific ones you've seen in the DFHack core? I thought I fixed all of them when building with GCC without plugins, but maybe MSVC is being different. I haven't looked at non-dev plugins much yet.

library/LuaApi.cpp
```diff
@@ -2373,7 +2373,7 @@ static int internal_setAddress(lua_State *L)
 
     // Print via printerr, so that it is definitely logged to stderr.log.
     uintptr_t iaddr = addr - Core::getInstance().vinfo->getRebaseDelta();
-    fprintf(stderr, "Setting global '%s' to %x (%x)\n", name.c_str(), addr, iaddr);
+    fprintf(stderr, "Setting global '%s' to %zu (%zu)\n", name.c_str(), addr, iaddr);
```
lethosor: I'm not sure about `%zu` here - it does seem to be newly-supported in MSVC 2015 (yay), but wouldn't `%p` make more sense?

plugins/ruby/ruby.cpp
```diff
@@ -1023,7 +1023,7 @@ static VALUE rb_dfmemory_set_clear(VALUE self, VALUE set)
 
 
 /* call an arbitrary object virtual method */
-#ifdef WIN32
+#if defined(_WIN32) && !defined(_WIN64)
```
lethosor: Yeah, this one is... tricky. It might compile with this change, but it'll probably break the plugin in mysterious ways on 64-bit Windows. It *might* be possible to move the assembly to a separate file, but even then, it would probably need more changes to work on x64. @jjyg would know best here.

depends/protobuf/CMakeLists.txt
lethosor: Oh, protobuf. Yeah, this is fine.

pronvit: @lethosor, it's protobuf-only CMakeLists.txt, it's probably fine to suppress the warnings in this case rather than fix anything in 3rd-party code.

CMakeLists.txt
Qartar: This change breaks that when generating for Visual Studio but I believe it's irrelevant since you cannot build Win32 targets using the Win64 generator and vice versa.

lethosor: Oh, true. I guess you'd have to make a new build folder for a new generator too, so the option isn't really useful at all with MSVC.

library/LuaApi.cpp
Qartar: MSVC complains when using `%p` for `uintptr_t` types, which makes sense as `uintptr_t` is defined to be an unsigned integer type capable of storing a pointer value, not a pointer type _per se_.

plugins/ruby/ruby.cpp
Qartar: Microsoft's [documentation](https://msdn.microsoft.com/en-us/library/windows/hardware/ff561499.aspx) shows that calling conventions are effectively identical:

>Unlike the x86, the C/C++ compiler only supports one calling convention on x64. This calling convention takes advantage of the increased number of registers available on x64:
>* The first four integer or pointer parameters are passed in the **rcx**, **rdx**, **r8**, and **r9** registers.
>* The first four floating-point parameters are passed in the first four SSE registers, **xmm0**-**xmm3**.
>* The caller reserves space on the stack for arguments passed in registers. The called function can use this space to spill the contents of registers to the stack.
>* Any additional arguments are passed on the stack.

...

>The calling convention for C++ is very similar: the **this** pointer is passed as an implicit first parameter. The next three parameters are passed in registers, while the rest are passed on the stack.

However this does mean functions expecting floating-point arguments are going to receive garbage data.

Qartar:
@lethosor What should I do with this pull request? mifki on IRC suggested casting to `void*` and keeping %p instead of using %zu which I am planning to change. Is there anything else? Should I hold off on the changes to raw_vfcall until somebody else verifies that it's correct? (It doesn't compile otherwise).

lethosor:
The raw_vfcall change is probably okay for now - the calling conventions on x64 (and x86) should be consistent now. There could be other issues with the Ruby plugin on Windows that I can't test, but this shouldn't make them worse.

I saw the discussion on IRC, and yeah, that's probably the way to go.

Qartar:
I push an update to the pull request for the format strings. There are definitely still issues with Ruby; the Perl script blows up because of integer fields with subtype long not having ld:bits defined (from e8fe72826f4bda2e54940960787958d6b0a965a2?). I hacked a fix on my local workspace to keep developing but I'm not sure I could make a real fix without making a research dive into Ruby interop which I'm not that excited to do.

lethosor:
It's https://github.com/DFHack/df-structures/commit/290244046d1ff36ce09f077dfa5b37f2da085617 that's the issue, and it'll probably need some changes to the Perl script(s).

PR #970, opened 2016-08-01
Update the example init file

PeridexisErrant:
After updating the extra init file in my pack, I realised that many of
the changes could be merged into the standard init.

This enables a few more UI improvements, adds neglected keybindings, and
removes the roses-init and binpatches sections (third-party only and
obsolete respectively).

### DIFF
--- a/dfhack.init-example
+++ b/dfhack.init-example
@@ -48,6 +48,10 @@ keybinding add Ctrl-F@dwarfmode/Default "dwarfmonitor stats"
 # export a Dwarf's preferences screen in BBCode to post to a forum
 keybinding add Ctrl-Shift-F@dwarfmode forum-dwarves
 
+# an in-game init file editor
+keybinding add Alt-S@title gui/settings-manager
+keybinding add Alt-S@dwarfmode/Default gui/settings-manager
+
 ##############################
 # Generic adv mode bindings  #
 ##############################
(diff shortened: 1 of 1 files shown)


dfhack.init-example
```diff
@@ -177,6 +181,7 @@ tweak import-priority-category
 # Misc. UI tweaks
 tweak block-labors              # Prevents labors that can't be used from being toggled
 tweak civ-view-agreement
+tweak eggs-fertile
```
lethosor: I have mixed feelings about enabling this by default. I said that I considered it cheating when it came up for discussion when I added it, and most of the other tweaks enabled by default are bugfixes or improvements.

PeridexisErrant: I think of it as a UI improvement, but if you want me to I can remove it.

dfhack.init-example
```diff
@@ -223,7 +228,7 @@ enable \
 # You can comment out the extension of a line.
 
 # enable mouse controls and sand indicator in embark screen
-embark-tools enable sand mouse
+embark-tools enable sticky sand mouse
```
lethosor: This one's really a personal preference, in my opinion. It also has issues with selecting uninhabitable sites, which vanilla DF avoids.

dfhack.init-example
```diff
@@ -232,14 +237,11 @@ embark-tools enable sand mouse
 # write extra information to the gamelog
 modtools/extra-gamelog enable
 
+# extended status screen (bedrooms page)
+enable gui/extended-status
```
lethosor: I really ought to expand this one eventually...

dfhack.init-example
lethosor: I could make it possible to toggle in the UI, I suppose (like `tweak hide-priority`).

Issue #971, opened 2016-08-01
Add default keybindings to documentation


PeridexisErrant:
Each keybinding in `dfhack.init-example` should be mentioned in the documentation for the relevant command, including the context/s the binding is limited to.  Travis should test that (a) each keybinding is documented, and (b) that each documented keybinding exists.

It would be fairly easy to enforce an ugly and labor-intensive version via text-matching.  However I'd prefer to use a custom Sphinx directive (or role), which would enable some nice formatting and automatically stay up-to-date with the init file.

PeridexisErrant:
Ugh.  Took way longer than I'd like, but I have a working prototype.  Pull request in a few days once I've cleaned up and done the misc work to *use* the directive everywhere it should be, remove old text, and so on.

PeridexisErrant:
The directive outputs a Sphinx error if the referenced keybinding doesn't exist, which would make the build fail in Travis ((b) above).  I've added logic that auto-annotates script docs where appropriate, and logic to enforce the note for plugins shouldn't be hard.

PeridexisErrant:
Closed by #972.

PR #972, opened 2016-08-02
Document keybinds in dfhack.init-example

PeridexisErrant:
Closes issue #971 - DFHack/scripts#5 is the complementary pull.

This pull:

* adds a new directive to Sphinx, for keybindings, with most information pulled from `dfhack.init-example`
* automatically adds the directive to script documentation if there is a keybinding at build time
* adds the directive to plugin docs as appropriate (by hand)
* causes a build of the docs to fail if there is an undocumented keybinding, and warns if a documented keybinding no longer exists (which also fails the Travis build)
* updates `NEWS.rst`, including for `dwarfvet`
* removes an obsolete special case from `autodump` (since keybinds may now have spaces in the command if quoted).  Further work, to remove the `autodump-destroy-item` case, would be welcome but is beyond my C++ skills.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -15,14 +15,69 @@ serve to show the default.
 
 # pylint:disable=redefined-builtin
 
-import fnmatch
 from io import open
-from itertools import starmap
 import os
 import re
 import shlex  # pylint:disable=unused-import
 import sys
 
+
+# -- Support :dfhack-keybind:`command` ------------------------------------
+# this is a custom directive that pulls info from dfhack.init-example
+
+from docutils import nodes
+from docutils.parsers.rst import roles
+
+
+def get_keybinds():
+    """Get the implemented keybinds, and return a dict of
+    {tool: [(full_command, keybinding, context), ...]}.
+    """
+    with open('dfhack.init-example') as f:
+        lines = [l.replace('keybinding add', '').strip() for l in f.readlines()
+                 if l.startswith('keybinding add')]
+    keybindings = dict()
+    for k in lines:
+        first, command = k.split(' ', maxsplit=1)
+        bind, context = (first.split('@') + [''])[:2]
+        if ' ' not in command:
+            command = command.replace('"', '')
+        tool = command.split(' ')[0].replace('"', '')
+        keybindings[tool] = keybindings.get(tool, []) + [
+            (command, bind.split('-'), context)]
+    return keybindings
+
+KEYBINDS = get_keybinds()
+
+
+# pylint:disable=unused-argument,dangerous-default-value,too-many-arguments
+def dfhack_keybind_role_func(role, rawtext, text, lineno, inliner,
+                             options={}, content=[]):
+    """Custom role parser for DFHack default keybinds."""
+    roles.set_classes(options)
+    if text not in KEYBINDS:
+        msg = inliner.reporter.error(
+            'no keybinding for {} in dfhack.init-example'.format(text),
+            line=lineno)
+        prb = inliner.problematic(rawtext, rawtext, msg)
+        return [prb], [msg]
+    newnode = nodes.paragraph()
+    for cmd, key, ctx in KEYBINDS[text]:
+        n = nodes.paragraph()
+        newnode += n
+        n += nodes.strong('Keybinding: ', 'Keybinding: ')
+        for k in key:
+            n += nodes.inline(k, k, classes=['kbd'])
+        if cmd != text:
+            n += nodes.inline(' -> ', ' -> ')
+            n += nodes.literal(cmd, cmd, classes=['guilabel'])
+        if ctx:
+            n += nodes.inline(' in ', ' in ')
+            n += nodes.literal(ctx, ctx)
+    return [newnode], []
+
+
+roles.register_canonical_role('dfhack-keybind', dfhack_keybind_role_func)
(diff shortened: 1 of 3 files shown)


PeridexisErrant:
...Has Travis been disabled or something?

PeridexisErrant:
@lethosor - any idea why Travis is silent?

lethosor:
It was disabled temporarily because of the current state of the develop branch (e.g. the ruby plugin won't compile anywhere, besides the issues on Windows addressed in another PR). I'll see if there's a better way to check PRs, but I doubt this PR would pass the compilation stage. I can definitely test this locally, though (although it might be a couple days).

PeridexisErrant:
Fixed merge conflicts, improved `docs/Core.rst`

docs/Core.rst
```diff
@@ -21,14 +27,51 @@ are used in the same way:
...
+the command name, and the remaining part of the line is used verbatim as
+the first argument.  This is very useful for the `lua` and `rb` commands.
+As an example, the following two command lines are exactly equivalent::
+
+  :foo a b "c d" e f
+  foo "a b \"c d\" e f"
+
+There are two ways to run DFHack commands from an OS terminal.
+
+* If DF and DFHack are already running, calling ``dfhack-run my command``
```
lethosor: Should be `./dfhack-run` on *nix

docs/Core.rst
```diff
@@ -21,14 +27,51 @@ are used in the same way:
...
+  in an external terminal is equivalent to calling ``my command`` in the
+  DFHack console.  This executable is provided to support external programs,
+  and is not intended for direct use.
```
lethosor: It's not? It's definitely useful if you're running another interactive program (lua, tiletypes, etc.), if you're using PRINT_MODE:TEXT, if you've disabled the console for some reason (dfterm3?), and for `kill-lua`. In fact, it's pretty much the *only* way to run kill-lua if a lua script has entered an infinite loop, since DF typically won't respond.

docs/Core.rst
```diff
@@ -21,14 +27,51 @@ are used in the same way:
...
+
+* If DF/DFHack is started with arguments beginning with ``+``, the remaining
+  text is treated as a command in the DFHack console.  For example, use
+  `load-save` to skip the main menu with ``./dfhack +load-save region1``.
```
lethosor: Might want to verify that `Dwarf Fortress.exe +load-save region1` works on Windows and document it too.

lethosor: Also, you can specify multiple commands with `+load-save region1 +devel/print-args foo`. The "a command" part here might be a little confusing.

lethosor:
`autodump-destroy-here` is actually slightly different from `autodump destroy-here` (although I thought they were the same myself until now) - the former uses `Gui::cursor_hotkey`, so DFHack will never even attempt to call it if a cursor isn't available. With the latter bound to a hotkey, you'll get an error logged to the console if you press Ctrl-Shift-K, which is an issue if you have Ctrl-Shift-K bound to another command that works in another UI state. 

Incidentally, you do get an error if you run either version from the console. As far as I know, plugin commands can't detect whether they're invoked by a keybinding (only the DFHack core can), so merging the logic into autodump_main and making it silent when invoked as a keybinding isn't possible. Even if it were, it's easier to delegate the task of determining whether the UI state is valid to the hotkey guards (Gui::cursor_hotkey, Gui::any_item_hotkey).

Anyway, thanks for documenting the `+`-style arguments - that showed up in 5-ish changelogs around 0.43.03, so I must have assumed it was documented already, but apparently not. :)

PeridexisErrant:
OK, looks like the core docs need some extra work - I'll split that out into a new pull request.  I'll also remove the autodump changes, and squash the rest for clarity.

conf.py
```diff
@@ -17,12 +17,68 @@
 
 import fnmatch
 from io import open
-from itertools import starmap
+from functools import lru_cache
```
lethosor: This isn't implemented in Python 2

lethosor: It looks like the only two things this is used for are zero-argument functions, so it might as well be replaced with a simple memoize/cache decorator.

PeridexisErrant:
Rebased on develop to remove merge conflict in NEWS.rst, and removed use of lru_cache (again).

PeridexisErrant:
@lethosor - I've fixed the merge conflict, any other blockers?

PeridexisErrant:
@lethosor - I've fixed the merge conflict, any other blockers?

lethosor:
I've been dealing with OS upgrades and changes to Lua GUI scripts/hooks for a few days here, and I'll probably be busy in real life later this week. Sorry for the timing being somewhat inconvenient, but I'll get to this when I have time. No need to delete/recreate the comment to notify me again - I do check the list of PRs periodically, and definitely before releases.

PeridexisErrant:
Sorry - just checking that it wasn't a problem that I needed to fix.

(I've also added a commit to expand the `workNow` docs, per a request via Reddit PM)

PeridexisErrant:
Rebased to fix merge conflict in NEWS.rst

lethosor:
Apparently Python 2 doesn't like the `maxsplit` keyword argument, but this looks nice otherwise!

PR #973, opened 2016-08-03
Develop

Fukken-Saved:
I'm probably doing this wrong, but at least it only changes autohauler.


lethosor:
Squashed and merged in 93ac937

PR #975, opened 2016-08-07
Add windows container sizes to ruby codegen

Qartar:
Sizes were calculated using Visual Studio 2015 Update 3.

### DIFF
--- a/plugins/ruby/codegen.pl
+++ b/plugins/ruby/codegen.pl
@@ -716,10 +716,8 @@ sub sizeof {
         my $subtype = $field->getAttribute('ld:subtype');
 
         if ($subtype eq 'stl-vector') {
-            if ($os eq 'linux') {
+            if ($os eq 'linux' or $os eq 'windows') {
                 return ($arch == 64) ? 24 : 12;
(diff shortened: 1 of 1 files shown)


plugins/ruby/codegen.pl
```diff
@@ -716,26 +716,24 @@ sub sizeof {
         my $subtype = $field->getAttribute('ld:subtype');
 
         if ($subtype eq 'stl-vector') {
-            if ($os eq 'linux') {
+            if ($os eq 'linux' or $os eq 'windows') {
```
lethosor: Oh, did this change in VS2015?

Qartar: [Visual Studio 2012](https://msdn.microsoft.com/en-us/library/hh409293(v=vs.110).aspx) actually:

>To optimize memory resource usage, containers are now smaller. For example, in x86 release mode with default settings, **std::vector** has shrunk from 16 bytes in Visual Studio 2010 to 12 bytes in Visual Studio 2012, and **std::map** has shrunk from 16 bytes in Visual Studio 2010 to 8 bytes in Visual Studio 2012.

Issue #987, opened 2016-08-28
Included ruby libraries are all 32-bit


dseguin:
After building the develop branch on amd64 using the -DDFHACK_BUILD_ARCH=64 cmake flag, the libruby.so that ends up being installed is 32-bit. The result is working on v0.43.05 64-bit, except for ruby which can't be loaded. Symlinking libruby.so to the system .so causes ruby scripts to hang.

```
$ uname -a
Linux LAPTOP01 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2+deb8u3 (2016-07-02) x86_64 GNU/Linux
```

Checklist (-@lethosor):
- [x] Windows
- [x] OS X
- [x] Linux: #989

lethosor:
Yep. We haven't updated the ruby libraries yet. If your system ruby library is 2.x, it probably won't work. You could try ruby 1.8 or 1.9, I suppose.

dseguin:
Cool, I'll try building ruby 1.8 and see if it works. Other than that, I'm surprised at how well it works on df 64-bit already.

lethosor:
Does multicmd work? There could be a bunch of issues with ruby scripts working with structures in 64-bit DF, but multicmd shouldn't run into those (at least if you avoid running ruby scripts with it).

dseguin:
multicmd works fine.

Other ruby scripts hang when trying to interact with the cursor.
i.e. This works:
`superdwarf list` - returns an empty list as expected
But this hangs:
`superdwarf add` with cursor on unit

And something like this `show-unit-syndromes` fails with 
```E: NoMethodError: undefined method `flags1' for nil:NilClass```

lethosor:
Updated title - may as well track the status of the other two platforms here as well.

I'll work on OS X when I get a chance. @jjyg or @BenLubar, a Windows build would be great if you get a chance too.

BenLubar:
@lethosor I can't figure out how to build it for 64-bit. It looks like there's a lot of assembly code specific to various 32-bit compilers. @jjyg may have more luck.

lethosor:
It sounds like fixing this on 64-bit Windows might require updating ruby to 2.x. I'm unclear on exactly what would need to be fixed for that to work.

lethosor:
[...]

PR #989, opened 2016-08-28
Included 64-bit build of libruby1.8.7 for linux

dseguin:
Built from https://github.com/ruby/ruby on branch ruby_1_8_7

Changes:
  - Added 64-bit libruby for linux
  - Changed ruby/CMakeLists.txt accordingly
    - Default behavior is the same (unpack 32-bit lib)
    - When `DFHACK_BUILD_ARCH`=64, it unpacks the 64-bit lib


lethosor:
Thanks for building this! Sorry for not responding earlier - I was busy with OS upgrades.

I'm looking to move away from storing binaries in this repo (our current/temporary? solution is dfhack/dfhack-bin), but I will add it there as soon as I get a chance, and update ruby/CMakeLists.txt accordingly (and add you to Authors.rst if you don't mind).

dseguin:
Hmm, maybe it would be better to add ruby as a submodule? Also, I forgot to strip the debug symbols, so the lib is larger than it should be. I think `strip --strip-unneeded <lib>` will reduce it by ~2/3.

> (and add you to Authors.rst if you don't mind)

Thanks, but I haven't contributed any code ;) (But yes, you can add me)

lethosor:
Ruby takes a lot longer to build than Lua (and we don't modify it), so it's nice to have pre-built binaries. Feel free to reduce the size of this one if you like.

dseguin:
Gotcha. Whenever you have time, you can merge this or copy the lib to dfhack/dfhack-bin, whatever works.

lethosor:
(Never mind, mixed this up with #987.)

I still need to handle this - sorry for the delay. Just checking: you've tested this (e.g. `multicmd`) on 64-bit Linux 0.43.05, yes?

dseguin:
Yes. Tested just now with a fresh clone.

lethosor:
Merged (sort of). Thanks!

PR #990, opened 2016-08-29
RemoteFortressReader changes

JapaMala:
I'm not sure how to exclude the submodule from the commit.


plugins/proto/RemoteFortressReader.proto
```diff
@@ -202,6 +225,7 @@ message MapBlock
     repeated int32 tree_y = 22;
     repeated int32 tree_z = 23;
     repeated TileDigDesignation tile_dig_designation = 24;
+	repeated SpatterPile spatterPile = 25;
```
lethosor: tab

PR #992, opened 2016-09-08
Added check to rescan zlib and pthread on linux when arch changes

### LINKED ISSUE #982: zlib scan isn't re-run on Linux when build architecture changes
This will result in trying (and failing) to link a 32-bit zlib to 64-bit DFHack:
```
cmake .. -DDFHACK_BUILD_ARCH=32
cmake .. -DDFHACK_BUILD_ARCH=64
```
It might be possible to use [get_property](https://cmake.org/cmake/help/v3.0/command/get_property.html) to detect changes to DFHACK_BUILD_ARCH, set ZLIB_LIBRARY to ZLIB_LIBRARY-NOTFOUND (or "", or maybe a different variable), which should re-trigger the scan.

(This isn't an issue on Windows, because the 32-bit and 64-bit zlibs are copied into the build directory appropriately, or on OS X, where the system zlib is universal.)

Reported by mbrigdan on IRC.

dseguin:
unset ZLIB_LIBRARY and CMAKE_HAVE_PTHREAD_H to rescan those libs
when arch changes from 32 to 64 (and vis-versa). There may be a
better way to do this. Fixes #982


CMakeLists.txt
```diff
@@ -274,6 +274,16 @@ endif()
 
 #### expose depends ####
 
+# check for change in arch (32<->64)
+if(UNIX)
+    GET_PROPERTY(DFHACK_CACHED_ARCH CACHE STRING PROPERTY DFHACK_BUILD_ARCH)
+    if("${DFHACK_BUILD_ARCH}" STREQUAL "${DFHACK_CACHED_ARCH}")
```
lethosor:
```cmake
if(NOT "${DFHACK_BUILD_ARCH}" STREQUAL "${DFHACK_CACHED_ARCH}")
```
should be a bit cleaner.

lethosor:
Nice! Just wondering, have you tested building fully when switching to either architecture? (`make` with plugins disabled should be enough.)

dseguin:
I've tested switching from 32->64 and 64->32. Although the only way I can get that issue with zlib is by having both 32 and 64 bit zlib-dev install, which is not possible on Debian, but is possible on stuff like Arch. Pthread is fixed for sure.

Also, I think it would be a good idea to change `FIND_PACKAGE(Threads)` to `FIND_PACKAGE(Threads REQUIRED)` in depends/protobuf/CMakeLists.txt, since protoc won't build without linking to pthread.

lethosor:
Oh, true. I did notice that it took a while to build stuff before failing with the wrong pthread, but I didn't remember to fix it, apparently.

lethosor:
This does not work. For one thing, the get_property syntax you're using is wrong - it should be 
```cmake
GET_PROPERTY(DFHACK_CACHED_ARCH CACHE DFHACK_BUILD_ARCH PROPERTY VALUE)
```
And even with this, it seems that the cache gets updated before CMakeLists.txt is run again, so DFHACK_CACHED_ARCH is always the new one.

dseguin:
Oh, I'm really sorry about that. I'm not that familiar with cmake. But it looks like you set it up properly, thank you!

lethosor:
Heh, I fight with cmake sometimes too. It looked like something that should've worked to me, but evidently I didn't notice before I merged it.

Anyway, the issue was mainly that the pthread/zlib scans were being run every time, since DFHACK_CACHED_ARCH wasn't being set, so it did technically fix #982.

PR #1006, opened 2016-10-20
Add documentation for manipulator professions

jlogsdon:
I never did get around to adding basic docs for the Professions feature of Manipulator. So, I guess here it is?

I'm not that familiar with RST but I think I did it right.


PeridexisErrant:
Looks good to me - thanks!

PR #1011, opened 2016-10-27
Remove plugins - all broken since 0.40

PeridexisErrant:
This changes nothing at all as far as any DFHack user is concerned, as these plugins have not been possible to build since DF 34.11 - and would have to be rewritten for compatiblity with new native-DF systems.  "treefarm" is additionally replaced by "autochop".

In the unlikely event that a developer wants the source code for something... that's what version control is for!

This pull is inspired by #735.


lethosor:
I'm fine with removing treefarm since it has a successor. Not sure about the others, since they're not obsolete or impossible to fix. I'll look into them and see what the issues are.

lethosor:
I looked into misery/advtools briefly and fixed a few issues, although they still don't work yet. I believe there are just a couple advtools subcommands that are obsolete and failing to compile, so I'll see if the rest work once those subcommands are removed. As for misery, I think the only thing needing to be changed is how bad thoughts/stress are generated.

PR #1019, opened 2016-11-18
Add job remove method

CannibalVox:
Job remove eliminates a job's worker & holder references, if any, puts
the worker on cd, if appropriate, removes the job's postings, eliminates
the job from the global linked list, and then finally deletes it.  This
code was tested by incorporating it into autochop and it does make the
plugin work.  However, chop jobs don't have holder building references,
and anyway, with DF being 90% edge case by volume, this could use a heck
of a lot more testing.

I saw elsewhere code that prevented worker removal if the job was a
special job, and that made me feel funny so I made the job remove method
not work if the job is a special job.

### DIFF
--- a/library/modules/Job.cpp
+++ b/library/modules/Job.cpp
@@ -212,7 +212,7 @@ void DFHack::Job::printJobDetails(color_ostream &out, df::job *job)
     out.color(job->flags.bits.suspend ? COLOR_DARKGREY : COLOR_GREY);
     out << "Job " << job->id << ": " << ENUM_KEY_STR(job_type,job->job_type);
     if (job->flags.whole)
-           out << " (" << bitfield_to_string(job->flags) << ")";
+        out << " (" << bitfield_to_string(job->flags) << ")";
     out << endl;
     out.reset_color();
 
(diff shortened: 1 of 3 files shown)


CannibalVox:
Apologies if I've missed something in the contributor document!  Please let me know if there's anything I can change here.

library/modules/Job.cpp
```diff
@@ -304,6 +304,68 @@ void DFHack::Job::setJobCooldown(df::building *workshop, df::unit *worker, int c
...
+            vector_erase_at(holder->jobs, jobIndex);
+    }
+
+    if (worker) {
+        if (worker->job.current_job == job)
+            worker->job.current_job = NULL;
+    }
+
+    //Wipe all refs out
+    while (job->general_refs.size() > 0) {
```
warmist: theoretically this needs checking. Each reference means that there is an object that points to this job at the other end. To be very safe- check if gen-refs are worker and holder and remove only them. If there are more general refs then we don't know how to remove them.

library/modules/Job.cpp
```diff
@@ -304,6 +304,68 @@ void DFHack::Job::setJobCooldown(df::building *workshop, df::unit *worker, int c
...
+            prev->next = next;
+
+        if (next)
+            next->prev = prev;
+
+        delete job->list_link;
+    }
+
+    delete job;
+}
```
warmist: This is also missing the part which walks over all items and unmarks them as "in_job". This would make them unusable in other jobs.

lethosor:
Yeah, `WRAPM(Job,removeJob)` around `WRAPM(Job,removePostings)` in library/LuaApi.cpp should do it.

library/modules/Job.cpp
CannibalVox: Even if we don't remove the refs, we're deleting the job so if there's a reverse-ref it'll crash regardless, I think.

CannibalVox:
Okay, lua bindings added & item disconnections fixed.  Should I stop deleting all general refs from the job?  I feel like if there's a return ref we're screwed either way so we might as well delete the refs.

warmist:
I think it's okay as is. Maybe a note would be nice. Also we could benefit from "universal general ref destroyer" which would know how to unlink ANY general_ref. That way a lot of minor thing would be easier (e.g. items in containers and so on).

CannibalVox:
Was just taking another look at this- job.job_items seems to be a vector of pointers, which means the contents of that vector need to be deleted & removed to avoid a memory leak, right?  Or are other objects using those items?

lethosor:
Those appear to be specific to the job structure (i.e. descriptions of items, not actual items). I want to say that they're not used anywhere else, but I'm not entirely sure.

warmist:
Yes they are only used by the job. Correct order of operations: for each job_item unmark linked item as in_job, delete the job_item, clear the vector.

CannibalVox:
I think you're thinking of the items vector, job_items don't seem to have any references to actual items

CannibalVox:
Is there anything you're waiting on me to do before this can be merged?

library/modules/Job.cpp
lethosor: Is this covered by https://github.com/DFHack/dfhack/pull/1019/files#diff-a460fb2c626d385f023be6225025afcdR358?

library/modules/Job.cpp
lethosor: Should it just refuse to delete the job if there are refs it can't handle?

lethosor:
I left a couple comments (apologies if they were answered already and I forgot)

library/modules/Job.cpp
CannibalVox: Yeah- I mark the item as not in job & remove the return reference, delete all item refs & clear the vector.

CannibalVox:
Not sure what's going on with the build- it looks like the linter is beefing?  Do I have bad formatting?

warmist:
./library/include/modules/Job.h: Contains trailing whitespace: lines 71, 75, 80

CannibalVox:
Whoops, thanks for that!  I think this is all set.. now...

lethosor:
The only other thing I noticed (and mentioned on IRC) was that the other two added functions could take `job` first for consistency. I can handle that along with a couple tests, though.

lethosor:
Got a crash with `dfhack.job.removeJob(job)` on the units screen with a "construct mechanisms" job selected.

warmist:
@lethosor imho we do not add checks for dangling pointers. This will quickly go insane. If you want to delete pointer from a screen that holds pointers - you are responsible for fixing it up. This is a tool and it should not be misused.
Maybe add another function in gui module? But i think this is already too much discussion for a quite simple feature.

CannibalVox:
I think the hardest factor in removing jobs without handling the GUI ourselves would be something like getplants which would need to remove jobs from the CLI - based on what you happen to be doing in game it could crash/not crash?  That seems bad.  I would like to iterate on this to at least handle those two viewscreens, but I agree that it's beyond the scope of this PR.  I'd like to do getplants after autochop's improvements are checked in, so perhaps I can investigate this then.

lethosor:
Yeah, I mostly agree with @warmist here. I just brought it up because it was the simplest way I thought of to access a job, and it triggered a crash.

Regarding getplants, I would say it should only be messing with jobs when the game is unpaused. It can't *already* be calling removeJob(), so hopefully implementing it so that it avoids messing with jobs when the game is paused isn't too hard. (I *think* workflow does this, but I could be wrong.)

CannibalVox:
Do you need me to reverse those params?  I'm more than happy to do it myself.

lethosor:
I meant to, sorry. I think I forgot to do that after I tested it in one of my forts.

If you want to change it, feel free. Otherwise, I'll be busy for a couple days, but I'll change it if necessary and merge it once I get back.

PR #1020, opened 2016-11-26
make labormanager know building instruments is furniture hauling

nocico:
I did test that furniture hauling is responsible for building instruments


ab9rf:
I had no idea that instruments could be built as constructions. Looks good to me.

PR #1022, opened 2016-11-26
Use attributes in calculating assignment weight

ab9rf:
Improves labormanager by using dwarven attributes (e.g. strength) in deciding who to assign jobs.


plugins/labormanager.cpp
```diff
@@ -2156,10 +2300,16 @@ class AutoLaborManager {
             {
                 skill_level = Units::getEffectiveSkill(d->dwarf, skill);
                 xp = Units::getExperience(d->dwarf, skill, false);
+
+                for (int pa = 0; pa < 6; pa++)
```
lethosor: It looks like the whitespace here and on line 2307 is the issue. I'm not sure why the Travis script is reporting 2305/2308.

lethosor: Here's the output on my end:
```
Contains trailing whitespace: lines 1829, 2304, 2307, 2582-2583, 2710-2711, 2763-2766, 2896
```
It doesn't look like you touched 1829 here, though, and it's fine on dfhack/dfhack:develop. It could be due to rebasing, although I'm not really sure.

PR #1025, opened 2016-11-29
Labormanager trap components and constructions with multiple materials

nocico:
Trap components were left to mechanic labor, which stalled my only carpenter's workshop when trying to craft a wooden corkscrew. Screw pumps would not get built unless the pipe section's material's labor matched the block's, and other multi-material constructions had similar problems. This fixes these issues.

### DIFF
--- a/plugins/labormanager.cpp
+++ b/plugins/labormanager.cpp
@@ -673,8 +673,15 @@ static df::building* get_building_from_job(df::job* j)
     return 0;
 }
 
-static df::unit_labor construction_build_labor (df::item* i)
+static df::unit_labor construction_build_labor (df::building_actual* b)
 {
+    if (b->getType() == df::building_type::RoadPaved)
+        return df::unit_labor::BUILD_ROAD;
+    auto a = (df::building_actual *) b;
+    // For screw pumps contained_items[0] = pipe, 1 corkscrew, 2 block
+    // For wells 0 mechanism, 1 rope, 2 bucket, 3 block
+    // Trade depots and bridges use the last one too
+    df::item* i = a->contained_items.back()->item;
     MaterialInfo matinfo;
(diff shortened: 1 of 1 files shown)


ab9rf:
I knew there was a stall somewhere in forge labors, but I couldn't remember what it was. Thanks for finding this, and the one with pumps too.

plugins/labormanager.cpp
```diff
@@ -673,8 +673,15 @@ static df::building* get_building_from_job(df::job* j)
     return 0;
 }
 
-static df::unit_labor construction_build_labor (df::item* i)
+static df::unit_labor construction_build_labor (df::building_actual* b)
 {
+    if (b->getType() == df::building_type::RoadPaved)
+        return df::unit_labor::BUILD_ROAD;
+    auto a = (df::building_actual *) b;
```
lethosor: Isn't `b` already a `df::building_actual*`?

ab9rf: This patch changed it to a df::item. It'll actually be a building_actual at that point, but C++ is too stupid to know this.

plugins/labormanager.cpp
```diff
@@ -673,8 +673,15 @@ static df::building* get_building_from_job(df::job* j)
...
+    // For screw pumps contained_items[0] = pipe, 1 corkscrew, 2 block
+    // For wells 0 mechanism, 1 rope, 2 bucket, 3 block
+    // Trade depots and bridges use the last one too
+    df::item* i = a->contained_items.back()->item;
```
ab9rf: A constructed building will always have at least one item in it. I doubt that's anything that will change any time soon.

lethosor: Yeah, not sure what I was thinking there. It's been a while since I've looked at buildings.

Issue #1018, opened 2016-11-15
getplants duplicates cutting/gathering orders if run multiple times


PatrikLundell:
The plugin designates the specified plants for gathering every time the command is executed. However, the designations are converted into jobs and it appears the designations are then reset, so when the command is executed again a while later it will designate all the plants for gathering again, resulting in new jobs targeting the same tiles.
To avoid this, getplants would have to iterate over the df.global.world.job_list to find duplicates and refrain from designating those tiles again.

I found this when using a lua script based on getplants' logic to spare one of each kind of booze supporting fruit tree on my embark, running the script whenever I thought a bit too many new trees had popped up (and with a wood cutter doing other things mostly), and then saw my wood cutter cut a tree, walk away, return with a tree cutting order to the empty tile, walk away...
I've verified that "getplants CASHEW" repeated a few times with DF running in between resulted in multiple tree cutting orders for the same set of 3 tiles (the number of cashews I happened to have at the time). While I only tested with trees, I assume the logic is the same with shrubs.

ab9rf:
This is closely related to the problem in #1014 and has to do with how Toady changed the way map designations are handled, somewhere in 0.40.x. Prior to 0.40.x, map designations persisted and map-related jobs (digging, tree cutting, etc) were spawned when a dwarf with the relevant labor enabled went looking for a job. Sometime after 0.34.11, Toady changed this: on unpause, map designations that become set while the game is paused are immediately cleared and converted into job postings. Setting new ones after that will be converted again into more job postings. The in-game map renderer checks the job posting queue when it determines if a map square should be shown as "designated".

The jobs are not posted to df.global.world.job_list initially; they're posted instead to df.global.world.job_postings. They get moved to the job list when a dwarf is actually assigned to the posting.

Getplants (and also autochop and likely many other plugins) are still relying on the map designations; anything that does so will probably behave unreliably. Anything that wants to _query_ map designations probably needs now to also scan both the job posting list and the job list for pending jobs. If Toady has a lookaside accelerator for this, I'm not aware of it.

PatrikLundell:
[...]

Issue #1014, opened 2016-10-31
getplants marks the wrong tile for trees sometimes


ab9rf:
getplants, when asked to mark trees for chopping, will incorrectly mark the center tile of a 3x3 or larger tree. As this tile is unreachable, the tree will never be chopped down.

lethosor:
Do you have a save handy with trees this large?

ab9rf:
Not at the moment, but I'll get one to you soon; they're easy to get (any embark with highwood will do).

ab9rf:
http://dffd.bay12games.com/file.php?id=12537

Issuing a "getplants highwood" will mark the highwood that's about 20 blocks from the embark wagon for cutting, but it will mark the center square, which is not reachable and thus will never lead to anyone cutting it down.

quietust:
Getplants was written back for version 0.31, and it was apparently never updated for multi-tile trees - it'll need to be adjusted to read the tree data and [de]select all trunk tiles on the "base" Z-level

PatrikLundell:
I'm not sure I agree with quietust (or misunderstand the comment). I've had the issue myself, and the problem is that the tile marked is the one the tree is located on, i.e. the central tile, but not the other ones. Additional logic is needed to determine that the trunk is 3*3 and select one of the surrounding tiles instead, preferably one that's reachable (i.e. not one on the outer side of a cliff, or one by a cliff edge. A corner should work most times).

lethosor:
Yeah, so the issue is that getplants hasn't been updated to handle multi-tile trees, like Quietust said.

lethosor:
This should be fixed in the next release. Autochop should also use the right tile (thanks for the save there, @ab9rf!).

Issue #531, opened 2015-02-05
`getplants -c [ID]` does not clear all corresponding designations (anymore)


bseiller:
After designating some shrubs (MUSHROOM_CUP_DIMPLE) with `getplants -s [ID]` and harvesting the reachables i tried to undesignate the rest (in the caverns which) `getplants -c [ID]`
They stayed designated for gathering, while the output said `Updated 153 plant designations`.
Undesignating some by hand and reissuing `getplants -s [ID]` flagged all of them again, using `getplants -c [ID]` after that properly cleared those that I changed manually, but again left those only manipulated by the initial getplants call unchanged - the output still claimed `Updated 153 plant designations`.
On request I can provide the save.

warmist:
Gathering designations are no longer in designations flags. It's now a 4000 flag in designation priority block event.

lethosor:
This should be fixed for the next release.

PR #1030, opened 2016-12-06
Fix autochop

CannibalVox:
Add job-removal functionality to autochop, this should make it actually
function.

### DIFF
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -251,21 +251,21 @@ static int do_chop_designation(bool chop, bool count_only)
         }
     }
 
-	if (!chop) {
-		//count and/or deadify all chop jobs
-		df::job_list_link *link = &world->job_list;
-		while (link) {
-			auto next = link->next;
+    if (!chop) {
+        //count and/or deadify all chop jobs
+        df::job_list_link *link = &world->job_list;
+        while (link) {
+            auto next = link->next;
 
-			if (link->item != NULL && link->item->job_type == df::job_type::FellTree && (count_only || watchedBurrows.isValidPos(link->item->pos))) {
-				++count;
-				if (!count_only)
-					DFHack::Job::removeJob(link->item);
-			}
+            if (link->item != NULL && link->item->job_type == df::job_type::FellTree && (count_only || watchedBurrows.isValidPos(link->item->pos))) {
+                ++count;
+                if (!count_only)
+                    DFHack::Job::removeJob(link->item);
+            }
 
-			link = next;
-		}
-	}
+            link = next;
+        }
+    }
 
     return count;
 }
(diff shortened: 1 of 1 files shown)


lethosor:
What's this supposed to fix? Job removal?

I noticed an issue before your changes where woodcutters would stay around trees they just cut down (even ones with single-tile trunks) with the "fell tree" job for a few seconds before moving on to the next one, and designations wouldn't flash after the first tree they cut down. That still seems to be an issue. Also, a new issue that I only saw after these changes is a bunch of "Miner cancels dig: inappropriate dig square" announcements. They're to the southwest corner of the most recently felled trees (so far), so it may be the corner of a 16x16 block.

ab9rf:
The "hanging around" problem is likely related to #1018 and #1014. What likely happens is that the tile is designated for chopping by the plugin by setting its designation bits. On unpause, DF clears the designation and queues a chop tree job posting. The next time the treechopping plugin runs, it sees the tree as not designated for chopping and designates it again. DF again queues a chop tree job posting. When the tree is finally cut down, DF removes the first job posting. The treechopping dwarf then looks for nearby treecutting jobs, finds one on the same square, and tries to start it, but gets canceled because there's no longer a tree there. This results in the "inappropriate dig square" announcement.

Thus, plugins that will set map designations need to check the job posting queue for jobs for that map square to avoid spamming the job queue. An option would be for the job to just post the job queue entry themselves, but I'm not sure we know how to do that yet.

lethosor:
I'm pretty sure it was a different dwarf cancelling the mining job (not the woodcutter). I could be wrong about that, though.

Checking the postings list is probably a good way to go, unless we can create jobs ourselves (in which case inserting a posting is fairly easy).

CannibalVox:
Yeah, this was primarily to deal with job removal.  Atm, if you set it into autochop mode, dwarfs will currently uncontrollably chop every tree on the map because there's no real way to kill the jobs once you have enough logs.  This makes "undesignate" and the autochop "i have enough logs" stuff properly remove existing jobs.  I can take care of those other things too though.

lethosor:
Okay, makes sense. I just did a brief test and didn't look at the cancellation features, apparently.

CannibalVox:
Is "inappropriate dig square" actually a new bug I introduced?  Not sure how I did that.

CannibalVox:
Oh also kind of a side effect of the job removal fix, but we also correctly count designated trees, even once the designations are converted to jobs.  Dupe jobs shouldn't be the norm in ordinary play unless a player manually redesignates everything (which is certainly a thing that can happen).

lethosor:
Does the "inappropriate dig square" issue occur on your end? I thought it could've been related to #1047, since it showed up around that time, but the issue is still occurring for me in alpha4 (64-bit OS X).

I'm also seeing my woodcutter remain next to a recently-felled tree with the "fell tree" job, or even return multiple times to the most recently-felled one, apparently trying to cut it down again. The lack of the usual flashing designation is also still an issue.

lethosor:
@CannibalVox have you seen the "inappropriate dig square" issue on your end?

lethosor:
Any luck with the announcement spam? I'd like to get this merged into r1, if possible.

lethosor:
I've figured out how to get rid of the duplicated jobs (which was causing the issue with woodcutters trying to chop down already-felled trees, also mentioned in #1076). I'll look into this fix too, although I may have to manually apply it.

lethosor:
I made a module based partly on this in 1e469453fcf947744c35c1280890482354cb39dc, hopefully to use for getplants too (#531, #1018, and similar issues). Thanks for the help!

Issue #1009, opened 2016-10-24
Linking against system tinyxml seems to cause immediate segfault on 0.43.05-alpha1


TC01:
I went to upgrade my packages to the 0.43.05 alpha and found that, while things compiled, attempting to launch dfhack immediately segfaulted:

```
$ dfhack
Sound devices available:
OpenAL Soft
Picking OpenAL Soft. If your desired device was missing, make sure you have the appropriate 32-bit libraries installed. If you wanted a different device, configure ~/.openalrc appropriately.
Perfect OpenAL context attributes GET
Loading bindings from data/init/interface.txt
New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

Resetting textures
/home/bjr/.dwarffortress/dfhack: line 66: 13880 Segmentation fault      (core dumped) setarch i386 -R env LD_PRELOAD="$PRELOAD_LIB" ./libs/Dwarf_Fortress "$@"
```

After some trial and error I eventually traced the problem to me setting -DEXTERNAL_TINYXML in the build process. Disabling that flag and using the included tinyxml sources seemed to have solved the issue.

This is obviously a low-priority issue but I figured I should file a bug. Poking briefly around the source tree it's not immediately clear to me what's changed to cause this. What version of tinyxml does dfhack bundle?

lethosor:
```cpp
const int TIXML_MAJOR_VERSION = 2;
const int TIXML_MINOR_VERSION = 5;
const int TIXML_PATCH_VERSION = 3;
```
This appears to be https://sourceforge.net/projects/tinyxml/, which is the original TinyXML (I think), *not* TinyXML2.

I don't imagine we depend heavily on a specific version of TinyXML, so upgrading shouldn't be too bad if that would help.

lethosor:
I wouldn't expect this to be the case, but is it possible that DFHack is still using its own TinyXML headers? I don't know why that wouldn't show up before 0.43.05, though, unless your system TinyXML isn't the same architecture as DFHack and CMake is switching things up on you.

TC01:
Hm, I suppose it's possible that the wrong headers are being picked up, but yeah, I'm not sure why that wouldn't have been an issue before this. I did double-check to make sure the 64-bit tinyxml is being detected properly by cmake: ```TinyXML_LIBRARY:FILEPATH=/usr/lib64/libtinyxml.so``` is definitely getting set.

[...]

PR #1033, opened 2016-12-07
Update bundled tinyxml version from 2.5.3 to 2.6.2

TC01:
Upstream has moved onto tinyxml2, so this is likely to be the last release of the original tinyxml (2.6.2 came out in 2011). dfhack should therefore probably ship 2.6.2, since it's unlikely there will ever be future fixes to this code. The upstream changelog (taken from "changes.txt" in upstream tarball) contains many bugfixes and is included in the commit message below.

This PR was originally motivated by #1009. I poked around the git history and it seems that, at one point, tinyxml was updated to [2.6.1, but then reverted soon after](https://github.com/DFHack/dfhack/commits/v0.31.25/library/depends/tinyxml) way back in 2011. The comment doesn't explain why, though, and I tested that dfhack runs successfully with the newer tinyxml substituted instead. (I generated a world and started a fortress and everything *seemed* to be working just fine).

2.5.4
- A TiXMLDocument can't be a sub-node. Block this from happening in the 'replace'. Thanks Noam.
- [ 1714831 ] TiXmlBase::location is not copied by copy-ctors, fix reported and suggested by Nicola Civran.
- Fixed possible memory overrun in the comment reading code - thanks gcarlton77

2.5.5
- Alex van der Wal spotted incorrect types (lf) being used in print and scan. robertnestor pointed out some problems with the simple solution. Types updated.
- Johannes Hillert pointed out some bug typos.
- Christian Mueller identified inconsistent error handling with Attributes.
- olivier barthelemy also reported a problem with double truncation, also related to the %lf issue.
- zaelsius came up with a great (and simple) suggestion to fix QueryValueAttribute truncating strings.
- added some null pointer checks suggested by hansenk
- Sami Väisänen found a (rare) buffer overrun that could occur in parsing.
- vi tri filed a bug that led to a refactoring of the attribute setting mess (as well as adding a missing SetDoubleAttribute() )
- removed TIXML_ERROR_OUT_OF_MEMORY. TinyXML does not systematically address OOO, and the notion it does is misleading.
- vanneto, keithmarshall, others all reported the warning from IsWhiteSpace() usage. Cleaned this up - many thanks to everyone who reported this one.
- tibur found a bug in end tag parsing

2.6.2
- Switched over to VC 2010
- Fixed up all the build issues arising from that. (Lots of latent build problems.)
- Removed the old, now unmaintained and likely not working, build files.
- Fixed some static analysis issues reported by orbitcowboy from cppcheck.
- Bayard 95 sent in analysis from a different analyzer - fixes applied from that as well.
- Tim Kosse sent a patch fixing an infinite loop.
- Ma Anguo identified a doc issue.
- Eddie Cohen identified a missing qualifier resulting in a compilation error on some systems.
- Fixed a line ending bug. (What year is this? Can we all agree on a format for text files? Please? ...oh well.)

### DIFF
--- a/depends/tinyxml/tinystr.cpp
+++ b/depends/tinyxml/tinystr.cpp
@@ -1,6 +1,5 @@
 /*
 www.sourceforge.net/projects/tinyxml
-Original file by Yves Berquin.
 
 This software is provided 'as-is', without any express or implied
 warranty. In no event will the authors be held liable for any
@@ -21,10 +20,6 @@ must not be misrepresented as being the original software.
 3. This notice may not be removed or altered from any source
 distribution.
 */
-
-/*
- * THIS FILE WAS ALTERED BY Tyge Løvset, 7. April 2005.
- */
 
 
 #ifndef TIXML_USE_STL
--- a/depends/tinyxml/tinystr.h
+++ b/depends/tinyxml/tinystr.h
@@ -1,6 +1,5 @@
 /*
 www.sourceforge.net/projects/tinyxml
-Original file by Yves Berquin.
 
 This software is provided 'as-is', without any express or implied
 warranty. In no event will the authors be held liable for any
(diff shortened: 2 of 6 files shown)


lethosor:
Yeah, this is mainly used for loading symbols.xml, and that seems to be working fine still.

PR #1043, opened 2016-12-15
labormanager: Add support for shell crafts.

ab9rf:
Add support for shell crafts to job inference engine.


lethosor:
I cherry-picked c74e872 in c99780aafd (to avoid including 5c1d333 again, which somehow wound up on your branch). Sadly, that means Github doesn't think this was merged, but your commit is merged into develop now.

ab9rf:
I don't know how that ended up on that branch either, but I'm fine with the cherry pick.

PR #1046, opened 2016-12-16
stockflow: Fix "integer expected" problem in stockflow

### LINKED ISSUE #1044: Error from stockflow
After saving and reloading a fort in which stockflow had been in use, got the following error some time after unpausing:

```
C:\Users\Kelly\df_43_05_win\hack\lua\plugins\stockflow.lua:1097: Cannot write field manager_order.amount_left: integer expected.
stack traceback:
        [C]: in metamethod '__newindex'
        C:\Users\Kelly\df_43_05_win\hack\lua\plugins\stockflow.lua:1097: in function 'plugins.stockflow.create_orders'
        C:\Users\Kelly\df_43_05_win\hack\lua\plugins\stockflow.lua:91: in function 'plugins.stockflow.finish_bookkeeping'
```
Doesn't crash, but the expected jobs are not added. The message recurs each time the bookkeeper updates stock records.

Stockflow was working prior to the save/reload.

DF 0.43.05, Win64, dfhack 0.43.05-alpha3. Save available upon request.

ab9rf:
This appears to resolve #1044.

### DIFF
--- a/plugins/lua/stockflow.lua
+++ b/plugins/lua/stockflow.lua
@@ -1136,8 +1136,7 @@ function check_stockpiles(verbose)
             local reaction = spec.entry.ints[entry_ints.order_number]
             local filled, empty = check_pile(spec.stockpile, verbose)
             local amount = trigger.filled and filled or empty
-            amount = (amount - (amount % trigger.divisor)) / trigger.divisor
-            amount = math.floor(amount)
+            amount = (amount - (amount % trigger.divisor)) // trigger.divisor
             result[reaction] = (result[reaction] or 0) + amount
         end
     end
--- a/plugins/lua/stockflow.lua
+++ b/plugins/lua/stockflow.lua
@@ -1137,6 +1137,7 @@ function check_stockpiles(verbose)
             local filled, empty = check_pile(spec.stockpile, verbose)
             local amount = trigger.filled and filled or empty
             amount = (amount - (amount % trigger.divisor)) / trigger.divisor
+            amount = math.floor(amount)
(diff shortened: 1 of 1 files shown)


dscorbett:
Instead of using `math.floor`, use `//` (floor division).

lethosor:
It's worth noting that using the // operator would break the lua syntax checking script, since it uses Lua 5.2 (5.3 isn't available on Travis). That's not a good reason not to use that operator, though. I'm looking into solutions on the Travis side - in the worst case, Lua doesn't take that long to build, and the binaries could probably be cached to avoid building it every time.

dscorbett:
Even this new version is unnecessarily complex: `(x - (x % y)) // y` can be simplified to `x // y` for any finite numbers `x` and `y`. You only need one line to calculate `amount`; see the code in my previous comment.

lethosor:
Sorry, not sure how I missed that. I'll change it, and probably use math.floor for now until I can get Lua 5.3 working.

dscorbett:
My recommendation of `//` was too hasty. `//` produces the right answer, but does not convert its result to an integer if either input is a float. I didn’t check whether it is safe to use this operator in this script.

lethosor:
Sure enough. I'll leave it as-is, then.

ab9rf:
They should be integers, but "should" is here a somewhat fuzzy concept given Lua's odd definition of "integer".

lethosor:
Lua 5.3 has more of a distinction (e.g. `3.0` and `3` will print differently but compare as equal, while they would both print as "3" in Lua 5.2).

PR #1063, opened 2017-01-20
Add DOWNLOAD_RUBY option to Ruby plugin cmake, defaulting to ON

TC01:
If DOWNLOAD_RUBY is set to OFF, the build will not attempt to download a prebuilt libruby.so and not fire a warning that one is not present. This may be desired if one wants to link dfhack against a newer system
libruby.so, now that dfhack supports ruby 2.x and also supports linking against system libruby (at least, on Linux), as per this recent [commit](https://github.com/DFHack/dfhack/commit/df9b5bca7311489748ab100dfc51080841c3f196).

This also allows the dfhack build to proceed without network access (once all submodules have been fetched), which at least some Linux distributions, like Fedora, require.

By default DOWNLOAD_RUBY is set to ON, so this should not change the behavior of default builds.

I tested that when DOWNLOAD_RUBY is set to OFF, dfhack still launches and that the Ruby plugin still functions, e.g. by running ```rb_eval puts('Hello world.')``` in the dfhack console.

### DIFF
--- a/plugins/ruby/CMakeLists.txt
+++ b/plugins/ruby/CMakeLists.txt
@@ -1,52 +1,59 @@
-IF (APPLE)
-    SET(RUBYLIB ${CMAKE_CURRENT_SOURCE_DIR}/osx${DFHACK_BUILD_ARCH}/libruby.dylib)
-    SET(RUBYLIB_INSTALL_NAME "libruby.dylib")
-    IF(${DFHACK_BUILD_ARCH} STREQUAL 64)
-        # MESSAGE("No ruby lib for 64-bit OS X yet")
+# Allow build system to turn off downloading of libruby.so.
+OPTION(DOWNLOAD_RUBY "Download prebuilt libruby.so for ruby plugin." ON)
+
+IF (DOWNLOAD_RUBY)
+
+    IF (APPLE)
+        SET(RUBYLIB ${CMAKE_CURRENT_SOURCE_DIR}/osx${DFHACK_BUILD_ARCH}/libruby.dylib)
+        SET(RUBYLIB_INSTALL_NAME "libruby.dylib")
+        IF(${DFHACK_BUILD_ARCH} STREQUAL 64)
+            # MESSAGE("No ruby lib for 64-bit OS X yet")
+        ELSE()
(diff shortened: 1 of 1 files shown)


lethosor:
Have you checked that a system libruby is located successfully on Linux?

plugins/ruby/CMakeLists.txt
```diff
@@ -1,3 +1,8 @@
+# Allow build system to turn off downloading of libruby.so.
+OPTION(DOWNLOAD_RUBY "Download prebuilt libruby.so for ruby plugin." ON)
+
+IF (DOWNLOAD_RUBY)
+
```
lethosor: Might want to indent this part

TC01:
@lethosor: yes, I have, sorry for not making that explicitly clear. The test I mentioned was ran with no libruby.so present anywhere in my DF/DH installation.

Note though that in order to link against the non-versioned libruby.so on at least Fedora, the ```ruby-devel``` package is required (which provides ```/usr/lib64/libruby.so``` as a symlink to the versioned .so file). This is a little ugly, but it prevents dfhack from needing to be rebuilt whenever libruby is updated.

plugins/ruby/CMakeLists.txt
TC01: Done; I re-pushed the commit with everything between the initial IF(DOWNLOAD_RUBY) and ENDIF() indented one level.

lethosor:
What's the name of the Ruby library on Linux? If it doesn't change too often, I wouldn't mind checking for specific versions.

TC01:
Well, on Fedora 25 , it's ```libruby.so.2.3.0```as Fedora currently ships Ruby 2.3.x. I'd assume that the so version would normally only gets bumped every time there's a major Ruby release. I'm not that familiar with the Ruby release cycle, but I know 2.4 was released recently and [the roadmap](https://bugs.ruby-lang.org/projects/ruby-trunk/roadmap) projects 2.5 in about a year's time, so I'd assume they currently release approximately annually?

Unfortunately, I just checked Debian, which seems to package it as ```libruby-2.1.so.2.1.0``` and doesn't even *have* a libruby.so symlink (they have a libruby-2.1.so symlink instead). So it's not like this is done consistently across distributions. :(

PR #1064, opened 2017-01-26
Creature renamer

JapaMala:
Renames generated creatures to have tokens that are useful to graphics packs, allowing giant snake forgotten beasts to have giant snake tiles.


lethosor:
I'd like to look into this more before enabling it by default. I don't think all users would need it anyway, so the risk of breaking saves isn't worth it. From what I remember about creature raws (which isn't much, admittedly), changing IDs can prevent saves from loading entirely. Is that different for generated creatures?

Also, this needs to be documented, ideally with an extra note if it's enabled by default.

plugins/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ if (BUILD_SUPPORTED)
     DFHACK_PLUGIN(follow follow.cpp)
     DFHACK_PLUGIN(forceequip forceequip.cpp)
     DFHACK_PLUGIN(fortplan fortplan.cpp LINK_LIBRARIES buildingplan-lib)
+	DFHACK_PLUGIN(generated-creature-renamer generated-creature-renamer.cpp)
```
lethosor: Inconsistent indentation

plugins/generated-creature-renamer.cpp
```diff
@@ -0,0 +1,159 @@
...
+DFHACK_PLUGIN_IS_ENABLED(is_enabled);
+
+DFhackCExport command_result plugin_enable(color_ostream &out, bool enable)
+{
+    is_enabled = enable;
+    return CR_OK;
+}
+
+#define NUM_DESC 227
+std::string descriptors[NUM_DESC] = { " blob", " quadruped", " humanoid", " silverfish", " mayfly", " dragonfly",
```
lethosor: I assume the order here is important - is there anywhere else (e.g. in df-structures) where these are defined? It might be better to use enum attributes, if possible.

Also, are there any here that don't start with a space? There aren't any that stand out to me, although I could have missed some.

JapaMala: The order isn't particularly important. It's just a list of possible names. They all have a space to make sure it only matches whole words. If there's a better way of doing that, I'm open to suggestions.

plugins/CMakeLists.txt
JapaMala: I'll change it.

plugins/generated-creature-renamer.cpp
lethosor: Oh, this is searching the description. Makes sense, although you could probably prepend the space later (instead of relying on it to be present in every entry here), e.g.
```c++
size_t pos = creatureRaw->caste[0]->description.find(" " + descriptors[j]);
```

I would probably also use an STL container instead of keeping track of the size manually, but it's not a big deal.

JapaMala: Okay, I've made it into a STL::Vector, removed the spaces, and added basic version checking in case more creatures are added later.

PR #1074, opened 2017-03-27
README.md: Fix broken markdown links and header.

pjf:
Github's markdown parser does not permit newlines between link titles
and link text. This change removes the newlines, resulting in a less
pleasing source, but one which renders when viewed on github.

Github also requires a space between the '#' and text of headings, so
we fix that, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,10 @@
-#DFHack Readme
+# DFHack Readme
 
-[![Build Status](https://travis-ci.org/DFHack/dfhack.svg?branch=develop)]
-(https://travis-ci.org/DFHack/dfhack)
-[![Documentation Status](https://readthedocs.org/projects/dfhack/badge)]
-(https://dfhack.readthedocs.org)
-[![License](https://img.shields.io/badge/license-ZLib-blue.svg)]
-(https://en.wikipedia.org/wiki/Zlib_License)
-[![Github Issues](http://githubbadges.herokuapp.com/DFHack/dfhack/issues)]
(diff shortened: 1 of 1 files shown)


lethosor:
Interesting. This must have been a fairly recent change, because I remember at least the images rendering correctly before. I'll get this merged into develop, and hopefully master too.

warmist:
@lethosor yes it is probably quite new. Recently Github released their markdown specs see: [This blog post](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown)

pjf:
Oops! I didn't realise there was a develop branch. I've rebased and changed merge targets appropriately. Hopefully this will also make travis happy. :)

lethosor:
Not sure why that failed again, but it looks good to me.

pjf:
> Not sure why that failed again, but it looks good to me.

My guess is because I rebased first, and changed merge targets second. Travis runs in response to the re-based commit (which fails, as it's not yet against `develop`), but *doesn't* run when the merge target changes as there's no new commit to trigger it.

If that's the case, then hitting "restart build" on the travis job should have it pass. :)

pjf:
And travis is happy now that I got it to restart. :)

(Ignore the "unverified", it's me discovering how github handles signed commits.)

lethosor:
I didn't even know you could change the base/target of an existing PR. Did you just do that by rebasing your local branch?

No worries about the signature - not many of our commits are signed anyway.

PR #1079, opened 2017-04-24
fix of relations errors.

Amostubal:
There is a lot more errors in this file, mostly called unk variables... I bet they have names, I've tried to figure out what they were renamed too, but I'm not fully understanding the xml.


lethosor:
This repo is different - we make all of the changes on the `develop` branch, and merge it into `master` for releases. I managed to change the base branch of this PR (above), but that completely confused Travis, so I cancelled the build. It looks good to merge, though. Thanks for the PR!

PR #1098, opened 2017-05-27
Add orders plugin for importing and exporting manager orders.

BenLubar:
```
orders - Manipulate manager orders
  orders export [name]
    Exports the current list of manager orders to a file named dfhack-config/orders/[name].json.
  orders import [name]
    Imports manager orders from a file named dfhack-config/orders/[name].json.
  orders clear
    Deletes all manager orders in the current embark.
```


plugins/orders.cpp
```diff
@@ -0,0 +1,788 @@
...
+    Json::Value orders(Json::arrayValue);
+
+    {
+        CoreSuspender suspend;
+
+        for (auto it = world->manager_orders.begin(); it != world->manager_orders.end(); it++)
+        {
+            Json::Value order(Json::objectValue);
+
+            order["id"] = (*it)->id;
```
lethosor: Do you need to save the IDs? This might cause issues when importing orders into different worlds.

BenLubar: The IDs are used for the order conditions so orders can reference other orders.

plugins/orders.cpp
```diff
@@ -0,0 +1,811 @@
...
+    CoreSuspender suspend;
+
+    std::map<int32_t, int32_t> id_mapping;
+    for (auto it : orders)
+    {
+        id_mapping[it["id"].asInt()] = world->manager_order_next_id;
+        world->manager_order_next_id++;
+    }
+
+    for (auto & it : orders)
```
BenLubar: The problem is that "order" and "condition" were already used inside the loops, so I didn't want to start using synonyms or other confusing stuff.

PR #1114, opened 2017-06-19
Enable asserts in RelWithDebInfo builds

BenLubar:
Removes the `NDEBUG` define from RelWithDebInfo builds.


CMakeLists.txt
```diff
@@ -185,7 +185,7 @@ IF(UNIX)
   # ensure compatibility with older CPUs
   # enable C++11 features
   add_definitions(-DLINUX_BUILD)
-  SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -g -Wall -Wno-unused-variable")
+  SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -g -Wall -Wno-unused-variable -DDFHACK_ASSERT")
```
lethosor: It looks like this means that the behavior with Debug builds would be the same as with Release builds.

library/include/assert.h
```diff
@@ -0,0 +1,34 @@
...
+3. This notice may not be removed or altered from any source
+distribution.
+*/
+
+// no pragma once
+
+#include "Error.h"
+
+#undef assert
+#ifdef DFHACK_ASSERT
```
lethosor: Is it possible to just use NDEBUG? I think CMake already sets that depending on the build mode, so we wouldn't have to add a definition.

CMakeLists.txt
lethosor: Also, maybe add_definitions() would work (if that can apply to non-Release builds).

library/include/assert.h
BenLubar: NDEBUG is used by the standard library, so it'll probably break ABI compatibility, and if it doesn't now, it probably will eventually.

lethosor: I mean that CMake already defines NDEBUG to enable assert() in Debug/RelWithDebInfo (?) builds. We could potentially use that instead of a separate flag.

BenLubar: NDEBUG is not-debug, and is defined in both Release and RelWithDebInfo. Is it guaranteed to not affect anything other than `assert`?

lethosor: Oh, right, I had that mixed up. I thought Release was the only different one, though (i.e. Debug and RelWithDebInfo share the NDEBUG setting), although I could be wrong about that. 

I don't *think* NDEBUG affects anything besides assert() (see also [here](https://stackoverflow.com/questions/5473556/what-is-the-ndebug-preprocessor-macro-used-for-on-different-platforms)). If you ask me, `-g` is more likely what could break ABI compatibility (and the Windows equivalent does, IIRC).

PR #1116, opened 2017-06-21
Fix forum thread link in README.md

alanbrady:
The README is pointing to the old thread, which in turn leads to an old download.


lethosor:
Thanks! Sorry for the mess there. I'm experimenting with putting documentation hotfixes in the `docs` branch so they can be merged into `master` and `develop` more easily (`master` is just for stable releases).

PR #1119, opened 2017-06-25
fixing getItemBaseValue() for cheese, sheets and instruments

pronvit:


lethosor:
Verified that this works with 995 items from a random caravan, so this looks good to me.

PR #1131, opened 2017-07-11
Make repeatUtil.cancel work even when called from the callback.

### LINKED ISSUE #1122: repeat-utils functions can't cancel themselves
This was pointed out by [Atkana](http://www.bay12forums.com/smf/index.php?topic=164123.msg7497491#msg7497491) on the forums:

> ```lua
> local blah = 1
> local id = "foo"
> local repeatUtil = require 'repeat-util'
> 
> local function doStuff()
>     print("Blah: " .. blah)
>     blah = blah +1
>     print("Blah is now " .. blah)
>     if blah > 10 then
>         print("Blah is more than 10. Stopping.")
>         print("Has supposedly cancelled = " .. tostring(repeatUtil.cancel(id)))
>     end
> end
> 
> repeatUtil.scheduleEvery(id,50,"frames",doStuff)
> ```

This doesn't work because repeat-utils schedules the next call of doStuff() *after* doStuff() returns, regardless of what doStuff() does. Scheduling it before doStuff() could result in error message spam, but tracking calls to cancel() and avoiding rescheduling doStuff() if it cancelled itself could work.

BenLubar:
Closes #1122.

### DIFF
--- a/library/lua/repeat-util.lua
+++ b/library/lua/repeat-util.lua
@@ -17,7 +17,9 @@ function cancel(name)
  if not repeating[name] then
   return false
  end
- dfhack.timeout_active(repeating[name],nil)
+ if repeating[name] ~= -1 then
+  dfhack.timeout_active(repeating[name],nil)
+ end
  repeating[name] = nil
  return true
 end
@@ -26,8 +28,11 @@ function scheduleEvery(name,time,timeUnits,func)
  cancel(name)
  local function helper()
   func()
-  repeating[name] = dfhack.timeout(time,timeUnits,helper)
+  if repeating[name] then
+   repeating[name] = dfhack.timeout(time,timeUnits,helper)
+  end
  end
+ repeating[name] = -1
  helper()
 end
 


library/lua/repeat-util.lua
```diff
@@ -17,7 +17,9 @@ function cancel(name)
  if not repeating[name] then
   return false
  end
- dfhack.timeout_active(repeating[name],nil)
+ if repeating[name] ~= -1 then
+  dfhack.timeout_active(repeating[name],nil)
```
lethosor: Does this work? It looks like this might be inverted, but I'm not sure.

BenLubar: I'm using -1 as a placeholder value for repeating tasks that are running for the first time and haven't been re-scheduled yet.

lethosor: But cancel() should still work for those

BenLubar: Yes, setting the timeout ID to nil prevents it from re-setting the timeout. The if on line 20 is just there to avoid an error message if `timeout_active(-1,nil)` ever starts responding with error messages.

lethosor: Ah, okay. I got confused because you weren't using nil, I think.

PR #1139, opened 2017-07-29
labormanager: fix #1103

### LINKED ISSUE #1103: labormanager idles the mayor too readily
There is reportedly some sort of bug with meetings (http://www.bay12games.com/dwarves/mantisbt/view.php?id=9262) and this may be leading labormanager to inappropriately idle the mayor

also, labormanager may be willing to idle the mayor to have meetings with people who are dead, asleep, resting, or on military duty; need to check for these conditions.

ab9rf:
This resolves #1103. 

Open PR #1137 is incorporated into this PR.


plugins/labormanager/labormanager.cpp
```diff
@@ -2078,15 +1212,28 @@ class AutoLaborManager {
...
+                        if (other && !(other->flags1.bits.dead ||
+                                       other->job.current_job->job_type == df::job_type::Sleep ||
+                                       other->job.current_job->job_type == df::job_type::Rest ||
+                                       ENUM_ATTR(profession, military, other->profession)))
+                        {
+                            dwarf->clear_all = true;
+                            if (print_debug)
+                                out.print("Dwarf \"%s\" has a meeting, will be cleared of all labors\n", dwarf->dwarf->name.first_name.c_str());
+                            break;
+                        }
```
BenLubar: You've got a trailing space here that's causing the build to fail.

PR #1140, opened 2017-07-29
Fix #1115 (penalize elves for woodcutting jobs)

### LINKED ISSUE #1115: labormanager should avoid assigning woodcutting jobs to elves
If there's a generalized way to find jobs that will make a unit unhappy to do, using that would be a Good Idea too.

ab9rf:
This resolves #1115.

Open PR #1137 is incorporated into this PR.


lethosor:
Should I just merge this instead of #1137, then?

lethosor:
Should I just merge this instead of #1137, then?

ab9rf:
Each of the bugfixes is independent of the others but all incorporate the refactoring. If you merge any of the bugfixes, the refactor will become a null merge.

PR #1150, opened 2017-08-12
Fix luasocket receive with byte count

warmist:
Did not correctly detect when you typed in a number instead of pattern.


lethosor:
"dev plugin" is for the plugins in plugins/devel, by the way.

PR #1156, opened 2017-08-28
max_barrels was serialized on max_bins position

ViTuRaS:
When I save and load stockpile settings with stockpile management plugin, then saved barrels count will be loaded in bins and barrels are zero. So I think this change should correct it.


ViTuRaS:
I think, I had to edit the file in develop branch and not in master. Sorry.

ViTuRaS:
OK, now the branch should be correct. Somehow I don't see any way to recheck by Travis CI.

lethosor:
Thanks!

PR #1159, opened 2017-09-01
Add embark-assistant plugin

PatrikLundell:
The plugin is actively invoked on the embark map to help select an embark site and is either actively disabled or automatically disabled as an embark is loaded or the embark map is unloaded. It contains a set of help screens describing it found in plugins/embark-assistant/help_ui.cpp, with the "main" plugin file being plugins/embark-assistant/embark-assistant.cpp.

The plugin provides three services:
- Display of normally invisible sites overlayed on the region map (i.e. lairs, shrines, etc).
- Enhanced embark resources display (more correct assessment of clay, flux, and aquifers, plus the set of metals and economic materials available). This functionality uses the logic from Prospector adapted for this plugin's purposes.
- Embark find functionality replacing the vanilla version (becomes available if the plugin is exited) containing a wider selection of search criteria.

I've developed it with a win32 DF and tested it (briefly, to see that it seems to work) with a win64 one. I have no access to other platforms, but I'm not aware of anything used making it platform dependent.

The plugin uses Ragundo's biome_type.cpp code with minor adjustments to make it usable in this plugin.


PatrikLundell:
It appears my commit somehow failed to provide biome_type.cpp,. I'll be back when I've sorted that out with a new request...

lethosor:
No, you had that file. The issue is this line in that file:

```
#include "../../include/dfhack.h"
```
I'm not sure what you're trying to include, but you should rely on DFHack's include paths if you can, e.g. by including just "DFHack.h". It might also be an issue with case sensitivity, because DFHack.h and dfhack.h are different files on Linux.

PatrikLundell:
Thanks lethosor. I would probably never have figured that out, since it does work locally. Since the code is copied from Ragundo, I haven't really looked at what it does, but I think I added an extra level of indirection because the file couldn't be found when placed in my plugin directory.
Anyway, your help gives me something to investigate.

lethosor:
You're including dfhack.h still, so that might not work

PatrikLundell:
Ah, sorry, I misunderstood you. Making a new attempt...

ragundo:
This is really good stuff
I need to dig into this code!
Love that more people start developing plugins

plugins/embark-assistant/biome_type.cpp
```diff
@@ -0,0 +1,752 @@
+/* The code is copied from Ragundo's repo referenced below.
+The changes are:
+- The addition of a .h file reference.
+- The simplification of the code using ofsub to remove the use of (and
```
ragundo: Yeah
That's was one artifact from IDA reverse code generation.
I never really understood its purpose, and I see now that is has no effects in the code removing it.
I take note of this and change the code in my code.
Thanks!!!

BenLubar:
Not necessarily as part of this PR, but it would be nice to get the biome thing split into a library like buildingplan-lib is currently set up.

PatrikLundell:
I definitely agree it would be very useful to have the biome determination code provided as a utility function (accessible from scripts as well: that would save me from half a dozen script copies...). I suggest my modification to allow the hosting world tile's y coordinate to be used for tropicality determination sould be included in that case.
@ragundo: I've made a few scripts that contain some stuff not in this code, and I'll continue to use scripts when I can, given the rather heavy and slow cycle of plugin deployment and bug fixing. In this case the script prototype was about an order of magnitude too slow (which I expected), so code was the only way to go. I'm not blaming you for the oddities in the reverse code generation: that's to be expected. I've done some additional minor prettyfying in the script version, but the OFSUB thing was the only one that bugged me enough to go into your code (it took me quite some time to figure out what it was supposed to do, and Lua doesn't do bitwise arithmetics, so I either had to understand it or emulate it. Once you know that it is it isn't much of an issue, though). If you have use for my comments, it's a small payback for you making the code available in the first place.

lethosor:
Seems to work on macOS in 0.44.05. I had to fix a couple fields, and might want to clean up some UI stuff (using getKeyDisplay for key names, adding spacing for consistency, etc.), but other than that it looks good.

PR #1168, opened 2017-09-23
Update the rendermax plugin with warmist's changes

Clyybber:
Warmists updated version contains some optimizations

### DIFF
--- a/plugins/rendermax/renderer_light.cpp
+++ b/plugins/rendermax/renderer_light.cpp
@@ -521,13 +521,16 @@ rgbf lightingEngineViewscreen::propogateSun(MapExtras::Block* b, int x,int y,con
     {
         ret*=matStairCase;
     }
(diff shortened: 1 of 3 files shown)


Clyybber:
Will probably need to remove the whitespace after the comment in renderer_light.cpp at line 1327 to make travis-ci content. It built successfully though.

warmist:
So okay i must comment why I haven't issued a pull-req:

- Not sure if it's faster for everyone and how much faster
- It has a TODO and i'm not sure if it's implemented
- It still needs a sun propagation rewrite, which would move the sunlight to threaded part of work
- MOAR profiling just to see what it slow
- I'm not sure if i like how it looks.

So it can be merged in but i'm hoping for some more work by someone?

warmist:
So sorry to do this, but i've made another branch which will use opencl with rendermax. Hopefully that will increase the fps for everyone :)

Clyybber:
No problem, thank you for your work!

PR #1173, opened 2017-10-03
Allow external connections to DFHack remote services.

JapaMala:
This needs a bit more work before it's ready to merge, but I figured I'd get some feedback here anyway.

The idea is that the more dangerous functions, like the "run any command" or "run arbitary lua code" functions are barred if they don't come from localhost, while relatively safe functions can be run from other computers or devices, like reading the map, or setting designations, etc.


lethosor:
Is this merge-ready? Also, are the conflicts irrelevant (i.e. can I just go with the develop branch versions of the RFR files), or are there important changes here?

It appears that spoofing localhost's IP isn't realistic, but as long as DFHack still listens only on 127.0.0.1 by default, this should be fine (anyone enabling this feature should understand the risks).

JapaMala:
I'll have to take a look.

I want both the enabled state, as well as the port, to be settable in a json file. I'll have to check to make sure they are.

lethosor:
Cool. Looks like the only build issue was a missing script, probably due to submodule stuff I can handle.

lethosor:
@JapaMala did you check?

JapaMala:
Looks like it's in there. Should be fine.

lethosor:
Getting this on a fresh install on OS X:
```
terminate called after throwing an instance of 'Json::LogicError'
  what():  Value is not convertible to bool.
```
I suspect it has something to do with the config file not existing.

JapaMala:
I'll look into it.

PR #1174, opened 2017-10-04
auto_ptr is deprecated

BenLubar:
use unique_ptr instead, ~~with the C++14 helper function make_unique.~~

### DIFF
--- a/library/RemoteServer.cpp
+++ b/library/RemoteServer.cpp
@@ -261,7 +261,7 @@ void ServerConnection::threadFn()
             break;
         }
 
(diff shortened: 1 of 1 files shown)


BenLubar:
Ugh, we're using C++14 on Windows but C++11 on Linux?

lethosor:
make_unique would be nice. Does GCC 4.8 not support that?

BenLubar:
It apparently requires GCC 4.9 or higher.

lethosor:
We could write our own pretty easily. I'm not sure forcing a GCC upgrade on people would be suitable just for that.

(This solution also works, of course.)

Issue #592, opened 2015-03-28
Autochop: optionally skip cutting down fruit trees


PeridexisErrant:
As a response to [this issue](http://redd.it/30m6da) on Reddit.

lethosor:
Done in #1180 thanks to @dtimm

PR #1180, opened 2017-10-17
Add tree product exclusions to autochop

dtimm:
Addresses issue #592 .

### DIFF
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -16,6 +16,7 @@
 #include "df/items_other_id.h"
 #include "df/job.h"
 #include "df/map_block.h"
+#include "df/material.h"
 #include "df/plant.h"
 #include "df/plant_raw.h"
 #include "df/tile_dig_designation.h"
@@ -48,6 +49,9 @@ static bool autochop_enabled = false;
 static int min_logs, max_logs;
 static const int LOG_CAP_MAX = 99999;
 static bool wait_for_threshold;
+static bool skip_fruit_trees;
+static bool skip_food_trees;
+static bool skip_cook_trees;
 
 static PersistentDataItem config_autochop;
 
@@ -179,6 +183,9 @@ static void save_config()
     config_autochop.ival(1) = min_logs;
     config_autochop.ival(2) = max_logs;
     config_autochop.ival(3) = wait_for_threshold;
+    config_autochop.ival(5) = skip_fruit_trees;
+    config_autochop.ival(6) = skip_food_trees;
+    config_autochop.ival(7) = skip_cook_trees;
 }
 
 static void initialize()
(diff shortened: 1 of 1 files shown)


lethosor:
Nice! I was considering using JSON or something for this, since we're running out of space in the persistent entry (as you can probably tell), but that would be considerably more effort to set up, and this looks good. 

I'll try to test it on my end soon, although I unfortunately haven't had a lot of time for DFHack things recently, so I can't make any promises.

plugins/autochop.cpp
```diff
@@ -629,13 +688,16 @@ class ViewscreenAutochop : public dfhack_viewscreen
...
                     x = std::max(x, prev_x + 10);
-                    for (size_t j = 0; j < sizeof(row.skeys)/sizeof(row.skeys[0]); ++j)
+                    for (size_t j = 0; j < sizeof(row.skeys) / sizeof(row.skeys[0]); ++j)
                         OutputString(COLOR_LIGHTGREEN, x, y, DFHack::Screen::getKeyDisplay(row.skeys[j]));
                     OutputString(COLOR_WHITE, x, y, ": Step");
                 }
                 OutputString(COLOR_WHITE, x, y, "", true, left_margin);
             }
             OutputHotkeyString(x, y, "No limit", CUSTOM_SHIFT_N, true, left_margin);
+            OutputToggleString(x, y, "Skip Fruit Trees: ", "f", skip_fruit_trees, true, left_margin);
```
lethosor: It might be better to use CUSTOM_F instead of "f" here (same for the two lines below). I can't think of a reason why anyone would bind CUSTOM_F to anything that isn't "f", but I've seen things like that before.

dtimm: I was trying to follow the style of the other `OutputToggleString` out of an abundance of caution. I'll clean up all the direct letter references here and test them out.

lethosor: Yeah, the ones that take an interface_key are newer than most plugins that use OutputToggleString(), etc.. Thanks!

plugins/autochop.cpp
```diff
@@ -197,6 +207,9 @@ static void initialize()
         min_logs = config_autochop.ival(1);
         max_logs = config_autochop.ival(2);
         wait_for_threshold = config_autochop.ival(3);
+        skip_fruit_trees = config_autochop.ival(4);
```
lethosor: Just noticed that you're writing to ival(5-7) but reading from ival(4-6). Anyway, this could probably use just one ival() entry, but it doesn't really matter since (hopefully) we'll use JSON or something else soon.

lethosor:
Thanks, and sorry for the delay!
I did fix one potential issue - the ival() fields default to -1, so although it appeared that you were trying to disable these new restrictions by default, they could have been getting enabled because of that. (I don't know for sure if that was happening, because I switched it to just use ival(4), but it was definitely an issue after I did that.)
Anyway, looks good now.

PR #1181, opened 2017-10-20
map-render plugin

warmist:
So this is stuff ~~stolen~~ inspired by and directly copied from mifkis [repo](https://github.com/mifki/df-twbt/blob/master/twbt.cpp)
I'm using it for remote server and currently only rendering is done. Maybe later this would benefit to being moved into a module as other plugins might need e.g. loading up a texture


warmist:
before merge, somebody explain how to rebase so there is only 1 (or 2) commits. The ones in the middle somehow made it in, not sure what happened.

moonheart08:
Personally going to also request a merging for this. It will be *very* helpful.

warmist:
Note to self: add used offsets to the xml lib

pronvit:
You should remove all that code you don't need, e.g. related to tileset loading.

lethosor:
@warmist we need 0.44.05 offsets for this, and ideally a way to find them quickly for future DF versions, or it might be hard to support for early builds for future versions.

pronvit:
@lethosor btw, now there's a script to automatically find twbt offsets, FWIW.

lethosor:
Not seeing it mentioned in https://github.com/mifki/df-twbt/commits/master - is it somewhere else?

lethosor:
Would another name be appropriate for this? I don't want people to confuse it with TWBT, and it's just a render_map implementation at the moment.
The symbols we had in df-structures are long gone, by the way. I'm still not sure how to automate finding them, and manually-located offsets have a pretty low priority (RFR uses one to load art images that has been missing for a few releases now).

warmist:
Okay so the initial idea was to add stuff twbt uses for other people to use (hence the name). However I haven't been as interested in the image loading or the other functions twbt uses and has offsets for. 

I've made an "attempt at automating offset finding" here: [gist](https://gist.github.com/warmist/26319f3b49dbd7ac4733306621bcb386) (needs to be called after 'unload -all' due to render method interpose)

The script itself needs some documentation and could use more checks (e.g. see if instructions around the call are similar to the ones in twbt page). For it to be as correct as possible, it needs a disassembler (or some part of one) (e.g. i've used this one: https://github.com/zyantific/zydis with some success) but IMO for one plugin to add another  (non-one-header) dependency seems a bit overkill. So if anything could use this too, it has my vote :)

plugins/twbt-utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+    CoreSuspender suspender;
+
+    int x = luaL_checkint(L, 1);
+    int y = luaL_checkint(L, 2);
+    int z = luaL_checkint(L, 3);
+    int w = luaL_checkint(L, 4);
+    int h = luaL_checkint(L, 5);
+    //backup state
+    uint8_t *s = df::global::gps->screen;
```
lethosor: Is this supposed to be a backup or just an alias?

plugins/twbt-utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#ifdef WIN32
+    // On Windows there's no parameter pointing to the map_renderer structure
+    typedef void(_stdcall *RENDER_MAP)(int);
+
+    RENDER_MAP _render_map;
+
+    void render_map(){ _render_map(0); }
+#else
+REQUIRE_GLOBAL_NO_USE(map_renderer)
```
lethosor: Might as well use `REQUIRE_GLOBAL`

plugins/twbt-utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+    int z = luaL_checkint(L, 3);
+    int w = luaL_checkint(L, 4);
+    int h = luaL_checkint(L, 5);
+    //backup state
+    uint8_t *s = df::global::gps->screen;
+    int32_t win_h = df::global::gps->dimy;
+    int32_t was_x = *window_x;
+    int32_t was_y = *window_y;
+    int32_t was_z = *window_z;
+    int32_t gx = df::global::init->display.grid_x;
```
lethosor: Missing `REQUIRE_GLOBAL(init)`

plugins/twbt-utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+    render_map();
+    //restore state
+    *window_x = was_x;
+    *window_y = was_y;
+    *window_z = was_z;
+    df::global::init->display.grid_x = gx;
+    df::global::init->display.grid_y = gy;
+
+    lua_createtable(L,w*h*4,0);
```
lethosor: I think a `PenArray` would work well here, but I'm not sure if there's currently an API to create those outside of LuaApi.cpp.

plugins/twbt-utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+DFHACK_PLUGIN_LUA_COMMANDS{
+    DFHACK_LUA_COMMAND(render_map_rect),
+    DFHACK_LUA_END
+};
+
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector<PluginCommand> &commands)
+{
+    auto addr =reinterpret_cast<RENDER_MAP>(Core::getInstance().vinfo->getAddress("twbt_render_map"));
```
lethosor: This is the only address needed, right? Your script finds two more (plus the vtable):
```
[DFHack]# find_twbt
Vtable: : 1022e8f50 1022e8f50
render : 100490bf0 100490bf0
dwarfmode_render_main : 100490590 100490590
A_RENDER_MAP : 100b1c040 100b1c040
```
(0.44.10 osx64)

plugins/twbt-utils.cpp
warmist: Initial idea was to have a std::vector for "local render buffer". However when i tried it, it did not work, not sure why though.
The rest of state (window_x/y/z, grid_x/y) is backuped.

plugins/twbt-utils.cpp
warmist: I forget how it works, but there was a system in place to push c++ stuff into lua.

warmist:
Not sure if i'll have time. Hopefully today (in 8 hours or so). Imho unsolved problems:

- New name
- Are offsets found in ALL supported builds. If not what can we do.
- Would it be better to add functions somewhere else (e.g. new module which would have all non-virtual DF functions?) so that other plugins could use it?

warmist:
So this is as done as i care to have it. Some notes:

- Maybe it would be nice to use Screen::PenArray as return but that just complicates setting it up. Also i'm using it to send to http clients so there is no difference (AFAIK). This would help if you want to have a little window that shows e.g. "what your dwarf is doing" on the side.
- Not sure how to add the required offset to dfhack procedure of finding offset when new version is out. This plugin needs `twbt_render_map` which could be found from twbt plugin or [my script](https://gist.github.com/warmist/26319f3b49dbd7ac4733306621bcb386) which i modified to print only the interested offset.

Issue #720, opened 2015-11-01
built-in commands aren't included in try_autocomplete() suggestions


lethosor:

PeridexisErrant:
Related to #696?

lethosor:
Sort of, but there isn't a good way to mention that (like on Mantis). This issue is just with the suggestions you see when typing an incomplete command, such as "di", which include commands like "diggingInvaders" but not "die".

lethosor:
(Forgot to close with #1189)

PR #1189, opened 2017-11-27
Autocomplete built in commands (die, etc.)

Pfhreak:
Addresses https://github.com/DFHack/dfhack/issues/720

Output:
cl is not recognized. Possible completions: cls cleanconst clean cleanowned clear-smoke
di is not recognized. Possible completions: die disable digexp digcircle digFlood diggingInvaders digfort
he is not recognized. Did you mean help?
... etc.

Commands still work correctly. (Tested with cls, help, die, etc.)

Happy to take feedback!

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -429,14 +429,37 @@ command_result Core::runCommand(color_ostream &out, const std::string &command)
         return CR_NOT_IMPLEMENTED;
 }
 
+// List of built in commands
+static const std::set<std::string> built_in_commands = {
+    "ls" ,
+    "help" ,
+    "type" ,
+    "load" ,
+    "unload" ,
+    "reload" ,
+    "enable" ,
+    "disable" ,
+    "plug" ,
+    "keybinding" ,
+    "alias" ,
+    "fpause" ,
+    "cls" ,
+    "die" ,
+    "kill-lua" ,
+    "script" ,
+    "hide" ,
+    "show" ,
+    "sc-script"
+};
+
 static bool try_autocomplete(color_ostream &con, const std::string &first, std::string &completed)
 {
     std::vector<std::string> possible;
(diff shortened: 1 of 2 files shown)


strainer:
Hi. If I can test being tricky about this - I think it is a possible source of confusion to call it autocomplete. Autocompletes typically save keystrokes by allowing selection of longer inputs while typing. This feature responds to entered invalid input with possible completions, which if sought or appropriate, will still need typed and entered in full. To my reckoning this is a 'search help on error' response. Its kind of saying "did you mean....?" or "are you (implicity) searching for .... ?"
So, I suspect this feature would be best as an explicit help command eg. input:"?cle" output:"completions for cle: clean-stuff, clean-all ...etc" rather than a potentially unsought suggestor on input errors.

To reduce keystrokes, short command aliases seem necessary, and if/when they are implemented a potential hazard of mis/typing 'help searches' that accidentally match short commands would arise from having 'error guessing' on default. It could be cleaner if the help search is done explicitly, ie: "For help on specific commands ?clea lists commands beginning with 'clea' "

lethosor:
This is an existing feature that's been around for years. It doesn't try to correct commands (although it used to at one point, I think) - it only displays suggestions. You can press the up arrow to avoid retyping previous commands. The only thing this PR adds is support for commands that are implemented in the DFHack core - these suggestions currently just include script and plugin commands.

strainer:
Thanks for filling me in, if its already set in that's fine. Its just not really a typical 'autocomplete' - for the sake of some feedback. Pressing up for previous commands is a terminal feature, which doesn't access these suggestions, but I can see they could be useful.

lethosor:
We had to implement the up arrow feature ourselves, actually, but it's true that these suggestions aren't used there.

PR #1197, opened 2017-12-14
Made console prompt respect color reset

### LINKED ISSUE #1123: The DFHack console can get its colors confused.
I suspect the DFHack console thread has an unsafe handover between the command executed and DFHack's own prompt.
A couple of small scripts used to produce odd results (at least on my system):
```lua
function y ()
  for i = 1, 1000 do
  dfhack.printerr ("Do I bleed?")
  end
  dfhack.println ("What?")
--  error ("Not really...")
--  dfhack.println ("What?")
--  dfhack.println ("What?")
end

y ()
```
and
```lua
function z ()
  dfhack.color (COLOR_BLUE)
  dfhack.println ("What color is this?")
end

z ()
```
Each test is run on a freshly started DF (with an associated DFHack terminal). Script output were as expected unless noted.
- Running y causes the DFHack prompt to become red (actually light red), and if I run z after that I then get a blue prompt. Typing garbage (in blue) resulted in a red error message and a red prompt.
- Running z followed by y caused the final standard color output line from y to become blue followed by a blue prompt.

PatrikLundell:
The function originally used writes characters while retaining whatever attributes the cells written to contained before, effectively rendering the color reset before the call to refresh the prompt useless.

Closes #1123


lethosor:
Looks good to me, although I can't test it on non-Windows. Does it fix the issue with the snippets you posted in #1123?

PatrikLundell:
Yes. It's the same code as the one used back then, although I've run a new (brief) test to verify it before making the pull request.
It obviously doesn't do anything about "naughty" scripts that leave a color designation behind, affecting all subsequent output, but that's a completely different issue (which presumably would be platform independent).

lethosor:
Cool. What if you run these two commands (in order)?
```
:lua dfhack.color(COLOR_BLUE) print('test')
:lua print('test')
```
Both should print in blue.

PatrikLundell:
Yes, they do (and the prompt is white). These commands give the same result in the original version as well, so I assume that's a regression test to verify things haven't been broken by the change.

lethosor:
Yeah, I just wanted to make sure it wasn't changing existing behavior. Sounds good!

PR #1201, opened 2017-12-21
Add cxxrandom plugin

lethosor:
By @cppcooper
For reference - this is an alternative to the dfhack.random module for people (like him) who need C++11's random functions for some reason.


plugins/cxxrandom.cpp
```diff
@@ -0,0 +1,312 @@
...
+    return CXXRNG::Instance().GenBool(p);
+}
+
+void BlastDistributions()
+{
+    CXXRNG::Instance().BlastDistributions();
+}
+
+DFHACK_PLUGIN_LUA_FUNCTIONS {
+    DFHACK_LUA_FUNCTION(ResetIndexRolls),
```
lethosor: Change to camel case for consistency

strainer:
I read into it a bit and its not exporting C11s methods its using C11s random in the course of introducing its own helper functions. Im not sure if its a most suitable addition to dfhacks helpers.

dfhacks random module currently lacks a normal/gaussian distribution method, but the
gaussian method here is trickily implemented so unclear if it works well without testing. 

The RollBool method is remarkably convoluted for what can be done simply by: rndbool  = random() > 0.5 f;

Its not clear what the purpose of the use of internal caching of "Distributions" is except for serving the public "RollIndex" method, which is ambiguously named, seems to return a random string, that could with care be used as an index, or used as anything else.

The BlastDistributions method should soberly be named ClearDistributions, or publicly something like ResetRndStrings. 

This private helper library also lacks a seeding method to get deterministic results, thats something that a games random helper lib should really have.

I suggest if a random helper lib is needed a popular tested lib should be found.

cppcooper:
@strainer 
The bool roller may be convoluted but guarantees consistent results.
With lua's built in random functionality all it takes is pulling random numbers somewhere else to skew the expectations. Take a look.
https://ideone.com/XlRx18

Using the C++11 random features, problems like that go away.

When I first laid out the plan for the plugin I actually didn't know how powerful the implementations were. First I didn't realize that the number distributions have no internal state, or at least this is my current understanding. Instead the random engine has an internal state, and despite this you can use the same engine on multiple distributions (with varying numbers of use) without skewing your results. Extremely powerful especially if you want to have an exact 66% chance of rolling true, or generating values within a normal distribution.

The bottom line I found was that you can use them in any way and not screw up your expectations for results.

Consequently now that I know this I've been planning to revise the code.
1) I'll be removing the distribution storage
2) I will add a seeding function to seed a global engine
3) The BlastDistributions, will be scrapped all together

I'd love to hear alternatives to RollIndex, because I was unable to think of a less ambiguous name. Speaking off, since that is not from c11 I figure I may actually just reimplement that in lua. On the other hand though I need the Unique ID mapping for my uses. So I'm a bit on the fence actually.

strainer:
Thanks for reply cppcooper, I did not intend my criticism to read as frosty there.

I did a fair bit of work rolling and testing my own personal prng and helper functions exercised  [here](http://strainer.github.io/Fdrandom.js/) 

In the course of that I found the kind of test which you linked which involves just a few thousand variates to be very deceptive. I dont think it shows what you think it has. Charting a few million results should be revealing. For one test I needed to assess trillions of results to mine for bias in the noise floor of a double generator, amazing that is possible on PCs these days !

It is my understanding that luas math.random should be very fine for all non cryptographic purposes, lua is a mature academic product, exceptionally well coded.

Thing is the distribution of a set of random variates cannot be influenced by skipping some unless you test them before skipping them - this is fundamental to the nature of a random sequence. Because of this sets of a certain distribution can have no special benefit from being stored together, each variate is as likely as any other to be found at any value of the distribution. They are all as they say " independant variables "

Ive not much experience of cpp so Im not sure but suspect using string as a convenient container of integers is problematic. 

I would call RollIndex something like RollRndBuffer, but I am confident there should be no benefit from sheparding these in any particular order.

cppcooper:
You're probably right, I didn't extensively test the lua functionality. So it may not be revealing what it appears to.

I did do so with the C++ random functionality however. I used a sample set of originally ~34 million, then tested under different usage patterns. Each test showed they narrowed down around the expected values.

So I've updated the code base for cxxrandom. I had been overly cautious in my implementation, but it is now more compact and avoids unnecessary caching of distributions. The simple number distributions are still cached, since they have a state to maintain.

I'm not sure what you mean by "using strings as a convenient container of integers" but I do assume you're reffering to the unordered_map in the `std::unordered_map<std::string, SimpleNumDistribution>`. I will quote cppreference.com: "Unordered map is an associative container that contains key-value pairs with unique keys. Search, insertion, and removal of elements have average constant-time complexity." 

std::string can be hashed out of the box, so there is no problem with this code. I thought about using ID numbers, but that would have made using it more difficult than it needs to be. It is the sort of code that gets used in loops to select random array elements. With that said, I'm not sure RollRndBuffer gets that point across either, so perhaps I'll renamed it to RollArrayIndex.

On the gaussian matter. Yesterday I modified the test code put on display at http://www.cplusplus.com/reference/random/normal_distribution/ in order to run testing on the std::normal_distribution. I performed a larger test with a sample size of exactly 13 million. It isn't perfect (I'm fairly sure I messed up the math to confine results within a range of 0-10, I should have probably rewritten it using a map but owell)
https://ideone.com/ZRbyM6

All of the C++ <random> library has been well tested, I don't think it possible for me to screw up expectations regarding random results. Although as established it is more than possible for me to unnecessarily complicate things, which I hope has been remedied.

--
I asked lethosor to make this pull request because I use `git push --force` frequently to rewrite commit history. However I may just create a branch of a branch and do my best to avoid merging too early to the main branch; I'm a bit rusty with my git habits. The main code repo is here
https://github.com/cppcooper/dfhack-cxxrandom/tree/cxxrandom

strainer:
I see the strings are keys to a map of vectors of shorts which are initialised with 1 to len and later, filled with random shorts maybe, perhaps that is what the exported lua function  rng.RollArrayIndex does. 

Maybe Ive been obtuse ill not complicate this further. I think a helper lib would ideally be clearer than this. I dont think there is any problem with pulling rnds from standards prng out of order, they are psuedo-not in any order. You should get that research reviewed by others rather than defend it here.

What I think matters not, just some critical input - happy hols'

lethosor:
Closing in favor of #1204

PR #1204, opened 2017-12-23
Add cxxrandom plugin

cppcooper:
This pull request is to offer some basic support for random number generation using C11's random library. Related to [pull request](https://github.com/DFHack/dfhack/pull/1201) #1201  a new pull request is being made so I can update the code directly if and when recommended, or required, changes are made.

The library offers 20 number distributions, but I've implemented only the most basic. More can be added if it is deemed necessary or desirable, with extreme ease. Currently only one random engine is used, the [std::default_random_engine](http://en.cppreference.com/w/cpp/numeric/random). Which is implementation-defined, so I've considered replacing it with the [std::mt19937](http://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine) (ie. std::mersenne_twister_engine).

So currently only the 4 distributions are implemented:
- [std::normal_distribution](http://en.cppreference.com/w/cpp/numeric/random/normal_distribution)
- [std::bernoulli_distribution](http://en.cppreference.com/w/cpp/numeric/random/bernoulli_distribution)
- [std::uniform_int_distribution](http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution)
- [std::uniform_real_distribution](http://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution)

Below are the helper functions exported for lua. There is currently one pressing issue: the name of `rollIndex` and `resetIndexRolls`

rollIndex is a helper function for randomizing the iteration of a table array. It has been agreed that it is not named clearly enough, but a precise alternative has not yet been arrived upon. It really just generates integers [1-N], they are kept in a vector and shuffled, and then the shuffled vector is iterated through for random numbers. This was meant to ensure each number in the series is returned exactly once in a loop of N iterations.

Since this method for rolling array index numbers requires internal states to be kept track of the distributions are paired with string identifiers which the user can setup at their own discretion. These are stored in an unordered map in case somebody gets extremely ambitious and for some reason wants a few hundred or hundred-thousand of these.
~~~~~~~~~
Implements helper functions for random number generation.
Implemented using C++11 <random> library.

Exported Lua Functions:
- seedRNG(seed)
- rollInt(min, max)
- rollDouble(min, max)
- rollNormal(mean, std_deviation)
- rollBool(chance_for_true)
- resetIndexRolls(string, array_length) --String identifies the instance of SimpleNumDistribution to reset
- rollIndex(string, array_length)       --String identifies the instance of SimpleNumDistribution to use
                                        --(Shuffles a vector of indices, Next() increments through then reshuffles when end() is reached)

 On branch cxxrandom-rel
 Changes to be committed:
   modified:   plugins/CMakeLists.txt
   new file:   plugins/cxxrandom.cpp
   new file:   plugins/lua/cxxrandom.lua


strainer:
I didnt pick up that actually was a shuffled index. Is it used something like, local myx = rng.ShuffledIndex(10)
...
rng.ReShuffle(myx)

That sure could be useful.
Usage example in help text too.

Also I read elsewhere that there may be issues with luas old ANSI rng. Sounds like we can trust C11s. Sorry I couldnt be convinced from your tests. 

I do prefer the plain term 'Rnd' to the more playful 'Roll' used here - a lone nitpick probably. I can appreciate the usefulness now thanks.

cppcooper:
I'm hoping I won't need to personally carve out some space in the help documentation file `index.rst` or something like that? The file is a bit daunting. However I will gladly explain how rollIndex works. Before that though, as a side note I tend to agree about Rnd vs Roll but this was for hacking a game so I deemed it ok for myself; perhaps when christmas has passed though =)

The way I use it is like so.
Main loop:
https://github.com/cppcooper/dfhack-lua-dorf-buffing/blob/ad3bfb91c9ed7b7980708066c6074e915161b418/pimp-it.lua#L481

GetRandomTableEntry function:
https://github.com/cppcooper/dfhack-lua-dorf-buffing/blob/ad3bfb91c9ed7b7980708066c6074e915161b418/pimp-it.lua#L97

I've never done a pull request before, so I'm not exactly sure if it would be better to post the code here directly, but I thought it would be cleaner this way (can't go wrong).

The TLDR and I didn't want to click a link version:
It basically goes like this..
~~~~~~~~
local tableLength = TableCounter(myTable)
rng.resetIndexRolls("Primary Identifier" .. secondary_identifier, tableLength) --I do this to make sure it starts in a fresh state
for i=1,tableLength do
    local index = rng.rollIndex("Primary Identifier" .. secondary_identifier, tableLength) --If you also read my code you know I made a function to handle getting any type of entry using rollIndex (not just arrays)
    myTable[index]...do-the-thing

strainer:
I think you'd be better just not listing those rollindex functions in the plugins description since no one will guess how to use them.

cppcooper:
Sure thing, leave it out of the description if it gets integrated. Ideally someone can find a way to clarify the purpose behind those functions, as it is a handy interface to have around. Particularly for persons, like myself, who find objects in lua to be a bit convoluted.

I've been wondering if it might be worth it to revise this plugin once more to implement an engine keeper for each distribution. Since the engine keeps state information and gets to choose what numbers get pulled next.. this makes me wonder if it doing this wouldn't create better results in smaller sample sizes.

strainer:
If there is not a known flaw with the RNG (as with the old C lib in some Lua distros) then generating many or few batches/'samples' of random numbers will not affect distribution at all. One tested-fast-seedable RNG is really the tool - only crypto has use for a whole keyring of PRNGs :) The distributions are the interesting thing for gaming.

lethosor:
Ok, I looked over this again and understand why it's useful now. I'm not sure I like the string-based ID approach, though. Moving it into the core would allow it to use metatables and perhaps be more similar to the existing dfhack.random API.

Do you know if creating a distribution instance for each function call has any significant overhead?

lethosor:
Oh, for (4), I wasn't concerned about speed. I was thinking having an actual object might be more Lua-idiomatic and/or consistent with the existing random API. Something like this:
```lua
gen = mkgen(array_length)
gen:rollIndex()
gen:rollIndex()
gen:resetIndexRolls()
```
or
```lua
gen = mkgen()
gen:rollIndex(array_length)
gen:rollIndex(array_length)
gen:resetIndexRolls(array_length)
```
I'm not sure why you would need to vary the array_length parameter for the same distribution, but the second example would make that possible.

Anyway, something like that could be implemented in cxxrandom.lua easily enough. If you have a better suggestion for the name than "mkgen" ("IndexRoller"?), I'd be willing to try implementing that.

As for (1)-(3), I was thinking more along the lines of DFHack's random API, which has a way (I think) to make standalone generators, like engines, but the idea is that you'd only have to do that once for each use you have for random numbers. It might be nice to have a way for different tools to use cxxrandom independently like that, but it's not a high priority.

cppcooper:
I like that idea, it would do all the things I want it to do. A generator object type for each type of distribution would be really nice. It would, hopefully, mean having a class with both a distribution and engine declared inside it, and thus provide the improvements **(3)** can hopefully grant users. It would keep the usage simple! And it would also remove the need for IDs.

I'm just dubious about object life cycles, and whether it would work the way I think it would. Would this be like exporting a class written in C++ to lua? If not how would we instantiate Engines and Distributions?

Also, this is currently a bit beyond my lua abilities. Though having recently learned javascript maybe I could make some sense of lua objects, though I think I'll still be short of what I need to enact this plan.

lethosor:
DFHack has its own class system. What I suggested in my comment above could just be a layer above the existing functions taking tokens, implemented in Lua. Something like this, maybe (untested):
```lua
Generator = defclass(Generator)
function Generator:init()
  self.key = tostring(math.random())  -- just need a unique key per instance here, could also use a global ID or something, maybe generated by the plugin to ensure that it's a unique string
end
function Generator:rollIndex(array_length)
  return rollIndex(self.key, array_length) -- note that this calls the plugin function because it's a global in cxxrandom.lua
end
function Generator:resetIndexRolls(array_length)
  resetIndexRolls(self.key, array_length)
end
```

strainer:
I beg the question, why not just :
```
cxxrandom = require "cxxrandom"
g = cxxrandom.Generator()
for i = 1, 5 do print(g:nextGaussian) end
```

A shuffled index is a useful thing that is difficult to make without a good shuffle function.  And various random distributions like Gaussian and Bernoulli are very useful and require implementation of researched algorithms to make them well...

but there is zero benefit from pre-buffering these distributions and initialising and resetting shuffled indexes to access them in 'extra' random order. 

That convolution spoils what could be a comprehendable and useful facility. 

If your api requires the user to manage this buffering and out of order access of what are already effectively completely disordered sequences, it will confuse as to what point there is to it - but there really is none. If there were a problem with the RNG or distribution transforms, re-disordeiring the output would only hide it in certain cases, and in this day&age there is no reason to employ buggy RNGs or distribution transform algorithms. 

The RNG is good here, the distribution transforms should be good, the convoluted out-of-order delivery doesn't fix anything. Leave it out and everything's right and proper. Keep it in and some day it'll dawn on you - it just doesn't do anything useful.

lethosor:
I'm not sure exactly what's convoluted about it. The shuffled vector is only reset when you explicitly call resetIndexRolls.

strainer:
Its not in itself convoluted, but out of order delivery of pseudo-disordered numbers is a pointless convolution.

If you dont believe it is pointless then ask, when is there any benefit from resetting this index? What is the purpose of declaring an index of length n and ever resetting it? Why rearrange transformed RNG output with an index of any length? 

There are no sensible answers to those questions, the theory and performance of modern RNGs and math transforms precludes it. I grok this but you dont have to take my word for it, just check if you really grok this re-indexing of a professionally developed RNG output and Academically researched distribution transforms.

lethosor:
Are you just complaining about the existence of resetIndexRolls? That I could understand, but I have certainly used things like rollIndex before, where independently-generated random numbers won't work.

strainer:
"wont work" - What is not working about any of these independantly generated random numbers? Nothing - the reindexing thing is just done on a hunch that if something isn't quite working it should improve things - its more likely to cause problems than anything.

I was not 'just complaining' (hey), i was trying to explain the folly here - that if other programmers who understand the generation of random distributions happen to use it, will get puzzled over and asked about in the future. Im done then - good luck with it. 

I wont bend your ears anymore since Im getting rhetorical questions instead of answers

lethosor:
If I want to choose each item in a container exactly once, randomly, calling randInt that number of times won't work. Maybe I'm misunderstanding what you're saying.

strainer:
"If I want to choose each item in a container exactly once, randomly..."  > you need to shuffle and index.

Sorry, I thought this was still being used to reorder the output of the distributions, and that the generator needed to be initialized with an index size, just to return gaussians etc. As I wrote before: a shuffled index is useful, there is just no need to require one be initialised for the generator to supply distributions of *independant* variates.

Sorry if that was clear already.

It is odd to initialize the generator with array length of a shuffled index that it should return later, while it also may just return independent variates :
```
gen = mkgen(a_length)
local shufixarray = gen:rollIndex() --will be of a_length 
local gaussianvar= gen:gaussian() --doesnt care about a_length
```

This is stanard:
```
gen = mkgen(optional_seed)
local shufixarray = gen:rollIndex(length)
local another_shufix = gen:rollIndex(another_length)
local gaussianvar= gen:gaussian() --no length ever involved
```

Also, this wouldnt work:
```
for i = 1, 5 do print(g:rollIndex(5)) end
```
g's rollIndex would keep getting re-initialised to length 5, or it would have to count how many times it has been called, and assume it is getting called in a single loop of fixed size.

So much easier to return the actual shuffled array for the user to employ. Even better return the function which shuffles an array:

```
for i = 1, 5 myindex[i]=i end
gen.shuffle(myindex) -- its shuffled
```

lethosor:
The generator API I suggested above is *only* for the random-index functionality. Everything else currently shares a random engine, e.g. https://github.com/DFHack/dfhack/pull/1204/files#diff-32ab864fba73fdda0f258d93d6d01347R158

lethosor:
As for your claim that
```
for i = 1, 5 do print(g:rollIndex(5)) end
```
wouldn't work, the generator state only gets reset if the array length passed in *changes*:
https://github.com/DFHack/dfhack/pull/1204/files#diff-32ab864fba73fdda0f258d93d6d01347R144

strainer:
"The generator API I suggested above is only for the random-index functionality."

Thats clear now.

"the generator state only gets reset if the array length passed in changes:"

So when I want to reroll an index of the same size, I have to know to first reroll it with a different size and then back again - thats "magic" .

lethosor:
That's what resetIndexRolls is for.

strainer:
You have function that creates a shuffled index of length (n) and returns its values in order, and recreates and resets itself only if you call it again with a different n.  Its called 'rollIndex' like its an 'index of rolls' (resetIndexRolls too) as well as the action 'roll an index'. Except the keys arent even what is commonly understood by 'rolls' because rolled things dont care what came before them...

And another function 'reset the index rolls' which which will recreate the shuffledindex (sic) even it is the same n as before, or maybe it just resets the hidden output counter to the beginning \:/

lethosor:
It's fairly obvious that resetIndexRolls resets the generator (producing new values) from reading the code. It takes the same handle as rollIndex. Sure, there are more idiomatic ways to inplement it, and better names to use, but it's still useful and functional in its current state, as far as I can tell.

cppcooper:
Strainer speaks a bit too academically for me, but if I followed things correctly the concern is about recreating a shuffled index distribution.

Nothing stops the user from using the same reference, with a different index length. This is why I have code to check the size used, if there is a mismatch with the distribution already in memory then it gets deleted under the assumption the user knows what they are doing. Realistically it should throw an error.

As for resetting shuffled index distributions, (**resetting** that is) this only takes place when the distribution is exhausted or if the user explicitly calls the reset function. I shouldn't really need to explain why this takes place, I hope.

It seems like `rollXY` is still a sticking point, if this does get transitioned into a lua class then I think it would be a nice thing to drop from the C code base. I only liked roll__ because it is short, and a bit playful. I would be far more in favor of class::range(min,max), class::shuffledIndex(ref,size), etc.

strainer:
fairly obvious from reading the code is not right for a public api. If people end up using it someone spends ages figuring out why they were only getting one pattern out of the index...

```
--common straightforward init, no unexpected hidden mechanics:
 gen = cxRand(seed)
 vf = gen:gaus() 
 vf = gen:bern()

--make new shuffle sequence and set output to first:
 gen:shuffleSeq(5) 

--print the seq
 for i=0,5 do print(gen:sequence()) end

 --reshuffle the seq
 gen:shuffleSeq(5) 

--get element three of the new seq
e=gen:sequence(3)

--get element e of another sequence (make cool patterns and stuff for proceeduraly generated landscapes etc)
e2=gen:sequence(gen_b:sequence())
```

To do that kind of thing with the api discussed, we'd need to copy the objects buffer into our own array, and then use arrays as we always have - May as well just return the array then, but if that's not straightforward (because its implemented as a vector or whatever) being able to index the sequence like an array is straightforward. Straightforward apis are what we need to put attention into being creative instead of just getting stuff to work.

strainer:
"I followed things correctly the concern is about recreating a shuffled index distribution"

Its about the clarity of creating and using this - I was confused that a sequence length was required to initialise the generator and perhaps employed to increase randomness in the distributions as it seemed last year when I questioned the boolean generation. But how its used internally is not a concern now, its the simplicity of the api which is pretty vital to how good a helper it is.

Its worthing hacking out.

cppcooper:
Alright I think I get it now. So you'd decouple the sequence from the shuffling of it? Thus making it possible to generate, perhaps as an example, a range of numbers and shuffle that range. Then only reshuffle it when explicitly stated?

We've been on and on about this, so I just want to make sure I understand where you're coming from before I start inputting new information/ideas.

strainer:
"Thus making it possible to generate, perhaps as an example, a range of numbers and shuffle that range. Then only reshuffle it when explicitly stated?"

Thats a good example of what someone might want to do with a shuffle. Also might want to loop through it in one direction, and then in reverse or another order. Or make an uneven pattern by reading  shfa(shfb()+shfc()). Or other less usual things. The automatic reshuffle would frustrate all but the simplest usage where the sequence is effectively disposed after one read.

"We've been on and on about this"
Well, this and other things like shuffling to generate booleans and to 'enhance' doubted RNG output. I use rngs for proceedural generation, (eg. [here](https://strainer.github.io/fancy/) is a 'galaxy') and developed a library to help so having an interest in the subject - I couldnt fathom why/how you packaged a shuffle feature in such an awkward way, including unexpected resetting and resizing mechanics just to use in one way - other ways would just have to know about them and work around them. With just a little refactoring youve got a helper that can be used in different ways and without having to know unexpected mechanics.

cppcooper:
Well I suppose I packaged it that way because I was creating a random helper library and not a general purpose helper library. So I was in the mindset of implementing things that provide a random number generation.

Perhaps the shuffle and range sequence could be added to another helper in dfhack? Then I could just rewrite and hook into those while preserving the library's intent.

If not I'm mostly ok with making it more generalized. Not sure how I'd go about returning an array to lua, or taking one in and shuffling it. (taking in an array actually sounds like it could be a bit dangerous tbh resulting in index out of bounds undefined behaviours)

strainer:
Amongst earlier ramblings I did mention how you can give array like access to the sequence without returning an array if that is difficult. I would just call it the "sequence" rather than rollindex, and just return element x when called by `r = genobject:sequence(x)`

The already created auto-increment behaviour could be preserved for when :sequence() is called without a parameter.

The auto-wipe and refill behaviour (when it gets to end of auto-incrementing or gets called with new parameter (?!)) seems like a gotcha-trap to me, that would need a warning and explaination in help. It is a critical operation to rewrite the sequence that deserves a clear function of its own, like: `genobj:shuffleSequence()` --its intuitive that command would reset the auto-return position and maintain the sequences size, and if given a parameter, it would change it to a new size as `genobj:shuffleSequence(5)`

One shuffleseq per generation object seems more convenient than involving an extra parameter to have multiple. I dont think performance overhead of having separate generation objects for separate shufflers would be significant enough to warrant the complications of managing multiple shufflers within same object. But if it were desired is should be a second optional parameter. 

When creating an object which will provide Random functions, an optional seed is the obvious parameter which is relevant to all its functions. The length of a shuffle sequence is not relevant to the Distribution returning functions, so is most understandably provided whne initialising the shuffle sequence rather than the 'genobject.

Initializing a shuffle sequence and resetting it have practically the same outcome so it could be called the same thing, eg:`genobj:shuffleSequence(5)`

I appreciate your patience, ive been terribly sloppy here. If Ive not managed to be clear and helpful this time -my bad, pay me no more heed. Cheers.

cppcooper:
strainer, perhaps you'd care to chat with me about some of these proposed changes. No private messaging on github, and you used a dummy email in your commits ;D - cooper.s.josh@gmail.com

strainer:
Thankyou cppcooper, I must fix my local git config. My email is andrewinput@gmail.com if you need anything. Good health and luck'

cppcooper:
Implemented all the stuff we talked about.

The latest commit message has the closest thing to documentation as there will be from me however. It is quite straight forward, and I'd be happy to explain it directly to anything who needs that.

Commit message follows:
```
Major Revision
update v2.0
=-=-=-=-=
Native functions(exported to lua):
-GenerateEngine:  returns engine id              (args: seed)
-DestroyEngine:   destroys corresponding engine  (args: rngID)
-NewSeed          re-seeds engine                (args: rngID, seed)
-rollInt          generates random integer       (args: rngID, min, max)
-rollDouble       generates random double        (args: rngID, min, max)
-rollNormal       generates random normal[gaus.] (args: rngID, avg, stddev)
-rollBool         generates random boolean       (args: rngID, chance)
-MakeNumSequence  returns sequence id            (args: start, end)
-AddToSequence    adds a number to the sequence  (args: seqID, num)
-ShuffleSequence  shuffles the number sequence   (args: rngID, seqID)
-NextInSequence   returns the next number in seq.(args: seqID)

Lua plugin functions:
-MakeNewEngine    returns engine id              (args: seed)

Lua plugin classes:
-crng
    methods:
        -init(id, df, dist)     :: constructor
            id                           - Reference ID of engine to use in RNGenerations
            df (optional)                - bool indicating whether to destroy the Engine when the crng object is garbage collected
            dist (optional)              - lua number distribution to use
        -__gc()                 :: destructor
        -changeSeed(seed)       :: alters engine's seed value
        -setNumDistrib(distrib) :: set's the number distribution crng object should use
            distrib                      - number distribution object to use in RNGenerations
        -next()                 :: returns the next number in the distribution
        -shuffle()              :: effectively shuffles the number distribution
-normal_distribution
    methods:
        -init(avg, stddev)      :: constructor
        -next(id)               :: returns next number in the distribution
            id                           - engine ID to pass to native function
-real_distribution
    methods:
        -init(min, max)         :: constructor
        -next(id)               :: returns next number in the distribution
            id                           - engine ID to pass to native function
-int_distribution
    methods:
        -init(min, max)         :: constructor
        -next(id)               :: returns next number in the distribution
            id                           - engine ID to pass to native function
-bool_distribution
    methods:
        -init(min, max)         :: constructor
        -next(id)               :: returns next boolean in the distribution
            id                           - engine ID to pass to native function
-num_sequence
    methods:
        -init(a, b)             :: constructor
        -add(num)               :: adds num to the end of the number sequence
        -shuffle()              :: shuffles the sequence of numbers
        -next()                 :: returns next number in the sequence
```

**Note 1:** lua classes no longer use `init(...)` as the `defclass()` system of class creation doesn't allow for regular constructor syntax ie. `ctor(arg1, arg2, arg3, ...)` but instead relies on a table of arguments. 

**Note 2:** The exported functions list (ie. native functions) excludes a single function `DebugSequence(seqID)` which attempts to print all the numbers in the distribution, in the order they're listed, to the dfhack console.

PR #1205, opened 2017-12-24
add more world mode options to world info

JapaMala:
This was breaking things before that relied on dwarf mode being set, and the fortress was unretired.


lethosor:
Isn't there also a "dwarf arena" mode?

JapaMala:
There is, though I didn't add it in because it's very different in usage from regular dwarf mode, so probably needs its own enum item, much like legends mode.

lethosor:
Ok. I was thinking you wanted this for RemoteFortressReader, but that's a good point.

PR #1214, opened 2018-01-12
Remote control

JapaMala:
RFR changes, adding adventure mode features, plus a few other misc.


lethosor:
Travis is failing because it can't fetch https://github.com/dfhack/df-structures/commit/618db6349ece5827d371a60b0fac1066b1d7989b, because that commit is only in your fork (it's not in any branches in dfhack/df-structures, although GitHub still manages to display it).
Also, you should probably just add that offset to symbols.xml, not df.globals.xml, like other code offsets (start_dwarf_count and the ones warmist added in #1181). Example: https://github.com/warmist/dfhack/blob/d6df9cd257ddada6394cb1e08af7cbe0416adef0/plugins/twbt-utils.cpp#L112

JapaMala:
I wasn't actually sure what was required for loading an address from symbols.xml, and make sure PR after adding instructions for finding the address. (it's actually pretty straightforward, just search for references to "art_image-", and back trace a function or two.

plugins/remotefortressreader/item_reader.cpp
```diff
@@ -402,7 +395,7 @@ void CopyItem(RemoteFortressReader::Item * NetItem, df::item * DfItem)
             case df::enums::improvement_type::ART_IMAGE:
             {
                 VIRTUAL_CAST_VAR(artImage, df::itemimprovement_art_imagest, improvement);
-                CopyImage(artImage->image, netImp->mutable_image());
+                CopyImage(artImage->getImage(DfItem), netImp->mutable_image());
```
lethosor: Does this mean you don't need the offset because there was a vmethod already that worked, or am I misinterpreting this?

JapaMala: The offsets I found were actually just for generating new, empty, art images, and allocating them, so they don't actually work for what I need.

The Vmethod only works for decorations with images on them.

Statues don't have such a handy vmethod (though I did find one that just returns the image reference) 
In DF, the method for statues is hidden deep inside the description generating function, so what I'll just do for that is keep pointer to the art_image-*.dat loading function in df-structures, bonus that it's easy to find, and use that when needed, and just ignore statues where the image isn't already generated yet.

lethosor:
@JapaMala good to merge?

JapaMala:
Not quite yet. Few things left to do that should be finished in the next day or so.

JapaMala:
Should be good now.

PR #1218, opened 2018-01-28
More cases for Gui::getSelectedUnit

danamlund:
add new cases for Gui::getSelectedUnit: (these cases are mostly from gui/unit-info-viewer.lua)
- report list, 
- combat log list, 
- military screen, 
- unit health, 
- unit custumize, 
- assigning to cage, 
- viewing cage, 
- pitting, 
- penning, 
- burrows, 
- look at corpse, 
- look at named corpse piece, 
- look at named spatter.

This is a better implementation of a closed pull request https://github.com/DFHack/scripts/pull/44 that made the same change in lua for specific scripts.

### DIFF
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -1065,7 +1065,7 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
             {
                 // loo(k) at blood/ichor/.. spatter with a name
                 MaterialInfo mat;
-                if (mat.decode(item->spatter_mat_type, item->spatter_mat_index))
+                if (mat.decode(item->spatter_mat_type, item->spatter_mat_index) && mat.figure)
                     return df::unit::find(mat.figure->unit_id);
             }
         }
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -49,16 +49,21 @@ using namespace DFHack;
 
 #include "df/announcement_flags.h"
 #include "df/assign_trade_status.h"
+#include "df/building_cagest.h"
 #include "df/building_civzonest.h"
 #include "df/building_furnacest.h"
 #include "df/building_trapst.h"
+#include "df/building_type.h"
 #include "df/building_workshopst.h"
 #include "df/d_init.h"
 #include "df/game_mode.h"
 #include "df/general_ref.h"
 #include "df/global_objects.h"
 #include "df/graphic.h"
+#include "df/historical_figure.h"
 #include "df/interfacest.h"
+#include "df/item_corpsepiecest.h"
+#include "df/item_corpsest.h"
 #include "df/job.h"
 #include "df/layer_object_listst.h"
 #include "df/occupation.h"
@@ -74,8 +79,10 @@ using namespace DFHack;
 #include "df/ui_unit_view_mode.h"
 #include "df/unit.h"
 #include "df/unit_inventory_item.h"
+#include "df/viewscreen_announcelistst.h"
 #include "df/viewscreen_assign_display_itemst.h"
 #include "df/viewscreen_buildinglistst.h"
+#include "df/viewscreen_customize_unitst.h"
 #include "df/viewscreen_dungeon_monsterstatusst.h"
 #include "df/viewscreen_dungeonmodest.h"
 #include "df/viewscreen_dwarfmodest.h"
@@ -89,6 +96,7 @@ using namespace DFHack;
 #include "df/viewscreen_layer_noblelistst.h"
 #include "df/viewscreen_layer_overall_healthst.h"
 #include "df/viewscreen_layer_stockpilest.h"
+#include "df/viewscreen_layer_unit_healthst.h"
 #include "df/viewscreen_layer_unit_relationshipst.h"
 #include "df/viewscreen_locationsst.h"
 #include "df/viewscreen_petst.h"
@@ -97,6 +105,7 @@ using namespace DFHack;
 #include "df/viewscreen_tradegoodsst.h"
 #include "df/viewscreen_unitlistst.h"
 #include "df/viewscreen_unitst.h"
+#include "df/viewscreen_reportlistst.h"
 #include "df/viewscreen_workquota_conditionst.h"
 #include "df/viewscreen_workshop_profilest.h"
 #include "df/world.h"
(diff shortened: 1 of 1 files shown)


library/modules/Gui.cpp
```diff
@@ -934,12 +946,63 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
         return NULL;
     }
 
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_reportlistst, top))
+        return vector_get(screen->units, screen->cursor);
+
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_announcelistst, top))
+    {
+        auto *report = vector_get(screen->reports, screen->sel_idx);
+        for (df::unit *unit : world->units.all)
```
danamlund: This loop could take a while. I hope we aren't running this method multiple times every frame. I did not notice any delay when this code is executed.

library/modules/Gui.cpp
```diff
@@ -948,16 +1011,66 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
...
+                    auto *corpsepiece = static_cast<df::item_corpsepiecest*>(item->item);
+                    return df::unit::find(corpsepiece->unit_id);
+                }
+            }
+            else if (item->type == df::ui_look_list::T_items::Spatter)
+            {
+                // loo(k) at blood/ichor/.. spatter with a name
+                MaterialInfo mat;
+                if (mat.decode(item->spatter_mat_type, item->spatter_mat_index) && mat.figure)
+                    return df::unit::find(mat.figure->unit_id);
```
danamlund: You mentioned we shouldn't implement the find unit in blood spatter. I assume you meant not to implement find unit for a general race spatter. This spatter lookup will always find a single unit.

library/modules/Gui.cpp
```diff
@@ -948,16 +1011,66 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
...
+    {
+        if (ui->burrows.in_add_units_mode)
+            return vector_get(ui->burrows.list_units, ui->burrows.unit_cursor_pos);
+
+        return NULL;
+    }
+    case QueryBuilding:
+    {
+        df::building *building = getAnyBuilding(top);
+        if (building->getType() == df::building_type::Cage)
```
lethosor: More DFHack-idiomatic:
```c++
if (VIRTUAL_CAST_VAR(cage, df::building_cagest, building))
{
  // do stuff with cage, which is non-null
}
```
it's equivalent to
```c++
if (auto cage = virtual_cast<df::building_cagest>(building)) {...}
```

library/modules/Gui.cpp
```diff
@@ -948,16 +1011,66 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
...
+        {
+            auto *cage = static_cast<df::building_cagest*>(building);
+            if (*ui_building_item_cursor < cage->assigned_units.size())
```
lethosor: Could check for this global being non-null somewhere (not sure if I missed it earlier)

library/modules/Gui.cpp
```diff
@@ -948,16 +1011,66 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
...
-        if (item && item->type == df::ui_look_list::T_items::Unit)
-            return item->unit;
-        else
-            return NULL;
+        if (item) {
+            if (item->type == df::ui_look_list::T_items::Unit)
+                return item->unit;
+            else if (item->type == df::ui_look_list::T_items::Item)
+            {
+                if (item->item->getType() == df::item_type::CORPSE)
```
lethosor: Same thing with VIRTUAL_CAST_VAR (or virtual_cast) here and the next `else if` - it'll protect against null pointers/mismatched types.

library/modules/Gui.cpp
lethosor: It seems a bit weird to me for getSelectedUnit() to return a unit when no unit is selected, but it'll do for now.

library/modules/Gui.cpp
```diff
@@ -934,12 +946,63 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
         return NULL;
     }
 
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_reportlistst, top))
+        return vector_get(screen->units, screen->cursor);
+
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_announcelistst, top))
+    {
+        auto *report = vector_get(screen->reports, screen->sel_idx);
```
lethosor: Should check for null. I don't think that should ever happen with this screen, but it could with raid reports, or it might change in the future. Also, `vector_get` is designed to return null instead of crashing (although admittedly that covers bad indices, nothing specific to vectors containing null pointers), so you might as well check for it.

library/modules/Gui.cpp
```diff
@@ -934,12 +946,63 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
         return NULL;
     }
 
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_reportlistst, top))
+        return vector_get(screen->units, screen->cursor);
+
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_announcelistst, top))
```
lethosor: This is also the main `a` screen, right? It looks like what you have handles `r`+Enter, which is good, but might not behave as expected on the `a` screen (disregard if you've tested that).

library/modules/Gui.cpp
lethosor: Hopefully. I'm not aware of a way to make it faster, unless unit IDs are recorded in reports somehow.

library/modules/Gui.cpp
danamlund: I had not thought of that. But I could not find a way to fetch unit from a screen reports, so i left that case blank.

danamlund:
Excellent finds. Added checks and performed code-cleanup based on feedback.

library/modules/Gui.cpp
lethosor: Ok, looks good. There was some discussion about that in the DFHack thread and I don't believe there is a way that we know of.

library/modules/Gui.cpp
```diff
@@ -934,30 +946,138 @@ df::unit *Gui::getAnyUnit(df::viewscreen *top)
...
+            // in (a)nnouncements
+            return NULL; // cannot determine unit from reports
+        }
+    }
+
+    if (VIRTUAL_CAST_VAR(screen, df::viewscreen_layer_militaryst, top))
+    {
+        if (screen->page == df::viewscreen_layer_militaryst::T_page::Positions) {
+            auto positions = getLayerList(screen, 1);
+            if (positions && positions->enabled && positions->active)
```
lethosor: I think the lists should always exist, even if they're empty, but can't hurt to check.

PR #1219, opened 2018-01-28
FPS counter that does not need to stabilize after unpausing

danamlund:
This FPS counter only counts frames when the game is not paused.

This commit enables it in dwarfmonitor by default. But it is only shown if [FPS:YES] is set in init.txt.
Its positioned on top of the default FPS counter (when its pushed by the "*PAUSED*" text).

It does not support changing FPS:YES at runtime.

### DIFF
--- a/dfhack.init-example
+++ b/dfhack.init-example
@@ -202,6 +202,9 @@ tweak hotkey-clear
 # Allows lowercase letters in embark profile names, and allows exiting the name prompt without saving
 tweak embark-profile-name
 
+# dwarf mode FPS counter that does not jump to FPS_CAP when paused
+tweak pausing-fps-counter
+
 # Misc. UI tweaks
 tweak block-labors              # Prevents labors that can't be used from being toggled
 tweak burrow-name-cancel
--- a/plugins/tweak/tweak.cpp
+++ b/plugins/tweak/tweak.cpp
@@ -98,6 +98,7 @@
 #include "tweaks/kitchen-prefs-empty.h"
 #include "tweaks/max-wheelbarrow.h"
 #include "tweaks/military-assign.h"
+#include "tweaks/pausing-fps-counter.h"
 #include "tweaks/nestbox-color.h"
 #include "tweaks/shift-8-scroll.h"
 #include "tweaks/stable-cursor.h"
@@ -232,6 +233,9 @@ DFhackCExport command_result plugin_init (color_ostream &out, std::vector <Plugi
         "    Preserve list order and cursor position when assigning to squad,\n"
         "    i.e. stop the rightmost list of the Positions page of the military\n"
         "    screen from constantly jumping to the top.\n"
+        "  tweak pausing-fps-counter [disable]\n"
+        "    Replace fortress mode FPS counter with one that stops counting \n"
+        "    when paused.\n"
         "  tweak shift-8-scroll [disable]\n"
         "    Gives Shift+8 (or *) priority when scrolling menus, instead of \n"
         "    scrolling the map\n"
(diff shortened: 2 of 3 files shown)


plugins/lua/dwarfmonitor.lua
```diff
@@ -146,6 +146,95 @@ function Widget_cursor:render_body(p)
...
+    df.global.gps.display_frames = 0
+
+    if not self.opts.average_over_seconds then
+        self.opts.average_over_seconds = 5
+    end
+
+    if df.global.pause_state then
+        self.prev_paused = true
+    else
+        local clock = os.clock()
```
lethosor: This is not portable - it returns the elapsed CPU time on Linux and macOS, not the absolute time, so I'm seeing stuff like "FPS: 342" and "FPS: 887" with a limit of 100.

lethosor: This is not portable - it returns the elapsed CPU time on Linux and macOS, not the absolute time, so I'm seeing stuff like "FPS: 342" and "FPS: 887" with a limit of 100.

danamlund: The counter is updated every second, and the value is the average of the latest five 1-second intervals of FPS. Without averaging, the FPS counter would jump unpredictable every time it was reset. I will fix it so the counter is always visible (but shows as "FPS:   -" the first second).

The counter seemed to work on my linux machine so I did not think to thoroughly understand os.clock(). 
To fix it I intend to add a variant of MiscUtils.cpp:getTimeMs64() to LuaApi.cpp:dfhack_module because lua does not provide millisecond precision timing by default. That way I can call dfhack.getTime() from lua that returns seconds since epoch with millisecond floating-point precision (similar to os.clock()).
Since lua stores numbers as 64 bit doubles, a 64 bit unsigned int will lose some precision in the year 200 million. I can live with this limitation.

And thank you very much for all your informative feedback as I get know the code base.

lethosor: Oh, was it working on Linux? I was guessing you were testing on Windows because os.clock returns absolute time there, but I guess it would effectively return absolute time on Linux if you're hitting the FPS cap.

Your idea for wrapping getTimeMs64 is good. Lua 5.3 supports 64-bit integers, and we had to switch to that for 0.43.05, so it won't lose precision until it wraps around.

danamlund:
I have added the new method dfhack.getTimeMs(). And used that method instead of os.clock() to get accurate timing.
The counter is now algo accurate when the fps is limited by maxfps.
I also now show "FPS: ..." for the first second before enough information have been gathered to show the real FPS.

danamlund:
I just realized that this counter removes the graphical fps. 
And more general, I dont like that this counter replaces an existing df feature by default. 
I will rework it to show the new fps count number in addition to the old fps counters.

danamlund:
Turns out i did not need a new time method, dfhack.global.enabler.clock is the number of milliseconds df have been running.
This FPS counter now also show graphical fps, so its a complete replacement for the default fps counter.

I decided to show FPS as average over the last 3 seconds. Showing over 1 second made it jump between 40 and 70 on my medium fort. 3 seconds show a stable number that still change quickly.

The new FPS counter is only used in dwarf mode, in the main menu and adventurere the default fps counter is used.

I am thinking this FPS counter fits better as a tweak rather than a dwarfmonitor setting, but then you would not be able to change how many seconds you want average over (like we can currenty in dfhack-config/dwarfmonitor.json). What do you think?

When i play, FPS is an important resource, so I think a more responsive FPS counter is useful enough to enable by default (assuming [FPS:YES]). If this counter should not be default, I will undo the changes to dfhack-config/dwarfmonitor.json.

danamlund:
Rewrote new FPS counter as a tweak. Rebased commits to hide away the dwarfmonitor code.

This FPS counter look just like the default one except it does not increase when paused and it shows average over the last 3 seconds. 
The new FPS counter is only for fortress mode, and only when [FPS:YES] is set in init.txt.

The one bug I cannot figure out how to fix is that disabling fps counter through gui/settings-manager.lua will not work with this new FPS counter. You have to remove [FPS:YES] and restart dwarf fortress.

lethosor:
Looks good to me. This seems like more of a personal preference sort of tweak to me, while most of the tweaks in dfhack.init-example are bugfixes and small UI additions, so I don't think it should be in there for now. I'll make sure people know about it, though, especially pack maintainers, so we'll see what happens.

Anyway, thanks for your recent contributions! (and sorry for being slow to merge things sometimes.)

danamlund:
Sounds good

PR #1223, opened 2018-02-05
Remote control

JapaMala:
For next version.


lethosor:
@JapaMala spotted a couple more issues and fixed them here - you might want to pull those.

PR #1224, opened 2018-02-07
Fix Error.cpp compile for Windows

BenLubar:


ab9rf:
Patch breaks dfhack-client:
```
9>     Creating library C:/Users/Kelly/Documents/GitHub/dfhack/build/win64/VC2015/library/RelWithDebInfo/dfhack-client.lib and object C:/Users/Kelly/Documents/GitHub/dfhack/build/win64/VC2015/library/RelWithDebInfo/dfhack-client.exp
9>MiscUtils.obj : error LNK2019: unresolved external symbol "public: __cdecl DFHack::Error::NullPointer::NullPointer(char const *)" (??0NullPointer@Error@DFHack@@QEAA@PEBD@Z) referenced in function "public: void __cdecl DFHack::Error::NullPointer::`default constructor closure'(void)" (??_FNullPointer@Error@DFHack@@QEAAXXZ)
9>MiscUtils.obj : error LNK2019: unresolved external symbol "public: __cdecl DFHack::Error::InvalidArgument::InvalidArgument(char const *)" (??0InvalidArgument@Error@DFHack@@QEAA@PEBD@Z) referenced in function "public: void __cdecl DFHack::Error::InvalidArgument::`default constructor closure'(void)" (??_FInvalidArgument@Error@DFHack@@QEAAXXZ)
9>MiscUtils.obj : error LNK2019: unresolved external symbol "public: __cdecl DFHack::Error::VTableMissing::VTableMissing(char const *)" (??0VTableMissing@Error@DFHack@@QEAA@PEBD@Z) referenced in function "public: void __cdecl DFHack::Error::VTableMissing::`default constructor closure'(void)" (??_FVTableMissing@Error@DFHack@@QEAAXXZ)
```

lethosor:
This patch doesn't change the signatures of anything. Have you tried building again?

ab9rf:
Ben changed some of the types from const char* to std::string, and that breaks the default constructor.

lethosor:
He only changed safe_str, and stl_concat was returning a string before.

BenLubar:
Is safe_str used from somewhere else?

lethosor:
No, it's local to that file. But the errors ab9rf is getting are with the exception constructors, whose types haven't changed in ages.

ab9rf:
I completely blew away the entire VC2015 tree in my work directory and rebuilt entirely from scratch. Same error.

Here's the kicker: commenting out `#include "Error.h` in MiscUtils.cpp eliminates the error. I have no meaningful explanation.

ab9rf:
With the makefile change in Ben's second commit, it now compiles for me.

I don't pretend to understand what is going on here, but my lack of comprehension is no reason to block the PR.

lethosor:
Ok, so for some reason the MSVC linker was deciding to check for symbols that weren't actually used. Anyway, the include in MiscUtils.cpp should probably be removed if it's not needed; I forgot to check there.

ab9rf:
MSVC is pulling the symbols in to make the class constructible, which become necessary because of Error.h.  Error.cpp doesn't have this problem because the required constructors are in Error.cpp itself.  Basically, MiscUtils.cpp (which dfhack-client depends on) depends on Error.cpp, but dfhack-client doesn't know that.

I don't understand why this doesn't happen under gcc, but gcc plays a lot looser with the language standards than MSVC does.

lethosor:
But none of those classes are actually constructed or used in MiscUtils.cpp in any way. The only things that could construct exception instances are the CHECK_NULL_POINTER and CHECK_INVALID_ARGUMENT macros, and the preprocessor should strip those out because they aren't used. That also doesn't explain the VTableMissing error, or why none of the classes derived from AllSymbols have similar issues.

Anyway, MSVC is weird, and we don't need the Error.h include anyway, and might not even need dfhack-client. I'll open a new issue for that.

PR #1231, opened 2018-02-25
Corrected reanimation detection

PatrikLundell:
Corrected reanimation detection

I've tried the modified code in a DF installation, and it does seem to work as intended.

### DIFF
--- a/build/win64/install-debug.bat
+++ b/build/win64/install-debug.bat
@@ -1,4 +1,4 @@
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+call "D:\Program (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
 cd VC2015
 msbuild /m /p:Platform=x64 /p:Configuration=RelWithDebInfo INSTALL.vcxproj
 cd ..
--- a/build/win64/install-debug.bat
+++ b/build/win64/install-debug.bat
@@ -1,4 +1,4 @@
-call "D:\Program (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
 cd VC2015
 msbuild /m /p:Platform=x64 /p:Configuration=RelWithDebInfo INSTALL.vcxproj
 cd ..
--- a/plugins/embark-assistant/defs.h
+++ b/plugins/embark-assistant/defs.h
@@ -33,6 +33,8 @@ namespace embark_assist {
             int16_t elevation;
             bool river_present = false;
             int16_t river_elevation = 100;
+            int8_t adamantine_level;  // -1 = none, 0 .. 3 = cavern 1 .. magma sea. Currently not used beyond present/absent.
+            int8_t magma_level;  // -1 = none, 0 .. 3 = cavern 3 .. surface/volcano
             int8_t biome_offset;
             uint8_t savagery_level;  // 0 - 2
             uint8_t evilness_level;  // 0 - 2
@@ -42,7 +44,6 @@ namespace embark_assist {
         };
 
         typedef std::array<std::array<mid_level_tile, 16>, 16> mid_level_tiles;
-//        typedef mid_level_tile mid_level_tiles[16][16];
 
         struct region_tile_datum {
             bool surveyed = false;
@@ -53,7 +54,6 @@ namespace embark_assist {
             uint8_t min_region_soil = 10;
             uint8_t max_region_soil = 0;
             bool waterfall = false;
-
             river_sizes river_size;
             int16_t biome_index[10];  // Indexed through biome_offset; -1 = null, Index of region, [0] not used
             int16_t biome[10];        // Indexed through biome_offset; -1 = null, df::biome_type, [0] not used
@@ -162,6 +162,22 @@ namespace embark_assist {
             Major
         };
 
+//        enum class adamantine_ranges : int8_t {
+//            NA = -1,
+//            Cavern_1,
+//            Cavern_2,
+//            Cavern_3,
+//            Magma_Sea
+//        };
+
+        enum class magma_ranges : int8_t {
+            NA = -1,
+            Cavern_3,
+            Cavern_2,
+            Cavern_1,
+            Volcano
+        };
+
         enum class yes_no_ranges : int8_t {
             NA = -1,
             Yes,
@@ -218,6 +234,10 @@ namespace embark_assist {
             yes_no_ranges evil_weather;  //  Will probably blow up with the magic release arcs...
             yes_no_ranges reanimation;
             yes_no_ranges thralling;
+            int8_t spire_count_min; // N/A(-1), 0-9
+            int8_t spire_count_max; // N/A(-1), 0-9
+            magma_ranges magma_min;
+            magma_ranges magma_max;
             int8_t biome_count_min; // N/A(-1), 1-9
             int8_t biome_count_max; // N/A(-1), 1-9
             int8_t region_type_1;   // N/A(-1), df::world_region_type
--- a/plugins/embark-assistant/defs.h
+++ b/plugins/embark-assistant/defs.h
@@ -162,6 +162,7 @@ namespace embark_assist {
             Major
         };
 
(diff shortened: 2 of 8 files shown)


plugins/embark-assistant/defs.h
```diff
@@ -162,6 +162,22 @@ namespace embark_assist {
             Major
         };
 
+//        enum class adamantine_ranges : int8_t {
```
lethosor: Are you planning on using this?

plugins/embark-assistant/finder_ui.cpp
```diff
@@ -1041,15 +1267,19 @@ namespace embark_assist {
...
-            embark_assist::screen::paintString(white_pen, 4, 1, ":Shift list");
-            embark_assist::screen::paintString(lr_pen, 16, 1, "8/2");
-            embark_assist::screen::paintString(white_pen, 19, 1, ":Up/down");
-            embark_assist::screen::paintString(lr_pen, 28, 1, "ENTER");
-            embark_assist::screen::paintString(white_pen, 33, 1, ":Select item");
-            embark_assist::screen::paintString(lr_pen, 46, 1, "f");
-            embark_assist::screen::paintString(white_pen, 47, 1, ":Find");
-            embark_assist::screen::paintString(lr_pen, 53, 1, "ESC");
-            embark_assist::screen::paintString(white_pen, 56, 1, ":Abort");
+            embark_assist::screen::paintString(white_pen, 4, 1, ":<->");
```
lethosor: Could use `"\x1b\x1a"` for actual arrows - see http://dwarffortresswiki.org/index.php/CP437

plugins/embark-assistant/help_ui.cpp
```diff
@@ -250,13 +253,18 @@ namespace embark_assist{
...
+                help_text.push_back("  that have been removed at world gen to fail to be generated. It's likely");
+                help_text.push_back("  this bug also affects magma pools. This plugin does not address this.");
+                help_text.push_back("Version 0.3 2018-02-26");
 
                 break;
             }
 
             //  Add control keys to first line.
-            embark_assist::screen::paintString(pen_lr, 1, 1, "TAB/Shift-TAB");
+            embark_assist::screen::paintString(pen_lr, 1, 1, "TAB");
```
lethosor: I didn't notice this before, but you should be passing interface_key enum items into Screen::getKeyDisplay somehow instead of hardcoding the human-readable form here. There are examples of OutputString and related functions in plugins/uicommon.h. Doing that would avoid confusion if people have changed their keybindings, and would also be more consistent (DF uses `Tab`, `Enter`, `Esc`, etc.).

lethosor:
Looks good overall (I commented on the few things I noticed). Out of curiosity, is [this](http://www.bay12games.com/dwarves/mantisbt/view.php?id=10267) the bug you're referring to? It would be good to mention it by number somewhere to avoid later confusion by people who aren't sure if it's fixed or not from the description alone.

plugins/embark-assistant/finder_ui.cpp
```diff
@@ -419,6 +553,56 @@ namespace embark_assist {
...
+                {
+                    embark_assist::defs::magma_ranges k = embark_assist::defs::magma_ranges::NA;
+                    while (true) {
+                        switch (k) {
+                        case embark_assist::defs::magma_ranges::NA:
+                            element->list.push_back({ "N/A", static_cast<int8_t>(k) });
+                            break;
+
+                        case embark_assist::defs::magma_ranges::Cavern_3:
+                            element->list.push_back({ "3:rd Cavern", static_cast<int8_t>(k) });
```
lethosor: Unsure if these correspond to `1`, `2`, `3` keys - if they don't, I would get rid of the colons.

build/win64/install-debug.bat
```diff
@@ -1,4 +1,4 @@
-call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+call "D:\Program (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
```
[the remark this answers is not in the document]
PatrikLundell: Obviously, "git commit -a" doesn't work as I thought. I manually did "git add" on the files that should be pushed... Will changing back, committing (and changing back so it actually works for me) work, or do I have to use yet another arcane git command to undo the damage?

plugins/embark-assistant/finder_ui.cpp
```diff
@@ -19,6 +21,8 @@
 
 using df::global::world;
 
+#define profile_file_name ".\\data\\init\\embark_assistant_profile.txt"
```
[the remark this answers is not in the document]
PatrikLundell: Willl do.

plugins/embark-assistant/finder_ui.cpp
PatrikLundell: I'll try that.

plugins/embark-assistant/help_ui.cpp
PatrikLundell: I would have if I'd known how to get at it. I do in Lua scripts. I'll look at the reference.

plugins/embark-assistant/finder_ui.cpp
PatrikLundell: That't the bug report. I'll add a reference to it.

I have trouble understanding the comment. There are lots of colons in there, and no explicit key press references. However after reading it several times my guess is that you refer to 1:st, 2:nd, and 3:rd? Do you think it would read better as First, Second, and Third?

plugins/embark-assistant/defs.h
PatrikLundell: I have no concrete plans for it. However, it matches the data actually collected for spires, and so I thought it's a good idea to prepare for it while updating the section.

plugins/embark-assistant/finder_ui.cpp
PatrikLundell: Good suggestion.  I also tried up/down arrows, but Phoebus uses a spruce like thingie instead of the up arrow, so I went back there.

PatrikLundell:
I believe I've addressed all review comments. If I've failed to do so, please help me by pointing out what I've missed.

lethosor:
Sorry, didn't have time to look at it when I got the notification. Will try tomorrow.

PatrikLundell:
It's just that it's a bit worrying when there is only silence and the next release window looks like it's about to close. Is requesting a review (thanks BenLubar!) the expected way to go about it?

lethosor:
That just pinged me again. It doesn't show up anywhere more obvious than a normal PR comment once I've seen the notification.

build/win64/install-debug.bat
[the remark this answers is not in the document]
lethosor: "git commit -a" is equivalent to "git add" on *every file* that has ever been added, followed by "git commit".
It looks like whatever you did worked.

plugins/embark-assistant/finder_ui.cpp
```diff
@@ -419,6 +575,56 @@ namespace embark_assist {
...
+                        case embark_assist::defs::magma_ranges::Cavern_3:
+                            element->list.push_back({ "Third Cavern", static_cast<int8_t>(k) });
+                            break;
+
+                        case embark_assist::defs::magma_ranges::Cavern_2:
+                            element->list.push_back({ "Second Cavern", static_cast<int8_t>(k) });
+                            break;
+
+                        case embark_assist::defs::magma_ranges::Cavern_1:
+                            element->list.push_back({ "Third Cavern", static_cast<int8_t>(k) });
```
lethosor: First, not Third

plugins/embark-assistant/finder_ui.cpp
lethosor: Yeah, I meant "3:rd", preferring "3rd" or "Third".

lethosor:
If you make a version of paintString that takes an interface_key (what we do in uicommon.h), that'll probably be easier in the long run.

PatrikLundell:
Thanks. I know nothing about sphinx, and don't have it installed if it wasn't installed as part of the standard non documentation installation. I'll try to look into it, though.

paintString: True. It's only a few places. Oh, only a few more. Oh, only a few more...
When/if there's an update I'll add that function. If more search options are added I'll have to implement scrolling of the options list, as I think the minimum screen height shouldn't be pushed further.

PR #1232, opened 2018-03-05
Remote reader

JapaMala:
More RFR changes.


plugins/proto/RemoteFortressReader.proto
```diff
@@ -1046,6 +1046,7 @@ enum FlowType
     OceanWave = 11;
     SeaFoam = 12;
     ItemCloud = 13;
+    CampFire = -1;
```
quietust: What's this about? Campfires aren't flows, but are distinct objects stored in world.campfires[] (along with vermin, fires, constructions, and various other special things)

JapaMala: Correct, they are not.

It's actually a fake flow, hence the negative number, for Armok Vision, so the same particle systems Dan handle it easily. 

Fires end up being treated the same way, because even though there's a fire flow type, the most commonly encountered fire is a tile type.

quietust: The "Fire" flow is for fire *breath* (e.g. from fire imps), which is probably why you don't see it for actual fires.

Issue #1240, opened 2018-03-18
AutoLaborManager crashes if Dig designation is placed above unallocated map_block


espresso4:
Every time I use digtype on gold it crashes either immediately or after I use it again on another gold nugget. It only happens with gold.

espresso4:
Actually, it crashed when I tried it on tetrahedrite. It crashes each time I try to use it on that ore.

Odd, I was able to use it on various iron ores and platinum without issue.

quietust:
What version of DFHack+DF, and on what platform+architecture?

espresso4:
DF 44.07 and DFHack 44.07-Alpha1

No other mods

Windows 10 64 bit

quietust:
I've just tested digtype with DFHack 0.44.07-alpha1 with the 64-bit Windows build and was unable to observe any crashes. However, I **did** observe it placing Dig designations on top of raw adamantine, semi-molten rock, and _slade_ where mineral inclusions likely would have appeared (had they not been overwritten by deep special tubes, the magma core, and the underworld).

espresso4:
Hmmm, I can replicate this issue quite easily.

Maybe I should add one last detail I thought was not important.

I binded the “digtype” command to Ctrl-Shift-Alt-V so I use those 4 buttons each time rather than typing it. This is in the dfhack init file where you edit shortcuts. I checked that no other command used this shortcut.

Can you give me some tips on what I should try to verify this as a bug? Should I try to replicate it on another computer? Try a fresh install (even though mine is just DF and DFhack). Let me know.

quietust:
You need to upload a copy of your fortress to DFFD or Dropbox (or some other file sharing site) so one of us can see what's causing it to crash - it's quite likely that the crash is related to the geology of your fortress's embark location.

espresso4:
http://dffd.bay12games.com/file.php?id=13584

Here is my save folder. I did create the world using advanced world gen. It may have something to do with that.

quietust:
An important detail you neglected to mention: the game does not crash until you **unpause** it.

The crash seems to be happening in AutoLaborManager::count_map_designations(void), on the following line:

```
                        df::coord p = bl->map_pos;
                        df::coord c(p.x, p.y, p.z - 1);
>                       if (Maps::getTileDesignation(c)->bits.hidden)
                            continue;
```
In this specific case, the coordinate "c" is (48,0,3), which is pointing at an _unallocated_ map block.

espresso4:
[...]

PR #1244, opened 2018-03-20
avoid crash in labormanager if map square on lowest level is designated+hidden (#1240)

ab9rf:


lethosor:
(added a clickable link, hope you don't mind)

PR #1251, opened 2018-04-04
fixed bug causing gui scripts to fail when sidebar is closed

ThiagoLira:
Tested with gui/teleport


library/lua/gui.lua
```diff
@@ -109,6 +109,7 @@ local function parse_inset(inset)
 end
 
 function inset_frame(rect, inset, gap)
+    if not rect then return mkdims_wh(0, 0, 0, 0) end
```
lethosor: Space at the end here. I'll take care of it.

PR #1253, opened 2018-04-07
Fix MSVC and GCC warnings in DFHack and core plugins

BenLubar:


depends/protobuf/CMakeLists.txt
```diff
@@ -191,7 +191,7 @@ LIST(APPEND LIBPROTOBUF_FULL_SRCS ${LIBPROTOBUF_LITE_SRCS})
 
 IF(CMAKE_COMPILER_IS_GNUCC)
     SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -Wno-sign-compare")
-    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-result")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-result -Wno-unused-local-typedefs -Wno-misleading-indentation")
```
lethosor: I think `-w` would be more future-proof

plugins/Plugins.cmake
```diff
@@ -73,42 +73,79 @@ MACRO(DFHACK_PLUGIN)
...
   FOREACH(pbuf ${PLUGIN_PROTOBUFS})
-    SET(PLUGIN_SOURCES ${PLUGIN_SOURCES} ${CMAKE_CURRENT_SOURCE_DIR}/proto/${pbuf}.pb.cc)
-    SET(PLUGIN_SOURCES ${PLUGIN_SOURCES} ${CMAKE_CURRENT_SOURCE_DIR}/proto/${pbuf}.proto)
-    SET(PLUGIN_PROTOCPP ${PLUGIN_PROTOCPP} ${CMAKE_CURRENT_SOURCE_DIR}/proto/${pbuf}.pb.cc)
+    LIST(APPEND PLUGIN_PROTOS ${CMAKE_CURRENT_SOURCE_DIR}/proto/${pbuf}.proto)
   ENDFOREACH()
 
-  # Tell CMake the source won't be available until build time.
-  SET_SOURCE_FILES_PROPERTIES(${PLUGIN_PROTOCPP} PROPERTIES GENERATED 1)
+  LIST(LENGTH PLUGIN_PROTOS NUM_PROTO)
```
lethosor: Is this stuff supposed to fix issues with protobuf-generated files not getting rebuilt? I'm assuming it's not warning-related. Was something wrong with what was in plugins/CMakeLists.txt? (Apparently I wrote it, although I didn't recognize it for a minute here.) I do like separating it by plugin, though.

plugins/autolabor.cpp
```diff
@@ -606,7 +606,7 @@ static void init_state()
     {
         string key = p->key();
         df::unit_labor labor = (df::unit_labor) atoi(key.substr(strlen("autolabor/labors/")).c_str());
-        if (labor >= 0 && labor <= labor_infos.size())
+        if (labor >= 0 && size_t(labor) <= labor_infos.size())
```
lethosor: Pretty sure this should be `<`, not `<=`

plugins/remotefortressreader/CMakeLists.txt
```diff
@@ -41,4 +42,4 @@ ELSE(UNIX)
     )
 ENDIF(UNIX)
 # this makes sure all the stuff is put in proper places and linked to dfhack
-DFHACK_PLUGIN(RemoteFortressReader ${PROJECT_SRCS} LINK_LIBRARIES ${PROJECT_LIBS} PROTOBUFS ${PROJECT_PROTO} )
+DFHACK_PLUGIN(RemoteFortressReader ${PROJECT_SRCS} LINK_LIBRARIES protobuf-lite ${PROJECT_LIBS} COMPILE_FLAGS_MSVC "/FI\"Export.h\"" COMPILE_FLAGS_GCC "-include Export.h -Wno-misleading-indentation" )
```
lethosor: What's with the `-include` here? Shouldn't the plugin source file(s) include that?

plugins/autohauler.cpp
```diff
@@ -624,7 +624,7 @@ static void init_state()
         df::unit_labor labor = (df::unit_labor) atoi(key.substr(strlen("autohauler/labors/")).c_str());
 
         // Ensure that the labor is defined in the existing list
-        if (labor >= 0 && labor <= labor_infos.size())
+        if (labor >= 0 && size_t(labor) <= labor_infos.size())
```
lethosor: Same </<= issue as autolabor

plugins/labormanager/labormanager.cpp
```diff
@@ -614,15 +614,15 @@ static void init_state()
     {
         string key = p->key();
         df::unit_labor labor = (df::unit_labor) atoi(key.substr(strlen("labormanager/2.0/labors/")).c_str());
-        if (labor >= 0 && labor <= labor_infos.size())
+        if (labor >= 0 && size_t(labor) <= labor_infos.size())
```
lethosor: </<=

plugins/Plugins.cmake
BenLubar: I rewrote it because everything that used it was using shared proto files in `/plugins/proto` which ended up meaning plugins that could actually utilize it to do something specific to the plugin were stuck making their own copy of the script.

PR #1265, opened 2018-05-08
this allows the blueprint plugin to be called from lua

db48x:


lethosor:
You need to change the line for blueprint in plugins/CMakeLists.txt to:
```cmake
    DFHACK_PLUGIN(blueprint blueprint.cpp LINK_LIBRARIES lua)
```
for this to be portable. I'm surprised it works on Linux (Travis didn't fail), but it doesn't work on OS X, and I'm not sure about Windows.

PR #1268, opened 2018-05-11
Fix unlikely GUI::getAnyUnit segfault

danamlund:
It seems unlikely we will end up on an `df::viewscreen_announcelistst` without a `world`. But we perform the same check elsewhere, and segfaults are bad.

I also verified that most cases in Gui::getAnyUnit still work in 44.10 (could not figure out what `df::viewscreen_dungeon_monsterstatusst` is).


lethosor:
viewscreen_dungeon_monsterstatusst is something in adventure mode - it looks simple enough that I'm not concerned about it being an issue.
I realize that I wasn't particularly clear when I pointed out the issues I ran into today (I was in the middle of investigating an unrelated crash). Thanks for taking the time to look through and improve this.

PR #1271, opened 2018-05-13
Simplify CMakeLists.txt for protobuf.

BenLubar:
All of the compilers we support have std::unordered_map and std::unordered_set in their standard locations.

Submitted as a PR so that I can attempt to fix the new Docker-based Windows compile setup without spamming the DFHack compile log any more than I already have.


lethosor:
Is removing /O2 something we should be concerned about (for performance stuff, although we don't care about that too much). Will release builds on buildmaster still have /O2?

PR #1278, opened 2018-05-18
enable lua api construction functions

grubsteak:
`designateRemove(pos)` cannot be invoked because it isn't registered.


lethosor:
Nice catch!

grubsteak:
whoops, wrong branch!

lethosor:
Yeah, no worries (as long as you get it right next time) - it's easy enough to merge manually.

PR #1281, opened 2018-05-20
lua: enabled shift cursor movement for fake workshop gui overlay

grubsteak:
I was trying to make a custom workshop interface and noticed you couldn't hold shift while moving the cursor because the input code wasn't passing it through to dwarf fortress.


lethosor:
The `MOVEMENT_KEYS` table in that file also has CURSOR_UP_Z_AUX and CURSOR_DOWN_Z_AUX - would it be worth it to add those keys (and do they cause issues)?

Issue #1274, opened 2018-05-17
Max quality for planned buildings


lethosor:
Suggested by JezaGaia at http://www.bay12forums.com/smf/index.php?topic=164123.msg7757035#msg7757035

> I can select the min quality of items I want used but not the max, so say I want to keep my wasterwork quality beds for my nobles and furnish the other rooms with any other quality I haven't found a way for it.

TotallyGatsby:
It seems like the ItemConstraint has a min_quality property. Would the approach here be to just add a max_quality property, and check/set it everywhere that min_quality is set in workflow.cpp?

If so, I can probably make an attempt at this.

lethosor:
This is about buildingplan, so stuff is spread between plugins/buildingplan.cpp and buildingplan-lib.{h,cpp}. In addition to what you mentioned, you'll need to display it in the sidebar and save it somehow (it looks like config.ival(1) through config.ival(3) are used, so config.ival(4) should be ok).

TotallyGatsby:
Ok, got the latest develop branch compiling. Lemme see what I can do.

TotallyGatsby:
Is there precedence for how min/max should interact while cycling? 

Example:
Min is set to 'FinelyCrafted', and I cycle max past 'Artifact'. Should it:
a) Go to 'Ordinary' (the bottom value)
b) Go to 'FinelyCrafted' (the current minimum value)

I can see either being reasonable.

lethosor:
I think DF has a quality selection in the workshop profile screen - you could look at that for ideas. I'm not sure if there's a good way to do this with just 2 keys instead of 4.

TotallyGatsby:
Tossed up a (perhaps naive?) implementation. Your comment about 4 keys makes me wonder if I'm missing a quality of life change. (Were you thinking that you'd want to cycle up and down with two keys per entry? e.g. Q - increase minimum, q - decrease minimum, W - increase maximum, w - decrease maximum)

lethosor:
If you think sticking with the two behaviors you listed is better, I think (b) is the more intuitive one, because you won't end up with combinations that exclude all items.

TotallyGatsby:
Cool. I can take a pass on the interface. Lemme know if there's anything else in the PR you think needs to be touched up.

TotallyGatsby:
The workshop manager lowers the min when the max goes lower than the min (and raises the max when the min goes above the max.) It uses four keys.  I'll match that behavior.

[...]

PR #1284, opened 2018-05-20
Add 'Max Quality' setting to buildingplan

TotallyGatsby:
Fulfills feature request #1274

Additionally, made it so one can set the minimum quality to Artifact (if it was intentional that buildingplan did not allow for placement of Artifact furniture, I can quickly fix that up.)

Only built/tested on Windows, but the scope of the changes is pretty small.

(Not sure where to drop the notice in the changelog.)

### DIFF
--- a/docs/Authors.rst
+++ b/docs/Authors.rst
@@ -131,6 +131,7 @@ Tim Walberg             twalberg
 Timothy Collett         danaris
 Tom Jobbins             TheBloke
 Tom Prince
+TotallyGatsby           TotallyGatsby
 Travis Hoppe            thoppe                  orthographic-pedant
 txtsd                   txtsd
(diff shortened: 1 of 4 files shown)


TotallyGatsby:
Derp, pointed this to master initially. Fixed.

lethosor:
I think it would be more consistent to have adjacent keys adjust the same setting in opposite directions. For example, qw for min and QW for max, instead of qQ for min and wW for max. @TotallyGatsby I'm most likely going to change this, but if there was a specific reason you chose not to do it that way, let me know.

TotallyGatsby:
I'm the newbie, you do what you think is most appropriate. Need me to make the change or is it easy enough to do as a part of the PR?

PR #1295, opened 2018-06-03
Add df-luacheck to Travis build.

BenLubar:


.travis.yml
```diff
@@ -51,6 +51,10 @@ script:
 - cd ..
 - cp travis/dfhack_travis.init "$DF_FOLDER"/
 - python travis/run-tests.py "$DF_FOLDER"
+- cd travis/df-luacheck
+- npm install
```
lethosor: Should probably be in some pre-build step if possible.

BenLubar: It requires the `codegen.out.xml` file, and I'm thinking of making a plugin that generates the `builtin.js` file.

myk002:
I think this has been superseded by @vallode 's efforts to get integrated type checking in the build

PR #1296, opened 2018-06-06
Add Lua profiler module

suokko:
Lua profiler module is pure lua profiler implementation. It support time based and call entry exit based profiling. The base profiler code from http://lua-users.org/wiki/PepperfishProfiler required some porting to lua 5.3 and dfhack specialization.

While I ported the code to dfhack I noticed that profiling hooks looked like doing a lot of unnecessary work. I decided to rewrite profiling hooks in hopes that lower profiling costs would improve profiling statistics quality.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -59,6 +59,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 - Removed Vermin module (unused and obsolete)
 
+## Lua
+- Added ``profiler`` module to measure lua performance
+
 # 0.44.10-r1
 
 ## New Scripts
--- a/library/lua/profiler.lua
+++ b/library/lua/profiler.lua
@@ -1,0 +1,616 @@
+--[[
+
+== Introduction ==
+
+  Note that this requires os.clock(), debug.sethook(),
+  and debug.getinfo() or your equivalent replacements to
+  be available if this is an embedded application.
+
+  Example usage:
+
+    profiler = newProfiler()
+    profiler:start()
+
+    < call some functions that take time >
+
+    profiler:stop()
+
+    local outfile = io.open( "profile.txt", "w+" )
+    profiler:report( outfile )
+    outfile:close()
+
+== Optionally choosing profiling method ==
+
+The rest of this comment can be ignored if you merely want a good profiler.
+
+  newProfiler(method, sampledelay):
+
+If method is omitted or "time", will profile based on real performance.
+optionally, frequency can be provided to control the number of opcodes
+per profiling tick. By default this is 100000, which (on my system) provides
+one tick approximately every 2ms and reduces system performance by about 10%.
+This can be reduced to increase accuracy at the cost of performance, or
+increased for the opposite effect.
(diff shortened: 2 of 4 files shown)


docs/changelog.txt
```diff
@@ -316,6 +316,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Structures
 - New globals: ``soul_next_id``
 
+## Lua
+- Added ``profiler`` module to measure lua performance
+
```
lethosor: Wrong section - this is under 0.44.05-r1.

It's possible that you based this on an old commit. If that's the case, it should be manageable enough to fix on our end.

library/lua/profiler.lua
```diff
@@ -0,0 +1,651 @@
...
+  self.stack = {}
+  self.depth = 0
+  if self.variant == "time" then
+    self.lastclock = os.clock()
+    debug.sethook( _profiler_hook_wrapper_by_time, "", self.sampledelay )
+  elseif self.variant == "call" then
+    debug.sethook( _profiler_hook_wrapper_by_call, "cr" )
+  else
+    print("Profiler method must be 'time' or 'call'.")
+    sys.exit(1)
```
lethosor: Not acceptable in a module (but I don't think this is defined, anyway). It should use error() instead.

library/lua/profiler.lua
```diff
@@ -0,0 +1,651 @@
...
+  if self.variant == "time" then
+    terms.capitalized = "Sample"
+    terms.single = "sample"
+    terms.pastverb = "sampled"
+  elseif self.variant == "call" then
+    terms.capitalized = "Call"
+    terms.single = "call"
+    terms.pastverb = "called"
+  else
+    assert(false)
```
lethosor: Might as well replace with `error()`, although I think the above error() should be enough.

suokko:
Clearly I don't know how to use github. I tried to make a new profile_v2 branch with commits change to fix your comments. But then I couldn't figure out how to make this pull request reflect the changes. I found a stackoverflow answer telling I need to update my own branch and changes will automatically appear to pull requests. Too bad it seems like github doesn't like force updates.

If this pull request is broken I will open a new one.

warmist:
@suokko I'm pretty sure you can force push to github. Just make sure that the branch was correct.

Also is this related/can be used for lua debugging/IDE integration?

suokko:
The profiler module doesn't help integrating debugger to dfhack. Debugger requires much more calls to the debug module of lua. Writing completely new debugger or IDE integration feels like a lot work even tough the debug module seems to provide a good and simple interface.

Easiest debugging aid tool to develop based on profiler would latrace like function call tracer.

warmist:
No, i meant this: https://studio.zerobrane.com/doc-remote-debugging

It seemed to me that we only need a functional lua-sockets?

lethosor:
@suokko generally you should continue working on the same branch you were before, which avoids the need to force push and keeps history intact. Looks like you addressed everything except the line in the license file, which we can add pretty easily.

PR #1297, opened 2018-06-07
Units dead

PatrikLundell:
See discussion in df-structures issue 247.
Link:  DFHack/df-structures#247

PR #1298, opened 2018-06-08
Dead plugins

PatrikLundell:
Addresses parts of df-structures issue 247.

### DIFF
--- a/plugins/cursecheck.cpp
+++ b/plugins/cursecheck.cpp
@@ -160,8 +160,9 @@ command_result cursecheck (color_ostream &out, vector <string> & parameters)
     {
         df::unit * unit = world->units.all[i];
 
+        // filter out all "living" units that are currently removed from play
         // don't spam all completely dead creatures if not explicitly wanted
-        if(unit->flags1.bits.dead && ignoreDead)
+        if((unit->flags1.bits.dead && !unit->flags2.bits.killed) || (unit->flags2.bits.killed && ignoreDead))
         {
             continue;
         }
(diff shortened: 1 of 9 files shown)


lethosor:
Link: DFHack/df-structures#247

(You can also just paste the full link to the issue and it'll display like that.)

plugins/cursecheck.cpp
```diff
@@ -160,8 +160,9 @@ command_result cursecheck (color_ostream &out, vector <string> & parameters)
     {
         df::unit * unit = world->units.all[i];
 
+        // filter out all "living" units that are currently removed from play
         // don't spam all completely dead creatures if not explicitly wanted
-        if(unit->flags1.bits.dead && ignoreDead)
+        if((unit->flags1.bits.dead && !unit->flags2.bits.killed) || (unit->flags2.bits.killed && ignoreDead))
```
lethosor: It would be clearer to use library functions from #1297 here once that's merged.

PatrikLundell: True, but I can't do that before then, at least not if I wan't it to compile in my end.

PatrikLundell:
As can be seen, I've switched to Units functions. I've compiled it by first copying the Units.h and Units.cpp over from the dead_units branch without adding/committing them.

PatrikLundell:
I did get one of the commit comments wrong (guess which one).
These changes should mean the files won't have to be changed when the "dead" flag is renamed, and all of the files did make use of other Units functions anyway.
Obviously, it still won't compile, though.

PR #1299, opened 2018-06-09
YCM fallback defaults if database lookup fails

suokko:
YCM flag function returning None result an error from YCM python code
everytime a file is opened. The None return can happen if database
lookup fails for any reason. To let YCM have resonable flags without
database entry the .ycm_extra_conf.py can include default fallback
flags.

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -4,11 +4,40 @@ https://github.com/Valloric/ycmd
 
 # pylint: disable=import-error,invalid-name,missing-docstring,unused-argument
 
-import os
+import os,platform
 import ycm_core
 
 def DirectoryOfThisScript():
     return os.path.dirname(os.path.abspath(__file__))
+
+default_flags = [
+    '-I','library/include',
+    '-I','library/proto',
+    '-I','plugins/proto',
+    '-I','depends/protobuf',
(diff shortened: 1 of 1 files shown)


.ycm_extra_conf.py
```diff
@@ -117,13 +146,19 @@ def FlagsForFile(filename, **kwargs):
...
+            'do_cache': True,
+        }
 
     final_flags = MakeRelativePathsInFlagsAbsolute(
         compilation_info.compiler_flags_,
         compilation_info.compiler_working_dir_
     )
 
+    # Make sure ycm reports more suspicuous code lines
+    final_flags.append('-Wextra')
```
lethosor: I don't think this flag is enabled anywhere in DFHack. How does it affect YCM?

suokko: It enables couple of clang warnings that will appear in diagnostic view. I have setup diagnostic to mark warnings and errors in left column next to source lines (by default there is vim relative numbering). The warnings that will be included are partially included in gcc -Wall (example sign compare). https://clang.llvm.org/docs/DiagnosticsReference.html#wextra

lethosor: Oh, it's using clang, not GCC. Makes sense.

PR #1300, opened 2018-06-10
Make stl_vsprintf return if there is an vsnprintf error

suokko:
vsnprintf man page claims:
"If an output error is encountered, a negative value is returned."

That means we has to call vsnprintf twice at most to have whole output
written to a string. But in case of error we return an empty string.

The code also optimizes an expected common case of outputting single
line with a small stack allocated buffer. If the stack buffer is too
small then it uses std::string::resize to allocate exactly enough memory
and writes directly to std::string.

Second call could use vsprintf because memory is known to be large
enough. But I think that difference isn't detectable outside micro
benchmarks.

### DIFF
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -27,6 +27,8 @@ distribution.
 #include "MiscUtils.h"
 
 #ifndef LINUX_BUILD
+// We don't want min and max macros
+#define NOMINMAX
     #include <Windows.h>
 #else
     #include <sys/time.h>
@@ -67,6 +69,8 @@ std::string stl_vsprintf(const char *fmt, va_list args) {
     // Allocate enough memory for the output and null termination
     rv.resize(rsz+1);
     rsz = vsnprintf(&rv[0], rv.size(), fmt, args);
+    if (rsz + 1 < static_cast<int>(rv.size()))
+      rv.resize(std::max(rsz+1,0));
     return rv;
 }
 
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -40,6 +40,7 @@ distribution.
 
 #include <sstream>
 #include <map>
+#include <array>
 
 std::string stl_sprintf(const char *fmt, ...) {
     va_list lst;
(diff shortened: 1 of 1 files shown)


library/MiscUtils.cpp
```diff
@@ -50,21 +51,23 @@ std::string stl_sprintf(const char *fmt, ...) {
...
+    int rsz = vsnprintf(&buf[0], buf.size(), fmt, args2);
+    va_end(args2);
+    if (rsz < 0)
+        return std::string(); /* Error occurred */
+    if (static_cast<unsigned>(rsz) < buf.size())
+        return std::string(&buf[0], rsz); /* Whole string fits to a single line buffer */
+    std::string rv;
+    // Allocate enough memory for the output and null termination
+    rv.resize(rsz+1);
+    rsz = vsnprintf(&rv[0], rv.size(), fmt, args);
```
lethosor: Do you care about `rsz` here (any error checking needed)?

library/MiscUtils.cpp
```diff
@@ -67,6 +67,8 @@ std::string stl_vsprintf(const char *fmt, va_list args) {
     // Allocate enough memory for the output and null termination
     rv.resize(rsz+1);
     rsz = vsnprintf(&rv[0], rv.size(), fmt, args);
+    if (rsz + 1 < static_cast<int>(rv.size()))
+      rv.resize(std::max(rsz+1,0));
```
suokko: Z:/home/buildmaster/work/_E69339/_D0/src/library/MiscUtils.cpp(71): error C2589: '(': illegal token on right side of '::'
Z:/home/buildmaster/work/_E69339/_D0/src/library/MiscUtils.cpp(71): error C2059: syntax error: '::'

I need to figure out what causes these error messages.

lethosor: `max` is a macro defined in some Windows header somewhere, I think. There's a definition you can use to disable it. I think we have it set in a few places, but maybe not something `MiscUtils.cpp` includes.

lethosor:
```
depends/tthread/fast_mutex.h:  #define NOMINMAX
depends/tthread/tinythread.h:  #define NOMINMAX
library/include/modules/Filesystem.h:    #define NOMINMAX
```

PR #1301, opened 2018-06-10
Request c++11 support from compilers (through cmake)

suokko:
All platforms seems to either have compiler supporting c++14. Windows
already requires msvc2015 that supports c++14 which should make it
minimal issue to require c++14 support from all compilers because gcc
is pretty easy to upgrade.

I decided to make the change and submit the pull request to raise discussion if there is any setup blocking the change. I tested it quickly by creating a new pocket world and making a very short embark to there. I didn't notice any issues which at least shows most common features work with c++14 selected for gcc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ if(WIN32)
 endif()
 
 # Ask for C++14 standard from compilers
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 11)
 # Require the standard support from compilers.
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # Use only standard c++ to keep code portable
@@ -58,7 +58,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 if(MSVC)
   # cmake 3.10 is first version that support msvc standard settings
   if(${CMAKE_VERSION} VERSION_LESS "3.10.0")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++14")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++11")
   endif()
 # disable C4819 code-page warning
 add_definitions( "/wd4819" )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,18 @@ if(WIN32)
   endif()
 endif()
 
+# Ask for C++14 standard from compilers
+set(CMAKE_CXX_STANDARD 14)
+# Require the standard support from compilers.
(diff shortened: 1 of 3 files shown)


lethosor:
I think you missed my reply on IRC, but we still support GCC 4.8 because DF uses that, and some older Linux systems rely on builds using it. From the Travis status, it looks like GCC 4.8 doesn't fail with whatever flags were set, but it doesn't have full C++14 support from what I remember.

suokko:
My IRC connection was lost nearly immediately after I left my computer today.

gcc 4.8 doesn't support c++14 at all. The support requires gcc 5 (http://en.cppreference.com/w/cpp/compiler_support).

My df is build using 4.8.4 while dfhack is build with 7.3.0. It works as expected. Only possible problems would be backward incompatible ABI change in libstdc++ that would prevent compiler and c++ library upgrades.

lethosor:
GCC 4.8 supports `-std=c++1y`, but I don't know how much that includes.
```
           c++1y
               The next revision of the ISO C++ standard, tentatively planned for
               2017.  Support is highly experimental, and will almost certainly change
               in incompatible ways in future releases.

           gnu++1y
               GNU dialect of -std=c++1y.  Support is highly experimental, and will
               almost certainly change in incompatible ways in future releases.
```
I know DFHack can be built with GCC 7 - that's what I'm using too. But the GCC 7 build simply does not work for some Linux users, and the 4.8 build does.

suokko:
I think there was fairly little support for most features because cppreference seems to list versions when features was added into c++1y or c++1z in gcc and clang. I even remember installing experimental gcc 5 when system default compiler was 4.8.

But if gcc 4.8 is still hard requirement then I can change the version in the patch to c++11. It would then require c++11 and make it easy to later require a newer standard version.

lethosor:
Yeah, I do like the idea of using CMake for this instead of manually adding flags, assuming that it works, of course. Can you make sure it's setting the right flags on your end? They might be in CMakeCache.txt.

suokko:
I used verbose build to check compiler flags. They switched between 17, 14 and 11 when I changed the CMakeLists.txt.  But I don't know how old cmake has good support for the feature. At least windows support is only very recent. Online documentation seems to be available from 3.1.3 while 3.0 is missing.

BenLubar:
The C++14 version of this PR didn't cause any compile errors (or even warnings) on GCC 4.8:

https://buildmaster.local.lubar.me/applications/17/builds/build?releaseNumber=1301&buildNumber=1

(The downloads have been/will soon be deleted because it only retains the latest build of open PRs, but the logs should stick around for a while at least.)

lethosor:
Right, but even though `-std=c++0x` worked on GCC 4.5, it was still lacking several C++11 features (range-based for, `override`, etc).

Issue #1066, opened 2017-02-10
liquids/tiletypes prevent DF from closing


epsilon-phase:
If the liquids prompt is left open and you attempt to close the game, the game refuses to close until you quite liquids.

I haven't tried it with any of the other commands with their own prompts.

This really isn't too much of a serious problem, it's easy to get out of, but I'd like to see it fixed.

lethosor:
I think tiletypes has a similar issue.

lethosor:
@jaked122 What platform(s) are you seeing this issue on? I'm getting it on OS X, and presumably Linux too, but haven't checked Windows.

jaked122:
I'm on Linux 64 bit.

I don't recall if I saw this behavior on Windows when I used it

expwnent:
I vaguely remember getting it on Windows a long time ago.

lethosor:
The issue appears to be some thread waiting for the plugin to unload, so I don't think it's platform-specific. You can also produce this issue by running `unload tiletypes` from command-prompt or with dfhack-run while tiletypes is running.

lethosor:
Fixed in #1304

PR #1304, opened 2018-06-12
Multibyte console support for posix and fix potential data races in console shutdown

suokko:
I noticed that console code doesn't support multibyte characters at all. This may lead messed up console state and invalid multibyte sequence passed into command handlers. I decided to fix the input handling to keep console state match printed characters and prevent any partial multibyte characters passed into command handlers.

While I was implementing the multibyte support I noticed select was getting a suboptimal nfds parameters (first).  The large nfs parameter doesn't affect anything but it is minor performance issue because kernel has to scan up to nfds -1 FD bits in the fdset.

Then I also noticed the data race potential in the shutdown code. I decided to implement a simple fix by joining the thread before Core frees or cleanups any state.

Windows change is untested. Can someone test if it works correctly?
My assumption is that FreeConsole closes the console which makes ReadConsoleInputA return non-zero value that indicates an error.

### DIFF
--- a/library/Console-windows.cpp
+++ b/library/Console-windows.cpp
@@ -285,7 +285,10 @@ namespace DFHack
                 INPUT_RECORD rec;
                 DWORD count;
                 lock->unlock();
-                ReadConsoleInputA(console_in, &rec, 1, &count);
+                if (ReadConsoleInputA(console_in, &rec, 1, &count) != 0) {
+                    lock->lock();
+                    return -2;
+                }
                 lock->lock();
                 if (rec.EventType != KEY_EVENT || !rec.Event.KeyEvent.bKeyDown)
                     continue;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,6 +137,12 @@ ${CMAKE_MODULE_PATH}
 # generates compile_commands.json, used for autocompletion by some editors
 SET(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+include(CheckIncludeFileCXX)
+CHECK_INCLUDE_FILE_CXX("cuchar" HAVE_CUCHAR)
+if(HAVE_CUCHAR)
+	add_definitions("-DHAVE_CUCHAR")
+endif()
+
 # mixing the build system with the source code is ugly and stupid. enforce the opposite :)
 if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
    message(FATAL_ERROR "In-source builds are not allowed.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,10 +137,19 @@ ${CMAKE_MODULE_PATH}
 # generates compile_commands.json, used for autocompletion by some editors
 SET(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-include(CheckIncludeFileCXX)
-CHECK_INCLUDE_FILE_CXX("cuchar" HAVE_CUCHAR)
-if(HAVE_CUCHAR)
-	add_definitions("-DHAVE_CUCHAR")
+include(CheckCXXSourceCompiles)
+CHECK_CXX_SOURCE_COMPILES("
+#include <cstdlib>
+#include <cuchar>
+int main(void) {
+    char32_t in = 0;
+    char out[MB_CUR_MAX];
+    std::mbstate_t state{};
+    std::c32rtomb(out, in, &state);
+    return 0;
+}" HAVE_CUCHAR2)
+if(HAVE_CUCHAR2)
+    add_definitions("-DHAVE_CUCHAR")
 endif()
 
 # mixing the build system with the source code is ugly and stupid. enforce the opposite :)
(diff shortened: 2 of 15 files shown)


suokko:
I tested cuchar fallback implementation locally with gcc 4.8.

There is still windows change that needs testing. Possible issues with changes is that dfhack hangs waiting console thread to exit when tying to shutdown. Good test case would be trying to shutdown dfhack with remote client or just using df main menu to quit. Both should result to a clean exit without hangs or crashes if my change is correct.

library/Console-windows.cpp
```diff
@@ -285,7 +285,10 @@ namespace DFHack
                 INPUT_RECORD rec;
                 DWORD count;
                 lock->unlock();
-                ReadConsoleInputA(console_in, &rec, 1, &count);
+                if (ReadConsoleInputA(console_in, &rec, 1, &count) != 0) {
+                    lock->lock();
+                    return -2
```
BenLubar: Missing a semicolon here.

suokko:
Some of those are available in common Linux keyboard layouts if dead keys are enabled. But then the first key press is never sent to applications. The accent key is then combined with following character which often result to a multibyte UTF-8 precombined character. Of course as a programer my default layout is an dead keys disabled variant. Then there is also Alt Gr+key. Example š comes directly from Alt Gr+s.

But issue would be characters that don't have precombined variation or input system doesn't know to generate the precombined character.

suokko:
Seems like mac has cuchar header but mbrtoc32/c32rtomb are missing. I guess it means that _GLIBCXX_USE_C11_UCHAR_CXX11 is not set, Does that mean mac doesn't have those functions in C library?

I guess I should improve cmake check to also see if functions exists.

lethosor:
Weird, it looks like it's there on my end:
```
/usr/local/Cellar/gcc@7/7.3.0/include/c++/7.3.0//cuchar:#undef c32rtomb
/usr/local/Cellar/gcc@7/7.3.0/include/c++/7.3.0//cuchar:  using ::c32rtomb;
/usr/local/Cellar/gcc@7/7.3.0/include/c++/7.3.0//cuchar:#undef mbrtoc32
/usr/local/Cellar/gcc@7/7.3.0/include/c++/7.3.0//cuchar:  using ::mbrtoc32;
```

lethosor:
Yeah, they might be missing from the C library, though - they're not in /usr/include.

BenLubar:
Mac and Linux should be using exactly the same version of GCC and glibc unless that function is defined in the Linux kernel headers somewhere.

`/osxcross/target/x86_64-apple-darwin14/include/c++/7.3.0/cuchar` is this file on the builder image: https://github.com/gcc-mirror/gcc/blob/gcc-7_3_0-release/libstdc%2B%2B-v3/include/c_global/cuchar

suokko:
Functions should be implemented in c-library. The included is uchar.h (/usr/include/uchar.h in linux)

lethosor:
I have /usr/include/wchar.h on macOS but no uchar.h that I can find (`gcc-7 -include uchar.h -x c -` also fails to find it).

lethosor:
This PR, presumably the threading rewrite, has made https://github.com/DFHack/dfhack/issues/1066 worse. Specifically, tiletypes and liquids now go into an infinite loop when quitting DF from the main menu. It appears that they're continuously reading empty lines, which causes at least tiletypes to produce an error message, which gets logged to stderr.log (my log grew to around 300 MB, and this would be bad for anyone who doesn't know to check that file).

Also, when the `lua` interpreter is running, quitting DF just hangs (presumably the same thing where the interpreter is reading blank lines, just not logging an error). Previously, the interpreter would exit just fine.

I haven't looked into #1066 much, and I'm afraid I don't have much insight into why the C++ tools hang and the Lua ones don't, but the behavior introduced here isn't better.

lethosor:
This also corrupts command history - running "help", then pressing [up] displays "help ", and pressing [enter] produces "help is not a recognized command." Looks like some hidden characters are breaking things.

lethosor:
Yeah, null terminators are making it into the command history somehow. From my dfhack.history:
```
die
help^@
help
die
lua^@
lua
foo^@
foo
^@
^@
tiletypes^@
lua^@
tiletypes^@
die^@
```

suokko:
I forgot add Fixes #1066 to commit messages but I tested that now exiting with liquids, tiletypes or lua interpreter doesn't hang.

suokko:
I added a "Use timed mutex to avoid CoreSuspender deadlock" as a forward compatible safety try_lock_for alternative in vmethods. The deadlock doesn't happen with current 0.44.10 but noting prevents core DF to change threading which could lead to deadlock chances in future.

BenLubar:
If DF changes its threading model, `CoreSuspender` will likely no longer be a thing, or at the very least it will need to be completely re-written.

suokko:
I dropped deadlock detection from this branch. It is still available from https://github.com/suokko/dfhack/tree/multibyte_console_deadlock_detection

lethosor:
Minor issue: Ctrl-C in lua, tiletypes, etc. exits the interpreter instead of just cancelling the current line (and prompting for a new line) like it used to. I think the lineedit() return value checks in the plugins can be changed to handle that, but I'm not sure about Lua.

PR #1306, opened 2018-06-12
Start script additions strace, gdbserver and df arguments to gdb

suokko:
I have made a few modifications to my dfhack script. Here are modifications if they look like something that would help others too.

I didn't include ltrace change because I don't know setup that would feel like useable. I either get very slow execution or not enough information. A compromise could be allowing the -x argument of ltrace to be passed for dfhack. But I don't know if it can be done with a simply change.

### DIFF
--- a/package/linux/dfhack
+++ b/package/linux/dfhack
@@ -68,11 +68,8 @@ setarch_arch=$(cat hack/dfhack_setarch.txt || printf i386)
 case "$1" in
   -g | --gdb)
     shift
-    echo "set environment LD_LIBRARY_PATH $LD_LIBRARY_PATH" > gdbcmd.tmp
-    echo "set environment LD_PRELOAD $PRELOAD_LIB" >> gdbcmd.tmp
-    echo "set environment MALLOC_PERTURB_ 45" >> gdbcmd.tmp
-    echo "set startup-with-shell off" >> gdbcmd.tmp
-    gdb $DF_GDB_OPTS -x gdbcmd.tmp ./libs/Dwarf_Fortress "$@"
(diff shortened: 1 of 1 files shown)


package/linux/dfhack
```diff
@@ -68,12 +68,37 @@ setarch_arch=$(cat hack/dfhack_setarch.txt || printf i386)
 case "$1" in
   -g | --gdb)
     shift
+    echo "set exec-wrapper env LD_LIBRARY_PATH='$LD_LIBRARY_PATH' LD_PRELOAD='$PRELOAD_LIB' MALLOC_PERTURB_=45" > gdbcmd.tmp
```
lethosor: Is this the same as `startup-with-shell off` below? I remember `startup-with-shell` being added recently to fix an issue, so I would like to stick with it unless they're equivalent.

suokko: exec-wrapper runs inside startup shell setting up environment. It improves argument handling because gdb argument handling without shell doesn't support spaces.  df -gen parameter is likely to require a parameter with a spaces. That why keeping shell and using exec-wrapper helps if someone wants to use parameters and gdb.

Too bad gdbserver connection doesn't handle spaces or exec-wrapper so my prefered debug setup won't handle arguments.

lethosor: #1280 is what I was referring to. So you've checked that `-g` still works with your changes?

suokko: It works. I rechecked it now:
./dfhack -g -gen 4 4 "POCKET ISLAND"
That opened df with dfhack loaded.

lethosor: It's really the GDB part I'm concerned about, but thanks for double-checking!

PR #1308, opened 2018-06-14
Fix for Unlikely screen show memory leak fix and check to avoid breaking the screen linked list

suokko:
Showing a screen which is already visible would mess up the linked list.
A simple check to prevent breaking the linked list is to let
Screen::show check if screen is in a list while looking for the parent
screen.

Fix unlikely memory leaks if Screen::show fails  …
The Screen::show takes ownership of the screen pointer. I decided to
switch the parameter to std::unique_ptr to make the pointer ownership
explicit. The unique_ptr then provides automatic screen destruction in
Screen::show unless pointer is inserted or is already in the linked list
that is managed by df.

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -2242,11 +2242,7 @@ int screen_show(lua_State *L)
 
     df::viewscreen *screen = dfhack_lua_viewscreen::get_pointer(L, 1, true);
 
-    bool ok = Screen::show(screen, before);
-
-    // If it is a table, get_pointer created a new object. Don't leak it.
-    if (!ok && lua_istable(L, 1))
-        delete screen;
+    bool ok = Screen::show(std::unique_ptr<df::viewscreen>{screen}, before);
 
     lua_pushboolean(L, ok);
     return 1;
--- a/library/include/MiscUtils.h
+++ b/library/include/MiscUtils.h
@@ -31,6 +31,7 @@ distribution.
 #include <vector>
 #include <sstream>
 #include <cstdio>
+#include <memory>
 
 using std::ostream;
 using std::stringstream;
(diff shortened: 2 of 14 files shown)


library/include/MiscUtils.h
```diff
@@ -44,6 +44,28 @@ using std::endl;
...
+ * standard version if one is available. The standard version gets imported with
+ * using.
+ */
+namespace dts {
+//  Check if lib supports the feature test macro or version is over c++14.
+#if __cpp_lib_make_unique < 201304 && __cplusplus < 201402L
+//! Insert c++14 make_unique to be forward compatible. Array versions are
+//! missing
+template<typename T, typename... Args>
+typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T> >::type
```
lethosor: Looks like you need to include <memory> in this file.

library/LuaApi.cpp
```diff
@@ -2242,11 +2242,7 @@ int screen_show(lua_State *L)
 
     df::viewscreen *screen = dfhack_lua_viewscreen::get_pointer(L, 1, true);
 
-    bool ok = Screen::show(screen, before);
-
-    // If it is a table, get_pointer created a new object. Don't leak it.
-    if (!ok && lua_istable(L, 1))
-        delete screen;
+    bool ok = Screen::show(std::unique_ptr<df::viewscreen>{screen}, before);
```
lethosor: get_pointer has separate logic for handling tables (Screen instances from gui.lua) and raw viewscreen pointers. Does this work appropriately with both of those?
It's uncommon for Lua code to call show() directly on native screens, but it should work. I think `open-legends` and `modtools/create-unit` do this.

suokko: Thanks. I was wondering about the native screen pointers from lua when I was looking the lua part of code. But then I forgot to find a script that uses them to test if it works. Now I checked than `open-legends` at least works correctly.

PR #1309, opened 2018-06-14
jsoncpp: Fix (U)Int64 to long int on 64bit linux

suokko:
Passing a 64bit integer to Json::Value fails to find matching. Even
tough long long int is also 64bits gcc considres long long int different
to long int. But fix can't be as simple as removing an long from type
because that would reduce 32bit builds to have only 32bit maximum int
size. But standard offers fixed width integer types that can be used to
get correct underlying type depending on platform.

### DIFF
--- a/depends/jsoncpp/jsoncpp.h
+++ b/depends/jsoncpp/jsoncpp.h
@@ -195,6 +195,8 @@ license you like.
 #define JSONCPP_DEPRECATED(message)
 #endif // if !defined(JSONCPP_DEPRECATED)
 
+#include <cstdint>
+
 namespace Json {
 typedef int Int;
 typedef unsigned int UInt;
@@ -208,8 +210,8 @@ typedef unsigned int LargestUInt;
 typedef __int64 Int64;
 typedef unsigned __int64 UInt64;
 #else                 // if defined(_MSC_VER) // Other platforms, use long long
-typedef long long int Int64;
-typedef unsigned long long int UInt64;
+typedef int64_t Int64;
+typedef uint64_t UInt64;
 #endif // if defined(_MSC_VER)
 typedef Int64 LargestInt;
 typedef UInt64 LargestUInt;


lethosor:
What type exactly were you "passing" to Json::Value? `long long` is 64 bits on all platforms that we support.

Also, this is a generated header. It's a small enough change that I think it's manageable to make it every time we upgrade json-cpp (which we haven't yet), but I don't want to be making lots of our own changes to this file.

suokko:
long unsigned was my type that failed to find matching constructor.

Issue is not width of long long but that it is different type to all default 64 bit types when compiling with gcc. Because long long and long is different types it allows overloads to define both even tough they both take exactly same hardware type on 64bit systems.

lethosor:
Ok, but why were you dealing with an `unsigned long` instead of `unsigned long long`? Does the compiler default to that or something?

I feel like this is something that could be addressed by adding conversion operator from `[unsigned] long` to `Json::Value`. I don't see how your change affects anything, since `int64_t` should be typedef'd to `long long` anyway.

suokko:
Conversion operator could actually work too. I will check if I can make it work.

But types in /usr/include/bits/types.h are defined like
#if __WORDSIZE == 64
typedef signed long int __int64_t;
typedef unsigned long int __uint64_t;
#else
__extension__ typedef signed long long int __int64_t;
__extension__ typedef unsigned long long int __uint64_t;
#endif

And those are then used to define stdint.h standard types.

lethosor:
Oh, right, I remember some inconsistency with those on Linux.

We have our own extensions in json-ex.h (I think that's the name, anyway). If it could go in there, that would be ideal.

lethosor:
From an IRC discussion: this was already fixed upstream: https://github.com/open-source-parsers/jsoncpp/commit/b9afdf190dbccbf3a59a3aba86eb295c0fa07068#diff-3c93282e96d1ef08ede694c0f5ea8108

I think a good solution is to merge this for now, then update jsoncpp when we've determined that it will work (one notable change is that FastWriter is apparently deprecated now).

PR #1312, opened 2018-06-16
Linked list

BenLubar:


library/include/BitArray.h
```diff
@@ -232,4 +234,299 @@ namespace DFHack
...
+        }
+        iterator end()
+        {
+            L *it = static_cast<L *>(this);
+            while (it->next && it->next->next)
+            {
+                it = it->next;
+            }
+            return iterator(static_cast<L *>(this), it, true);
+        }
```
suokko: end should be as simple as return iterator(nullptr) because the end should point to element after the last valid element. But that wouldn't be perfectly valid because std::prev(end()) is required to work.

library/include/BitArray.h
```diff
@@ -232,4 +234,299 @@ namespace DFHack
...
+        {
+            return static_cast<const L *>(this)->next == nullptr;
+        }
+        size_t size() const
+        {
+            size_t n = 0;
+            for (I *i : *this)
+                n++;
+            return n;
+        }
```
suokko: maybe return std::distance(begin(),end());

library/include/BitArray.h
lethosor: For forward iterators? I don't think we have a fast way to go backwards (or start at the end of a list) here.

suokko: dflist is double linked list expect its root isn't correctly linked to end and begin. But if we would provide only forward iterator and simple implement `begin() { return iterator(root->next)}` and `end() {return iterator(nullptr);}` and adapt iterator to simple keep pointer to df::list_link. Then end iterator would compare equal only when iterator has been increment one past the end.

lethosor: Oh, right, I thought it was singly-linked (probably because Ben was implementing a forward iterator).

suokko: For practical use it is same as singly-linked list except it makes erase a bit simpler. But the link between root and end would need to be added to have all double-linked list features.

BenLubar: Is `std::prev` required to have a specific time complexity? I could make the end of the list be a special case where it computes it on demand.

suokko: complexity requirements and other are mostly only for stl implementations. DF list can't be made easily conform to std::list requirements. The hardest requirement is that size() must be O(1). But std::forward_list requirements are something that DF list can fairly easily match. There it would have only forward iterator which would mean that std::prev(end()) isn't valid use. But then we could mix that with std::list interface for simpler usage as we actually have double-linked list which makes it easy to implement erase and insert with only current list_link pointed by iterator.
http://en.cppreference.com/w/cpp/container/forward_list
http://en.cppreference.com/w/cpp/container/list

I think that would then allow dfhack code use our "std::list" like container with most important stuff like range-for loop, stl algorithms and std::list like erase and insert interface.

suokko: Actually. There is the performance issue of calling list::insert(end(), new_entry). But we could disallow use of that kind of construct.

library/include/BitArray.h
```diff
@@ -232,4 +234,299 @@ namespace DFHack
             m_data[idx] = item;
         }
     };
+
+    template <typename L, typename I>
```
suokko: If you want this could just take a single template parameter and then require example requirements would be
<pre>
template&#60;typename ValueType&#62;
struct DfLinkedList {
using link_type = DfLinkedList&#60;ValueType&#62;::type;
};
// This would then be required only once for each linked list type
template<>
struct DfLinkedListLinkType&#60;job*&#62; {
using type = job_list_link;
};
// List could be then declared with familiar syntax.
DfLinkedList&#60;job*&#62; list;
</pre>

BenLubar: Since the item field is always named `item`, we could also use `decltype(L::item)`.

library/include/BitArray.h
BenLubar: You can't subtract the iterators in constant time because it's a linked list.

library/include/BitArray.h
BenLubar: I've changed it so the only performance drawback on `end()` will be when moving it. You can compare with it for basically free now.

library/include/BitArray.h
```diff
@@ -526,27 +407,125 @@ namespace DFHack
...
+            else if (link->prev)
+            {
+                link->prev->next = nullptr;
+            }
+
+            if (root->next == link)
+            {
+                root->next = next;
+            }
+
```
suokko: Can we reduce branches?

Maybe something like following would work:
<pre>
if (link->prev) link->prev->next = link->next;
else root->next = link->next;
if (link->next) link->next->prev = link->prev;
</pre>

library/include/BitArray.h
```diff
@@ -526,27 +407,125 @@ namespace DFHack
...
+            auto root = static_cast<L *>(this);
+            auto link = new L();
+            link->prev = nullptr;
+            if (root->next)
+            {
+                root->next->prev = link;
+                link->next = root->next;
+            }
+            link->item = item;
+            root->next = link;
```
suokko: Did you notice that push_front could call insert_after?
<pre>
insert_after(iterator(root), item);
</pre>
Only problem is when assigning newlink->prev = link which should be instead null for when link == root

library/include/BitArray.h
```diff
@@ -526,27 +407,125 @@ namespace DFHack
...
+            if (newlink->prev)
+            {
+                newlink->prev->next = newlink;
+            }
+            else if (link == root->next)
+            {
+                root->next = newlink;
+            }
+            newlink->item = item;
+            return iterator(root, newlink);
```
suokko: I wonder if insert could work even if link->prev == nullptr (inserting to the begin). But not if iterator is end().
<pre>
newlink->prev = link->prev;
newlink->next = link;
if (link->prev) link->prev->next = newlink;
else root->next = newlink;
link->prev = newlink;
</pre>

library/include/BitArray.h
BenLubar: Actually, after trying it, we can't use `std::remove_pointer<decltype(L::item)>::type` because `L` is just the name of a class at the point when that code would be evaluated by the compiler and there aren't any fields associated with that class yet at that point.

suokko: That sounds like an alternative. Only issue to me is that I feel link_type should be implementation detail that isn't "exposed" outside wrapper class.

suokko: using value_type = std::remove_pointer<decltype(L{}.item)>::type; might be the syntax you need to make it work. But link_type should be always fully available type when template class is instantiated in the code using it. If it isn't fully defined type then whole class would fail. `DfLinkedListLinkType<job*> list;`  should be replacement for `job_list_link list;` where both require job_list_link to be fully defined.

PR #1313, opened 2018-06-18
 Update jobs when committing MapCache changes

### LINKED ISSUE #1229: digv can sometimes result in "Inappropriate dig square" cancellation
From #481's fix
Some tiles are dug without flashing, and then the neighboring tiles produce this announcement. Note that this does not actually remove the designation, and usually the tile will get mined shortly after (maybe it's double-designated somehow)?
See also #1228

suokko:
While I was implementing the fix I decided to make a minor change to index_tile. I tough that index_tile should automatically determine return type from the array element type because c++11 allows it easily. That change would conflict with the fix so I included it as a separate commit to same pull request.

The map_block->designation.{dig,smooth} are reset to zeros when a job
posting is created for the designation. The job is then used to override
the designation state in the map_block. To make the new designation set
propogate to jobs the job structure would require updating. The update
would be possible a complex operation. The simple alternative is to
remove the job and let df create a new job in the next tick.

Fixes #1229

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -48,6 +48,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `ban-cooking`: fixed errors introduced by kitchen structure changes in 0.44.10-r1
 - `remove-stress`: fixed an error when running on soul-less units (e.g. with ``-all``)
 - `revflood`: stopped revealing tiles adjacent to tiles above open space inappropriately
+- `dig`: Fix "Inappropriate dig square" announcements if digging job has been posted
 
 ## Misc Improvements
 - Added script name to messages produced by ``qerror()`` in Lua scripts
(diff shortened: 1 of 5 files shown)


suokko:
I noticed that this fails if there is unrelated jobs like hauling in the tile.

lethosor:
Slight annoyance: this cancels jobs that have workers assigned. Not sure if that's always necessary to get the priorities to update, but I don't think it's consistent with how DF reacts when changing priority of a tile that a miner is digging out.

lethosor:
Honestly, I only complained because re-prioritizing a `digv` designation caused my miners to stop digging the tiles they digging and go eat instead. Once a tile is being dug, adjusting the priority probably won't do anything. Maybe it could skip cancelling/recreating designation jobs with workers entirely? (Avoiding recreating them might be harder.)

PR #1315, opened 2018-06-19
Add cmake configuration to select if stubs include symbols

suokko:
G++ generates structure debug symbols for a few df namespace classes to
generated stub source files. I decided to test how much symbols from
those files would increase binary size. When the result was about double
size I decided to add cmake configuration option to let user easily
select if they prefer complete symbols or reduced size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ else(CMAKE_CONFIGURATION_TYPES)
 endif(CMAKE_CONFIGURATION_TYPES)
 
 OPTION(BUILD_DOCS "Choose whether to build the documentation (requires python and Sphinx)." OFF)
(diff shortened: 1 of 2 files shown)


lethosor:
By "stubs" are you referring to the constructors/field data defined in these files or something else? Some of the constructors do things (they're mostly limited to initializing fields with their `init-value` XML attributes, from what I remember).

suokko:
Yes, Constructors and static member functions. They were called stubs in cmake comment so I used same name in the commin message

lethosor:
Ah, yeah, that comment was there before. Maybe it's outdated. I was mostly wondering what symbols you thought were useful from there.

suokko:
structure symbols for a few classes are generated there. Now I can at least remember df::job_handler, df::building and classes derived df::building.

Issue #1276, opened 2018-05-17
[0.44.10-beta1 x64] prospect causes crash


Tharn:
Using a savegame grandfathered in from 0.43.X (not sure if this is relevant or not), the game crashes immediately and without any error popup if I run `prospect hell all` or even just `prospect`.

The last line in stderr.log reads ``Invoking: prospect``

lethosor:
I can't reproduce this with saves between 0.40 and 0.44. Can you upload your save somewhere?

Tharn:
Here's the save: https://mega.nz/#!NXoz0aRL!rt0J7cg2OQFLejHyz1UIKElloMHql4duUEwcq0AWbao

lethosor:
That helps a lot, thanks. I got output through the `>>> TOTAL = ` line, then a backtrace:
```
Thread 7 Crashed:
0   prospector.plug.dylib         	0x000000010df3b716 printVeins(DFHack::color_ostream&, std::map<short, matdata, std::less<short>, std::allocator<std::pair<short const, matdata> > >&, DFHack::Materials*, bool) + 230
1   prospector.plug.dylib         	0x000000010df3e9d5 prospector(DFHack::color_ostream&, std::vector<std::string, std::allocator<std::string> >&) + 7717
```
so this should help us figure out what's going on.

lethosor:
There's something weird with vein materials on your map that I haven't figured out yet. I feel like this has come up before, but the closest thing I've found is a45438d17273988fd00cc79896578fc55fbb5b82, which is a fix for 3dveins but not the same cause.
Anyway, the crash should be replaced with a warning now. `prospect` and `prospect hell all` don't crash any more on your save. I'll create another issue for tracking down the vein weirdness, so please keep your save up if you can.

PR #1316, opened 2018-06-19
Fix stockpile inventory not to include inventory items

### LINKED ISSUE #1288: [0.44.10-r1 x64] Stockpile 'i' function lists military worn inventory
I have a game with very specialized stockpiles. Browsing through them with DFHack's `i`nventory function will display ghost entries of stuff that isn't actually in the bins. My stockpiles for finished goods, wooden shields, bone weapons and the various armour and weapons segregated by metal quality will *all* include items worn in the current inventory of military units.

And these entries are not properly sorted by stockpile, i.e. the finished goods stockpile won't just list waterskins that have been taken from there but whole military inventories, including steel gear, quivers, waterskin contents and so forth. And it's the same with all other stockpiles.

Just noticed it now, but it's probably visible in the save I already uploaded. Subterran level 3 has finished goods stockpiles on the top row sorted by quality. Subterran level 12 has all the weapon and armour stockpiles. Easiest way to see this is to hide all stockpile filters except 'In Inventory'.

The weirdest example is `raw mussel [2]` found in the *furniture* inventory that has had large pots / food storage disabled for in-game *decades*. No idea how that listing got there, as there's not even a reason for military dwarves to interact with that stockpile.

suokko:
Inventory items have stale position information. The coordinates can be
very old if item is owned and continuesly carried for long time. The fix
checks if item is carried. To avoid filtering carried items that are
assigned (containers) to the stockpile the assignment needs to be
checked.

Fixes #1288

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -49,6 +49,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `remove-stress`: fixed an error when running on soul-less units (e.g. with ``-all``)
 - `revflood`: stopped revealing tiles adjacent to tiles above open space inappropriately
 - `stockpiles`: ``loadstock`` sets usable and unusable weapon and armor settings
+- `stocks`: Remove carried items from stockpile where they were picked up
 
 ## Misc Improvements
 - Added script name to messages produced by ``qerror()`` in Lua scripts
--- a/plugins/uicommon.h
+++ b/plugins/uicommon.h
@@ -348,10 +348,22 @@ public:
 
     bool inStockpile(df::item *i)
     {
-        df::item *container = Items::getContainer(i);
+        // Check if the item is assigned to a stockpile
+        if (i->isAssignedToStockpile())
+            return i->isAssignedToThisStockpile(sp->id);
+
+        // If this is in a container check where the container item is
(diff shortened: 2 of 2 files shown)


lethosor:
Are you aware of `Items::getPosition()`?

suokko:
Thanks. I didn't notice it.

There is still the issue that getPosition would return position inside stockpile when unit is walking over stockpile. I propose that inStockpile still needs different code than getPosition and the updated version in the patch would be at least close to correct.

lethosor:
Yeah, in this case an inventory check sounds good to me. I'm just wondering if some of it can be changed to use getPosition instead. But as long as this addresses the issue, it's fine with me.

suokko:
I tough about if getPosition could be used but issue is that stockpile check wants to follow only container holder chain while all other chains mean that the item can't be in a stockpile.

lethosor:
Unless I'm mistaken somehow, this doesn't appear to fix the specific issue mentioned with the furniture stockpile in #1288 (save from #1276, furniture stockpile to the right of the initial cursor, "raw mussel [2]" still present).

lethosor:
Never mind - I was sure I had installed and reloaded the plugin, but installing everything and restarting DF fixed it. Looks good!

PR #1318, opened 2018-06-20
Change gui.widgets.List:setChoices to keep state in internal table 

### LINKED ISSUE #1171: The widgets.lua List:setChoices trashes the supplied list
The operation takes the list of strings supplies and then rehacks it into a list of list that have a string element. That would be perfectly fine if it copied the list first, but it actually trashes the list given to it, which means you can't e.g. use the original list and add elements to it to pass in a new set.
It's reasonably easy to work around it by making sure you always pass a copy of your list to the operation for it to rip apart, but there's no warning label for the unwary, and the operation really ought to do the copying itself (and it isn't really a good idea to create external references to its internal data without a clear warning either).

suokko:
Requires https://github.com/DFHack/scripts/pull/65 to keep adventure fort script working.

Fixes #1171

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -458,7 +458,7 @@ end
 function List:setChoices(choices, selected)
     self.choices = {}
 
-    for i,v in ipairs(choices) do
+    for i,v in ipairs(choices or {}) do
         local l = utils.clone(v);
         if type(v) ~= 'table' then
             l = { text = v }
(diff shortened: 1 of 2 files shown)


library/lua/gui/widgets.lua
```diff
@@ -456,15 +456,17 @@ function List:init(info)
 end
 
 function List:setChoices(choices, selected)
-    self.choices = choices or {}
+    self.choices = {}
 
-    for i,v in ipairs(self.choices) do
+    for i,v in ipairs(choices) do
```
lethosor: This breaks the ability to do `setChoices()` to clear the list. I don't know if that's important (`setChoices({})` still works), but maybe `choices or {}` here would be better.

suokko: Thanks. Fixed my error.

PR #1319, opened 2018-06-20
Temporary lower command-prompt when executing the command

### LINKED ISSUE #1194: spotclean doesn't work from the in-game terminal
Selecting a tile with `k` and attempting to run `spotclean` with the in-game terminal, [command-prompt](https://dfhack.readthedocs.io/en/stable/docs/Plugins.html#id232), fails with the following error:

```
Could not invoke spotclean: unsuitable UI state.
```

Typing `spotclean` into the actual command prompt succeeds, naturally. It seems like the presence of the in-game command prompt is incorrectly recognized as an unsuitable UI state.

suokko:
command-prompt viewscreen may affect command execution if they are
looking for UI state. To make commands execute simillary to hotkeys or
console commands the viewscreen needs to removed from the top position.

Fixes #1194

I'm unsure if LowerScreen should be part of Core or module/Screen. I first put it to the Core but I can moved it to Screen if it sounds like a better place for this functionality.

What about lowering position? I only lower it a level but it might be better to lower to bottom (gview->view.child = screen_)

### DIFF
--- a/library/include/Core.h
+++ b/library/include/Core.h
@@ -70,6 +70,11 @@ namespace DFHack
         class df_window;
     }
 
+    namespace Screen
+    {
+        struct Hide;
+    }
+
     enum state_change_event
     {
         SC_UNKNOWN = -1,
@@ -273,6 +278,7 @@ namespace DFHack
         void *last_world_data_ptr;
         // for state change tracking
         void *last_local_map_ptr;
+        friend struct Screen::Hide;
         df::viewscreen *top_viewscreen;
         bool last_pause_state;
         // Very important!
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -49,6 +49,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `remove-stress`: fixed an error when running on soul-less units (e.g. with ``-all``)
 - `revflood`: stopped revealing tiles adjacent to tiles above open space inappropriately
 - `stockpiles`: ``loadstock`` sets usable and unusable weapon and armor settings
+- `command-prompt`: Allow executing commands that require specific screen to be visible
 
 ## Misc Improvements
 - Added script name to messages produced by ``qerror()`` in Lua scripts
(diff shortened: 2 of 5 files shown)


library/include/Core.h
```diff
@@ -198,6 +198,14 @@ namespace DFHack
 
         static void cheap_tokenise(std::string const& input, std::vector<std::string> &output);
 
+        //! Temporary lower a screen until constructor is called
```
lethosor: constructor -> destructor?

library/Core.cpp
```diff
@@ -2769,6 +2769,37 @@ std::string Core::GetAliasCommand(const std::string &name, const std::string &de
...
+    df::viewscreen* bc = b->child;
+
+    ap->child = b;
+    b->parent = ap;
+
+    a->parent = b;
+    b->child = a;
+
+    if (bc) bc->parent = a;
+    else Core::getInstance().top_viewscreen = a;
```
lethosor: Does anything important actually use this? I was under the impression that things like spotclean used getCurViewscreen(), which doesn't use this.

suokko: Gui::cursor:hotkey() gets it from Plugin::invoke() which uses Core::getTopViewscreen()

suokko:
Completely unlinking sounds safer. I will change to that. If I set links carefully I think that any stale pointers to an old top viewscreen would still handle it sanely. I just need to figure if keeping screen->parent  or setting it to null is safer.

lethosor:
Also, forgot to mention it, but I think this belongs in the screen module. An exception might be if you definitely need to work with core internals, like top_viewscreen, but even then I might prefer just allowing the thing in the screen module to deal with top_viewscreen directly (if needed).

lethosor:
This completely breaks Stonesense, because Stonesense is garbage. Not your fault - apparently there's an X11 Screen struct, and Stonesense has `using namespace DFHack` in several places before including X11 headers.

suokko:
The stonesense change that I originally forgot to commit and push and then link to here: https://github.com/DFHack/stonesense/pull/59

lethosor:
Oops. Well, hopefully it'll be more resilient to future changes that conflict with X11/Allegro/whatever

PR #1322, opened 2018-06-22
Reset dsgn_check_cooldown in Block::write

suokko:
To make sure that DF generates jobjs from desgination as soon as
possible the dsgn_check_cooldown should be reset when designations
change.

Combined with a janitor change:
Use constexpr to prevent attempts of linking static variable

### DIFF
--- a/library/include/modules/MapCache.h
+++ b/library/include/modules/MapCache.h
@@ -55,7 +55,7 @@ class Block;
 
 struct BiomeInfo {
     // Determined by the 4-bit index in the designation bitfield
-    static const unsigned MAX_LAYERS = 16;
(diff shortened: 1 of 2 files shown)


library/include/modules/MapCache.h
```diff
@@ -55,7 +55,7 @@ class Block;
 
 struct BiomeInfo {
     // Determined by the 4-bit index in the designation bitfield
-    static const unsigned MAX_LAYERS = 16;
+    static constexpr unsigned MAX_LAYERS = 16;
```
ragundo: Cool trick
I didn't know that constexpr is equivalent to inline for static variables

PR #1328, opened 2018-06-29
 Support dlsym loading from libgraphics vtables

suokko:
Uses changes from https://github.com/DFHack/df-structures/pull/282

### DIFF
--- a/library/PlugLoad-windows.cpp
+++ b/library/PlugLoad-windows.cpp
@@ -41,8 +41,7 @@ distribution.
  */
 namespace DFHack
 {
-    //! \todo What windows HMODULE matches RTLD_DEFAULT?
-    DFLibrary* GLOBAL_NAMES = 0;
+    DFLibrary* GLOBAL_NAMES = (DFLibrary*)GetModuleHandle(nullptr);
     DFLibrary * OpenPlugin (const char * filename)
     {
         return (DFLibrary *) LoadLibrary(filename);
--- a/library/PlugLoad-windows.cpp
+++ b/library/PlugLoad-windows.cpp
@@ -41,6 +41,8 @@ distribution.
  */
 namespace DFHack
 {
+    //! \todo What windows HMODULE matches RTLD_DEFAULT?
+    DFLibrary* GLOBAL_NAMES = 0;
     DFLibrary * OpenPlugin (const char * filename)
     {
         return (DFLibrary *) LoadLibrary(filename);
--- a/library/VersionInfoFactory.cpp
+++ b/library/VersionInfoFactory.cpp
@@ -35,6 +35,7 @@ using namespace std;
 #include "VersionInfo.h"
 #include "Error.h"
 #include "Memory.h"
+#include "PluginManager.h"
 using namespace DFHack;
 
 #include <tinyxml.h>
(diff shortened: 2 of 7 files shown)


BenLubar:
Automated PR generator is here: https://github.com/BenLubar/dfhack-raft/blob/otter/modules/Update-DF-Structures

BenLubar:
Ok, automated PR generator should be okay until the pull request is merged: https://github.com/BenLubar/dfhack-raft/commit/c1205bd677ca88953c39fe9b3fcb1a088bfd3c0c

library/PlugLoad-windows.cpp
```diff
@@ -41,6 +41,8 @@ distribution.
  */
 namespace DFHack
 {
+    //! \todo What windows HMODULE matches RTLD_DEFAULT?
+    DFLibrary* GLOBAL_NAMES = 0;
```
BenLubar: From what I've read, it seems to be `GetModuleHandle(NULL)`.

lethosor: Sure, but this should only be necessary on Linux (for now, anyway)

BenLubar:
@suokko can you rebase this on top of https://github.com/DFHack/dfhack/commit/aad4bbe65c73ae2c9af12fbe7c0e0e74259d460b to make Travis happy?

library/PlugLoad-windows.cpp
BenLubar: I don't think there's any downside to using `GetModuleHandle(NULL)` right now, since it'll just add a reference to the current process that will go away when the current process exits anyway. It'll make it easier if we need it in the future.

suokko:
I will now that I got my next change ready to commit so I go to fix the windows todo which I forgot and rebase it

library/PlugLoad-windows.cpp
suokko: How is windows dfhack loading done?
I'm wondering if GetModuleHandle(NULL) is safe in static variable initialization

library/VersionInfoFactory.cpp
```diff
@@ -133,18 +134,30 @@ void VersionInfoFactory::ParseVersion (TiXmlElement* entry, VersionInfo* mem)
...
             if ((is_vtable && no_vtables) || (!is_vtable && no_globals))
                 continue;
-#ifdef DFHACK64
-            uintptr_t addr = strtoull(cstr_value, 0, 0);
-#else
-            uintptr_t addr = strtol(cstr_value, 0, 0);
-#endif
+            uintptr_t addr;
+            if (cstr_value) {
+                if (sizeof(addr) == sizeof(unsigned long))
```
lethosor: This is bad on Windows because `sizeof(unsigned long) == 4` regardless of the architecture.

suokko: else branch uses strtoull if sizeof unsigned long doesn't match sizeof uintptr_t. That means 32bit unix and windows uses the true branch calling strtoul which returns 32bit integer. In 64bit unix it also takes true branch and calls strtoul but this time the function returns 64bit integer. Then 64bit windows takes else branch and calls stroull which also returns 64 bit integer.

It would be possible to write c++ template function for all stdto* functions which would just check signens and sizeof integer. But this code uses only two alternatives so I kept the check inline.

There appears to be conflicts. I can rebase this later today.

lethosor: Oh, right, strtoul(l) will return something matching sizeof(addr). Makes sense

PR #1331, opened 2018-06-30
Add dfhack.printall_recurse to quickly print df containers

suokko:
I often want to see multiple items quickly when trying to figure out
what states actually matter to an issue that I debug. I decided to make
it easier to quickly dump df structures with substructures and
containers. It will generate large amount of data which can be sometimes
slow to process manually. But processing can be automated using
dfhack-run lua ^<df data to inspect> and pipe to other tools (eg grep,
sed, perl, sort, uniq etc)

No changelog or documentation yet. I can add them when this looks acceptable. I still wonder if I should change some.

I'm not entirely happy how I formatted object keys.  I wanted to use format that is very close to printall(table) format. Basically only adding recursion to subjects and pointers. The original printing code used format that closely matched lua table definition where curly brackets helped key table printing. But as table keys is a very rare keys I decided not to try to improve it.

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -180,6 +180,116 @@ function printall_ipairs(table)
     end
 end
 
+local do_print_recurse
+
+local function print_string(printfn, v, seen, indent)
+    local str = tostring(v)
+    printfn(str)
+    return #str;
+end
+
+local fill_chars = {
+    __index = function(table, key, value)
+        local rv = string.rep(' ', 23 - key) .. ' = '
+        rawset(table, key, rv)
+        return rv
+    end,
+}
+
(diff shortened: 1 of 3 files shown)


library/lua/dfhack.lua
```diff
@@ -180,6 +180,115 @@ function printall_ipairs(table)
...
+    end,
+}
+
+setmetatable(fill_chars, fill_chars)
+
+local function print_fields(value, seen, indent, prefix)
+    local ok,f,t,k = pcall(pairs,value)
+    if not ok then
+        dfhack.print(prefix)
+        dfhack.println('<Type dosen\'t support iteration with pairs>')
```
lethosor: doesn't

PR #1335, opened 2018-07-01
Runtime debug prints (only filtering code and command line plugin to control it)

suokko:
I think I have managed to get the runtime debug print filtering code good enough that it can considered for merging. 

Important features:
* DebugCategory provides light impact filtering code for debug prints. Code actually uses wrapper macros TRACE, DEBUG, INFO, WARN and ERR to print the messages. Constructo registers the category and destructor removes it from the DebugManager.
* DebugManager provides management interface for the plugin to set filtering levels based on json configuration and user commands.
* plugin/debug.cpp implements debugfilter command to manage filters levels. It uses regular expression to allow single filter to match multiple alternatives easily. It also stores persistent filters into dfhack-config/
* Signal provides a generic signals and slots implementation. It can be helpful when implementing functionality where some events may have unknown number of handlers in runtime. First use case is DebugManager events. But it could be useful in c++ widget module or any similar event based interface. Design is influenced by boost/Signal2 but I choose a bit different data race quarentees and implementation.

I tried to keep comments up to date while my design evolved from simple implementation to proper data race safe and static initialization order safe version.

After this becomes good enough to be merged then I can work on adding some basic debug prints to core. Those can be then used as example by other to add more debugs when ever they have to debug something complex.

Example output:
23:24:55.081:t2:debug:filter: sig type: 2 core:debug
23:24:55.081:t2:core:debug: modify DebugCategory 'debug' from 'core' allowed 0
23:24:55.081:t2:debug:command: Save config to 'dfhack-config/runtime-debug.json'

Example config:
{
	"configVersion" : 1,
	"filters" : 
	{
		"1" : 
		{
			"categoryText_" : ".",
			"enabled_" : true,
			"level_" : 0,
			"pluginText_" : "."
		}
	},
	"nextId_" : 3
}


suokko:
shrieker> hmm. I could avoid the static initializer order fiasco with constexpr constructor. I forgot that completely. (https://akrzemi1.wordpress.com/2012/05/27/constant-initialization/)
shrieker> I guess I should use that to solve the fiasco instead of template class trick because constexpr constructor and separate class handling registering would be simpler

suokko:
I couldn't go to sleep while this was hanging in the air. I clearly hated my first solution using template class static functions to manage life time of objects.

The new patch switched initialization to use constexpr constructor. I only gave it minimal testing but gdb shows that now runtime constructors are generated for DebugCategory.

PR #1336, opened 2018-07-03
Main thread ownership for df ai

suokko:


BenLubar:
For those who weren't part of the IRC discussion, this is required for df-ai's lockstep mode to work after the changes in #1304.

df-ai's lockstep mode works like this:

1. Pause Dwarf Fortress game loop during `plugin_onupdate`
2. Override certain system functions that affect animations. Generally, this just means the "get current time" functions.
3. Replace the renderer with one that injects a reimplementation of the game loop into the render thread.
4. Send an asynchronous `zoom` command to the renderer, causing the render thread to run the code that was just injected.
5. Game loop is now running in the render thread. The old game loop thread gives stub responses to async requests.

To revert:

6. System function overrides are removed.
7. The renderer is changed back to the non-wrapped version.
8. If the plugin delayed a disable-plugin request or a game exit, it performs the action now.
9. The game loop resumes as if the entire time lockstep was running had been one long df-ai `plugin_onupdate` call.

In order for this to work, the `plugin_onupdate` call that gets interrupted cannot hold the lock on CoreSuspender while the game loop is running on the render thread, but it needs to be holding the lock before df-ai returns control to DFHack.

PR #1337, opened 2018-07-03
Allow canceling lineedit with ctrl+c

suokko:
@lethosor Is this closer to behavior you expect from lineedit?

Windows seems to not to support ctrl+c at the moment.

### DIFF
--- a/library/Console-posix.cpp
+++ b/library/Console-posix.cpp
@@ -423,13 +423,13 @@ namespace DFHack
                 int count;
                 if (enable_raw() == -1) return 0;
                 if(state == con_lineedit)
-                    return -1;
+                    return Console::FAILURE;
                 state = con_lineedit;
                 count = prompt_loop(lock,ch);
                 state = con_unclaimed;
                 disable_raw();
                 print("\n");
-                if(count != -1)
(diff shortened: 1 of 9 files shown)


lethosor:
Ctrl-c in liquids and tiletypes still exits to the DFHack prompt. Ctrl-C in Lua gives a new Lua prompt, though. I'm not sure why that would happen - from reading the code, it looks like the -3 return value should be passed through Console::lineedit properly, but I don't have a lot of time to investigate right now.

suokko:
It should have worked for liquids and tiletypes (at least they worked for me). But I noticed other bugs:
* console returning strings and RETRY or SHUTDOWN.
* The dfhack core prompt not handling RETRY.
* Too many magic numbers. I decided to change return values to constants.
* A few more cases which probably should support RETRY.

lethosor:
I still have the same issue, and will try to investigate. +1 on the constants, though.

suokko:
You could try setting break point on Console::lineedit return line (gdb "b Console-posix.cpp:941") Then I could see the returned string from the std::string passed as reference parameter and I could print the return value (or use finish to return and print the value).

PR #1338, opened 2018-07-03
Add documentation for plugin proto files

BenLubar:
This documentation is intended for automated tools (such as [dfhackrpc](https://github.com/BenLubar/dfhackrpc)) to be able to parse.

I also removed the execute bit from a few files that didn't make sense having it set and renamed the `proto` argument to `plugin` in the RPC client implementation to better reflect its purpose.


lethosor:
So this isn't actually documentation in the DFHack repo, then? Looks like pretty minor changes to me, so I have no problem merging this, although the comments in RFR files might be easy to forget to keep up-to-date.

lethosor:
You know more about this than I do

BenLubar:
I followed the same format that the other proto files were already using, and it let me generate this: https://godoc.org/github.com/BenLubar/dfhackrpc/dfproto/plugins/remotefortressreader#CallCheckHashes

lethosor:
Yeah, there were some other instances of this in .proto files already. @JapaMala, any ideas on a better place? Otherwise, might as well merge this.

JapaMala:
Yeah, merge it.

PR #1341, opened 2018-07-04
Fix memory return calling convention in ruby bindings

### LINKED ISSUE #1209: Ruby codegen can't handle vmethods that return std::string's
enabler::GetKeyDisplay() is new in https://github.com/DFHack/df-structures/commit/559c23baa06f68d62a0ba674669389273bfcfb97#diff-cee8bfd3ccf90f6603c9dc081d54d943

This produces a `vmethod unkret DFHack.vmethod_call(self, 0, InterfaceKey.int(arg0))` message in ruby/codegen.pl, and it looks like the generated function calls `rb_dfget_rtti_classname`, which can crash. The output below is from 64-bit OS X 0.44.03; note that the returned value for "2" is 0x101ee35c0, which is the address of `enabler`.
```
[DFHack]# rb puts df.enabler.GetKeyDisplay 2
4327355840
[DFHack]# rb puts df.enabler.GetKeyDisplay 3
(segfault)
```

suokko:
Calling convention for memory return is that caller allocates (in stack)
memory to hold returned object. Then caller passes the pointer to callee
as implicit first parameter.

references:
https://github.com/hjl-tools/x86-psABI/wiki/X86-psABI
https://msdn.microsoft.com/en-us/library/7572ztz4.aspx

Fixes #1209

### DIFF
--- a/plugins/ruby/codegen.pl
+++ b/plugins/ruby/codegen.pl
@@ -462,8 +462,13 @@ sub render_class_vmethods {
             push @lines_rb, "def $name(" . join(', ', @argnames) . ')';
             indent_rb {
                 my $args = join('', map { ", $_" } @argargs);
-                my $call = "DFHack.vmethod_call(self, $voff$args)";
                 my $ret = $meth->findnodes('child::ret-type')->[0];
+                my $call;
+                if (!$ret or $ret->getAttribute('ld:meta') ne 'primitive') {
+                    $call = "DFHack.vmethod_call(self, $voff$args)";
+                } else {
+                    $call = "DFHack.vmethod_call_mem_return(self, $voff, rv$args)";
+                }
                 render_class_vmethod_ret($call, $ret);
             };
             push @lines_rb, 'end';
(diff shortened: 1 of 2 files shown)


plugins/ruby/codegen.pl
```diff
@@ -534,6 +539,18 @@ sub render_class_vmethod_ret {
         };
         push @lines_rb, "end._at(ptr) if ptr != 0";
     }
+    elsif ($retmeta eq 'primitive')
```
BenLubar: is this condition inverted?

BenLubar: is stl-string a primitive? I would assume no, but I might be wrong.

suokko:
lower-1.xslt has following:
         <prim-type ld:meta='primitive' ld:subtype='stl-string'/>

plugins/ruby/codegen.pl
lethosor: I didn't think stl-strings were considered primitives in df-structures, at least, but I could be wrong about that.

suokko: lower-1.xslt defines stl-string and stl-fstream to be primitive

lethosor: Oops, I was looking at the diff and didn't see your other comment (below here).

PR #1343, opened 2018-07-07
Remove Core.h include from DataDefs.h

suokko:
Idea is to avoid rebuilding statics when rebuild isn't required.

This depends on https://github.com/DFHack/df-structures/pull/286


BenLubar:
@suokko can you rebase this and/or the structures version? I added support for building PRs that require other PRs, but it's failing due to linked lists being used on one but not the other.

PR #1344, opened 2018-07-07
Tsan race warning fixes

suokko:
Make lua data race free 

Fixes tsan trace report between lua viewscreen and other threads running
lua without CoreSuspender lock. But I would assume similar races exists
when using lua from console thread, remote thread and vmethods same time.

Fix data race between threaded init and EventManager

The initial run_dfhack_init loads shared state information that is used
by EventManager when state changes. There is a small risk that
EventManager can handle events while run_dfhack_init is still running.

Make ServerMain and ServerConnection data race free

RemoteServer and PluginManager side would need complete redesign to be
data race free and concurrent. But as that would be unlikely to be
required from DFHack I decided simpler solution that is fixing data
ownership to a thread and all ServerConnection share a single lock which
allows access to PluginManager and Core.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,7 @@ if(WIN32)
   endif()
 endif()
 
+option(EXTERNAL_LIBSTDCXX "macOS only: Avoid installing the DFHack-provided libstdc++." OFF)
 if(APPLE)
   # libstdc++ (GCC 4.8.5 for OS X 10.6)
   # fixes crash-on-unwind bug in DF's libstdc++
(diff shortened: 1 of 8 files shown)


suokko:
I tested this now after rebase. Ingame labor selection is broken. I have no idea what broke it because my changes don't look like they should affect labor selection at all.

I will debug the issue and try to figure out where the issue is coming from

suokko:
Seems like issue was that autolabor was enabled when I expected it shouldn't be enabled.

lethosor:
Rebased onto develop. 6af912cb4 was the original commit, and 3a837d17 was my botched attempt at merging. Sorry for all the noise. This is back at 4 commits now, so I'll take a look and wait for CI to finish.

lethosor:
Getting this on startup, didn't run into it it on develop:
```
dyld: Symbol not found: __ZSt14__once_functor
  Referenced from: ./hack/libdfhack.dylib (which was built for Mac OS X 10.13)
  Expected in: /Users/alan/DF/0.44.12/hack/libstdc++.6.dylib
 in ./hack/libdfhack.dylib
./dfd: line 27: 91251 Segmentation fault: 11  DYLD_INSERT_LIBRARIES="$libs" ./dwarfort.exe "$@"
```

PR #1350, opened 2018-07-08
Use unique_ptr and shared_ptr for keeping member variables

sgayda2:


library/Core.cpp
```diff
@@ -1489,11 +1489,10 @@ Core::~Core()
     }
     if (d->iothread.joinable())
         con.shutdown();
-    delete d;
 }
 
 Core::Core() :
-    d{new Private},
+    d(new Private),
```
suokko: Even tough exceptions can't happen here. It is good practice to use make_unique&#60;T>(args...) when initializing unique_ptr. The template function makes sure that allocation is exception safe (not leaking in any scenario). There is forward compatible version in dts namespace. (that is c++14 feature)

library/VersionInfoFactory.cpp
```diff
@@ -51,31 +51,26 @@ VersionInfoFactory::~VersionInfoFactory()
...
 }
 
-VersionInfo * VersionInfoFactory::getVersionInfoByMD5(string hash)
+const VersionInfo * VersionInfoFactory::getVersionInfoByMD5(string hash)
 {
     for(size_t i = 0; i < versions.size();i++)
     {
         if(versions[i]->hasMD5(hash))
-            return versions[i];
+            return versions[i].get();
```
suokko: If you use shared_ptr you could return shared_ptr or shared_ptr reference here. Then callers need to just copy the shared_ptr instead of VersionInfo structure.

Alternative would be to use unique_ptr in versions vector. You can insert unique_ptr to container with versions.emplace_back(std::move(version));

I haven't studied the code enough to recommend either approach. But if vinfo is modified by caller then I suspect the later is better solution while in other cases copy of shared_ptr is cheaper than copy of VersionInfo.

library/Process-darwin.cpp
```diff
@@ -59,21 +59,17 @@ Process::Process(VersionInfoFactory * known_versions)
...
     memset(first_kb, 0, sizeof(first_kb));
     // get hash of the running DF process
     my_md5 = md5.getHashFromFile(real_path, length, (char *) first_kb);
     // create linux process, add it to the vector
-    VersionInfo * vinfo = known_versions->getVersionInfoByMD5(my_md5);
+    const VersionInfo * vinfo = known_versions->getVersionInfoByMD5(my_md5);
     if(vinfo)
     {
-        my_descriptor = new VersionInfo(*vinfo);
+        my_descriptor.reset(new VersionInfo(*vinfo));
```
suokko: Maybe `my_description = std::move(dts::make_unique<VersionInfo>(*vinfo)`

Alternative would be making my_descriptor shared_ptr and just assigning returned `shared_ptr<VersionInfo>&` to my_descriptor.

library/Core.cpp
sgayda2: I was unsure what version of C++ was available so i didnt want to use make_unique. But im glad you have an alternative within dts.

library/VersionInfoFactory.cpp
sgayda2: I do want to modify this further to remove some more of the pointers and new/delete calls but I wanted to do it one thing at a time.

library/Process-darwin.cpp
sgayda2: As mentioned in the other comment, I don't want to make too many changes to the code just yet. I did however use the make_unique call.

sgayda2:
Added a few more fixes

warmist:
Any reason for shared_ptrs everywhere? I thought core exists beyond lifetime of anything else, so you would not "share" parts of it thus pointers are very valid (just don't delete them). 

Unique_ptr is usually ok everywhere.

lethosor:
I think of the three modules that are still pointers, only Materials is still relevant/used - the others (Notes and Graphic) can probably be removed, but I'll double-check and do that after merging this.

library/VersionInfoFactory.cpp
```diff
@@ -52,33 +52,28 @@ VersionInfoFactory::~VersionInfoFactory()
...
-    for(size_t i = 0; i < versions.size();i++)
+    for (const auto& version : versions)
     {
-        if(versions[i]->hasMD5(hash))
-            return versions[i];
+        if(version->hasMD5(hash))
+            return version;
     }
-    return 0;
+    return nullptr;
```
lethosor: +1 here as well - we used to not be able to use `nullptr` because of GCC 4.5, but some plugins defined their own versions of it, which turned into a mess.

library/include/VersionInfo.h
```diff
@@ -169,5 +170,19 @@ namespace DFHack
         {
             return OS;
         };
+
+        void ValidateOS() {
```
lethosor: Nice, this is a better place - thanks for cleaning it up.

lethosor:
@warmist unique_ptrs can't be copied (only moved) - shared_ptrs are better here for things like VersionInfo where individual instances need to be dealt with in several places. Granted, raw pointers would work there, even though unique_ptrs wouldn't, assuming Core never frees the objects in question, but if it ever does, that could cause crashes later on.

PR #1351, opened 2018-07-08
Use std::call_once to ensure thread safety for windows sdl hooks

sgayda2:
…alize the SDL pointers

### DIFF
--- a/library/Hooks-windows.cpp
+++ b/library/Hooks-windows.cpp
@@ -26,6 +26,7 @@ distribution.
 
 #include <windows.h>
 #include <stdint.h>
+#include <mutex>
 #include <vector>
 #include <string>
 #include "Core.h"
@@ -40,8 +41,8 @@ distribution.
 
 // we don't know which of the SDL functions will be called first... so we
 // just catch the first one and init all our function pointers at that time
-bool FirstCall(void);
-bool inited = false;
+void InitSDLPointers(void);
+std::once_flag inited;
 
 /// wrappers for SDL 1.2 functions used in 40d16
 /***** Condition variables
@@ -612,8 +613,7 @@ DFhackCExport int SDL_NumJoysticks(void)
 static void (*_SDL_GL_SwapBuffers)(void) = 0;
 DFhackCExport void SDL_GL_SwapBuffers(void)
 {
-    if(!inited)
-        FirstCall();
+    InitSDLPointers()
     _SDL_GL_SwapBuffers();
 }
 
@@ -621,16 +621,14 @@ DFhackCExport void SDL_GL_SwapBuffers(void)
 static int (*_SDL_Flip)(void * some_ptr) = 0;
 DFhackCExport int SDL_Flip(void * some_ptr)
 {
-    if(!inited)
-        FirstCall();
+    InitSDLPointers()
     return _SDL_Flip(some_ptr);
 }
 
 static int (*_SDL_Init)(uint32_t flags) = 0;
 DFhackCExport int SDL_Init(uint32_t flags)
 {
-    if(!inited)
-        FirstCall();
+    InitSDLPointers()
     return _SDL_Init(flags);
 }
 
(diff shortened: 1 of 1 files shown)


library/Hooks-windows.cpp
```diff
@@ -612,25 +613,22 @@ DFhackCExport int SDL_NumJoysticks(void)
 static void (*_SDL_GL_SwapBuffers)(void) = 0;
 DFhackCExport void SDL_GL_SwapBuffers(void)
 {
-    if(!inited)
-        FirstCall();
+    InitSDLPointers()
```
suokko: semicolons seem to be missing.

library/Hooks-windows.cpp
```diff
@@ -40,8 +41,8 @@ distribution.
 
 // we don't know which of the SDL functions will be called first... so we
 // just catch the first one and init all our function pointers at that time
-bool FirstCall(void);
-bool inited = false;
+void InitSDLPointers(void);
+std::once_flag inited;
```
suokko: How about static linkage for variable and function?

sgayda2: done

library/Hooks-windows.cpp
sgayda2: i was compiling on linux and did not notice that this file was 'windows' specific until i had already pushed. Sorry about that.

library/Hooks-windows.cpp
```diff
@@ -640,110 +638,96 @@ MORE CRAP
...
 {
-    if(!inited)
-        FirstCall();
+    InitSDLPointers();
     return _SDL_strlcat(dst, src, maxlen);
 }
 
-// FIXME: this has to be thread-safe.
-bool FirstCall()
+void FirstCall()
```
lethosor: From https://buildmaster.local.lubar.me/executions/execution-details?executionId=70119#S104.3170

```
Z:/home/buildmaster/work/_E70119/_D0/src/library/Hooks-windows.cpp(740): error C2562: 'FirstCall': 'void' function returning a value
make[2]: *** [library/CMakeFiles/dfhack.dir/Hooks-windows.cpp.obj] Error 2
make[2]: *** Waiting for unfinished jobs....
Z:/home/buildmaster/work/_E70119/_D0/src/library/Hooks-windows.cpp(730): note: see declaration of 'FirstCall'
library/CMakeFiles/dfhack.dir/build.make:998: recipe for target 'library/CMakeFiles/dfhack.dir/Hooks-windows.cpp.obj' failed
```
I can't link to the problem, but it's "return 0" under the "if" block below. In theory, I'd like to keep the return values intact, but we're obviously not using them, and have never used them as far as I know (and nobody has reported the error below), so I guess replacing it with just "return" would work.

library/Hooks-windows.cpp
```diff
@@ -836,6 +820,9 @@ bool FirstCall()
     _SDL_EnableUNICODE(1);
 
     fprintf(stderr,"Initized HOOKS!\n");
-    inited = true;
-    return 1;
+}
+
+void InitSDLPointers()
+{
+    std::call_once(inited, [](){ FirstCall(); });
```
lethosor: Is there a reason for the lambda instead of just `call_once(inited, FirstCall)`? From https://en.cppreference.com/w/cpp/thread/call_once, the second parameter is a Callable, and I'd expect a function pointer to qualify for that, but I haven't tried it myself, so I could be wrong.

lethosor:
Out of curiosity, are you able to compile your PRs on Windows (or with MSVC some other way)? Now that we have Windows CI I'm not too concerned if you can't, but it's nice to know for reviewing purposes.

library/Hooks-windows.cpp
sgayda2: Since I can't compile on windows and since I have used a lambda before in this kind of case I just figured it would be easier, I have no strong reason for using one over the other.

library/Hooks-windows.cpp
sgayda2: Done

sgayda2:
I do not have a setup to compile on windows. I am mostly relying on catching it myself as well as the CI.

PR #1353, opened 2018-07-09
Exclude jsoncpp default CMake targets.

BenLubar:
Removes the lib and include directories from the DFHack package.

~~Removes the status message that lists the jsoncpp version number.~~


lethosor:
For reference: #1310

BenLubar:
Ah, I saw that it didn't show up when I ran `make` a second time, but the way I tested it didn't make any sense.

Issue #1354, opened 2018-07-09
Typing too much into the console causes DF to crash.


keupo:
Using the 0.44.12-alpha1 64-bit mac version.

Typing too long of a command, 32 characters, into the console causes dfhack and the game to crash.

The in-game command line from ctrl+shift+p will accept these longer commands and serves as a workaround.

lethosor:
I've never had this happen. Do you know what you were typing? I'm assuming 0.44.12, in any case.

lethosor:
```
DFHack version 0.44.12-alpha1 (release) on x86_64
Type in '?' or 'help' for general help, 'ls' to see all commands.
[DFHack]# 1234567890qwertyuiopasdfghjklzxcvbnm1234567890qwertyuiopasdfghjklzxcvbnm
1234567890qwertyuiopasdfghjklzxcvbnm1234567890qwertyuiopasdfghjklzxcvbnm is not a recognized command.
```
Can't reproduce this

lethosor:
Ok, looks like #1355 addresses this (I had to enter ~90 characters - your terminal window must be pretty small).

lethosor:
Fixed in #1355

PR #1355, opened 2018-07-09
Fix crash when editing lines that are exactly console width

suokko:
plen+cooked_cursor==cols => begin = -1 which is passed to substr. The
sign is incorrect as code should be removing a character from begin
instead of trying to add a character.


suokko:
Missing from the description and commit message:
Fixes #1354

PR #1357, opened 2018-07-10
Fix windows crash when _exit calls static destructors in dll unload

### LINKED ISSUE #1356: "die" crashing and/or taking a long time on Windows
The symptoms seem to vary quite a lot:
* Saiko Kila in the [forum thread](http://www.bay12forums.com/smf/index.php?topic=164123.msg7803045#msg7803045) reports that DFHack hangs
* clinodev (same thread) reports that it takes around 15 seconds to quit
* Quietust has had it crash, but only on 32-bit Windows (64-bit is fine)
* Japa has had occasional hangs, but not consistently
* BenLubar on IRC traced an exception to the Core destructor. I'm not sure why _exit is calling destructors - it shouldn't, as far as I know, and it hasn't been an issue before.
exception screenshot

suokko:
Fixes #1356

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - Fixed special characters in `command-prompt` and other non-console in-game outputs on Linux/macOS (in tools using ``df2console``)
 - `dwarfmonitor`, `manipulator`: fixed stress cutoffs
 - `startdwarf`: fixed on 64-bit Linux
+- `die`: Fixed windows crash in the exit handling
 
 ## API
 - Added to ``Units`` module:
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1163,7 +1163,7 @@ command_result Core::runCommand(color_ostream &con, const std::string &first_, v
         }
         else if (builtin == "die")
         {
-            _exit(666);
+            std::_Exit(666);
         }
         else if (builtin == "kill-lua")
         {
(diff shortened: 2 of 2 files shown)


docs/changelog.txt
```diff
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - Fixed special characters in `command-prompt` and other non-console in-game outputs on Linux/macOS (in tools using ``df2console``)
 - `dwarfmonitor`, `manipulator`: fixed stress cutoffs
 - `startdwarf`: fixed on 64-bit Linux
+- `die`: Fixed windows crash in the exit handling
```
lethosor: Single backticks in RST produce links, in case you were unaware. This requires an anchor named "die" in another file. I think there is one, but there isn't always for C++-implemented commands, so I'll double-check. Just something to keep in mind.

lethosor: Core.rst:
```rst
.. _die:

die
---
Instantly kills DF without saving.
```
looks good

suokko: I checked it exists by generating documentations (I have documentations enabled always to verify links which have been broken couple of times for me.)

lethosor: There's also docs/build.sh if you don't want to build docs every time that they change when you're working on other things (particularly scripts).

suokko: I would just forget it. It is better for me to have everything behind as few commands as possible. If I want to optimize my testing temporary then I prefer to use a specific build target to avoid reduce the incremental build time.

PR #1365, opened 2018-07-16
building_hacks.lua: fix custom animation tables

KlonZK:
move "local frames" from inside the block to outside, otherwise there will not be an animation.

### DIFF
--- a/plugins/lua/building-hacks.lua
+++ b/plugins/lua/building-hacks.lua
@@ -147,13 +147,13 @@ function registerBuilding(args)
     end
     --animations and connections next:
     local gears
+    local frames
 
     local frameLength
     local animate=args.animate
     if not auto_gears then
         gears=args.gears or {}
-        frameLength=1
-        local frames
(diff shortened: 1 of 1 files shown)


plugins/lua/building-hacks.lua
```diff
@@ -147,13 +147,13 @@ function registerBuilding(args)
...
     local gears
+    local frames
 
     local frameLength
     local animate=args.animate
     if not auto_gears then
         gears=args.gears or {}
-        frameLength=1
-        local frames
+        frameLength=1
```
lethosor: Travis was complaining because you have 8 extra spaces at the end of this line. It's easy to fix on my end, so I'll do that, but just something to keep in mind for next time.

lethosor: Fixing that also made the diff cleaner, incidentally: https://github.com/DFHack/dfhack/commit/61df5e52867f8ada817078853f673a090be1e024

PR #1367, opened 2018-07-18
Change CMakeSettings.json to use Ninja. Add DFHACK_INCLUDE_CORE option for easier development.

BenLubar:


BenLubar:
Fixed install path for DF not being the root folder and rebased.

CMakeLists.txt
```diff
@@ -493,6 +493,45 @@ endif()
...
+        IF(DFHACK_BUILD_32)
+            STRING(APPEND DF_CORE_FILENAME "32")
+        ENDIF()
+        STRING(APPEND DF_CORE_FILENAME ".zip")
+    ENDIF()
+    SET(DF_CORE_URL "http://bay12games.com/dwarves/${DF_CORE_FILENAME}")
+    IF (NOT EXISTS "${CMAKE_BINARY_DIR}/${DF_CORE_FILENAME}")
+        FILE(DOWNLOAD "${DF_CORE_URL}" "${CMAKE_BINARY_DIR}/${DF_CORE_FILENAME}" SHOW_PROGRESS)
+    ENDIF()
+    FILE(REMOVE_RECURSE "${CMAKE_BINARY_DIR}/df-core")
```
lethosor: This will run whenever CMake runs - I assume this isn't intended for personal use, then? What is it intended for? As far as I know, DFHack only depends on a couple files in the DF folder at most, and they don't change often.

BenLubar: The idea is that DFHack would install the correct version of DF automatically for easier development.

Is there a way it could extract the archive only if it's not already extracted from the same version?

lethosor: It would be difficult, and would probably rely on Toady updating the correct files. Or you could write your own file containing the version number and check that.
Still, clearing the DF folder would happen on new releases. It's not meant for people to play with, so what's it meant for?

BenLubar: This isn't meant to be used with `make package`, but rather with `make install`. It removes the need for the DFHack dev to update their installation of Dwarf Fortress manually. For example, in Visual Studio, it's stored in `C:\Users\Owner\CMakeBuilds\f59d9b5c-4f13-7f38-a962-38d2a89cfda6\install\MSVC 64 Debug` for me.

lethosor: And why do you need to `make install` if you're not going to be using the resulting build? Does it help with packaging somehow?

lethosor: Oh, ok, so df-core is cleared sometimes but the installation directory isn't, and that's what's useful. Makes sense now, thanks.

PR #1368, opened 2018-07-18
Ignore CMakeLists.custom.txt and create it if it does not exist during cmake.

BenLubar:


BenLubar:
The .gitignore method also doesn't suffer from wiping out your changes if you `git checkout .`

PR #1372, opened 2018-07-22
Add an identity to dfhack_viewscreen and dfhack_lua_viewscreen.

BenLubar:


library/modules/Screen.cpp
```diff
@@ -918,3 +918,31 @@ df::plant *dfhack_lua_viewscreen::getSelectedPlant()
...
 }
+
+#define STATIC_FIELDS_GROUP
+#include "../DataStaticsFields.cpp"
+
+using df::identity_traits;
+
+#define CUR_STRUCT dfhack_viewscreen
+static const struct_field_info dfhack_viewscreen_fields[] = {
+    { METHOD(OBJ_METHOD, is_lua_screen), 0, 0 },
```
lethosor: Are these useful in Lua or are they just necessary for struct_identity instances?

BenLubar: They're for the virtual_identity macros. I could also just not define any fields, but I thought I'd be "complete".

lethosor: Cool, that's what I thought. I think that virtual_identity::find won't end up using these due to the mangled name not matching, but it's good to have in the future.

Issue #1366, opened 2018-07-16
(feature request) building_hacks: add support for furnaces


KlonZK:
the lua frontend seems to be ready for furnaces already but the c++ part is not.

warmist:
IIRC you need to reimplement everything [here](https://github.com/DFHack/dfhack/blob/master/plugins/building-hacks.cpp#L67) to furnace. Probably by moving code out to be shared for both.

warmist:
Some more info after looking into the headers: you can't have furnaces connected to machines as they don't have machine info. So i'm not sure if it's even worth doing this?

KlonZK:
so it is entirely impossible to do, or can it be done by handling the machine stuff directly?

warmist:
Nothing is impossible per se. Just very very difficult.

In this case almost everything can be done easily (getMachineInfo could return some part of our memory - custom tracked or in some evil casting nightmare) but the hard part is how machines connect to each other (which needs work either way - it still has order of operations bug). The vmethod that is used for that is `canConnectToMachine` and i think in furnace case it might be not the same (i.e. return always false) as workshop.

lethosor:
Adding machine info to furnaces is probably more trouble than it's worth, at least for now. It might not get saved properly even if we can construct it.

myk002:
In-progress work available in https://github.com/DFHack/dfhack/pull/1376

PR #1376, opened 2018-07-29
WIP adding furnaces to building hacks

warmist:
This (some day will) implements #1366

### DIFF
--- a/plugins/building-hacks.cpp
+++ b/plugins/building-hacks.cpp
@@ -11,6 +11,7 @@
 
 #include "df/building_doorst.h"
 #include "df/building_workshopst.h"
+#include "df/building_furnacest.h"
 #include "df/machine.h"
 #include "df/machine_tile_set.h"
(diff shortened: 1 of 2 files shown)


lethosor:
How in-progress is this?

warmist:
As noted in #1366 this lacks the support for machine in furnaces. Adding such thing would be very hard. Features that are left might be valuable. Also as is noted in code - there could be use from adding furnace specific features (afaik only melted metal fractions is unique to furnaces).

Today i've commited last part that was missing. So what is left is testing (can't remember if i've tested this enough) and doc updates (probably to the new inlined docs)

lethosor:
Yeah, partial support for furnaces is better than none. Doc updates would be great. Feel free to merge when it's ready and tested.

myk002:
I'll close this PR for now, but the related issue is still there and keeps a reference to this PR when we look at it again

PR #1380, opened 2018-07-31
(labormanager) only assign valid labors (fix #1142)

### LINKED ISSUE #1142: labormanager shouldn't assign intelligent, grasping pets to labors
Partially because they aren't able to be assigned labors any other way (!!) but also because they tend to hold on to axes and picks after being done with their job.

ab9rf:
Add a number of tests to ensure that only assignable labors are assigned
to units, and that only units that are eligible to be assigned labors
are considered at all.

I borrowed logic from Dwarf Manipulator to identify units eligible to be
assigned labors.

This should prevent pets., ghosts, visitors, or any other noneligible
entities from being assigned labors by labormanager, and should thus fix
#1142.

### DIFF
--- a/plugins/labormanager/labormanager.cpp
+++ b/plugins/labormanager/labormanager.cpp
@@ -945,6 +945,15 @@ private:
     {
         if (labor >= 0 && labor <= ENUM_LAST_ITEM(unit_labor))
         {
+            if (!Units::isValidLabor(dwarf->dwarf, labor))
+            {
+                debug("WARN(labormanager): Attempted to %s dwarf %s with ineligible labor %s\n",
+                    value ? "set" : "unset",
+                    dwarf->dwarf->name.first_name.c_str(),
+                    ENUM_KEY_STR(unit_labor, labor).c_str());
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


plugins/labormanager/labormanager.cpp
```diff
@@ -1967,13 +1998,14 @@ class AutoLaborManager {
...
                     continue;
 
-                set_labor(*d, l,
-                    (l >= df::unit_labor::HAUL_STONE && l <= df::unit_labor::HAUL_ANIMALS) ||
-                    l == df::unit_labor::CLEAN ||
-                    l == df::unit_labor::HAUL_WATER ||
-                    l == df::unit_labor::REMOVE_CONSTRUCTION ||
-                    l == df::unit_labor::PULL_LEVER ||
-                    l == df::unit_labor::HAUL_TRADE);
+                if (Units::isValidLabor((*d)->dwarf, l))
```
lethosor: FYI, you could probably use a range-based for loop here to avoid the double-dereference (`for (auto d : available_dwarfs)`).

PR #1382, opened 2018-08-01
(labormanager) respect NATURE value

ab9rf:
Units with values regarding NATURE will be preferenced or dispreferenced
for jobs which might give them a good/bad feeling because of that value

### DIFF
--- a/plugins/labormanager/labormanager.cpp
+++ b/plugins/labormanager/labormanager.cpp
@@ -70,6 +70,7 @@
 #include <df/personality_facet_type.h>
 #include <df/cultural_identity.h>
 #include <df/ethic_type.h>
+#include <df/value_type.h>
 
 #include "labormanager.h"
 #include "joblabormapper.h"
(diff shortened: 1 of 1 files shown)


plugins/labormanager/labormanager.cpp
```diff
@@ -1518,6 +1519,26 @@ class AutoLaborManager {
...
+
+        // Favor/disfavor BUTCHER (covers slaughtering), HAUL_ANIMALS (covers caging), and CUTWOOD based on NATURE value
+
+        if (labor == df::unit_labor::BUTCHER || labor == df::unit_labor::HAUL_ANIMALS || labor == df::unit_labor::CUTWOOD)
+        {
+            int nature = 0;
+            for (auto i = d->dwarf->status.current_soul->personality.values.begin();
+                i != d->dwarf->status.current_soul->personality.values.end();
+                i++)
+            {
```
lethosor: Another possible range-based-for use case, a la #1380
(I know this plugin predates our move to C++11-supporting compilers, so it's not really an issue.)

PR #1383, opened 2018-08-05
Embark assistant

PatrikLundell:
changelog.txt presumably requires manual merging.

### DIFF
--- a/plugins/embark-assistant/embark-assistant.cpp
+++ b/plugins/embark-assistant/embark-assistant.cpp
@@ -143,7 +143,7 @@ DFhackCExport command_result plugin_init (color_ostream &out, std::vector <Plugi
         "  the embark rectangle as well as normally undisplayed sites in the\n"
         "  current embark region. It also has a site selection tool with more\n"
         "  options than DF's vanilla search tool. For detailed help invoke the\n"
-        "  in game info screen. Requires 46 lines to display properly.\n"
+        "  in game info screen. Prefers 46 lines to display properly.\n"
     ));
     return CR_OK;
 }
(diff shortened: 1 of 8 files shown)


plugins/embark-assistant/embark-assistant.cpp
```diff
@@ -143,7 +143,7 @@ DFhackCExport command_result plugin_init (color_ostream &out, std::vector <Plugi
         "  the embark rectangle as well as normally undisplayed sites in the\n"
         "  current embark region. It also has a site selection tool with more\n"
         "  options than DF's vanilla search tool. For detailed help invoke the\n"
-        "  in game info screen. Requires 46 lines to display properly.\n"
+        "  in game info screen. Prefers 46 lines to display properly.\n"
```
lethosor: Did you change this requirement earlier? I'm not seeing changes that look like this in this PR. Might be worth adding to the changelog, in any case.

PatrikLundell:
The reason I changed "requires" to "prefers" is that since scrolling of the conditions has been implemented, it's possible to use a shallower window (but I don't quite know how much, as I don't know how large the resource list for the embark can become [and it can grow further if DF is modded]). I think that change was made in the previous version (I can't find it in the changes either), but I didn't think of changing the help text at the time. Hard to remember when things were done, as a fair number of months have passed since the changes were made...

lethosor:
Okay, thanks. This looks good to me.

lethosor:
Minor nitpick: the changelog entries should go under "misc improvements". "Fixes" is traditionally just bugfixes. I've fixed that on my end; no need to worry about it.

PR #1386, opened 2018-08-10
labormanager: add option to disable the management of a labor.

billw2012:
Also switching to case insensitive labor name matching.


billw2012:
How about this? I have been using it for a few hours and it seems fine, although I never noticed any obvious issues before this change (I don't really know what to look for).

PR #1387, opened 2018-08-11
labormanager: make labor names case-insensitive

ab9rf:
This is semi-shamelessly picked out of @billw2012's changes in PR #1386. While the changes proposed there to manage a subset of labors are not ready to commit for the reasons I set forth on the PR, there is no reason not to implement case-insensitive labor specifications on the command line.

### DIFF
--- a/plugins/labormanager/labormanager.cpp
+++ b/plugins/labormanager/labormanager.cpp
@@ -390,13 +390,13 @@ struct labor_info
     int idle_dwarfs;
     int busy_dwarfs;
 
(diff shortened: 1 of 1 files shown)


lethosor:
I merged #1386 since it was approved, and I'm not seeing anything significant in here that isn't present in 3e2c898e1a93e0c04a1a341e531a192f53357312 (there are minor differences in lookup_labor_by_name, but those should only have an effect when there are labor names that vary only in case, which isn't the case). I'm closing this, but let me know if I missed anything important.

Issue #1392, opened 2018-08-15
Proposal: Move biome type inference engine from embark-assistant to main library


ab9rf:
There is no straightforward way to determine the biome at a particular map tile square in fort mode, but it is possible. @ragundo worked out an inference algorithm for exportmaps in 2016: [see here](https://github.com/ragundo/exportmaps/blob/master/cpp/df_utils/biome_type.cpp). Ragundo's code was imported into dfhack as part of embark-assistant, at https://github.com/DFHack/dfhack/blob/master/plugins/embark-assistant/biome_type.cpp.

I would like to move the (fairly complex) logic in embark-assistant into one of the general library modules, probably Maps. In addition to being used by embark-assistant, I need this logic to make autofarm work correctly (at present, it does a crappy job of assigning seeds to above-ground farms because it can't tell what plants can be planted where). It's also used by df-ai. Rather than replicate the same code across multiple plugins, it makes more sense to move this generic function to a library.

Comments?

lethosor:
Improving code reuse is good. I think the Maps module mostly deals with player-level maps (fort/adv mode), but I guess I can't really think of a better place for world-level map utilities.

ragundo:
Je,Je
That code was a direct translation from IDA.

lethosor:
Closing since it looks like #1394 addresses this fully. Please reopen if I'm mistaken.

PR #1394, opened 2018-08-16
Add function to infer biome type to Maps module

ab9rf:
This moves code intended to infer biome type currently living in a couple of plugins into the Maps module, so that this code can be shared more easily by multiple plugins, as discussed in #1392.

I heavily refactored the existing code, as it was fuggly as hell. In limited testing, it appears to work, but it's possible that I made mistakes. Since this is new functionality that nothing currently relies on, adding it will not break anything.


library/modules/Maps.cpp
```diff
@@ -711,3 +712,398 @@ bool Maps::canStepBetween(df::coord pos1, df::coord pos2)
...
+            if (wdata->world_height == 17)
+                v6 *= 16;
+            else if (wdata->world_height == 33)
+                v6 *= 8;
+            else if (wdata->world_height == 65)
+                v6 *= 4;
+            else if (wdata->world_height == 129)
+                v6 *= 2;
+
+            is_possible_tropical_area_by_latitude = v6 > 170;
```
lethosor:
```
/home/buildmaster/work/_E70955/_D0/src/library/modules/Maps.cpp:799:56: warning: 'v6' may be used uninitialized in this function [-Wmaybe-uninitialized]
```
Is "no poles" an option?

This looks like it was reverse-engineered, but it'd be nice to get rid of the warning anyway.

ab9rf: It is possible to get a world with no poles. That possibility is covered by the prior code block; this code will never execute for a no-pole world.  But I'll force v6 to zero just to silence the warning.

And, yes, this is reverse-engineered code. I refactored it so it didn't look quite as fugly.

PR #1395, opened 2018-08-16
move nestboxes out of devel

ab9rf:
Nestboxes has been stable for years, no reason for it to remain in devel.

I did give it a nonderpy description, though.


plugins/CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ if (BUILD_SUPPORTED)
     DFHACK_PLUGIN(misery misery.cpp)
     DFHACK_PLUGIN(mode mode.cpp)
     DFHACK_PLUGIN(mousequery mousequery.cpp)
+	DFHACK_PLUGIN(nestboxes nestboxes.cpp)
```
lethosor: image

lethosor:
I had issues writing a similar plugin where setting the forbidden flag didn't cancel existing egg hauling jobs. Have you tested for that? It might be less common due to your update loop running more often (maybe that could be checked for performance impact...), but it could still be an issue if a dwarf gets assigned a job within 5 ticks of the eggs being laid.

We have Jobs::removeJob() and df::job_handler::cancel_job() now - the former might not work here (haven't tested the latter, but I think removeJob didn't work for me last time).

ab9rf:
I've never made any effort to cancel hauling jobs. The short scan interval makes it unlikely to happen.

The performance impact should be fairly minimal; it scans the building list rather than the item list, and the building list is usually not that large. I've never noticed a frame rate drop from enabling the plugin. It could probably be improved by using `buildings.other[buildings_other_id::NEST_BOX]` instead of `buildings.all`.

PR #1400, opened 2018-08-26
labormanager: add option to use a specified labor as an ignore marker.

billw2012:
Any labor can be specified for use as an ignore marker. When specified, any dwarf with that labor enabled will be treated as always busy using the same logic as burrows or military activation. This is easier to manage than burrows for people using Dwarf Therapist, and makes combining automatic and manual job assignments easier (in combination with Dwarf Therapist and the disable labor option in #1386).

### DIFF
--- a/plugins/labormanager/labormanager.cpp
+++ b/plugins/labormanager/labormanager.cpp
@@ -868,6 +868,10 @@ DFhackCExport command_result plugin_init(color_ostream &out, std::vector <Plugin
         "    Change the assignment priority of a labor (default is 100)\n"
         "  labormanager reset <labor>\n"
         "    Return a labor to the default handling.\n"
+        "  labormanager ignoremarker <labor>\n"
+        "    Ignore any dwarves that have this labor enabled.\n"
+        "  labormanager ignoremarker none\n"
+        "    Disable the ignore marker.\n"
         "  labormanager reset-all\n"
         "    Return all labors to the default handling.\n"
         "  labormanager list\n"
@@ -2368,6 +2372,7 @@ command_result labormanager(color_ostream &out, std::vector <std::string> & para
             return CR_FAILURE;
         }
 
+
         bool need_comma = 0;
         for (int i = 0; i < NUM_STATE; i++)
         {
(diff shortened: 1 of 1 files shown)


lethosor:
If ival(n) defaults to -1 (like the corresponding language_name fields do) then I'm pretty sure it should work.

billw2012:
> I don't see anything in what you've provided to prevent labormanager from setting the marker labor on one of the dwarfs it is managing.
True, if only there were some system that allowed one to disable a labor from being managed entirely :) I made the changes you requested in the other PR that does provide this functionality, if/when that goes in I can piggy back it to make setting a disablemarker imply disabling that labor from management. Until then I will hold off on this, or I will be making conflicting changes.

> handle the situation where there is a persisted configuration that was saved by a prior version of this plugin.
Yeah I thought I had confirmed that the data was always initialized but a second look gives a different impression. So I could bump the minor number of the existing storage and copy the known values over, or I could just use different storage (although I guess it is more polite to be optimal with storage, I don't know how much there is?). Which would you prefer?

ab9rf:
@BenLubar and I were talking yesterday about redoing the way persistent storage is implemented (see #955). I'm not sure what the best approach is. However, I think @lethosor is probably correct about uninitialized ivals defaulting to -1, which is what NONE is mapped to, so that will probably be ok. You still need to address excluding the "marker labor" from being managed.

Be aware that I am seriously reconsidering a major refactoring of all three automatic labor managers (autolabor, autohauler, and labormanager). All three share a ton of common code, and someone should only run at most one of them at a time, so combining the three of them into a single framework, with selectable submodules, is likely the trajectory we're heading on here.

plugins/labormanager/labormanager.cpp
```diff
@@ -2281,6 +2307,26 @@ command_result labormanager(color_ostream &out, std::vector <std::string> & para
...
+    {
+        if (!enable_labormanager)
+        {
+            out << "Error: The plugin is not enabled." << endl;
+            return CR_FAILURE;
+        }
+
+        df::unit_labor labor = lookup_labor_by_name(parameters[1]);
+
+        if (labor == df::unit_labor::NONE)
```
lethosor: You're also using NONE to mean "not set", so this makes it impossible to actually unset the marker. Can you change that? A special case for non-case-sensitive "NONE" might be enough... not sure if the underlying data that `lookup_labor_by_name` uses makes it easy to check for invalid names.

missionz3r0:
@billw2012  Coming over here from the issue that was commented on. Has any thought been put into having a custom defined role be the indicator for the dwarf to not to be utilized, because using alchemy would break eventually once that is put in place. Though, that may be less of a concern because of things breaking anyways whenever DF updates.

myk002:
closing this as labormanager is being merged with autolabor, and the whole thing is due for a rewrite.

PR #1402, opened 2018-08-26
Add Persistence module.

BenLubar:
- Alter World to use Persistence instead of storing data in historical figures.
- Fake historical figures will be converted to the new format when a world is loaded.
- Added plugin_save and plugin_load functions to the plugin API.
- Made the weird int7/int28 code that was in the old persistence API slightly safer.


library/Core.cpp
```diff
@@ -2042,6 +2041,11 @@ void Core::doUpdate(color_ostream &out, bool first_update)
     // Execute per-frame handlers
     onUpdate(out);
 
+    if (df::global::ui->main.autosave_request || strict_virtual_cast<df::viewscreen_optionst>(screen))
```
lethosor: Are you sure the options screen is a good place to save persistent data? What if the user doesn't save the game?

pronvit: If user doesn't save, histfigs won't be saved either. However, with this, will it save on every tick when on that screen?

BenLubar: Then the data goes into `data/saves/current` and doesn't actually get saved. Is there a better way to detect hard saves?

Come to think of it, this won't save data when a game ends, so it probably should be a different screen. Would viewscreen_savegamest work? I don't remember why I didn't use that for df-ai.

lethosor: It's possible that data isn't in a safe state when viewscreen_savegamest is shown, although it could be safe if we hook in at the first tick. #955 hooks in to the general_ref_nemesis::write_file vmethod, which could run later, though.
Does viewscreen_savegamest display when abandoning a game (or when a fort fails)?

library/Core.cpp
```diff
@@ -2041,7 +2040,7 @@ void Core::doUpdate(color_ostream &out, bool first_update)
     // Execute per-frame handlers
     onUpdate(out);
 
-    if (df::global::ui->main.autosave_request || strict_virtual_cast<df::viewscreen_optionst>(screen))
+    if (df::global::ui->main.autosave_request || strict_virtual_cast<df::viewscreen_savegamest>(screen))
```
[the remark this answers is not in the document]
BenLubar: oh, whoops!

library/Core.cpp
ab9rf: Was a resolution ever reached on this point?

BenLubar: Yes. The current patch uses the first tick of `viewscreen_savegamest`.

https://github.com/DFHack/dfhack/pull/1402/files#diff-ec49c48896f67f205270e832ab410b1eR2043

library/Core.cpp
```diff
@@ -2042,6 +2040,11 @@ void Core::doUpdate(color_ostream &out, bool first_update)
     // Execute per-frame handlers
     onUpdate(out);
 
+    if (df::global::ui->main.autosave_request || (vs_changed && strict_virtual_cast<df::viewscreen_savegamest>(screen)))
+    {
+        doSave(out);
```
lethosor: Any chance we can make this only run once each time autosave_request is set? This assumes autosaves are synchronous, which won't necessarily be the case in future DF versions (see e.g. 0.40.01).

library/include/modules/Persistence.h
```diff
@@ -0,0 +1,232 @@
...
+        // Fills the vector with references to each persistent item.
+        DFHACK_EXPORT void getAll(std::vector<PersistentDataItem> &vec);
+        // Fills the vector with references to each persistent item with a key that is
+        // greater than or equal to "min" and less than "max".
+        DFHACK_EXPORT void getAllByKeyRange(std::vector<PersistentDataItem> &vec, const std::string &min, const std::string &max);
+        // Fills the vector with references to each persistent item with a key that is
+        // equal to the given key.
+        DFHACK_EXPORT void getAllByKey(std::vector<PersistentDataItem> &vec, const std::string &key);
+
+#if defined(__GNUC__) && __GNUC__ < 5
```
lethosor: Perhaps this could go in a utility file - I've seen a compiler workarounds header, but probably in an unmerged PR.

plugins/skeleton/skeleton.cpp
```diff
@@ -86,6 +89,25 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
 }
 */
 
+// If you need to save or load world-specific data, define these functions.
```
lethosor: Good to have these

lethosor:
Just to confirm, this replaces PRs #955 and #927?

library/PluginManager.cpp
```diff
@@ -197,6 +197,9 @@ Plugin::Plugin(Core * core, const std::string & path,
     plugin_rpcconnect = 0;
     plugin_enable = 0;
     plugin_is_enabled = 0;
+    plugin_save = 0;
+    plugin_load = 0;
```
lethosor: Perhaps plugin_load_game and plugin_save_game would be better - load could be confused with the "load" command

library/Core.cpp
```diff
@@ -2285,6 +2288,27 @@ void Core::onStateChange(color_ostream &out, state_change_event event)
...
+    }
+}
+
+void Core::doSave(color_ostream &out)
+{
+    plug_mgr->doSave(out);
+    Persistence::Internal::save();
+}
+
+void Core::doLoad(color_ostream &out)
```
lethosor: Maybe `doLoadGame`, similarly

library/Core.cpp
lethosor: Done in https://github.com/DFHack/dfhack/pull/1402/commits/100e584c3435edca39216550c05c5997a0ef401f

lethosor:
Tested some more with various actions: saving, loading without quitting, abandoning, entering/exiting/saving arena/adventurer mode, etc.. Haven't been able to cause any issues.

A basic Lua test snippet, for reference:
```lua
dfhack.persistent.save({key='a', ints={1,2,3}})
```

PR #1404, opened 2018-08-31
WIP libcurl with lua->json rest calls.

billw2012:
This adds libcurl for cross platform RESTful things. It also adds a few functions with lua bindings that allow lua objects to be sent as json to a url target.
For myself this is specifically to support sending analytics to external programs/services. Initially it was elasticsearch (which can be targeted directly with this code), and now I am working on a python program to run a custom frontend for DF analytics. 
More generally REST is a very common protocol and lowers the barrier to entry for any integrations with other programs or services.
This is marked WIP as I want to a) test the waters, b) test my code!
libcurl itself is a hugely popular and mature cross platform library, to it is my preference for REST support, however there are potentially some alternatives that include less code, however they are less popular, supported and mature.


.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "depends/jsoncpp"]
 	path = depends/jsoncpp-sub
 	url = ../../open-source-parsers/jsoncpp.git
+[submodule "depends/curl"]
+	path = depends/curl
+	url = ../curl.git
```
BenLubar: This should have `../../[GitHub username]/curl.git` so it doesn't try to load from whoever forks the DFHack repository's account.

billw2012:
What is going on here? My file looks like this:
https://github.com/billw2012/dfhack/blob/libcurl/library/modules/RestApi.cpp
The CodeFactor errors are relating to an old version of that file, why isn't it looking at the latest one? Or does it check each commit separately? If so how can I stop that? Make a new branch with only my latest?

lethosor:
Codefactor is not a required check, so ignore it if its output doesn't make sense. I'll look through this later.

billw2012:
Actually once it is passing I will resubmit from a fresh branch, and remove the vscode config files.

billw2012:
Cleaned up, and squashed to #1406.

PR #1414, opened 2018-12-02
Add some stuff to RFR, and some stonesense fixes.

JapaMala:


plugins/remotefortressreader/remotefortressreader.cpp
```diff
@@ -256,6 +255,252 @@ command_result dump_bp_mods(color_ostream &out, vector <string> & parameters)
...
+                            if(tile.bits.connection_east)// East Connection
+                            {
+                                if(tile.bits.connection_south) // South Connection
+                                {
+                                    if(tile.bits.connection_west) // West Connection
+                                    {
+                                        if(tile.bits.connection_north) // North Connection
+                                            {
+                                                if(tile.bits.trunk)
+                                                    output << "╬";
```
lethosor: I would prefer escapes here (worried about this not working with some compilers or for some developers, although that might be unfounded)

plugins/remotefortressreader/remotefortressreader.cpp
```diff
@@ -2793,6 +3078,11 @@ static command_result GetPartialCreatureRaws(color_ostream &stream, const ListRe
 
             CopyMat(send_tissue->mutable_material(), orig_tissue->mat_type, orig_tissue->mat_index);
         }
+
+        for (int i = 0; i <= ENUM_LAST_ITEM(creature_raw_flags); i++)
```
lethosor: Should use FOR_ENUM_ITEMS or at least ENUM_FIRST_ITEM for consistency (FOR_ENUM_ITEMS would avoid the cast below)

JapaMala:
Huh, somehow didn't get notified of this review. I'll look into tings

plugins/remotefortressreader/df_version_int.h
```diff
@@ -1,2 +1,2 @@
 #pragma once
-#define DF_VERSION_INT 44002
+#define DF_VERSION_INT 44010
```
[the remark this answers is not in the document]
JapaMala: It's only used to block out code sections that I want to keep for backwards compatibility, so it doesn't get changed often.

JapaMala:
jsoncpp change isn't needed. honestly, I thought it was pulled in from the develop branch.

plugins/remotefortressreader/remotefortressreader.cpp
JapaMala: Solved it by getting rid of that function entirely.

JapaMala:
I can't figure out why Travis keeps failing.

plugins/Plugins.cmake
```diff
@@ -111,6 +111,7 @@ MACRO(DFHACK_PLUGIN)
     SET_SOURCE_FILES_PROPERTIES( ${PLUGIN_PROTO_HDRS} PROPERTIES HEADER_FILE_ONLY TRUE )
     LIST(APPEND PLUGIN_SOURCES ${PLUGIN_PROTO_HDRS})
     LIST(APPEND PLUGIN_SOURCES ${PLUGIN_PROTO_SRCS})
+    LIST(APPEND PLUGIN_SOURCES ${PLUGIN_PROTOS})
```
lethosor: Why? This is passed into ADD_LIBRARY below - wouldn't this try to compile .proto files?

lethosor:
Buildmaster has some warnings about comparing an int to `vec.size()`, by the way

plugins/Plugins.cmake
JapaMala: No, the compilers recognize them as not being source files, but this way they show up in the file list in visual studio as an extra file, which makes editing them together with the rest of the stuff much more convenient.

lethosor: Oh, right, like header files

PR #1432, opened 2019-02-09
Improve readme, dependencies, and gitignore

LordGolias:
This basically does small improvements to the build and maintenance of the rep.

See individual commits for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,10 @@ build/install_manifest.txt
 build/_CPack_Packages
 build/dfhack-*.zip
 build/dfhack-*.bz2
+build/*ninja*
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -353,6 +353,9 @@ if(UNIX)
         unset(ZLIB_LIBRARY CACHE)
         unset(CMAKE_HAVE_PTHREAD_H CACHE)
     endif()
+
+    # on windows, SDL dependency is not handled by cmake.
+    find_package(SDL REQUIRED)
```
lethosor: This should not be used on macOS

PR #1437, opened 2019-04-27
Auto clothing

JapaMala:
Allows the user to set a how many of each clothing type every citizen should have.


JapaMala:
I'm removing the wip tag from this because it's got everything I was hoping for in it. Still needs longer testing though.

lethosor:
@JapaMala adding this to the r3 list of things, and will test it some, although probably not as much as you did. Let me know if it's really in a state where it shouldn't be merged, though.

JapaMala:
It's pretty much in a usable state right now. Maybe some feedback on the commands could be useful.

ab9rf:
I have a similar project in https://github.com/ab9rf/dfhack/tree/tailor. I haven't had a chance to see if/how what you've done overlaps (or not) with what I'd done.

JapaMala:
Hm, the functionality is similar but different. Yours will check each dwarf to replace old and worn out clothing, and replace them. This will just make sure there's the same number of cloth pants as citizens in your fort.

plugins/autoclothing.cpp
```diff
@@ -0,0 +1,643 @@
...
+        stream << item_subtype << " ";
+        stream << material_category.whole << " ";
+        stream << needed_per_citizen;
+        return stream.str();
+    }
+
+    void Deserialize(std::string s)
+    {
+        stringstream stream(s);
+        stream >> (int16_t&)job_type;
```
lethosor: This is throwing a warning (see buildmaster). I personally don't think it should throw a warning, but it is. Maybe a better solution is defining a custom operator>> for job_type and item_type, although that should probably be done for all DF enums if we're going that route.

JapaMala: I changed it to serialize the enums to string instead of to numbers. It won't be as fast, but it's only when saving and loading, so it should be fine.

lethosor: Cool, thanks for that. Wish the original way had worked... might look into adding I/O operators anyway.

PR #1440, opened 2019-05-17
mousequery map boundary detection fixes

lethosor:
This changes mousequery to use the map dimensions returned from Gui::getDwarfmodeViewDims() to determine map boundaries rather than some complicated (and inconsistent) math with other fields. Granted, mousequery predates the map_x2/y2 fields. However, TWBT breaks the old checks because it modifies the result of getMousePos() to support different map tile sizes. In the case where the map tiles are larger than the surrounding tiles, mousequery would think that the mouse wasn't close to the bottom/right edges of the map when it actually was.

This hopefully fixes some edge detection issues reported on the forums, as well as potential similar issues with other features (tracking and right-button scrolling).

### DIFF
--- a/plugins/mousequery.cpp
+++ b/plugins/mousequery.cpp
@@ -572,12 +572,11 @@ struct mousequery_hook : public df::viewscreen_dwarfmodest
         static decltype(enabler->clock) last_t = 0;
 
         auto dims = Gui::getDwarfmodeViewDims();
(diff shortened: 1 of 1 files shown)


lethosor:
Pvt Pirate reports that this works on the forums, and I can't see any difference from the original behavior without TWBT, so I'm merging it.

PR #1441, opened 2019-05-21
Make unit_ishostile always return a non-nil value

janxious:
Have been messing with this function in the dfhack repl and when it returns nil it's hard to know what that means unless you have also already read the code. So this handles a couple spots in `unit_ishostile` that previously just returned false-ey `nil`s to return `false`

### DIFF
--- a/plugins/ruby/unit.rb
+++ b/plugins/ruby/unit.rb
@@ -134,7 +134,7 @@ module DFHack
         # does not include ghosts / wildlife
         def unit_ishostile(u)
             # return true if u.flags3.ghostly and not u.flags1.inactive
-            return unless unit_category(u) == :Others
+            return false unless unit_category(u) == :Others
 
             case unit_other_category(u)
             when :Berserk, :Undead, :Hostile, :Invader, :Underworld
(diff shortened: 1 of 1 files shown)


lethosor:
Not sure why Travis is failing (besides the develop branch check). Can you try rebasing your branch on the current DFHack develop branch? I'm thinking yours might be based on something too old to run properly on Travis.

janxious:
@lethosor done. sorry. I have now read the contribution guidelines (after making the PR) so I think that messed me up.

lethosor:
No worries. We ought to make the `develop` branch the default to make it more obvious. And changing the base here to develop should be enough, but Travis is weird and remembers the fact that the PR used to be based on `master`... interesting that the force-push resolved that issue.

Anyway, this looks good to me as long as you've tested it, which I assume you have.

janxious:
> Anyway, this looks good to me as long as you've tested it, which I assume you have.

Insofar as I knew how, which was travis passes and it worked for all the testcases I could produce in a fort with some hackery. Mostly had a bunch of undead I made and some goblin invaders. I didn't try it with other dwarf civs or elves or humans, so I could do that if you want.

lethosor:
Sounds good, thanks. (Travis doesn't actually test this part besides checking syntax of all .rb files.)

PR #1443, opened 2019-06-06
Embark assistant

PatrikLundell:


lethosor:
Are you making those commits locally? You can include more than one file in a commit if so.

PatrikLundell:
Yes, I do commit them locally. I wasn't aware that you could add multiple files to a commit. Also, that's more typing than recalling the previous command and changing the last part of the name, keeping the path...

lethosor:
You don't even have to specify filenames. Passing -a to `git commit` will include all files that have been committed before (or -p will let you select changes to commit). I'd prefer to have commits be logical changes so they can be inspected all at once, instead of jumping between commits for each file, but it's not a big enough issue to change here.

PatrikLundell:
Ah, using "-a" sounds like an improvement to my side as well. I'll try to remember it for the next time.

lethosor:
It's a bit difficult to word well. "git commit -a" will commit all files with changes, as long as they are already tracked in the repo (i.e. it won't commit any new files). If you try to commit a file that hasn't been changed, git will tell you there aren't any changes to commit and it won't do anything. It might make more sense if you run through "git add" and "git status" first.

PatrikLundell:
Thanks. I'll try to remember it to the next time.

build/win64/install-debug.bat
```diff
@@ -1,4 +1,4 @@
-call "%ProgramFiles(x86)%\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+call "D:\Program (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
```
lethosor: Can you revert the changes to this file?

lethosor: I'm guessing this was added because of my recommendation above. Something that would have worked would have been:
```sh
git add docs plugins
git commit
```
this would only add changes to the "docs" and "plugins" folder to the commit.

lethosor:
Some whitespace complaints from the linter in 8248832 (so you don't have to go digging through the logs):
```
./plugins/embark-assistant/survey.cpp: Contains trailing whitespace: line 1748
./plugins/embark-assistant/survey.h: Contains trailing whitespace: line 56
./plugins/embark-assistant/embark-assistant.cpp: Contains trailing whitespace: line 278
```

Don't pay too much attention to CodeFactor.

PatrikLundell:
I'll try to restore the file contents and deal with the white space.

So I take it just 'git commit -m "xxx"' would take what's added, and there'd be no need to mention the files again once they've been added.

lethosor:
Correct. I happened to have https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository handy, which goes into a decent amount of detail about the various options/subcommands for committing changes, if that's useful.

PR #1447, opened 2019-07-17
CMake cleanup

lethosor:
Some formatting consistency things at the moment. Will take a pass through and clean out some ancient/unused options later, maybe.

Also need to clean up *.cmake files.


lethosor:
The first commit should just be that (minus a couple irrelevant comments that got removed). I wasn't sure if macros were case-sensitive at first, but they're not, and everything seems to build fine. I've got some more to do, I think.

PR #1454, opened 2019-09-02
Change dwarfmonitor to use global lua state.

warmist:
This needs more CoreSuspends but they are needed either way. This way
you can access other script environments and access this plugin enviroment
from lua.


warmist:
Some thoughts and possible future improvements:

- After this change the internal lua class might be not that useful anymore but i've left it as is
- There might be merit to just hooking main screens (i.e. dwarf mode, adventure mode, arena mode) and calling some lua function that does the drawing
- In that case this could have functionality of "collect and monitor data" while other plugins would supply other data and custom lua would display it e.g. in margins of screens.

PR #1456, opened 2019-09-23
Added 'fileresult' optional parameter, improved world traversal

PatrikLundell:
- The optional 'fileresult' parameter causes the plugin to read the saved search profile, perform a search (twice) and then write the number of matching world tiles to the file <DF>/data/init/embark_assistant_fileresult.txt. This file was deleted when the plugin was started with this parameter.
Intended usage is with an external harness (no relation to the plugin at all) that generates worlds, searches them, and selects the ones with matches.
- Improved world tile traversal logic to speed up the second search onward by skipping moving to some tiles if not needed.

### DIFF
--- a/plugins/embark-assistant/embark-assistant.cpp
+++ b/plugins/embark-assistant/embark-assistant.cpp
@@ -253,6 +253,13 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
 
 command_result embark_assistant(color_ostream &out, std::vector <std::string> & parameters)
 {
+    bool fileresult = false;
+
+    if (parameters.size() == 1 &&
+        parameters[0] == "fileresult") {
+        remove(fileresult_file_name);
+        fileresult = true;
+    } else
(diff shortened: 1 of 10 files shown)


PatrikLundell:
No comment added to changelog.txt as it's a bug fix within the current request.

plugins/embark-assistant/embark-assistant.cpp
```diff
@@ -253,6 +253,13 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
 
 command_result embark_assistant(color_ostream &out, std::vector <std::string> & parameters)
 {
+    bool fileresult = false;
+
+    if (parameters.size() == 1 &&
+        parameters[0] == "fileresult") {
+        remove(fileresult_file_name);
+        fileresult = true;
+    } else
```
lethosor: Style: `else if` should be on the same line
I'll fix this, but it's easy to misinterpret the next line as a standalone `if`

PR #1458, opened 2019-09-30
lua/widgets filtered list input fix

warmist:
FilteredList was not consuming input events.


warmist:
Note/Todo: check there scripts:
* gui/workflow
* gui/gm-editor
* gui/autogems

PR #1463, opened 2019-10-06
Output Castes on Error

jimcarreer:
These values are actually pretty difficult to determine, since they're iterated anyway for the check, we might as well collect them and output them when the user enters an invalid value.

### DIFF
--- a/plugins/createitem.cpp
+++ b/plugins/createitem.cpp
@@ -311,12 +311,14 @@ command_result df_createitem (color_ostream &out, vector <string> & parameters)
 
         for (size_t i = 0; i < world->raws.creatures.all.size(); i++)
         {
+            string castes = "";
             df::creature_raw *creature = world->raws.creatures.all[i];
             if (creature->creature_id == tokens[0])
             {
                 for (size_t j = 0; j < creature->caste.size(); j++)
                 {
                     df::caste_raw *caste = creature->caste[j];
+                    castes += " "+creature->caste[j]->caste_id;
                     if (creature->caste[j]->caste_id == tokens[1])
                     {
                         mat_type = i;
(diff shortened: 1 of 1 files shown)


jimcarreer:
@lethosor I'm not really setup to build + test this but its pretty straight forward (I am interested in maybe contributing more).  Anyway, I've updated the wiki as well to give an example for pond turtles, I've tried this command with 'MUSSEL' though and I'm starting to suspect some animals _have no castes_ to input here, if that's the case I might have a bigger fix.

lethosor:
Forgot to reply after I added it to the project - looks good! Glad you figured out why CI wasn't passing. My only comment would be a stylistic one (I prefer spaces around binary operators, i.e. " + " instead of "+", but no need to change that). "In progress" just refers to the column in https://github.com/orgs/DFHack/projects/7 - it'll move to "Done" once it's merged.

https://dfhack.readthedocs.io/en/stable/docs/Compile.html has build instructions - they're a bit long, but let us know if you have questions (on Bay12, IRC, etc.).

lethosor:
As for mussels, they appear to have a DEFAULT caste, at least for a world generated in 0.44.05:
```
[lua]# ~world.raws.creatures.all[537].name
<string[]: 0x119022008>
0                      	 = mussel
1                      	 = mussels
2                      	 = mussel
[lua]# ~world.raws.creatures.all[537].caste[0]
<caste_raw: 0x1108cc000>
caste_id               	 = DEFAULT
caste_name             	 = <string[]: 0x1108cc008>
...
```

jimcarreer:
@lethosor Thanks! I'll give this a shot after work, are there any bugs / issues in the backlog that might be good for me to tackle / you'd really like taken care of?

lethosor:
I was able to create some mussel eggs with `createitem EGG MUSSEL:DEFAULT`. I also improved it so that leaving out the caste entirely displays the same help. Looks good!

There are a bunch of open bugs, if you're interested - not particularly organized other than that, but there's a "good first issue" tag for some that should be easier to tackle.

PR #1464, opened 2019-10-24
export-datatypes command (Ghidra import)

jimbowhales:
Code for importing df-structures into Ghidra. It'd be nice base for importing into any other external tool too.

I only just now got it to work for the first time, and haven't really even tested it, so I assume there are major oversights. It's definitely got rough edges, but I'm fed up with working with it, so it's not WIP either. I don't really care if this is merged, and it'd be just dead weight for most users anyway. Just letting you know it exists.

I hope something pretty much exactly like this didn't already exist that I just failed to find with a search engine.

This also needs changes to df-structures code: https://github.com/DFHack/df-structures/pull/327


ab9rf:
I started to work on something to do this a while ago, but haven't looped back to it for a while. (My code also totally sucks.) When I get a chance, I'll be glad to take a look at this, because I've been wanting this functionality for quite some time as well.

warmist:
@ab9rf : he mentions that "but I'm fed up with working with it," so not sure how useful/supported this is but looking over it would be appreciated - including fixing build issues.

Also for future tool integration, maybe generating directly from xml would be easier as you would not need to generate code that emits things that are already in xml. I'm all for adding more info into c++ side however it tends to bloat the compile time and size (probably not much though?).

ab9rf:
The code I have in my embryonic tool is in Python and reads the XML directly. Not sure that it makes any sense at all for this to incorporated into the core application when it can just as readily be an auxiliary tool.

warmist:
> 
> 
> The code I have in my embryonic tool is in Python and reads the XML directly. Not sure that it makes any sense at all for this to incorporated into the core application when it can just as readily be an auxiliary tool.

Well there is one reason: somebody did all the work already :D

lethosor:
> reads the XML directly

Out of curiosity, are you reading the source XML files or codegen.out.xml?

ab9rf:
> > reads the XML directly
> 
> Out of curiosity, are you reading the source XML files or codegen.out.xml?

Directly from the `df.*.xml` files (as well as `symbols.xml`).

The ghidra python script is [here](https://gist.github.com/ab9rf/f3909401671e679cd03f059d93a26e8a); at the moment it only imports the global symbols; it parses the types (probably incorrectly, at that) but does not yet create structure definitions.

I didn't use `codegen.out.xml` because, at the time I started working on this, I didn't have a working DFhack build environment, and thus did not have a `codegen.out.xml` to work from.

BenLubar:
I've tied this PR to the structures one and vice versa in BuildMaster. Looks like the generated `static.fields-*.inc` files don't compile.

lethosor:
Do some of the more recent additions to df_misc, like https://github.com/DFHack/df_misc/blob/master/import_df_structures.java, obsolete this and https://github.com/DFHack/df-structures/pull/327?

myk002:
is this PR still viable/useful? If not, I might close it out to clear our PR list.

ab9rf:
I don't know if there's anything in this that isn't covered by dfhack/df_misc#16. I suspect `df_import_structures.java` referenced in that PR may yet need some small adjustments (I seem to recall that it didn't set alignment correctly on some structure) but it was mostly working last I checked.

myk002:
Ack. Ok, I'll close this PR. The branch still exists if we find we need to reopen.

PR #1468, opened 2019-11-18
Autofarm script reimplemented correctly in C++ 

### LINKED ISSUE #1144: autofarm (script) does not seem to run in background
Presumably there was a change somewhere in the API for this that prevents this script from retriggering itself.

### LINKED ISSUE #1421: Autofarm is not managing barley
It looks like autofarm is not managing or seeing barley at all. In fact, it disables farms that I have manually set to grow barley. Setting a threshold manually does not help. Looking into it, it seems to be because `df.ui.tasks.discovered_plants` is returning false for barley, though I'm not sure why. This is with 0.44.12-r2 on Win10 64 bit. Can anyone reproduce? I can provide a save file if needed.

ab9rf:
This reimplements autofarm in C++, using the biome inference module that was added in July to properly determine crop assignments on surface farms.

This replaces the Ruby autofarm script which does not work correctly (it never has) and which should be deleted; however, we can do that separately after this is approved. I find pull requests across multiple modules confusing.

Fixes #1144
Fixes #1421

### DIFF
--- a/plugins/devel/CMakeLists.txt
+++ b/plugins/devel/CMakeLists.txt
@@ -3,6 +3,7 @@ DFHACK_PLUGIN(vectors vectors.cpp)
 endif()
 
 ADD_DEFINITIONS(-DDEV_PLUGIN)
+DFHACK_PLUGIN(autofarm autofarm.cpp)
 DFHACK_PLUGIN(buildprobe buildprobe.cpp)
 DFHACK_PLUGIN(color-dfhack-text color-dfhack-text.cpp)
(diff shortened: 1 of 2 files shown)


ab9rf:
I would prefer not to try to make the method that CodeFactor is complaining about "less complex"; doing so would require splitting closely-coupled code segments from one another, which would make the code actually harder to follow than it is currently.

lethosor:
CodeFactor isn't a required check; feel free to ignore it

PR #1475, opened 2019-11-27
Use Windows code page when processing console.

### LINKED ISSUE #1474: dfhack console on windows does not handle cp 65001 correctly
I have my console code page set to 65001 (UTF-8); however, the DFHack console on Windows appears to assume code page 1252 and thus does not properly display characters outside the Basic Latin block. These arise fairly frequently in DF names, and at least one plugin uses the gender markers ♂ and ♀. The characters in names typically do not render at all, while the gender markers render as a replacement character.

Possibly relatedly, trying to paste a UTF-8 character into the DFHack console when the console code page is set to 65001 causes the dfhack console to close, although without crashing DF itself.

BenLubar:
- Read input as UTF-16.
- Store history as UTF-8.
- Write output as current code page.
- Add helper functions for converting UTF-8 strings to console encoding.

Fixes #1474.

*Untested; submitting as a pull request because it's easier to compile this way.*

### DIFF
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -321,7 +321,45 @@ static uint16_t character_table[256] = {
     0xB0,   0x2219, 0xB7,   0x221A, 0x207F, 0xB2,   0x25A0, 0xA0
 };
 
-std::string DF2UTF(const std::string &in)
+#ifdef WIN32
+std::string UTF16ToUTF8(const std::wstring &in)
+{
+    // get buffer size in UTF-8
+    int size = WideCharToMultiByte(CP_UTF8, 0, &in[0], in.size(), nullptr, 0, nullptr, nullptr);
+
+    // convert to active codepage
+    std::string out(unsigned(size), 0);
+    WideCharToMultiByte(CP_UTF8, 0, &in[0], in.size(), &out[0], out.size(), nullptr, nullptr);
+
+    return out;
+}
+
+std::wstring UTF8ToUTF16(const std::string &in)
+{
+    // get buffer length
+    int size = MultiByteToWideChar(CP_UTF8, 0, &in[0], in.size(), nullptr, 0);
+
+    // convert to active codepage
+    std::wstring out(unsigned(size), 0);
+    MultiByteToWideChar(CP_UTF8, 0, &in[0], in.size(), &out[0], out.size());
+
+    return out;
+}
+
+std::string UTF16ToCurrentCodePage(const std::wstring &in)
+{
+    // get buffer size in current codepage
+    int size = WideCharToMultiByte(CP_THREAD_ACP, 0, &in[0], in.size(), nullptr, 0, nullptr, nullptr);
+
+    // convert to active codepage
+    std::string out(unsigned(size), 0);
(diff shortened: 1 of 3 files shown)


lethosor:
Did this get tested?

BenLubar:
I did not have a chance to test it yet. @ab9rf might have an easier time knowing what to test.

lethosor:
I think I'll hold off on merging this until we can verify that it works with both the default encoding and Unicode (I don't want to risk breaking the console encoding on Windows).

lethosor:
@quietust would you be able to test this on Windows if you get a chance? In addition to basic console functionality, it would be nice to try `dfhack.df2console()` from Lua with a CP437 string (e.g. a dwarf name with special characters)

quietust:
The console still seems to work on Windows 8.1, though in my case it's already using CP437 as its encoding so everything just works. As far as I'm aware, there's no way to change an individual console window's codepage unless cmd.exe is running inside it (so that you can run `CHCP`), aside from changing the entire system locale (which I'm not willing to do, since it requires a reboot and would likely break many of my existing applications).

quietust:
I just tried this (after cherry-picking it onto 0.47.04), and `dfhack.df2console()` is returning an empty string whenever the source string contains accented characters (in my case, I fed it `unit.name.first_name` from a dwarf named "ùshrir Sodelnil").

library/MiscUtils.cpp
```diff
@@ -380,8 +418,30 @@ std::string UTF2DF(const std::string &in)
...
+    return out.is_console() ? UTF2CONSOLE(in) : in;
+}
+
 DFHACK_EXPORT std::string DF2CONSOLE(const std::string &in)
 {
+#ifdef WIN32
+    // convert to UTF-16
+    std::wstring win(in.size(), 0);
+    for (unsigned i = 0; i < in.size(); i++)
+        win[i] = character_table[in[i]];
```
quietust, line 441: `in[i]` needs to be cast to `uint8_t`, otherwise extended characters will be negative.

library/MiscUtils.cpp
```diff
@@ -321,7 +321,45 @@ static uint16_t character_table[256] = {
...
+    std::wstring out(unsigned(size), 0);
+    MultiByteToWideChar(CP_UTF8, 0, &in[0], in.size(), &out[0], out.size());
+
+    return out;
+}
+
+std::string UTF16ToCurrentCodePage(const std::wstring &in)
+{
+    // get buffer size in current codepage
+    int size = WideCharToMultiByte(CP_THREAD_ACP, 0, &in[0], in.size(), nullptr, 0, nullptr, nullptr);
```
quietust, line 352: CP_THREAD_ACP isn't actually the correct codepage - it's "ANSI", while console windows typically use "OEM"

quietust, line 352: `CP_THREAD_ACP` isn't actually the correct codepage - it's "ANSI", while console windows typically use "OEM". Changing this (and the one below) to `CP_OEMCP` allows it to work correctly on Windows 8.1, though I don't know how it'll behave on Windows 10.

library/MiscUtils.cpp
quietust, line 441: `in[i]` needs to be cast to `uint8_t`, otherwise extended characters will be negative.

PR #1476, opened 2019-12-08
add Units::getPhysicalDescription

nornagon:
This returns the physical unit description seen on the unit screen, e.g.

> His very long sideburns are braided.  His very long moustache is arranged in double braids.  His very long beard is arranged in double braids.  His hair is clean-shaven.  He has a scratchy voice.  His free-lobed ears are splayed out.  His nose is narrow.  His lips are very thin.  His nose bridge is slightly convex.  His dark peach skin is slightly wrinkled.  His emerald eyes have slightly thin irises.  His hair is ecru with a touch of gray.

There's no easy way to reproduce this string directly from the body information, as there's a substantial amount of logic that goes into deciding what appears in this string and in what order.

Requires a new symbol in symbols.txt, which I've currently only found for macos x64 (0x100cbb890), but I'll find it on other platforms if this is plausibly mergeable! I found it by setting a memory-read breakpoint on a constant string that appears in that description (e.g. "incredibly short"), then jumping up the stack trace until I found this function.

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -541,20 +541,13 @@ string Units::getRaceName(df::unit* unit)
     return getRaceNameById(unit->race);
 }
 
-#ifdef _WIN32
-#define THISCALL __thiscall
-#else
-#define THISCALL
-#endif
-
-typedef void (THISCALL *df_unit_physical_description_fn)(df::unit*, string*);
 void df_unit_physical_description(df::unit* unit, string* out_str)
 {
-    static df_unit_physical_description_fn fn =
-        reinterpret_cast<df_unit_physical_description_fn>(
+    static auto* const fn =
+        reinterpret_cast<void(THISCALL *)(df::unit*, string*)>(
             Core::getInstance().vinfo->getAddress("unit_physical_description"));
-    if (fn)
-        fn(unit, out_str);
+    CHECK_NULL_POINTER(fn);
+    fn(unit, out_str);
 }
 
 string Units::getDescription(df::unit* unit)
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -541,7 +541,13 @@ string Units::getRaceName(df::unit* unit)
     return getRaceNameById(unit->race);
 }
 
-typedef void (*df_unit_physical_description_fn)(df::unit*, string*);
+#ifdef _WIN32
+#define THISCALL __thiscall
+#else
+#define THISCALL
+#endif
+
+typedef void (THISCALL *df_unit_physical_description_fn)(df::unit*, string*);
 void df_unit_physical_description(df::unit* unit, string* out_str)
 {
     static df_unit_physical_description_fn fn =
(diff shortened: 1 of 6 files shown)


quietust:
In case it wasn't already clear, this new symbol is for *code*, not data. The function appears to be a (non-virtual) method of the "unitst" class, and it takes 1 parameter aside from `this` (to take the string).

The prototype you provided will **not** work in the 32-bit Windows binary (whose address happens to be 0xDB87F0) - you must specify a calling convention of `__thiscall` to ensure that the first parameter ("this") is passed in the `ecx` register instead of on the stack.

nornagon:
Ah, good find. Thanks. I tested this on macos and windows x64 but not on any 32bit platforms.

I think this can be made to work using something like:

```cpp
class dummy_unitst {
  public:
    void physical_description(std::string *out_str);
};

// ...
std::string str;
reinterpret_cast<dummy_unitst*>(unit)->physical_description(&str);
```

I'll experiment with this soon and update the PR.

nornagon:
... oh wait, can we just write `__thiscall` on windows and it'll work?

lethosor:
I believe so - exportmaps and old versions of TWBT might have examples

nornagon:
Still just experimenting here, I'll try this out for real when I'm back at my windows machine.

lethosor:
Looks about right, but you might want to verify that _WIN32 isn't also defined on 64-bit Windows (I think WIN32 is, at least)

nornagon:
Per comment in the commit, __thiscall does nothing on x64 anyway.

library/modules/Units.cpp
```diff
@@ -541,6 +541,23 @@ string Units::getRaceName(df::unit* unit)
     return getRaceNameById(unit->race);
 }
 
+void df_unit_physical_description(df::unit* unit, string* out_str)
+{
+    static auto* const fn =
+        reinterpret_cast<void(THISCALL *)(df::unit*, string*)>(
+            Core::getInstance().vinfo->getAddress("unit_physical_description"));
+    CHECK_NULL_POINTER(fn);
```
lethosor: I would actually keep the do-nothing behavior that you had earlier if fn is null - when global addresses are missing, most other functions also fall back to some default behavior. Setting out_str to an empty string is also an option.

nornagon: done 👍

nornagon:
What do you think of calling the symbol something like
`unit::getPhysicalDescription(std::string*)`?
-- 

j

lethosor:
Special characters in symbol names might break things - they won't be valid code if they end up in generated code, and there might be some tools we have that can't handle that, so I'd avoid it

nornagon:
One more potential alternative, since there might in future be other function pointers in the symbols that aren't getters: `unit_get_physical_description`? i.e. put the class name first. Or perhaps even `unit__get_physical_description`, having `__` stand in for `::`?

lethosor:
Good point. I like `unit_get_physical_description` - it's consistent with some other "groups" of global addresses, like `ui_`, `standing_orders_`, etc.

Issue #1479, opened 2020-01-02
getplants will mark plants for harvest that don't have usable growths


Rekov3D:
If you use getplants to designate shrubs for harvest, it will do so even if there is nothing currently there to harvest. This occurs when the only usable components of the plants are growths.

Steps to reproduce:
1: Embark on a site with bilberrys
2: getplants -s BILBERRY
3: The dwarves will go around 'harvesting' bilberries. The shrubs will be destroyed. No bilberry fruits are actually collected.

lethosor:
I'm not entirely familiar with this, so to clarify - are the dwarves collecting growths (other than berries) or nothing at all? Does designating the bilberry bushes for gathering manually lead to the same result? If that's the case, I'm not sure if there's a way to tell in advance whether a plant will produce berries when harvested, but we can investigate.

RekovPDX:
No. If you manually designate the bilberry bushes with d - p they aren't selected unless there is a fruit there to collect. Similarly if you put these plants in a gather zone, dwarves will not attempt to harvest them unless there's a fruit or whatever. 

From the wiki: 
> Edible or otherwise usable growths should have [STOCKPILE_PLANT_GROWTH] in their material definitions for proper stockpiling. This also lets them be collected from plant gathering and farming jobs. 

I don't really know if they are 'collecting' the leaf/flower growths to be honest. Nothing is produced out of it either way.

I'm roughly trying to repeat [what Patrik Lundell told me on the forum.](http://www.bay12forums.com/smf/index.php?topic=175283.0) His explanation might be more helpful.

lethosor:
Aha, thanks for the forum link. So getplants is making designations that aren't even possible to make in vanilla, then. I'm not sure how complicated this is to address, but yeah, it's definitely a DFHack bug in that case.

PatrikLundell:
I've thought a bit about the problem, and came up with a few complications that will have to be considered.
The Bilberry case is reasonably straightforward, as the basic plant can't be used for anything, and thus there's no vanilla reason to pick a Bilberry at that stage. However:
- If the "structure" part of the Bilberry plant actually is collected, it's ought to be possible to create custom reactions that could use it (like Papyrus doesn't have any usefulness flags associated with it [at least none have been identified], but has to be detected through its paper making reaction).
[...]

PR #1481, opened 2020-01-15
Partial correction of issue 1479 and added verbose flag

PatrikLundell:
I've been unable to identify where DF stores information about actual growths on plants, and so the correction only removes plants where growths are out of season, but not those already gathered.

(#1479)

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -52,6 +52,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `embark-assistant`:
     - fixed bug causing crash on worlds without generated metals (as well as pruning vectors as originally intended).
     - fixed bug causing mineral matching to fail to cut off at the magma sea, reporting presence of things that aren't (like DF does currently).
+- `getplants`: fixed designation of plants out of season and added verbose flag, but failed to identify picked plants (which are still designated incorrectly)
 - `gui/autogems`: fixed error when no world is loaded
 - `gui/companion-order`:
     - fixed error when resetting group leaders
--- a/plugins/devel/CMakeLists.txt
+++ b/plugins/devel/CMakeLists.txt
@@ -1,6 +1,8 @@
 IF(UNIX)
 DFHACK_PLUGIN(vectors vectors.cpp)
-endif()
+ENDIF()
+
+INCLUDE(FindThreads)
 
 ADD_DEFINITIONS(-DDEV_PLUGIN)
 DFHACK_PLUGIN(buildprobe buildprobe.cpp)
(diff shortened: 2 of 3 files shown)


BenLubar:
@PatrikLundell can you rebase this? I think the merge conflict is confusing GitHub into not making the `refs/pull/1481/merge` pseudo-branch and that's causing BuildMaster to fail.

PatrikLundell:
@BenLubar Sorry, but I'm not sure I understand what you want me to do. My best guess is that it requires a "proper" reference to df-structures, as I just made the change quietust made after my download locally to make things compile (since df-structures isn't actually part of the project I thought it would be sufficient to have the correct contents). My second best guess is that changing the branch from "master" to "develop" after changing the code, but before making a new branch somehow caused git to sulk (and if that's the case copying the changed files, nuking dfhack, and download everything from scratch, changing things in the proper order, and then copy the saved files would probably be the only way I could achieve it, as I certainly don't know the commands to do it nicely).

lethosor:
image

Basically the changes you've made to changelog.txt here conflict with changes that were made on develop after you created this branch. If resolving those on the command line is tricky, you can just undo the change you made to changelog.txt, paste the line here, and I'll add it manually once I merge this. That should get rid of the conflict, at least.

lethosor:
That looks good to me - thanks for handling the merge + fixes! The error from the tests looks like it's `title-folder` which #1482 should address (but it's not something related to your code anyway).

PatrikLundell:
Yes, it was rather obvious complaints about TWBT had nothing to do with the changes.

Merging the changelog file changes was fairly easy, once I found where you'd "save" the changes (and you had to do it twice).

I belong to the obnoxious group of people who think that code should result in no warnings, or you'll eventually get so many of them that you won't see the cases where they're actually relevant. However, butchering the logic to satisfy an arbitrary "complexity" measurement is going too far (and I think we've touched on that one before).

BenLubar:
Yeah, it's safe to ignore test results from BuildMaster right now if you're not working directly on the DFHack unit test suite (which is really tiny at the moment)

lethosor:
> I belong to the obnoxious group of people who think that code should result in no warnings, or you'll eventually get so many of them that you won't see the cases where they're actually relevant. However, butchering the logic to satisfy an arbitrary "complexity" measurement is going too far (and I think we've touched on that one before).

Agreed, and that's why CodeFactor isn't a required check (it doesn't have to pass to merge this PR). It gives useful feedback sometimes, but the current feedback it's giving on this PR isn't that useful. I wish there were a way to make it less noisy, but I haven't figured one out yet.

PR #1482, opened 2020-01-15
Don't print an error about not being able to get the SDL title in text mode.

BenLubar:


lethosor:
(We could probably get this to work with the appropriate terminal escape sequences, actually, but this works)

PR #1484, opened 2020-01-22
Added switches for farm seeds and plant counts

PatrikLundell:
The file contains a comment at the beginning trying to explain the logic used and the source of that logic.

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -2298,16 +2298,19 @@ by spaces.
 
 Options:
 
-:``-t``: Select trees only (exclude shrubs)
-:``-s``: Select shrubs only (exclude trees)
-:``-c``: Clear designations instead of setting them
-:``-x``: Apply selected action to all plants except those specified (invert
+:``-t``: Tree: Select trees only (exclude shrubs)
+:``-s``: Shrub: Select shrubs only (exclude trees)
+:``-f``: Farming: Designate only shrubs that yield seeds for farming. Implies -s
+:``-c``: Clear: Clear designations instead of setting them
+:``-x``: eXcept: Apply selected action to all plants except those specified (invert
      selection)
-:``-a``: Select every type of plant (obeys ``-t``/``-s``)
-:``-v``: Lists the number of (un)designations per plant
+:``-a``: All: Select every type of plant (obeys ``-t``/``-s``/``-f``)
+:``-v``: Verbose: Lists the number of (un)designations per plant
+:``-n *``: Number: Designate up to * (an integer number) plants of each species
 
-Specifying both ``-t`` and ``-s`` will have no effect. If no plant IDs are specified,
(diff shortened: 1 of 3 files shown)


lethosor:
It would be nice to also have the docs in docs/Plugins.rst updated (but I know that requires Python and some other dependencies, so I can do that if you don't want to deal with those)

PatrikLundell:
I certainly don't know how to rebuild docs. However, I tried to install the Python and Sphinx stuff the "Compiling DFHack" document refers to, followed by docs\gen_changelog (from the dfhack directory), and nothing happened, i.e. the command did not return any feedback and "git status" only mentions the auto generated "depends/jsoncop-sub" file as modified.

It also seems to be a deficiency in the process to require generated documents to be generated for each change and tracked as changes that have to be merged: it should be sufficient to have the source under version control and let the (full) build process generate the resultant ones (as is done with C header files).

lethosor:
Documentation is built by Sphinx as part of the build process, if you set BUILD_DOCS to ON or 1 in your cmake config. gen_changelog.py only produces a .rst file in docs/_auto, which is not tracked in version control, but is included in the Sphinx build to generate the final (HTML) documentation.

> It also seems to be a deficiency in the process to require generated documents to be generated for each change and tracked as changes that have to be merged

This is not the case - I'm not sure what you're referring to. .rst files are the source files for the documentation (.html files are the output, and are definitely not tracked). Some .rst files are generated (e.g. by gen_changelog.py or from scripts) but those are also not tracked.

PatrikLundell:
Your comment misled me to believe the .rst file was generated from the changelog, give the reference to tools. However, if I understand it correctly, the file is the documentation (as you said, and I see when I look at it), and the tools are only needed to verify the file can actually be used for further generation. It OUGHT to be possible to just edit the file in a text editor, push it up, and have the build process spit out errors if it's incorrect, but I'll edit it and see if I can get generation to work.

docs/Plugins.rst
```diff
@@ -2317,7 +2320,13 @@ all valid plant IDs will be listed.
     This leads to some shrubs being designated when they shouldn't be, causing a
     plant gatherer to walk there and do nothing (except clearing the
     designation). See :issue:`1479` for details.
-
+    
+    The implementation another known deficiency: it's incapable of detecting that
+    RAW definitions that specify a seed extraction reaction for the structural part
+    but has no other use for it cannot actually yield any seeds, as the part is
+    never used (parts of bug tracker report 6940, e.g. Red Spinach), even though DF
```
lethosor, line 2327: For reference, similar to `:issue:` in the paragraph above (which links to DFHack issues on GitHub), `:bug:` links to Mantis issues. A full list can be found in [conf.py](https://github.com/DFHack/dfhack/blob/1546d9ec1a3c5673dfa49795bb8a956bf17c3b2d/conf.py#L199-L209)

PR #1485, opened 2020-01-25
Issue #1262. Added Items::getTitle and used in stocks

### LINKED ISSUE #1262: The search enhancement of the stocks screen and the enhanced stock view fail on tool books and some "books"
I've been trying to locate a particular codex/scroll/quire in my fortress, and found neither the search enhancement nor the enhanced stock view is able to find it. I think it's a scroll or quire, as searching seems to work fine with codices.

Try a fortress with a lot of books. Go into the tools category and locate one of those. Try to get the 's'earch function to shrink the list ('the' *should* result in a number of books, but you'll typically only find feather wood things).

I suspect it's caused by the wacky way quire and scrolls are "named", i.e. you have to follow the link to the contents to find the title, and I suspect neither of the enhancements do that, in which case they'd likely match on scroll and quire. Testing that shows that yes, "scr" provides a list of titles, presumably on scrolls, while "quire" does the same with quires (and a lot of blank ones). All of those are unsorted, of course.

Searching for "quire" with the enhanced stocks view gave 72 quires in one collection marked with a '*', which probably are the quires written on.

Edit: The work I was looking for turned out to be a codex, so those aren't found properly either. However, the enhanced view shows a number of titles (and searching on those narrows the enhanced view list down properly), so there seems to be some kind of category of written works that end up being handled properly. Yes. It's an item_bookst with a title, while most codices have a blank title.

PatrikLundell:
Items::getDescription only gets the title of codices that have their titles in their title field. However, some codices and all tool "books" have the title buried in their improvements. Items::getTitle returns the first title found, either from the title field or buried, or a blank string if no title is found (as is the case for non "books").
It can be noted that it might have been useful to include slab descriptions, but I didn't have access to a written one, so I don't know if it would be too lengthy. Also, it seems one of the unmapped "item" functions ought to provide the functionality of this operation.

The stocks plugin calls getTitle first and getDescription if no title was found. A few codices were found not to get any title, but further investigation showed their contents was actually untitled.

Another issue encountered was that several music instrument parts were missing from the stocks display, but investigation eventually showed the missing ones were marked with the Trader flag (while the displayed ones weren't), and thus were suppressed by design. I don't know why some of them were marked and some weren't, as they seemed to all be in the trade depot with the visiting caravan.
This observation is just that, an observation, not part of the issue at hand.

Fixes #1262

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -38,10 +38,10 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 # Future
 
 ## Fixes
-- `stocks`: fixed display of titles by using the new ``Items::`getTitle`` operation before the original one, issue 1262
+- `stocks`: fixed display of titles by using the new ``Items::getBookTitle`` operation before the original one
 
 ## API
-- Added ``Items::`getTitle`` to get titles of "books". Catches titles buried in improvements, unlike getDescription.
+- Added ``Items::getBookTitle`` to get titles of "books". Catches titles buried in improvements, unlike getDescription.
 
 # 0.44.12-r3
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -35,6 +35,14 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ======== IMPORTANT: rename this, and add a new "future" section, BEFORE ========
 ======== making a new DFHack release!                                   ========
 ================================================================================
+# Future
+
+## Fixes
+- `stocks`: fixed display of titles by using the new ``Items::`getTitle`` operation before the original one, issue 1262
+
+## API
+- Added ``Items::`getTitle`` to get titles of "books". Catches titles buried in improvements, unlike getDescription.
+
 # 0.44.12-r3
 
 ## New Plugins
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -688,7 +688,7 @@ static void addQuality(std::string &tmp, int quality)
 }
 
 //  It's not impossible the functionality of this operation is provided by one of the unmapped item functions.
-std::string Items::getTitle(df::item *item)
+std::string Items::getBookTitle(df::item *item)
 {
     CHECK_NULL_POINTER(item);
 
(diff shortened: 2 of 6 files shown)


lethosor:
Would `getBookTitle` be a better name, maybe, or are you anticipating that other things might also have titles?

PatrikLundell:
I hesitate to use the term "book" when the game has codices, scrolls, and quires, but not actually the term "book". However, I'm fine with such a change.

library/modules/Items.cpp
```diff
@@ -681,6 +687,81 @@ static void addQuality(std::string &tmp, int quality)
...
+//  It's not impossible the functionality of this operation is provided by one of the unmapped item functions.
+std::string Items::getTitle(df::item *item)
+{
+    CHECK_NULL_POINTER(item);
+
+    std::string tmp;
+
+    if (item->getType() == df::item_type::BOOK)
+    {
+        if (virtual_cast<df::item_bookst>(item)->title != "")
```
BenLubar: `virtual_cast` is a non-trivial function, so you should call it once and store the result in a local variable.

PatrikLundell:
Thanks. I'll change that.

library/modules/Items.cpp
BenLubar: you can also do `if (auto book = virtual_cast<df::item_bookst>(item))` as the first condition and then just use `book`

BenLubar:
> So virtual_cast won't blow up, but just return null if it fails?

Correct, `virtual_cast` is basically equivalent to `dynamic_cast` in normal C++ but it has some special stuff to work properly for DFHack where we don't know the vtable locations from the compiler (we get them from `symbols.xml`)

PatrikLundell:
Didn't quite follow BenLubar's advice, as I assume virtual_cast is still a fair bit more expensive than getType, so the operations were only invoked when the item have been determined to belong to the appropriate types. Given that pages and writings should dominate as "book" improvements, the lazier approach was taken there.

lethosor:
Note that there is a risk here when DFHack isn't up-to-date - we have to update the enum that getType() returns manually, but the vtable address updates (which virtual_cast uses) are automated, so they could become out of sync when DFHack is unstable. Your way works, though as does BenLubar's suggestion, and both are used in various places.

docs/changelog.txt
```diff
@@ -35,6 +35,14 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ======== IMPORTANT: rename this, and add a new "future" section, BEFORE ========
 ======== making a new DFHack release!                                   ========
 ================================================================================
+# Future
+
+## Fixes
+- `stocks`: fixed display of titles by using the new ``Items::`getTitle`` operation before the original one, issue 1262
```
lethosor: Should update this - and also, not sure what the backtick after the `::` is doing here

lethosor:
```suggestion
- `stocks`: fixed display of titles by using the new ``Items::getBookTitle`` operation before the original one
```

(we historically haven't put issues in here, and the changelog script probably won't handle links well... I suppose I could address that)

PR #1505, opened 2020-02-24
dwarfvet log only if enabled (fix #1504)

### LINKED ISSUE #1504: dwarfvet too verbose?
A plugin seems to be too verbose, I see this line a lot in the dfhack console:

```
Clearing all animal hospitals
```

I guess this line can be removed:
https://github.com/DFHack/dfhack/blob/546178bc56e6d70ae8a03d0801b1c0ab60ba39fe/plugins/dwarfvet.cpp#L417

pierredavidbelanger:
naive fix, just to get rid of the log :)


lethosor:
Thanks!

PR #1511, opened 2020-03-03
Updated embark-assistant. Lt aquifer, trees, neighbors

PatrikLundell:


plugins/embark-assistant/survey.cpp
```diff
@@ -2068,7 +2147,7 @@ void embark_assist::survey::survey_embark(embark_assist::defs::mid_level_tiles *
     bool incursion_processing_failed = false;
     df::world_data *world_data = world->world_data;
 
-    if (!use_cache) {  //  For some reason DF scrambles these values on world tile movements (at least in Lua...).
+    if (!use_cache) {  //  DF scrambles these values on world tile movements, while embark-tools stabilizes the movement, but its changes to the value are done after we've read them.
```
lethosor: You could probably run this code after embark-tools - if you're calling DF's render() or feed() after this code, calling it before might work. On the other hand, people can enable/disable embark-tools whenever they want, so handling this here is good.

PR #1533, opened 2020-04-02
Update embark-profile-name tweak

### LINKED ISSUE #1532: tweak embark-profile-name displays the wrong abort key
The embark-profile-name key shows shift-space to abort but this actually inserts a space into your embark profile name. The abort key is esc. This key should also be shown in red.

ymber:
This changes the embark-profile-name tweak to remove the undocumented abort key and reenable the abort key shown in game.

Fixes #1532

### DIFF
--- a/plugins/tweak/tweaks/embark-profile-name.h
+++ b/plugins/tweak/tweaks/embark-profile-name.h
@@ -13,6 +13,9 @@ struct embark_profile_name_hook : df::viewscreen_setupdwarfgamest {
         if( in_save_profile && ch >= 33 && ch <= 126 ) {
             profile_name.push_back( ( char )ch );
         } else {
+            if( input->count( df::interface_key::LEAVESCREEN ) ) {
+                input->insert( df::interface_key::SETUPGAME_SAVE_PROFILE_ABORT );
+            }
             INTERPOSE_NEXT( feed )( input );
         }
     }
--- a/plugins/tweak/tweaks/embark-profile-name.h
+++ b/plugins/tweak/tweaks/embark-profile-name.h
@@ -1,22 +1,21 @@
 #include "df/viewscreen_setupdwarfgamest.h"
-using namespace DFHack;
+
 struct embark_profile_name_hook : df::viewscreen_setupdwarfgamest {
     typedef df::viewscreen_setupdwarfgamest interpose_base;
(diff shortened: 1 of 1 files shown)


plugins/tweak/tweaks/embark-profile-name.h
```diff
@@ -1,22 +1,21 @@
 #include "df/viewscreen_setupdwarfgamest.h"
-using namespace DFHack;
+
 struct embark_profile_name_hook : df::viewscreen_setupdwarfgamest {
     typedef df::viewscreen_setupdwarfgamest interpose_base;
-    DEFINE_VMETHOD_INTERPOSE(void, feed, (std::set<df::interface_key> *input))
-    {
+
+    DEFINE_VMETHOD_INTERPOSE( void, feed, ( std::set<df::interface_key> *input ) ) {
```
lethosor, line 6: Could you maintain the style here? Specifically, we usually don't usually use spaces immediately inside parentheses, but do after `if`/`for`/etc. (e.g. `if (x)`). Per https://docs.dfhack.org/en/stable/Contributing.html, opening braces on their own line are acceptable (in fact, that used to be the *only* style mentioned there, so you'll see a lot of older DFHack code using it). I don't have a strong preference about the braces personally, but I'd rather keep what's already here.

plugins/tweak/tweaks/embark-profile-name.h
```diff
@@ -1,22 +1,21 @@
...
-            ch = Screen::keyToChar(*it);
-        if (in_save_profile && ch >= 32 && ch <= 126)
-        {
-            profile_name.push_back((char)ch);
+        for( auto it = input->begin(); ch == -1 && it != input->end(); ++it ) {
+            ch = Screen::keyToChar( *it );
         }
-        else
-        {
-            if (input->count(df::interface_key::LEAVESCREEN))
```
lethosor, line 16: I may not have been clear about this on IRC - while the esc key wasn't mentioned on-screen, I would prefer to keep it because it's a conventional key to leave a screen/dialog (so people will tend to press it), and it doesn't do anything in this particular dialog otherwise (so supporting it won't conflict with DF)

ymber:
Restored the esc abort key. Esc and shift-space both work now and all character inputs in the profile name work as before.

PR #1538, opened 2020-04-04
Add astyle config and check style in CI

### LINKED ISSUE #1535: Code style automation
DFHack has a code style specification and developing would be easier if things around that were automated. A config for astyle and CI checks to make sure C++ files are formatted would cover it.
```
# "One True Brace Style"
--style=1tbs

# 4 spaces per indentation level and per tab stop
--indent=spaces=4

# replaces tabs with spaces in non-indent sections, except in strings
--convert-tabs

# force LF line endings
--lineends=linux

# attach brackets to class and struct inline function definitions.
--attach-inlines

# 'char *foo' instead of 'char* foo'
--align-pointer=name

# maximum length for a single unbroken line
--max-code-length=100

# 'foo ||\nbar' instead of 'foo\n|| bar'
--break-after-logical

# indent 'public:' 'protected:' 'private:' deeper than 'class' or 'struct'
--indent-classes

# indent 'case' lines deeper than 'switch'
--indent-switches

# indent later lines of multi-line preprocessor directives
--indent-preproc-define

# indent comments on column 1 to match the code block they are in
--indent-col1-comments

# indent multi-line control block headers like 'if (foo\nbar)'
--min-conditional-indent=0

# add space around operators with two operands
--pad-oper

# add brackets to unbracketed one line conditional statements
--add-brackets

# remove extra space padding around parentheses
--unpad-paren

# insert space padding around parentheses on the inside only
--pad-paren-in
```
The first 4 options in there will make files meet the style guide in Contributing.rst except for sorting includes and file encoding. The rest aren't specified in the style guide but it's worth considering standardizing them.

CI checks for formatting could use astyle with `--dry-run` to check if any files would be changed.

ymber:
Adds a standard astyle config and a CI check for code style. The first 4 points of the config here will enforce current style requirements as specified in Contributing.rst. The rest of it is some things to consider standardizing that have worked well for us on the CDDA team. Anyone who's regularly developing here should check out how it fits their preferences.

Closes #1535

To do:
- [ ] Decide on any additions to standard project style for the config
- [ ] If any changes are made that current code doesn't meet add a file blacklist so CI doesn't always fail until standard style is met by all files
- [ ] Update any relevant documentation

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,0 +1,50 @@
+# "One True Brace Style"
+--style=1tbs
+
+# 4 spaces per indentation level and per tab stop
+--indent=spaces=4
+
+# replaces tabs with spaces in non-indent sections, except in strings
+--convert-tabs
+
+# force LF line endings
+--lineend=linux
+
+# attach brackets to class and struct inline function definitions.
+--attach-inlines
+
+# 'char *foo' instead of 'char* foo'
+--align-pointer=name
+
+# maximum length for a single unbroken line
+--max-code-length=100
+
+# 'foo ||\nbar' instead of 'foo\n|| bar'
+--break-after-logical
+
+# indent 'public:' 'protected:' 'private:' deeper than 'class' or 'struct'
+--indent-classes
+
+# indent 'case' lines deeper than 'switch'
+--indent-switches
+
+# indent later lines of multi-line preprocessor directives
+--indent-preproc-define
+
+# indent comments on column 1 to match the code block they are in
+--indent-col1-comments
+
+# indent multi-line control block headers like 'if (foo\nbar)'
+--min-conditional-indent=0
+
+# add space around operators with two operands
+--pad-oper
+
+# add brackets to unbracketed one line conditional statements
+--add-brackets
+
+# remove extra space padding around parentheses
+--unpad-paren
+
+# insert space padding around parentheses on the inside only
+--pad-paren-in
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ script:
 - sh travis/git-info.sh
 - sphinx-build -qW -j3 . docs/html
 - python travis/pr-check-base.py
+- travis/astyle-check.sh
 - python travis/lint.py
 - python travis/authors-rst.py
 - python travis/script-docs.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ addons:
       - libxml-libxslt-perl
       - ninja-build
       - zlib1g-dev
+      - astyle
 matrix:
   include:
     - env: GCC_VERSION=4.8
--- a/travis/astyle-check.sh
+++ b/travis/astyle-check.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+sources=$(find plugins library -type f -regextype posix-awk -regex ".*\.(h|cpp)")
+
+output=$(astyle --options=.astylerc --dry-run -Q $sources)
+
+if [ -n "$output" ]
+then
+    printf "Files modified by astyle\n$output"
+    exit 1
+fi
+exit 0


PatrikLundell:
The rigid CDDA "Artistic" style enforcement is utter garbage. It's "artistic" as in ugly, contrary to common conventions, in complete disregard for what's practical and useful, and undocumented apart from some feeble examples that sort of hints at what some of the rules are, enforcing usage of a specific command line tool (the installation instructions are out of date, relying on a magic population of the tool into a list where it isn't available, so command line usage to uglify the code is required, at least for Windoze).

lethosor:
I share some of your concerns about `astyle`'s ease of use. I think #1535 is still relevant (it was inspired by feedback on a PR whose style differed significantly from the rest of the DFHack codebase) so we could discuss alternatives there. I've kept this open to serve as a reminder to take a more thorough pass on this specifically tool+config (sorry I haven't yet) even if we don't ultimately settle on it.

PatrikLundell:
If some kind of style enforcement is implemented, it would have to:
- Focus on the important bits, not an obsessively detailed OCD level (that hits you with full force in the face when you modify anything as the existing code doesn't abide by it).
- Be compatible with the general style used in DFHack, rather than something completely different (which should also be compatible with how code editors generally format things, rather than the complete opposite of it).
- Be properly documented (every rules explained in a manner understandable by the average reader). If you're not prepared to explain it, don't mandate it!
- Provide extensive and useful feedback to inform the victim of exactly where and why it rejected the code with references to the corresponding rule(s).
- Have a set of rules the average contributor could learn with relative ease so they're not required to run some obscure black box "fixer" thing that might have done something (or not) to contribute anything.

ymber:
I just threw this together to show how styling would easily fit into the existing CI after we were talking about it on the IRC. Normally I use clang-format in my own stuff. You get more versatility out of it than astyle. If lundell cant read cata style code or operate the tools just use a different config with clang-format.

PatrikLundell:
You miss the point that if you mandate something you force everyone to abide by it, so I wouldn't have any option to format things in a sensible way if the CDDA rules were enforced.

Saying that if the tool installation instructions are incorrect you can then investigate how to use some other tool to do the enforcement you didn't want in the first place and then research what the rules enforced are and how to code them to fit that other tool just doesn't make any realistic sense.

It's not that I can't decode the code that results from what CDDA rules mandates, it's that I can't consistently write it in that eye hurting manner and the compiler automatically reformat it into a more readable (and thus rejected) format anyway, leading to a rejection whenever a line break is introduced at any time in the process. And, of course, that most of the rules are hidden so I don't even know what they are.

ymber:
You miss my point. This PR was demonstrating concept. I'm not invested in what formatting standard you use for your project or which tool it uses. I just found when I was writing my other PR last year that automated style checks and a formatting config for my editor would have caught things I had to fix sooner.

lethosor:
I'm definitely on board with _something_ for style formatting. Thanks for putting this up! We have since switched away from Travis, so it would need some modifications in any case, but I'm happy to handle that.

Actually coming up with formatter configs that match our existing code style will be somewhat complicated. I think it's important to minimize the burden for existing contributors (I think Patrik is getting at this too), both in terms of running the formatter and producing code that matches our style guidelines in the first place. I would generally like to have something that's easy to run locally on all of our target platforms, so if `astyle` has portable binaries, that could be a good option. I know we'll also get contributors who don't install formatters locally, so some way to correct formatting (through CI?) might also be useful (but that depends less on the exact formatter we use). In any case, it's something that will need to be documented and accessible (so probably on the Contributing page).

From my personal experience, I've found that running code through formatters locally tends to lower the effort required on my end in the long run, since I don't have to worry quite as much about fixing stylistic issues manually.

myk002:
closing this PR as we're moving towards pre-commit.ci-based solutions

PR #1539, opened 2020-04-04
Units update need coordination with structure identity

PatrikLundell:
This PR requires careful reviewing and probably modification, as well as coordination with DFHack/df-structures#389, where much of the background can be found.

When it comes to remotefortressreader I've done the minimum modification of just using the new field name. It probably should use the same logic as the corresponding part of Units.cpp

Units.cpp:
setNickname: Avoided some changes years ago to get nicknames to show where they ought to because of not understanding the references to demons. With the discovery of the enum, I think that refers to the Demon_Alias, and possibly Unk_1 cases, which means that we ought to be able to have nicknames show up in identities for the other identified cases. Unfortunately, I don't have a save to test it on.

isHidingCurse: just does the same as before, adapted to the number now being an enum value. Unless the new intelligent ghouls use their own enum value, the logic should still be valid.

getAge: the previous code just didn't make sense (which doesn't make sense). The check against a negative histfig id would never kick in in the worlds I checked (all positive), and only processing if there's a curse seems to be retained from before the introduction of false identities for non curse hiding purposes (0.42? 0.43?). The changed code returns a false age if there's one available, which seems straightforward. Again, I don't have a suitable world to test in.

It would be very useful if someone who know why the code looked like it did could review the changes.


BenLubar:
(if you ever need a structures and core PR linked, @ me)

PatrikLundell:
I just discovered exportlegends also uses the identity structure, and so it needs to be updated as well.

PatrikLundell:
As you can see, I had to update some things affected by structure changes made by @BenLubar a few hours ago. The DFHack/df-structures#389 connection also affects DFHack/scripts#144, but I don't think linking is necessary as @lethosor is aware of the connection, and the PRs won't have to be active for long.

library/modules/Units.cpp
```diff
@@ -722,12 +733,11 @@ double Units::getAge(df::unit *unit, bool true_age)
...
-    {
-        if (auto identity = getIdentity(unit))
-        {
-            if (identity->histfig_id < 0)
-                birth_time = identity->birth_year + identity->birth_second/year_ticks;
+    if (!true_age) {
+        if (auto identity = getIdentity(unit)) {
+            if (identity->birth_year != -1) {
+                birth_time = identity->birth_year + identity->birth_second / year_ticks;
+            }
```
lethosor, line 740: Thanks for cleaning this up

plugins/remotefortressreader/remotefortressreader.cpp
```diff
@@ -1693,20 +1693,7 @@ static command_result GetUnitListInside(color_ostream &stream, const BlockReques
...
-        {
-            if (auto identity = Units::getIdentity(unit))
-            {
-                if (identity->histfig_id < 0)
-                    birth_time = identity->birth_year * year_ticks + identity->birth_second;
-            }
-        }
-
-        send_unit->set_age(cur_time - birth_time);
+       send_unit->set_age(Units::getAge(unit, false));
```
lethosor, line 1696: Looks like a space got deleted here; I'll double-check locally and fix if needed

Issue #1436, opened 2019-04-16
Expose xml attributes (e.g. ref-target) to Lua


lethosor:
Making `ref-target` attributes available to Lua would make IDs easier to navigate in gui/gm-editor. The `comment` attribute might also be useful. I can't think of other attributes besides the Lisp ones, like `aux-value` (probably difficult to parse) and `instance-vector` (already accessible via `get_vector()`).

PatrikLundell:
Moving from issue #DFHack/df-structures/372:

Another attribute that would be useful to export would be "is-union", especially if paired with a "discriminant". If is-union's parameter can be changed to be the name of the field that determines which of the versions of the overlapping data that's applicable when known, and e.g. "nil" (or retain "true") when not known, it would be sufficient to export it as is.
If is-union has to have a boolean value, the introduction of a new attribute for that purpose would be useful.

If this is paired with the convention of ensuring each overlaying variant of the union data is a single field (by collecting groups in structs, etc.), and defining the overlaying fields in the same order as the discriminant field does, it would allow scripts (e.g. gui/gm-editor) to determine and use the appropriate variant when the discriminant is a simple 0-based enum/number field.

One technical way to export the attributes of interest would be to define them as explicit attributes connected to the appropriate fields of types, so e.g. df.incident.T_unk_v42_1 would get the attribute "is-union" with value="subtype" (and a number of structs with only padding would be defined to pad out the union to align the variant with the discriminant).

cvuchener:
Here is an union example (one I know because I use it in Dwarf Therapist): `unit_preference`
https://github.com/DFHack/df-structures/blob/0.47.03-beta1/df.units.xml#L2097-L2123

`type` tells you how to read the anonymous union. It is a nice enum starting at 0 with consecutive values, but the mapping is not one to one. `LikeMaterial` does not use the union, `LikeCreature` and `HateCreature` both use `creature_id`, `LikeFood` and `LikeItem` both use `item_type`, `LikePlant` and `LikeTree` both use `plant_id`.

Would you duplicate fields (with the same type and the same `ref-target`) to match the enum? There would also be the need for a special field meaning "not initialized" (like `std::monostate` is for `std::variant`).

PatrikLundell:
If I remember correctly XML schema can actually handle this, because it's got a union construct, but it's not used for DFHack, and some high level languages also supports it (high level assembly doesn't, of course).
Anyway, I'd work around the problem with:
´´´
<compound is-union='true'>
[...]

PR #1543, opened 2020-04-09
Add ref_target attribute to primitive field references

lethosor:
Still needs some cleanup to behave "properly" like the `value` field

See #1436


library/LuaTypes.cpp
```diff
@@ -706,6 +707,16 @@ static type_identity *find_primitive_field(lua_State *state, int field, const ch
...
+    const char *attr = lua_tostring(state, -1);
+    if (attr == std::string("ref_target")) {
+        const struct_field_info *field_info = get_object_ref_header(state, 1)->field_info;
+        if (field_info && field_info->extra && field_info->extra->ref_target) {
+            LookupInTable(state, field_info->extra->ref_target, &DFHACK_TYPEID_TABLE_TOKEN);
+        } else {
+            lua_pushnil(state);
+        }
+        return 1;
+    }
```
lethosor, line 719: This part in particular could use some cleaning up - one example is the special `whole` field for bitfields, defined [here](https://github.com/DFHack/dfhack/blob/0.47.04-beta1/library/LuaTypes.cpp#L1414) and implemented [here](https://github.com/DFHack/dfhack/blob/0.47.04-beta1/library/LuaTypes.cpp#L958-L965)

library/LuaTypes.cpp
```diff
@@ -706,6 +707,16 @@ static type_identity *find_primitive_field(lua_State *state, int field, const ch
  */
 static int meta_primitive_index(lua_State *state)
 {
+    const char *attr = lua_tostring(state, -1);
+    if (attr == std::string("ref_target")) {
```
BenLubar, line 711: since you're comparing two c strings here, you can just do `!strcmp(attr, "ref_target")` to avoid a string allocation

lethosor: I could indeed, but I was lazy and I expect to change this to not use string comparisons at all (like the bitfield-handling code I mentioned)

PR #1558, opened 2020-04-27
Implementation for df-other-vectors-type

BenLubar:
In C++. Still needs Lua and Ruby support.

See DFHack/df-structures#402.

### DIFF
--- a/library/LuaTypes.cpp
+++ b/library/LuaTypes.cpp
@@ -1422,6 +1422,28 @@ void struct_identity::build_metatable(lua_State *state)
     SetPtrMethods(state, base+1, base+2);
 }
 
+void other_vectors_identity::build_metatable(lua_State *state)
+{
+    int base = lua_gettop(state);
+    MakeFieldMetatable(state, this, meta_struct_index, meta_struct_newindex);
+
+    EnableMetaField(state, base+2, "_enum");
+
+    LookupInTable(state, index_enum, &DFHACK_TYPEID_TABLE_TOKEN);
+    lua_setfield(state, base+1, "_enum");
+
+    auto keys = &index_enum->getKeys()[-index_enum->getFirstItem()];
+
+    for (int64_t i = 0; i < index_enum->getLastItem(); i++)
+    {
+        lua_getfield(state, base+2, keys[i]);
+        lua_rawseti(state, base+2, int(i));
+    }
+
+    SetStructMethod(state, base+1, base+2, meta_struct_field_reference, "_field");
+    SetPtrMethods(state, base+1, base+2);
+}
+
 void global_identity::build_metatable(lua_State *state)
 {
     int base = lua_gettop(state);
--- a/library/include/BitArray.h
+++ b/library/include/BitArray.h
@@ -574,6 +574,8 @@ namespace DFHack
     {
         std::vector<I *> & operator[](O other_id)
         {
+            CHECK_INVALID_ARGUMENT(size_t(other_id) < sizeof(*this) / sizeof(std::vector<I *>));
+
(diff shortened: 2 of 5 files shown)


BenLubar:
This PR is ready for review. The structures PR isn't ready yet, and these changes won't make anything visibly change without the structures changes.

library/LuaTypes.cpp
```diff
@@ -1422,6 +1422,28 @@ void struct_identity::build_metatable(lua_State *state)
...
+{
+    int base = lua_gettop(state);
+    MakeFieldMetatable(state, this, meta_struct_index, meta_struct_newindex);
+
+    EnableMetaField(state, base+2, "_enum");
+
+    LookupInTable(state, index_enum, &DFHACK_TYPEID_TABLE_TOKEN);
+    lua_setfield(state, base+1, "_enum");
+
+    auto keys = &index_enum->getKeys()[-index_enum->getFirstItem()];
```
lethosor, line 1435: I'm not sure I follow the logic here - why is this negated? Wouldn't this end up being `1` for types with `NONE = -1`?

library/LuaTypes.cpp
```diff
@@ -1422,6 +1422,28 @@ void struct_identity::build_metatable(lua_State *state)
...
+
+    for (int64_t i = 0; i < index_enum->getLastItem(); i++)
```
lethosor, line 1437: This is probably fine since the only enums we care about have meaningful values starting at 0, but it may be worth specifying this in the XML (or generated metadata) somehow rather than hardcoding 0.

library/LuaTypes.cpp
lethosor: Ok, I think I see why - getKeys() starts with the first enum item (-1) at index 0, but you want keys starting at the second enum item (to skip NONE), which is index 1 in the array.
I guess this actually does work for enums starting at any non-positive value, although I think it could stand to be clarified a bit.

library/LuaTypes.cpp
lethosor: Following up on my other comment thread: if this were e.g. `first_useful_value = 0`, you could use `first_useful_value - index_enum->getFirstItem()` as the array index and `i = first_useful_value` in the loop initializer. I would be fine with hardcoding that in this function for now, assuming adding that to the XML is complicated.

library/LuaTypes.cpp
BenLubar: yeah, we're taking the address of key 0

it assumes that the enum isn't a complex enum, so that should probably at least have an assert

library/LuaTypes.cpp
BenLubar: we don't need to know what the first useful value is because we know that the first field in this structure is `enum[0]`

lethosor: Oh, you're right, that's how index-enum works

lethosor:
I'm getting a SIGABRT in `StockpileMonitor::doCycle()` (presumably from the stockpiles plugin) when loading a fortress. Doesn't happen on develop, so I'm guessing something is breaking compatibility here.
(I can investigate if you're unable to reproduce on your end)

lethosor:
That crash is still happening in a fresh world, by the way (now that I'm looking at it again, looks like this update was just a rebase, not a change, so that would make sense)

lethosor:
@BenLubar were you unable to reproduce the crash?

BenLubar:
@lethosor I didn't see the comment about the crash - I'll try to reproduce

BenLubar:
Fixed in f20446534bb7f39425e102bd70daec46e328004f

lethosor:
Thanks! I probably should have submitted an actual review, I guess.

PR #1564, opened 2020-05-02
adapted to enum value name change

### LINKED ISSUE #1479: getplants will mark plants for harvest that don't have usable growths

PatrikLundell:
@BenLubar : This PR is dependent upon PR  DFHack/df-structures#400, and so needs linking.

Edit: Ignore the below. Attribute linking won't help in this case.
(Also, it would be better to make use of the new attributes rather than to use a switch statement, with the added benefit that no change of the code would be required when the last(?) enum value is identified, but I don't know how to get at them from C++, and this is probably as good a way to ask as any).

Edit: fix #1479


lethosor:
For reference, [these macros](https://github.com/DFHack/dfhack/blob/2762f75d69e51e8faf9febc515cc45e84999bdbd/library/include/DataDefs.h#L829-L830) would help with accessing enum attributes in C++. But you are correct that the attribute of identity_type doesn't map to the cases here.

PatrikLundell:
Thanks. Useful info.

PatrikLundell:
Since the getplants code relies on structures mapped in the same pending structures PR as the Units one does, both changes were crammed into the same PR.

The getplants code makes use of the logic provided by @quietust to determine the random actual growth period, although that hasn't been tested (would require an embark in the right time window and locating suitable plants within it, which is possible to do). The check to not designate plants that have already been picked was tested against horned melons in season. Running the command after all designations had been cleared designate 38 plants. Clearing the designations and manually designate one of them for picking and running the command once that plant had been picked resulted in 37 designation (and not the picked plant).

plugins/getplants.cpp
```diff
@@ -224,6 +231,106 @@ selectability selectablePlant(const df::plant_raw *plant, bool farming)
     }
 }
 
+//  Formula for determination of the variance in plant growth maturation time, determined via disassembly.
+//  The x and y parameters are in tiles relative to the embark.
+bool ripe(int32_t x, int32_t y, int32_t start, int32_t end) {
+    int32_t time = ((435522653 - (((y + 3) * x + 5) * ((y + 7) * y * 400181475 + 289700012))) & 0x3FFFFFFF) % 2000 + (*cur_year_tick % 403200);
```
quietust, line 237: The `% 403200` should be applied _after_ adding `cur_year_tick` to the delta value (since `cur_year_tick` is already guaranteed to be within that range).

PatrikLundell: Thanks. It looked odd.

PR #1565, opened 2020-05-02
Fixing some crash bugs.

RosaryMala:


lethosor:
I might be missing something - why was this crashing?

RosaryMala:
If you try to set an enum in profobuf to an invalid value, it crashes, and I was previously casting df enums to protobuf enums. This changes it to just use ints instead, and the receiving software will have to worry about decoding it.

PR #1569, opened 2020-05-05
Embark assistant

PatrikLundell:
Fixed 2 incursion handling bugs.


lethosor:
Thanks for the changelog updates!

It looks like you also updated submodules here (specifically stonesense and scripts), which are causing a merge conflict - I usually recommend not adding submodule changes unless you need them for the build to work for this reason, and I don't think either of those would have affected embark-assistant, but it's pretty easy for me to resolve locally.

For a more thorough explanation: you updated the scripts submodule, which was pointing at dfhack/scripts@47d9b3c9, to dfhack/scripts@7e9d1755 (I assume due to your unrelated work on dfhack/scripts#153), but DFHack's develop branch has it pointing at dfhack/scripts@4881acf now, and git doesn't know how to resolve the conflict. (One way I like to think of it: we'd run into basically the same issue if "scripts" were a single-line file containing a link to a commit in the scripts repo, or a commit hash.) I should probably add this to the "Contributing" docs...

PR #1585, opened 2020-06-02
Add a new tweak, "reaction-gloves"

quietust:
When enabled, custom reactions will begin to produce gloves in sets, based
on the number of hands the job performer's race has, and set the
Handedness flags accordingly.

The "createitem" plugin already contains a simpler workaround (which
doesn't check body plan but instead just produces pairs), but it shouldn't
trigger when this tweak is enabled (unless you use it on a creature which
has been modded to only have "neutral" hands).

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -332,6 +332,7 @@ Subcommands that persist until disabled or DF quits:
 :stone-status-all:      Adds an option to toggle the economic status of all stones
 :title-start-rename:    Adds a safe rename option to the title screen "Start Playing" menu
 :tradereq-pet-gender:   Displays pet genders on the trade request screen
+:reaction-gloves:       Fixes reactions to produce gloves in sets with correct handedness
 
 .. _fix-armory:
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,6 +39,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 # Future
 
+## New Tweaks
+- `tweak` reaction-gloves: adds an option to make reactions produce gloves in sets with correct handedness
+
 ## Fixes
 - Fixed a segfault when attempting to start a headless session with a graphical PRINT_MODE setting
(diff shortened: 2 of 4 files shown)


FalcoGer:
Does this also work with boots or other armor types if the creature has, for example, 7 heads? What happens when the race has no hands? How are gloves special?

quietust:
Gloves are special because they need to be marked as right-handed or left-handed in order to be equippable - all other armor types produced by custom reactions work as-is.

ab9rf:
Would it be possible to base the decision on the designated race of the armor item instead of the race of unit doing the creating? I would not be surprised if that `entity` (if not null) points to the race designated for the armor item. While the unit doing the creating is the default, this can be overridden through the job manager. I use this in the `tailor` plugin (see https://github.com/DFHack/dfhack/blob/develop/plugins/tailor.cpp#L398).

quietust:
Additional research would be required to determine exactly how that's handled - the "entity" parameter seems unlikely, since that refers to an entire **civilization** rather than a single creature, so it might be one of the 2 unknown parameters near the end.

lethosor:
Thanks for handling the docs updates! There's a special `:bug:` prefix you can use to link to DF bugs from the docs, so I added that in 6cafd230c5f12a0bb4d0ad0dd4d56ea4b72a270a

ab9rf:
> Edit: the logic inside reaction_product_itemst::produce() **always** sets the "maker race" of the resulting item based on the unit performing the reaction, so I don't see how the "tailor" plugin would be able to do it differently (unless it explicitly modifies the items after the job has been completed, which would put it outside the control of this tweak).

No, it doesn't do anything like that; it simply sets the maker race in the job order. Everything after that is handled by "vanilla" code; `tailor` only creates job orders. I have no idea what DF does in the process of turning a job order into actual items.

quietust:
My guess is that when you create a work order with a specific race, DF only permits it to be assigned to a creature _of_ that race so that the products will be sized accordingly.

PR #1596, opened 2020-06-21
fix labor inferences for custom furnaces [labormanager,WIP]

### LINKED ISSUE #1595: labormanager doesn't adduce correct build labor for custom furnaces
I got a report via Discord of labormanager failing to allocate an architect to build a Masterwork "finishing forge". On research, I've determined that this is a custom furnace. Labormanager is not respecting the designated custom build labor when inferring the required build labor, which results in starving finishing forges (and likely other Masterwork custom furnaces) of the architect required.

ab9rf:
Also clean up how paved roads are handled (the way it was was stupid)

fixes #1595 

WIP b/c not tested yet

### DIFF
--- a/plugins/labormanager/joblabormapper.cpp
+++ b/plugins/labormanager/joblabormapper.cpp
@@ -200,8 +200,6 @@ static df::building* get_building_from_job(df::job* j)
 
 static df::unit_labor construction_build_labor(df::building_actual* b)
 {
-    if (b->getType() == df::building_type::RoadPaved)
-        return df::unit_labor::BUILD_ROAD;
     // Find last item in building with use mode appropriate to the building's constructions state
     // For screw pumps contained_items[0] = pipe, 1 corkscrew, 2 block
     // For wells 0 mechanism, 1 rope, 2 bucket, 3 block
@@ -313,14 +311,27 @@ public:
                 return workshop_build_labor[ws->type];
         }
         break;
+        case df::building_type::Furnace:
+        {
+            df::building_furnacest* frn = (df::building_furnacest*) bld;
+            if (frn->design && !frn->design->flags.bits.designed)
+                return df::unit_labor::ARCHITECT;
+            if (frn->type == df::furnace_type::Custom)
+            {
+                df::building_def* def = df::building_def::find(frn->custom_type);
+                return def->build_labors[0];
+            }
+            else
+                // cast to building_actual should be safe here because at this point the building has been designed
+                return construction_build_labor((df::building_actual*)bld);
+        }
+        break;
         case df::building_type::Construction:
             return df::unit_labor::BUILD_CONSTRUCTION;
-        case df::building_type::Furnace:
         case df::building_type::TradeDepot:
         case df::building_type::Bridge:
         case df::building_type::ArcheryTarget:
         case df::building_type::WaterWheel:
-        case df::building_type::RoadPaved:
         case df::building_type::Well:
         case df::building_type::ScrewPump:
         case df::building_type::Wagon:
@@ -334,6 +345,8 @@ public:
             return construction_build_labor(b);
         }
         break;
+        case df::building_type::RoadPaved:
+            return df::unit_labor::BUILD_ROAD;
         case df::building_type::FarmPlot:
             return df::unit_labor::PLANT;
         case df::building_type::Chair:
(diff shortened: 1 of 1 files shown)


plugins/labormanager/joblabormapper.cpp
```diff
@@ -313,14 +311,27 @@ class jlfunc_construct_bld : public jlfunc
                 return workshop_build_labor[ws->type];
         }
         break;
+        case df::building_type::Furnace:
+        {
+            df::building_furnacest* frn = (df::building_furnacest*) bld;
```
lethosor, line 316: Maybe `virtual_cast` would be safer?

plugins/labormanager/joblabormapper.cpp
```diff
@@ -313,14 +311,27 @@ class jlfunc_construct_bld : public jlfunc
...
+            if (frn->design && !frn->design->flags.bits.designed)
+                return df::unit_labor::ARCHITECT;
+            if (frn->type == df::furnace_type::Custom)
+            {
+                df::building_def* def = df::building_def::find(frn->custom_type);
+                return def->build_labors[0];
+            }
+            else
+                // cast to building_actual should be safe here because at this point the building has been designed
+                return construction_build_labor((df::building_actual*)bld);
```
lethosor, line 326: Also, building_furnacest inherits from building_actual, so you could probably pass in `frn` here without needing the cast

plugins/labormanager/joblabormapper.cpp
ab9rf: Probably. There are a ton of "potentially unsafe" casts in this code. If I "fix" this one, I should probably fix the one that's like ten lines above it in the code for custom workshops, too...

lethosor: Fair, I didn't expand the diff here. `virtual_cast` would check the actual type, but in any case, this should only ever break due to a structures issue (e.g. a new building_type added somewhere in the middle), which would be pretty easy to catch since it would break a lot of other things too.

ab9rf: Also, `virtual_cast` isn't really safer here. All `virtual_cast` is is a wrapper that either casts the pointer to the type or returns null. The code still has to do a null check. But (with current DF structures, at least) we're guaranteed that if `bld->getType()` is Furnace, then `bld` is a `building_furnacest`.

lethosor: virtual_cast also checks DF's type information (not sure if that's what you meant; it's basically dynamic_cast, but we can't use that for DF-defined types). That's based on the `<vtable-address>` entries in symbols.xml, which are a bit more reliable than our enum definitions. But in order for this code to fail, a new building subclass would need to be added whose `getType()` returns what was previously `building_type::Furnace`, and our  `building_type` would need to be outdated, which should really only be possible soon after a DF release. So I agree, it's not really worth changing.

lethosor:
Fine with me if you merge this once you're satisfied that it works (I figure you have a better idea of how to test it)

lethosor:
@ab9rf does this one still just need to be tested? I can make an attempt at that if needed.

ab9rf:
Sorry, I haven't had the opportunity to look at this yet.

lethosor:
I played around with it. Building and destroying a custom furnace in the save from #1595 (thanks, @Tokeli!) seemed to work. However, in that save and in a clean embark, I couldn't get a paved road built. In the former, it got designed, but the one road builder never built it, and in the latter, it never even got designed (nobody wound up with architecture enabled after a few in-game days). I'm not very familiar with labormanager, and I haven't tested whether this happened before this PR yet, but I could be doing something wrong.

lethosor:
Yeah, this seems to break paved roads. I was able to get them built in a fresh save before merging these changes, but not after. Not a huge deal - I suspect rolling back the road-related changes here will fix it. I'll try that tomorrow.

plugins/labormanager/joblabormapper.cpp
```diff
@@ -334,6 +345,8 @@ class jlfunc_construct_bld : public jlfunc
             return construction_build_labor(b);
         }
         break;
+        case df::building_type::RoadPaved:
+            return df::unit_labor::BUILD_ROAD;
```
lethosor, line 349: I think this is the issue - paved roads require architects, which the old case handled. I'll see if that fixes it.

Issue #499, opened 2015-01-22
`fortplan` plugin should support stockpiles and query mode, like Quickfort


PeridexisErrant:
DFHack already includes `digfort` for designations, and `fortplan` for building.  With the new `blueprint` plugin creating complete QF blueprints, it would be good to have a native way to lay out the generated blueprints in-game.

For "place" mode, all basic stockpiles need to be settable.  Custom stockpiles are not supported, IIRC.  For "query" mode, it would be good to support all options - but a majority of cases would be covered by `r+` for making a room.

danaris:
I am working on fortplan, a plugin designed to implement all of Quickfort's basic functionality. At present, it only supports build mode, and only furniture built with items that share the furniture's name. However, I intend to expand it to support all modes, and at least all basic buildings, stockpiles, and queries. It's just going to take some time :-)

myk002:
danaris: are you still working on this? I might start poking around in my spare time if this is still open

danaris:
Sadly, I have not had time to devote to anything dwarfy in a few years. I would be delighted to know that some of the stuff I was poking at was being worked on by those who do.

myk002:
I put together a project plan, though it might take me a very long time to implement it. Would you mind taking a look and leaving comments? You've spent more time thinking about this than I have, I'm sure, and I'd love to hear your thoughts on what's a good idea and what isn't. I'd especially like to find out what parts of the plan are going to be difficult/impossible so I can start thinking about alternatives!
https://docs.google.com/document/d/17EjZlGsUOJ6E8WtqkCRZi9BBKjL9L1IdHfkB3Ajqas8

myk002:
@lethosor If you have time, I would also appreciate your input on this before I do further planning or bring it to the forums.

lethosor:
I left a couple more general comments - I'm not that familiar with the buildings APIs myself.

myk002:
@lethosor Thanks! That helps put some things in perspective. I see @angavrilov and @BenLubar wrote most of Buildings.cpp -- could either of you share your thoughts on the best way to apply building blueprints, especially with buildings that take multiple source components, like wells? Doc link for comment: https://docs.google.com/document/d/17EjZlGsUOJ6E8WtqkCRZi9BBKjL9L1IdHfkB3Ajqas8

myk002:
[...]

PR #1607, opened 2020-07-16
blueprint: write blueprints to blueprints/ subdir

myk002:
Change blueprints plugin to write to the blueprints/ subdir as its root instead of the dfhack root dir. This keeps blueprints organized in a dedicated folder and sets up for the quickfort plugin to read blueprints from this directory in the future.

To enable writing to a subdir that may not exist, blueprint now automatically creates folder trees. E.g. ``blueprint 30 30 1 rooms/dining dig`` will create the file ``blueprints/rooms/dining-dig.csv``). Previously it would fail if the ``blueprints/rooms/`` directory didn't already exist.

Part 1 of many for #499 (implement quickfort)


lethosor:
I see you've dealt with some compatibility issues - thanks for working through that! We are a bit limited in terms of C++14/17 features - we could probably bump the Linux compiler up to GCC 5 or 6 by now, but MSVC 2015 is a hard requirement on the Windows side.

Is this at a point where you'd be okay with merging it? Multiple isolated PRs are fine with me (maybe preferred, even) but I don't want to merge something too soon.

myk002:
Thumbs up for review and merge.

I prefer small, incremental PRs as well. It's much easier to get focused feedback and make corrections before wasting too much time going down a wrong path. As a reviewer, I also know it's much easier to review a small PR than a huge one : )

What works well for you? Should I leave an explicit "Ready for review" comment? Or is "ready for review" implicit if the buildmaster checks are clean and "WIP" is not in the title?

myk002:
btw, the buildmaster is awesome! -- I was really worried about my inability to test on windows -- especially once I found out I couldn't use the C++11 std::filesystem class

lethosor:
> What works well for you? Should I leave an explicit "Ready for review" comment? Or is "ready for review" implicit if the buildmaster checks are clean and "WIP" is not in the title?

It can vary by PR, author, etc - I would suggest marking PRs as drafts if they're not ready to go (assuming that CI runs on non-draft PRs, which I think it does), and I'll merge non-drafts that are passing.

myk002:
Awesome. Thanks!

lethosor:
FYI, I added mentions to the `mkdir_recursive` addition in a couple places: https://github.com/DFHack/dfhack/commit/0cc6f40bd6bb936665edd4e8f6ca3fb507476796

PR #1609, opened 2020-07-18
quickfort config, docs, and data

myk002:
#499 Initial configuration and data files for the quickfort script.

Installed file layout is what was agreed upon in https://docs.google.com/document/d/17EjZlGsUOJ6E8WtqkCRZi9BBKjL9L1IdHfkB3Ajqas8/edit#heading=h.as2pioa5qx8r

but are you comfortable with directories in data/ being installed to non-sibling locations?
data/blueprints -> blueprints
data/quickfort -> hack/data/quickfort

We can always change the source layout later, of course, but I wanted your check if you'd prefer a different organization.

aliases-common.txt is forked from https://github.com/joelpt/quickfort/blob/master/qfconvert/config/aliases.txt for compatibility with existing Quickfort blueprints.

I refactored Filesystem::listdir_recursive() to support removing the path prefix from the returned files list to simplify the new recursive default config implementation. There are no current calls that make use of the existing prefix parameter, so I converted it into a boolean. Current usages will use the new default parameter and will not see any changed behavior.


myk002:
@lethosor does this PR look good to you?

lethosor:
Yeah, it looks good. I've been working a bit on merging other changes for a 0.47.04-r2 release at some point in the near future, and I'm guessing that the entire new quickfort won't be ready by then, so do you think it would make more sense to have a separate long-lived branch to merge quickfort-related work into? (This PR as-is would probably be fine in a release, although the `blueprints` folder might confuse people if it's unused...)

docs/changelog.txt
```diff
@@ -51,7 +51,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `confirm`: added a confirmation dialog for convicting dwarves of crimes
 
 ## API
-- Added ``Filesystem::mkdir_recursive``
+- Added ``Filesystem::mkdir_recursive`` (returns true if directories to create already exist)
```
lethosor, line 54: Nitpicking a bit, but I would probably avoid including implementation details in the changelog - a description of the function is fine (although I left it out earlier because I thought the name was self-explanatory), but for details about arguments, return values, etc., I would say that's what the API docs are for (in this case, `Lua API.rst` and whatever's in `Filesystem.h`).

myk002:
I would be absolutely fine with making a long term branch for the quickfort script in the scripts repo, but it would be nice if we could merge code here into develop, so I can share the in-progress script with beta testers and actually have it work for them.

The blueprints dir will still be used by the blueprints plugin, at least, so it won't be a useless directory.

lethosor:
Oh, yeah, a script that's partly done is better than no script, as long as it works. I forgot most of the development was taking place in the scripts repo - as long as whatever's merged is relatively stable, that's fine with me.

Good to know about the blueprints dir. I'll merge this as-is, then.

lethosor:
I made some changes to error handling in https://github.com/DFHack/dfhack/commit/cafdc1620b5e57ac744064c448682bd417cabcb4 (I was the one who originally implemented this, so I didn't want to ask you to fix it, but it may be relevant if you write more Lua bindings). Basically, when I was testing yesterday, I tried passing in a boolean as the second argument to listdir_recursive, which was silently discarded because of the `lua_type` check. Generally I would recommend using the `luaL_check` family for argument validation; unfortunately, there is no `luaL_checkboolean`, but existing functions are pretty lenient about converting values to booleans as needed anyway.

myk002:
I actually thought the absence of luaL_checkint was intentional -- I was calling filesystem.listdir_recursive from Lua, passing nil as the depth parameter so I could set include_prefix but still use the default depth:

local paths = dfhack.filesystem.listdir_recursive(
        quickfort_common.settings['blueprints_dir'], nil, false)

I'll change my usage to pass an explicit 10 as the depth param, but passing nil to get the default seemed cleaner.

lethosor:
That's actually a good point, and not something I anticipated when I originally wrote it. I'll change it back. Most of our existing Lua API functions don't support default arguments at all (trailing boolean arguments that default to `false` work, but that's just because Lua automatically converts `nil` to `false`). Named arguments (like Python) would be nice here, but Lua doesn't really support that.

lethosor:
I settled on allowing `nil` to be used in the cases you mentioned, but still keeping the type check so that a second argument of `'x'` or `true` raises an error.

PR #1612, opened 2020-07-25
Support internal scripts that don't show up in #DFHack ls

myk002:
#499 Enables the quickfort script to span multiple .lua files without polluting the output of #DFHack ls

This could also be useful for advfort, which is split into a main file and a module, but the module gets uselessly listed when we run ls -a

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,7 +56,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - Added a ``ref_target`` field to primitive field references, corresponding to the ``ref-target`` XML attribute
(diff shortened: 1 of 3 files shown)


lethosor:
Woah, that advfort module is old. I should probably update it to use reqscript instead of mkmodule/require (and the docs don't really describe that it's a module either).

As for this PR, it's something that a couple other scripts could benefit from too (dwarf-op and view-item-info, at least). I like the name `internal`, but how about making it a subfolder instead? That would allow organizing multiple supporting files like `internal/blueprint/x.lua` without taking up as much space in the top-level scripts folder, and would probably support languages other than Lua without additional changes.

myk002:
that would be much more generally applicable compared to what I had : )  done. I updated DFHack/scripts#159 with the new "internal" directory structure. Am I using modules correctly there?

PR #1614, opened 2020-07-30
corrected incursion bugs

PatrikLundell:


plugins/embark-assistant/survey.cpp
```diff
@@ -1675,11 +1708,44 @@ uint8_t  embark_assist::survey::translate_corner(embark_assist::defs::world_tile
...
+            return 5;
+        }
+
+        //  Neither tile will automatically yield to the other
         if (corner_location == 7) {
             return 4;
         }
         else {  //  Can only be corner_location == 8
-            return 3;
+            return 5;
```
lethosor, line 1748: Do these constants correspond to enums? Even if not, maybe introducing an enum of your own could help avoid bugs like these. There could be a fair amount of overhead involved, so it's not a huge deal - just a suggestion.

PatrikLundell: Yes, it's the result of that
I've wanted to make an enum, but, as you say, it's a bit of work, and since RedDwarfStepper is working on improving the logic I don't want to change more than a minimum now.

The "enum" is one of the 3(?) tile referencing systems DF uses (hoping the map rewrite will consolidate them into one...):
- 0-8 (this one, which is my preferred one): NW, N, NE, W, Self, E, SW, S, SE
- 1-9 (which is actually used in structures <comments about 0 not being used>): SW, S, SE, W, Self, E, NW, N, NE
- 0-8: Self, NW, N, NE, W, E, SW, S, SE

lethosor: My understanding is that it's going to be more of a cosmetic improvement than an internal one, but yeah, that's really confusing.

PatrikLundell: Not really. It's likely the incorrect specification of the air biome is a direct result of thinking the third addressing type was used, while it is the first one (all tiles, except those that are bugged to have a random value, have a value of 0 rather than 4). Using a single addressing scheme eliminates the risk of trying to use the wrong one.

PR #1615, opened 2020-08-04
Add Goal to Manipulator Detail Modes

ryanbennitt:
Added ability to toggle view so that the dwarf's dream/goal is displayed in Dwarf Manipulator, making it easier to e.g. ensure those who want to craft masterworks have appropriate skills set. This can also be sorted allowing like-minded dwarfs to be grouped, e.g. for match-making those wanting to start families by setting their skills the same so they have a chance to meet. Text and highlighting shows those who have already achieved their goal/dream, and for those wanting to start families also highlights their gender.


library/include/modules/Units.h
```diff
@@ -180,6 +181,10 @@ DFHACK_EXPORT std::string getCasteProfessionName(int race, int caste, df::profes
 DFHACK_EXPORT int8_t getProfessionColor(df::unit *unit, bool ignore_noble = false);
 DFHACK_EXPORT int8_t getCasteProfessionColor(int race, int caste, df::profession pid);
 
+DFHACK_EXPORT df::goal_type getGoalType(df::unit *unit, int goalIndex = 0);
+DFHACK_EXPORT std::string getGoalName(df::unit *unit, int goalIndex = 0);
+DFHACK_EXPORT bool isGoalAchieved(df::unit *unit, int goalIndex = 0);
```
lethosor, line 186: If you could also add these to the units module in `LuaApi.cpp` (and maybe even document them in `docs/Lua API.rst`), that would be great.

plugins/manipulator.cpp
```diff
@@ -32,9 +32,12 @@
 #include "df/caste_raw.h"
 #include "df/historical_entity.h"
 #include "df/entity_raw.h"
+#include "df/goal_type.h"
+#include "df/unit_personality.h"
 
 #include "uicommon.h"
 #include "listcolumn.h"
+#include "manipulator.h"
```
lethosor, line 40: Looks like you either forgot to add this file or this was a holdover from an earlier change:
https://github.com/DFHack/dfhack/pull/1615/checks?check_run_id=946915877#step:10:383

ryanbennitt: Oops, that's from a refactor I backed out of, Visual Studio trying too be too helpful. Will remove...

library/include/modules/Units.h
ryanbennitt: Thanks, I'll look to add these tomorrow.

lethosor: There's a hopefully-helpful page about building the docs here, if you want to see how they look when built: https://docs.dfhack.org/en/latest/docs/Documentation.html
It could be a lot of overhead relative to the size of your changes, so I'd also be willing to build them and make sure the formatting looks ok.

library/modules/Units.cpp
```diff
@@ -1437,6 +1437,73 @@ int8_t Units::getCasteProfessionColor(int race, int casteid, df::profession pid)
...
+    CHECK_NULL_POINTER(unit);
+
+    df::goal_type goal = getGoalType(unit, goalIndex);
+    bool achieved_goal = isGoalAchieved(unit, goalIndex);
+
+    switch (goal)
+    {
+    default:
+    case df::goal_type::STAY_ALIVE:
+        return achieved_goal ? "Stayed Alive" : "Stay Alive";
```
quietust, line 1464: I'm wondering if maybe these strings should get added to the Structure XML files (though the "Start a Family" one might be tricky since it's apparently gender-specific).

ryanbennitt: Not seeing any structure xml files. Can you point me at an example?

ryanbennitt: Oh, you mean df.units.xml? goal_type is defined there, but I guess I need to add additional sub-elements to it?

lethosor: Yeah, enum items can have "attributes". [Here](https://github.com/DFHack/df-structures/blob/0.47.04-r1/df.buildings.xml#L21) is an example, and [here](https://github.com/DFHack/dfhack/blob/0.47.04-r1/plugins/remotefortressreader/building_reader.cpp#L81) is an example of code that reads it.
In this case, there could be two attributes for the unachieved and achieved names. Maybe `name` and `achieved_name` or `caption` and `achieved_caption`. The case for `START_A_FAMILY` could use "Parent", although I haven't checked what the game displays - you could replace "Parent" in this function dynamically, or use a placeholder in the attribute definition, which [some unit_thought entries](https://github.com/DFHack/df-structures/blob/0.47.04-r1/df.unit-thoughts.xml#L667) do.

This isn't required, of course, but it might make the function more future-proof - two new goals were added in 0.47.01, and if someone adds more to the XML files in the future, they'll be more likely to update attributes if they're right there.

lethosor: Another benefit of using enum attrs: I made the naming change Quietust suggested in https://github.com/DFHack/df-structures/pull/409#discussion_r467296259, and it didn't require any code changes on the DFHack side at all.

PR #1616, opened 2020-08-05
Update quickfort docs and configs

myk002:
#499 I moved blueprints/README.txt to blueprints/README.md. My plan is to move the user guide text to that file. Markdown will be much better for formatting than txt, and much of the source material will be drawn from existing markdown files.

this PR also significantly expands on the aliases list and the aliases documentation, though I may end up moving/duplicating some of the alias docs to the user guide.


myk002:
My plan, at least for the short term, is just to link straight to the README.md in the repository. Github will render it. It won't be part of docs.dfhack.org but it will still be accessible and usable. I can see how having all the docs built together is preferable, though. We can look into converting it with sphinx plugins or, if all else fails, reformatting it in RST when we get closer to an official release

lethosor:
Yep, that should be fine. Reformatting can be a bit painful too, although markdown typically isn't very complicated. Probably a good thing to keep this out of the main user-facing DFHack docs if we're not sure when it'll be released. (There's not a lot left that I want to get into r2 - maybe it'll be out this weekend?)

PR #1617, opened 2020-08-07
Quickfort user guide

myk002:
#499 Add the quickfort user guide.

The script command documentation that shows how to *apply* blueprints is still in the quickfort.lua file so it shows up in the official docs and in the quickfort help output. This user guide is more about understanding and building blueprint files.

You can see how GitHub will render the file by copying the whole thing into a comment box and then hitting 'Preview'


lethosor:
> You can see how GitHub will render the file by copying the whole thing into a comment box and then hitting 'Preview'

There's also this option, which takes you [here](https://github.com/DFHack/dfhack/blob/03b28f4e083e7f49a9e7734eb05cc74beebb32cb/data/blueprints/README.md):

image

myk002:
> There's also this option
That's much nicer, and you can actually check to make sure the section links work that way. Thanks!

myk002:
> I think this looks good! Far more comprehensive than most scripts' docs. I would like to convert it to RST and integrate it with the rest of the docs at some point before r3, although it's not as high-priority. The downside is that GitHub won't render it as well - inter-doc links typically won't work, for one thing. We do distribute HTML versions of the docs in the hack/docs folder, though, so users should be able to get this locally without it being in the data folder.
> (Does that sound good, or am I missing a reason why you set it up this way?)

Thanks! The reason I put the doc in the blueprints folder is so people could find it "organically" by looking through their local files, perhaps after saving a blueprint with the blueprints plugin. I'd be perfectly happy with a README.txt file that points people to the online documentation, though. I don't think there's a need to have the whole thing locally.

The reasons I wrote it in markdown are: 1) A good chunk of the original guide came from Joel Thompson's Quickfort User Guide, which is written in markdown, and I wanted to build on that. and 2) I'm much more fluent in markdown than RST

But I agree that converting it into RST and properly integrating it into the DFHack docs is the way to go. It's fine if github doesn't render it well since we'll just point all our links to it on docs.dfhack.org.

lethosor:
Ok, a link of some sort in the data folder makes sense to me. I'll merge as-is and deal with reformatting later.

PR #1620, opened 2020-08-09
Quickfort xlsxio integration

myk002:
#499 Context: I'm attempting to pull in a C library that supports reading from .xslx files so the quickfort script can use it.

At a high level, here is what this PR does:

1) integrates the xlsxio library and its dependencies into the build
2) implements a glue layer that makes the xlsx functionality available to dfhack
3) exposes the API to Lua

The libraries I am adding are:
* [libzip](https://github.com/nih-at/libzip) (BSD license)
* [libexpat](https://github.com/libexpat/libexpat) (MIT license)
* [xlsxio](https://github.com/brechtsanders/xlsxio) (MIT license)

I've added them as submodules in the depends/ directory, checked out to a version that builds on all CI environments.

I configured them to build static libs so we don't add to our list of shared library files in libs/

I use CMake's ExternalProject module to call out to the dependencies' build scripts to keep them as separate as possible from DFHack's CMake config. ExternalProject ensures the dependency inherits the existing build environment (like compiler settings) but prevents the dependency's environment from leaking back into DFHack's. I thought this was appropriate since we don't control these dependencies and we can't predict what changes they'll make in the future. I also ran a test where I included the new deps with add_subdirectory(), and the number of options they added to the project made configuring dfhack with ccmake very very confusing.

The glue layer is implemented as a plugin.


.gitmodules
```diff
@@ -16,3 +16,15 @@
 [submodule "depends/jsoncpp"]
 	path = depends/jsoncpp-sub
 	url = ../../DFHack/jsoncpp.git
+[submodule "depends/xlsxio"]
+	path = depends/xlsxio
+	url = ../../brechtsanders/xlsxio.git
+	shallow = true
```
lethosor, line 22: > 1b) I plan to fix the dependencies to a particular branch/commit instead of having them track master/HEAD. I don't want them changing out from underneath us. Is this ok?

Technically, that's what submodules always do. For example, a fresh clone of DFHack 0.47.04-r2 (commit https://github.com/DFHack/dfhack/commit/f2b0f012c9895aa3fd743ad50fab9e6fb78b479d) will check out https://github.com/dfhack/scripts/commit/568c586 in the scripts folder. You can check out the `master` branch or the `0.47.04-r2` tag, but since it points to the same commit, Git won't recognize a change in the DFHack repo. It only keeps track of submodules' commit hashes in the parent project, so they'll always be locked to whatever we specify (at least in an unmodified clone) without doing anything special.

The only different thing I see here (compared to other submodules) is `shallow = true`. That wouldn't change the behavior of these submodules, other than making the downloaded history smaller (it's equivalent to `--depth 1` when cloning), which looks like it's a good thing for at least expat, which has a large history. I assume that's why you added this. A shallow clone of submodules *might* make updating to a new version trickier, since only a single commit would have been fetched originally - I would likely have to experiment to understand how that works.

CMakeLists.txt
```diff
@@ -415,6 +417,30 @@ endif()
 
 # build the lib itself
 if(BUILD_LIBRARY)
+    ExternalProject_Get_Property(xlsxio_project INSTALL_DIR)
+    set(XLSXIO_INSTALL_DIR ${INSTALL_DIR})
+    include_directories(${XLSXIO_INSTALL_DIR}/include)
+    add_library(xlsxio_read STATIC IMPORTED)
```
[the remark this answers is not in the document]
myk002: I'll play around with this a little to see if I can get it back into depends/CMakeLists.txt.  The issue is described here: https://stackoverflow.com/a/23566351 : Imported targets do not follow the same visibility rules as non-imported targets

library/LuaApi.cpp
```diff
@@ -2480,6 +2481,65 @@ static const LuaWrapper::FunctionReg dfhack_kitchen_module[] = {
...
+
+// internal function to factor out handle extraction
+static void * get_xlsxreader_handle(lua_State *L)
+{
+    if (lua_gettop(L) < 1 || lua_isnil(L, 1))
+    {
+        luaL_error(L, "invalid xlsxreader handle");
+    }
+    luaL_checktype(L, 1, LUA_TLIGHTUSERDATA);
+    return lua_touserdata(L, 1);
```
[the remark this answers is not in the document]
myk002: well, I would say I *chose* to use light userdata. I actually originally typechecked it as userdata, but the check failed because it was being passe din as light userdata, so I changed the check : p Lua data shuttling is an entirely new area for me, so I'm kinda looking to you to define what "good enough" is here. The current code works, I've verified, when used properly.

.gitmodules
myk002: ah, ok. I didn't know they were locked to a specific commit by design. chalk that misunderstanding up to my inexperience with creating git submodules. I suggest, then, that the policy be that we never update these dependencies unless we actually need to.

Yes, I added the shallow=true lines since I figured we don't really care about the commit histories, and it reduces the download size. I don't seem to have any trouble switching branches or pulling updates in those `--depth 1` repos.

myk002:
- ok, static libs it is

- I *could* add the glue layer directly into LuaApi.cpp, but I'm kinda wary of letting xlsxio's header files be visible to any more code than it has to. I reviewed the code for general correctness, but it's a lot messier than I would like, and I don't really feel comfortable including it in such an important file as LuaApi.cpp. I could just put it in a file in library/ as a peer to LuaApi.cpp, if it's not useful as a module (and I agree as a module it's kinda an odd man out).

CMakeLists.txt
```diff
@@ -415,6 +417,30 @@ endif()
 
 # build the lib itself
 if(BUILD_LIBRARY)
+    ExternalProject_Get_Property(xlsxio_project INSTALL_DIR)
```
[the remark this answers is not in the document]
myk002: I tested it with just add_subdirectory(), and although it worked fine, it *really* polluted our option space with junk. it becomes very difficult to configure dfhack with ccmake since there are so many options

myk002:
Actually, would this work well as a plugin? It's "utility"-like, like the cxxrandom plugin

lethosor:
Avoiding header pollution is a good point. I think a plugin is a good choice (luasocket is another example, in addition to cxxrandom).

myk002:
Ok, it finally builds everywhere it's supposed to. I also switched to using CMAKE vars for library prefixes and suffixes and modified xlsxreader.cpp to use USERDATA instead of LIGHTUSERDATA as per your suggestions in #dfhack

Thank you so much for your help with getting this done!

docs/Plugins.rst
```diff
@@ -2939,10 +2939,11 @@ Lua API
 =======
 
 Some plugins consist solely of native libraries exposed to Lua. They are listed
-in the `lua-api` file under `lua-plugins`:
+in the `lua-api` file under `lua-plugins`::
```
lethosor, line 2942: The extra colon is making this a code block instead (at least when I make this change on `develop`):

image

plugins/xlsxreader.cpp
```diff
@@ -0,0 +1,181 @@
...
+};
+
+struct xlsx_sheet_handle {
+    const xlsxioreadersheet handle;
+    xlsx_sheet_handle(xlsxioreadersheet handle): handle(handle) {}
+    static handle_identity _identity;
+};
+
+handle_identity xlsx_file_handle::_identity;
+handle_identity xlsx_sheet_handle::_identity;
```
lethosor, line 69: It might be a good idea to give these separate names, i.e. changing the parameter that's hardcoded to `"handle"` above. As far as I know, the names are still human-facing, but it could be confusing to have two different types, both named "handle", with functions that only accept one of them.

Out of curiosity, have you tried unloading this plugin while Lua code has access to a handle? If that crashes, I wouldn't be too surprised, but there are probably other ways to cause a crash when unloading some plugins, so it's probably an acceptable risk for now.

I haven't yet figured out if simply creating an object with these identities would be enough to cause a crash later if the plugin is unloaded, either when garbage collection kicks in or if some internal code collects a list of identities as they're registered and tries to access them later. If you haven't encountered a crash after some time in-game after running quickfort, and maybe after calling `collectgarbage()` from Lua, I suspect it's fine. I'd like to do a bit more research into this, though.

docs/Plugins.rst
myk002: whoops -- that was the rendering I wanted when I wrote the example code. obviously I didn't check how it affected this block. will fix. one sec

plugins/xlsxreader.cpp
myk002: I tested reloading the plugin while a handle is open, and everything worked ok:
DFHack# lua
[lua]# xlsxreader = require('plugins.xlsxreader')
[lua]# xlsx_file = xlsxreader.open_xlsx_file('blueprints/industry.xlsx')
[lua]# xlsx_sheet = xlsxreader.open_sheet(xlsx_file, 'dig')
[lua]# printall(xlsxreader.get_row(xlsx_sheet))
[expected output]
[lua]# printall(xlsxreader.get_row(xlsx_sheet))
[expected output]

then from a separate window:
$ ./dfhack-run reload xlsxreader

then back in DFHack:
[lua]# collectgarbage()
[lua]# printall(xlsxreader.get_row(xlsx_sheet))
[expected output]
[lua]# printall(xlsxreader.get_row(xlsx_sheet))
[expected output]
[lua]# xlsxreader.close_sheet(xlsx_sheet)
[lua]# xlsxreader.close_xlsx_file(xlsx_file)

I think it's expected that it works since xlsxreader holds all the state in the allocated handle. it doesn't have any global state to lose in a reload. the handles are just malloc'd, so it's not like the memory references become bad -- glibc is still the same

lethosor: The tricky part is that the addresses of the identities (like `xlsx_file_handle::_identity`) can change across reloads. I think those are used at least somewhere in the Lua wrapper layer. Your investigation sounds better than I had hoped for, though!

myk002: ah, that's a very good point. Should I move the statics into the core library?

myk002: library/include/PluginStatics.h and library/PluginStatics.cpp?

lethosor: I usually prefer to keep plugin things out of non-plugins, but I think this can be an exception (at least for now, until there's a better way to handle plugin-defined types).

depends/CMakeLists.txt
```diff
@@ -17,3 +17,62 @@ option(CLSOCKET_SHARED "Build clsocket lib as shared." OFF)
...
+
+if(WIN32)
+  set(EXPAT_LIB_NAME_SUFFIX "MD")
+  set(XLSXIO_C_FLAGS "${CMAKE_C_FLAGS} /DXML_STATIC")
+else()
+  set(EXPAT_LIB_NAME_SUFFIX "")
+  set(XLSXIO_C_FLAGS "${CMAKE_C_FLAGS} -DXML_STATIC")
+endif()
+
+set(LIBEXPAT_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/libexpat)
```
lethosor, line 45: I tried building without setting up the new submodules and got a somewhat-unhelpful error message (where the suggested solutions aren't applicable to users who have forgotten this step):

```
CMake Error at /usr/share/cmake-3.16/Modules/ExternalProject.cmake:2630 (message):
  No download info given for 'libexpat_project' and its source directory:

   /path/to/dfhack/depends/libexpat/expat

  is not an existing non-empty directory.  Please specify one of:

   * SOURCE_DIR with an existing non-empty directory
   * DOWNLOAD_COMMAND
   * URL
   * GIT_REPOSITORY
   * SVN_REPOSITORY
   * HG_REPOSITORY
   * CVS_REPOSITORY and CVS_MODULE
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/ExternalProject.cmake:3236 (_ep_add_download_command)
  depends/CMakeLists.txt:48 (ExternalProject_Add)
```

There's an existing check for some other submodule files here: https://github.com/DFHack/dfhack/blob/f2b0f012c9895aa3fd743ad50fab9e6fb78b479d/CMakeLists.txt#L181

Do you have thoughts on a clean way to perform a similar check here? I was considering a check for the CMakeLists.txt in each folder before the call to ExternalProject_Add, but was wondering if you had better ideas.

Another option, of course, is to leave this as-is and document the error message in Compile.rst. My concern here is mainly that this is the first time we've added submodules in a while, and the first time using ExternalProject with them, so people with existing DFHack clones might not realize how to fix this issue if they run into it.

plugins/xlsxreader.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+void close_xlsx_file(xlsx_file_handle *file_handle) {
+    CHECK_NULL_POINTER(file_handle);
+    xlsxioread_close(file_handle->handle);
+    delete(file_handle);
+}
+
+// returns XLSXIOReaderSheet object or NULL on error
+xlsx_sheet_handle* open_sheet(xlsx_file_handle *file_handle,
```
lethosor, line 29: I haven't done a lot of debugging yet, but this function segfaults (due to a null `handle` passed to `xlsxioread_sheet_open()`) if the xlsx file failed to open, due to it not existing (which happened to me when testing originally) or the user lacking permission to read it. Quickfort seems to have pretty robust checks for the former, i.e. not opening files that don't exist, but the latter might cause issues. I haven't gotten `quickfort list` to segfault yet (it prints the `no sheet with a #mode marker detected` message for non-readable files), but other user code might.

Is there a good place to handle this? Maybe if `xlsxioread_open` returns null, that should throw a Lua error in `open_xlsx_file`? I think that should prevent Lua code from obtaining a null handle in the first place.

depends/CMakeLists.txt
myk002: Adding additional checks to the top-level CMakeLists.txt seems like a good idea -- I'd rather add it there than in depends/CMakeLists.txt since we already have the error message in place directing people to Compile.rst, which is still applicable here.

plugins/xlsxreader.cpp
myk002: I think this can be solved with better use of CHECK_NULL_POINTER. I'm currently using it to verify my *wrapper*, but not the inner handle itself. Let me test that a bit and make sure that protects successfully against segfaults

myk002: the problem was that I didn't update my NULL detection when I switched to using the _identity-compliant wrappers. I wasn't able to induce any NULL-related segfaults with the updated checks. Thank you for catching this!

plugins/CMakeLists.txt
```diff
@@ -175,8 +189,11 @@ if(BUILD_SUPPORTED)
     add_subdirectory(tweak)
     dfhack_plugin(workflow workflow.cpp LINK_LIBRARIES lua)
     dfhack_plugin(workNow workNow.cpp)
+    dfhack_plugin(xlsxreader xlsxreader.cpp LINK_LIBRARIES lua xlsxio_read zip expat ${LIB_Z_LIB})
     dfhack_plugin(zone zone.cpp LINK_LIBRARIES lua)
 endif()
+target_include_directories(xlsxreader PRIVATE ${XLSXIO_INSTALL_DIR}/include)
+add_dependencies(xlsxreader xlsxio_project)
```
lethosor, line 196: A bit late here, but this should have gone in the `if (BUILD_SUPPORTED)` block above as well. For instance, df-ai builds with BUILD_SUPPORTED=OFF, and this fails because the `xlsxreader` target isn't created: https://buildmaster.lubar.me/executions/execution-details?executionId=504238. I'll push a fix.

myk002: Ack. Thanks!

PR #1622, opened 2020-08-12
quickfort: expose basic buildingplan functionality to Lua

myk002:
#499 
More changes are coming to support more buildings and better filtering, but I wanted to submit this as a separate PR so we could discuss any general concerns you might have about extending buildingplan functionality.


plugins/buildingplan-lib.cpp
```diff
@@ -623,11 +623,6 @@ bool Planner::allocatePlannedBuilding(df::building_type type)
         return false;
     }
 
-    for (auto iter = newinst->jobs.begin(); iter != newinst->jobs.end(); iter++)
-    {
-        (*iter)->flags.bits.suspend = true;
```
lethosor, line 628: Out of curiosity, why did this move to addPlannedBuilding?

plugins/lua/buildingplan.lua
```diff
@@ -0,0 +1,3 @@
+local _ENV = mkmodule('plugins.buildingplan')
+
```
lethosor, line 2: Lua API docs would be great - simple is fine for these functions, just as long as people know that the functions exist. Listing them in `Lua API.rst` is probably more important than listing them here, but seeing this reminded me. (If you wanted to wait for a  confirmation before documenting these functions, they look sensible to me.)

plugins/buildingplan-lib.cpp
myk002: addPlannedBuilding is an API method, but it didn't seem to enforce the supposed invariant that planned buildings are suspended. Refactoring the functionality into the method itself seemed cleaner than making it a separate API method or reimplementing the suspension in Lua.

plugins/lua/buildingplan.lua
myk002: done. thanks!

plugins/buildingplan-lib.cpp
lethosor: Ah, I see allocatePlannedBuilding calling addPlannedBuilding now (the door check was between this removed section and that call, but didn't need to be).

PR #1624, opened 2020-08-16
Properly calculate valid tiles for civzones in the Buildings module

myk002:
#499 
I ran into trouble in the quickfort script while constructing activity zones. Zones that happened to be taken up by trees on the surface were failing to get created. Creating such zones in the UI works fine, so I investigated.

It turns out checkFreeTiles doesn't handle civzones properly. It assumes that no "building" can be placed in a wall, which is usually true, but not for civzones.  There was no other way to get whether the building is a civzone in that function, so I added an optional parameter to pass the information in. I would have liked to handle it at a higher level, but checkFreeTiles has useful side effects (like allocating extents) that are difficult to factor out.

### DIFF
--- a/library/include/modules/Buildings.h
+++ b/library/include/modules/Buildings.h
@@ -143,7 +143,7 @@ DFHACK_EXPORT bool checkFreeTiles(df::coord pos, df::coord2d size,
                                   df::building_extents *ext = NULL,
                                   bool create_ext = false,
                                   bool allow_occupied = false,
-                                  bool is_civzone = false);
+                                  bool allow_wall = false);
 
 /**
  * Returns the number of tiles included by the extent, or defval.
--- a/library/include/modules/Buildings.h
+++ b/library/include/modules/Buildings.h
@@ -141,7 +141,9 @@ DFHACK_EXPORT bool getCorrectSize(df::coord2d &size, df::coord2d &center,
  */
 DFHACK_EXPORT bool checkFreeTiles(df::coord pos, df::coord2d size,
                                   df::building_extents *ext = NULL,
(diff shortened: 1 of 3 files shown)


library/include/modules/Buildings.h
```diff
@@ -141,7 +141,9 @@ DFHACK_EXPORT bool getCorrectSize(df::coord2d &size, df::coord2d &center,
  */
 DFHACK_EXPORT bool checkFreeTiles(df::coord pos, df::coord2d size,
                                   df::building_extents *ext = NULL,
-                                  bool create_ext = false, bool allow_occupied = false);
+                                  bool create_ext = false,
+                                  bool allow_occupied = false,
+                                  bool is_civzone = false);
```
lethosor, line 146: Would `allow_wall` be a better name, in case other building types run into this?

myk002: I wasn't sure how general this would be, but that works for me. The specific call that it's bypassing is HighPassable(). Is that close enough to "isWallI()"?

lethosor: I had to look it up, and I'm still a bit confused what "high passable" means - the comment in the XML file for the passable_high attribute is "tile is missing a roof", which I guess means the "main" part of the tile (not the floor) is open?

I would lean towards a more generic name in this case, and `allow_wall` is (IMO) more generic than `is_civzone`.

myk002: agreed. done.

PatrikLundell:
I think "allow_blocking_tiles" would have been better, as trees are not walls, and there may be additional non wall blockers as well.
However, the current name will likely cause a minimum of confusion.

PR #1630, opened 2020-08-21
Pass CMAKE_GENERATOR_TOOLSET (cmake -T argument) to external projects

lethosor:
The dependencies introduced in #1620 cause an error when building natively on Windows, according to @ragundo. CMake is looking for build tools `v100`, which are part of VS2010 and won't work with DF/DFHack, even if they were installed.

Some of the Windows build scripts pass in `-T v140_xp` (from VS2015), so passing this option through to subprojects may help: https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html

Opening a PR to double-check that this doesn't break Buildmaster


lethosor:
Repeating some findings from IRC: this still doesn't work with msbuild from the command line (which still tries to use v100, i.e. VS2010), and we haven't been able to figure out why. It does build from within Visual Studio, though. Maybe this PR helps, and it doesn't seem to hurt anything, so I'll go ahead and merge it.

We may need to bundle these dependencies in `depends` and write our own `CMakeLists.txt`s (that don't make config changes we don't want) to fully resolve the issue. @myk002 I'm assuming you don't have any changes pending related to third-party libs, right?

myk002:
> @myk002 I'm assuming you don't have any changes pending related to third-party libs, right?

No changes planned. I'm sorry I caused you this compatibility pain : (

lethosor:
No worries. As far as I'm concerned, CMake or msbuild are causing the issue here. I likely won't change it unless other people also run into this issue on Windows.

RosaryMala:
Can confirm I'm getting the same issue, and not able to find any reference to VS2010 in any of the files, so no idea where it's coming from.

PR #1633, opened 2020-08-25
Switch trade screen search to use search_multicolumn_modifiable

GitOnUp:
…so that item counts are updated correctly with other information.


GitOnUp:
Worked with a repro save here: https://dffd.bay12games.com/file.php?id=15184

Behavior pre-fix was that an entire stack would be traded, verified in both directions that the correct amount is transferred post-fix.

PR #1634, opened 2020-08-27
quickfort: document marker mode and buildingplan

myk002:
#499
also a bit of cleanup and reorg

### DIFF
--- a/data/blueprints/README.md
+++ b/data/blueprints/README.md
@@ -486,7 +486,7 @@ Start positions specify a cursor offset for a particular blueprint, simplifying 
 
 will build the workshop *centered* on the cursor, not down and to the right of the cursor.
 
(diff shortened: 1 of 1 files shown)


myk002:
> Are you planning on making changes to this in the near future? I would like to convert this to RST and get it into the main DFHack docs so that other pages can cross-reference it instead of linking to the page on GitHub, but I don't want to conflict with any changes you're making at the same time.

No immediate changes to this file planned. I'm writing a section on advanced blueprints, but I'm currently just getting a draft together as a google doc. once it's ready, I can add it to whatever format the user manual is in then.

myk002:
Is it possible to put the rst version of this guide in the scripts repo? It would be nice to be able to document a feature in the same PR that I implement it

lethosor:
It's technically possible, yeah, since the documentation "root" is the root of the DFHack repo (leading to the /docs/ prefix in many pages like https://docs.dfhack.org/en/latest/docs/Documentation.html, which I'd like to get rid of, but that's another matter). There are a couple pages from df-structures included in the docs:
* https://docs.dfhack.org/en/latest/library/xml/SYNTAX.html
* https://docs.dfhack.org/en/latest/library/xml/how-to-update.html

These are linked to from https://docs.dfhack.org/en/latest/docs/Structures-intro.html links to them ([source](https://github.com/DFHack/dfhack/blob/develop/docs/Structures-intro.rst))

myk002:
Totally ok with me to move the user guide content to the scripts repo. Maybe just change this file to contain a pointer to the online location of the user guide.

PR #1637, opened 2020-09-01
Quick'n'Dirty autocomplete for Posix console

PassionateAngler:
Hello.
I have added quick'n'dirty support for TAB auto-completion for Linux (Posix) console.
Since I have almost no experience with programming for windows I am not able to add such support for M$ OS.
This auto-completion is based on statically generated list off all commands I mean builtin, plugins and scripts.
If someone drops a script when DF is running it will not be included in autocomplete list.
That is most simple and naive implementation with no sub-command support.

Cheers
P

### DIFF
--- a/library/Console-posix.cpp
+++ b/library/Console-posix.cpp
@@ -213,6 +213,10 @@ namespace DFHack
             supported_terminal = false;
             state = con_unclaimed;
         };
+        void update_command_list(const CommandList& cl)
+        {
+            command_list = cl;
+        };
         virtual ~Private()
         {
             //sync();
@@ -542,26 +546,17 @@ namespace DFHack
                 }
                 lock->lock();
                 const int old_cursor = raw_cursor;
-                /* Only autocomplete when the callback is set. It returns < 0 when
-                 * there was an error reading from fd. Otherwise it will return the
-                 * character that should be handled next. */
+
                 if (c == 9)
                 {
-                    /*
-                    if( completionCallback != NULL) {
-                        c = completeLine(fd,prompt,buf,buflen,&len,&pos,cols);
-                        // Return on errors
-                        if (c < 0) return len;
-                        // Read next character when 0
-                        if (c == 0) continue;
-                    }
-                    else
+                    std::string part = toLocaleMB(raw_buffer);
+                    auto command = command_list.lower_bound(part);
+                    if(command != command_list.end() && command->substr(0, part.size()) == part)
                     {
-                        // ignore tab
-                        continue;
+                        raw_buffer = u32string(command->begin(), command->end());
+                        raw_cursor = command->size();
+                        prompt_refresh();
                     }
-                    */
-                    // just ignore tabs
                     continue;
                 }
 
@@ -813,6 +808,7 @@ namespace DFHack
             }
             return raw_buffer.size();
         }
+        
         FILE * dfout_C;
         bool supported_terminal;
         // state variables
@@ -832,6 +828,8 @@ namespace DFHack
         // thread exit mechanism
         int exit_pipe[2];
         fd_set descriptor_set;
+
+        CommandList command_list;
     };
 }
 
@@ -1002,4 +1000,9 @@ bool Console::show()
 {
     //Warmist: don't know if it's possible...
     return false;
+}
+
+void Console::update_command_list(const CommandList& command_list)
+{
+    d->update_command_list(command_list);
 }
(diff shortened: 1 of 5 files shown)


library/dfhack-run.cpp
```diff
@@ -43,6 +43,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <fstream>
 #include <istream>
 #include <string>
+#include <set>
```
lethosor, line 46: Just curious, did you need this to get it to compile on your end?

library/Console-posix.cpp
```diff
@@ -542,26 +546,17 @@ namespace DFHack
...
-                    if( completionCallback != NULL) {
-                        c = completeLine(fd,prompt,buf,buflen,&len,&pos,cols);
-                        // Return on errors
-                        if (c < 0) return len;
-                        // Read next character when 0
-                        if (c == 0) continue;
-                    }
-                    else
+                    std::string part = toLocaleMB(raw_buffer);
+                    auto command = command_list.lower_bound(part);
```
lethosor, line 553: Nice use of lower_bound here

library/Console-posix.cpp
```diff
@@ -542,26 +546,17 @@ namespace DFHack
...
+                    if(command != command_list.end() && command->substr(0, part.size()) == part)
                     {
-                        // ignore tab
-                        continue;
+                        raw_buffer = u32string(command->begin(), command->end());
+                        raw_cursor = command->size();
```
lethosor, line 557: Have you checked what happens if the cursor isn't originally at the end of the line when tab is pressed? (I'm not entirely sure why you'd do that, so weird behavior is fine with me for now, as long as it doesn't crash)

library/Core.cpp
```diff
@@ -512,6 +512,36 @@ static bool try_autocomplete(color_ostream &con, const std::string &first, std::
...
+    }
+    
+    std::map<string, string> scripts;
+    listAllScripts(scripts, true);
+    for (auto iter = scripts.begin(); iter != scripts.end(); ++iter)
+    {
+        commands.insert(iter->first);
+    }
+
+    return commands;
```
lethosor, line 542: This doesn't appear to include aliases - they probably aren't that commonly-used, but `Core::ListAliases()` appears to be a safe way to get a list of them.

PassionateAngler: I will change that in spare time.

library/Console-posix.cpp
PassionateAngler: It does not crash. If cursor is in the middle of command it jumps to the end if command is a mach. If it is not a mach it just stays where it was

library/Console-posix.cpp
PassionateAngler: Thx :)

library/dfhack-run.cpp
PassionateAngler: Yest it does. But I am on:
gcc version 10.2.1 20200723 (Red Hat 10.2.1-1) (GCC) 

I see that you test compilation on 4.8 and it fails. That may be the issue ?

lethosor: Our CI currently uses both GCC 4.8 and 7 (but on Ubuntu). I know some Fedora fixes have been made already for GCC 10, like 7bdf958518d2892ee89a7173224a069c4a2190d8, so I was surprised that this wasn't done already. Thanks for the fix, anyway!

lethosor: Oh, I found the error on Buildmaster: https://buildmaster.lubar.me/executions/execution-details?executionId=507413&noAutoRedirect=y#S87.6619
```
Sphinx parallel build error:
ninja exited with code 1 (failure)
```

This is a known (and intermittent) issue when building the documentation in the GCC 4.8 image, which is also currently stuck on an older version of Python and Sphinx. From https://buildmaster.lubar.me/applications/17/builds/build?buildId=8445, the 64-bit GCC 4.8 build succeeded (the error only happened in the 32-bit build), so I'm not worried about it.

lethosor:
Did you intentionally close this, or was it closed because you deleted your develop branch? (It looks like it still exists at https://github.com/PassionateAngler/dfhack/tree/develop, so I'm not quite sure what happened, actually)

PassionateAngler:
My lack of experience with GitHub happen :) I have closed because I have removed my "develop"branch it and moved to another branch called "autocomplete" in my fork. I can reopen pull request but I did not fixed aliases, actually I did but fix did not work. Probably because I create list of commands before aliases are initialize...

PassionateAngler:
Here it is. https://github.com/PassionateAngler/dfhack/tree/autocomplete

lethosor:
Thanks! It's a good starting point - I'll try to get it (or something like it) into r3.

Issue #1640, opened 2020-09-06
buildingplan: support all building types


myk002:
The next big feature I have planned for quickfort is support for all buildings in buildingplan.

I have a few questions, though, and I'm not on IRC reliably enough to ask them there:

1) How do we match an item against a building construction job item filter?

In order to support all buildings, I'll need to support a wider variety of building items, including generic building materials. buildingplan goes through all free game items and assigns pertinent ones to buildingplan-managed buildings. This is easy to do for the furniture types buildingplan currently supports, but how can we do it generically for any filter returned from buildings.lua? Say the filter says "fire-safe building material", what is the best way to select a matching item in the game world?

Edit: I just discovered dfhack.job.isSuitableItem(). Let me investigate that.
Edit2: that looks like it will work. Nice.

2) How do we assign just one of several items in a construction job?

This might become obvious once I start implementing, but just in case it's difficult, are there any special steps I have to take for assigning items to a suspended well construction job if not all items are available at the same time? Can I assign one item and keep the job suspended? are there any negative consequences of doing that?

I haven't fully designed the data structures and item selection algorithm yet, but my overall plan is this:

- general cleanup of buildingplan code to make the code easier to modify (no logic or API changes) -- move business logic out of header files, split buildingplan-lib.h into buildingplan-planner.h and buildingplan-rooms.h, move gui code back into buildingplan.cpp
- tighten up the API -- make all publicly unused methods private, remove direct access to mutable fields (this will be required for the persistent item filters feature request that was on the forums recently)
- Use lua to allocate buildingplan-created buildings so we get appropriate item filters
- When assigning an item to a construction job, filter both by construction job filters and buildingplan filters
- Generate gui material selection pages for each required item, use getFilters() to get how many items, display serialized filter at top of screen so players know what they’re selecting for, left, right arrow keys to switch between items

myk002:
btw, could you assign this bug to me? I plan to implement it myself, just posting as an issue for tracking purposes.

lethosor:
Assigned + labeled, hopefully appropriately. Not sure on (2) - you might need to add a generic_ref or specific_ref to keep the item "attached" to a job, but that might be something you need to do even if the job isn't suspended. I haven't taken a close look at buildings/jobs in a while, but maybe someone else knows more.

myk002:
> Not sure on (2) - you might need to add a generic_ref or specific_ref to keep the item "attached" to a job, but that might be something you need to do even if the job isn't suspended.

[...]

PR #1641, opened 2020-09-08
reorganize buildingplan code - no logic changes

myk002:
first step in #1640

- no API or logic changes, just moving code around and simplifying structure in preparation for changes to come
- split buildingplan-lib into planner and rooms files
- move business logic from .h files to .cpp files
- move gui code out of buildingplan-lib back into buildingplan.cpp


plugins/fortplan.cpp
```diff
@@ -88,7 +102,7 @@ DFhackCExport command_result plugin_init ( color_ostream &out, vector <PluginCom
 #define DAY_TICKS 1200
 DFhackCExport command_result plugin_onupdate(color_ostream &out)
 {
-    if (Maps::IsValid() && !World::ReadPauseState() && world->frame_counter % (DAY_TICKS/2) == 0)
+    if (Maps::IsValid() && !World::ReadPauseState() && df::global::world->frame_counter % (DAY_TICKS/2) == 0)
```
lethosor, line 105: One existing convention is to use `REQUIRE_GLOBAL(world)`, usually near the `DFHACK_PLUGIN` call, which imports the specified global into the global namespace and also tracks which globals a plugin uses. In practice, recent DFHack versions have all globals available, but using REQUIRE_GLOBAL makes it a bit safer, so I would recommend that. Keeping the `df::global::` prefix isn't strictly necessary with it, but it's fine to have it here. I'm just bringing it up because I noticed it in the diff. (Did the reorganization break some code that was compiling previously?)

myk002: added the REQUIRE_GLOBAL() calls to fortplan. the reorg removed the `using namespace` lines that were in buildingplan-lib.h (`using namespace` in header files can lead to some tricky bugs). I could have added them back in fortplan.cpp, but I just went with the long form since adding a `using namespace df::global` wouldn't have saved much.

lethosor: Yeah, I think the original buildingplan plugin was split up into a couple files in #464 for fortplan's purposes - at the time, it was more intended to be shared code than a proper library. Thanks for cleaning it up!

lethosor (approved):
Looks good! (Ran through a couple basic tests including a save/load cycle)

lethosor:
Whoops, looks like you pushed up a merge after I merged this locally. Will merge again.

PR #1642, opened 2020-09-09
Upgrade clsocket

lethosor:
Opening to test a Windows build, per https://github.com/DFHack/clsocket/pull/17#issuecomment-680125404


lethosor:
It seems to work on Buildmaster, but I would rather hold off on upgrading clsocket until after the next DFHack release (likely 0.47.04-r3) to minimize breakage, and that upgrade would hopefully also include DFHack/clsocket#4.

PR #1645, opened 2020-09-14
use more precise item vectors for building item filters

myk002:
Filled in some vector ids where they were not specified. ANY_GENERIC35 used to be a conglomerate of a bunch of types. Now it doesn't even exist, so using it was effectively not setting a vector id at all. I'm not sure why the more specific vectors weren't used in the first place. Perhaps they didn't exist at the time the code was originally written?


library/lua/dfhack/buildings.lua
```diff
@@ -176,9 +176,9 @@ local building_inputs = {
             vector_id=df.job_item_vector_id.TRACTION_BENCH
         }
     },
-    [df.building_type.Slab] = { { item_type=df.item_type.SLAB } },
-    [df.building_type.NestBox] = { { has_tool_use=df.tool_uses.NEST_BOX, item_type=df.item_type.TOOL } },
-    [df.building_type.Hive] = { { has_tool_use=df.tool_uses.HIVE, item_type=df.item_type.TOOL } },
+    [df.building_type.Slab] = { { item_type=df.item_type.SLAB, vector_id=df.job_item_vector_id.SLAB } },
+    [df.building_type.NestBox] = { { has_tool_use=df.tool_uses.NEST_BOX, item_type=df.item_type.TOOL, vector_id=df.job_item_vector_id.TOOL } },
+    [df.building_type.Hive] = { { has_tool_use=df.tool_uses.HIVE, item_type=df.item_type.TOOL, vector_id=df.job_item_vector_id.TOOL } },
```
lethosor, line 181: These three `vector_id`s (well, two unique ones) are nil:
```
[lua]# ~df.job_item_vector_id.SLAB
nil
[lua]# ~df.job_item_vector_id.TOOL
nil
```
I see comments linking these to ANY_125 and ANY_124 (respectively) in df.item-vectors.xml, but those just point to `items.all`, which I think these were falling back to before anyway (but I'm having a hard time tracking down what actually uses the `vector_id` field...). At any rate, this doesn't hurt anything, since the value of `vector_id` was and is `nil`, but it could be confusing.

lethosor (commented):
The others look good (I tested by comparing the contents of world.items.other[type] with all items in world.items.all.with the corresponding type).
My guess is that ANY_GENERIC35 was renamed to ANY_GENERIC36 (or higher) at some point. This is a relatively common problem that comes up when naming enum items that don't have another distinguishing feature after their value, and is unfortunately harder to catch in Lua.

library/lua/dfhack/buildings.lua
myk002: Ah, I see I was assuming since the OTHER id exists, the vector id must as well. https://github.com/DFHack/df-structures/blob/master/df.item-vectors.xml#L669 mentions that SLAB (and similarly TOOL) is SKIPPED -- what does that mean (I don't see SKIPPED defined as a syntax item in SYNTAX.rst), any do you know why it would be skipped?

I'll revert these three back to undefined. Thank you for finding this!

lethosor: It just means "skipped", like how [EGG](https://github.com/DFHack/df-structures/blob/5756bf8599630891fc460b1509bd0660b71c0809/df.item-vectors.xml#L776) is [later](https://github.com/DFHack/df-structures/blob/5756bf8599630891fc460b1509bd0660b71c0809/df.item-vectors.xml#L744) than it would be if the order of `job_item_vector_id` matched `items_other_id`.

My understanding of `job_item_vector_id` is that it is a rough indicator of what types of items certain jobs require, given that it is only actually used in `job_item::vector_id`. It tends to correspond to `items_other_id` (the names deliberately match on our end when applicable, and sometimes the raw values happen to match too), but there are some entries interspersed that refer to ANY or different vectors altogether. Not sure why DF does it this way. I suspect that for "skipped" item types without an entry in `job_item_vector_id`, DF searches `world.items.all` instead - it would be slower, but any speedups from adding an entry would be relatively small. It's also possible that DF added a value at the end of `job_item_vector_id`, and we simply haven't noticed yet (since it's not used as an index the way `items_other_id` is).

lethosor: As a test, I created an "engrave memorial slab" job, and `job.job_items[0].vector_id` is 125 (job_item_vector_id::ANY_125). 
https://github.com/DFHack/df-structures/blob/5756bf8599630891fc460b1509bd0660b71c0809/df.item-vectors.xml#L772
It's possible that other job types also use `125`, so it can't always be resolved to `world.items.other.SLAB`. @quietust might be able to provide a better explanation.

lethosor (approved):

PR #1650, opened 2020-09-17
Change dependencies to use add_subdirectory()

lethosor:
Intended to fix some Windows issues related to #1620 and #1630.

This now involves all three libraries added in #1620. 

- libzip and xlsxio have DFHack forks with some CMake changes, mostly renaming options to avoid conflicting with each other and DFHack.
- xlsxio has been changed to not be installed by default
- Confirmed that no libraries are installed by default
- Confirmed that libexpat is built statically and linked with xlsxreader
- Some libraries add some extra variables to CMakeCache.txt from "find function"/"have function" checks (e.g. libexpat adds SIZE_T). I don't think these are worth changing.

To do:
- libexpat prints a bunch of stuff at CMake time. Might want to fork just to get rid of this.
- libzip prints CMake warnings because we're not building with any sort of crypto.
- possibly other CMake warnings I've forgotten
- set compiler flags to silence warnings in buildmaster?


lethosor:
I _think_ this is mostly done now. @ragundo and @RosaryMala, if you have a chance, I would greatly appreciate if you could see if this resolves your issues with building on Windows (from #1630).

myk002 (approved):
Unhappy that the library integration caused you so much pain, but this change LGTM

note on the buildmaster failure: I ran into this before. the version of cmake on that image is ancient.

ragundo:
Great!.
I'll try to compile it this afternoon.

RosaryMala:
I'll try compiling it tomorrow and see if it works.

lethosor:
> note on the buildmaster failure: I ran into this before. I suspect it is because the version of cmake on that image is ancient.

Do you happen to remember what you did to fix it?

I looked into this and [add_compile_definitions](https://cmake.org/cmake/help/git-master/command/add_compile_definitions.html) was added in CMake 3.12. The issue is here: https://github.com/DFHack/libzip/blob/4d7582db02a7f40e62f311d33db968460dc3ab1f/CMakeLists.txt#L228-L231
What's strange is that this didn't fail before. There are other `if(MSVC)` checks in the DFHack codebase that are working properly, so the only explanation I can think of is that libzip *wasn't being built with MSVC*. I have no idea what other non-MSVC compiler could be available in the image that would build a working Windows library, but if that's the case, the issue in #1630 was probably present on Buildmaster this whole time, but it happened to be using a compiler that existed and worked.

myk002:
> > note on the buildmaster failure: I ran into this before. I suspect it is because the version of cmake on that image is ancient.
> 
> Do you happen to remember what you did to fix it?

ah, I remember more now. This is the reason I pinned libzip to the commit that I did. revert the update and that should fix it.

lethosor:
Ah, thanks! I was wondering why it was pinned to that instead of a release. I've forked libzip already, so assuming there's an easy workaround, I can fix it, and possibly submit a patch upstream.

RosaryMala:
I tested this on my windows machine, it compiles fine.

lethosor:
Thanks!
Unfortunately Buildmaster is unhappy:
```
FAILED: plugins/xlsxreaderMD.plug.dll
: && /opt/cmake/bin/cmake -E vs_link_dll --intdir=plugins/CMakeFiles/xlsxreader.dir --manifests  -- /usr/local/bin/link /nologo plugins/CMakeFiles/xlsxreader.dir/xlsxreader.cpp.obj  /out:plugins/xlsxreaderMD.plug.dll /implib:plugins/xlsxreaderMD.lib /pdb:plugins/xlsxreaderMD.pdb /dll /version:0.0 /machine:X86 /INCREMENTAL:NO  library/SDLMD.lib library/dfhack-versionMD.lib depends/lua/lua.lib depends/xlsxio/xlsxio_readMD.lib depends/libzip/lib/zipMD.lib depends/libexpat/expat/libexpatMD.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib  && :
LINK: command "/usr/local/bin/link /nologo plugins/CMakeFiles/xlsxreader.dir/xlsxreader.cpp.obj /out:plugins/xlsxreaderMD.plug.dll /implib:plugins/xlsxreaderMD.lib /pdb:plugins/xlsxreaderMD.pdb /dll /version:0.0 /machine:X86 /INCREMENTAL:NO library/SDLMD.lib library/dfhack-versionMD.lib depends/lua/lua.lib depends/xlsxio/xlsxio_readMD.lib depends/libzip/lib/zipMD.lib depends/libexpat/expat/libexpatMD.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:plugins/xlsxreaderMD.plug.dll.manifest" failed (exit code 157) with the following output:
LINK : fatal error LNK1181: cannot open input file 'depends/libzip/lib/zipMD.lib'
```

Can you see what files you have in depends/libzip/lib (in your build folder)? Unsure what the `MD` suffix means, but this might be a difference between release and debug builds.

RosaryMala:
For me, that lives in `depends\libzip\lib\RelWithDebInfo`

lethosor:
Interesting. Do you have `depends/xlsxio/xlsxio_readMD.lib` in your build folder or is that also in a `RelWithDebInfo` subfolder?

Not sure why CMake doesn't know where the build library wound up, but I'll see what I can change.

RosaryMala:
Yup, they're also in a RelWithDebInfo subfolder, just like all the other libs.

lethosor:
Ok, thanks, so CMake must be accounting for that component correctly (maybe the subfolder was added in a newer CMake version than what Buildmaster uses?).
Looks like Buildmaster is building a *shared* libzip for some reason, so it's the extension that's incorrect (well, and the file type):
```
[230/639] Linking C shared library depends/libzip/lib/zipMD.dll
```
You do have a static libzip built locally, right?
The other libraries are static on Buildmaster, so maybe it's another CMake incompatibility.

RosaryMala:
Yeah, it's built locally, as far as I can tell.

lethosor:
libexpat and xlsxio both call `add_library` with `STATIC` when a static library is requested - I couldn't figure out how libzip was attempting to set up a static library, so I patched its `add_library` call to do the same. That seems to have fixed Buildmaster!

https://github.com/DFHack/dfhack/pull/1650/commits/65b8071941f60b4a4dbb83c0e91c028c4da71225 was another fix that Ragundo helped with significantly - I'm not sure if this would have failed to compile on a proper Windows system before this PR as well.

PR #1652, opened 2020-09-19
enumnick: command added to 'zone' plugin and other QoL improvements

PassionateAngler:
Small improvements for 'zone' plugin.
I have added enumnick command which create nick-s for creatures from given prefix and position on the list.
I have also changed maxage/minage so they accept float.
uinfo displays "slaughter" flag.
uinfo displays "Matched creatures" i.e number of creatures matched by filters at the end of the list.

### DIFF
--- a/plugins/zone.cpp
+++ b/plugins/zone.cpp
@@ -133,6 +133,7 @@ const string zone_help =
     "                 with filters named units are ignored unless specified.\n"
     "  unassign     - unassign selected creature(s) from zone or cage\n"
     "  nick         - give unit(s) nicknames (e.g. all units in a cage)\n"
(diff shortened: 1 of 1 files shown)


lethosor (approved):
Nice! I can deal with docs updates.

plugins/zone.cpp
```diff
@@ -1489,6 +1493,8 @@ command_result df_zone (color_ostream &out, vector <string> & parameters)
     bool cagezone_assign = false;
     bool nick_set = false;
     string target_nick;
+    bool enum_nick = true;
```
lethosor, line 1496: I suspect this needed to be initialized to `false` - see #1709. I'm working on a fix, but this looks like a likely cause at the moment.

PR #1655, opened 2020-09-20
ignore libzip build artifact

myk002:


lethosor:
Actually, I meant to patch libzip to put this in CMAKE_CURRENT_BINARY_DIR (which would be under `build/depends`) instead of CMAKE_BINARY_DIR. Not quite sure if changing that alone is enough, though - I'll try it.

lethosor:
https://github.com/DFHack/libzip/commit/be76fa5086bfe6b1a5e83c9855e39f98edc1f066 seems to work for me - there were only two places that referred to this file - so hopefully it will work on Windows as well. Thanks for catching this, though!

PR #1656, opened 2020-09-20
document generated isEnabled and setEnabled lua functions for plugins

myk002:


docs/Lua API.rst
```diff
@@ -3616,8 +3616,9 @@ buildingplan
 
 Native functions:
 
-* ``bool isPlannableBuilding(df::building_type type)`` returns whether the building type is handled by buildingplan
-* ``void addPlannedBuilding(df::building *bld)`` suspends the building jobs and adds the building to the monitor list
+* ``bool isEnabled()`` returns whether the plugin is enabled.
```
lethosor, line 3619: I'm already seeing an `isEnabled()` (and `setEnabled()`) function when I `require('plugins.buildingplan')` on develop, and it looks like those functions are created [here](https://github.com/DFHack/dfhack/blob/d43bf0ce4474a8e994d59d552469d16dbd79f4a6/library/PluginManager.cpp#L774-L785) for all plugins that support the enable/disable API. I assume that you've tested this and it doesn't break anything, but I don't think we need to add this specifically to buildingplan.

I _do_ think it should go in the documentation, since it evidently wasn't clear that this was happening. Maybe the right place to mention them is under the "Plugins" header, before the first plugin?

myk002: ha, I didn't think to test isEnabled() before I wrote the function, so I didn't notice that it was already there : P

I'll revert the code changes in this PR and replace them with doc changes.

lethosor: I just noticed because I tested this PR with `~require("plugins.buildingplan")` in Lua, and noticed that `setEnabled` was already there. Then tried with another plugin...
In short, there are definitely gaps in our documentation, so thanks for adjusting this PR!

lethosor (approved):

PR #1657, opened 2020-09-24
buildingplan: step1: refactor gui-related logic and state out of the lib

myk002:
#1640

- no behavior changes
- move the planmode_enabled map and quickfort-related bits over to buildingplan.cpp so it can live with the rest of the gui code.
- don't bother initing the planmode_enabled map -- keys get created when they are accessed and the default value is false
- replace saved_planmode with an || quickfort_mode -- same effect, less mess (I guess there is a slight user-visible behavior change here. building types that are manually enabled while quickfort mode is active will no longer be lost when quickfort mode is disabled again)
- fix misspelling of "plannable" in API


myk002:
There are several more PRs in this chain, but I don't really want to make you have to manually retest each one. How about if I open PRs for all the steps so you only have to test the final one, but we can review the code step by step? I'll open them as WIP and we can discuss.

lethosor:
Sorry I haven't gotten to this one yet. Looks like a lot of stuff in the pipeline here, and I was thinking of releasing r3 with quickfort somewhat soon. How much of this does quickfort depend on? Would it be okay to merge this PR and then have the rest in r4, or should they all be combined? 

My main concern is breaking buildingplan right before a release, but I assume you've tested it pretty well. I could always put out a call for more testing on the forums, but I'm unsure how much attention that would get.

myk002:
These changes solve the number one usability problem with quickfort -- buildings disappearing after a #build blueprint is run due to insufficient materials. I've documented the workaround -- run `quickfort orders` and prepare all materials before applying a #build blueprint -- but nobody will pay attention to that and I foresee user unhappiness when the issue bites them.

that being said, this change to buildingplan is almost a total rewrite of the entire plugin and the risk that it breaks something is definitely not zero. I have sanity checked it, in particular to ensure that the buildings it previously supported still all work, but a call for community testing would be welcome. it is totally reasonable to put this off until the next release.

lethosor:
Ok, I think my plan for now will be to merge other outstanding PRs, put up r3 (hopefully around this weekend), then take a closer look at these. That might change as time allows.

In any case, thank you for tackling this! It will certainly be appreciated.

lethosor (approved):
Yep, looks like fortplan didn't need this to be shared - looks good

lethosor:
I got rid of the merges that were previously after 1c2761f to clean up the history a bit, and force-pushed to this branch so that GitHub would realize that I'm merging the PR (as opposed to just a subset) - hopefully that's all right. As for the other PRs, I'd prefer not to have quite as many merge commits as in #1665, for instance. I'm not sure what the best way to get prune them while keeping the remaining commits the same across PRs is (e.g. a simple `git rebase develop` in each PR would lead to 4 different versions of 4d7f4d80ad478d2735a815aec5a5edb2dea665d0). I could attempt to come up with a strategy another day, though, if you like.

Looks like GitHub Actions is having some package mirror issues, which I've seen maybe once or twice before - not sure why it's consistently just the `build` and `lint` jobs, but since it was passing before, I'm going to ignore it for now.

myk002:
Absolutely fine to clean things up, but if I may ask, why not use squash and merge when you merge the pr? Wouldn't that reduce the pr to a single commit in develop?

I can squash the remaining PRs manually, or, rather, I can make new branches with squashed commits (I actually did do that for the current branches, but then there were more commits...)

lethosor:
I don't mind keeping history - I would just prefer not to keep the extra merges around. I also tend to merge locally, so I'm not sure if GitHub would recognize a squash merge that isn't done through the web UI. I could try squashing the next PR to see what happens, although it may end up requiring the other PRs to be updated to resolve merge conflicts that wouldn't exist otherwise.

myk002:
Understood. I'll pre-squash the PRs, making sure to call out important individual commits in the combined commit.

These PRs are particularly merge-y since every time I committed a fix to an earlier step, I had to propagate the change up the stack for integration testing. This kind of situation should be rare, but I'll try to be mindful of "merge commit spam" in the future.

PR #1658, opened 2020-09-24
Normalize search

### LINKED ISSUE #1598: search: normalize non-ASCII characters when matching
Search strings are already normalized to lowercase, so searching for "dumat" will match "Dumat Timnärbomrek". Since DF doesn't really offer options for typing special characters on keyboards without them, it would be nice if "timnar" (a instead of ä) would also match.
(This should probably be added to MiscUtils or something for ease of use in other plugins.)

GitOnUp:
Normalize searches so that words/names with accented characters and diphthongs are more easily found with a standard keyboard.

This will fix #1598

### DIFF
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -128,6 +128,45 @@ std::string toLower(const std::string &str)
     return rv;
 }
 
+static const char *normalized_table[256] = {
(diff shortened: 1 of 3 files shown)


lethosor (approved):
Works well! My plans for this next are to add it to the `stocks` plugin (and maybe others that could use it?) and expose it to Lua, similar to df2utf and related functions. I'm not sure what Lua scripts, if any, currently need this, or if there's a good centralized place to use it, but having it available will help.
Thanks for tackling this!

PR #1660, opened 2020-09-29
fix bug: buckets are missing lye_milk_free attribute in item matcher

myk002:
otherwise the bucket for wells (and asheries and dyer's workshops) can never be matched


lethosor (approved):
Wow, nice catch!

This is probably worth mentioning in the changelog, so I'll do that. I considered adding a test for this too, but even if this function were exposed to Lua (which it doesn't appear to be), a test would probably have to create an item specifically designed to pass/fail the test at this point, so it's probably not worth doing until we have the ability to test with actual DF objects.

PR #1662, opened 2020-10-05
buildingplan: step2: prep buildingplan for core algorithm changes

myk002:
#1640 
Lots of refactoring and reorganizing, with only cosmetic player-visible changes.

- show quickfort mode hotlkey label regardless of whether the current building type has buildingplan enabled. before, it was only shown after the user enabled buildingplan for the current building. this eliminates the extra step when enabling quickfort mode, which force-enables all building types.
- changed signature of lua-exported isPlannableBuilding to take subtype
  and custom type in addition to building type. this is only used by
  quickfort, and it already sends all three params in preparation for
  this change
- added lua-exported scheduleCycle(), which is like doCycle(), but only
  takes effect on the next non-paused frame. this lets quickfort
  run only one buildingplan cycle regardless of how many #build
  blueprints were run
- declared a few dfhack library methods and params const so buildingplan
  could call them from const methods
- converted buildingplan internal debug logging fn to have a printf api
- reshaped buildingplan-planner API and refactored implementation in
  preparation for upcoming core algorithm changes for supporing all
  building types (no externally-visible functionality changes)
  - changed df::building_type params to type, subtype, custom tuple keys
  - introduced capability to return multiple filters per building type
    (though the current buildings all only have one filter per)
- split monolith hook functions in buildingplan.cpp into one per scope.
  this significantly cleans up the code and preps the hooks to handle
  iterating through multiple item filters.
- got rid of send_key function and replaced with better reporting of
  whether keys have been handled


lethosor:
> * removed text that showed up if you used the wrong hotkeys. no other
>   dfhack screen does this, and it seems unneeded. can add back if others
>   think otherwise, though

Can you point me to this? I haven't been able to locate it in the diff, at least.

myk002:
> > * removed text that showed up if you used the wrong hotkeys
> 
> Can you point me to this? I haven't been able to locate it in the diff, at least.

show_help and its related logic
https://github.com/DFHack/dfhack/pull/1662/files#diff-1d44d0616ca136a4bcd031fd4f814388L30

lethosor:
Ah, I see. I would probably keep it in, if it's not too hard to add back. I assume there's a good reason for using all uppercase keys in that screen, which is a bit unusual.

I actually just tracked this down in the original buildingplan thread: http://www.bay12forums.com/smf/index.php?topic=121858.msg4337947#msg4337947
I'm a little confused whether this only occurred when using the external Quickfort, but the core DF bindings haven't really changed since then.

lethosor:
The GCC 4.8 build appears to be failing because it can't hash building_type.  https://github.com/DFHack/dfhack/pull/1662/commits/5087b6d2ae88efcac5efef7943daddd5d8cba754#diff-089cb08a6cf4d84b50aafbc7e9f04571R122 is the closest related thing I noticed in the first build that failed. Maybe some other plugins using unordered maps have addressed this already?

myk002:
> Ah, I see. I would probably keep it in, if it's not too hard to add back. I assume there's a good reason for using all uppercase keys in that screen, which is a bit unusual.

kk will do.

> I actually just tracked this down in the original buildingplan thread: http://www.bay12forums.com/smf/index.php?topic=121858.msg4337947#msg4337947
> I'm a little confused whether this only occurred when using the external Quickfort, but the core DF bindings haven't really changed since then.

Yeah, why would 'f' or 'd' even be sent on that screen.. you know, I'm starting to develop a theory for why dfhack-added hotkeys can't be scripted by sendkeys. I'm suspecting a race condition between when a key is processed and when an input handling hook is added to the chain. Is the hook management code handled synchronously with the input feed?

Anyway, key conflicts are still an issue. Buildingplan uses 'q' and 'w' to set min quality. Rollers use them to set speed. You can work around the conflict by temporarily disabling buildingplan for rollers, setting the speed, then re-enabling, setting the min quality, then enter to place. That won't work very well for Python quickfort, though. I don't really want to change the q/w hotkeys that people are used to, but a conflict is a conflict. Maybe just change the min quality hotkeys for rollers? Change all min quality hotkeys to 'A' and 'S'? What are your thoughts?

myk002:
> The GCC 4.8 build appears to be failing because it can't hash building_type.

I thought I fixed this (and the init order warnings) in https://github.com/DFHack/dfhack/pull/1662/commits/9116b331b7805b2902bf623daa4d7464e96c2b48

did it not pick up the latest commit? I'll try merging upstream/develop and force a rebuild to see.

myk002:
Ok, help text restored and build fixed

lethosor:
> I'm starting to develop a theory for why dfhack-added hotkeys can't be scripted by sendkeys. I'm suspecting a race condition between when a key is processed and when an input handling hook thinks it is on the correct screen. Could that be possible?

Is this specifically for buildingplan hotkeys?
It's certainly possible that a specific plugin is doing something weird, but most plugins use CUSTOM_ keys, which should be possible to simulate with e.g. `devel/send-key CUSTOM_X` - that would be a bug if not.

Not sure on the GCC issue - I could probably add GCC 4.8 to GitHub actions, which would at least enable quicker feedback.

myk002:
> > I'm starting to develop a theory
> 
> Is this specifically for buildingplan hotkeys?

I've actually run into it most often while searching in the stockpiles menu. F. ex. disabling all tallow. Search for tallow, disable all. This tends to not work when scripted. In the aliases-common file the alias for disabling tallow expands to down down enter down down enter....

lethosor:
Ah, the search plugin has some complicated logic, but it ultimately checks for the key returned by `Screen::charToKey('s')` to initiate searches (unless overridden), which returns STRING_A115. So I was able to search for "t" in that screen by sending STRING_A115 followed by STRING_A116 with `devel/send-key`. I'm not sure if that helps.

myk002:
Well, when I tested it, it always worked when the screen was already loaded, but it would miss the search hotkey when the entire stockpile configuration sequence was sent and df was flicking through multiple screens. It's on my list to investigate. I'll see if I can come up with something reproducible

lethosor:
Ah, ok. I should have mentioned that I didn't see anything obvious that would cause that when I was looking through the search plugin - there doesn't seem to be anything that would ignore the first run of `feed()`, at least. There is some initialization (`init()`) called on the first render, which needs to run before search input will be accepted, but it's also run on the first call to `feed()`. There could definitely be some source of a delay like this elsewhere in the search plugin, though.

Between this and the `{togglesequence2 811}` I see in aliases-common.txt (which will leave out some autogenerated creatures' materials in larger worlds, although likely nobody will ever obtain those), this is a situation where I think a more robust solution would be to set the stockpile settings directly instead of through keyboard input. But I understand why you're doing it this way.

lethosor:
> ~Not sure on the GCC issue~ - I could probably add GCC 4.8 to GitHub actions, which would at least enable quicker feedback.

Not entirely related, but since I mentioned it here: as of https://github.com/DFHack/dfhack/commit/c1e5ecf77393ce72f7c72272f7de9866c9393523, GitHub actions runs tests with three different GCC versions (in parallel): https://github.com/DFHack/dfhack/actions/runs/294754044

myk002:
> > ~Not sure on the GCC issue~ - I could probably add GCC 4.8 to GitHub actions, which would at least enable quicker feedback.
> 
> GitHub actions runs tests with three different GCC versions

Nice! That will make future changes much less painful. Thank you!

plugins/buildingplan.cpp
```diff
@@ -670,13 +693,17 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
     return CR_OK;
 }
 
+static bool cycle_requested = false;
+
 #define DAY_TICKS 1200
 DFhackCExport command_result plugin_onupdate(color_ostream &)
 {
-    if (Maps::IsValid() && !World::ReadPauseState() && world->frame_counter % (DAY_TICKS/2) == 0)
+    if (Maps::IsValid() && !World::ReadPauseState()
```
myk002, line 701: I've noticed that World::ReadPauseState() returns true when you hit space to pause the game and then go into query mode, but it returns false when you go directly into query mode and the game pauses as a side effect. Is there a good way to tell when the game is actually paused?

lethosor: I think any ui.main.mode value other than Default (or any viewscreen other than viewscreen_dwarfmodest) pauses the game. Unsure if this is wrapped in a library function anywhere.

myk002: ok. I'll investigate this and improve pause detection in a future PR.

myk002:
replaced by pre-squashed #1673

PR #1663, opened 2020-10-05
WIP: buildingplan: step3: construct buildings from lua

myk002:
#1640 
Replace C++ building construction code with lua constructBuilding so we can get the proper job_item filters set. these filters will be used when we replace the core buildingplan algorithm in the next PR.

Code can be reviewed after #1662 is merged


myk002:
replaced by #1674

PR #1664, opened 2020-10-05
WIP: buildingplan: step4: generalize core buildingplan algorithm to handle all building types

myk002:
#1640 
but restrict to only the currently supported set so we can still assume only one filter is required for each building.

Code can be reviewed after #1663 is merged.

changes:
- update buildingplan plugin version to 2.0
- new serialization format for planned buildings
- old persistent data is automatically migrated to new format on load
- algorithm now respects job_item filters; items must match job_item filter and buildingplan ItemFilter
- more invalid items are now filtered out, like items encased in ice. are there any others we should be checking (see BadFlags struct)
- items are sorted before job is unsuspended so final item ordering is correct regardless of what order the items were matched and attached
- fixes memory leak when the building needs an architect design
- allows artifacts to be matched (ItemFilter defaults now top out at Masterful -- Artifact is selectable but must be manually specified)
- add gui to switch between items for buildings that require multiple item types


plugins/buildingplan-planner.cpp
```diff
@@ -1,552 +1,922 @@
...
+        // TODO: make this configurable
+        && !(item->getType() == df::item_type::BOX && item->isBag());
 }
 
-bool Planner::inQuickFortMode() { return quickfort_mode; }
+static bool matchesFilters(df::item * item,
+                           df::job_item * job_item,
+                           const ItemFilter & item_filter)
+{
+    if (job_item->item_type > -1 && job_item->item_type != item->getType())
```
myk002, line 707: these checks feel like they should be part of isSuitableItem(), but the function signature would need to be changed. Is there any reason basic attributes like item type are *not* checked in isSuitableItem()? or does isSuitableItem() assume that the item vector will always be set correctly, and that all items in that vector will have the same type?

PR #1665, opened 2020-10-05
WIP: buildingplan: step5: support all building types

### LINKED ISSUE #1640: buildingplan: support all building types

myk002:
Closes #1640 

Allow buildingplan to handle all building types, update the docs, and add in any extra fixes to ensure all the new types work correctly.

Code can be reviewed after #1664 is merged.

There is a keybinding conflict with rollers that we should address. buildingplan takes over interpretation of 'q' and 'w' to adjust the min quality filter, and the roller speed can never be set. You can temporarily disable buildingplan for rollers, set the speed, enable buildingplan, and place the building, but that is a poor user experience. It might also break python quickfort blueprints that place rollers.

I don't really want to change the q/w hotkeys that people are used to, but a conflict is a conflict. Maybe just change the min quality hotkeys for rollers? Change all min quality hotkeys to 'A' and 'S'? What are your thoughts?


myk002:
replaced by #1676

PR #1661, opened 2020-10-04
Integrate quickfort docs into main docs

lethosor:
Closes #1653

PR #1666, opened 2020-10-07
update quickfort docs

myk002:
default dig
notes blueprints
buildingplan

### DIFF
--- a/docs/guides/quickfort-user-guide.rst
+++ b/docs/guides/quickfort-user-guide.rst
@@ -139,8 +139,11 @@ in DF. The following "mode" keywords are understood:
    zone    Activity zones menu (i)
    query   Set building tasks/prefs menu (q)
 
-There are also "meta" blueprints, but we'll talk about those
-`later <#meta-blueprints>`__.
+If no modeline appears in the first cell, the file or sheet is interpreted as a
+``#dig`` blueprint.
+
+There are also "meta" and "notes" blueprints, but we'll talk about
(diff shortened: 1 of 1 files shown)


lethosor (approved):
This reminded me that I wanted to switch to "proper" RST section links in #1661 instead of raw links to sections (starting with `#`). Switching would have a couple advantages, like triggering errors if a section name changes and working in the PDF build (I think).

I'll deal with that once I get this and #1667 (and any others) merged in, though.

myk002:
> I'll deal with that once I get this and #1667 (and any others) merged in, though.

No more major doc updates planned in the immediate future.

lethosor:
Ok, I took care of that in 2fdd6f7a (and corrected `messages()` to `message()` in two places, which is hopefully correct - only the latter appears in the source code).

myk002:
Thanks! And yes, "message" is correct.

lethosor:
Another semi-related point: I noticed that code blocks were all being highlighted, so I made some changes - code blocks use a custom syntax that basically just recognizes `#` as a prefix for comments. This doesn't affect the Quickfort guide much, but if you want these snippets to all be just plain text, either a `.. code-block:: text` directive in front of each or `.. highlight:: text` at the top of the page would do it. (I didn't see any actual code blocks in this page, but I might have missed one.)

image

myk002:
I'll keep that in mind, but just from the example you posted above, I'm thinking a bit of color actually makes the blocks a little easier to read : )

PR #1667, opened 2020-10-07
quickfort docs: dreamfort case study

myk002:
- add the dreamfort case study to the quickfort user guide
- fix one bad link and some missing text around labels (<word> syntax was lost by the doc conversion process)
- make sure the guide is actually built by sphinx (the guides dir was missing from the build path)

I learned in the PR that urls can be split over multiple lines in `label <url>`__ format and they still get resembled properly. Not in .. _label: url format, though.


lethosor (approved):
Nice!

PR #1676, opened 2020-10-16
[buildingplan] step5: support all building types

### LINKED ISSUE #1640: buildingplan: support all building types

myk002:
Closes #1640

Allow buildingplan to handle all building types, update the docs, and add in little extra fixes to ensure all the new types work correctly.

Code can be reviewed after #1675 is merged.

There is a keybinding conflict with rollers that we should address. buildingplan takes over interpretation of 'q' and 'w' to adjust the min quality filter when planning mode is enabled for rollers, and then the roller speed can never be set. You can temporarily disable buildingplan for rollers, set the speed, enable buildingplan, and place the building, but that is a poor user experience. It might also break python quickfort blueprints that place rollers.

I don't really want to change the q/w hotkeys that people are used to, but a conflict is a conflict. Maybe just change the min quality hotkeys for rollers? Change all min quality hotkeys to 'A' and 'S'? What are your thoughts?

Edit: solved by changing keybindings for min and max quality.

There is also a confusing UX interaction with the automaterial plugin for constructions. currently, if planning mode is enabled, buildingplan takes precedence and places a single tile of the construction on {Enter}, regardless of what settings are set for automaterial (e.g. box select). I think the solution here is:
- for automaterial-managed constructions, let automat handle placement (that is, buildingplan doesn't intercept the {Enter} key if automaterial is enabled)
- if buildingplan planning mode is enabled for the construction type, automaterial will
  - not show the "Auto Mat-select", "Reselect Type", or "Open Placement" options
  - construct the buildings with filters instead of showing its material selection screen
  - route the building back through buildingplan to let it select materials according to its filter

But I think that can be tackled in a future PR (after I hear your thoughts on this)

### DIFF
--- a/data/blueprints/library/dreamfort.csv
+++ b/data/blueprints/library/dreamfort.csv
@@ -2,26 +2,28 @@
 Welcome to the Dreamfort Walkthrough!
 It can be difficult applying a set of blueprints that you did not write yourself. This walkthrough will guide you through the high-level steps of building Dreamfort.
 ""
-"The final fort will have a walled-in area on the surface for livestock, trading, and military. Be sure to bring some blocks with you for the initial set of workshops! One z-level down is the farming level, with related workshops and vents up to the surface for miasma control. The farming level also has a miniature dining hall and dormitory for use until you get the services and apartment levels set up."
+"The final fort will have a walled-in area on the surface for livestock, trading, and military. Be sure to bring some blocks with you for the initial set of workshops! One z-level down is the farming level, with related workshops and vents up to the surface for miasma control. The farming level also has a miniature dining hall and dormitory for use until you get the services and beds levels set up."
 ""
 "Beyond those two, the other layers can be built in any order, at any z-level, according to your preference and the layout peculiarities of your embark site:"
 "- The industry level has a compact, but complete set of workshops and stockpiles (minus what is already provided on the farming level)"
 "- The services level has dining, hospital, and justice services, including a well system. This level is 4 z-levels deep."
-"- The guildhall level has many rooms for building libraries, temples, and guildhalls"
+"- The guildhall level has many empty rooms for building libraries, temples, and guildhalls"
 - The suites level has fancy rooms for your nobles
 - The apartments level(s) has small but well-furnished bedrooms for your other dwarves
 ""
+"Run the ""help"" blueprints for each level for more details."
+""
 "Dreamfort has a central staircase-based design. Choose a tile for your staircase on the surface in a nice, flat area. For all blueprints, the cursor will start on this tile on the z-level where you want to apply the blueprint."
 ""
-"Blueprints that require manual steps (like 'assign animals to pasture') will leave a message telling you so after you apply them. Blueprints will also leave messages with hints when you might want to run ""quickfort orders"" on specific blueprints to start manufacturing needed items."
+"Blueprints that require manual steps (like 'assign minecart to hauling route') will leave a message telling you so after you apply them. Blueprints will also leave messages with hints when you might want to run ""quickfort orders"" on specific blueprints to start manufacturing needed items."
 ""
 "Directly after embark, apply /surface1 on a flat area on the surface and /industry1 somewhere underground (but not immediately below the surface -- that will be for the farming level). Work your way through the steps for those levels: apply /surface2 when /surface1 is done, /industry2 when /industry1 is done, etc. Once you channel out parts of the surface with /surface3, you can start the farming sequence with /farming1. You can start the services, guildhall, suites, and apartments sequences as your fort needs those levels."
 ""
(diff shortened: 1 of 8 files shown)


lethosor:
I played with this a bit after looking at #1675. Looks awesome!

> There is a keybinding conflict with rollers that we should address. buildingplan takes over interpretation of 'q' and 'w' to adjust the min quality filter when planning mode is enabled for rollers, and then the roller speed can never be set. You can temporarily disable buildingplan for rollers, set the speed, enable buildingplan, and place the building, but that is a poor user experience. It might also break python quickfort blueprints that place rollers.

> I don't really want to change the q/w hotkeys that people are used to, but a conflict is a conflict. Maybe just change the min quality hotkeys for rollers? Change all min quality hotkeys to 'A' and 'S'? What are your thoughts?

Yeah, I would be okay with `a` and `s`. Not sure if there are other special cases where _those_ conflict, though. I suppose one option would be to accept q/w *and* a/s for most building types, unless there is a conflict.

In any case, I think the max quality keys should also match (a/s and Q/W seem confusing to me, particularly since their positions on a QWERTY keyboard don't match their on-screen positioning). a/s and z/x could also be an option, although that opens the door to more possible conflicts. Combinations with modifiers (Shift in this case) are probably less likely to result in conflicts.

I guess I don't have a strong suggestion, but I also doubt many people have memorized these specific key combinations, and they're displayed on-screen, so I would be fine with changing them entirely if that's easiest. (I'm also not sure if you've already made code changes since writing up the PR, so apologies if I missed something.)

myk002:
> I played with this a bit after looking at #1675. Looks awesome!

Thanks :D I'm really happy with how it came out.

> Combinations with modifiers (Shift in this case) are probably less likely to result in conflicts.
> [...]
> I would be fine with changing them entirely if that's easiest.

ok, I think I'll go with {Shift}q and {Shift}w for adjusting max quality and {Shift}a and {Shift}s for adjusting min quality, then. Just 'a' and 's' conflict with trackstop speed and roller direction, respectively.

lethosor (approved):

lethosor:
As for the automaterial conflict: communicating between plugins is a bit tricky. Interposed vmethods do support priorities, though (`IMPLEMENT_VMETHOD_INTERPOSE_PRIO`), so you might be able to use that to run certain logic before/after automaterial. Not sure how much that would help.

I'll try to take a closer look tomorrow at how automaterial is dealing with this - I'm assuming it didn't support planning constructions without the necessary materials before, right?

myk002:
> As for the automaterial conflict: communicating between plugins is a bit tricky.

I was thinking of using the Lua layer for communication -- checking to see what's enabled on both sides, automat registering the building with buildingplan, etc.

> I'll try to take a closer look tomorrow at how automaterial is dealing with this - I'm assuming it didn't support planning constructions without the necessary materials before, right?

It currently uses the UI to build, with an alternate path through Buildings::constructWithItems() for handling hollow rectangles. My preliminary plan was to always construct via the API so we have more control over it. But we'll see. I haven't studied the code very deeply yet.

lethosor:
Yeah, communicating through Lua could definitely work. I was thinking of past inter-plugin communication attempts where one or both of the plugins haven't exposed Lua functionality.

PR #1683, opened 2020-10-23
Add filter support for bookcases, display cases, and offering places

myk002:
Code added to https://github.com/DFHack/dfhack/pull/1675#discussion_r506725773 for applying these filters.

changelog entry for quickfort attached to DFHack/scripts#212 (though technically buildingplan and anything else that directly or indirectly uses dfhack.buildings.getFiltersByType() will benefit)


lethosor (approved):
They all work! Looks like the refactor in #1676 and co really paid off - thanks again.

I also tested hives for fun, which worked, once I realized that I needed a beekeeper to built them...

PR #1684, opened 2020-10-23
[buildingplan] global settings and building material filters

myk002:
This PR can be reviewed after #1682 is merged
￼
Migrate quickfort's buildings_use_blocks functionality to buildingplan, where it fits in with buildingplan's other material selection functionality.

- remove buildings_use_blocks setting from quickfort config file
- add a new Buildingplan Global Settings dialog to house global settings
- move Quickfort Mode (for legacy Python Quickfort) into that dialog
- add four settings to control how generic building materials are matched:
  - blocks
  - boulders
  - logs
  - bars
- ajust the buildingplan algorithm to register duplicate tasks for building material item filters, one for each type. since we track how many items we've matched for a filter, the first matched item will "win" and the extras will get detected as invalid and popped off the queue.
- ensure boulders, logs, and bars are scanned last, and in that order
- more global settings planned for the future! see http://www.bay12forums.com/smf/index.php?topic=176889.msg8202679#msg8202679

### DIFF
--- a/plugins/lua/buildingplan.lua
+++ b/plugins/lua/buildingplan.lua
@@ -194,6 +194,7 @@ function GlobalSettings:init()
         self:make_setting_label_token('Bars', 'CUSTOM_R', 'bars', 10),
             self:make_setting_value_token('bars'), '\n',
         '  Changes to these settings will be applied to newly-planned buildings.\n',
+        '  If no types are enabled above, then any building material is allowed.\n',
         '\n',
         self:make_setting_label_token('Legacy Quickfort Mode', 'CUSTOM_F',
                                       'quickfort_mode', 23),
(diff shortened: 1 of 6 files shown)


plugins/lua/buildingplan.lua
```diff
@@ -92,4 +94,114 @@ function construct_building_from_ui_state()
...
+    if self.settings[name] then return 'On' end
+    return 'Off'
+end
+
+function GlobalSettings:is_setting_enabled(name)
+    return self.settings[name]
+end
+
+function GlobalSettings:make_setting_label_token(text, key, name, width)
+    return {text=text, key=key, key_sep=': ', key_pen=COLOR_GREEN,
```
lethosor, line 137: COLOR_LIGHTGREEN is usually used for keys, so I would use that (some DFHack bindings also use COLOR_LIGHTRED, but that's less necessary when the entire screen is implemented by DFHack)

lethosor (approved):

plugins/lua/buildingplan.lua
myk002: Ack. Thanks for fixing!

PR #1686, opened 2020-10-25
[quickfort] Add more blueprints to the dfhack blueprint library

myk002:
~~WIP for now since I'm still evaluating which other blueprints to port and want to discuss library distribution issues with PeridexisErrant.~~
I'll sync with PeridexisErrant later. I think I'm done adding blueprints for now and this shouldn't block the -r4 release.

@PeridexisErrant - I'm thinking now maybe it's better to keep the two blueprints directories in the Starter Pack separate. As I port more of the Starter Pack blueprints over, there will be more and more duplication, which would be confusing for DFHack quickfort users looking through the library.

mostly ported from the Starter Pack blueprint library, reformatted for clarity and repackaged so all related blueprints are in one file. Also some blueprint bugs fixed (like incorrect start positions). Most Starter Pack blueprints are just #dig blueprints, leaving the furnishing and configuration up to the player. I created new #build and #query blueprints to make the existing #dig blueprints more useful.


lethosor:
What are the `.png` files used for? I'm not seeing references to them in documentation - are they just intended for users looking in the blueprints folder?

myk002:
Yeah, I'm ambivalent about adding the pngs. My semi-goal is to populate our library with the best stuff from the starter pack blueprints library (which is mostly dig blueprints), fix the bugs, and add accompanying build and query blueprints. I added the pngs because the starter pack has them and I wanted to make our library recognizable to users who know the starter pack library.

The starter pack pngs, however, come from the bedroom designs page on the wiki. Pngs are useful for someone browsing the blueprints directory, as you said, but it might also be acceptable to just add a link to the wiki page in the blueprint itself.

Right now, I just want to get /something/ in the library so it isn't so empty. I would have preferred to get input from users on which blueprints to include, but there aren't enough engaged users yet. This week probably be all I as to the library for now until I get better direction.

lethosor:
Yeah, I think comments/notes in the blueprints would be a good place to refer to the images, and if done just there, links to the wiki could be a better choice than including the images here (since users might be reading the notes without knowing where the source files and images are).

myk002:
after tracking down all the images online, I realized that we probably don't need images at all. I reformatted the blueprints themselves so they are in a grid, making them their own image. I included the commit with the image links so we can go back to those if we want to, though.

lethosor:
I think the links could be helpful at least for the blueprints with multiple types of designations - telling `d` and ` ` apart is somewhat easier (for me, anyway) than `d` and `i`, particularly if you don't know which letters you're looking for or how many. So I'll probably add them back, if that's okay with you.

myk002:
I can do it - the commit with the comments actually has a side effect that I'd like to avoid. Since the #dig modeline is no longer on the first line, the "modeless blueprints are #dig blueprints" takes effect and you end up with an extra, empty #dig blueprint. I'll move the comments down one line and adjust the start cursor position.

lethosor:
Ah, sounds good, I probably would not have caught that!

myk002:
done

lethosor (approved):
Nice!

PR #1687, opened 2020-10-25
[buildingplan] persist global settings across map reloads

myk002:
this ensures that a player's preferences are saved across map reloads. this is particularly important for the building material filters, since item fulfillment tasks are regenerated on map load, and any changes in settings when buildingplan is reset will change the item vectors that will be searched for planned buildings. if the settings were allowed to reset, then a player who thought all walls would be made out of blocks would be surprised to find boulders and logs being used after the map is reloaded.

the quickfort_mode setting is intentionally not persisted, as it seems like that would be a "per session" kinda setting. I'd be happy to change that if people think differently, though. It might also depend on what other global settings we end up adding in the future, and whether it makes sense to persist those.

Also fix a small bug where buildingplan was attempting to scan persisted buildings when the plugin was enabled before a map was loaded

### DIFF
--- a/plugins/buildingplan-planner.cpp
+++ b/plugins/buildingplan-planner.cpp
@@ -20,6 +20,7 @@
 
 static const std::string planned_building_persistence_key_v1 = "buildingplan/constraints";
 static const std::string planned_building_persistence_key_v2 = "buildingplan/constraints2";
+static const std::string global_settings_persistence_key = "buildingplan/global";
 
 /*
  * ItemFilter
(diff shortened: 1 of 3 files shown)


myk002:
btw, should I be using labels for PRs instead of prefixing the PR title with the related plugin/script name? or are labels intended just for issues? On that note, could you possibly create a 'quickfort' label?

lethosor (approved):
Works! Omitting the Python quickfort mode setting from this makes sense to me.

PR #1689, opened 2020-10-29
[automaterial] [buildingplan] solve confusing UX interactions

### LINKED ISSUE #1688: [buildingplan][automaterial] Fix confusing UX interaction on construction placement screens
Filing this as a formal issue so it blocks -r4. I think the current state qualifies as "broken". I didn't catch this earlier since I was testing with 1x1 buildings, where the issue doesn't manifest.

Currently, if planning mode is enabled, buildingplan takes precedence and places a single tile of the construction on {Enter}, regardless of desired size or what settings are set for automaterial (e.g. box select). I think the solution here is:
- for automaterial-managed constructions, let automat handle placement (that is, buildingplan doesn't intercept the {Enter} key
- if planning mode is enabled, automaterial will
  - not show the "Auto Mat-select", "Reselect Type", or "Open Placement" options
  - construct the buildings with filters instead of showing its material selection screen
  - route the building back through buildingplan to let it select materials according to its filter

if buildingplan is enabled and automat is not, we stil have to handle the case where the requested size is > 1x1

myk002:
Fixes #1688

solves the confusing behavior when both automaterial and buildingplan are enabled for constructions. the two plugins now communicate with each other over the Lua layer to negotiate consistent behavior.

if neither plugin is enabled, the standard DF UI acts as normal

if automaterial is enabled but buildingplan is not, then automaterial behavior is unchanged.

if buildingplan is enabled and automaterial is not then behavior is the same as other buildings with buildingplan (no material selection screen, screen stays on building placement screen after placement). this commit fixes a bug, though, where buildingplan would only lay down a single tile of contruction instead of a solid block when a block is requested.

if both plugins are enabled but buildingplan is not enabled for the building type then automaterial is unchanged from previous behavior, except for an additional header showing the separation between automaterial hotkeys and buildingplan hotkeys.

finally, if both plugins are enabled and buildingplan is enabled for the building type then buildingplan behavior prevails, but the box select and hollow designations features of automaterial are still usable and useful. the 'Auto Mat-select', 'Reselect Type', and "Open Placement" automaterial hotkeys are hidden in the UI and ignored in the feed. This is because buildingplan takes over material selection, so 'Auto Mat-select' doesn't make sense. Buildingplan also already stays on the placement screen after placement, so 'Reselect Type' is not necessary. And all buildingplan-placed buildings have relaxed placement restrictions (e.g. they can be built in mid-air) so 'Open Placement' is also not necessary. The missing options are replaced with blank lines so the vertical alignment of all other options stays constant.

we also remove a few extra lua_pop() calls that are made superfluous by the StackUnwinder.


lethosor (approved):
These seem to work really well together now! Sorry for losing track of the PRs (real life caught up to me, as it does), and thanks for the nudge.

lethosor:
A side note: when I was testing this, I wasn't able to build constructions out of metal bars (even when enabling everything in buildingplan's global settings and allowing all materials in the material filter). I'm pretty sure that's unrelated to this PR, and it's entirely possible that I've overlooked something.

myk002:
> A side note: when I was testing this, I wasn't able to build constructions out of metal bars (even when enabling everything in buildingplan's global settings and allowing all materials in the material filter). I'm pretty sure that's unrelated to this PR, and it's entirely possible that I've overlooked something.

Bars are checked last, so if everything was enabled in buildingplan's settings, it will select blocks if possible, then boulders, then wood, then finally bars.

I just double-checked this by enabling just bars, and I was successfully able to build a track stop out of coke. Could you try this one more time? If there's something different about your setup, I'd like to make sure I'm handling it.

lethosor:
Yeah, I was able to use coke bars to build a floor. Possibly worth mentioning that I'm testing this with buildingplan + box-select to create large sections.

Interestingly, floors do not seem to get built out of bars if bars were globally disabled, then enabled after the floors were planned. The inverse is not true, though - floors planned when bars are globally enabled are still built even if bars are disabled immediately after planning the floors. Not sure if this is intentional.

(I should probably have made this a standalone issue, now that I think about it)

myk002:
Settings but getting retroactively applied is intentional. Documented (kinda) here: https://github.com/DFHack/dfhack/blob/7958c1f9e3369d055ec8d143607e66e887472204/plugins/lua/buildingplan.lua#L241

I have it on my [backlog](https://docs.google.com/document/d/17EjZlGsUOJ6E8WtqkCRZi9BBKjL9L1IdHfkB3Ajqas8/edit?usp=drivesdk) to add a method for retroactive application, but I wanted to see if the current implementation made anyone unhappy first.

myk002:
I tracked the problem with using metal bars down to our old friend:
`bool ItemTypeInfo::matches(const df::job_item &item, MaterialInfo *mat, bool skip_vector)`
I'm suspicious of this line:
```
    case BAR:
        OK(3,any_raw_material);
```
is this a flag that we should be setting in buildings.lua everywhere we specify generic building material?

myk002:
Ok, tracked it down. Metal bars do not get the non_economic flag in Materials.cpp:516 in
`void MaterialInfo::getMatchBits(df::job_item_flags2 &ok, df::job_item_flags2 &mask)`

`TEST(non_economic, !inorganic || !(ui && vector_get(ui->economic_stone, index)));`

Should iron bars get "non_economic" set? From the context, I think maybe they should. I need to look into why coal *does* work. Perhaps it gets marked as organic?

Even after fixing this, though, the buildingplan material matching code *also* fails to match metal bars as building materials. I'll open a new bug for these issues.

lethosor:
> Settings not getting retroactively applied is intentional. Documented (kinda) here:
> 
> https://github.com/DFHack/dfhack/blob/7958c1f9e3369d055ec8d143607e66e887472204/plugins/lua/buildingplan.lua#L241

Of course. I saw that message, noticed that it matched the length of the line below it really nicely, and forgot what it said.

I'm a bit confused why only _enabling_ bars appears to be retroactive, though, but I might not have been reproducing it consistently.

PR #1691, opened 2020-10-31
[buildingplan] don't show quality properties for no-quality items

myk002:
for items that cannot have a quality or be decorated:

in place mode, don't show quality adjustment hotkeys (or isDecorated flag hotkey) and don't interpret the associated keycodes if they are pressed.

in query mode, don't show quality or decorated fields.

this prevents players from erroneously requiring that items have a quality when they never will.


lethosor (approved):
Good catch!

PR #1692, opened 2020-10-31
[buildingplan] ensure we don't run cycles while the game is paused

myk002:
The code previously only checked World::ReadPauseState(), but this flag is not set to true if the game was paused implicitly by going to a submenu or switching modes.

we now also check if we're in a mode other than the default map or squads or if we're in a subscreen.


lethosor (approved):
Nice catch on `Squads`. Maybe this could go in the `Gui` module eventually.

PR #1700, opened 2020-11-13
identify bars as non_economic items

### LINKED ISSUE #1699: Buildingplan can't match metal bars as construction materials
See initial investigation in #1689

The issue is that if you try to build something that requires a building_material item, buildingplan never chooses a metal bar, even when it should. For example, if you try to build a track stop in buildingplan and set the material filter to 'metal', it will never choose an iron bar, even if that is the only building material available.

There are two separate bugs that need to be fixed. The first is in [Materials.cpp:516](https://github.com/DFHack/dfhack/blob/develop/library/modules/Materials.cpp#L516) in
`void MaterialInfo::getMatchBits(df::job_item_flags2 &ok, df::job_item_flags2 &mask)`:

`TEST(non_economic, !inorganic || !(ui && vector_get(ui->economic_stone, index)));`

It looks like iron bars should get "non_economic" set.

The second bug is that buildingplan's ItemFilter itself fails to match an iron bar as a valid metal material. I haven't tracked down the root cause for that one yet.

myk002:
Fixes #1699

allows MaterialInfo to match metal bars as a building material. This allows metal bars to be used for construction in buildingplan.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -50,6 +50,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## API
 - `buildingplan`: added Lua interface API
+- `job-module`: add item type param to dfhack.job.isSuitableMaterial() so the non_economic flag can be properly handled (it was being matched for all item types instead of just boulders)
 
 # 0.47.04-r3
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -50,7 +50,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## API
 - `buildingplan`: added Lua interface API
-- `job-module`: add item type param to dfhack.job.isSuitableMaterial() so the non_economic flag can be properly handled (it was being matched for all item types instead of just boulders)
+- add item type param to dfhack.job.isSuitableMaterial() so the non_economic flag can be properly handled (it was being matched for all item types instead of just boulders)
 
 # 0.47.04-r3
 
--- a/library/include/modules/Job.h
+++ b/library/include/modules/Job.h
@@ -102,7 +102,9 @@ namespace DFHack
                                          int filter_idx = -1, int insert_idx = -1);
 
         DFHACK_EXPORT bool isSuitableItem(df::job_item *item, df::item_type itype, int isubtype);
-        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type, int mat_index, df::item_type itype = df::item_type::NONE);
+        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type,
+                                              int mat_index,
+                                              df::item_type itype);
         DFHACK_EXPORT std::string getName(df::job *job);
     }
 
--- a/library/include/modules/Job.h
+++ b/library/include/modules/Job.h
@@ -102,7 +102,7 @@ namespace DFHack
                                          int filter_idx = -1, int insert_idx = -1);
 
         DFHACK_EXPORT bool isSuitableItem(df::job_item *item, df::item_type itype, int isubtype);
-        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type, int mat_index);
+        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type, int mat_index, df::item_type itype = df::item_type::NONE);
         DFHACK_EXPORT std::string getName(df::job *job);
     }
 
--- a/library/include/modules/Materials.h
+++ b/library/include/modules/Materials.h
@@ -78,7 +78,6 @@ namespace DFHack
 
         int16_t type;
         int32_t index;
-        df::item_type itype;
 
         df::material *material;
 
@@ -99,7 +98,7 @@ namespace DFHack
         df::historical_figure *figure;
 
     public:
-        MaterialInfo(int16_t type = -1, int32_t index = -1, df::item_type itype = df::item_type::NONE) { decode(type, index, itype); }
+        MaterialInfo(int16_t type = -1, int32_t index = -1) { decode(type, index); }
         MaterialInfo(const t_matpair &mp) { decode(mp.mat_type, mp.mat_index); }
         template<class T> MaterialInfo(T *ptr) { decode(ptr); }
 
@@ -114,7 +113,7 @@ namespace DFHack
         bool isAnyInorganic() const { return type == 0; }
         bool isInorganicWildcard() const { return isAnyInorganic() && isBuiltin(); }
 
-        bool decode(int16_t type, int32_t index = -1, df::item_type itype = df::item_type::NONE);
+        bool decode(int16_t type, int32_t index = -1);
         bool decode(df::item *item);
         bool decode(const df::material_vec_ref &vr, int idx);
         bool decode(const t_matpair &mp) { return decode(mp.mat_type, mp.mat_index); }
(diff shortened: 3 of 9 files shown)


myk002:
if this approach is accepted, I'll update docs in this PR as well

library/include/modules/Job.h
```diff
@@ -102,7 +102,7 @@ namespace DFHack
                                          int filter_idx = -1, int insert_idx = -1);
 
         DFHACK_EXPORT bool isSuitableItem(df::job_item *item, df::item_type itype, int isubtype);
-        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type, int mat_index);
+        DFHACK_EXPORT bool isSuitableMaterial(df::job_item *item, int mat_type, int mat_index, df::item_type itype = df::item_type::NONE);
```
lethosor, line 105: `job_item` objects have an `item_type` field - would that work, or is it different / not set in the case you care about?

library/include/modules/Materials.h
```diff
@@ -78,6 +78,7 @@ namespace DFHack
 
         int16_t type;
         int32_t index;
+        df::item_type itype;
```
lethosor, line 81: I'm a bit reluctant to introduce item-specific state in a material-specific class (I think this is basically what you brought up in https://github.com/DFHack/dfhack/issues/1699#issuecomment-726472912). The only place I see this state used is in Job::isSuitableMaterial(), specifically when the MaterialInfo is passed to `iinfo.matches(*item, &minfo);`

Would it be possible to make `ItemTypeInfo::matches()` take the item type instead of storing it in `MaterialInfo`? Possibly even better, `matches()` already takes a `job_item`, which has a `item_type` field per above - if this field is already set to the correct value, that could work (but I'm not sure if it is).

lethosor (commented):
I left some comments on what I think might be a possible alternative.

----

Another, potentially less-useful possibility I thought of when replying to #1699, before taking a closer look at this PR:

Would it make sense to change the non_economic flag around here? Or is that too early/late?
https://github.com/DFHack/dfhack/blob/fce10e598e61634824fd1a851e8dbc666c648931/library/modules/Items.cpp#L348-L352

This comment seems relevant, although I'm not sure where exactly the flag is being checked: https://github.com/DFHack/dfhack/blob/fce10e598e61634824fd1a851e8dbc666c648931/library/modules/Items.cpp#L327-L329

From what I've gathered, `xmask` indicates flags that should _not_ be checked in certain cases, but I'm not sure if that works here.

library/include/modules/Materials.h
lethosor: I did some investigating on my end, and it looks like ItemTypeInfo::matches is getting a job_item with `item_type = NONE` after creating a metal bar with `createitem BAR IRON`, enabling bars with `b-C-f-G-r`, and planning a new floor. Tables do seem to result in `item_type = TABLE` being set. I probably misunderstood the purpose of job_item - I'm guessing its fields make up more of a template and less of a concrete item.

library/modules/Job.cpp
```diff
@@ -608,15 +608,16 @@ bool Job::isSuitableItem(df::job_item *item, df::item_type itype, int isubtype)
...
+    df::job_item *item, int mat_type, int mat_index, df::item_type itype)
 {
     CHECK_NULL_POINTER(item);
 
     if (mat_type == -1 && mat_index == -1)
         return true;
 
     ItemTypeInfo iinfo(item);
-    MaterialInfo minfo(mat_type, mat_index);
+    MaterialInfo minfo(mat_type, mat_index, itype);
```
lethosor, line 620: Since `job_item.item_type` doesn't seem to be as useful as I had hoped, I think an alternative "clean" solution might be to pass `itype` to `matches()` below instead of storing it in `MaterialInfo`. (I think this is the only place where the stored `itype` is actually relied on, but I could be missing something - most of the other additions appear to just be passing it along.)

myk002:
I think passing the item type to matches() is a good idea. Will rejigger.

For reference, the job_item is the filter returned from buildings.lua. table has the item type set because it is set in the filter. Any construction material can be any of several types, so the filter doesn't set a type, just flags describing the required properties.

myk002:
> https://github.com/DFHack/dfhack/blob/fce10e598e61634824fd1a851e8dbc666c648931/library/modules/Items.cpp#L327-L329
> 
> From what I've gathered, `xmask` indicates flags that should _not_ be checked in certain cases, but I'm not sure if that works here.

I'm seeing a lot of similarities between Items.cpp and Materials.cpp. They both check the same flags, but Items.cpp seems to be better at it. In particular, 
https://github.com/DFHack/dfhack/blob/fce10e598e61634824fd1a851e8dbc666c648931/library/modules/Materials.cpp#L447-L463 does pretty much the same check, but without an xmask. The solution here appears to be to route itype to *that* matches() function and introduce an xmask for job_item_flags2. An even better solution might be to unify related logic in Items.cpp and Materials.cpp, but that's a larger issue.

myk002:
the only script usage of getSuitableMaterial() that I found is in gui/workshop-job. I'll update that script in a separate PR.

myk002:
I added the change to isSuitableMaterial() to the API section, though it might also belong in the 'fixes' section..

lethosor (approved):
Looks like it works on my end now!

Thanks for clarifying the job_item usage. As for the changelog, the API section is fine since that's the only user-facing change since r3 (constructions weren't plannable before).

library/modules/Job.cpp
```diff
@@ -605,10 +605,11 @@ bool Job::isSuitableItem(df::job_item *item, df::item_type itype, int isubtype)
     ItemTypeInfo iinfo(itype, isubtype);
     MaterialInfo minfo(item);
 
-    return iinfo.isValid() && iinfo.matches(*item, &minfo);
+    return iinfo.isValid() && iinfo.matches(*item, &minfo, true, itype);
```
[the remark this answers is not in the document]
myk002: sorry, yes, intentional, but not for this PR. I'll revert it here and discuss why I think this is a good idea in a future PR.

docs/changelog.txt
```diff
@@ -50,7 +50,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## API
 - `buildingplan`: added Lua interface API
-- `job-module`: add item type param to dfhack.job.isSuitableMaterial() so the non_economic flag can be properly handled (it was being matched for all item types instead of just boulders)
+- add item type param to dfhack.job.isSuitableMaterial() so the non_economic flag can be properly handled (it was being matched for all item types instead of just boulders)
```
lethosor, line 53: We could add an anchor like `.. _job-module:` to `Lua API.rst` (although it would only document what's exposed to Lua). The reason I don't usually do this is that changelogs are essentially reused verbatim as Markdown for GitHub and Reddit, causing anything in backticks to show up as code. It's fine for tools like `buildingplan` but could be weird or confusing for `job-module` to show up as-is. (I manually corrected the quickfort user guide link for r3, but a longer-term solution would be to actually generate proper markdown links somehow.)

An alternative could be to start with ``dfhack.job.isSuitableMaterial()`` (in double backticks, so it's not an RST link). Doesn't need to be part of this PR - I plan to work on changelog-related documentation before r4, so I'll do it then.

library/modules/Job.cpp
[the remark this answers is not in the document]
lethosor: Ok, sounds good, and floors with metal bars still work with that reverted.

docs/changelog.txt
myk002: ok, I'll use the double backtick approach in the future in situations like this. Thanks!

Issue #1702, opened 2020-11-13
Autounsuspend can no longer identify planned buildings


myk002:
Need to add a buildingplan lua api for querying whether a building is a planned building and call it from the Ruby autounsuspend script. The old method of detecting an item type of NONE no longer works.

Question: how does one call a lua function from ruby? Or do I need to export the isPlannedBuilding() function specifically to Ruby?

lethosor:
Is this an issue for all buildings or just specific types?

There's not an established way for plugins to expose new functions to Ruby. This would be easiest if there were another way to identify planned bindings without communicating directly with buildingplan, but it might be possible to add support for calling Lua functions from Ruby (I've done this from Python before, but Ruby has some additional constraints on function names that could make this a bit tricky)

lethosor:
As a side note, the unsuspend script doesn't appear to handle planned buildings at all, so it should also be changed.

myk002:
It's an issue for any building type that has an item type defined in its filter, e.g. https://github.com/DFHack/dfhack/blob/develop/library/lua/dfhack/buildings.lua#L35

Constructions, notably, will not get resumed by autounsuspend's logic since their job_items lack item_types, but all building types that autounsuspend used to skip (i.e. furniture) will get prematurely resumed and generate cancel spam.

On that note, what is the 'resume' command that I see referenced in autounsuspend and unsuspend? is it an old script that has since been removed?

So the options seem to be:
1. Create infrastructure to allow C++ plugin functions from Ruby
1. Add a marker somewhere in job_item or building state
1. Rewrite autounsuspend and unsuspend in Lua

(1) seems like it could have long-lasting value, but is hard
(2) might be possible, but seems fragile. DF might clean garbage state (now or in the future) when a save is reloaded, and we can never be sure that the marker we set won't ever have a game-visible effect.
(3) I'm not usually a fan of rewriting things from scratch (yeah, I know, quickfort), but these scripts are just a few lines long. and autounsuspend can be wholly replaced with calls to 'repeat unsuspend'. At the moment, this is the solution I'm leaning towards.

lethosor:
`resume` is https://github.com/DFHack/dfhack/blob/develop/plugins/resume.cpp. It appears to use the same logic, and I'm not quite sure how best to update it - if it can integrate with libbuildingplan to detect planned buildings, that might work? Alternatively, I suppose a similar Lua-based approach to automaterial could work.
[...]

PR #1703, opened 2020-11-13
[buildingplan] add buildingplan.isPlannedBuilding()

myk002:
#1702
so it can be checked from Lua in DFHack/scripts#223


lethosor (approved):
Looks good, thanks for handling docs!

PR #1704, opened 2020-11-13
Fixed bug missing world tiles lacking incurson supplied resources

PatrikLundell:


PatrikLundell:
Fixed the lint error (and added a changelog entry). My feedback on the lint error is that it would be easier to find the extra space if the "Check whitespace" report provided the file name and not only the line number (I didn't change many files, so it wasn't an in my case)...

lethosor:
Yeah, I noticed that. We added support for making `lint.py` and some other CI scripts print messages in a special format that GitHub Actions recognizes, so its warnings will show up in comments under the "Files changed" tab or on individual commits, like https://github.com/DFHack/dfhack/pull/1704/commits/6014a452353b61374408c6fc1bf1a085a3e7981b:

image

The problem appears to be that, after we implemented this, GitHub Actions changed the way it prints messages and no longer makes the filename/line/column visible (only the message we provide is logged, which happens to also include the line number). I'll take a pass over our scripts and make them print messages in both formats at the same time, which should hopefully make the log output clearer as well.

plugins/embark-assistant/matcher.cpp
```diff
@@ -1515,16 +1518,20 @@ namespace embark_assist {
...
                     }
                     break;
 
                 case embark_assist::defs::aquifer_ranges::None_Plus_At_Least_Light:
-                    if (!(static_cast<int8_t>(tile->aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::None)) ||
-                        !(static_cast<int8_t>(tile->aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::Light_Heavy))) {
+                    if ((!(static_cast<int8_t>(tile->aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::None)) &&
+                        (!(static_cast<int8_t>(tile->neighboring_aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::None)))) ||
+                        (!(static_cast<int8_t>(tile->aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::Light_Heavy)) &&
+                            (!(static_cast<int8_t>(tile->neighboring_aquifer) & static_cast<int8_t>(embark_assist::defs::aquifer_sizes::Light_Heavy))))) {
```
lethosor, line 1534: I might suggest wrapping some of these checks in a function, if there's a way to do that that makes sense (it could help reduce the `)))))` at the end, and possibly abstract away some of the negations). Up to you.

I'm also wondering if `aquifer_sizes` would make more sense as a bitfield instead of an enum, but there's no need to change it here (just pointing out that checking for specific bits in a bitfield could avoid the need for casting to `int8_t`).

lethosor (commented):
I'm guessing this is related to http://www.bay12forums.com/smf/index.php?topic=169634.255 - just linking for reference.

I'll merge once I've tested with a couple embarks containing sand/clay, unless you think there's anything else I should look at.

docs/changelog.txt
```diff
@@ -44,6 +44,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `search`: fixed an issue causing item counts on the trade screen to display inconsistently when searching
 - `stockpiles`: fixed a crash when loading food stockpiles
 - `stockpiles`: fixed an error when saving furniture stockpiles
+- `embark-assistant`: Fixed issue causing incursion resource matching to skip the checks if those resources were provided only through incursions.
```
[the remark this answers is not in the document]
PatrikLundell: "The checks" are the complete set of embark matching ones. The problem here was that the high level preliminary checks excluded world tiles that shouldn't have been excluded. Checks (or matching) are performed in two steps; first a world tile match to skip world tile that can't have any matches (which is where the logic failed), and then a detailed check of each possible embark within the remaining world tiles.

It's odd that my work is based on an older version as I thought I did get the latest before I started (earlier the same day), and getting those things cause a whole lot of stuff to be changed, so I have no idea why it wouldn't be based on the latest data.

plugins/embark-assistant/matcher.cpp
PatrikLundell: The reason for the use of an enum is primarily that this is what makes sense in the UI, so that is what was used throughout. However, it would probably have been better to translate it into a bit field once it gets out of the UI.
As usual, it started with only a few cases where it mattered, and then I had to add a whole bunch more...

PatrikLundell: As you can see, I reworked the aquifer bit handling from an enum to named bit constants, but didn't go all the way to a bit field. At the same time, I realized I could merge the native tile aquifer info with the neighboring tiles one so there would only be a single check against the criterion at hand. I think it was only one section of the code where the checks
became slightly longer, where several became considerably shorter.

docs/changelog.txt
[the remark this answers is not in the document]
lethosor: Got it, thanks.

Hm, maybe you pulled from the `master` branch? At least in this repo, that points to the latest stable release, and `develop` has changes made since then.

For reference, if it helps:
* 0.47.04-r3 history: https://github.com/DFHack/dfhack/commits/0.47.04-r3
* This PR's history: https://github.com/DFHack/dfhack/commits/0ef56ba859c2328957228385112da5f42b44d074 (basically, your commits + 0.47.04-r3)
* `develop` history: https://github.com/DFHack/dfhack/commits/develop

It'll merge successfully now, in any case - I'll just fix the changelog manually. Basing work on `develop` would help avoid merge conflicts, but in this case, it looks like no conflicting changes to embark-assistant have been made since r3. (I realize other DFHack repos like df-structures use `master` for everything - perhaps we should make them more consistent...)

plugins/embark-assistant/matcher.cpp
lethosor: Yeah, I figured it was UI-related. I like what you came up with here!

PatrikLundell: Thanks.

I assume it is as you assumed, i.e. that I may have pulled the master rather than develop. It sure would lessen confusion if the same convention was used throughout, though.

And thanks for sorting things out.

lethosor: For reference, I think https://docs.dfhack.org/en/stable/docs/Contributing.html#pull-request-guidelines is up-to-date (and would change if the default branch names change)

PatrikLundell: If you read the documentation every time, and aren't cocky enough to think you remember from the last time, a number of months earlier, yes...

lethosor (approved):

PR #1705, opened 2020-11-13
detect buildingplan buildings in the resume plugin

myk002:
#1702
so they won't get erroneously resumed


lethosor (approved):
Looks good! Simpler than I expected.

Issue #1417, opened 2018-12-19
Game crashes with dwarfvet and clean patient job


way-hu:
My game (windows x64, df 44.12, DFHack, revised mod) crashes about a second after a vet diagnosed a patient dog. I think that the crash is connected to a "Clean Patient" job being issued. My fort does not have empty buckets, and the crash disappears when I build a few buckets before trying to diagnose.

Might be connected to issue #1270 but I am not sure.

lethosor:
Probably related to #1270. I don't think we fixed anything directly related to dwarfvet in 0.44.12-r2, but can you check with that version, and if it still crashes, upload a save?

way-hu:
I've attached a save file. There's a wounded dog in the animal hospital area. Typing dwarfvet enable on the dfhack console would make the dog sleep at the topleft corner and the diagnostician goes there. As soon as the diagnostician leaves the dog the game crashes. 
[DVCrash.zip](https://github.com/DFHack/dfhack/files/2761174/DVCrash.zip)

way-hu:
Witnessig more crashes I came to think that it is pretty much related to not having the proper items available in the fort. No bucket for cleaning, no thread for suturing, etc.

Issue #1711, opened 2020-11-16
dwarfvet: Occasional crash upon discharging patient from hospital zone


Nilsolm:
I noticed that dwarfvet is a bit crashy currently so I've been trying to track down the possible causes. I got one of them which sometimes leads to a crash when an animal is discharged from the hospital after having been treated:

```
(gdb) bt
#0  0x00007fffd37f6d4a in std::_Bit_reference::operator=(bool) (__x=false, this=<optimized out>)
    at /usr/include/c++/9/bits/stl_bvector.h:237
#1  AnimalHospital::dischargePatient(Patient*, DFHack::color_ostream&)
    (this=this@entry=0x7fffce2f7da0, patient=<optimized out>, out=...) at ../plugins/dwarfvet.cpp:350
#2  0x00007fffd37f7636 in AnimalHospital::processPatients(DFHack::color_ostream&) (this=0x7fffce2f7da0, out=...)
    at ../plugins/dwarfvet.cpp:388
#3  0x00007fffd37f84bb in tickHandler(DFHack::color_ostream&, void*) (out=..., data=data@entry=0x76c)
    at ../plugins/dwarfvet.cpp:752
#4  0x00007ffff7bcd702 in manageTickEvent(DFHack::color_ostream&) (out=...) at ../library/modules/EventManager.cpp:351
#5  0x00007ffff7bcac11 in DFHack::EventManager::manageEvents(DFHack::color_ostream&) (out=...)
    at ../library/modules/EventManager.cpp:336
#6  0x00007ffff79370a8 in DFHack::Core::onUpdate(DFHack::color_ostream&) (this=this@entry=
    0x7ffff7f5aae0 <DFHack::Core::getInstance()::instance>, out=...) at ../library/Core.cpp:2118
#7  0x00007ffff794646d in DFHack::Core::doUpdate(DFHack::color_ostream&, bool)
    (this=0x7ffff7f5aae0 <DFHack::Core::getInstance()::instance>, out=..., first_update=<optimized out>)
    at ../library/Core.cpp:2069
#8  0x00007ffff79497ac in DFHack::Core::Update() (this=0x7ffff7f5aae0 <DFHack::Core::getInstance()::instance>)
    at ../library/Core.cpp:2100
#9  0x00007ffff6e0c202 in enablerst::async_loop() () at /home/david/Games/dev/df_linux/libs/libgraphics.so
[...]

Issue #1270, opened 2018-05-13
Dwarfvet causes game crash


bessgeor:
### Software in use
- PyLNP win x64 0.13
- DF 0.44.09
- DFHACK 0.44.09-r1 (release) on x86-64
- TWBT 6.41

### Problem
Running `dwarfvet enable` causes the game to crash after some ticks.

### Steps to reproduce
Download [my save](https://files.fm/f/x9rjfnq9), run it, run `dwarfvet enable`, wait a couple of seconds.

[Original issue in the plugin repo](https://github.com/NCommander/dfhack-dwarfvet/issues/2)

lethosor:
From #1272: https://drive.google.com/file/d/1_sHgznx1Bqz_LHqJvsP-w95W3uou5Biq/view?usp=sharing

lethosor:
@bessgeor I haven't gotten your save to crash in 0.44.10 (osx64) yet, using DFHack 0.44.10-beta1. Does it happen for you in that version?

Also, I'm getting this logged to the DFHack console around once a second:
```
Telling intelligent unit 7511 to report to the hospital!
Telling intelligent unit 7512 to report to the hospital!
Telling intelligent unit 8008 to report to the hospital!
Telling intelligent unit 8010 to report to the hospital!
```
as well as these announcements in-game:
```
Stray hunting Dog (Tame) cancels Rest: Area inaccessible.
Stray war Dog (Tame) cancels Rest: Area inaccessible.
```
Did you see any of those before the crash?

bessgeor:
@lethosor I don't know how to run 0.44.10 until new PyLNP release.

I have been getting the following logs on .09 just before the crash:
```
[DfHack]# dwarfvet enable
Populating hospital cache:
    Found animal hospital 11186  at x1: 137, y1: 161, z: 52 from valid hospital list
Accepted patient 7511 at hospital 11186
Accepted patient 7512 at hospital 11186
Accepted patient 8008 at hospital 11186
Accepted patient 8010 at hospital 11186
Telling intelligent unit 7511 to report to the hospital!
Telling intelligent unit 7512 to report to the hospital!
Telling intelligent unit 8008 to report to the hospital!
Telling intelligent unit 8010 to report to the hospital!
```

[Here](https://i.imgur.com/rFq34hk.jpg) is a screenshot of those, and [here](https://videobin.org/+skk/xd5.html) is a video of the crash (may take some time to be processed by the hosting. Logs and crash at 0:22).

[...]

PR #1712, opened 2020-11-17
dwarfvet - initialize spot_index (#1711), fix types in Patient

quietust:
Make sure `Patient::spot_index` actually gets initialized, as I can only assume was the original intent.

Note that I don't have the means to test this fix right now, and given how long this functionality has been broken, it's possible that there are other bugs in the plugin that rely on it.

### DIFF
--- a/plugins/dwarfvet.cpp
+++ b/plugins/dwarfvet.cpp
@@ -72,21 +72,22 @@ struct hospital_spot {
 class Patient {
   public:
     // Constructor/Deconstrctor
-    Patient(int32_t id, int spot_index, int32_t x, int32_t y, int32_t z);
+    Patient(int32_t id, size_t spot_index, int32_t x, int32_t y, int32_t z);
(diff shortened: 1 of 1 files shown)


Nilsolm:
I tested this for a bit. 

I found a way to reproduce the original issue by manually setting the needs_healthcare and rq_diagnosis flags on a dog to send it back to the hospital repeatedly. The crash typically happened after a few tries. No crash after applying this PR after 15-20 tries though.

This _may_ have also fixed #1417. I've seen a few clean patient jobs performed with no crashes, but that would need a bit more testing.

lethosor (approved):
Glad to hear it works! Looks good from a static review on my end. (spot_index is only used in one place, as a vector index, and is only assigned once, as a valid index into that same vector).

I was unable to reproduce #1270 recently with the provided save, but I will try to take a look at #1417.

PR #1713, opened 2020-11-17
Corrected world level incursion search criteria

PatrikLundell:
In addition to correcting the top level filtering, a few other things were made:
- Split local screen lair sites into mound, burrow, and general lair (wilderness location).
- Introduced enums for orientations rather than magic numbers, and placed them in their own name spaces to allow them to use the "same" name for the same direction (with different underlying numbers).
- Merged the top level incursion info into with the rest of the data for the top level as there's actually no need for any separation: that level is used only for preliminary filtering to weed out world tiles that can't have any matches.
- Realized that it's actually possible to use the collected data to perform a "proper" incursion influence on the top level, rather than a this might be present" one.


plugins/embark-assistant/matcher.cpp
```diff
@@ -1173,7 +1173,7 @@ namespace embark_assist {
                else if (k < start_y + finder->y_dim - 1) { //  We've already covered the SE tile's SE corner, with its complications.
                    process_embark_incursion_mid_level_tile
                    (embark_assist::survey::translate_corner(survey_results,
-                       8,
+                       embark_assist::defs::directions::Southwest,
```
lethosor, line 1176: 8 is Southeast, not Southwest - was this changed intentionally?

plugins/embark-assistant/embark-assistant.cpp
```diff
@@ -314,7 +314,19 @@ command_result embark_assistant(color_ostream &out, std::vector <std::string> &
             embark_assist::main::state->survey_results[i][k].survey_completed = false;
             embark_assist::main::state->survey_results[i][k].neighboring_clay = false;
             embark_assist::main::state->survey_results[i][k].neighboring_sand = false;
-            embark_assist::main::state->survey_results[i][k].neighboring_aquifer = embark_assist::defs::Clear_Aquifer_Bits;
+            for (uint8_t l = 0; l <= ENUM_LAST_ITEM(biome_type); l++) {
```
lethosor, line 317: There's also a FOR_ENUM_ITEMS macro in DataDefs.h, in case you were unaware - you're perfectly welcome not to use it, though (for stylistic reasons or otherwise - it does yield values of the enum type instead of the base integer type, although I think either would work here).

docs/changelog.txt
```diff
@@ -50,6 +50,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `buildingplan`: new global settings for whether generic building materials should match blocks, boulders, logs, and/or bars. defaults are everything but bars.
 - `quickfort`: The Dreamfort sample blueprints now have complete walkthroughs for each fort level and importable orders that automate basic fort stock management
 - `quickfort`: More blueprints have been added to the blueprints library: several bedroom layouts, the Saracen Crypts, and the complete fortress example from Python Quickfort: TheQuickFortress
+- `embark-assistant`: split the lair types displayed on the local map into mound, burrow, and lair
```
lethosor, line 53: Are the other changes in this PR essentially a continuation of #1704?

plugins/embark-assistant/matcher.cpp
PatrikLundell: No, that's a bug (the danger of improving stuff that isn't broken). I'll fix it.

plugins/embark-assistant/embark-assistant.cpp
PatrikLundell: Thanks. I didn't know that.

docs/changelog.txt
PatrikLundell: Yes, it's a continuation of that PR, so the line in the changelog covers it. The magic number->named number change is too internal and marginal to get a description, as it doesn't modify the functionality (apart from introduced bugs...).

lethosor: Makes sense, thanks for confirming

plugins/embark-assistant/matcher.cpp
lethosor: All of the other changes looked good, by the way (forgot to mention that)

lethosor (approved):

Issue #1544, opened 2020-04-10
buildingplan: Auto-allocation toggles while navigating with numpad


Nilsolm:
It seems that navigating with the Numpad while having a designated furniture selected toggles auto-allocation if the assigned key is pressed. Pressing '1' in this case allocates the room to the relevant noble position instead of moving the cursor. 

This is admittedly a rather minor thing, but it can lead to confusion if players accidentally auto-allocate rooms to certain nobles. It happened to me until I've found out what caused it.

**How to reproduce:** 

1. Select a relevant furniture in a designated room (bed, table, chair).
2. Use the Numpad to navigate while pressing one of the assigned keys.

**DFHack version:**

- 0.47.04-beta1

**Platform & OS:** 

- Linux64, Ubuntu 20.04

I haven't tried other platforms. It's possible that this is specific to Linux.

**Possible solutions:**

- Use different keys? a, f, r, l and x are already bound on that screen.
- Don't toggle allocation if Numpad keys are used? Would that be even possible?

lethosor:
This might not be that hard to fix, actually. The feed() method gets a list of all interface_keys that are used, so you could ignore anything if a CURSOR_* key is present. However, those might also trigger with the default keybindings if the regular number keys are used... seems like a problem we've dealt with before, but I don't remember how off the top of my head.

Nilsolm:
Alright, I'll look into it sometime next week and see if I can figure it out.

lethosor:
[`tweak shift-8-scroll`](https://github.com/DFHack/dfhack/blob/develop/plugins/tweak/tweaks/shift-8-scroll.h) is one I thought of, but that's doing something else (resolving a conflict between a SECONDSCROLL and CURSOR key by picking the former).

I don't have a numpad, so testing this is a bit tricky. One thing you could try: using [this script](https://github.com/lethosor/dfhack-scripts/blob/master/gui/input-monitor.lua), check what keys are produced when you press regular 1 vs numpad 1 (with as close to the default keybindings as possible). Hopefully at least something is different. If not, we may have to rework the interface to not conflict with numpad scrolling (e.g. by hiding these options in a submenu).

Nilsolm:
It seems that both "1" and "Numpad 1" are bound to the same set of commands by default: A_CONV_PERSUADE, CURSOR_DOWNLEFT, A_MOVE_SW, STRING_A049, so no easy solution there. It seems putting the option in a submenu that inhibits navigation is the better option.

[...]

PR #1714, opened 2020-11-18
buildingplan: Fix crash upon pressing 0 when a noble room is queried

Nilsolm:
Small fix to an issue identified in #1544


plugins/buildingplan.cpp
```diff
@@ -864,7 +864,7 @@ struct buildingplan_room_hook : public df::viewscreen_dwarfmodest
             && last_token <= interface_key::STRING_A058)
```
lethosor, line 864: Good catch! I think the key check here is also too lenient (it should accept keys from '1' to '9', but accepts them from '0' to ':' inclusive), so I'll change that.

lethosor (approved):

PR #1716, opened 2020-11-19
add libexpat, libzip, and xlsxio to LICENSE.rst

myk002:
missed this when I was adding the libraries


lethosor (approved):
Thanks for catching this!

PR #1717, opened 2020-11-21
Lua run_command improvements and documentation

lethosor:
A couple changes:

* Made `run_command()` send output directly to the console, instead of intercepting and printing it from the Lua side. This is faster and fixes some issues where commands run through `run_command()` would be unable to detect the console encoding (I noticed this when `warn-starving` printed some malformed text to the console on Linux when run under `repeat`).
* Added some documentation that was missing (as pointed out by Fleeting Frames)

I did just notice that `run_command()` can now run interactive commands, like `tiletypes` - I'm not sure if this is desirable, and it could probably be prevented by using a `color_ostream_proxy` instead.


lethosor:
Confirmed that running `tiletypes` with `repeat` leads to issues with console output after it runs the second time (i.e. the first time not directly as a result of running a console command). This should probably be addressed before merging.

lethosor (commented):
Tested on Windows, works there too

PR #1723, opened 2020-12-03
Corrected river size handling

PatrikLundell:
The world tile level river size determination doesn't provide the complete picture as smaller rivers may join larger ones and rivers change size along their courses. Thus, river size determination has to be performed on the MLT level to be correct.


lethosor (approved):
Looks good - thanks for updating the changelog!

PR #1724, opened 2020-12-06
[quickfort] overhaul quickfort alias documentation

myk002:
editorial review would be appreciated, though I'm sure the community will eventually find any errors too : p

- move the alias syntax and usage docs from dfhack-config/quickfort/aliases.txt to a proper guide written in RST. Add examples and more details.
- move the alias library docs from data/quickfort/aliases-common.txt to the new guide
- reorder aliases in aliases-common to match the order in the docs
- factor out the character used to enter the stockpile config screen so we can use the same aliases for stockpiles and hauling routes (use 's' for stockpiles and '{Enter}' for hauling routes)
- reference the new guide in the quickfort user guide
- do an editorial pass of the quickfort user guide
  - change name to "Quickfort Blueprint Guide", but only in the text, not the filename, so we don't change the URL
  - add `quickfort-blueprint-guide` as a label, in addition to the existing `quickfort-user-guide`
  - changed table-like lists to actual tables
  - changed "grid" tables into "simple" tables where possible
  - used ':kbd:' markers whenever we refer to a single character
  - turned Meta blueprints and Notes blueprints sections into subsections of a new "Other blueprint modes" section, in preparation for a few new modes coming in -r5.
  - removed out-of date caveat about bookcases, display furniture, and offering places not being supported


lethosor:
Going to try reopening this to see if that kicks off a ReadTheDocs build

lethosor:
ReadTheDocs seems to be successfully building PRs now - builds are listed at https://readthedocs.org/projects/dfhack/builds/, and you can find this particular one at https://dfhack--1724.org.readthedocs.build/en/1724/
I haven't figured out why it's not reporting build status back to GitHub yet, though - I suspect it needs to be authorized at the organization level, but RTD isn't making it particularly clear what needs to change.

myk002:
That's beautiful! details link takes you directly to the main index. I like : )

I found a bug in the script code while testing the changes in this PR, so DFHack/scripts has been updated. This PR is now ready for review.

docs/guides/quickfort-alias-guide.rst
```diff
@@ -0,0 +1,850 @@
...
+With aliases, blueprints are much easier to read and understand. They also
+save you from having to copy the same long key sequences everywhere.
+
+Alias definition files
+----------------------
+
+DFHack comes with a library of aliases for you to use that are always
+available when you run a ``#query`` blueprint. Many blueprints can be built
+with just those aliases. This "standard alias library" is stored in
+:source:`data/quickfort/aliases-common.txt`. The aliases in that file are
```
lethosor, line 62: Minor: may be worth clarifying that this is `hack/data/...` at runtime (since DF has its own top-level `data` folder). If it's intended to be read-only, this is probably not as important, but you could add something like "in the `hack` folder" if you want to avoid duplicating the file path.

docs/guides/quickfort-alias-guide.rst
```diff
@@ -0,0 +1,850 @@
...
+add your custom aliases to :source:`dfhack-config/quickfort/aliases.txt`.
+Definitions in this file take precedence over any definitions in the standard
+library.
+
+Alias syntax and usage
+----------------------
+
+The syntax for defining aliases is:
+
+::
```
lethosor, line 76: There is a shorthand you can use in cases like these (not mandatory, of course) - ending the line above with a double colon will render a single colon _and_ begin a code block. https://raw.githubusercontent.com/DFHack/dfhack/develop/docs/Compile.rst has some examples of this.

There's also a more explicit form, `.. code-block::`, that you can use to set syntax highlighting on a per-block basis, but the blocks in this document all seem to be displaying reasonably with the default "DFHack" highlighting (which basically just styles lines starting with `#` differently).

docs/guides/quickfort-alias-guide.rst
```diff
@@ -0,0 +1,850 @@
...
+alias nine times and the ``Down`` keycode five times, respectively.
+
+Modifier keys
+~~~~~~~~~~~~~
+
+Ctrl, Alt, and Shift modifiers can be specified for the next key by adding
+them into the key sequence. For example, Alt-h is written as ``{Alt}h``.
+
+Note that DF does not handle keys that have more than a single modifier, so
+sequences like ``{Ctrl}{Alt}a`` will result in an error.
```
lethosor, line 172: This isn't quite true - the `interface_key`s starting with `CUSTOM_` are only defined for up to one modifier, but DF supports binding any interface_key to combinations with multiple modifiers (e.g. `ctrl-alt-e` gets saved as `[SYM:6:e]`, where `6 == 4 | 2 == CTRL | ALT`).

Perhaps "DF" should be replaced with "quickfort" here? In any case, the default interface.txt (well, mine, which has minimal modifications) only uses `1`, `2`, and `4` modifiers for `SYM` entries, so this is a relatively unimportant feature to support in my opinion.

docs/guides/quickfort-alias-guide.rst
```diff
@@ -0,0 +1,850 @@
...
+Sub-aliases
+~~~~~~~~~~~
+
+You can specify sub-aliases that will only be defined while the current alias
+is being resolved. This is useful for "injecting" custom behavior into the
+middle of a larger alias. For example:
+
+::
+
+    {quantumstopfromeast name="Trash Dump"}
```
lethosor, line 215: It wasn't initially clear to me from reading this section that the `quantumstopfromeast` alias would need to include `{name}` in its definition. Could help to provide an example (minimal) definition of this macro.

Also, what happens if the `name` parameter is omitted? In the `quantum_enable` example below, there is an existing alias of that name that serves to provide default behavior, but it's less clear to me what would happen in this case (I assume it would expand to an empty string?).

And finally, including underscores in the names of aliases here might make them a bit easier to read, but that's not hugely important (especially if you're trying to match aliases defined elsewhere).

lethosor (commented):
I am behind on these, apparently. I figured the comments I have so far might be useful (I haven't gotten past the "sub-aliases" section yet) - I'm hoping to get through the whole thing by tomorrow. Definitely an improvement! 

Since this is documentation for an existing feature, I would like to get it into r4 if possible (also delayed due to real life and such) but it is also possible to roll out docs updates later to the "stable" branch without making a new DFHack release (they'd just be left out of the offline copy that users get).

docs/guides/quickfort-alias-guide.rst
myk002: done

docs/guides/quickfort-alias-guide.rst
myk002: thank you for this tip! several hundred superfluous lines are now deleted.

for the ``.. code-block::`` format, how do you set the language for the syntax highlighting? I believe the default is python, which isn't always appropriate

docs/guides/quickfort-alias-guide.rst
myk002: I removed the note. Thanks! FTR, quickfort code does support the general case of multiple modifiers. I just noticed that interface.txt didn't have any multiple-modifier keys when I tried to test the functionality. I forgot that custom definitions could be added.

docs/guides/quickfort-alias-guide.rst
myk002: good feedback. text updated. I have it as a future task for myself to allow empty alias definitions so the "optional sub-alias" case is handled more elegantly.

underscores (and hyphens) are relatively new to alias names. all the "original" aliases are pure alphanumeric (since they used the same convention as Python Quickfort). I could rename the existing ones, but I would risk breaking blueprints that use them. I don't *think* there are many blueprints that would break at this point (there hasn't been very much time yet for users to create blueprints that use the alias library), but I can't be sure.

docs/guides/quickfort-alias-guide.rst
lethosor: `.. code-block:: language`
I made a couple changes in 09f262a0c234ad34da176ffb6bd9cd125384572a and a5f85e256d5f443bbc979969b37bc268122fb6e3
I think the default looks good here, but you can also change the default for all following code blocks with the `.. highlight:: language` directive (Compile.rst sets this to `shell` initially)

myk002:
updates pushed

lethosor (approved):
The rest of the new stuff looks good!

PR #1728, opened 2020-12-16
corrected index usage

PatrikLundell:
Don't think this small bug fix warrants another line in the change log, as there is already a bug fix one.


lethosor (approved):
Looks good - [forum discussion](http://www.bay12forums.com/smf/index.php?topic=169634.msg8223681#msg8223681) for reference
(and agreed on the changelog)

PR #1729, opened 2020-12-16
only reset extents in Buildings::setSize() if they are unusable

myk002:
This allows callers of `Buildings::setSize()` to "pre-initialize" the extents to declare non-rectangular structures. The main beneficiary of this change is quickfort, which currently has a bug where non-rectangular structures *appear* correct on the map, but the occupancy is set incorrectly for non-occupied tiles within the bounds of the extent.

From inspection, the existing code appears to clear existing extents to protect against the case where the passed-in size does not match the size eventually assigned to the building. We can check for that case, though so the extents can still be reset as before if the size of the building doesn't match the caller's expectations.

This commit also fixes a memory leak when `setSize()` allocates memory for extents, but the memory is not deallocated if the building is ultimately invalid for some reason.

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -1679,7 +1679,11 @@ Low-level building creation functions:
   Returns *false* if the building cannot be placed, or *true, width,
   height, rect_area, true_area*. Returned width and height are the
   final values used by the building; true_area is less than rect_area
-  if any tiles were removed from designation.
+  if any tiles were removed from designation. You can specify a non-rectangular
+  designation for building types that support extents by setting the
+  ``room.extents`` bitmap before calling this function. The extents will be
+  reset, however, if the size returned by this function doesn't match the
+  input size parameter.
 
 * ``dfhack.buildings.constructAbstract(building)``
 
@@ -1768,7 +1772,9 @@ Among them are:
 
   - ``fields = { ... }``
 
-    Initializes fields of the building object after creation with ``df.assign``.
+    Initializes fields of the building object after creation with
+    ``df.assign``. If ``room.extents`` is assigned this way and this function
+    returns with error, the memory allocated for the extents is freed.
 
   - ``width = ..., height = ..., direction = ...``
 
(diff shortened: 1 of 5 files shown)


lethosor:
Do you happen to have an example blueprint that's fixed by this? (Would a non-rectangular stockpile be enough?)

myk002:
sorry, yes:

#place
a,u,a
u,a,u
a,u,a
,
a,a,a
a,u,a
a,a,a
,
a,a,a
a, ,a
a,a,a
#zone
g,g,g
g,h,g
g,g,g
#build
p,p,p
p,p(1x1),p
p,p,p

but it requires some changes to the script as well (we need to assign extents in the "fields" param in order to make use of the new functionality in Buildings::setSize(). I'll put up another PR with those changes.

myk002:
> (Would a non-rectangular stockpile be enough?)
non-rectangular stockpiles are not enough by themselves. they look correct on the game map. the problem appears when you try to build a second stockpile in the (seemingly) unclaimed space

myk002:
sorry I'm so fragmented here. more info: the root cause of the problem is that when setSize clears the extents, it allows Buildings::markBuildingTiles to assume a solid rectangular footprint. Then tile occupancy is set with that footprint. when quickfort later overrides the extents, it doesn't update the occupancy.

myk002:
closing and reopening to get the buildmaster to try again

lethosor:
Thanks for the clarification - I hadn't seen where occupancy was handled before, and the relationship between what you changed and `markBuildingTiles` wasn't clear to me at first.

I've verified that this fix (combined with https://github.com/DFHack/scripts/pull/236) works with your sample blueprint. Going to try to familiarize myself with the buildings module to make sure that the fix makes sense here, but I'll try not to delay it too much (it does seem worth including in r4 to me, unless you think it's not important enough).

myk002:
I vote for inclusion in -r4. I think this change solves a real bug (in -r3 quickfort) and is low risk. `buildings.constructBuilding` is the only caller of `setSize` (that I can find), and nobody other than quickfort does anything with extents.

lethosor (approved):

lethosor:
setSize looks like a good place to handle this to me

PR #1731, opened 2020-12-17
high level docs for the quickfort blueprint library

myk002:
no pictures yet, but just something to describe what is available in the blueprint library so it's more accessible. pictures will take more work -- v2 of this guide.


lethosor (approved):
Looks good (and links all work)! Nice to have this described in one place.

PR #1734, opened 2020-12-27
Embark assistant

PatrikLundell:
The discussion regarding this bug starts here: [http://www.bay12forums.com/smf/index.php?topic=169634.msg8226685#msg8226685](http://www.bay12forums.com/smf/index.php?topic=169634.msg8226685#msg8226685).


plugins/embark-assistant/survey.cpp
```diff
@@ -432,17 +432,17 @@ namespace embark_assist {
             }
 
             if (world->world_data->world_height == 17) {
-                divisor = (57 / steps * lat + 0.4);
+                divisor = ((lat * 57) / steps + 0.4);
```
lethosor, line 435: Per your comment [here](http://www.bay12forums.com/smf/index.php?topic=169634.msg8227674#msg8227674), would it make more sense to use floating-point math here? I think changing 57 to 57.0 might be enough, but casting somewhere would also do the trick:
```suggestion
                divisor = (double(lat * 57) / steps + 0.4);
```
(It's also possible I misinterpreted your comment and this is something else)

PatrikLundell: I think you misinterpreted my comment. I believe Toady uses integer math, which is why the formulae uses integer math. The integer logic has been shown to match for every latitude checked (about half of them, I think, for all world sizes, with every latitude for the two smallest ones ["every" includes only samples for the "other" hemisphere for dual pole worlds, to verify that the assumption that the hemispheres are completely mirrored]). Changing it to floating point math should be done only after verifying that the result would fit every case of a sample of at least the same size, and then probably only if it results in the removal of the fudge factors. The error I made in this code was to place the factors in the wrong order (the parenthesis shouldn't have any effect, but makes the intent clearer).

lethosor: Oh, so you're replicating a DF algorithm that potentially has that issue - got it.

PR #1736, opened 2020-12-29
Add an error message when LoadLibrary fails

### LINKED ISSUE #1735: Print error message when LoadLibrary fails
Unlike the posix version, the windows version of `OpenPlugin` does not print an error message when `LoadLibrary` fails. This makes it harder to diagnose some user issues (e.g. http://www.bay12forums.com/smf/index.php?topic=138754.msg8227430#msg8227430).

cvuchener:
fix #1735

I wonder if I should use std::system_category instead of FormatMessage (not portable but DFHack is only compiled with MSVC), and rtrim instead of a custom algorithm (it requires the creation of a std::string but performance is not important here).

### DIFF
--- a/library/PlugLoad-windows.cpp
+++ b/library/PlugLoad-windows.cpp
@@ -24,6 +24,7 @@ distribution.
 
 #define DFhackCExport extern "C" __declspec(dllexport)
 
+#define NOMINMAX
 #include <windows.h>
 #include <stdint.h>
 #include <vector>
@@ -35,6 +36,7 @@ distribution.
 
 #include "tinythread.h"
 #include "modules/Graphic.h"
+#include "../plugins/uicommon.h"
 
 /*
  * Plugin loading functions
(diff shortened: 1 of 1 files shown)


lethosor:
If you can get information that's at least as descriptive out of std::system_category, I would be okay with that, although I'm not entirely sure how it would replace FormatMessageA here. I also thought FormatMessageA was Windows-specific, not MSVC-specific, but I could be wrong about that.

Up to you how you want to do string manipulations, but `rtrim` could be a bit more readable (and creating a string here for that is unlikely to cause issues, performance-related or otherwise).

cvuchener:
Yes, FormatMessage is Win32 API, not MSVC-specific. The issue is with std::system_category, the standard says "errors reported by the operating system", but GCC uses POSIX errno on all platforms (including Windows), while MSVC uses Windows error codes.

It would look like this:
```c++
DWORD error = GetLastError();
std::string message = std::system_category().message(error);
std::cerr << "[...]" << rtrim(message) << std::endl;
```
This code would not work if compiled with GCC.

lethosor:
Ah, I misunderstood which thing your parenthetical applied to. We only need to worry about MSVC on Windows, and there is a fair amount of other code that depends on MSVC as well (e.g. VTableInterpose). I guess I would rather go with the current Win32 API you're using, just to make sure it's interpreting the error codes correctly.

`rtrim()` could be an improvement, though (it's defined in `uicommon.h`, which would pull in a number of other headers, but hopefully that wouldn't cause other issues - if it does, maybe it should be moved to `MiscUtils.h`).

Anyway, I'll try to double-check this on Windows in a bit.

cvuchener:
When including `uicommon.h` (as `../plugins/uicommon.h`, is there a better way?), I get some errors because `windows.h`'s min and max. Where is the best place to define `NOMINMAX`? Before the include or more globally in `CMakeLists.txt` ?

lethosor:
Interesting... looks like `library/include/modules/Filesystem.h` and `depends/tthread/tinythread.h` already define NOMINMAX, so I guess setting it in CMakeLists.txt would work. `Export.h` is another place where some project-wide macro definitions live, if you prefer setting it there.

lethosor:
Looks good to me on Windows. I do also get this dialog, even without your changes, but maybe this is configuration-dependent:

Screenshot from 2021-01-06 20-26-37

lethosor (approved):

PR #1738, opened 2020-12-30
Fix crash in look menu search

### LINKED ISSUE #1737: search: crash when navigating map after searching `k` sidebar
Reported by Fleeting Flames (via Kitfox Discord)

To reproduce:
1. With `k`, select a tile with at least 3 items listed in the sidebar (floors/walls included)
2. Search and narrow down the list to under 3 items (e.g. `s`, `beer`, `Enter`)
3. Navigate away from this tile with a key that has an ASCII representation (`<`, `>`, or `1-9` - *not* an arrow key), or with F1-F8 hotkeys (edit: the crash with hotkeys seems to occur when they are used in step 3, followed by arrow keys in step 4)
4. Navigate to an unrevealed tile (or maybe a tile with fewer items than in step 2?) with the arrow keys (~or via an announcement,~ etc.)

Step 4 usually crashes because the search state isn't cleared in step 3 due to the check added in 87b1524ae. 

The logic in one of these places should probably change: https://github.com/DFHack/dfhack/blob/a7232499ac78ae85346cac43254eea4ee156dc90/plugins/search.cpp#L1956
https://github.com/DFHack/dfhack/blob/a7232499ac78ae85346cac43254eea4ee156dc90/plugins/search.cpp#L121 
We only need to intercept ASCII keys when the search field is active (step 2).

Note that changing the behavior of `cursor_key_pressed` may also affect the behavior of `search_generic`, so the proper fix here might be complicated:
https://github.com/DFHack/dfhack/blob/a7232499ac78ae85346cac43254eea4ee156dc90/plugins/search.cpp#L252

lethosor:
Fixes #1737

### DIFF
--- a/plugins/search.cpp
+++ b/plugins/search.cpp
@@ -115,13 +115,16 @@ static string get_unit_description(df::unit *unit)
     return desc;
 }
 
-static bool cursor_key_pressed (std::set<df::interface_key> *input)
+static bool cursor_key_pressed (std::set<df::interface_key> *input, bool in_entry_mode)
 {
-    // give text input (e.g. "2") priority over cursor keys
-    for (auto it = input->begin(); it != input->end(); ++it)
+    if (in_entry_mode)
     {
-        if (Screen::keyToChar(*it) != -1)
-            return false;
+        // give text input (e.g. "2") priority over cursor keys
+        for (auto it = input->begin(); it != input->end(); ++it)
+        {
+            if (Screen::keyToChar(*it) != -1)
+                return false;
+        }
     }
     return
     input->count(df::interface_key::CURSOR_UP) ||
@@ -249,7 +252,7 @@ public:
                 // ENTER or ESC: leave typing mode
                 end_entry_mode();
             }
-            else if (cursor_key_pressed(input))
+            else if (cursor_key_pressed(input, entry_mode))
             {
                 // Arrow key pressed. Leave entry mode and allow screen to process key
                 end_entry_mode();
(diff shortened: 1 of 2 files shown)


lethosor:
Fleeting Frames reports that this doesn't handle F1-F8 hotkeys (i.e. step 4 from #1737). Fixing those is likely doable, but fixing navigation from the announcements screen (likely broken in the same way) is harder, since from my understanding, search state is still saved. Unsure what other screens would lead to similar issues - likely zooming to units from the `u` screen, at least. Maybe detecting new screens and resetting search state is a usable approach here.

lethosor:
Fleeting Frames clarified that announcements zooming was purely speculative - if DF ever allows the game to run when in the `k` menu, announcements could cause the position of the screen to change, leading to similar crashes. That's not something we can address yet if it doesn't happen.

plugins/search.cpp
```diff
@@ -249,7 +252,7 @@ class search_generic
                 // ENTER or ESC: leave typing mode
                 end_entry_mode();
             }
-            else if (cursor_key_pressed(input))
+            else if (cursor_key_pressed(input, entry_mode))
```
lethosor, line 255: On second glance, I think this call isn't technically correct - most screens (other than fortress mode) use STANDARDSCROLL keys to scroll (or maybe even SECONDSCROLL), but cursor_key_pressed checks for CURSOR keys. Worth addressing, but out of scope here, and it works with default keybindings.

PR #1746, opened 2021-01-11
[Lua] getopt-compatible lua script args parser

myk002:
Happy new year! This PR introduces a more flexible alternative to the existing `processArgs()` function in utils.lua

In addition to a more powerful argument handling syntax, this implementation don’t force "flag" options to be at the end. For example, with processArgs, `quickfort list dreamfort -l` is legal but `quickfort list -l dreamfort` is not. This causes confusion when users try the second (IMHO, more natural) syntax.

The new implementation is based on https://github.com/LuaDist/alt-getopt (MIT license), modified to support non-flag options and to replace calls to `os.exit()` with `qerror()`.

The new `processArgs2()` function is a wrapper of the core getopt logic and provides a handler callback interface for processing parameters. Documentation for the new function is in utils.lua. Is there anywhere else that these functions should be documented?

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed order of factors when calculating min temperature
 
+## API
+- ``processArgs2()`` added to utils.lua, providing a callback interface for parameter parsing and getopt-like flexibility for parameter ordering and combination (see docs in ``library/lua/utils.lua`` and ``library/lua/3rdparty/alt_getopt.lua`` for details).
+
 # 0.47.04-r4
 
 ## Fixes
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,7 +36,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed order of factors when calculating min temperature
 
-## API
+## Lua
 - ``processArgs2()`` added to utils.lua, providing a callback interface for parameter parsing and getopt-like flexibility for parameter ordering and combination (see docs in ``library/lua/utils.lua`` and ``library/lua/3rdparty/alt_getopt.lua`` for details).
 
 # 0.47.04-r4
@@ -1171,4 +1171,3 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - The ``ui_menu_width`` global is now a 2-byte array; the second item is the former ``ui_area_map_width`` global, which is now removed
 - The former ``announcements`` global is now a field in ``d_init``
 - ``world`` fields formerly beginning with ``job_`` are now fields of ``world.jobs``, e.g. ``world.job_list`` is now ``world.jobs.list``
-
--- a/library/lua/3rdparty/alt_getopt.lua
+++ b/library/lua/3rdparty/alt_getopt.lua
@@ -1,0 +1,171 @@
+-- Copyright (c) 2009 Aleksey Cheusov <vle@gmx.net>
+--
+-- Permission is hereby granted, free of charge, to any person obtaining
+-- a copy of this software and associated documentation files (the
+-- "Software"), to deal in the Software without restriction, including
+-- without limitation the rights to use, copy, modify, merge, publish,
+-- distribute, sublicense, and/or sell copies of the Software, and to
+-- permit persons to whom the Software is furnished to do so, subject to
+-- the following conditions:
+--
+-- The above copyright notice and this permission notice shall be
+-- included in all copies or substantial portions of the Software.
+--
+-- THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+-- EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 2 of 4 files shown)


library/lua/utils.lua
```diff
@@ -613,6 +613,55 @@ function processArgs(args, validArgs)
...
+        end
+        sh_opts = sh_opts .. sh_opt
+        if optionAction.hasArg then sh_opts = sh_opts .. ':' end
+        handlers[sh_opt] = optionAction.handler
+        if long_opt then
+            long_opts[long_opt] = sh_opt
+            handlers[long_opt] = optionAction.handler
+        end
+    end
+    local getopt = require('3rdparty.alt_getopt').get_ordered_opts
```
lethosor, line 657: It's fine to use `require()` at the global level - in fact, maybe preferable, since CI would catch any load-time issues with it.

docs/changelog.txt
```diff
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed order of factors when calculating min temperature
 
+## API
```
lethosor, line 39: Since it's Lua-only:
```suggestion
## Lua
```

library/lua/3rdparty/alt_getopt.lua
```diff
@@ -0,0 +1,169 @@
+-- Copyright (c) 2009 Aleksey Cheusov <vle@gmx.net>
```
lethosor, line 1: I was about to suggest `internal/alt_getopt.lua` like the JSON library, but it looks like I named the original implementation `json/internal.lua` instead of `internal/json.lua` - guess we don't have a standard for this after all. `3rdparty` is good with me (maybe more accurate too), as long as starting module names with numbers doesn't cause issues (Python would complain, but I think Lua doesn't care).

The other third-party module I know of is `dumper.lua`, but that's different because it isn't wrapped at all, so putting it in the top-level `library/lua` folder makes sense.

library/lua/3rdparty/alt_getopt.lua
```diff
@@ -0,0 +1,169 @@
...
+--
+-- Permission is hereby granted, free of charge, to any person obtaining
```
lethosor, line 3: If you could also add an entry for this in `docs/LICENSE.rst`, that would be great.

library/lua/utils.lua
```diff
@@ -613,6 +613,55 @@ function processArgs(args, validArgs)
...
+-- local open_readonly = false
+-- local nonoptions = processArgs2(args, {
+--   {'r', handler=function() open_readonly = true end},
+--   {'f', 'filename', hasArg=true,
+--    handler=function(optarg) filename = optarg end}
+-- })
+--
+-- when args is {'first', '-f', 'fname', 'second'}
+-- then filename will be fname and nonoptions will contain {'first', 'second'}
+function processArgs2(args, optionActions)
```
lethosor, line 638: Perhaps `processArgsGetopt`, `getopt`, or some variation of `getopt` would be better? If we ever introduce another implementation (however unlikely), I would be hesitant to name it `processArgs3`.

lethosor:
I forgot to leave a top-level review comment - this looks useful! And I like how it's used in https://github.com/DFHack/scripts/pull/238 - some extra boilerplate, yes, but it looks like it'll allow for a lot more flexibility.

library/lua/utils.lua
myk002: done

library/lua/3rdparty/alt_getopt.lua
myk002: done. sorry for forgetting that, especially after I explicitly added the others for xlsxio...

library/lua/utils.lua
myk002: hehe true. I'll go with `processArgsGetopt` to highlight the fact that it's an alternate to `processArgs`

docs/changelog.txt
```diff
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed order of factors when calculating min temperature
 
+## Lua
+- ``processArgs2()`` added to utils.lua, providing a callback interface for parameter parsing and getopt-like flexibility for parameter ordering and combination (see docs in ``library/lua/utils.lua`` and ``library/lua/3rdparty/alt_getopt.lua`` for details).
```
lethosor, line 40: Looks like this needed to be changed too - changing on my end

library/lua/utils.lua
```diff
@@ -613,6 +614,60 @@ function processArgs(args, validArgs)
...
+-- {'first', '--filename', 'fname', 'second'} (note the double dash in front of
+-- the long option alias), then filename will be fname and nonoptions will
+-- contain {'first', 'second'}.
+function processArgsGetopt(args, optionActions)
+    local sh_opts, long_opts = '', {}
+    local handlers = {}
+    for _,optionAction in ipairs(optionActions) do
+        local sh_opt,long_opt = optionAction[1], optionAction[2]
+        if not sh_opt or type(sh_opt) ~= 'string' or #sh_opt ~= 1 then
+            qerror('optionAction missing option letter at index 1')
```
lethosor, line 650: IMO these should be changed to `error` because they're dev-facing errors - I'll adjust locally

lethosor (approved):
Thanks!

PR #1747, opened 2021-01-11
[buildingplan] set global settings from DFHack# prompt

myk002:
Add commandline parsing/validation and documentation for buildingplan global settings.

More info in the Plugins.rst documentation, but quick example usage:

`on-new-fortress buildingplan set boulders false; buildingplan set logs false`

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -723,11 +723,11 @@ While placing a building, you can set filters for what materials you want the bu
 out of, what quality you want the component items to be, and whether you want the items to
 be decorated.
 
-If a building type takes more than one item to construct, use
-:kbd:`Ctrl`:kbd:`Left` and :kbd:`Ctrl`:kbd:`Right` to select the item that you
-want to set filters for. Any filters that you set will be used for all buildings
-of the selected type from that point onward (until you set a new filter or clear
-the current one).
+If a building type takes more than one item to construct, use :kbd:`Ctrl`:kbd:`Left` and
+:kbd:`Ctrl`:kbd:`Right` to select the item that you want to set filters for. Any filters that
+you set will be used for all buildings of the selected type placed from that point onward
+(until you set a new filter or clear the current one). Buildings placed before the filters
+were changed will keep the filter values that were set when the building was placed.
 
 For example, you can be sure that all your constructed walls are the same color by setting
 a filter to accept only certain types of stone.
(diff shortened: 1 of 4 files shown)


docs/Plugins.rst
```diff
@@ -745,6 +745,42 @@ Note that Quickfort mode is only for compatibility with the legacy Python Quickf
 DFHack `quickfort` script does not need Quickfort mode to be enabled. The `quickfort` script
 will successfully integrate with buildingplan as long as the buildingplan plugin is enabled.
 
+Global settings
```
lethosor, line 748: This might be worth cross-referencing from the "Buildingplan integration" section of the quickfort user guide (I briefly assumed it was there already)

plugins/buildingplan.cpp
```diff
@@ -911,26 +911,75 @@ IMPLEMENT_VMETHOD_INTERPOSE(buildingplan_query_hook, render);
 IMPLEMENT_VMETHOD_INTERPOSE(buildingplan_place_hook, render);
 IMPLEMENT_VMETHOD_INTERPOSE(buildingplan_room_hook, render);
 
+DFHACK_PLUGIN_IS_ENABLED(is_enabled);
+
+static void setSetting(std::string name, bool value);
+
+static bool isTrue(std::string val)
+{
+    return val == "on" || val == "true" || val == "y" || val == "yes";
```
lethosor, line 920: Some commands accept `1` and `0` as well.

Since you're already processing the value passed in here with `toLower` on the caller's side, I might suggest moving that here, since I assume you don't ever want `YES` to be treated as false. (Unlikely, sure, but since I'm already suggesting things...)

plugins/buildingplan.cpp
```diff
@@ -911,26 +911,75 @@ IMPLEMENT_VMETHOD_INTERPOSE(buildingplan_query_hook, render);
...
+        show_debugging = isTrue(toLower(parameters[1]));
+        out << "buildingplan debugging: " <<
+                ((show_debugging) ? "enabled" : "disabled") << endl;
+    }
+    else if (cmd == "set")
+    {
+        if (!is_enabled)
         {
-            out << "Building Plan" << endl << "Version: " << PLUGIN_VERSION << endl;
+            out << "ERROR: buildingplan must be enabled before you can read or "
```
lethosor, line 944: Errors could use `out.printerr()` (which doesn't support `<<` syntax, but it looks like both messages are just raw strings)

plugins/buildingplan.cpp
```diff
@@ -911,26 +911,75 @@ IMPLEMENT_VMETHOD_INTERPOSE(buildingplan_query_hook, render);
...
+                << (quickfort_mode ? "true" : "false") << endl;
+        }
+        else if (parameters.size() == 3)
         {
-            show_debugging = (toLower(parameters[1]) == "on");
-            out << "Debugging " << ((show_debugging) ? "enabled" : "disabled") << endl;
+            // set a setting
+            std::string setting = toLower(parameters[1]);
+            bool val = isTrue(toLower(parameters[2]));
+            setSetting(setting, val);
```
lethosor, line 976: Does this do any sort of error detection? I felt like trying the following commands, and they all did nothing. Not hugely important if it's hard to add.
```
buildingplan set a
buildingplan set a b
buildingplan set a b c
```

I see that the second one does print a debug message, in `planner.setGlobalSetting()` but I suspect routing that to the correct output stream, and only when run directly from the console (not from a Lua frontend) would be tricky. Handling the first and third examples might be nice, for completeness (just an `else` below here). Maybe the second could be handled by checking for the key's presence in `planner.getGlobalSettings()`, now that I'm looking a bit more closely?

lethosor (approved):
Looks like it works well and integrates with the buildingplan Lua frontend like I'd expect!

docs/Plugins.rst
myk002: done. I actually thought it was already there too..

plugins/buildingplan.cpp
myk002: done and done.

plugins/buildingplan.cpp
myk002: Changed it all. And for the better. Have I ever mentioned how much I dislike c++ stream insertion for string output?

plugins/buildingplan.cpp
myk002: done. planner.setGlobalSetting() already returns false if the key is not found, so I can just use that return value.

docs/Plugins.rst
lethosor: I was actually thinking of a link, but I can add that. Thanks for updating it!

plugins/buildingplan.cpp
lethosor: They don't translate quite as directly as I'd like - specifically, `print()`/`printerr()` require newlines in the messages (I took care of adding the ones that were missing).

PR #1748, opened 2021-01-11
scroll Lua Label text when height exceeds viewrect

myk002:
Allows Label text to scroll with Up/Down/PgUp/PgDown keys, just like the ListBox classes.

Not being able to scroll Label text became an issue for quickfort because blueprint help text is displayed in a MessageBox (which uses a Label to display the text). Blueprint help text is user-defined, so long blueprint walkthroughs (I'm looking at you, dreamfort) could not be fully displayed.

I would have loved to accept Space as equivalent to PageDown, but that interacts poorly with text input fields (depending on the order components' onInput() functions are called, it could end up scrolling every time you typed a space into a text field)

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -436,7 +436,7 @@ function Label:onInput(keys)
                 v = -self.frame_body.height
             end
             self:scroll(v)
-            return true
+            return false
         end
     end
     return check_text_keys(self, keys)
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -234,11 +234,10 @@ end
 
 function render_text(obj,dc,x0,y0,pen,dpen,disabled)
     local width = 0
-    for iline,line in ipairs(obj.text_lines) do
-        if dc and obj.start_line_num and iline < obj.start_line_num then
-            goto continue
-        end
-        local x = 0
+    -- note that lines outside of the containing frame are not displayed, so
+    -- we only have to bound the start condition, not the end condition
+    for iline = dc and obj.start_line_num or 1, #obj.text_lines do
+        local x, line = 0, obj.text_lines[iline]
         if dc then
             local offset = (obj.start_line_num or 1) - 1
             dc:seek(x+x0,y0+iline-offset-1)
@@ -327,7 +326,6 @@ function render_text(obj,dc,x0,y0,pen,dpen,disabled)
             token.x2 = x
         end
         width = math.max(width, x)
-        ::continue::
     end
     obj.text_width = width
 end
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -235,9 +235,13 @@ end
 function render_text(obj,dc,x0,y0,pen,dpen,disabled)
     local width = 0
     for iline,line in ipairs(obj.text_lines) do
+        if dc and obj.start_line_num and iline < obj.start_line_num then
+            goto continue
+        end
         local x = 0
         if dc then
-            dc:seek(x+x0,y0+iline-1)
+            local offset = (obj.start_line_num or 1) - 1
+            dc:seek(x+x0,y0+iline-offset-1)
         end
         for _,token in ipairs(line) do
             token.line = iline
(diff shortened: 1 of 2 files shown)


library/lua/gui/widgets.lua
```diff
@@ -399,16 +413,33 @@ function Label:onRenderBody(dc)
...
+end
+
 function Label:onInput(keys)
-    if not is_disabled(self) then
-        if keys._MOUSE_L_DOWN and self:getMousePos() and self.on_click then
-            self:on_click()
-        end
-        if keys._MOUSE_R_DOWN and self:getMousePos() and self.on_rclick then
-            self:on_rclick()
+    if is_disabled(self) then return false end
```
lethosor, line 424: Good cleanup here!

library/lua/gui/widgets.lua
```diff
@@ -235,9 +235,13 @@ end
 function render_text(obj,dc,x0,y0,pen,dpen,disabled)
     local width = 0
     for iline,line in ipairs(obj.text_lines) do
```
lethosor, line 237: I think changing this to a numeric for loop might lead to slightly cleaner logic than the current approach (continuing until the desired start index is reached), e.g.:
```lua
for iline in (obj.start_line_num or 1), #obj.text_lines do
  local line = obj.text_lines[iline]
```
(I'm not quite sure what the `dc` check is for, but it might also need to be included in the first expression there)

library/lua/gui/widgets.lua
```diff
@@ -235,9 +235,13 @@ end
...
+        if dc and obj.start_line_num and iline < obj.start_line_num then
```
lethosor, line 238: Also, do you need a check for an "end line"?

lethosor (commented):
How does this behave on screens with other widgets that are also scrollable? I think the main page of `gui/gm-editor` would be an example.

myk002:
> How does this behave on screens with other widgets that are also scrollable? I think the main page of `gui/gm-editor` would be an example.

I had onInput() return false when a scroll key is detected, so nothing would eat the key. since we have no way to set focus on a particular widget, if there are multiple scrollable widgets on the same view, they will all scroll.

library/lua/gui/widgets.lua
myk002: overflow text just isn't displayed. I could have stopped processing it, but I didn't want to complicate the code if I didn't have to. I don't think there is a detectable performance difference. I can add a comment to make this decision explicit, though.

library/lua/gui/widgets.lua
myk002: done. the dc check is because render_text is called with nil dc from Label:getTextWidth() and List:getContentWidth(), and in those cases it is important to let it "render" all the text so obj.text_width can be set at the end of the function

library/lua/gui/widgets.lua
lethosor: Several Painter drawing calls check `isValidPos()`, so they should be at least somewhat faster when drawing text outside the visible region.

library/lua/gui/widgets.lua
lethosor: Ah, thanks for the explanation

library/lua/gui/widgets.lua
myk002: fair point. done.

lethosor: Oh, I was agreeing with you, not suggesting a change - should have made that clearer. But thanks for the update!

library/lua/gui/widgets.lua
```diff
@@ -399,16 +412,33 @@ function Label:onRenderBody(dc)
     render_text(self,dc,0,0,text_pen,self.text_dpen,is_disabled(self))
 end
 
+function Label:scroll(nlines)
+    local n = self.start_line_num + nlines
+    n = math.min(n, self:getTextHeight() - self.frame_body.height)
```
lethosor, line 417: Looks like this needs a `+1` at the end - I can address

lethosor (approved):

library/lua/gui/widgets.lua
myk002: Ah, I was letting it overlap a line so the reader wouldn't be unsure if they were missing anything

lethosor: To clarify, I was seeing a cut-off line at the end ("modifications!" can't be viewed):

image


which this fixed:
```suggestion
    n = math.min(n, self:getTextHeight() - self.frame_body.height + 1)
```

PR #1750, opened 2021-01-11
[quickfort] Dreamfort blueprint improvements

myk002:
- add checklist for all required dreamfort commands
- give names to stockpiles, levers, bridges, and zones
- ensure diggers don't dig deeper than they need to using dig priorities


myk002:
I think the doc build failure is a flake. docs build fine locally.

lethosor:
Yeah, this is a known intermittent issue with the older Linux environment (and upgrading Sphinx requires upgrading Python, and it's a bit tricky)
> Sphinx parallel build error:

lethosor (approved):

PR #1752, opened 2021-01-12
[autofarm] fallow farms when no further plants requested

### LINKED ISSUE #1741: Request: autofarm: let farms fallow if all plants reached desired count
Currently, as I understand it, `autofarm` 

1. searches for farms that are planting in excess: these are the farms that are currently set to plant plants to which (more than) the desired amount of farm space has already been allocated. Then, `autofarm`
2. checks all plants to see which ones still require more farm space to be allocated to them, and sets the farms that were planting in excess to plant those plants instead.

This leaves open the possibility of _all_ counts of desired plants having been reached, while the farms are still planting plants.
In that case, `autofarm` is currently happy to let the farms keep planting plants, with undesired plant yields as a result.

I propose that, when all plant counts set by the user have already been attained, we set the remaining farms that were planting in excess to fallow.

This could probably be achieved by changing the `set_farms` function to set all remaining farms marked as `toChange` to fallow upon reaching all desired planting allocations.

abstern:
Change the autofarm plugin to set any farm to fallow whenever no plant remains that both
- can be planted on it
- and has not yet reached its user-configured threshold.

This prevents serious excess plant yields when total farm production potential exceeds total plant consumption.

Resolves #1741.

### DIFF
--- a/plugins/autofarm.cpp
+++ b/plugins/autofarm.cpp
@@ -201,11 +201,26 @@ public:
     {
         // this algorithm attempts to change as few farms as possible, while ensuring that
         // the number of farms planting each eligible plant is "as equal as possible"
-
-        if (farms.empty() || plants.empty())
-            return; // do nothing if there are no farms or no plantable plants
-
+        
         int season = *df::global::cur_season;
+        
+        if (farms.empty() || plants.empty())
+        {
+            // if no more plants were requested, fallow all farms
+            // if there were no farms, do nothing
+            for (auto farm : farms)
+            {
+                int o = farm->plant_id[season];
+                if (o != -1)
+                {
+                    farm->plant_id[season] = -1;
+                    out << "autofarm: changing farm #" << farm->id <<
+                        " from " << ((o == -1) ? "NONE" : world->raws.plants.all[o]->name) <<
+                        " to NONE" << endl;
+                }
+            }
+            return;
+        }
 
         int min = farms.size() / plants.size(); // the number of farms that should plant each eligible plant, rounded down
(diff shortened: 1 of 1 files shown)


plugins/autofarm.cpp
```diff
@@ -201,11 +201,26 @@ class AutoFarm {
...
+            // if no more plants were requested, fallow all farms
+            // if there were no farms, do nothing
+            for (auto farm : farms)
+            {
+                int o = farm->plant_id[season];
+                if (o != -1)
+                {
+                    farm->plant_id[season] = -1;
+                    out << "autofarm: changing farm #" << farm->id <<
+                        " from " << ((o == -1) ? "NONE" : world->raws.plants.all[o]->name) <<
```
lethosor, line 218: A more "DFHack-y" way would be to use `df::plant_raw::find(o)` here, which would give you a `plant_raw*`, set to null if the ID is invalid (like -1). It wouldn't quite be a drop-in replacement (we don't have an idiomatic way to access a field of a possibly-null pointer that I know of), but it would look something like this:
```c++
df::plant_raw *raw = df::plant_raw::find(o);
// ...
out << (raw ? raw->name : "NONE")
```

abstern: Very well!

Since similar code is used elsewhere as well (to set the farms when new plants _are_ requested), I have opted to factor out the simple functions `get_plant_name` and `set_farm`.

lethosor (approved):
Looks good to me from some basic testing. @ab9rf any thoughts?

plugins/autofarm.cpp
lethosor: Cool, thanks!

PR #1753, opened 2021-01-16
don't turn off seedwatch on map load

### LINKED ISSUE #682: Seedwatch ignores map load event, prevents autoloading of plugin.
# Summary:

If you try to auto-load `seedwatch` it will always fail, because it ignores DF's map load event.
# Fix

I think the only code change required is:

```
case SC_MAP_LOADED:
case SC_MAP_UNLOADED:
```

to:

```
case SC_MAP_LOADED:
    running = true;
    out.print("seedwatch supervision started.\n");
    break;

case SC_MAP_UNLOADED:
```

Because [`seedwatch start` only toggles a bool](https://github.com/DFHack/dfhack/blob/master/plugins/seedwatch.cpp#L144).
# Details:

Placing into one of the automatically ran config files such as `onLoadWorld.init` or `PyLNP_dfhack_onLoad.init` (From the PyLNP pack) will not work. Which gives the following DFHack output error:

```
seedwatch deactivated due to game load/unload
```

 This is the config used:

```
# autobutcher runs fine; yet seedwatcher fails.
enable seedwatch
enable autobutcher
seedwatch start
autobutcher autowatch
autobutcher start

# Debug output shows seedwatcher fails
seedwatch info
autobutcher list
```

Even with this config every game save and load you must retype `seedwatch start`.

This is because [seedwatcher does not load on `DFHack::SC_MAP_LOADED`](https://github.com/DFHack/dfhack/blob/master/plugins/seedwatch.cpp#L286), compared to [autobutcher which does](https://github.com/DFHack/dfhack/blob/master/plugins/zone.cpp#L295). (Seedwatcher toggles itself off when the event fires).

Moth-Tolias:
should fix #682 =]

i've also removed the comment that states seedwatch doesn't work with linux. surprised it took over a decade to be removed!

[i don't really know how c++ works, so it's possible i got this wrong - please let me know if anything needs to be changed!]

### DIFF
--- a/plugins/seedwatch.cpp
+++ b/plugins/seedwatch.cpp
@@ -283,6 +283,9 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
 {
     switch (event) {
     case SC_MAP_LOADED:
+		if(running)
+			out.print("seedwatch supervision started.\n");
+		break;
     case SC_MAP_UNLOADED:
         if (running)
             out.printerr("seedwatch deactivated due to game load/unload\n");
--- a/plugins/seedwatch.cpp
+++ b/plugins/seedwatch.cpp
@@ -1,4 +1,3 @@
-// This does not work with Linux Dwarf Fortress
 // With thanks to peterix for DFHack and Quietust for information http://www.bay12forums.com/smf/index.php?topic=91166.msg2605147#msg2605147
 
 #include <map>
--- a/plugins/seedwatch.cpp
+++ b/plugins/seedwatch.cpp
@@ -142,8 +142,17 @@ command_result df_seedwatch(color_ostream &out, vector<string>& parameters)
         }
         else if(par == "start")
         {
-            running = true;
-            out.print("seedwatch supervision started.\n");
+            if(Core::getInstance().isWorldLoaded())
+            {
+                running = true;
+                out.print("seedwatch supervision started.\n");
+            } else {
+                out.printerr(
+                    "This plugin needs a fortress to be loaded and will deactivate automatically otherwise.\n"
+                    "Activate with 'seedwatch start' after you load the game.\n"
+                    );
+            }
+
         }
         else if(par == "stop")
         {
--- a/plugins/seedwatch.cpp
+++ b/plugins/seedwatch.cpp
@@ -100,7 +100,23 @@ string searchAbbreviations(string in)
 
 DFhackCExport command_result plugin_enable(color_ostream &out, bool enable)
 {
-    running = enable;
+    if(enable == true)
+    {
+        if(Core::getInstance().isWorldLoaded())
+        {
+            running = true;
+            out.print("seedwatch supervision started.\n");
+        } else {
+            out.printerr(
+                "This plugin needs a fortress to be loaded and will deactivate automatically otherwise.\n"
+                "Activate with 'seedwatch start' after you load the game.\n"
+                );
+        }
+    } else {
+        running = false;
+        out.print("seedwatch supervision stopped.\n");
+    }
+
     return CR_OK;
 }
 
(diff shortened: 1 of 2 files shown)


Moth-Tolias:
oh whoops, looks like my editor added tabs. what's the best way to resolve that? add another commit removing them?

myk002:
thank you for this : )  could you add a docs/changelog.txt entry recording the change?

Moth-Tolias:
should be good now !

plugins/seedwatch.cpp
```diff
@@ -284,6 +283,9 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
 {
     switch (event) {
     case SC_MAP_LOADED:
+        if(running)
+            out.print("seedwatch supervision started.\n");
+        break;
     case SC_MAP_UNLOADED:
         if (running)
             out.printerr("seedwatch deactivated due to game load/unload\n");
```
lethosor, line 291: A minor suggestion:
```suggestion
            out.printerr("seedwatch deactivated due to game unload\n");
```

lethosor (commented):
Thanks for working on this! I wasn't quite sure what the best solution here was originally - it wasn't entirely clear to me why the plugin only allows itself to be enabled when a world is loaded, but I suppose `workflow` and `autolabor` are some examples of plugins that behave similarly. It is likely that users would want this enabled on a per-world basis, not per-install, so allowing this to be run in an `onLoad.init` file makes sense (and that can be either global or per-world). 

However, I think a better way to handle this would be to prevent the plugin from being enabled in the first place if a world isn't loaded. Specifically, this would involve adding a check for `Core::getInstance().isWorldLoaded()` [here](https://github.com/DFHack/dfhack/blob/4a7546c03a40a73953b9e1348a85f3bccefcb9e3/plugins/seedwatch.cpp#L144-L148) and perhaps printing an error message with `out.printerr()` if no world is loaded instead of enabling the plugin.

Additionally, [plugin_enable](https://github.com/DFHack/dfhack/blob/4a7546c03a40a73953b9e1348a85f3bccefcb9e3/plugins/seedwatch.cpp#L102) appears to be another possible route to enable the plugin without a check for a world or a message - maybe the world-loaded check and message printing could be moved to `plugin_enable()`, and the "start" and "stop" commands could be refactored to call it?

There are also [docs](https://github.com/DFHack/dfhack/blob/4a7546c03a40a73953b9e1348a85f3bccefcb9e3/docs/Plugins.rst#id115) to be aware of, although I think they would still be accurate with these changes.

Interestingly enough, I'm pretty sure that putting `seedwatch start` in an `onLoadMap.init` file would have had the desired fix earlier - it wasn't working in `onLoad.init` because that runs on world load events, which happen before map load events, but state change events like these are passed to plugins before `.init` scripts are called. But that's a detail that users typically shouldn't need to worry about.

Moth-Tolias:
i'll try and work on this today. sorry for taking so long.

lethosor:
No worries - I've also been slow on some things. Thanks for getting back to this!

Moth-Tolias:
> Additionally, [plugin_enable](https://github.com/DFHack/dfhack/blob/4a7546c03a40a73953b9e1348a85f3bccefcb9e3/plugins/seedwatch.cpp#L102) appears to be another possible route to enable the plugin without a check for a world or a message - maybe the world-loaded check and message printing could be moved to `plugin_enable()`, and the "start" and "stop" commands could be refactored to call it?

how do i call plugin_enable() ? it looks like i need a color_ostream, but i don't know what that is.

lethosor:
It's a special type of output stream that we have. `df_seedwatch()` already takes one, named `out`, so you could just pass it through.

Moth-Tolias:
should be good now.

plugins/seedwatch.cpp
```diff
@@ -284,9 +298,12 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
 {
     switch (event) {
     case SC_MAP_LOADED:
+        if(running)
+            out.print("seedwatch supervision started.\n");
```
lethosor, line 302: This does result in a duplicate message when `seedwatch start` is placed in `onLoad.init`; other plugins don't tend to log their state when a world is loaded, so I'll probably remove this one.

lethosor (approved):
Looks good!

PR #1755, opened 2021-01-18
[embark-assistant] removing 2 dead stores to speed up survey::survey_mid_level_tile

bseiller:
as it says in the title

details:
- survey.cpp: removing maps layer_bottom and layer_top, which are never read, but slow down survey_mid_level_tile significantly because entries are added quite often into the tree map structure
- survey.h: removing now obsolete include for map

forum discussion for reference
http://www.bay12forums.com/smf/index.php?topic=169634.msg8128622#msg8128622
and more current
http://www.bay12forums.com/smf/index.php?topic=169634.msg8235898#msg8235898

@PatrikLundell: Just remembered that it helps to "@" the reviewer... :facepalm:

### DIFF
--- a/plugins/embark-assistant/survey.cpp
+++ b/plugins/embark-assistant/survey.cpp
@@ -996,7 +996,6 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data 
 
             base_z = elevation - 1;
             features = details->features[i][k];
-            std::map<int, int> layer_bottom, layer_top;
             mlt->at(i).at(k).adamantine_level = -1;
             mlt->at(i).at(k).magma_level = -1;
 
@@ -1027,8 +1026,6 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data 
                     feature->min_z != -30000) {
                     auto layer = world_data->underground_regions[feature->layer];
 
-                    layer_bottom[layer->layer_depth] = feature->min_z;
-                    layer_top[layer->layer_depth] = feature->max_z;
                     base_z = std::min((int)base_z, (int)feature->min_z);
 
                     if (layer->type == df::world_underground_region::MagmaSea) {
--- a/plugins/embark-assistant/survey.h
+++ b/plugins/embark-assistant/survey.h
@@ -1,5 +1,4 @@
 #pragma once
-#include <map>
 
 #include "DataDefs.h"
 #include "df/coord2d.h"


PatrikLundell:
Accepting this is a nobrainer.

lethosor (approved):

lethosor:
Indeed, although I like to wait for the builds to finish successfully (which they have), e.g. in case removing `#include`s breaks on a platform you didn't test on.

PatrikLundell: I believe you should also have the ability to formally review changes (under the "Files changed" tab -> "Review changes"), although there's not a significant difference for our purposes.

lethosor:
FYI, I try to include any notable user-facing changes in the changelog - https://github.com/DFHack/dfhack/commit/ed3c48e64b6cffdeefd91331dcf1e2ec48b69c5e#diff-59c023da8c40bbe8277d0f469e56591bf153585bf70b86c0f652240ce178044c for an example.

(I realize that this is not documented particularly well, and I still haven't addressed that, but I will hopefully do that soon.)

PatrikLundell:
Thanks for the pointer regarding reviewing. As you say, it doesn't make much difference in this case, but it certainly doesn't hurt to try to do things properly.

bseiller:
@PatrikLundell: Yes I see what you mean by high level and "reuse" - I'll keep it in mind.
@lethosor: Thanks for pointing out those things - One more questions: Will you take care of "rebase" the other two pull requests (see here https://stackoverflow.com/questions/51629464/how-to-handle-multiple-pull-requests and here https://stackoverflow.com/questions/16696528/how-to-submit-multiple-pull-requests-in-github-when-they-may-conflict-slightly - depending on the perspective)?
Also I'll take care to add an addition to the changelog.txt the next time if there is no entry yet that fits the bill.

The next PR won't contain any changes that will make any difference for the users though - it will be a real refactoring as preparation for the index integration - might take a while tough...

PatrikLundell:
I don't think there will be any problems with accepting the PRs without any rebasing, as the changes don't seem to overlap, and thus shouldn't conflict with each other. However, it's not wrong if you refresh your local copies with the accepted changes and push them back up to the PRs.

It's good to get an indication that there will be some time before the next PR is issued.

bseiller:
@PatrikLundell: Merged the other 2 PRs/branches with dfhack/develop and pushed them both into their respective branches. It was far easier than I expected - hopefully I did it right :)

PatrikLundell:
Yes, if you manage to use the right commands it's rather easy. Neither of them show any changes, which is good, as there shouldn't be anything new to commit. I looked at the diffs of one of them, and that showed the accepted PR's changes as well as some buildingplan and documentation stuff, which is about what I would expect. I didn't look at the diffs of the other one, just noted that there weren't any changes in it either.

PR #1756, opened 2021-01-18
[embark-assistant] early return from embark_update during an active search to improve performance

bseiller:
details: 
- embark-assistant.cpp: checking if a search is active, if so return early

Especially for large worlds the repeated calls to embark_update during the survey phase can burn a significant amount of cycles while the user often can't see the results before they are reset with the next ones.

@PatrikLundell: ... and here as well.

### DIFF
--- a/plugins/embark-assistant/embark-assistant.cpp
+++ b/plugins/embark-assistant/embark-assistant.cpp
@@ -57,6 +57,12 @@ namespace embark_assist {
         //===============================================================================
 
         void embark_update() {
+            // not updating the embark overlay during an active find/match/survey phase
+            // which leads to better performance
+            if (state != nullptr && state->match_iterator.active) {
+                return;
+            }
+
             auto screen = Gui::getViewscreenByType<df::viewscreen_choose_start_sitest>(0);
             embark_assist::defs::mid_level_tiles mlt;
             embark_assist::survey::initiate(&mlt);
--- a/plugins/embark-assistant/matcher.cpp
+++ b/plugins/embark-assistant/matcher.cpp
@@ -3065,6 +3065,12 @@ uint16_t embark_assist::matcher::find(embark_assist::defs::match_iterators *iter
         iterator->active = !(iterator->i > world->worldgen.worldgen_parms.dim_y / 16);
 
         if (!iterator->active) {
+            // if the cursor was positioned in the lower right corner before the search it has to be moved to a neighbouring tile manually 
+            // to force another call to embark_update when all (incursion) data is finally collected to make sure this specific world tile is properly reevaluated
+            // see the embark_update() in embark-assistant
+            if (iterator->x == world->worldgen.worldgen_parms.dim_x - 1 && iterator->y == world->worldgen.worldgen_parms.dim_y - 1) {
+                embark_assist::matcher::move_cursor(iterator->x - 1, iterator->y);
+            }
             embark_assist::matcher::move_cursor(iterator->x, iterator->y);
 
             if (!survey_results->at(0).at(0).survey_completed) {  // Every world tile has gone through preliminary survey, so add possible incursion resources to each tile.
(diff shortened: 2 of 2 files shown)


PatrikLundell:
This makes sense. However, I can see one edge case where this might result in an erroneous result, although it probably doesn't:

After surveying the cursor is moved back to where it was before the survey started, triggering a generation of the embark info, except in the case where the cursor already was at lower right corner world tile, in which case no movement is performed (as that's where the surveying ends). I think the embark info generated before the survey is correct as it doesn't rely on any incursion info, and nothing should replace it in the process. However, that edge case should probably be tested.

bseiller:
Isn't that true for all other starting positions as well, not just for the  lower right corner world tile?

Hm, just thinking aloud: It might be possible, if not very clean, to additionally check against the values iterator->x and iterator->y in the condition of the exit guard of embark_update and if those match the current position + `iterator->active == false` we don't exit early.
As a function this would be badly called 
`bool isSearchActiveButNotReturningToOriginPosition(state/iterator, screen.x, screen.y)`
I'll have to check that with some outputs and come back to you.

PatrikLundell:
Since it's a single time possible case, I think it's wasteful to check for it in embark_update, and it's better to check for it in matcher::find, so instead of
"if (!iterator->active) {
            embark_assist::matcher::move_cursor(iterator->x, iterator->y);"

we'd use
"if (!iterator->active) {
            if (iterator->x == world->worldgen.worldgen_parms.dim_x - 1 && iterator->y == world->worldgen.worldgen_parms.dim_y - 1) {
            embark_assist::matcher::move_cursor(iterator->x - 1, iterator->y);
}
            embark_assist::matcher::move_cursor(iterator->x, iterator->y);"

(I hope this horrible formatting is still possible to interpret).

bseiller:
I was literally just looking at those lines of code :rofl:

I'll add it + a comment that explains this very special solution.

PatrikLundell:
Looks good to go.

bseiller:
It seems that the ci build failed with an error similar to what can be found in at least one other recent PR with failed checks:
https://github.com/DFHack/dfhack/pull/1750
https://buildmaster.lubar.me/executions/execution-details?executionId=664852&noAutoRedirect=y

```
FAILED: cd /home/buildmaster/work/_E0/build32-4.8 && /usr/local/bin/sphinx-build -a -E -q -b html /home/buildmaster/work/_E0/src /home/buildmaster/work/_E0/src/docs/html -w /home/buildmaster/work/_E0/src/docs/_sphinx-warnings.txt -j 2
/usr/local/lib/python3.4/dist-packages/sphinx/util/nodes.py:94: FutureWarning:
   The iterable returned by Node.traverse()
   will become an iterator instead of a list in Docutils > 0.16.
  for classifier in reversed(node.parent.traverse(nodes.classifier)):
/usr/local/lib/python3.4/dist-packages/sphinx/util/nodes.py:94: FutureWarning:
   The iterable returned by Node.traverse()
   will become an iterator instead of a list in Docutils > 0.16.
  for classifier in reversed(node.parent.traverse(nodes.classifier)):
/usr/local/lib/python3.4/dist-packages/sphinx/util/nodes.py:94: FutureWarning:
   The iterable returned by Node.traverse()
   will become an iterator instead of a list in Docutils > 0.16.
  for classifier in reversed(node.parent.traverse(nodes.classifier)):

Sphinx parallel build error:
ninja exited with code 1 (failure)
```

PatrikLundell:
Sphinx deals with the documentation, and there are no changes to that in the PR, so I'd say it's something that's been brought in from the develop branch.

bseiller:
Ah, good to know!
I'm find it peculiar though that the two other PRs are fine, both of them also coming from the develop branch.
Something about this feels arbitrary.

PatrikLundell:
In another Github project there seems to be problems with some parts of the process being run in parallel, which apparently can cause some resources to be locked out, resulting in "arbitrary" failures.

lethosor:
Yep, same docs issue as https://github.com/DFHack/dfhack/pull/1750#issuecomment-767309873 - nothing you've introduced here, and looks like it worked the most recent time.

lethosor (approved):
Approving per @PatrikLundell's feedback

PR #1757, opened 2021-01-18
[embark-assistant] reduce memory footprint

bseiller:
By defining an additional struct "mid_level_tile_incursion_base" that only contains those attributes that are actually used during incursion processing - notably not the vectors for the inorganics (e.g. metals, economic stone and minerals) as those had the biggest impact on the memory footprint even when empty (48 bytes per vector on Windows) - it is possible to reduce memory consumption of the plugin significantly. 
While users likely won't notice this for small world, for large world (257*257) this amounts to savings of at least 580MiB/608MB.
(48 bytes * 3 vectors * 64 mlt_per_world_tile * world_width * world_height)
Using this new struct as parent for embark_assist::defs::mid_level_tile allows this adaptation with only minimal changes in some classes.

forum discussion for reference
http://www.bay12forums.com/smf/index.php?topic=169634.msg8128622#msg8128622
and more current
http://www.bay12forums.com/smf/index.php?topic=169634.msg8235898#msg8235898

@PatrikLundell: The name of the new struct is absolutely open for discussion. I originally called it "mid_level_tile_basic" but that name is not even half the truth.

### DIFF
--- a/plugins/embark-assistant/defs.h
+++ b/plugins/embark-assistant/defs.h
@@ -146,10 +146,10 @@ namespace embark_assist {
             std::vector<bool> minerals;
             std::vector<int16_t> neighbors;  //  entity_raw indices
             uint8_t necro_neighbors;
-            mid_level_tile north_row[16];
-            mid_level_tile south_row[16];
-            mid_level_tile west_column[16];
-            mid_level_tile east_column[16];
+            mid_level_tile_incursion_base north_row[16];
+            mid_level_tile_incursion_base south_row[16];
+            mid_level_tile_incursion_base west_column[16];
+            mid_level_tile_incursion_base east_column[16];
             uint8_t north_corner_selection[16]; //  0 - 3. For some reason DF stores everything needed for incursion
             uint8_t west_corner_selection[16];  //  detection in 17:th row/colum data in the region details except
                                                 //  this info, so we have to go to neighboring world tiles to fetch it.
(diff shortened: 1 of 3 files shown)


PatrikLundell:
I don't find "mid_level_tile_incursion_basic" to be a fantastic name either, but it's good enough, and I can't come up with a superior one, I guess it's a bit of a conflict between a name that describes its use and one that describes a base on which the full structure is built.

survey.cpp:
- The assignments to fields that aren't present in mid_level_tile_incursion_base should be removed completely, not just commented out.

Thus, with removal of that piece of commented out history I think it's good to go.
It feels good to get this process started!

PatrikLundell:
Looks like this PR is ready for acceptance.

plugins/embark-assistant/defs.h
```diff
@@ -68,24 +68,29 @@ namespace embark_assist {
             Woodland,
             Heavily_Forested
         };
-
-        struct mid_level_tile {
+        
+        // only contains those attributes that are being handled during incursion processing
+        struct mid_level_tile_incursion_base {
```
lethosor, line 73: Yeah, I wasn't able to come up with a better name either - if there were multiple derived structs, perhaps `_common` instead of `_base` could work (but there aren't)

lethosor (approved):

plugins/embark-assistant/defs.h
bseiller: Nonetheless: Thank you for putting some thought into it!

PR #1759, opened 2021-01-24
fix crash: avoid infinite parentage when linking rooms

### LINKED ISSUE #1758: Quickfort created zones not working and can cause DF to crash.
I have been doing a DreamFort run through seeing how the features work in a play mode.   I have found that zones created are not working.  The animal pen on the surface the dwarfs will not move the dogs to the pen on the stairs nor the zone for the grazing animals.  In fact these zones were causing the game to crash very quickly within a few game days after I added the animals.  When I deleted the zone and added them via manually the crashing went away and animals were properly penned. 

Dwarfs would not add the crutches, splints, bandages, thread to hospital zone on the services level.  I deleted the zone add it manually then everything started getting added.

myk002:
Fixes #1758 

see linked bug for root cause analysis

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed order of factors when calculating min temperature
 - `embark-assistant`: improved performance of surveying
+- `quickfort`: creating zones no longer causes eventual crashes
 
 ## Misc Improvements
 - `buildingplan`: set global settings from the ``DFHack#`` prompt: e.g.  ``buildingplan set boulders false``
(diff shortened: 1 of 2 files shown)


lethosor (approved):
Thanks for the detailed writeup in https://github.com/DFHack/dfhack/issues/1758#issuecomment-766392467!

My only cause for concern was this line, which makes it look like this function might have been reverse-engineered originally:
https://github.com/DFHack/dfhack/blob/2d5eb9c29d3e4d2f6a945d514c10f6a89f92b230/library/modules/Buildings.cpp#L895

Looking closer at 2303a25bdefab30bea67f32d27d35e5002448e8a, though, I think DFHack's building creation logic is intended to provide more flexibility than what DF has. And if a reasonable workflow causes a crash, that should obviously be fixed. It seems clear that DF shouldn't ever create self-references if it crashes on self-references, so we shouldn't allow it either.

It's a bit hard to test this at the source, but I've confirmed that the save from https://github.com/DFHack/dfhack/issues/1758#issuecomment-766375261 doesn't crash if self-references are removed.

Issue #1725, opened 2020-12-07
Search prompt doesn't appear the second time you enter a screen


myk002:
This is preventing quickfort blueprints from using search prompts, which makes stockpile configuration in blueprints much more verbose than it would be otherwise.

It appears to be a race condition since I can only reproduce it by programmatically sending the keys. I've never been able to manually type fast enough to trigger the problem.

Here's a blueprint that shows the search prompt working the first time you enter a screen:

```
#query
{animalsprefix}e{Right}s
```

Save this text to a file named, say, `searchtest.csv` in the `blueprints/` directory in your DF installation. Run the blueprint via `quickfort run searchtest.csv` at the `DFHack#` prompt.

The blueprint will end with the cursor active in the search widget. quickfort will report an error since we didn't end up on the main screen, but that's expected here.

And here's a blueprint that shows the search prompt missing when you enter a screen for a second time:

```
#query
{animalsprefix}e{Right}^{animalsprefix}{Right}s
```

The search prompt doesn't even appear.

Looking through [search.cpp](https://github.com/DFHack/dfhack/blob/develop/plugins/search.cpp), it's not immediately obvious why this is happening. My suspicion is that when screens are entered and exited so quickly, the wrong screen is detected at some point.

This is an ancient bug. I've been running into it for years, though this is the first time I've investigated enough to discover the missing search prompt. No need to block release, but I'd appreciate any insights you might have about debugging the issue.

lethosor:
In the second case, there is an extra (dismissed) `viewscreen_layer_stockpilest` on the screen stack, presumably the one that was dismissed earlier. DF takes a tick to clean these up, and only cleans up ones at the top of the stack. I believe the search plugin has some logic that ties a search instance ("module") to a specific instance of a screen, and doesn't allow it to appear on more than one screen at a time, so it's probably only appearing on the first instance of the screen.

lethosor:
It looks to me like changing `is_live_screen()` to exclude dismissed screens seems to trigger an update elsewhere in the search plugin, and makes your second blueprint work on my end:
```diff
--- a/plugins/search.cpp
+++ b/plugins/search.cpp
@@ -97,7 +97,7 @@ void make_text_dim(int x1, int x2, int y)
 static bool is_live_screen(const df::viewscreen *screen)
 {
     for (df::viewscreen *cur = &gview->view; cur; cur = cur->child)
-        if (cur == screen)
[...]

PR #1760, opened 2021-01-24
[quickfort] update quickfort library aliases

myk002:
discovered by ldog on the [forums](http://www.bay12forums.com/smf/index.php?topic=176889.120)

fixed aliases for tallow, iron, copper, and steel weapons; added aliases for bronze and crafts

### DIFF
--- a/data/quickfort/aliases-common.txt
+++ b/data/quickfort/aliases-common.txt
@@ -315,6 +315,10 @@ disablefinishedgoods: {finishedgoodsprefix}d^
 crafts: {finishedgoodsprefix}{Right}f{Right}{Down 9}{togglesequence 9}^
 jugs:   {finishedgoodsprefix}{Right}f{Right}{Up 2}&{Left}{Down 2}f{Down}f{Down}f^
 
+forbidcrafts: {finishedgoodsprefix}{Right 2}{Down 9}{togglesequence 9}^
+
+permitcrafts: {forbidcrafts}
+
 
 ##################################
 # cloth
--- a/data/quickfort/aliases-common.txt
+++ b/data/quickfort/aliases-common.txt
@@ -384,18 +384,21 @@ disablearmor: {armorprefix}d^
 metalarmor:  {forbidotherarmor}
 otherarmor:  {forbidmetalarmor}
 ironarmor:   {metalarmor}{forbidmetalarmor}{permitironarmor}
+bronzearmor: {metalarmor}{forbidmetalarmor}{permitbronzearmor}
 copperarmor: {metalarmor}{forbidmetalarmor}{permitcopperarmor}
 steelarmor:  {metalarmor}{forbidmetalarmor}{permitsteelarmor}
 
 forbidmetalarmor:  {armorprefix}{Right}{Down 6}f^
 forbidotherarmor:  {armorprefix}{Right}{Down 7}f^
 forbidironarmor:   {armorprefix}{Right}{Down 6}{Right}&^
+forbidbronzearmor: {armorprefix}{Right}{Down 6}{Right}{Down 6}&^
 forbidcopperarmor: {armorprefix}{Right}{Down 6}{Right}{Down 3}&^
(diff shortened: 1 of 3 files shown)


data/quickfort/aliases-common.txt
```diff
@@ -104,7 +104,7 @@ plants:         {foodprefix}b{Right}{Down 4}p^
 booze:          {foodprefix}b{Right}{Down 5}p{Down}p^
 seeds:          {foodprefix}b{Right}{Down 9}p^
 dye:            {foodprefix}b{Right}{Down 11}{Right}{Down 28}{togglesequence 4}^
-tallow:         {foodprefix}b{Right}{Down 13}{Right}{Down}{togglesequence2 811}^
+tallow:         {foodprefix}b{Right}{Down 13}{Right}{Down}{togglesequence2 822}^
```
lethosor, line 107: This feels like something that could be highly dependent on raws to me - have you verified that this works with vanilla raws now?

This is one of those cases that makes me think some special logic for certain screens could be a good idea, to make sure that aliases like these work for people with mods. In this case, the screen has an `item_names` vector (which the search plugin uses, incidentally) that could be easy to search.

I realize this is way out of the scope of this PR, and keeping compatibility with existing blueprints would be important, but might be worth spinning out into a separate issue (if there's not already one I'm forgetting)

lethosor: Update: this actually depends on how many *generated* creatures there are. I have a save with `815*2` entries in the "Fat" list. so 822 would wrap around and re-enable some fat entries at the top. Raws themselves are vanilla.

Not an easy screenshot to read, but here's what happens with 822:
image

811 leaves some entries on at the bottom:

image

myk002: *that* would explain why the number was off. I was wondering how in the world could I have gotten it wrong when I wrote it the first time.

so I see three options here:

1. use the minimum number of creature types (0 generated types, just vanilla raws) and just say that's "good enough" -- how often do generated creature types show up on fortress maps? This if course can fall flat if a modded game has fewer creature types...
2. use the search plugin. now that we've solved the disappearing prompt bug (#1725), this is feasible. I was hoping to avoid any hard dependencies on other plugins (even buildingplan is optional), but search is enabled by default so it is a relatively safe dependency (right?)
3. implement "alias functions" that are invoked like aliases but run software functions instead of just expand into static keycode sequences. I was intending to do this anyway for features like "assign minecart to hauling route", but I have a lot of open questions about the design of this feature, and I'm not sure if I want to commit myself to a particular implementation yet.

so unless the search plugin isn't as safe a dependency as I think it is, I think I'll go with option 2 in the short term and maybe convert to option 3 in the long term.

myk002: I implemented option 2. Do you think using search is sufficiently reliable (that is, are there any classes of users who don't have the plugin enabled by default?)

Where is the best place to document the dependency? I mentioned it in the alias guide, but that is targeted at blueprint creators. Where will regular users see it? It seems slightly out of place in the command reference, but if there is no place better, I can add it there

myk002: I'm not sure what failed on buildmaster.. I don't see any error messages, and each of the individual builds seems to have succeeded.

lethosor: Ooh, using `p` and `f` is smart! I didn't remember that would work. I think (3) is what I was leaning towards in my first comment as well, but this is a good shorter-term solution.

As for docs, maybe something along the lines of "some stockpile functionality currently requires the search plugin to be enabled" in more user-facing docs, then details in the alias guide could work.

I only see 5/8 of the expected build outputs at https://buildmaster.lubar.me/applications/17/builds/build?buildId=9219 - Buildmaster went down for maintenance around when you commented, so that may be related. (In fact, I think this PR might have been the built that hung for a couple hours and triggered said maintenance.) I don't see any issues, so I'll ignore it.

lethosor (approved):

PR #1761, opened 2021-01-24
[buildingplan] add 'enable all' option

myk002:
A quick way to enable planning mode for all building types. This kind of functionality is much more important now than it used to be since now there are so many supported building types.

Also modified the 'Planning Mode' status on the building placement screen to reflect whether we're in quickfort mode, enable all mode, or whether just the building type is enabled.

this setting is not persisted (just like quickfort_mode is not persisted), but it can be set from onMapLoad.init

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -765,29 +765,32 @@ and displayed with::
 
 The available settings are:
 
-+----------------+---------+---------------------------------------+
-| Setting        | Default | Description                           |
-+================+=========+=======================================+
-| blocks         | true    | Allow blocks, boulders, logs, or bars |
-+----------------+---------+ to be matched for generic "building   |
-| boulders       | true    | material" items                       |
-+----------------+---------+                                       |
-| logs           | true    |                                       |
-+----------------+---------+                                       |
-| bars           | false   |                                       |
-+----------------+---------+---------------------------------------+
-| quickfort_mode | false   | Enable compatibility mode for the     |
-|                |         | legacy Python Quickfort (not required |
-|                |         | for DFHack quickfort)                 |
-+----------------+---------+---------------------------------------+
++----------------+---------+-----------+---------------------------------------+
+| Setting        | Default | Persisted | Description                           |
++================+=========+===========+=======================================+
+| all_enabled    | false   | no        | Enable planning mode for all building |
+|                |         |           | types.                                |
++----------------+---------+-----------+---------------------------------------+
+| blocks         | true    | yes       | Allow blocks, boulders, logs, or bars |
++----------------+---------+           | to be matched for generic "building   |
+| boulders       | true    |           | material" items                       |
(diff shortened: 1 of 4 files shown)


plugins/buildingplan.cpp
```diff
@@ -765,8 +768,16 @@ struct buildingplan_place_hook : public df::viewscreen_dwarfmodest
...
+        OutputHotkeyString(x, y, "Planning Mode", interface_key::CUSTOM_SHIFT_P);
+        OutputString(COLOR_WHITE, x, y, ": ");
+        if (quickfort_mode)
+            OutputString(COLOR_YELLOW, x, y, "Quickfort", true, left_margin);
+        else if (all_enabled)
+            OutputString(COLOR_YELLOW, x, y, "All", true, left_margin);
+        else if (planmode_enabled[key])
+            OutputString(COLOR_GREEN, x, y, "On", true, left_margin);
+        else
+            OutputString(COLOR_GREY, x, y, "Off", true, left_margin);
```
myk002, line 780: Does this color scheme gel well with our best practices?

docs/changelog.txt
```diff
@@ -39,6 +39,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Misc Improvements
 - `buildingplan`: set global settings from the ``DFHack#`` prompt: e.g.  ``buildingplan set boulders false``
+- `buildingplan`: add 'enable all' option for buildingplan (so you don't have to enable all building types individually). this setting is not persisted (just like quickfort_mode is not persisted), but it can be set from onMapLoad.init
```
lethosor, line 42: These details could also go in `docs/Plugins.rst`

plugins/buildingplan.cpp
lethosor: Yep, it matches OutputToggleString, and yellow looks good to me for the non-boolean options

lethosor (approved):
Looks like it works!

docs/changelog.txt
myk002: they totally should. Thanks -- will add

myk002: done

docs/Plugins.rst
```diff
@@ -765,29 +765,32 @@ and displayed with::
...
+|                |         |           | types.                                |
++----------------+---------+-----------+---------------------------------------+
+| blocks         | true    | yes       | Allow blocks, boulders, logs, or bars |
++----------------+---------+           | to be matched for generic "building   |
+| boulders       | true    |           | material" items                       |
++----------------+---------+           |                                       |
+| logs           | true    |           |                                       |
++----------------+---------+           |                                       |
+| bars           | false   |           |                                       |
++----------------+---------+-----------+---------------------------------------+
```
lethosor, line 781: I didn't know this worked - cool!
image

lethosor (approved):

PR #1764, opened 2021-01-27
[stockpile, stockflow] resolve hotkey help text conflict

myk002:
They were overwriting each other when both plugins are enabled.

I removed stockpiles's dynamic placement code as well. it attempted to move the hotkey help text down if it covered any stockpile links, but this will no longer work since other hotkey text already takes up all the lines below stockpiles' hotkey text.

there are five plugins (that I know of) outputting hotkey help text at the bottom of stockpile screens: automelt, autotrade, stockpile, stocks, and stockflow. I suspect this is not the last conflict between that set that we'll see.

### DIFF
--- a/plugins/stockpiles/stockpiles.cpp
+++ b/plugins/stockpiles/stockpiles.cpp
@@ -409,18 +409,7 @@ struct stockpiles_import_hook : public df::viewscreen_dwarfmodest
         auto dims = Gui::getDwarfmodeViewDims();
         int left_margin = dims.menu_x1 + 1;
         int x = left_margin;
-        int y = dims.y2 - 3;
-
-        int links = 0;
-        links += sp->links.give_to_pile.size();
-        links += sp->links.take_from_pile.size();
-        links += sp->links.give_to_workshop.size();
-        links += sp->links.take_from_workshop.size();
-        if ( links + 12 >= y )
-        {
-            y += 1;
-        }
-
+        int y = dims.y2 - 7; // above automelt, autotrade, stocks, and stockflow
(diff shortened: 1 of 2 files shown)


plugins/stockpiles/stockpiles.cpp
```diff
@@ -409,18 +409,7 @@ struct stockpiles_import_hook : public df::viewscreen_dwarfmodest
...
-        links += sp->links.give_to_pile.size();
-        links += sp->links.take_from_pile.size();
-        links += sp->links.give_to_workshop.size();
-        links += sp->links.take_from_workshop.size();
-        if ( links + 12 >= y )
-        {
-            y += 1;
-        }
-
+        int y = dims.y2 - 7; // above automelt, autotrade, stocks, and stockflow
```
lethosor, line 412: Unfortunately this overlaps with autodump now (it's not always enabled, but I guess I have it enabled?)

https://github.com/DFHack/dfhack/blob/eed1f2db6ae78440c6447e8abc4ca898ce3846d8/plugins/autodump.cpp#L215

It looks like there's one spot left above DF's `ESC: Done` row, so I'll put it there for now. Definitely something that we've considered resolving somewhere more centrally, though.

image

lethosor (approved):
Addressed locally

plugins/stockpiles/stockpiles.cpp
lethosor: What I ended up going with:
image

myk002:
Be aware that when you *do* have a job selected, there's a J hotkey under the j

Issue #1733, opened 2020-12-19
Request: Save event accessible to scripts


Atkana:
It would be useful for scripts that save their own data if the pre-save hook open to plugins (or something similar) was also available in scripts. Currently, scripts can work around the lack of hooks for saving when the world is unloaded, but require hacky workarounds if they also want to save their data during autosaves.

lethosor:
For reference:

* Relevant logic that fires callbacks defined in plugins: https://github.com/DFHack/dfhack/blob/de21e0cdb86b9002cdcadabd5e8f1dbb68270467/library/Core.cpp#L2009-L2013
* Handling of the `onStateChange` event in Lua (a similar approach could probably work here - unsure if an event under `dfhack` or an eventful event would make more sense yet): https://github.com/DFHack/dfhack/blob/de21e0cdb86b9002cdcadabd5e8f1dbb68270467/library/Core.cpp#L2314 https://github.com/DFHack/dfhack/blob/de21e0cdb86b9002cdcadabd5e8f1dbb68270467/library/LuaTools.cpp#L1849-L1850

PR #1765, opened 2021-01-30
Add script-data.lua - A method for lua scripts to save data directly as json

Atkana:
This module is a resource for persistently saving lua tables directly with the json implementation. It handles either saving data globally (available anywhere) or as world data (specific to the save). World data is saved automatically whenever the world is unloaded, as well as whenever the game is saved via autosave or quicksave.

Ideally it needs somebody to go over and review a few things:

- Does the documentation make sense? Can you understand how to use the module from it? (I'm not great at articulating things >.>)
- Will the files save fine on every operating system (I've only checked for the OS I have - Windows 10)? I've not dabbled much in file system stuff before, and I can half-remember from somewhere that OSes can sometimes get picky about which slashes you use in file paths(?).
- Is this an okay way of handling this?

### DIFF
--- a/library/lua/script-data.lua
+++ b/library/lua/script-data.lua
@@ -1,0 +1,344 @@
+-- Stores lua script data persistently, either globally, or on a per-save basis, without being limited by dfhack.persistent restrictions
+-- (I've got no idea how making modules, I'm just copying what I see in other scripts :P)
+
+local _ENV = mkmodule("script-data")
+---------------------------------------------------------------------
+-- DOCUMENTATION
+--[====[
+
+script-data
+===========
+A service for saving and loading data within lua tables, providing a method of storing data persistently. This is similar to ``persist-table`` as well as the persistence module, but uses DFhack's json implementation directly.
+
+Data can be set to be global (available across every save), or world-based (where each save game can have different information saved). World-based data will automatically be saved whenever the world is exited, and can also be configured to automatically save during autosaves/quicksaves (which is enabled by default).
+
+Global data is stored in ``hack/data/script-data``, while world data is saved within ``[path to save]/script-data``. Data is stored within a json file named after the ID that the table was given by a script.
+
+Main Usage
+----------
+Storing data is done by placing any information you want to store within the lua table that's returned by ``load_global`` (for global data) or ``load_world`` (for world-based data), which will then be saved to a file when prompted. Both functions take a unique ID key to use as an identifier for that table of data. If there's pre-existing data using that ID (currently already loaded, or previously saved), the table that's returned by the functions will contain that data (in the case that it's already loaded, the table that's returned is the *exact lua table* rather than a copy of it).
+
+You should still use the load functions even if you know that no data by the given ID exists yet, as this is how you get a persistent table to store data under that ID to begin with.
+
+* :load_world(id, do_autosave): Loads world data of the given ID. World data is available only within the currently loaded save. ``do_autosave`` is an optional boolean argument, which governs whether the data should also be saved during autosave/quicksaves. This defaults to true if omitted, and there's unlikely any reason you'd need to disable it.
+
+* :load_global(id): Loads global data of the given ID. Global data is available across all saves. Global data doesn't save automatically, and so should be manually saved with the use of the ``save`` function as necessary.
+
+Data of both global and world types are both saved by the same function:
+
+* :save(id): Triggers the loaded data of the given ID to be saved. This process is done automatically for world type data, but needs to be triggered manually to save global type data.
+
+World data will be saved automatically when the world is unloaded, unless ``quit_without_saving`` has been set to true. ``quit_without_saving`` is set to true automatically during arena mode, as there's no non-hacky way to detect if the player chose to save and quit, or quit without saving.
+
+Notes/Limitations
+-----------------
+Notes on IDs:
+
+* IDs should be uniquely named. There is overlap between world and global IDs, so if your script uses both they should each be uniquely named (e.g. ``my_script_global`` and ``my_script_world``, rather than both being named ``my_script``).
+* Don't use characters that aren't file name friendly, as the IDs are used as the names for their json save files.
+
+There are a few limitations of DFhack's particular json implementation (``jf-json``) to bear in mind:
+
+* You cannot use both string keys and numerical keys within the same table.
+* Don't use numerical keys for non-array tables, and don't make sparse arrays, as ``jf-json`` will interpret any table with numbered keys as an array, and attempt to fill in every missing entry when writing to a file. For example, if you were to make a table that only has an entry at index 1000, when it's saved to a file the code will write in additional entries of ``nil`` for all 999 indexes leading up to that entry, bloating the file. If you're using numerical IDs as your keys (like unit IDs, etc.), consider storing them as strings rather than numbers.
+
+Example
+-------
+This is a simple counter that counts how many times the code has been run in the current save, maintaining its count between sessions. Because it's world data, the value will be differ between each save game::
+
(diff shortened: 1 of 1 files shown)


library/lua/script-data.lua
```diff
@@ -0,0 +1,344 @@
...
+
+    -- Stop monitoring autosaves
+    if loop_timer ~= nil then
+      dfhack.timeout_active(loop_timer, nil)
+    end
+  end
+end
+
+function start_loop()
+  loop_timer = dfhack.timeout(1, "frames", autosave_check_loop)
```
lethosor, line 310: I'm tempted to say that we should hold off on this until we have a solution for #1733. But I'll test it as-is for now, and I'll try to prioritize a proper solution there.

library/lua/script-data.lua
```diff
@@ -0,0 +1,344 @@
...
+  recorded_save_path = dfhack.getSavePath()
+
+  -- Begin monitoring for autosaves / quicksaves
+  -- Since autosaves are only actually relevant in fortress mode, there's no reason to start checking for them outside of that mode
+  if dfhack.world.isFortressMode() then
+    start_loop()
+  end
+
+  -- Automatically disable saving on exit if in arena mode
+  -- This is because it's the only current mode(?) that allows for quitting without saving, and there's no way (without some hackery) to really detect that
```
lethosor, line 269: I was about to suggest checking `df.global.save_on_exit`, which should theoretically be the inverse of your `quit_without_saving`, but it looks like that is set to true in arena mode, unfortunately, now that arena mode supports saving. I suspect it's probably immediately set to false if you select "abort game" in the options menu, but I haven't confirmed. Could be worth keeping in mind in case it's useful elsewhere. See below for more.

library/lua/script-data.lua
```diff
@@ -0,0 +1,344 @@
...
+  end
+end
+
+dfhack.onStateChange["script-data"] = function(event)
+  if event == SC_WORLD_LOADED then
+    world_load_setup()
+
+  elseif event == SC_WORLD_UNLOADED then
+    -- Trigger the saves for all `world` entries, provided quitting without saving is disabled
+    if quit_without_saving == false then
```
lethosor, line 284: This appears to be the only place you're reading from `quit_without_saving` - could it just be replaced with `isArena()` here? That would avoid the need to keep track of additional global state, if it works.

If you're doing this because isArena() returns false by the time SC_WORLD_UNLOADED fires, maybe it would work under SC_MAP_UNLOADED? Are there any situations where you'd want to support saving data if a map isn't loaded? World generation is the only one I can think of off the top of my head.

In any case, our global "save" event should fire at the right times currently, so #1733 could again make this logic unnecessary (I hope).

library/lua/script-data.lua
```diff
@@ -0,0 +1,344 @@
...
+  end
+
+  json.encode_file(get_loaded_data(id).data, path)
+end
+
+-- Trigger saves for all persistent tables which have the autosave flag currently set
+-- This is naturally triggered by this script whenever an autosave / quicksave is detected
+function save_autosave()
+  for id, entry in pairs(loaded_data) do
+    if entry.autosave == true then
```
lethosor, line 224: A minor stylistic point: I typically suggest replacing `if x == true` and `if x == false` with `if x` and `if not x`, respectively. `nil` and `false` are the only things that implicitly evaluate to false in Lua when treated as a boolean, so you could also replace `if x == nil` with `if not x` in most cases (although the fact that 0 evaluates to true could be confusing, particularly to anyone less familiar with Lua implementation details, so I don't tend to push this one as hard).

lethosor (commented):
I took a look over this briefly - you're doing modules the right way, and the docs make sense to me (thanks for the example!). I'll try to test more thoroughly as well.

I left a couple comments just from looking at it on here. Most of them are not particularly useful as they involve me hoping for a solution for #1533, but some are unrelated to that.

library/lua/script-data.lua
Atkana: After a bit of thought, I've remembered that actually both Fortress and Adventure mode can also be exited without saving, so they'd also likely make use of `quit_without_saving`... Next time I get around to editing this current form I'll have to throw something together (either checking `df.global.save_on_exit` that you brought up before, or defaulting to `quit_without_saving` starting off `false` until a map is loaded).

For use, the only thing I can really think of too is worldgen, but I find it doesn't hurt to provide more options to modders in case there's some niche circumstances I didn't foresee.

Though yeah, if/when that event gets added, this'll be a lot more straightforward :P

library/lua/script-data.lua
Atkana: I've been trying to intentionally structure checks like this nowadays to improve readability, though at the slight cost of elegance (and I believe it's safe to do in this check because autosave can only ever be `true` or `false`, never `nil`). I originally learned lua - and how to use DFhack especially - by reading through scripts made by other people, and I personally find that making the conditions obvious means it's easier to tell what the checks mean to somebody reading the script who's learning lua, or transitioning from another language.

library/lua/script-data.lua
lethosor: > After a bit of thought, I've remembered that actually both Fortress and Adventure mode can also be exited without saving, so they'd also likely make use of `quit_without_saving`

Oh, are you referring to the ability to abandon a game? Took me a second to think of that too.

I think it would be worth double-checking what our current plugin-facing "save" event does in this case. There is an argument to be made for saving data in this case, even if the user didn't select the "save" option from the menu. After all, fortresses and adventurers remain part of the world (or at least history), even the changes made between the last save and abandoning/retiring.

Atkana: Ah, I was more thinking about how you can load a save after worldgen or losing / abandoning / retiring and go through the world update screen (where time advances by 2 weeks), but then back out without making a fort / adventurer, with none of what happened during the world update being saved. I've always thought it's a bit weird that it works that way, enough so that I often forget it's a thing :b

lethosor: Ah, that's another good case to consider. Pretty sure our plugin-facing save event would not fire in that case. (I'm not actually sure if the "load" event would, though - I'll double-check). I am surprised that loading a fortress or just opening up the embark site selection screen both set the game mode and game type to the same value... I suppose that could throw off some scripts/tools, but they should be able to handle this by only enabling when SC_MAP_LOADED fires or if isMapLoaded() returns true.

lethosor: Update: loading a world from the "start playing" menu *does* fire a "load" event, but the "save" event only fires when the actual DF save screen comes up (or on autosaves). I think the intention is for tools to watch for the "save" event and write any last-minute changes to the persistent tree before it's written to disk. Not seeing any example of plugins that do this currently, but I think this is basically the behavior you're looking to match here, yes?

Atkana: Yep! Saving whenever the game does (be that during save + quit, or autosaves), and skipping saving when the game skips it, is what's desired (for the world-based data, at least).

cppcooper:
I guess I missed the boat to do this huh. I just converted my code for this into a class, I was gonna see about making it internal and a bit more generalized next. https://github.com/cppcooper/dfhack-scripts/blob/d07ce6688ecea340d4724646dbe319eca75006ea/dwarf-op2.lua#L543

myk002:
I'm going through older PRs, and this looks interesting, but I'm a little confused. What benefits does this offer over the current feature set of `persist-table` and `json`? For example, would `gui/launcher` or `emigration` benefit from using `script-data` instead of the existing APIs?

myk002:
This looks cool, but I think it has been superseded by the `dfhack.persistent` API

PR #1767, opened 2021-01-31
[embark-assistant] Refactor survey::survey_mid_level_tile and fix small bugs identified during the refactoring

bseiller:
@PatrikLundell
As discussed here: http://www.bay12forums.com/smf/index.php?topic=169634.msg8241576#msg8241576
and here: http://www.bay12forums.com/smf/index.php?topic=169634.msg8242311#msg8242311


PatrikLundell (commented):
There are a lot of "survey_results->at(x).at(y)" and "mlt->at(i).at(k)" left here. I think they would benefit from local variable treatment as well (being careful to not screw up the local variable for the cases where the indices are e.g. "i - 1")? Or have you already done an assessment and determined that there wouldn't be any benefit?

bseiller:
Yes, good, valid advice.
And: Way ahead of you, at least for the second part :) 
Have a look at the timestamps here:
https://github.com/bseiller/dfhack/blame/c9fc554f9362cc8163ebfa3a8c2ec47309f317bc/plugins/embark-assistant/survey.cpp#L2253

PatrikLundell:
Strange I haven't seen that before ;)

Good to see our opinions align.

plugins/embark-assistant/survey.cpp
```diff
@@ -1351,92 +1348,92 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data
...
-        tile->north_row[i].biome_offset = mlt->at(i).at(0).biome_offset;
-        tile->south_row[i].biome_offset = mlt->at(i).at(15).biome_offset;
-        tile->west_column[i].biome_offset = mlt->at(0).at(i).biome_offset;
-        tile->east_column[i].biome_offset = mlt->at(15).at(i).biome_offset;
-
-        tile->north_row[i].trees = mlt->at(i).at(0).trees;
-        tile->south_row[i].trees = mlt->at(i).at(15).trees;
-        tile->west_column[i].trees = mlt->at(0).at(i).trees;
-        tile->east_column[i].trees = mlt->at(15).at(i).trees;
+        tile.north_row[i].aquifer = mlt->at(i).at(0).aquifer;
```
bseiller, line 1363: @PatrikLundell:
As I've said I refactored this in my index feature branch already, but thinking about it again I found my solution lacking.
My pet peeve is code duplication/repetition starting with the third time, as twice is often kind of okay.
And in this loop the same code is repeated 4 times.
So I measured what happened if we introduce a function to handle the copy process of the incursion data in comparison to the straight assignments.
Guess what :)
The following code is even faster than the original - not that it takes long anyways in the first place (13ms vs. 12,2ms for a large region on my system) - but the refactored version is easier to maintain as well.

```
void inline copy_incursion_values(embark_assist::defs::mid_level_tile_incursion_base &target, const embark_assist::defs::mid_level_tile_incursion_base &source) {
    target.aquifer = source.aquifer;
    target.clay = source.clay;
    target.sand = source.sand;
    target.soil_depth = source.soil_depth;
    target.elevation = source.elevation;
    target.biome_offset = source.biome_offset;
    target.trees = source.trees;
    target.savagery_level = source.savagery_level;
    target.evilness_level = source.evilness_level;
}

void refactored_incursion_loop(embark_assist::defs::region_tile_datum_new &tile, const embark_assist::defs::mid_level_tiles_new *mlt,
    const int8_t biome_corner[16][16], const int8_t biome_x[16][16], const int8_t biome_y[16][16]) {
    for (uint32_t i = 0; i < 257 * 257; i++) {
        for (uint8_t i = 0; i < 16; i++) {
            copy_incursion_values(tile.north_row[i], mlt->at(i).at(0));
            copy_incursion_values(tile.south_row[i], mlt->at(i).at(15));
            copy_incursion_values(tile.west_column[i], mlt->at(0).at(i));
            copy_incursion_values(tile.east_column[i], mlt->at(15).at(i));

            tile.north_corner_selection[i] = biome_corner[i][0];
            tile.west_corner_selection[i] = biome_corner[0][i];
            tile.north_row_biome_x[i] = biome_x[i][0];
            tile.west_column_biome_y[i] = biome_y[0][i];
        }
    }
}
```

PatrikLundell: I agree introduction of the function is a better implementation that's easier to maintain. So, please, go ahead.

I wouldn't trust the performance measurement too much, though, as a compiler COULD optimize the code in a way that won't happen when used as intended (the corner assignments can be moved outside of the outer loop so it's done only 16 times). However, given the times we're talking about performance improvements aren't important.

bseiller: > I agree introduction of the function is a better implementation that's easier to maintain. So, please, go ahead.

Splendid - will do.

> I wouldn't trust the performance measurement too much, though, as a compiler COULD optimize the code in a way that won't happen when used as intended 

Very true.
> (the corner assignments can be moved outside of the outer loop so it's done only 16 times). 

Oh, the outer loop is just a helper construct to simulate all calls to `survey_mid_level_tile` in a large world. And in the feature branch I had to change around some things and ended up with the following lines, which remove the need for a loop completely:
```
std::copy(&details->edges.biome_x[0][0], &details->edges.biome_x[0][0] + 16 * 16, &tile.biome_x[0][0]);
std::copy(&details->edges.biome_y[0][0], &details->edges.biome_y[0][0] + 16 * 16, &tile.biome_y[0][0]);
std::copy(&details->edges.biome_corner[0][0], &details->edges.biome_corner[0][0] + 16 * 16, &tile.biome_corner[0][0]);
```

plugins/embark-assistant/survey.cpp
```diff
@@ -1261,60 +1257,61 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data
...
+                    tile.min_river_size = mid_level_tile.river_size;
                 }
 
-                if (survey_results->at(x).at(y).max_river_size < mlt->at(i).at(k).river_size) {
-                    survey_results->at(x).at(y).max_river_size = mlt->at(i).at(k).river_size;
+                if (tile.max_river_size < mid_level_tile.river_size) {
+                    tile.max_river_size = mid_level_tile.river_size;
                 }
 
                 if (i < 15 &&
```
bseiller, line 1301: > 2.iteration: 3.iteration dealt with "mlt->at(i).at(k)", but I think the survey.cpp line 1305 clause could use a local const for "mlt->at(i+1).(k)" by splitting the "if" statement into one for "i<15" and another one inside it for the rest after the const had been declared. The corresponding treatment could be applied to the "k<15" "if" clause immediately after.

You mean like so?
```
if (i < 15) {
	const embark_assist::defs::mid_level_tile &left_neighbour = mlt->at(i + 1).at(k);
	if (left_neighbour.river_size != embark_assist::defs::river_sizes::None &&
		abs(mid_level_tile.river_elevation - left_neighbour.river_elevation) > tile.max_waterfall) {
		tile.max_waterfall = abs(mid_level_tile.river_elevation - left_neighbour.river_elevation);
	}
}
```

plugins/embark-assistant/survey.cpp
PatrikLundell: I realized it was a simulation, but pointed at a possible loop hole in it, which a smart compiler could do.

I'm not familiar with std::copy (looking at a definition page), but from what I can understand you copy 256 elements from each of biome_x/y/corner rather than the 64 at the edges, which I guess means you store 256 rather 64 values as well..

plugins/embark-assistant/survey.cpp
PatrikLundell: Yes

plugins/embark-assistant/survey.cpp
bseiller: > I realized it was a simulation, but pointed at a possible loop hole in it, which a smart compiler could do.

Ah, now I get it - thanks for the clarification!

> which I guess means you store 256

Correct, to make the concurrent processing of internal and external incursions possible all `biome_*` data has to be available even if the "cursor" has moved to the next world tile already.

PatrikLundell (approved):
Looks good.

PatrikLundell (approved):
A nobrainer, as it's already been discussed.

PatrikLundell (approved):
I assume you've seen the whitespace complaint in 7.iteration. Apart from that that iteration looks good.

PatrikLundell (approved):
Looks good.

PatrikLundell (approved):
Again, makes sense.

PatrikLundell (approved):
Looks good.

bseiller:
Okay, that's it for the refactoring of survey::survey_mid_level_tile for now.
I have one little tweak performance left that speeds up the survey phase, but it needs changes in matcher as well, so I'll put it into a separate PR.
Is the entry in the changelog comprehensible?

PatrikLundell (commented):
I assume this refers to "fixing use of wrong indices, discovered during refactoring...". If so, I'd probably say "fixed bug in soil depth determination for ocean tiles".

You can get 0 when it shouldn't be and non zero when it should be, so you can get failures in both directions (on top of criteria that can be defined in both directions).

PatrikLundell (approved):
Looks good.

PR #1769, opened 2021-02-02
[autohauler] allow alchemist skill (aka the "no hauling labors" flag) to be toggled in-game

myk002:
since autohauler uses it as a flag.


quietust:
Would it not make more sense to change autolabor to _not_ use the Alchemy labor (since it's possible that a mod might be legitimately using it with custom reactions)?

myk002:
that sounds like a good idea, though we probably still don't want to allow alchemy to be enabled along with its group in manipulator. how can I get autohauler to change the raws?

quietust:
Modifying the raws can be done by basically copying the `Units::isValidLabor` function and changing it to write instead of reading.

I'm not sure it makes sense to exclude Alchemy from Toggle Group, because that would make it inconsistent with DF's own labor menu. Also, we shouldn't assume that autohauler is the **only** thing that uses the Alchemy labor - players might be using special mods that use that labor for actual tasks, and they would expect Shift+Enter to toggle it along with everything else.

quietust:
Example code for marking the labor as permitted:
```
    df::historical_entity *entity = df::historical_entity::find(ui->civ_id);
    if (entity && entity->entity_raw))
        entity->entity_raw->jobs.permitted_labor[unit_labor::ALCHEMIST] = true;
```
The only issue is that the above code only works with your own civilization type - in order to handle resident Humans/Elves/etc., it would have to iterate across world->raws.entities[] and update all of them.

myk002:
done, though I have to admit I don't fully understand the implications of the filtering. why is it important that alchemist *not* become valid for non-controllable races like vault guardians or kobolds? Can we simplify the filter to just
```
if (!world->raws.entities[i])
    continue;
```
? I didn't copy over the `controllable_civs` counting logic from [embark_assistant/finder_ui.cpp](https://github.com/DFHack/dfhack/blob/develop/plugins/embark-assistant/finder_ui.cpp#L355) since it didn't seem applicable, but I don't really have the background to know if that's the right choice.

plugins/manipulator.cpp
```diff
@@ -2291,6 +2301,9 @@ DFhackCExport command_result plugin_enable(color_ostream &out, bool enable)
             return CR_FAILURE;
 
         is_enabled = enable;
+
+        if (is_enabled)
```
myk002, line 2305: Is there any harm in leaving alchemist enabled even if manipulator is later disabled? it seems disabling it could lead to "dueling plugins" if some other plugin wants to keep it enabled for some reason.

docs/Lua API.rst
```diff
@@ -1270,6 +1270,15 @@ Units module
   Returns the age of the unit in years as a floating-point value.
   If ``true_age`` is true, ignores false identities.
 
+* ``dfhack.units.isValidLabor(unit, unit_labor)``
+
+  Returns whether the indicated labor is settable for the given unit.
+
+* ``dfhack.units.setLaborValidity(unit_labor)``
```
lethosor, line 1277: Missing `, isValid` here

plugins/manipulator.cpp
```diff
@@ -2306,6 +2319,20 @@ DFhackCExport command_result plugin_init ( color_ostream &out, vector <PluginCom
     return CR_OK;
 }
 
+DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_change_event event)
+{
+    switch (event) {
+    case SC_MAP_LOADED:
+        if (is_enabled)
+            enable_alchemist(out);
```
lethosor, line 2327: I feel like this should be done by autohauler instead. Not everyone who uses manipulator necessarily uses autohauler, and this will affect other things (e.g. `tweak block-labors`).

plugins/manipulator.cpp
lethosor: I think it would make sense to leave enabled (although perhaps from autohauler, per my other comment)

docs/Lua API.rst
myk002: done

plugins/manipulator.cpp
myk002: done

plugins/manipulator.cpp
myk002: thanks! moved to autohauler and left enabled

PatrikLundell:
Comment on earlier discussion regarding civ identification:

Residents ought to follow the rules for the player civ as they become members of it when they become residents. The Embark Assistant code finds neighbors, and thus has to filter out "neighbors" that don't apply according to DF, which is why vault guardians and cavern civs are removed from the list: there is no deeper meaning behind it than to emulate DF's behavior (but still adding kobolds...).

If you were/will be able to affect characters belonging to other civs you'd need to be able to modify the appropriate flags for them, but otherwise I would guess that you only need to modify the player civ, but I haven't any direct knowledge of it being so.

myk002:
What would be the practical effect of making the alchemist labor valid for all races? Even if it were valid for a non-playable race, how could a player even affect the value of the flag? Say we do make alchemist a valid labor for vault guardians - what does that mean for the player?

PatrikLundell:
You're manipulating civs, not races: individuals may belong to a race with one entity definition->set of civs, but still be a member of a different civ, and thus be bound by that civ's restrictions (a goblin player fortress member still won't be able to make loincloths, for instance).
Making a flag the game doesn't use natively available to a civ you're not playing likely has no effect at all. If you DID change it, I assume you might be able to manipulate it on invaders/traders/visitors, but since it's not tied to any game activity you'd still have to hack any activity/effect in yourself, in which case you might just perform the hack without the flag.

myk002:
So what I'm hearing is that it's ok to make the alchemist labor valid for all civs, and I can simplify this PR by removing the copied filter code. Is that right?

PatrikLundell:
It's what I believe, but I have no insight whatsoever in how the PR code is used, so it would be good with confirmation from that side of things. I would use quietust's first snippet to affect only my own civ, but would test it with both residents and full civ members of other races (ideally checked to come from non dwarven civs). On the other hand, I can't see any harm coming from affecting other civs, apart from future maintenance when people try to figure out why it was implemented that way (e.g. when alchemy is actually implemented and the code will have to be reviewed again). Testing for side effects would be tricky, though, as you'd have no idea what to look for.

myk002:
I rewrote it locally to just allow for player civ. Will test with acquired citizens on my next playtesting runthrough of dreamfort.

PatrikLundell:
Sounds like a good approach.

myk002:
Still trying to get a situation where I can test other civs joining my fort. I had a number of long-term residents, but then a visiting poet turned into a werepanther and ate everybody. Such is life. Trying again.

lethosor:
For what it's worth, I've found that buggy saves from Mantis/DFFD can be a good way to test certain scenarios without having to reproduce them yourself. Of course, the trick is finding saves that don't crash and don't have bugs relating to the things you want to test.

myk002:
Thanks -- I'll give that a try. I'm up to 33 long term residents and nobody seems to want to join my fort permanently!

PatrikLundell:
The second petition comes two years after the first (accepted) one, and merc and monster slayers don't petition for full residency. Scholars and performers do, however, although depression can block it.

myk002:
Played for another two years (Thanks Patrik!) and I now have two human citizens, and they can have the alchemist labor toggled. I pushed the simpler implementation that only enables the labor for the player civ.

myk002:
On that note, if anyone would like to take a peek at a fully built dreamfort, I uploaded my game to dffd: https://dffd.bay12games.com/file.php?id=15434

lethosor (approved):
Looks like it works! Your save was easy to test on :) (turns out there's a pending citizenship request by a human named Istrul, and once accepted, toggling alchemy worked as expected)

plugins/autohauler.cpp
```diff
@@ -565,10 +565,20 @@ static void cleanup_state()
     labor_infos.clear();
 }
 
+static void enable_alchemist(color_ostream &out)
+{
+    if (!Units::setLaborValidity(unit_labor::ALCHEMIST, true))
+    {
+        // informational only; this is a non-fatal error
+        out.printerr("manipulator: Could not flag Alchemist as a valid skill; Alchemist will not"
```
lethosor, line 573: I think `manipulator` should probably be changed to `autohauler` - can do on my end

myk002: whoops! thanks for the fix

PR #1770, opened 2021-02-02
[orders] don't crash on importing malformed json

myk002:
to reproduce: take a regular exported orders file, remove a comma between two elements, try to import.

this PR also adds the filename to error messages that have to do with file read errors.


lethosor (approved):
Good catch!

myk002:
> Good catch!

Ummm was that a pun 😛

PR #1771, opened 2021-02-02
[embark-assistant] Improve performance of surveying => faster search

bseiller:
@PatrikLundell: as announced.
I'll do some comparisons between the current version and the one with the following changes "in place" before removing the "WIP" just to be sure - but the code is mostly taken from the index feature branch which I scrutinized repeatedly concerning the performance of those changes.

Replace the local/automatic mid_level_tiles variable in matcher::match_world_tile with a dynamic one that is created and initialized once during the setup phase. The dynamic part of the contained 768 vectors is being allocated on the heap in both cases - which made the repeated instantiations of the automatic variable so slow/expensive in the first place.

Also replace calls to vector<bool>.resize in nested loops with direct assignments to those vectors, which curiously even after a lot of profiling is the fastest way to reset the inorganic vectors - at least on Windows - as there seems not to exist a template specialization for std::fill in MSVS C++11 in regards to std::vector<bool>.

- embark-assistant.cpp: Replace 2 local/automatic mid_level_tiles variables with a single dynamic variable that is created and initialized during setup as well; add calls to matcher::setup() and matcher::shutdown()
- matcher.cpp/.h: add state with mid_level_tiles member; add setup and shutdown functions
- survey.cpp: add function reset_mlt_inorganics as replacement for the looped calls to vector::resize as all inorganic vectors are now expected to have the proper size when entering survey::survey_mid_level_tile

### DIFF
--- a/plugins/embark-assistant/matcher.cpp
+++ b/plugins/embark-assistant/matcher.cpp
@@ -64,6 +64,12 @@ namespace embark_assist {
             bool mineral_3;
         };
 
+        struct states {
+            embark_assist::defs::mid_level_tiles mlt;
+        };
+
+        static states *state;
+
         //=======================================================================================
 
         void process_embark_incursion(matcher_info *result,
@@ -2681,14 +2687,13 @@ namespace embark_assist {
             uint16_t y) {
 
 //            color_ostream_proxy out(Core::getInstance().getConsole());
-            embark_assist::defs::mid_level_tiles mlt;
 
             embark_assist::survey::survey_mid_level_tile(geo_summary,
                 survey_results,
-                &mlt);
+                &state->mlt);
 
             mid_level_tile_match(survey_results,
-                &mlt,
+                &state->mlt,
                 x,
                 y,
                 finder,
@@ -2715,6 +2720,19 @@ namespace embark_assist {
 
 //=======================================================================================
 //  Visible operations
+//=======================================================================================
+
+void embark_assist::matcher::setup() {
+    embark_assist::matcher::state = new(embark_assist::matcher::states);
+    embark_assist::survey::initiate(&state->mlt);
+}
+
+//=======================================================================================
+
+void embark_assist::matcher::shutdown() {
+    delete state;
+}
+
 //=======================================================================================
 
 void embark_assist::matcher::move_cursor(uint16_t x, uint16_t y) {
--- a/plugins/embark-assistant/matcher.cpp
+++ b/plugins/embark-assistant/matcher.cpp
@@ -68,7 +68,7 @@ namespace embark_assist {
             embark_assist::defs::mid_level_tiles mlt;
         };
 
-        static states *state;
+        static states *state = nullptr;
 
         //=======================================================================================
 
@@ -2731,6 +2731,7 @@ void embark_assist::matcher::setup() {
 
 void embark_assist::matcher::shutdown() {
     delete state;
+    state = nullptr;
 }
 
 //=======================================================================================
--- a/plugins/embark-assistant/survey.cpp
+++ b/plugins/embark-assistant/survey.cpp
@@ -902,6 +902,21 @@ void embark_assist::survey::high_level_world_survey(embark_assist::defs::geo_dat
 
 //=================================================================================
 
+void reset_mlt_inorganics(embark_assist::defs::mid_level_tiles *mlts, const uint16_t max_inorganic) {
+    for (uint8_t i = 0; i < 16; i++) {
+        for (uint8_t k = 0; k < 16; k++) {
+            embark_assist::defs::mid_level_tile &mlt = mlts->at(i).at(k);
+            for (uint16_t i = 0; i < max_inorganic; i++) {
+                mlt.metals[i] = false;
+                mlt.economics[i] = false;
+                mlt.minerals[i] = false;
+            }
+        }
+    }
+}
+
+//=================================================================================
+
 void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data *geo_summary,
     embark_assist::defs::world_tile_data *survey_results,
     embark_assist::defs::mid_level_tiles *mlt) {
@@ -935,13 +950,7 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data 
         tile->minerals[i] = 0;
     }
 
-    for (uint8_t i = 0; i < 16; i++) {
-        for (uint8_t k = 0; k < 16; k++) {
-            mlt->at(i).at(k).metals.resize(state->max_inorganic);
-            mlt->at(i).at(k).economics.resize(state->max_inorganic);
-            mlt->at(i).at(k).minerals.resize(state->max_inorganic);
-        }
-    }
+    reset_mlt_inorganics(mlt, state->max_inorganic);
 
     for (uint8_t i = 1; i < 10; i++) survey_results->at(x).at(y).biome_index[i] = -1;
 
--- a/plugins/embark-assistant/survey.cpp
+++ b/plugins/embark-assistant/survey.cpp
@@ -75,7 +75,7 @@ namespace embark_assist {
             uint16_t max_inorganic;
         };
 
-        static states *state;
+        static states *state = nullptr;
 
         //=======================================================================================
 
@@ -2548,5 +2548,6 @@ void embark_assist::survey::survey_embark(embark_assist::defs::mid_level_tiles *
 
 void embark_assist::survey::shutdown() {
     delete state;
+    state = nullptr;
 }
 
--- a/plugins/embark-assistant/survey.cpp
+++ b/plugins/embark-assistant/survey.cpp
@@ -902,14 +902,15 @@ void embark_assist::survey::high_level_world_survey(embark_assist::defs::geo_dat
 
 //=================================================================================
 
-void reset_mlt_inorganics(embark_assist::defs::mid_level_tiles *mlts, const uint16_t max_inorganic) {
+void reset_mlt_inorganics(embark_assist::defs::mid_level_tiles &mlts) {
+    const uint16_t size = mlts[0][0].metals.size();
     for (uint8_t i = 0; i < 16; i++) {
         for (uint8_t k = 0; k < 16; k++) {
-            embark_assist::defs::mid_level_tile &mlt = mlts->at(i).at(k);
-            for (uint16_t i = 0; i < max_inorganic; i++) {
-                mlt.metals[i] = false;
-                mlt.economics[i] = false;
-                mlt.minerals[i] = false;
+            embark_assist::defs::mid_level_tile &mlt = mlts[i][k];
+            for (uint16_t l = 0; l < size; l++) {
+                mlt.metals[l] = false;
+                mlt.economics[l] = false;
+                mlt.minerals[l] = false;
             }
         }
     }
@@ -950,7 +951,7 @@ void embark_assist::survey::survey_mid_level_tile(embark_assist::defs::geo_data 
         tile->minerals[i] = 0;
     }
 
-    reset_mlt_inorganics(mlt, state->max_inorganic);
+    reset_mlt_inorganics(*mlt);
 
     for (uint8_t i = 1; i < 10; i++) survey_results->at(x).at(y).biome_index[i] = -1;
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `embark-assistant`: fixed bug in soil depth determination for ocean tiles
 
+## Misc Improvements
+- `embark-assistant`: slightly improved performance of surveying
+
 # 0.47.04-r5
 
 ## Fixes
--- a/plugins/embark-assistant/embark-assistant.cpp
+++ b/plugins/embark-assistant/embark-assistant.cpp
@@ -285,7 +285,6 @@ command_result embark_assistant(color_ostream &out, std::vector <std::string> & 
     }
 
     embark_assist::main::state = new embark_assist::main::states;
-    embark_assist::survey::initiate(&embark_assist::main::state->mlt);
 
     embark_assist::main::state->match_iterator.active = false;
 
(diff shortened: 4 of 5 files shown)


PatrikLundell (changes requested):
survey.cpp line 909-912: I dislike overloading loop variables due to the risk of confusion and introduction of bugs when updating the code, so I suggest changing the loop variable name to "l" (i.e. lower case Lima).

Also, do you really need the max_inorganic parameter? It ought to be possible to use mlt.metals.size() instead, as the code relies on them being the same (I guess it's a relic from when resize was used).

bseiller:
Good findings!
I was considering to use "n" instead" of "l" as it is sometimes hard to not mistake it for an upper-case "I" (as in Indigo), but it seemed fine.

Looking into the (constant) costs of std::vector::size() I stumbled upon this link:
https://www.acodersjourney.com/6-tips-supercharge-cpp-11-vector-performance/
which had an interesting point concerning std::vector::at()
After actually measuring a small difference in speed I incorporated it as well.
Also it looks/reads much nicer, don't you think?

PatrikLundell (approved):
Looks good. Interesting find regarding "at" vs "[]". I really don't like "at" at all as it's ugly and a function compilers apparently aren't smart enough to handle efficiently.

The main reason I've use "at" is that "[]" can't be used when you start with a "normal" variable, and I failed to massage it on the fly to morph the initial "->" pointer into the "." that can be substituted by "[]".

Of course, this opens up for the rest of the code to get the same treatment...

bseiller:
Ok - after some funny hide-and-seek with the sources from 0.47.04-r5 tag I was able to take some measurements.
Of  course I had to fix a crash first - caused by the one line that was not in the feature branch... 
Anyways the original version takes ~520 seconds to complete the search within a large world and a specific set of search criteria.
The adapted version takes ~500 seconds with the same world and search criteria. 
So it's 3,8% faster. Not much, but the change isn't big as well.

Both the original and the adapted version already include the changes from #1755, #1756 and #1757.
I'll try to find the time to measure the difference between the current version (all changes applied) and the version before all PRs.

#
> The main reason I've use "at" is that "[]" can't be used when you start with a "normal" variable, and I failed to massage it on the fly to morph the initial "->" pointer into the "." that can be substituted by "[]".

I was wondering about that when I started looking into the code way back.
 
> Of course, this opens up for the rest of the code to get the same treatment...

Haha, sure. Some extended search-and-replace and a 30 day evaluation of ReSharper C++ that currently sleeps on my system should help with that - but I'd like to get the remaining 2,5 (?) issues in the index branch fixed and then the merging done first. Otherwise I'll make merging harder again instead of easier.

PatrikLundell:
I think a 20 second cut of the search time is quite significant, even though it might not be a huge percentage.

Improving the code with use of indexing rather than function calls for indices isn't urgent, just something to remember when there's an opening for doing it. Although it ought to improve the performance to a small extent, I think legibility is an even more important driver for that.

PatrikLundell (approved):
Looks good

plugins/embark-assistant/survey.cpp
```diff
@@ -908,6 +908,22 @@ void inline copy_incursion_values(embark_assist::defs::mid_level_tile_incursion_
...
+void reset_mlt_inorganics(embark_assist::defs::mid_level_tiles &mlts) {
+    const uint16_t size = mlts[0][0].metals.size();
+    for (uint8_t i = 0; i < 16; i++) {
+        for (uint8_t k = 0; k < 16; k++) {
+            embark_assist::defs::mid_level_tile &mlt = mlts[i][k];
+            for (uint16_t l = 0; l < size; l++) {
+                mlt.metals[l] = false;
+                mlt.economics[l] = false;
+                mlt.minerals[l] = false;
+            }
```
lethosor, line 920: `.at()` also checks the validity of the provided index and raises an `out_of_range_` exception if the index is out of bounds; `[]` does not do this, which could explain some of the speedups you're seeing if you've replaced `at()` with `[]`. I had assumed part of the reason for using `at()` earlier was for safety, but it seems like that may be less important here.

You are making an assumption that all of these vectors match the size of `mlts[0][0].metals`, though, and losing the safety checks from earlier. I might suggest this instead - it should be pretty well-optimized, and it will always use the right size:
```suggestion
            std::fill(mlt.metals.begin(), mlt.metals.end(), false);
            std::fill(mlt.economics.begin(), mlt.economics.end(), false);
            std::fill(mlt.minerals.begin(), mlt.minerals.end(), false);
```

plugins/embark-assistant/matcher.cpp
```diff
@@ -2717,6 +2722,19 @@ namespace embark_assist {
...
 
+void embark_assist::matcher::setup() {
+    embark_assist::matcher::state = new(embark_assist::matcher::states);
+    embark_assist::survey::initiate(&state->mlt);
+}
+
+//=======================================================================================
+
+void embark_assist::matcher::shutdown() {
+    delete state;
```
lethosor, line 2733: I realize there's not a destructor where this could be easily placed, so to prevent accidental double-frees down the road, I might suggest also setting `state = nullptr` here.

PatrikLundell:
So "at" is an attempt to introduce some high level language functionality into C(++)...

All 3 of the vectors are assumed to be the length of the non generated part of the inorganics DF vector and are indexed with the same indices (i.e. the indices of the minerals found in the geo biome). If they don't have that size there is a bug, and entering elements into the bugged vectors will not work properly.
However, I don't have any objections to using std::fill (in theory it would be possible for std::fill to be more efficient than a loop by filling multiple elements at the same time using 64 bit assignments for as long as possible. However, I don't expect compilers/the std implementations to actually be smart enough to do that).

Guarding against double frees is good practice, so that's a good suggestion.

plugins/embark-assistant/survey.cpp
bseiller: First off :+1:  for taking the time to look into the PR in such detail.

I'd love to use `std::fill` in fact I did initially in the [feature branch](https://github.com/bseiller/dfhack/blob/c9fc554f9362cc8163ebfa3a8c2ec47309f317bc/plugins/embark-assistant/survey.cpp#L1512) with the big rewrite, but as I wrote in the commit message I actually did performance comparisons between `.at()` , `[]` and `std::fill`, even looking at `std::vector.assign`.
And shockingly enough and counterintuitively `std:fill` takes more than twice as long to process the data than `[]`, `std::vector.assign` around 3 times more, at least on Windows and for C++11 (or is it C++14?) which we use for dfhack.
Of course this is specific to my machine, but for 16.908.544 tiles (large world) and 256 inorganics (assignments to the vector where applicable ) the different variants perform as follows:
- `[]` takes around 8 seconds. 
- `.at` takes around 9 seconds. 
- `std::fill` takes around 17 seconds. 
- `std::vector.assign` takes between 23 and 24 seconds.
- `std::vector.resize` (current solution previous to this PR) takes about as long as `std::vector.assign`.
- the hack as described at [[3]](https://stackoverflow.com/a/64075271/167865) takes 0,5 seconds, but it very convoluted...

And that is in standalone code, mind you.
The reason for this behavior concerning  `std::fill` probably is that std::vector<bool> has been neglected in MS C++/MSVC, missing the needed template specialization for the template functions such as `std::fill` => [[1]](https://isocpp.org/blog/2012/11/on-vectorbool) + [[2]](https://stackoverflow.com/questions/12433154/are-stdfill-stdcopy-specialized-for-stdvectorbool)

The use and performance of `std::vector<bool>` in context of the inorganics has been on my mind again and again during my work at embark_assistant. I thought about replacing it with lists that only contain the indices/ids of the found inorganics, even about a custom data structure, that is both fast and has a small memory footprint just like `std::vector<bool>` but doesn't need to be iterated over completely for access and reset. I event looked into providing the needed template specialization, but I couldn't find a good starting point for that.
I haven't found the "perfect" solution so far for this, one that does not include a lot of rewriting.

If you still favor the use of `std::fill`, I'll change it. Happily so if I knew that the performance would improve when migrating to a more recent version of the C++ Standard in the future.
Also I wonder if `std::vector<bool>` and `std::fill` perform better on *nix with GCC or Clang - do you have any insight in this? 
Alternatively to using `std::fill` I could call `size()` for every vector and use `.at` instead of `[]`, as the vectors are initialized once with `embark_assist::survey::initiate` and meant to have all the same size - which of course isn't guaranteed as in the future someone could call `embark_assist::survey::survey_mid_level_tile` with uninitialized vectors. But in that case there would be an error further down in `embark_assist::survey::survey_mid_level_tile` anyway even if the use `std::fill` prevents that initially.

links for reference:
[1] https://isocpp.org/blog/2012/11/on-vectorbool
[2] https://stackoverflow.com/questions/12433154/are-stdfill-stdcopy-specialized-for-stdvectorbool
[3] https://stackoverflow.com/a/64075271/167865

plugins/embark-assistant/matcher.cpp
bseiller: That is a good idea, something like 
```
void embark_assist::matcher::shutdown() {
    if (state != nullptr)  {
        delete state;
        state = nullptr;
    }
}
```
Will do.

PatrikLundell: You can make a C array of uint8_t (number of elements / 8 rounded up), iterate by 64 (bits) and assign 0 (uint64_t or int64_t) to the address of the first, 8:th, ... byte through casting until don't have 8 bytes left, at which time you can iterate over bytes (or try a quad byte, a double byte, and a byte to squeeze absolutely everything out of it). That's definitely not pretty, though, so I wouldn't do it for the few seconds saved, but instead keep the "original" code (i.e. the one in this PR) and ditch std::fill.

plugins/embark-assistant/survey.cpp
lethosor: I haven't actually done any of my own profiling on it. I did a bit of research into it yesterday, and it looked like `[]`, `assign()`, and `fill()` had nearly identical performance under GCC with `-O3` (what our release builds use). There's a chance that it was for more "normal" vectors, i.e. not `vector<bool>`. In any case, you've clearly done more testing here, so I'm fine going with the current implementation.

bseiller: As I was curious myself I did a simple comparison with `std::vector<char>` not changing anything else in the benchmark setup:
- `[]` takes around 4 to 5 seconds
- `std::fill` takes around 0,4 to 0,6 seconds as does `std::fill_n`
- `std::vector.assign` takes around 0,7 seconds
- `std::memset` takes around 0,5 seconds - this one can't be used for `std::vector<bool>` at all as it doesn't change the 
So `std::vector<bool>` is really a very special case: highly optimized in regards to the memory footprint but with a significant overhead for bulk manipulation, at least in my version of MSVC.
The introduction and the notes at the end [here](https://en.cppreference.com/w/cpp/container/vector_bool) convey the idea of how special `std::vector<bool>` pretty well.

And just to make sure `std::vector<char>` isn't a special case itself as well I also checked with `std::vector<int>`:
- `[]` takes around 6 seconds
- `std::fill` takes around 4 seconds
- `std::fill_n` takes around 2 seconds
- `std::vector.assign` takes around 3 seconds
- `std::memset` takes around 1 second
So `std::vector<char>` might be another special case with tailored template specializations or there might just be CPU instructions that can process chars faster than ints, but the bulk operations (`std::fill`, `std::fill_n` and `std::memset`) are faster for `std::vector<int>` than looped `[]` nonetheless. 

By now I'm thinking that the easiest and cleanest solution for this might be to replace the 3 `std::vector<bool>` in [mid_level_tile](https://github.com/DFHack/dfhack/blob/60c6c2a05976edfab329ff3e2b0d3b33f7476432/plugins/embark-assistant/defs.h#L86) with `std::vector<char>` and use `std::fill` indead.
As PR #1757 reduced the pressure to keep the storage of inorganics occurrence as memory efficient as possible and there are only 2 instances of https://github.com/DFHack/dfhack/blob/60c6c2a05976edfab329ff3e2b0d3b33f7476432/plugins/embark-assistant/defs.h#L99 and thus 512 instances of mid_level_tile the additional memory consumption would be about 336kiB, assuming there are 256 non-generated inorganics. I think that is acceptable considering it will save another 7 seconds of search time.
@lethosor, @PatrikLundell: What do the both of you think?
I'll look into it, but that is definitely a new PR.

plugins/embark-assistant/matcher.cpp
bseiller: @PatrikLundell: That is pretty much, was the hack I mentioned does as well - by manipulating the underlying data structure directly it is able to do a bulk reset. But in both cases the resulting code is not pretty... :)
But have a look at the numbers I generated in the other comment above. I thought about changing vector<bool> to another type but the memory cost would have been significant - but that is no longer true.

plugins/embark-assistant/survey.cpp
PatrikLundell: I agree. Using a vector over char rather than bool isn't butt ugly, saves a bit of time, and costs insignificant amounts of memory if there are only two copies of the set. I'd make sure to make a comment to explain the rationale for usage of char rather than bool, though.

plugins/embark-assistant/matcher.cpp
PatrikLundell: I didn't read the hack in detail, but it looked like it was using bytes rather than octo-bytes (and I think it would have to if a vector is used rather than an array, as I think a vector isn't guaranteed to be contiguous).

bseiller: Ah I see what you mean now - yes - it operates on bytes not on octo-bytes - but the idea is the same, which is what I meant. Actually I think the hack is absolutely not guaranteed to be portable either (though it works at least with two versions of MSVC) so the contiguousness of the data is probably the least problem if one tries to use this in a code base like dfhack :D

PatrikLundell: Yes, making assumptions of what the internal structure of the vector implementation is would definitely enter risky territory, which is another reason for using an array rather than vector, as I think the array has a precise definition. However, this is an academic discussion at this point.

lethosor: For reference: as of C++03, vector data is required to be stored contiguously, except for `vector<bool>`.

lethosor (approved):

plugins/embark-assistant/survey.cpp
lethosor: Makes sense to me too, although I'd suggest `vector<int8_t>` or `vector<uint8_t>` for consistency with existing DF/DFHack code.

bseiller: Thanks for pointing out that convention considering `vector<int8_t>`/`vector<uint8_t>` - I haven't spent much time looking into other plugins or the DFHack core itself, so getting this kind of hint is much appreciated & I'll heed this advice.

PR #1774, opened 2021-02-08
improved pre embark mineral estimates

### LINKED ISSUE #1772: Prospect pre-embark Blue Diamond overestimation
In my most recent world several/many embark locations give greatly overestimated predictions for blue diamonds using the prospect plugin pre-embark.  Once embarked accurate results are reported.

I saw this using df47.05 with dfhack-0.47.05-alpha0-210131001-Linux-64bit-gcc-7.

[Pre-embark savegame](https://dffd.bay12games.com/file.php?id=15400)

The issue is present in a number of embarks across the world, although many/most embarks do not have the issue.  A selection of locations to see/reproduce the issue in the above save game follows.  The numbers predicted are reproduced for convenience.

If top left of map is (1,1) and embark rectangle is default 4x4 then 
(1,25) with embark rectangle bottom right corner   (diamond_blue 894  diamond_fy 128)
(65,41) with embark rectangle top left corner        (diamond_blue 1287 diamond_fy 307)
(14,35) with embark rectangle centred                  (diamond_blue 704  diamond_fy 128)

The save (world generation) was done with an entirely vanilla linux dwarf fortress (running up to date archlinux) with the above version of dfhack entirely unmodified (even using the example init).  It was made, however, with a custom advanced worldgen setting which is attached.  (Attempted to reproduce with a standard worldgen but my patience was tried - a couple of hours and I did not find any predicted single occurence diamonds at all.)  [The specific seeds are a result of reproducing the save in vanilla df since I originally noticed the issue while using CLA graphics/tileset which did modify a few raws.]

[world_gen.txt](https://github.com/DFHack/dfhack/files/5920799/world_gen.txt)

This was originally reported [on the bay12 forums here](http://www.bay12forums.com/smf/index.php?topic=164123.msg8243749#msg8243749) (and following) which is a better place to get hold of me if more information is needed.  (The figures are different from those reported there since I was originally using a 3x3 embark rectangle.)

PatrikLundell:
Fix #1772 
It can be noted that I've changed the somewhat arbitrary "average" amounts a little from what I've written in the issue.

Apart from adjusting the calculations, it turned out to be necessary to adjust their summing up, as the previous code applied integer truncation to the parts, resulting in e.g. a faint yellow diamond estimate drop from 128 to 0 when the factor was changed from 42 to 16.

As the comments added indicate, there are more intricacies that can be added, but with doubtful benefit. If someone would perform an analysis to determine more precise "average" cluster, vein, and small cluster sizes it would probably have a greater impact.

The main objective, i.e. to reduce cluster one estimates from 10 times the number of the small cluster material tiles it was enclosed within seems to have been reached, with an estimate of 10 vs an actual number of 9 at the first issue embark indicated in the issue.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -33,6 +33,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 # Future
 
+## Fixes
+- `prospector`: improved pre embark rough estimates
+
 # 0.47.04-r5
 
 ## Fixes
--- a/plugins/prospector.cpp
+++ b/plugins/prospector.cpp
@@ -49,7 +49,7 @@ struct matdata
     const static int invalid_z = -30000;
     matdata()
     {
-        count = 0;
+        count = 0.0;
         lower_z = invalid_z;
         upper_z = invalid_z;
     }
@@ -59,7 +59,7 @@ struct matdata
         lower_z = copyme.lower_z;
         upper_z = copyme.upper_z;
     }
-    unsigned int add( int z_level = invalid_z, int delta = 1 )
+    float add(int z_level = invalid_z, float delta = 1.0)
     {
         count += delta;
         if(z_level != invalid_z)
@@ -75,7 +75,7 @@ struct matdata
         }
         return count;
     }
-    unsigned int count;
+    float count;
     int lower_z;
     int upper_z;
 };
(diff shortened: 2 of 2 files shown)


lethosor (approved):
Looks good! I tested with a couple embarks, and most of the pre-embark gem counts seem closer to the post-embark counts now (admittedly a subjective assessment, and there were some that were still off, but that's fine, and I didn't see anything as bad as described in #1772)

quietust:
Just ran into a rather unfortunate side effect of this change: in suitably large (i.e. tall) embarks, the STONE and MINERAL counts in the "Base materials" section can be so large that they now display in **scientific notation**.

PR #1777, opened 2021-02-13
Upgrade Ruby

### LINKED ISSUE #1718: Upgrade Ruby to a consistent version on all platforms
**Update:** See #2081. Our current plan is to remove Ruby support instead.

Different platforms use different precompiled versions of libruby, currently (as of 0.47.04-r3):
* 2.0.0 on win64
* 1.8.7 on win32, osx32, and all Linux
* Whatever is provided by the system on osx64

These were manually built and are somewhat tricky to rebuild (particularly since they need to be built externally), so they are mostly outdated. Changes across different versions can lead to compatibility issues across platforms that users don't expect, like #1613 and https://github.com/DFHack/scripts/pull/227#discussion_r528072627. Update: our current solution also doesn't work on macOS 11, per #1742

Coming up with a CI pipeline to build Ruby with the correct toolchains/flags for DFHack could be a good long-term solution here.

lethosor:
This is intended to eventually fix #1718 - see issue for details. Currently it just updates the `ruby` plugin to work with Ruby 2.7. Builds of Ruby 2.7 are available at https://github.com/lethosor/ruby-build/actions and are not yet downloaded/installed on this branch.

Opening a PR primarily to trigger CI builds for other platforms.

### DIFF
--- a/plugins/ruby/ruby.cpp
+++ b/plugins/ruby/ruby.cpp
@@ -64,10 +64,11 @@ DFhackCExport command_result plugin_init ( color_ostream &out, std::vector <Plug
 {
     onupdate_active = 0;
 
-    // fail silently instead of spamming the console with 'failed to initialize'
-    // if libruby is not present, the error is still logged in stderr.log
     if (!df_loadruby())
-        return CR_OK;
+    {
+        out.printerr("df_loadruby failed - see stderr.log for details\n");
+        return CR_FAILURE;
+    }
 
     // the ruby thread sleeps trying to lock this
     // when it gets it, it runs according to r_type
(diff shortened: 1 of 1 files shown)


ab9rf:
Let's not forget that there was a weird issue with 64-bit Ruby on Windows (see #1041). I haven't looked, but we should make sure whatever we do continues to work on Windows, given the hacky way I fixed this back then.

lethosor:
The plugin still loads (well, without my ruby_sysinit changes) so I think that part is fine, but still a good thing to check.

myk002:
Ruby has been turned down. closing old PRs and bugs

PR #1780, opened 2021-02-25
[dreamfort] separate sand bags into their own stockpile

myk002:
since quantum stockpiling them dumps out their contents

last change before 0.47.05-r1. i swear.

### DIFF
--- a/data/quickfort/aliases-common.txt
+++ b/data/quickfort/aliases-common.txt
@@ -158,6 +158,14 @@ bags:    {furnitureprefix}de{Right}f{Right}{Up 10}&{Left}{Down}f{Down}f{Down}f{R
 buckets: {furnitureprefix}de{Right}f{Right}{Up 12}&^
 sand:    {furnitureprefix}de{Right}f{Right}{Up}&^
 
+forbidpots:    {furnitureprefix}{Right 2}{Up 5}&^
+forbidbuckets: {furnitureprefix}{Right 2}{Up 12}&^
+forbidsand:    {furnitureprefix}{Right 2}{Up}&^
+
+permitpots:    {forbidpots}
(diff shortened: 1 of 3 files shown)


data/quickfort/aliases-common.txt
```diff
@@ -158,6 +158,14 @@ bags:    {furnitureprefix}de{Right}f{Right}{Up 10}&{Left}{Down}f{Down}f{Down}f{R
 buckets: {furnitureprefix}de{Right}f{Right}{Up 12}&^
 sand:    {furnitureprefix}de{Right}f{Right}{Up}&^
 
+forbidpots:    {furnitureprefix}{Right 2}{Up 5}&^
+forbidbuckets: {furnitureprefix}{Right 2}{Up 12}&^
+forbidsand:    {furnitureprefix}{Right 2}{Up}&^
+
+permitpots:    {forbidpots}
+permitbuckets: {forbidbuckets}
+permitsand:    {forbidsand}
```
lethosor, line 167: Are these intentionally the same as the `forbid` aliases?

myk002: yeah, all permit/forbid pairs that just toggle an item in the third column are identical keystroke-wise. I just provide both so they can be used with semantically appropriate names. e.g.:

`{pots}{permitbuckets}` adds on buckets because it was disabled by `pots`, but `{forbidbuckets}` by itself would toggle a by-default enabled bucket item to off.

lethosor: Interesting. Too bad there isn't a way to consistently enable/disable them from an unknown initial state, but I think this makes sense the way it is.

lethosor (approved):

Issue #1690, opened 2020-10-30
Automated tests for quickfort and its dependencies


myk002:
Long term, what I’m envisioning is running quickfort tests as part of the automated regression suite so that the tests prevent changes from being merged if they break any existing functionality.
Short term, I just want an easy way to discover bugs and unhandled edge cases in quickfort and its dependencies, and to make sure *I* don’t break anything as I make changes in quickfort..

So I’m thinking this project can be broken down to two phases:
1. Infrastructure and code for manually-initiated locally-run quickfort regression tests
1. Integrate quickfort tests into DFHack’s automated regression testing

In the first phase, I’d like to:
- Enable fast iteration during dev (must at least run on Linux) by running tests interactively on an already-loaded map
- Collocate test files with code files (same directory, or at least same repo) so tests can be reviewed and merged with the code that they test
- Provide Lua code coverage visualization
- Achieve at least 75% code coverage for quickfort

And in the second phase, I’d like to:
- Gate code submission (must show a pass on GitHub before PR can be merged)
- Run on all platforms

## Questions to answer for phase 1
- What should the interface be for the developer to run the tests? What manual setup should be required?
- Can we easily use the existing test harness in DFHack/dfhack to run the tests? If not, what should we use?
- How do we get coverage reports?
- How do we structure the tests so that the code is clean but local functions can be accessed for testing?
- Is it ok to add (tens of?) thousands of lines of test code to DFHack/scripts? Can/should they be included in the distribution? What about test blueprints?
- How can the work we do in this phase positively impact the rest of DFHack? Can we make writing and running tests easier for contributing developers?

Here are my initial thoughts (as a starting point for discussion):

### Interface

The primary use case in this phase will be me writing and executing tests while actively developing for a PR. Quick edit-save-run cycles. It should take no more than a few seconds to run tests, and it should be really easy to rerun the most recently-run test suite.

The developer would be required to load up a fort within which the tests will run. Test set-up functions can forbid all existing items and call tiletypes and createitem to normalize the environment.

The test command, including filters and test output parameters, would be run from the DFHack# prompt. I suspect it would be best to implement the command within quickfort for now: `quickfort test …` while we figure out the test command feature set. If we want to make this into a framework for all scripts, we can move the logic into a common location later.

### Test harness selection

The test harness in DFHack/dfhack/test/ is not an obvious fit for the ad-hoc interactive test running of phase 1. It has the following shortcomings (as seen from the phase 1 use case):
[...]

PR #1787, opened 2021-03-06
[luacov] add luacov as a DFHack dependency and install it into hack/lua/luacov

myk002:
#1690

This is just the pure Lua implementation of luacov. I'll investigate possibly integrating with cluacov later. cluacov replaces the debug hook and "improves accuracy", but in my tests I've been perfectly happy with the accuracy of the lua implementation of the debug hook.

I installed it in `hack/lua/luacov` as a generally accessible library even though luacov will not be accessed directly by the majority of scripts. It still needs to be accessible to the luacov report generation script (`scripts/devel/luacov`) and to the C++ code, which calls `luacov.runner.init()` and manages installing the luacov debug hook in coprocesses.

I fixed the luacov repo to the recent 0.15.0 release. This release is mistakenly named '0.14.0' internally, but this is cosmetic.

### DIFF
--- a/depends/CMakeLists.txt
+++ b/depends/CMakeLists.txt
@@ -22,6 +22,8 @@ option(CLSOCKET_DEP_ONLY "Build for use inside other CMake projects as dependenc
 add_subdirectory(clsocket)
 ide_folder(clsocket "Depends")
 
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/luacov/src/luacov/ DESTINATION ${DFHACK_DATA_DESTINATION}/lua/luacov)
+
 # set the default values of libexpat options - the descriptions are left empty
 # because later option() calls *do* override those
 set(EXPAT_BUILD_EXAMPLES OFF CACHE BOOL "")
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,4 @@
 [submodule "depends/luacov"]
 	path = depends/luacov
 	url = ../../keplerproject/luacov.git
+	shallow = true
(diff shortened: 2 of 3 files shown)


depends/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ option(CLSOCKET_DEP_ONLY "Build for use inside other CMake projects as dependenc
 add_subdirectory(clsocket)
 ide_folder(clsocket "Depends")
 
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/luacov/src/luacov/ DESTINATION ${DFHACK_DATA_DESTINATION}/lua/luacov)
```
lethosor, line 25: I'll add a check to make sure the submodule exists with the other checks, since that can trip up people with existing clones sometimes.

lethosor (approved):
Looks like it installs in the right place, and as a plus, https://github.com/keplerproject/luacov/blob/v0.15.0/src/luacov/runner.lua#L9 (which this uses) has the right version number too.

lethosor:
Ran into an issue merging this: I have `push.recurseSubmodules` set to `check` for safety, and this breaks because the commit we have pinned is only referenced by a tag, not a branch upstream. (It also usually breaks on shallow clones, but I'm fine "un-shallowing" submodules on my end.)

I'm thinking of forking luacov just for the purpose of creating a `dfhack` branch, similar to other dependencies like libzip, but not introducing changes otherwise. (We haven't changed anything in https://github.com/dfhack/jsoncpp either, as far as I know.) I'll revisit tomorrow.

PR #1788, opened 2021-03-06
[WIP][confirm] use the standard lua interpreter context in the confirm plugin

myk002:
Since double-calling `Lua::Open()` causes some coverage stats to go to the wrong debug hook.

`Lua::Open()` is also called in `plugins/devel/memutils.cpp`, but there is no Lua accessed from that plugin (and also every time I touch that file I make things crash).


lethosor:
Perhaps #1789 would be a better place to discuss the details about `Lua::Open`, actually (I didn't take a very thorough pass through everything first)

myk002:
rescinding PR -- fixed the issue in #1789

PR #1789, opened 2021-03-06
[luacov] enable luacov coverage monitoring if DFHACK_ENABLE_LUACOV is set

myk002:
#1690

Use the DFHACK_ENABLE_LUACOV environment var to control whether we initialize and patch luacov in to Lua threads owned by the DFHack core context.

This PR depends on DFHack/scripts#261 for the `devel/luacov` doc header reference (the docs added by this PR will fail until that PR is merged).

On initialization, we load `luacov.runner` and initialize it directly from C++. We also need to ensure the debug hook is installed correctly for coroutine threads. Since creating coroutines in C++ bypasses the [monkey patching luacov does](https://github.com/keplerproject/luacov/blob/master/src/luacov/runner.lua#L449) to inject itself into coroutine contexts, we compensate by reimplementing the monkey patch in our code. We use our own little lua helper function to do this because luacov's similar `with_luacov()` function calls `has_hook_per_thread()`, which would cause an infinite loop when it creates its test coroutine to figure out whether it needs to install its wrapper.

If multiple instances of luacov were to be instantiated, we would be in one of two problematic situations:
1 the `tick` option is not specified in `.luacov` (this is the default). then the coverage metrics gathered by the runner that is not in the same context as the scripts/devel/luacov report generation script are effectively inaccessible, and therefore useless.
2 the `tick` option is specified in a customer `.luacov` and the two runners conflict when they attempt to write to the same output files.

To address this, we only initialize luacov for the core context. Lua files accessed from other contexts will not be included in the coverage reports. If this is a problem, we can implement the proper plumbing to differentiate the configurations among the active luacov runners and make them all accessible for report generation. A dev could also hack it into working by directly calling `devel/luacov` directly from the script in the appropriate context.

### DIFF
--- a/library/LuaTools.cpp
+++ b/library/LuaTools.cpp
@@ -554,6 +554,17 @@ static int dfhack_error(lua_State *L)
     return lua_error(L);
 }
 
+// replaces os.exit with a thrown exception
+static int dfhack_exit_error(lua_State *L)
+{
+    int exit_code = luaL_checkint(L, 1);
+    lua_pop(L, 1);
+    lua_pushfstring(L,
+                    "prevented execution of os.exit(%d); please use error()"
+                    " or qerror() instead.", exit_code);
+    return dfhack_error(L);
+}
+
 static int dfhack_exception_tostring(lua_State *L)
 {
     luaL_checktype(L, 1, LUA_TTABLE);
@@ -775,7 +786,22 @@ static int dfhack_cowrap (lua_State *L) {
     return 1;
 }
 
+static void add_luacov_coroutine_wrapper(lua_State *L)
+{
+    color_ostream *out = Lua::GetOutput(L);
+    if (!Lua::PushModulePublic(*out, L, "luacov_helper", "with_luacov"))
+    {
+        out->printerr("Failed to load luacov_helper.with_luacov.\n");
+        return;
+    }
+    lua_swap(L);
+    if (!Lua::SafeCall(*out, L, 1, 1))
+        out->printerr("Failed to set luacov monitoring wrapper.\n");
+}
+
 lua_State *DFHack::Lua::NewCoroutine(lua_State *L) {
+    if (getenv("DFHACK_ENABLE_LUACOV"))
+        add_luacov_coroutine_wrapper(L);
     lua_State *NL = lua_newthread(L);
     lua_swap(L);
     lua_xmove(L, NL, 1); /* move function from L to NL */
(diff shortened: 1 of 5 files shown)


lethosor:
I'd like to take a look at the coroutine/context issues too. What's a good way of checking whether luacov is "working"? Is `debug.gethook()` enough? From #1788, if only "some" coverage stats are being recorded incorrectly, I'm not sure if simply checking the hook once is good enough. Essentially, how have you been determining that there's an issue?

myk002:
checking for the debug hook from within lua in the context of the target script should be enough to ensure that luacov is working, but I usually verified by generating a report and verifying that I got expected data for the script.

I also instrumented luacov's debug hook to print out what runner stats were being added to for each file to get a better understanding of how the contexts interacted.

I found the contexts are kept properly separated -- the luacov initialized in confirm's state tracked usage for all files loaded by that state and the core context tracked everything else (with some overlap for files loaded by both states, like dfhack.lua). Contrary to what I thought before, the setting of the interrupt hook in the confirm state does *not* affect the luacov-set debug hook in the core context. Everything would be fine, but luacov can be configured to automatically flush stats to `luacov.stats.out` and two simultaneous flushers would overwrite each other's output.

Your suggestion to only initialize luacov in the core context, however, works just fine. We won't get coverage stats for confirm.lua (or any other plugin that creates its own Lua state), but that's not so bad. If we really want to cover them, we can init luacov in non-core contexts and provide a mechanism for them to read separate config files so they don't try to write to the same output files.

With this latest commit (which I will push momentarily), I'm happy with how luacov is behaving.

myk002:
Significantly updated the PR comment with current behavior and rationale

library/LuaTools.cpp
```diff
@@ -554,6 +554,17 @@ static int dfhack_error(lua_State *L)
     return lua_error(L);
 }
 
+// replaces os.exit with a thrown exception
+static int dfhack_exit_error(lua_State *L)
```
lethosor, line 558: Was disabling `os.exit()` needed to debug a luacov issue? I'm totally fine with keeping this - nobody should be using it anyway.

lethosor (approved):

library/LuaTools.cpp
myk002: I should have explained that, sorry. Unmodified luacov can call os.exit() on configuration error. It won't cause df to exit, but it will bork the dfhack console. By default, this should never happen, but if users/developers create a .luacov file with errors, it can come up.

I also thought it was a good general protection against unruly scripts calling os.exit(), which, as you said, would never make sense for DFHack anyway.

PR #1791, opened 2021-03-06
[test] test/main.lua enhancements

myk002:
#1690

new test/main API is backwards compatible with existing usage.

- allow configuration set in test_config.json to be dynamically overridden on the commandline
- make `expect.table_eq()` diff recursively (all existing tests that use it still pass)
- fix implementation of testing for multiple test name filters. the original implementation filtered out too many tests if more than one filter was specified.
- add ability to filter by mode (title, none, etc.)
- add ability to rerun tests that have already been run (I'm actually not sure what the purpose of `test_status.json` is. why would we want to skip a test if it has already been run? If you're running the tests again, haven't you just edited the code and wouldn't you want to see if tests that previously passed are now failing?)
- done command can now be passed as space-separated tokens or as one quoted string
- added output text that says what your test dir is and what your filters are
- added tests for new table_eq implementation

### DIFF
--- a/test/main.lua
+++ b/test/main.lua
@@ -2,8 +2,41 @@ local json = require 'json'
 local script = require 'gui.script'
 local utils = require 'utils'
 
-local args = {...}
-local done_command = args[1]
+local help_text =
+[[
+Run DFHack tests.
+
+Usage:
+    test/main [<options>] <post-test command>
+
+Options:
+    -h, --help      display this help message and exit.
+    -n, --nocache   don't skip tests marked as completed in test_status.json.
+    -m, --modes     only run tests in the given comma separated list of modes.
+                    valid modes are 'none' and 'title'. if not specified, no
+                    modes are filtered.
+    -t, --tests     only run tests that match one of the comma separated list of
+                    patterns. if not specified, no tests are filtered.
+
+Examples:
+    test/main                 runs all tests that haven't been run before
+    test/main -n              reruns all tests
+    test/main -nm none        reruns tests that don't need the game to be in a
+                              specific mode
+    test/main -nt quickfort   reruns quickfort tests
+]]
+
+local help, nocache, mode_filter, test_filter = false, false, {}, {}
+local done_command = utils.processArgsGetopt({...}, {
+        {'h', 'help', handler=function() help = true end},
+        {'n', 'nocache', handler=function() nocache = true end},
+        {'m', 'modes', hasArg=true,
+         handler=function(arg) mode_filter = arg:split(',') end},
+        {'t', 'tests', hasArg=true,
+         handler=function(arg) test_filter = arg:split(',') end},
+    })
+
+if help then print(help_text) return end
 
 local CONFIG_FILE = 'test_config.json'
 local STATUS_FILE = 'test_status.json'
@@ -152,6 +185,10 @@ function load_test_config(config_file)
         error('Invalid test folder: ' .. config.test_dir)
     end
 
+    -- override config with any params specified on the commandline
+    if #mode_filter > 0 then config.modes = mode_filter end
+    if #test_filter > 0 then config.tests = test_filter end
+
     return config
 end
 
@@ -207,7 +244,7 @@ function get_test_files(test_dir)
 end
 
 function load_test_status()
-    if dfhack.filesystem.isfile(STATUS_FILE) then
+    if not nocache and dfhack.filesystem.isfile(STATUS_FILE) then
         return json.decode_file(STATUS_FILE)
     end
 end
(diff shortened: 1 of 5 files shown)


test/main.lua
```diff
@@ -162,6 +245,7 @@ function build_test_env()
             mode = 'none',
         },
         expect = {},
+        expect_raw = expect,
```
lethosor, line 248: I would suggest importing `test/main.lua` from the test instead, since `expect` is a global already. It would need to be [tweaked slightly](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#importing-scripts) to only call `main()` when invoked as a command, but I think it should work when that's done.

myk002: good idea. I rewrote main.lua to be importable.

BenLubar:
If you're going to pass the done_command as an array, you need to also update https://github.com/DFHack/dfhack/blob/develop/travis/run-tests.py#L76 to generate it as two arguments rather than one argument with a space

myk002:
done. thanks!

test/test_test.lua
```diff
@@ -0,0 +1,36 @@
+local expect_raw = reqscript('test/main').expect
```
lethosor, line 1: > rename test.lua to test_test.lua since I plan to rename main.lua to test.lua in a future PR

I think we should reconsider where the test runner lives in the future, in that case. I didn't really think through where the runner should live, but it shouldn't be in the same namespace as the tests themselves. Essentially, my proposal is to:
* move test/main.lua to somewhere else in the DFHack repo and install it to hack/scripts if BUILD_TESTS is enabled. (Now that I think about it, it probably lives at `test/main` because I couldn't think of another place to put it in the DFHack repo, but maybe the `ci` folder could work.)
* don't add the `test` folder to the script path, since none of the test cases themselves are intended to be run directly (and they wouldn't work if they were). This would allow tests to have the same names as other scripts.

This is something I can handle if it's not too disruptive to your test work. Just wanted to make sure our plans were aligned - feel free to suggest something else.

myk002: that sounds good to me. Changes to the test runner don't conflict with anything I have in the pipeline. I'm just writing unit tests ATM.

Now that we're going to install the test runner directly into the `scripts/` directory, I do suggest still renaming `main.lua` to `test.lua`. Running tests with `test/main` made sense. Running tests with just `main` isn't quite as intuitive.

lethosor: Oh, definitely - forgot to mention it, but I was thinking the same thing.

lethosor:
> * add ability to rerun tests that have already been run (I'm actually not sure what the purpose of `test_status.json` is. why would we want to skip a test if it has already been run? If you're running the tests again, haven't you just edited the code and wouldn't you want to see if tests that previously passed are now failing?)

I missed this earlier. The purpose is to allow DF to be restarted and resume tests where they left off if a single test causes DF to crash. This is somewhat more likely in our case than in other projects because we have less control over the environment that tests are running in.

test/main.lua
```diff
@@ -292,55 +379,75 @@ function run_test(test, status, counts)
...
+
+    local status = {}
+    if not config.nocache then
+        status = load_test_status() or status
+        for i = #tests, 1, -1 do
+            local test = tests[i]
+            if not status[test.full_name] then
+                status[test.full_name] = TestStatus.PENDING
+            elseif status[test.full_name] ~= TestStatus.PENDING then
+                print(('skipping test: %s: state = %s)'):format(
```
lethosor, line 438: Heh, thanks for preserving the `)`, but that's a typo from e6c9db0a18a49ab3777d904151a40bb1d660a398 :)

test/main.lua
```diff
@@ -354,6 +461,55 @@ function main()
...
+        })
+
+    if help then print(help_text) return end
+
+    local done_command = table.concat(other_args, ' ')
+    local config = load_test_config(CONFIG_FILE)
+
+    -- override config with any params specified on the commandline
+    if test_dir then config.test_dir = test_dir end
+    if nocache then config.nocache = true end
```
lethosor, line 485: Honestly, I might end up inverting this. CI always runs tests through `run-tests.py`, which could pass a `--cache` flag to `test/main.lua` fairly easily and wouldn't need CI changes.

lethosor (approved):

test/main.lua
myk002: haha, I noticed, but I thought it was just stylized formatting -- it actually looks reasonable in the output

test/main.lua
myk002: that works. maybe rename `-n, --nocache` to `-r, --resume`. Want me to do that?

lethosor: Makes sense to me. Feel free

PR #1792, opened 2021-03-06
[test] exempt lua test files from docstring requirements

myk002:
#1690 

exempt lua test files from docstring requirements and don't display them with 'ls'

this prevents test files from cluttering up the ls output. users can't run them directly anyway, and as far as I can see, it doesn't seem that seeing the list of test files in ls output is valuable.

this also removes test/main from ls output. is that ok?

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 - Added more client library implementations to the `remote interface docs <remote-client-libs>`
 
+## Internals
+- Lua test files are no longer required to start with standard script documentation (since they aren't runnable scripts)
+
 # 0.47.05-r1
 
 ## Fixes
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -317,7 +317,7 @@ static void listScripts(PluginManager *plug_mgr, std::map<string,string> &pset, 
                 pset[key] = help;
             }
         }
-        else if (all && !files[i].empty() && files[i][0] != '.' && files[i] != "internal")
+        else if (all && !files[i].empty() && files[i][0] != '.' && files[i] != "internal" && files[i] != "test")
         {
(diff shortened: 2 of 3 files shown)


lethosor:
Maybe we should move `test/main` outside of the test folder, since it's not a test. (I didn't really think this through when I put it there, apparently - it already has an exception to avoid running itself...)

myk002:
I'll do this in a separate PR so I don't generate merge conflicts with the current set, but how about this:
- rename test/main.lua to test/test.lua (the existing test.lua is renamed to test_test.lua in #1791)
- if BUILD_TESTS then install test/ to hack/scripts/test/ and exclude test.lua
- if BUILD_TESTS then install test/test.lua to hack/scripts/test.lua
- move `test/core.lua`, `test/encoding.lua`, and `test/gui.lua` into `test/core/` to keep the top `test/` dir clean

myk002:
I'm not sure if the CI failure is due to this PR or not -- script not found, but only on latest GCC on Linux 64? It could be related, but I'm not clear how.

lethosor:
Almost certainly unrelated, although it's a bit hard for me to re-trigger now.

docs/changelog.txt
```diff
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 - Added more client library implementations to the `remote interface docs <remote-client-libs>`
 
+## Internals
+- Lua test files are no longer required to start with standard script documentation (since they aren't runnable scripts)
```
lethosor, line 40: The problematic files didn't exist before, so this doesn't need to be mentioned. (Any tests in the DFHack repo were exempt from this, and they were the only ones that existed previously.)

lethosor (approved):
Looks similar enough to #1612 that I'm comfortable getting this in quickly to fix CI on https://github.com/DFHack/scripts/pull/263. I'll address my other comment locally.

PR #1796, opened 2021-03-08
[xlsxreader] add Lua class wrappers

myk002:
makes the API easier to use for Lua clients

this change also allows the API to limit the number of tokens returned from get_row

The new class API will be used in DFHack/scripts#263

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -4307,12 +4307,14 @@ xlsxreader
 Utility functions to facilitate reading .xlsx spreadsheets. It provides the
 following low-level API methods:
 
-- ``open_xlsx_file(filename)`` returns an open file_handle
-- ``close_xlsx_file(file_handle)``
+- ``open_xlsx_file(filename)`` returns a file_handle or nil on error
+- ``close_xlsx_file(file_handle)`` closes the specified file_handle
 - ``list_sheets(file_handle)`` returns a list of strings representing sheet
   names
-- ``open_sheet(file_handle, sheet_name)`` returns a sheet_handle
-- ``close_sheet(sheet_handle)``
+- ``open_sheet(file_handle, sheet_name)`` returns a sheet_handle. This call
+  always succeeds, even if the sheet doesn't exist. Non-existent sheets will
+  have no data, though.
+- ``close_sheet(sheet_handle)`` closes the specified sheet_handle
 - ``get_row(sheet_handle, max_tokens)`` returns a list of strings representing
   the contents of the cells in the next row. The ``max_tokens`` parameter is
   optional. If set to a number > 0, it limits the number of cells read and
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -4307,17 +4307,40 @@ xlsxreader
 Utility functions to facilitate reading .xlsx spreadsheets. It provides the
 following low-level API methods:
 
- - ``file_handle open_xlsx_file(filename)``
+ - ``open_xlsx_file(filename)`` returns a file_handle
  - ``close_xlsx_file(file_handle)``
- - ``sheet_names list_sheets(file_handle)``
- - ``sheet_handle open_sheet(file_handle, sheet_name)``
+ - ``list_sheets(file_handle)`` returns a list of strings representing sheet
+  names
+ - ``open_sheet(file_handle, sheet_name)`` returns a sheet_handle
  - ``close_sheet(sheet_handle)``
- - ``cell_strings get_row(sheet_handle, max_tokens)`` The ``max_tokens``
-   parameter is optional. If set to a number > 0, it limits the number of cells
-   read and returned for the row.
+ - ``get_row(sheet_handle, max_tokens)`` returns a list of strings representing
+  the contents of the cells in the next row. The ``max_tokens`` parameter is
+  optional. If set to a number > 0, it limits the number of cells read and
+  returned for the row.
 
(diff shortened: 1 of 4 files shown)


plugins/lua/xlsxreader.lua
```diff
@@ -1,47 +1,55 @@
...
+    return get_row(self.sheet_handle, max_tokens)
+end
+
+XlsxioReader = defclass(XlsxioReader, nil)
+XlsxioReader.ATTRS{
+    -- full or relative path to the target .xlsx file. required.
+    filepath = DEFAULT_NIL,
+}
+
+function XlsxioReader:init()
```
lethosor, line 30: I think most file-opening functions just take the filename as a positional argument. That is somewhat limiting for classes defined with `defclass`, since `init()` only takes one argument. `json.lua` appears to handle this by implementing a separate global `open()` factory function - thoughts on doing that here?

docs/Lua API.rst
```diff
@@ -4305,41 +4305,44 @@ xlsxreader
...
  - ``sheet_handle open_sheet(file_handle, sheet_name)``
  - ``close_sheet(sheet_handle)``
- - ``cell_strings get_row(sheet_handle)``
+ - ``cell_strings get_row(sheet_handle, max_tokens)`` The ``max_tokens``
+   parameter is optional. If set to a number > 0, it limits the number of cells
+   read and returned for the row.
 
- Example::
+ Lua users would be better off using the Lua class wrappers, though. For
+ example::
```
lethosor, line 4320: It would be nice to have docs for the wrappers long-term, although I think the example is good and that's fine with me for now. This phrasing sounds to me like the classes were mentioned already - were they? If not, it would be worth at least calling out that the `XlsxReader` class (or whatever the recommended method for creating one is) exists before we get into the example. Perhaps mentioning that `XlsxSheetReader` exists in advance would help too, to minimize confusion with `XlsxReader`.

lethosor (approved):
Played around with it a bit, looks good! One minor usability suggestion, but I'm not committed to it.

plugins/lua/xlsxreader.lua
myk002: sgtm. done.

docs/Lua API.rst
myk002: good point. docs added.

plugins/xlsxreader.cpp
```diff
@@ -114,6 +116,8 @@ int get_row(lua_State *L) {
         auto cells = std::vector<std::string>();
         while (get_next_cell(sheet_handle, value)) {
             cells.push_back(value);
+            if (max_tokens > 0 && cells.size() >= max_tokens)
+                break;
```
lethosor, line 120: I'm seeing an interesting issue (I think it's an issue, not 100% sure) where passing `max_tokens <= num_cells_in_row` results in every other call to `get_row()` returning an empty row.

Specifically, I have a sheet with 8 columns (of irrelevant data). Calling `sheet:get_row(8)` returns a table with 8 elements, then a table with 0, then a table of 8, and so on. `sheet:get_row(3) behaves similarly, but `sheet:get_row(9)` always returns a table with 8 elements. I've compared to the data that's in the spreadsheet and confirmed that it isn't skipping any rows - it's just returning empty rows for every other call in the cases I've described.

Is this something you can reproduce on your end? I can attach a test file if not. My suspicion is that this might be going unnoticed if quickfort simply ignores empty rows, but it's worth looking into just in case other scripts end up using this feature.

lethosor:
Docs look great - thanks for updating those!

plugins/xlsxreader.cpp
lethosor: As far as I can tell, this is an issue upstream, even as of https://github.com/brechtsanders/xlsxio/commit/e3acace39e5fb153f5ce3500a4952c2bf93175bd

This patch seems to fix the issue, although it is a little less efficient when `max_tokens < num_cells_in_row`:
```diff
diff --git a/plugins/xlsxreader.cpp b/plugins/xlsxreader.cpp
index d9a781f1..12a59113 100644
--- a/plugins/xlsxreader.cpp
+++ b/plugins/xlsxreader.cpp
@@ -115,9 +115,9 @@ int get_row(lua_State *L) {
         std::string value;
         auto cells = std::vector<std::string>();
         while (get_next_cell(sheet_handle, value)) {
-            cells.push_back(value);
-            if (max_tokens > 0 && cells.size() >= max_tokens)
-                break;
+            if (max_tokens <= 0 || cells.size() <= max_tokens) {
+                cells.push_back(value);
+            }
         }
         Lua::PushVector(L, cells, true);
     }
```

myk002: Thanks for finding this! when quickfort reads less than all the cells, it's because it's scanning for blueprint modelines, so the extra empty rows don't affect anything. when the file is read for actual blueprint data, all cells are read so the issue doesn't manifest. I'll use your fix and add an explanatory comment.

lethosor:
Going to ignore the osx32 Buildmaster failure, which looks unrelated (`Unhandled exception: System.Data.SqlClient.SqlException (0x80131904): Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding. ---> System.ComponentModel.Win32Exception (0x80004005): The wait operation timed out`)

I will add an `int()` cast around `.size()` to silence the warnings, though. Maybe we should adjust our default warning flags...

lethosor (approved):

PR #1801, opened 2021-03-18
Update Lua API.rst

gaberau:
Corrected the dfhack.run_script documentation and added an example.


warmist:
Afaik both (the old way and new one) are valid but different ways to call a script. The original means that you put arguments into different variables, while the new one is you just pass a string. Though now that i look at it it's run_command example for run_script. Even then according to [this](https://github.com/DFHack/dfhack/blob/6db851a2ad74ea8fe839d1cf3d0a978d709b7593/library/lua/dfhack.lua#L738) it should work. If it does not the bug is in the code.

gaberau:
Yeah, I am not really sure if the old method works. The docs [here](https://docs.dfhack.org/en/stable/docs/Lua%20API.html?highlight=dfhack.run_script#general-script-api), simply list the command as "dfhack.run_script(name[,args...])". Unfortunately, that isn't very descriptive. It provides no examples and I am not sure if the name, args, or the entire thing is supposed to be a string, array, or some custom data type. I submitted a pull request because I found a method that worked and could be understood. How does the old method work? Do you have an example?

warmist:
Okay so reading over this a few times, i have these issues with the pullreq:

- The command syntax change (i.e the first line) is not in line with how they are usually presented. The current line indicates that name is required and the arguments can be any number (including 0). See this [code-syntax example](https://developers.google.com/style/code-syntax#optional-arguments) and this [example](https://developers.google.com/style/code-syntax#arguments-that-can-repeat)
- The example is for `run_command` not `run_script`.
- Run script can be used (unsure if any script uses this) in more flexible way. I.e. you can pass anything as argument(s). Your example shows only string but you could e.g. do `dfhack.run_script("add_to_the_event", dfhack.unit.find(unit_id), dfhack.historical_event.find(birth_of_theking_event_id))`

On the other hand:

- Current version has NO example.
- I'm unsure on what would be a common use-case of run_script
- Many scripts that i looked over do their own string parsing which makes it not as flexible as it could be
- I haven't looked into this very much so i could be wrong.

docs/Lua API.rst
```diff
@@ -4079,10 +4079,11 @@ DFHack core invokes the scripts in the *core context* (see above);
...
 from other scripts) in any context, via the same function the core uses:
 
-* ``dfhack.run_script(name[,args...])``
-
+* ``dfhack.run_script("name arg1 arg2 arg3")``
+   
   Run a lua script in hack/scripts/, as if it was started from dfhack command-line.
-  The ``name`` argument should be the name stem, as would be used on the command line.
+  The argument should be the exact command that would be written to the command line.
+  For example: dfhack.run_command("createitem WEAPON:ITEM_WEAPON_PICK COPPER 9999")
```
lethosor, line 4086: This should be under the run_command docs, not run_script. Also, per those docs, passing multiple arguments is recommended over a single argument.

lethosor:
Oh, I didn't realize there was a discussion here already when I commented on the diff earlier. I appreciate you going to the effort to address this, but there is already a similar example for `run_command()` [here](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#c-function-wrappers), and this example doesn't work for `run_script`.

I think I have a good idea of how to improve the docs `run_script`, even if it's not what you were trying to use. Is it okay with you if I do that and close this PR?

Current docs for run_command:

image

Regarding the syntax used in the docs, `args...` is our convention meaning "one or more arguments", and `[args...]` means "zero or more arguments".

@warmist `run_script` is actually what DFHack uses to run Lua scripts internally. For that reason, it has slightly less overhead than `run_command` when calling Lua scripts (because `run_command` ends up calling `run_script` internally). You're also correct that it doesn't do any type-checking. Core.cpp always passes string arguments to it, but you could probably cause some scripts to fail by passing other data types.

gaberau:
Alright, I didn't realize that the syntax for run_command and run_script were so similar. Should I edit the suggested revision?

lethosor:
Yeah, if you want to change this to an example for `run_script`, feel free. It would need to be a Lua script, so maybe something like this?
```lua
dfhack.run_script("weather", "clear")
```
I'd also request that we keep the `dfhack.run_script(name[,args...])` syntax for the signature.

gaberau:
Yeah, keeping the syntactical version would be best. All I really wanted was an example with some explanation for usage.

gaberau:
Alright, changes made and old syntax reinstated. Good to go?

docs/Lua API.rst
```diff
@@ -4079,10 +4079,22 @@ DFHack core invokes the scripts in the *core context* (see above);
...
   Run a lua script in hack/scripts/, as if it was started from dfhack command-line.
-  The ``name`` argument should be the name stem, as would be used on the command line.
+  The syntax for the argument is similar to dfhack.run_command.
+  The command can be passed as a table, multiple string
+  arguments, or a single string argument (not recommended - in this case, the
+  usual DFHack console tokenization is used).
+  The following examples are equivalent::
+        dfhack.run_script({'ls', '-a'})
+        dfhack.run_script('ls', '-a')
+        dfhack.run_script('ls -a')  -- not recommended
```
lethosor, line 4092: Looks like I forgot to explain this, sorry. `run_script` and `run_command` are not similar, and only the second example you have here is valid syntax. It won't actually work either, because `ls` isn't a Lua script (it's a built-in command). That's why I suggested the `weather` example, which _is_ a Lua script, and it's also fairly simple and unlikely to change in the future.

docs/Lua API.rst
```diff
@@ -4079,10 +4079,22 @@ DFHack core invokes the scripts in the *core context* (see above);
...
+  
+  Unlike ``dfhack.run_command``, ``dfhack.run_script`` is capable of taking the 
+  return value of a lua script as an argument. 
+  For example::
+        dfhack.run_scipt("process", dfhack.gui.getDwarfmodeViewDims());
```
lethosor, line 4097: Also to clarify @warmist's comment: while this is possible, it isn't expected to work, so I don't think we want to include it here. I believe passing non-strings to `run_command()` throws an error immediately, and if we were to reimplement `run_script()` today, it probably would as well.

docs/Lua API.rst
```diff
@@ -4079,10 +4079,22 @@ DFHack core invokes the scripts in the *core context* (see above);
 however it is possible to call them from any lua code (including
 from other scripts) in any context, via the same function the core uses:
 
-* ``dfhack.run_script(name[,args...])``
-
+* ``dfhack.run_script(command[, ...])``
+   
   Run a lua script in hack/scripts/, as if it was started from dfhack command-line.
-  The ``name`` argument should be the name stem, as would be used on the command line.
```
lethosor, line 4085: I'd suggest keeping `name` (see other comment) because it is the name of a script specifically, not an arbitrary command. Perhaps "name stem" could be replaced with "name of the script without the `.lua` extension" for clarity, though.

lethosor:
I just realized that you were targeting the `docs` branch with this PR. That branch is actually quite old and left over from some docs updates that we wanted to get on https://docs.dfhack.org/en/stable/ (as opposed to https://docs.dfhack.org/en/latest) without making a new release. Per https://docs.dfhack.org/en/stable/docs/Contributing.html, changes should generally be made to the `develop` branch instead.

Were you seeing the version of the documentation that you are editing somewhere? I'm asking so that I can make sure we're pointing people to the right version of the docs.

Anyway, I've adjusted this PR to merge into `develop`, and noticed some merge conflicts in the process. You are correct that the `docs` branch didn't have documentation for `run_command()` previously, but `develop` does, and it also has better docs for `run_script()` already, including some suggestions I made earlier:
image

I think the only thing from this PR that isn't in the latest docs yet is an example.

docs/Lua API.rst
gaberau: Not to be a contrarian, but passing commands to run_script actually does work without an error. 
Example:

test.lua:
`dfhack.run_script("personal/print", dfhack.gui.getDwarfmodeViewDims());
`
personal/print.lua:
```
local args = {...}
print(args)
```

Return from print.lua:
`table: 00000202BD97FAE0`

Although, I really don't know what use this has, and proving its existence is really little more than a proof of concept.

gaberau:
I found these docs just by visiting the GitHub page, navigating to the definition of dfhack.run_script, and clicking edit. Github automatically forked and let me edit the file online. In retrospect, I should have ensured that I was working off of the development branch. That shouldn't be an issue anymore, I have the branch on my machine now and everything should be set. 

As an aside, why do `dfhack.run_command()` and `dfhack.run_script()` exist as separate commands? Why not combine the two into one preferred function? 

Alright, I made a third commit with merge conflicts fixed. I still kept the notes about passing a function to `dfhack.run_script` because I found it useful to know. However, I have placed a blurb recommending against that usage.

gaberau:
Not sure why "Build / docs (pull_request)" is failing with error code 0. I have been editing the docs, is it cross comparing with the remotely stored version and triggering an issue because they are different?

myk002:
here are the errors from the build action (from the `details` link)

```
/home/runner/work/dfhack/dfhack/docs/Lua API.rst:4420: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/runner/work/dfhack/dfhack/docs/Lua API.rst:4423: WARNING: Unexpected indentation.
```

lethosor:
Yep, and if you click on "Details" next to the "docs/readthedocs.org" line, you can see what the docs look like here: https://dfhack--1801.org.readthedocs.build/en/1801/docs/Lua%20API.html

I'm pretty sure the issue is that code blocks need one blank line before and after them. Some of the ones you've added don't have either.

lethosor:
> I found these docs just by visiting the GitHub page, navigating to the definition of dfhack.run_script, and clicking edit. Github automatically forked and let me edit the file online.

Ok, did you click on the branch dropdown and select "docs" first? If so, we should probably just get rid of the branch to avoid confusion, since it wasn't intended to be long-lived. I went ahead and deleted it.

> As an aside, why do `dfhack.run_command()` and `dfhack.run_script()` exist as separate commands? Why not combine the two into one preferred function?

They do different things. `run_script` is only available in Lua and only runs Lua scripts. `run_command` can run any type of command, and as I mentioned before, it calls `run_script` internally if that command happens to be a Lua script.

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
...
+  from ``dfhack.run_script``.
+  Example::
+     dfhack.run_script("gaydar", "-all")
+
+  Unlike ``dfhack.run_command``, ``dfhack.run_script`` allows a different
+  different script to be executed and its value passed as an argument to the main
+  function called by ``dfhack.run_script``.
+
+  Example::
+      dfhack.run_script("some_script", dfhack.gui.getDwarfmodeViewDims());
```
lethosor, line 4430: Regarding terminology: this is passing in the value of a "Lua expression", not the result of a script (`getDwarfmodeViewDims` is a function, not a script).

There are two reasons I'm recommending against including this:
* You can do this with any Lua function, not just `run_script`. `run_command(someFunctionThatReturnsAString())` is equally valid, for instance.
* It behaves poorly if the arguments aren't a string, as in this example. *Some* scripts might work, but many scripts will not. For example, any script using `processArgs()` will fail if an argument isn't a string:
```
[lua]# dfhack.run_script('teleport', {})
./hack/lua/utils.lua:596: bad argument #1 to 'sub' (string expected, got table)
stack traceback:
	[C]: in function 'string.sub'
	./hack/lua/utils.lua:596: in function 'utils.processArgs'
	../../code/dfhack/scripts/teleport.lua:66: in local 'script_code'
	./hack/lua/dfhack.lua:680: in function 'dfhack.run_script_with_env'
	(...tail calls...)
```

In short: we shouldn't advertise that functions can be called with arguments of the wrong types, because that isn't supposed to work, and it often doesn't work.

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
...
+
+  While this implmenation is currently a functional, its purpose and stability
+  are limited. As such, ``dfhack.reqscript(name)`` should be used.
```
lethosor, line 4433: This isn't accurate. `run_script` *runs* a script, which is useful and stable (it is used every time any user runs a script). `reqscript` *imports* a script. Perhaps we add a cross-reference to `reqscript` here, but it is already in the "Importing Scripts" section immediately below this one.

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
...
+  Example::
+     dfhack.run_script("armoks-blessing", "all")
+  Scipts that use a dash(``-``) in their arguments still require one when called
+  from ``dfhack.run_script``.
+  Example::
+     dfhack.run_script("gaydar", "-all")
+
+  Unlike ``dfhack.run_command``, ``dfhack.run_script`` allows a different
+  different script to be executed and its value passed as an argument to the main
+  function called by ``dfhack.run_script``.
```
lethosor, line 4427: I'm not sure what this is supposed to mean - could you clarify? `dfhack.run_command("armoks-blessing")` and `dfhack.run_script("armoks-blessing")` should do essentially the same thing. The main differences I know of are that `run_script` propagates exceptions (already mentioned) and that `run_command` returns a command_result (mentioned in the `run_command` docs).

Typo: "different" is duplicated.

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
   command-line. The ``name`` argument should be the name of the script without
   its extension, as would be used on the command line.
 
-  Note that this function lets Lua errors propagate to the caller.
+  Example::
+     dfhack.run_script("armoks-blessing", "all")
+  Scipts that use a dash(``-``) in their arguments still require one when called
+  from ``dfhack.run_script``.
```
lethosor, line 4421: This is somewhat more specific than it needs to be - something like "arguments must be entered in the same way as in the DFHack console" could work, but I'm not sure we even need this at all, because there isn't a way to run commands that does change arguments before they get passed to commands.

The only meaningful difference I can think of between `run_script` and the console is that you can pass in spaces within an argument to `run_script` without needing to escape them, but I'm not sure of a script that would benefit from this off the top of my head.

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
...
+     dfhack.run_script("gaydar", "-all")
+
+  Unlike ``dfhack.run_command``, ``dfhack.run_script`` allows a different
+  different script to be executed and its value passed as an argument to the main
+  function called by ``dfhack.run_script``.
+
+  Example::
+      dfhack.run_script("some_script", dfhack.gui.getDwarfmodeViewDims());
+
+  While this implmenation is currently a functional, its purpose and stability
```
lethosor, line 4432: Typo: "implementation", "a"

docs/Lua API.rst
```diff
@@ -4415,12 +4415,29 @@ General script API
   command-line. The ``name`` argument should be the name of the script without
   its extension, as would be used on the command line.
 
-  Note that this function lets Lua errors propagate to the caller.
+  Example::
+     dfhack.run_script("armoks-blessing", "all")
```
lethosor, line 4419: For example, a change like this should fix the docs build issues:
```suggestion
  Example::
  
     dfhack.run_script("armoks-blessing", "all")
     
```

gaberau: Alright, sorry for the delay, a lot to read through. Your earlier comments are completely correct. Reviewing some of the things that I wrote, they don't make a ton of sense. I have removed some of my additions, fixed the spelling issues, and updated the example to be more accurate.

lethosor (approved):
Yeah, that looks better, and I confirmed that the examples work and that the docs build looks good. Thanks!

gaberau:
Happy to help in any way that I can

PR #1815, opened 2021-03-22
[test] move test harness out of test dir

myk002:
#1690

and install to test.lua in top-level scripts directory instead of test/main.lua

This is the plan of record from https://github.com/DFHack/dfhack/pull/1791#discussion_r593944821

but there is another option, if you'd prefer. We could move test.lua to the scripts repo and treat it like any other script (but only install it if BUILD_TESTS is set)

This PR depends on #1814


lethosor:
Unfortunately this breaks the `reqscript('test/main')` call in `test/test.lua`, which went undetected by CI for a couple reasons:
* Tests under Buildmaster failed at an early stage, because Buildmaster uses a cached copy of `run-tests.py` (see [1](https://github.com/BenLubar/build-env/blob/b4d88f135da72f23b6faf9dfebcc72a3c94393bf/gcc7/Dockerfile#L116), [2](https://github.com/BenLubar/build-env/blob/5ea53d9226d769821175aa197920fe7996b6528b/gcc7/dfhack-test#L27)), because the test environment just has access to the installed copy of DFHack (not the source)
* Tests under GitHub Actions passed (!), because `run-tests.py` determines its exit code from `test_status.json`, which did not contain any entries for tests that lived in a file that failed to import, because they couldn't be listed in the first place. Specifically, these messages went undetected by CI, and presumably also when you ran tests locally:
  * https://github.com/DFHack/dfhack/runs/2168768177#step:10:31: `Error when running file: ./hack/lua/dfhack.lua:625: Could not find script test/main`
  * https://github.com/DFHack/dfhack/runs/2168768177#step:10:167: `1 test files failed to load`

I fixed the latter in 9a29e5f1ce9a133972a79670f798dc25bdf4131a and updated this PR because GitHub was detecting a "conflict" (git merged it just fine locally, so I'm guessing it was just the rename that confused GitHub), and CI is now failing as it should.

As for how to fix this, I'm thinking of moving `expect` into a library file - maybe `library/lua/test/expect.lua`? I want to discourage tests from including it accidentally, since the raw/unwrapped versions of `expect.*()` won't work in tests, but documentation is probably enough. Thoughts? I can handle this separately, of course, just wanted to keep you updated on why CI is failing here.

myk002:
Ah, good catch. Yes, I missed that error.

I think either approach would fix the issue:

1. refactoring expect into `library/lua/test/expect.lua` or
2. moving `test.lua` into the scripts repo

While I like the idea of a `library/lua/test/` directory for test-related utility files, I think in this case moving `test.lua` (and test_test.lua, I suppose) into the scripts repo would be cleaner. As you said, the raw expect functions wouldn't be of any use for other scripts to include.

lethosor:
I think that in terms of maintenance, I'd prefer to keep things related to running tests all in one repo (so this one). I guess I should have phrased my concern a bit better: I don't want tests (distinct from scripts) to include the raw `expect` implementation unless they need to, but the current solution already allows them to include it, and some tests (well, one currently) need to include it, so I don't think there is much point in me actively trying to prevent tests from including it as long as it's sufficiently documented.

I'll come up with a proposed solution - is it okay with you if I update this branch? That's probably the easiest way to get CI to run on it and fix CI on this branch simultaneously.

lethosor:
Here's my proposal: 62776f55
(I'm heading offline, but hopefully it passes CI in my fork: https://github.com/lethosor/dfhack/runs/2181388727)

myk002:
Solution looks good, and absolutely feel free to update this branch!

PR #1816, opened 2021-03-22
[test] force reloading of scripts under test

myk002:
and invalidate scripts once tests are complete. this ensures that the
IN_TEST flag is respected.

note that this invalidates *all* scripts after running tests, even those that did not affect the test itself. we could potentially make this more targeted by remembering which scripts were loaded while IN_TEST was true and only reload those.

ooooorr, we could change the real reqscript to force a reload of the script if the value of IN_TEST in the calling environment doesn't match the value of IN_TEST that was in the environment when the script was loaded.

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -82,6 +82,14 @@ local function clean_require(module)
     return require(module)
 end
 
+-- similar to clean_require above; forces clean load of scripts directly
+-- included from a test file. note that this does *not* force transitively
+-- loaded scripts to be reloaded.
+local function clean_reqscript(name)
+    dfhack.internal.scripts[dfhack.findScript(name)] = nil
+    return reqscript(name)
+end
+
 local function ensure_title_screen()
     if df.viewscreen_titlest:is_instance(dfhack.gui.getCurViewscreen()) then
         return
@@ -129,6 +137,7 @@ local function build_test_env()
         expect = {},
         delay = delay,
         require = clean_require,
+        reqscript = clean_reqscript,
     }
     local private = {
         checks = 0,
@@ -182,8 +191,18 @@ local function save_test_status(status)
     json.encode_file(status, STATUS_FILE)
 end
 
+-- causes scripts to be reloaded the next time they are reqscript()'d. this
+-- allows scripts that change their behavior based on the value of
+-- dfhack.internal.IN_TEST to return to non-test behavior after tests are run.
+local function invalidate_scripts()
+    for k,_ in pairs(dfhack.internal.scripts) do
+        dfhack.internal.scripts[k] = nil
+    end
+end
+
 local function finish_tests(done_command)
     dfhack.internal.IN_TEST = false
+    invalidate_scripts()
     if done_command and #done_command > 0 then
         dfhack.run_command(done_command)
     end
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -82,14 +82,16 @@ local function clean_require(module)
     return require(module)
 end
 
--- similar to clean_require above; forces clean load of scripts directly
--- included from a test file. note that this does *not* force indirectly loaded
--- scripts (that is, scripts that are reqscript()'d by the scripts that are
--- reqscript()'d by the test file) to be reloaded.
+-- forces clean load of scripts directly or indirectly included from the test
+-- file. we use our own scripts table instead of the one in dfhack.internal so
+-- we don't affect the state scripts that are used outside the test harness.
+local test_scripts = {is_test_scripts=true}
+local test_envvars = {}
 local function clean_reqscript(name)
-    dfhack.internal.scripts[dfhack.findScript(name)] = nil
-    return reqscript(name)
+    return dfhack.script_environment(name, true, test_envvars, test_scripts)
 end
+test_envvars.require = clean_require
+test_envvars.reqscript = clean_reqscript
 
 local function ensure_title_screen()
     if df.viewscreen_titlest:is_instance(dfhack.gui.getCurViewscreen()) then
@@ -192,18 +194,8 @@ local function save_test_status(status)
     json.encode_file(status, STATUS_FILE)
 end
 
--- causes scripts to be reloaded the next time they are reqscript()'d. this
--- allows scripts that change their behavior based on the value of
--- dfhack.internal.IN_TEST to return to non-test behavior after tests are run.
-local function invalidate_scripts()
-    for k,_ in pairs(dfhack.internal.scripts) do
-        dfhack.internal.scripts[k] = nil
-    end
-end
-
 local function finish_tests(done_command)
     dfhack.internal.IN_TEST = false
-    invalidate_scripts()
     if done_command and #done_command > 0 then
         dfhack.run_command(done_command)
     end
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -85,7 +85,7 @@ end
 -- forces clean load of scripts directly or indirectly included from the test
 -- file. we use our own scripts table instead of the one in dfhack.internal so
 -- we don't affect the state scripts that are used outside the test harness.
-local test_scripts = {is_test_scripts=true}
+local test_scripts = {}
 local test_envvars = {}
 local function clean_reqscript(name)
     local path = dfhack.findScript(name)
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -83,8 +83,9 @@ local function clean_require(module)
 end
 
 -- similar to clean_require above; forces clean load of scripts directly
--- included from a test file. note that this does *not* force transitively
--- loaded scripts to be reloaded.
+-- included from a test file. note that this does *not* force indirectly loaded
+-- scripts (that is, scripts that are reqscript()'d by the scripts that are
+-- reqscript()'d by the test file) to be reloaded.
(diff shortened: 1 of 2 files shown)


test/main.lua
```diff
@@ -303,8 +312,18 @@ local function save_test_status(status)
     json.encode_file(status, STATUS_FILE)
 end
 
+-- causes scripts to be reloaded the next time they are reqscript()'d. this
+-- allows scripts that change their behavior based on the value of
+-- dfhack.internal.IN_TEST to return to non-test behavior after tests are run.
+local function invalidate_scripts()
+    for k,_ in pairs(dfhack.internal.scripts) do
+        dfhack.internal.scripts[k] = nil
```
lethosor, line 320: This would affect scripts that rely on global variables persisting across runs - values of those would be cleared permanently. Not a likely issue, but see `devel/lua-example` for a simple example of a script that would be reset by this logic.

test/main.lua
```diff
@@ -203,6 +203,14 @@ local function clean_require(module)
     return require(module)
 end
 
+-- similar to clean_require above; forces clean load of scripts directly
+-- included from a test file. note that this does *not* force transitively
```
lethosor, line 207: I'm not sure quite how many levels you mean by "transitively" here, but I'm pretty sure that *any* script that calls `reqscript()` would not use our patched version here, because reqscript in its global environment would be coming from `BASE_G`: https://github.com/DFHack/dfhack/blob/dbd35359be68f98c0a8a0cd1fbf783235b3ee84a/library/lua/dfhack.lua#L657

lethosor (commented):
> ooooorr, we could change the real reqscript to force a reload of the script if the value of IN_TEST in the calling environment doesn't match the value of IN_TEST that was in the environment when the script was loaded.

This is an interesting idea, and spent a little time looking into ways that it could be implemented. I think that both that approach and the current implementation end up mixing the implementation details of the test runner and the script loader, though. Either the test runner would rely on how `reqscript()` uses `internal.scripts` (the current implementation) or `reqscript()` would rely on how `internal.IN_TEST` is set by the test runner (this proposal). Yeah, both pieces of state are in the same `dfhack.internal` table, but that's a bit of a mess with no clear ownership at the moment. Given that the test runner's override for `require()` works without requiring changes to `require()` internally, I'd like to go with something similar for overriding `reqscript()`, i.e. the current implementation. (Side note: the behavior of `package.*` is more-or-less defined by the Lua standard, so it has stronger compatibility guarantees than stuff in `dfhack.internal`.)

One option that *might* produce better results could be to replace `internal.scripts` with an empty table before loading tests, then restore it when tests finish running. This would avoid affecting (and potentially breaking) scripts that track their own global state. My thoughts:
* Saving scripts could be done in `main()` before `run_tests()` is called
* Restoring scripts could be done in `finish_tests()`
* `clean_reqscript()` could be removed entirely (I think)
* This _might_ require changing all references to `scripts` in dfhack.lua to `internal.scripts` instead, due to Lua scoping rules (otherwise that file will use the original `internal.scripts` table)
* Transitively-included scripts should work automatically (even multiple levels). An unlikely use-case, for sure, but nice to avoid potential issues for things like `quickfort`.

Any thoughts on this? The current implementation works too, of course.

test/main.lua
myk002: I'll clarify -- I meant that while scripts directly reqscript'd by the test file will be reloaded, scripts reqscript'd by the reqscript'd scripts would not.

myk002: done

test/main.lua
myk002: that is true. is that desired behavior or undesired, though?  If you're running tests, is it more useful to keep or reset global state?

myk002:
> One option that _might_ produce better results could be to replace `internal.scripts` with an empty table before loading tests, then restore it when tests finish running.

This actually describes my first attempt at this PR, but the caching of the table internal to dfhack.lua scared me off. I was trying to avoid changing that file. I do agree that the handling of transitively loaded scripts is attractive, though. I'll give this implementation a try.

> Edit: another option I thought of would be to allow calling `script_environment()` with an additional table to store script environments in instead of `internal.scripts`. This would result in `clean_reqscript()` being a wrapper around e.g. `script_environment(name, true, isolated_script_envs)`, and would still require the `scripts -> internal.scripts` change I mentioned above, plus a handful of other changes in dfhack.lua to pass this additional table around, which could become a bit messier.

the benefit here being that we don't need to break into dfhack.internal from the test runner, right? we might even be able to leverage this to set IN_TEST without referencing dfhack.internal as well. let me think about this a bit.

lethosor:
> This actually describes my first attempt at this PR, but the caching of the table internal to dfhack.lua scared me off. I was trying to avoid changing that file. I do agree that the handling of transitively loaded scripts is attractive, though. I'll give this implementation a try.

Looks like the local reference (`scripts = dfhack.internal.scripts`) was introduced in 7eb4fc19de542db0d3e271123f24773e0c8c481e, possibly to prevent scripts from changing `internal.scripts` to a non-table and breaking dfhack.lua, but that's just a guess. (It could very well just be to cut down on typing - I made a similar change for that reason.)



> the benefit here being that we don't need to break into dfhack.internal from the test runner, right? we might even be able to leverage this to set IN_TEST without referencing dfhack.internal as well. let me think about this a bit.

Right. Not quite sure what you mean about IN_TEST, though.

I looked through dfhack.lua a bit more, and I think that instead of modifying `script_environment()`, we could: 
1. Modify `run_script_with_env()` to take an optional `scripts` (?) parameter in `flags` to override `scripts` locally - this should avoid breaking backwards-compatibility
2. Implement `clean_reqscript(name)` as `dfhack.run_script_with_env(nil, name, {module=true, module_strict=true, scripts=some_tmp_script_table})`. `some_tmp_script_table` would be owned by `test.lua` somewhere.

Yet another option: it doesn't look to me like anything uses the first parameter of `run_script_with_env`, so it _could_ be possible to change it to ignore the contents of `internal.scripts` if that parameter is set. That would allow skipping (1) and replacing (2) with ``dfhack.run_script_with_env({}, ...`. I expected this to work before I tried it, but there is more risk of breaking compatibility by making it work this way. (`run_script_with_env` is undocumented, though...)

myk002:
> 2. Implement `clean_reqscript(name)` as `dfhack.run_script_with_env(nil, name, {module=true, module_strict=true, scripts=some_tmp_script_table})`. `some_tmp_script_table` would be owned by `test.lua` somewhere.
> 
> Yet another option: it doesn't look to me like anything uses the first parameter of `run_script_with_env`, so it _could_ be possible to change it to ignore the contents of `internal.scripts` if that parameter is set. That would allow skipping (1) and replacing (2) with `dfhack.run_script_with_env({}, ...`. I expected this to work before I tried it, but there is more risk of breaking compatibility by making it work this way. (`run_script_with_env` is undocumented, though...)

I actually ended up driving `clean_reqscript` through `dfhack.script_environment` instead of `dfhack.run_script_with_env` and using the envVars parameter to propagate the `reqscript` override for transitively loaded scripts. I could potentially make fewer changes to dfhack.lua by duplicating more of `dfhack.script_environment` in `clean_reqscript`, but I thought this way would be more maintainable. Totally fine if you prefer the other way, though.

> Not quite sure what you mean about IN_TEST

I meant that IN_TEST could just be part of the environment, referenced by just `IN_TEST`, as opposed to putting it in `dfhack.internal.IN_TEST`, thus removing all references to `dfhack.internal` from outside of `dfhack.lua`. The changes to dfhack.lua could propagate the IN_TEST value just like they now do for the `reqscript` override.

I haven't pushed these changes yet; still testing and trying to understand all the code paths. Just giving you an update on current progress.

lethosor:
> I actually ended up using the envVars parameter to propagate the `reqscript` override for transitively loaded scripts.

Oh, that's interesting - I guess `envVars` turned out to be useful after all! This does assume all scripts use `reqscript` for their script-importing needs, but I think that's fine - scripts can be changed a bit to support tests if they don't already, so we don't need to go to a lot of effort to support scripts that call `run_script_with_env` in weird ways, for instance (not sure if that's feasible at all, but fortunately there aren't any scripts that do this).

> I meant that IN_TEST could just be part of the environment, referenced by just `IN_TEST`, as opposed to putting it in `dfhack.internal.IN_TEST`, thus removing all references to `dfhack.internal` in the test harness.

Ah. Unfortunately, Lua libraries loaded with `require()` don't have an equivalent to `envVars` that I know of (`plugins/lua/workflow.lua` was the original inspiration for `dfhack.internal.IN_TEST`). 

At any rate, this looks clean to me as-is. Thanks for dealing with this (and apologies for the years of messy code that make it hard)!

I'll try to kick off Buildmaster again, if I can figure out which webhook call failed.

myk002:
Idea for another PR: perhaps we could replace the current clean_require wrapper with a call that swaps out the package table for the duration of the tests. This would not only solve the transitive load problem for require (the solution in this PR solves reqscript -> require but not require -> require) but would also ensure module state is restored after the testing is done. Currently, modules loaded for tests retain their IN_TEST status after the tests are done.

lethosor:
Ha, you're right, the wrapper for `require()` is nowhere near as robust.

lethosor (approved):

PR #1818, opened 2021-03-28
tagged union support for lua

BenLubar:
- [x] handle enums with custom sizes
- [x] attach tagged union metadata when accessing a union field
- [ ] pass on tagged union data when retrieving items from a vector of unions
- [ ] filter union field iterator to only return tagged field (or nothing)

### DIFF
--- a/library/DataDefs.cpp
+++ b/library/DataDefs.cpp
@@ -159,6 +159,14 @@ enum_identity::enum_identity(size_t size,
     }
 }
 
+enum_identity::enum_identity(enum_identity *base_enum, type_identity *override_base_type)
+    : enum_identity(override_base_type->byte_size(), base_enum->getScopeParent(),
+                    base_enum->getName(), override_base_type, base_enum->first_item_value,
+                    base_enum->last_item_value, base_enum->keys, base_enum->complex,
+                    base_enum->attrs, base_enum->attr_type)
+{
+}
+
 enum_identity::ComplexData::ComplexData(std::initializer_list<int64_t> values)
 {
     size_t i = 0;
--- a/library/include/LuaWrapper.h
+++ b/library/include/LuaWrapper.h
@@ -127,6 +127,8 @@ namespace LuaWrapper {
     struct DFRefHeader {
         void *ptr;
         const struct_field_info *field_info;
+        const void *tag_ptr;
+        const type_identity *tag_identity;
     };
 
     /**
--- a/library/LuaTypes.cpp
+++ b/library/LuaTypes.cpp
@@ -616,6 +616,15 @@ static int meta_struct_index(lua_State *state)
     if (!field)
         return 1;
     read_field(state, field, ptr + field->offset);
+    if (field->mode == struct_field_info::SUBSTRUCT || field->mode == struct_field_info::CONTAINER)
+    {
+        auto struct_type = (struct_identity*)get_object_identity(state, 1, "read", false);
+        if (auto tag_field = find_union_tag(struct_type->getFields(), field))
+        {
+            get_object_ref_header(state, -1)->tag_ptr = ptr + tag_field->offset;
+            get_object_ref_header(state, -1)->tag_identity = tag_field->type;
+        }
+    }
     return 1;
 }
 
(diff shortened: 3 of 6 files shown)


library/include/LuaWrapper.h
```diff
@@ -127,6 +127,8 @@ namespace LuaWrapper {
     struct DFRefHeader {
         void *ptr;
         const struct_field_info *field_info;
+        const void *tag_ptr;
+        const type_identity *tag_identity;
```
lethosor, line 131: Is this something that could go in `struct_field_info` instead? Or is this a "back-reference" from the field being tagged to the field acting as the tag? Since (as far as I know) DFRefHeader objects are created every time a DF object/field is accessed from Lua, I'd like to keep the size as small as possible (but I suppose this might not have a significant impact if they are garbage-collected).

library/DataDefs.cpp
```diff
@@ -159,6 +159,14 @@ enum_identity::enum_identity(size_t size,
     }
 }
 
+enum_identity::enum_identity(enum_identity *base_enum, type_identity *override_base_type)
+    : enum_identity(override_base_type->byte_size(), base_enum->getScopeParent(),
+                    base_enum->getName(), override_base_type, base_enum->first_item_value,
+                    base_enum->last_item_value, base_enum->keys, base_enum->complex,
+                    base_enum->attrs, base_enum->attr_type)
+{
+}
```
lethosor, line 168: Nice! This actually appears to work for fields even outside of vectors now, e.g. you can modify any fields in this structure from `gui/gm-editor` and see enum metadata:
```xml
        <enum name='season8' type-name='season' base-type='int8_t'/>
        <enum name='season16' type-name='season' base-type='int16_t'/>
        <enum name='season32' type-name='season' base-type='int32_t'/>
        <stl-vector name='v_season8'>
            <enum type-name='season' base-type='int8_t'/>
        </stl-vector>
        <stl-vector name='v_season16'>
            <enum type-name='season' base-type='int16_t'/>
        </stl-vector>
        <stl-vector name='v_season32'>
            <enum type-name='season' base-type='int32_t'/>
        </stl-vector>
```

Was there a reason for bundling this with this PR? If we're confident in it, I can see it being useful as a standalone change.

(One oddity on my end is that inserting an element into the 16-bit and 32-bit vectors in gui/gm-editor with `Alt-i`, `Enter` seems to produce uninitialized elements, but not the 8-bit vector. Not sure if that's an issue somewhere else, though.)

lethosor (commented):
> filter union field iterator to only return tagged field (or nothing)

Thoughts on whether we would want to keep the current logic that skips union fields with vtables? My understanding is that this is only used in a few cases, like `ui_look_list`, where even trying to print a field that isn't valid will crash. However, as far as I know, we don't currently distinguish between whether accessing a certain union field when the tag is set to the corresponding value is _always_ safe, or whether the union field could still be uninitialized depending on other factors.

library/DataDefs.cpp
BenLubar: This can be cherry-picked if desired. The reason it's here is that I was working on this:

https://github.com/DFHack/df-structures/blob/9b7ee3879aa07fe9f0348052db97e62a23d05863/df.history.xml#L543-L556

library/include/LuaWrapper.h
BenLubar: If you're accessing a vector item, I'm pretty sure only the first field of DfRefHeader is currently non-null. The two added fields are only used (in code that has yet to be written) in three cases: when a tagged union is accessed from a struct, when a vector of tagged unions is accessed from a struct, and when an item of that vector is accessed.

If there's some other way to store the information (like how type identities are stored), I'm not familiar enough with lua to know how it's done, but it seems like that could be a more memory-efficient way to handle it.

lethosor:
This only affects `pairs()`, right? i.e. will accessing other union elements that don't match the tag still work? If not, this would definitely be worth mentioning in the changelog.

I could also take on a test adjustment - for the test you changed, we probably also want to check that other union items _aren't_ listed by `pairs()`.

PR #1823, opened 2021-04-02
[test] error on unmatched printerr output during a test

myk002:
- provides `expect.printerr_match()` for matching `dfhack.printerr()` output
- fails tests if `printerr` is called outside the `printerr_match` wrapper
- changes api of `expect.error` and `expect.error_match` to mirror the new `printerr_match` api
- adds unit tests for `expect.error`, `expect.error_match`, and the new `expect.printerr_match`

I also moved `expect.lua` from `library/lua/test/` to `library/lua/test_util` since luacov is configured to ignore any file with `/test/` in its path. We can fix this in luacov if you'd rather keep the path as is, though. Moving the file makes the diff more difficult to read in github, unfortunately, but commandline `git diff` will properly show the diff across the rename.

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -1,7 +1,7 @@
 -- DFHack developer test harness
 --@ module = true
 
-local expect = require 'test.expect'
+local expect = require 'internal.expect'
 local json = require 'json'
 local script = require 'gui.script'
 local utils = require 'utils'
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -141,6 +141,49 @@ local function load_test_config(config_file)
     return config
 end
 
+local function run_expect_func(func, ...)
+    local args = {...}
+    local saved_printerr = dfhack.printerr
+    local printerr_called = false
+    dfhack.printerr = function(msg) printerr_called = true end
+    return dfhack.with_finalize(
+        function() dfhack.printerr = saved_printerr end,
+        function()
+            local ret = {func(table.unpack(args))}
+            if printerr_called then
+                return {false,
+                        "dfhack.printerr was called outside of" ..
+                        " expect.printerr_match(). please wrap your test" ..
+                        " with expect.printerr_match()."}
+            end
+            return ret
+        end
+    )
+end
+
+local function wrap_expect(func, private)
+    return function(...)
+        private.checks = private.checks + 1
+        local ret = run_expect_func(func, ...)
+        local ok = table.remove(ret, 1)
+        if ok then
+            private.checks_ok = private.checks_ok + 1
+            return
+        end
+        local msg = ''
+        for _, part in pairs(ret) do
+            if part then
+                msg = msg .. ': ' .. tostring(part)
+            end
+        end
+        msg = msg:sub(3) -- strip leading ': '
+        dfhack.printerr('Check failed! ' .. (msg or '(no message)'))
+        local info = debug.getinfo(2)
+        dfhack.printerr(('  at %s:%d'):format(info.short_src, info.currentline))
+        print('')
+    end
+end
+
 local function build_test_env()
     local env = {
         test = utils.OrderedTable(),
@@ -157,26 +200,7 @@ local function build_test_env()
         checks_ok = 0,
     }
     for name, func in pairs(expect) do
-        env.expect[name] = function(...)
-            private.checks = private.checks + 1
-            local ret = {func(...)}
-            local ok = table.remove(ret, 1)
-            local msg = ''
-            for _, part in pairs(ret) do
-                if part then
-                    msg = msg .. ': ' .. tostring(part)
-                end
-            end
-            msg = msg:sub(3) -- strip leading ': '
-            if ok then
-                private.checks_ok = private.checks_ok + 1
-            else
-                dfhack.printerr('Check failed! ' .. (msg or '(no message)'))
-                local info = debug.getinfo(2)
-                dfhack.printerr(('  at %s:%d'):format(info.short_src, info.currentline))
-                print('')
-            end
-        end
+        env.expect[name] = wrap_expect(func, private)
     end
     setmetatable(env, {__index = _G})
     return env, private
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -141,30 +141,14 @@ local function load_test_config(config_file)
     return config
 end
 
-local function run_expect_func(func, ...)
-    local args = {...}
-    local saved_printerr = dfhack.printerr
-    local printerr_called = false
-    dfhack.printerr = function(msg) printerr_called = true end
-    return dfhack.with_finalize(
-        function() dfhack.printerr = saved_printerr end,
-        function()
-            local ret = {func(table.unpack(args))}
-            if printerr_called then
-                return {false,
-                        "dfhack.printerr was called outside of" ..
-                        " expect.printerr_match(). please wrap your test" ..
-                        " with expect.printerr_match()."}
-            end
-            return ret
-        end
-    )
-end
-
+-- we have to save and use the original dfhack.printerr here so our test harness
+-- output doesn't trigger its own dfhack.printerr usage detection (see
+-- detect_printerr below)
+local orig_printerr = dfhack.printerr
 local function wrap_expect(func, private)
     return function(...)
         private.checks = private.checks + 1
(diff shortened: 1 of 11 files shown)


library/lua/test/expect.lua
```diff
@@ -101,25 +101,130 @@ function expect.table_eq(a, b, comment)
             ('key %s: "%s" ~= "%s"'):format(keystr, diff[1], diff[2])
 end
 
-function expect.error(func, ...)
-    local ok, ret = pcall(func, ...)
+function expect.error(func, comment)
```
lethosor, line 104: +1 on this change. I [previously mentioned](https://github.com/DFHack/scripts/pull/263#discussion_r603772718) `expect.eq` taking `...` params, but it actually takes a single `comment` param too (as I'm sure you noticed). Much simpler to deal with here.

library/lua/test/expect.lua
```diff
@@ -101,25 +101,130 @@ function expect.table_eq(a, b, comment)
...
+                end
+                goto continue
+            end
+        end
+        if true_count > 0 then
+            true_count = true_count - 1
+            goto continue
+        end
+        return false, comment, ('unmatched printerr message: "%s"'):format(msg)
+        ::continue::
```
lethosor, line 212: I think the syntax error here is because Lua doesn't allow any code after a `return` statement, other than `end`. This is a syntax error too, for similar reasons:
```lua
function f(n)
  return n 
  print(n) -- syntax error: expected 'end'
end
```

In this case, couldn't you put the `return` in an `else` block and get rid of the `goto` a couple lines above?

ci/test.lua
```diff
@@ -141,6 +141,49 @@ local function load_test_config(config_file)
...
+    local args = {...}
+    local saved_printerr = dfhack.printerr
+    local printerr_called = false
+    dfhack.printerr = function(msg) printerr_called = true end
+    return dfhack.with_finalize(
+        function() dfhack.printerr = saved_printerr end,
+        function()
+            local ret = {func(table.unpack(args))}
+            if printerr_called then
+                return {false,
```
lethosor, line 154: Is this supposed to be a table? Most `expect` functions return multiple values instead of a table.

lethosor (commented):
> code is untested; seeking feedback on the api and general approach before proceeding.

Sorry, I totally overlooked this and went right into code-review mode. I think the API makes sense to me. Thanks for refactoring the runner too!

ci/test.lua
myk002: it's a table so that we capture the variable number of comment values. It's used in `wrap_expect` below. [1] is true/false, [2] and on are comments.

library/lua/test/expect.lua
myk002: done

library/lua/test/expect.lua
```diff
@@ -101,25 +101,130 @@ function expect.table_eq(a, b, comment)
...
+-- callback is run. the check passes if all printerr messages are matched by
+-- specified matchers and no matchers remain unmatched.
+--
+-- matcher can be:
+--  a string, interpreted as a lua pattern that matches a message
+--  a function that takes the string message and returns a boolean (true means
+--    match)
+--  the literal value true, which matches any message
+--  the literal value false, nil, or an empty table, which match the absence of
+--    printerr messages
```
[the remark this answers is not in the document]
myk002: agreed. `true` handling has been removed. simplifies the code too.

ci/test.lua
```diff
@@ -1,7 +1,7 @@
 -- DFHack developer test harness
 --@ module = true
 
-local expect = require 'test.expect'
+local expect = require 'internal.expect'
```
lethosor, line 4: I saw your reasoning in 75da527c9031e3ab202ce3ddaf87057b0b887d29, but I kind of still feel that this filename should have `test` somewhere in its name to reflect the fact that it's only intended to be used for tests. It doesn't have to have `/test/`, though - maybe `test_util/expect.lua`?. It's also not strictly related to the existing `dfhack.internal` module of the same name, so I'd tend away from `internal`.

By the way, where is the luacov configuration that was causing the original issue here? Is it built-in/overridable? Searching for `/test` or `test/` in the DFHack and luacov repos didn't seem to turn up anything related. (Asking to determine if my naming suggestion works and/or is necessary.)

myk002: The luacov config is totally under our control:
https://github.com/DFHack/scripts/blob/c0e0c307971da9299022cf9c39f3c610c5c0432a/devel/luacov.lua#L95-L100

The intention there is that test files themselves shouldn't contribute to the coverage report.

myk002: also I moved `expect.lua` from `internal/` to `test_util/`

myk002:
example usage of new APIs in https://github.com/DFHack/scripts/pull/263/commits/429a843f4b933c1316c181b715bda4b8c5e14d34

ci/test.lua
lethosor: Ah, thanks (I forgot to check there...). Ideally I'd like to have it use the actual path to tests, but communicating that between scripts could get messy since the test path could change at runtime. Checking for `/test/` is totally fine with me.

library/lua/test/expect.lua
```diff
@@ -101,25 +101,130 @@ function expect.table_eq(a, b, comment)
...
+end
+
+-- matches errors thrown from the specified function. the check passes if an
+-- error is thrown and the thrown error matches the specified matcher.
+--
+-- matcher can be:
+--  a string, interpreted as a lua pattern that matches the error message
+--  a function that takes an err object and returns a boolean (true means match)
+--  the literal value true, which matches any thrown error
+function expect.error_match(matcher, func, comment)
```
lethosor: I might have missed this earlier - I see this changed from `func, matcher, comment` to `matcher, func, comment`. Is that intended to match the `expected, actual, comment` API of some of the other checks, like `expect.eq()`? (I know it's meant to match the API of `expect.printerr_match()`). Totally fine with me to standardize on something; I don't think I really considered this earlier, and making this change now with limited callers is better than making it later.

library/lua/test_util/expect.lua
```diff
@@ -0,0 +1,234 @@
...
+-- if a mapped matcher is matched, it will have its value decremented by 1.
+function expect.printerr_match(matcher, func, comment)
+    local saved_printerr = dfhack.printerr
+    local messages = {}
+    dfhack.printerr = function(msg) table.insert(messages, msg) end
+    dfhack.with_finalize(
+        function() dfhack.printerr = saved_printerr end,
+        func)
+    if type(matcher) ~= 'table' then matcher = {matcher} end
+    while messages[1] do
```
lethosor, line 167: Nitpick: should this be `while #messages > 0`? It's technically possible to call `printerr(nil)` or `printerr(false)` (although nobody _should_ do that, but maybe something equivalent to `printerr()` could happen accidentally). This is further complicated by the fact that `ipairs()` just... silently stops iterating if it encounters a `nil` value (`#` does not do that, fortunately), so I suspect `unmatched_messages` would end up empty.

Maybe something like this could work to at least error out in weird cases?
```lua
while #messages > 0 do 
    local msg = messages[1]
    assert(msg, 'printerr() called with no message')
```

lethosor:
FYI, the earlier Buildmaster failure appears to have been an internal failure - I re-ran it successfully (and tests passed!): https://buildmaster.lubar.me/applications/17/builds/build?buildId=9870

library/lua/test/expect.lua
myk002: that's right, it's now `expected, actual, comment`, but it wasn't just a swap of the first two params. The original `error_match` was `func, matcher, func_args..` with no provision for a comment.

library/lua/test_util/expect.lua
myk002: I did not know that about '#' -- I thought that it stopped short at a `nil` like `ipairs` would. I'll get this handled. Thanks!

myk002: ok, so `nil` can never be in `messages` because `table.insert(messages, nil)` is a noop, so just `while #messages > 0` should be sufficient. need to add a few unit tests tho.

myk002: done

lethosor (approved):
Thanks!

PR #1828, opened 2021-04-06
[test] wrap dfhack.run_script for tests

myk002:
so `dfhack.internal.scripts` isn't polluted with `IN_TEST` envs

this patches `dfhack.run_script`. there are other holes, though, like `dfhack.run_command`, `dfhack.enable_script` and, of course, `dfhack.run_script_with_env` itself. This minimal fix *might* be good enough for now, but a more robust solution would be to deprecate the global `internal.IN_TEST` and instead have the `IN_TEST` flag and the `scripts` table automatically inherited from parent environments. That way we wouldn't have to whack-a-mole patch all the entrypoints.

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -82,11 +82,23 @@ local function clean_require(module)
     return require(module)
 end
 
--- forces clean load of scripts directly or indirectly included from the test
--- file. we use our own scripts table instead of the one in dfhack.internal so
--- we don't affect the state scripts that are used outside the test harness.
+-- clean_run_script and clean_reqscript force a clean load of scripts directly
+-- or indirectly included from the test file. we use our own scripts table
+-- instead of the one in dfhack.internal so we don't affect the state scripts
+-- that are used outside the test harness.
 local test_scripts = {}
 local test_envvars = {}
+
+-- clean_run_script is accessed via the dfhack table, not directly from the env.
+-- therefore we use this function in wrap_test() below and not in test_envvars.
+local function clean_run_script(name, ...)
+    return dfhack.run_script_with_env(
+        test_envvars,
+        name,
+        {scripts=test_scripts},
+        ...)
+end
+
 local function clean_reqscript(name)
     local path = dfhack.findScript(name)
     if test_scripts[path] then return test_scripts[path].env end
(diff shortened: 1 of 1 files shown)


lethosor:
Do we need to patch `run_script()` (or `run_command()` for that matter)? The overrides for `require()` and `reqscript()` are primarily (from my understanding, anyway) to allow modules/scripts to export different things for tests to access, not necessarily to change their runtime behavior, so it's fine (and maybe preferable?) for commands run under e.g. `run_script()` to not know that they're being run under tests.

I'm assuming this is inspired by tests like [this one](https://github.com/DFHack/scripts/blob/ee6ae4b91174df3b2a965a6bdc95ff3c4d3a8303/test/quickfort/notes_unit.lua#L3-L7) - let me know if I'm missing something.

myk002:
Spot on on which test highlighted the issue. The problem is that when run_script runs a script without the overrides, it loads it into the real scripts table when IN_TEST is set.

I don't think the approach in the PR would work for run_command, but what we could do to handle things uniformly is to wrap run_script and run_command and set IN_TEST to false for the duration of the called script/command.

lethosor:
Ah, thanks. Not as much of an issue as the inverse (loading a module with `IN_TEST=false` when running tests), but still worth fixing for consistency.

Another idea I had: we could only set IN_TEST at test-file-load time, so it would only apply to top-level `require()`/`reqscript()` calls.

Ultimately, it may be more effort than it's worth to get `IN_TEST` set properly at import time for both `require()` and `reqscript()` (and related functions). I'm thinking a simpler long-term fix could be to have modules export a global table named `__test_data` or something like that unconditionally, which would make module loading in particular less complicated. It might be possible to do some metatable tricks to hide this table outside of tests, or we could just rely on devs/scripts to be well-behaved.

That would potentially allow removing `IN_TEST` entirely, since it's currently only used at import time. If there are valid use-cases for it at runtime, maybe we could keep it (but I'm struggling to think of ones that couldn't be handled somehow by mocks).

lethosor (approved):

lethosor:
I split the above comment into a new issue: #1830

PR #1831, opened 2021-04-09
Add some mock-related test helpers

lethosor:
This is currently a relatively low-level `patch()` implementation, inspired by pytest's `mock.patch()`, since I noticed several recent tests using the same `with_finalize()` pattern to patch builtins (note: most of these tests are in the scripts repo, so I didn't modify them here, but see `ci/test.lua` for example usage). Open to adding other things that might be useful.

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -281,19 +281,20 @@ local function sort_tests(tests)
 end
 
 local function wrap_test(func)
-    local saved_printerr, saved_run_script = dfhack.printerr, dfhack.run_script
+    local saved_printerr = dfhack.printerr
     local printerr_called = false
-    dfhack.printerr = function(msg)
-            if msg == nil then return end
-            saved_printerr(msg)
-            printerr_called = true
-        end
-    dfhack.run_script = clean_run_script
-    return dfhack.with_finalize(
-        function()
-            dfhack.printerr = saved_printerr
-            dfhack.run_script = saved_run_script
-        end,
+    local printerr_wrapper = function(msg)
+        if msg == nil then return end
+        saved_printerr(msg)
+        printerr_called = true
+    end
+
+    return mock.patch(
+        {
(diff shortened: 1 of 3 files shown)


myk002 (approved):
Lgtm. This will let me clean up a number of tests!

myk002 (approved):
Approve for mock.func(). If needed, we can extend this later by letting the return_value param be a table of function when(args...) -> return_value

lethosor:
I was leaning more towards allowing an actual callback in place of https://github.com/DFHack/dfhack/pull/1831/files#diff-0f065c2f1def43f3b29dc03f2f8dada2cd1a359270761e389fddb0f5daf1a4d8R62, but didn't come up with a way to allow e.g. both `callback` and `return_value` args. I guess a separate factory function could work (`mock.func_callback()` or something like that). In any case, this should be adequate for current testing needs.

To avoid conflicting with your work: are you planning to adjust tests to use these? Or is it ok if I do that?

myk002:
> To avoid conflicting with your work: are you planning to adjust tests to use these? Or is it ok if I do that?

Feel free to adjust current tests. I have everything merged at this point so there won't be any conflicts.

lethosor:
Done: https://github.com/DFHack/scripts/commit/55ac6f7269a10a1b3e525da2c3f117675b529289

I opted not to change [this patch](https://github.com/DFHack/scripts/commit/55ac6f7269a10a1b3e525da2c3f117675b529289#diff-8ebb7c772a88f28943ac09901f6413686d323c762e1132d302e6137966a91221R34) for `print()` significantly, because that would require changing checks to e.g. `expect.eq({{'line 1'}, {'line 2'}}, print_wrapper.call_args)` which is somewhat more verbose. Maybe another specialization of `mock.func()` for single-argument functions could be useful later to avoid some of the added boilerplate, if this pattern becomes more common.

PR #1840, opened 2021-04-19
[test] support fortress mode

myk002:
#1690 

- implement navigation function for loading a fortress from the title screen (requires a fortress save to be ready to load).
- navigation function also handles case where a fortress map is already loaded
- ensure we don't try repeatedly to enter a mode that we can't reach (such as getting back to the title screen from fortress mode). failing to enter the mode once will skip all remaining tests in that mode.

This PR combines the VALID_MODES and MODE_NAVIGATE_FNS tables into a new MODES table that contains mode order, state, and related functions

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -166,7 +166,7 @@ local function ensure_fortress(config)
             dfhack.gui.resetDwarfmodeView(true)
             return
         end
-        local scr = dfhack.gui.getCurViewscreen()
+        local scr = dfhack.gui.getCurViewscreen(true)
         if focus_string == 'title' or
                 focus_string == 'dfhack/lua/load_screen' then
             -- qerror()'s on falure
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -27,13 +27,25 @@ Options:
     -d, --test_dir  specifies which directory to look in for tests. defaults to
                     the "hack/scripts/test" folder in your DF installation.
     -m, --modes     only run tests in the given comma separated list of modes.
-                    valid modes are 'none' (test can be run on any screen) and
-                    'title' (test must be run on the DF title screen). if not
-                    specified, no modes are filtered.
+                    see the next section for a list of valid modes. if not
+                    specified, the tests are not filtered by modes.
     -r, --resume    skip tests that have already been run. remove the
                     test_status.json file to reset the record.
+    -s, --save_dir  the save folder to load for "fortress" mode tests. this
+                    save is only loaded if a fort is not already loaded when
+                    a "fortress" mode test is run. if not specified, defaults to
+                    'region1'.
     -t, --tests     only run tests that match one of the comma separated list of
                     patterns. if not specified, no tests are filtered.
+
+Modes:
+
+    none      the test can be run on any screen
+    title     the test must be run on the DF title screen. note that if the game
+              has a map loaded, "title" mode tests cannot be run
+    fortress  the test must be run while a map is loaded. if the game is
+              currently on the title screen, the save specified by the save_dir
+              parameter will be loaded.
 
 Examples:
 
@@ -145,7 +157,7 @@ end
 -- Requires that a fortress game is already loaded or is ready to be loaded via
 -- the "Continue Playing" option in the title screen. Otherwise the function
 -- will time out and/or exit with error.
-local function ensure_fortress()
+local function ensure_fortress(config)
     local focus_string = dfhack.gui.getCurFocus(true)
     for screen_timeout = 1,10 do
         if is_fortress(focus_string) then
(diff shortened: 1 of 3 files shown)


ci/test.lua
```diff
@@ -116,29 +114,87 @@ end
...
+
+local function is_fortress(focus_string)
+    focus_string = focus_string or dfhack.gui.getCurFocus(true)
+    return focus_string == 'dwarfmode/Default'
+end
+
+-- Requires that a fortress game is already loaded or is ready to be loaded via
+-- the "Continue Playing" option in the title screen. Otherwise the function
+-- will time out and/or exit with error.
+local function ensure_fortress()
```
lethosor, line 148: Calling out to the `load-save` script might be a better solution here. It should work from the title screen. I'm not sure if it will be easy to distinguish between failures due to no matching save vs. no saves at all, for instance, but it should work.

Incidentally, does this actually look for `region1`, or just any save? Maybe the region folder name could be a parameter.

myk002: It doesn't look specifically for `region1`, it just selects the first region in the list. I hadn't thought of using `load-save` - we could certainly add a parameter to the test harness to control which region is loaded (only used if one isn't loaded already, of course).

myk002: added `--save_dir` parameter and changed to use `load-save` script to load

ci/test.lua
```diff
@@ -415,12 +482,22 @@ local function filter_tests(tests, config)
...
 
-local function run_tests(tests, status, counts)
+local function run_tests(tests, status, counts, config)
     print(('Running %d tests'):format(#tests))
     for _, test in pairs(tests) do
-        MODE_NAVIGATE_FNS[test.config.mode]()
-        local passed = run_test(test, status, counts)
-        status[test.full_name] = passed and TestStatus.PASSED or TestStatus.FAILED
+        status[test.full_name] = TestStatus.FAILED
+        if MODES[test.config.mode].failed then goto skip end
```
myk002, line 489: Is this desired behavior? Unrunnable tests are recorded as failed? Or should they just not be recorded at all?

ci/test.lua
```diff
@@ -415,12 +482,22 @@ local function filter_tests(tests, config)
...
+        if not MODES[test.config.mode].detect() then
+            local ok, err = pcall(MODES[test.config.mode].navigate, config)
+            if not ok then
+                MODES[test.config.mode].failed = true
+                dfhack.printerr(tostring(err))
+                goto skip
+            end
+        end
+        status[test.full_name] = run_test(test, status, counts) and
```
lethosor, line 498: Could probably replace with `if run_test(...) then status[...] = PASSED`

ci/test.lua
lethosor: The current approach wouldn't print an error for each test, right? My understanding is that saving a `FAILED` status to the status file would just prevent the test from re-running again on the next DF run.

Actually, if status for these tests remain `PENDING`, leaving them untouched would allow for them to run again on future DF runs, which might be a reasonable behavior (e.g. if DF gets stuck on a screen for some reason and the issue goes away when DF restarts).

ci/test.lua
myk002: done

ci/test.lua
myk002: done. I also printed a line for how many tests were in unreachable modes. otherwise they would be completely silently skipped

myk002:
Is this PR good for merge? Until we have a fortress setup for CI, perhaps we could specify `--modes=none,title` in [travis/run-tests.py](https://github.com/DFHack/dfhack/blob/develop/travis/run-tests.py#L76) to prevent fortress-mode tests from breaking CI

lethosor:
I'm really behind, sorry. I think it looks good, but I wanted to try it out and haven't gotten around to it yet. I'll try to get to this tomorrow (this looks like the oldest one so I'll try to get to it first).

Your `--modes` suggestion makes sense to me. CI is passing now, but won't pass if any fortress tests are added until we also have a save to use.

On a semi-related note, I've been working on support for running our test suite on PRs across repos to help with this PR and https://github.com/DFHack/scripts/pull/262. That's still a work in progress at https://github.com/lethosor/dfhack-ci, though.

myk002:
No worries! I'll add in the `--modes`` line to `run_tests.py` too.

Thank you for putting dfhack-ci together! Is there anything I can help with?

lethosor (approved):
Looks good!

lethosor:
> Thank you for putting dfhack-ci together! Is there anything I can help with?

I'm mostly just making (slow) progress on a script to clone DFHack with arbitrary submodule revisions, but I'll let you know if I think of anything and/or when it's ready.

PR #1841, opened 2021-04-25
[blueprint] speed up dig blueprint creation by 70%

myk002:
remove the unneeded cache layer. the cache is for writing. we're just reading. all the cache is doing is adding latency as it makes its copies of map data structures.

generating a 190x190x100 dig blueprint:
before change: 1.7s
after change:  1.0s

the performance gains aren't as important here as the reduced complexity of the algorithm, though. for reasonably-sized blueprints, the time savings are unnoticeable. I'm not sure if this even needs a changelog entry.

this PR also includes minor editing to the header comment and the short description displayed in `ls`. if these are at all controversial, I can move them to a different PR.

### DIFF
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -81,7 +81,8 @@ pair<uint32_t, uint32_t> get_building_size(df::building* b)
 
 char get_tile_dig(int32_t x, int32_t y, int32_t z)
 {
-    df::tiletype_shape ts = tileShape(*Maps::getTileType(x, y , z));
+    df::tiletype *tt = Maps::getTileType(x, y , z);
+    df::tiletype_shape ts = tileShape(tt ? *tt : tiletype::Void);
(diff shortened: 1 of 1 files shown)


lethosor:
LGTM from here - will test locally and wait for CI. My understanding of MapCache is that it sped up map accesses back when DFHack was out-of-process. It's probably unnecessary in most cases now, although there might be a couple convenience functions that are only provided by MapCache that lead people to use it.

myk002:
How does one re-trigger CI?

lethosor:
For BuildMaster, it's not super easy. Pushing another commit is easiest (or just amending an existing commit). I did it by retriggering the webhook under https://github.com/DFHack/dfhack/settings/hooks/21601290 but I think that requires admin privileges. Pinging BenLubar is another way.

myk002:
rebased and force pushed to poke CI

Issue #1842, opened 2021-04-28
Blueprint overhaul plan


myk002:
Design doc and tracking bug for the blueprint plugin overhaul. Public discussion in the [blueprint thread](http://www.bay12forums.com/smf/index.php?topic=176889.msg8274004). This text covers my ideas for:

- Objectives
- Use cases
- Features and commandline API
- GUI design
- Architecture and algorithm
- Project plan

## Objectives
The point of this overhaul is to make blueprint a more productive member of the quickfort ecosystem. Blueprint is the easiest way to create useful, valid quickfort blueprints, especially for people new to quickfort. However, there are a lot of opportunities to make blueprints easier to create, edit, and play back. There are also many ways to improve the quality of the generated blueprints via capturing more map information.

Here are the goals in list form, with tags that we can use later to categorize the project plan tasks:

1. `[Reliability]` Ensure the blueprint -> quickfort workflow works reliably
2. `[Generate]` Make it easier to generate usable blueprints from a map
3. `[Edit]` Make it easier to edit generated blueprints
4. `[Playback]` Make use of DFHack quickfort usability features to make the blueprints easier to play back
5. `[Completeness]` Capture as much map information as possible in the generated blueprints

## Use cases
There are two major groups of users that this work is targeting:

1. Users who want to generate a blueprint and use it with quickfort without having to understand anything about the format
2. Users who want to use generated blueprints as a starting point for creating more advanced custom blueprints

I am catering the default settings to members of group 1, with most of the non-default options intended for members of group 2.

For group 1, the expected workflow is:

1. run `blueprint gui` or `gui/blueprint` to generate a minimal .csv blueprint file
2. run `quickfort gui` or `quickfort run` at some later time to apply it

I want everything to be as simple as possible for this group. Quickfort has the capability to display messages after a blueprint is run, so we can autogenerate a walkthrough for replaying the blueprints later.

Group 2 will use a greater variety of options and spend time editing the blueprints before replaying them with quickfort. The features that make blueprints easier to read, understand, and edit are for this group. Group 2's expected workflow is:

1. run `blueprint gui` or `gui/blueprint`, fiddle with options, and generate pretty-printed .csv or .xlsx files
2. edit the files in a text editor or spreadsheet program
3. save/export the modified blueprints
4. run them through `quickfort`

[...]

PR #1846, opened 2021-05-04
[lua] treat negative numbers as non-options in getopt; fix bugs found with unit tests

myk002:
this allows the positional `depth` parameter in `blueprint` to be negative (in preparation for reworking the plugin in #1842)

also add unit tests, which found a few bugs in the parsing code:
- `--` was not being handled properly as a "all subsequent parameters are non-options" marker
- failed to detect missing required values for long-form options

Is the changelog entry in the right place? should the fixes be under `Fixes` or `Lua`?


lethosor:
Negative numbers have been giving us trouble in `processArgs()` before, so I think this makes sense. I know of some Unix commands that do take numbers as flags (`ping -4`, `traceroute -6`, `ls -1`) but that's probably uncommon enough in DFHack that we can go with your approach.

Thanks for writing the unit tests! This will be first on my list tomorrow.

FYI: GitHub is having [intermittent issues](https://www.githubstatus.com/) that have caused both GitHub Actions and Buildmaster to fail, so CI might fail here. Happy to kick it off as needed or just ignore any unrelated failures.

lethosor:
Looks like CI passed after all!

As for the changelog, a bit of a grey area, but since you're working with a library and the primary users are Lua devs, I think putting these changes under "Lua" is fine. (It does affect some scripts, though - but IMO, listing them all in the changelog would be unnecessarily tedious.)

lethosor (approved):

PR #1847, opened 2021-05-04
[tiletypes] fix label misspelling

myk002:
trivial spelling fix. I don't think this needs a changelog entry (but tell me if you think otherwise).


lethosor:
Yeah, not really worth noting either, IMO

PR #1848, opened 2021-05-04
[lua] fix edge detection in Viewport:isVisibleXY()

myk002:
I ran into this while writing `gui/blueprint`. I was using `Viewport:isVisibleXY()` to determine whether the mouse was clicked on the visible game map (and not, for example, the screen frame or the side panel). To my surprise, it came back true every time, even when clicked at what would be tile map coordinate `-1`.

Looking at the code, it seems that the coordinates are being modified before being compared. This looks to me like incorrect behavior -- if you send a coordinate outside of the game map, *shouldn't* `isVisibileXY()` return `false`?

There are other usages of `math.min` and `math.max` in the file, but those usages look correct -- processing *should* stop at the map edges in those cases. Perhaps `isVisibleXY` was copy-pasted from one of these other functions and then modified? This is very old code, and this bug hasn't come up before, so maybe my understanding of what is supposed to happen here is incorrect, though.

### DIFF
--- a/library/lua/gui/dwarfmode.lua
+++ b/library/lua/gui/dwarfmode.lua
@@ -154,10 +154,10 @@ end
 function Viewport:isVisibleXY(target,gap)
     gap = gap or 0
 
-    return math.max(target.x-gap,0) >= self.x1
-       and math.min(target.x+gap,world_map.x_count-1) <= self.x2
-       and math.max(target.y-gap,0) >= self.y1
-       and math.min(target.y+gap,world_map.y_count-1) <= self.y2
+    return target.x - gap >= self.x1
+       and target.x + gap <= self.x2
+       and target.y - gap >= self.y1
+       and target.y + gap <= self.y2
 end
 
 function Viewport:isVisible(target,gap)
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -43,6 +43,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - ``gui.Painter``: fixed error when calling ``viewport()`` method
+- ``guidm.Viewport:isVisibleXY()``: properly return false for coordinates outside of the viewport
 - `xlsxreader`: Added Lua class wrappers for the xlsxreader plugin API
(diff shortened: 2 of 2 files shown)


library/lua/gui/dwarfmode.lua
```diff
@@ -154,10 +154,10 @@ end
 function Viewport:isVisibleXY(target,gap)
     gap = gap or 0
 
-    return math.max(target.x-gap,0) >= self.x1
-       and math.min(target.x+gap,world_map.x_count-1) <= self.x2
-       and math.max(target.y-gap,0) >= self.y1
-       and math.min(target.y+gap,world_map.y_count-1) <= self.y2
```
lethosor, line 160: I think this function was doing something slightly different. I haven't placed my finger on it exactly, though, but if you haven't yet, I'd recommend taking a look at `reveal()` below. I _think_ `isVisibleXY` really means "would `target` be visible within `gap` tiles of all edges of the viewport, after being clamped to a valid map coordinate?". `reveal()` is used to simulate how DF's `CURSOR_` movement keys work in `simulateCursorMovement()`, I think - DF only recenters the viewport if the cursor gets "too close" to the edge of the map (determined by `gap` here?) but also clamps the cursor within the map boundaries (which I think is what the `min()` and `max()` calls here were doing before).

This all assumes that there _is_ a situation where `reveal()` will get a `false` return value from `isVisibleXY()`, though. I think an off-screen target (e.g. `x=0, y=0` when the viewport is in the lower right) would do it, but I haven't checked.

As for what you _should_ use: it sounds to me like you want to determine whether the mouse is over the map segment of the screen, regardless of the position of the map itself within that segment. In that case, `getPanelLayout().map` is probably what you want to check against. I'm not sure if there's a convenient bounds-checking function there - maybe `is_in_rect()` would work?

myk002: Thanks! yes, `gui.is_in_rect(guidm.getPanelLayout().map, dfhack.screen.getMousePos())` is exactly what I was looking for. I'll close this PR.

PR #1849, opened 2021-05-05
[blueprint] Support new commandline syntax and semantics

myk002:
#1842

This is the first stage of the blueprint plugin overhaul.

Depends on #1846 for support for negative depths and #1841 for the MapCache change (though the MapCache commits are included in this PR as well to avoid future merge conflicts)

The `blueprint gui` command has a soft dependency on DFHack/scripts#281. You'll be able to run the `blueprint gui` command, but it won't do anything until the `gui/blueprint` script has something in it.

User visible changes to blueprint:
- `depth` and `name` parameters are now optional (defaulting to `1` and `blueprint` if not specified)
- `depth` can now be negative to indicate that we should iterate through the levels top-down instead of the previous hard-coded bottom-up
- new `--cursor` option for specifying start position (game cursor is not needed if this option is specified)

I tried to touch the original logic as little as possible, but I did add bounds checking for the input and clipping of the translation area at the map edge.

All docs are updated as well.

**question:**
I kept the lua functions that the blueprint plugin previously exported (`dig()`, `place()`, `build()`, and `query()`) though personally I'm not convinced they add any value over just running `blueprint` via `run_command`. In fact, the new implementations of these functions are just a wrapper around `run_command`. I don't see any callers in the repo, but I kept the functions for the sake of backward compatibility.

I should note, though, that I'm planning on changing the semantics of the phases in the future, including splitting the current `query` phase into `config` and `rooms`, so I don't think perfect backward compatibility will be feasible. Should I keep those functions or just remove them?

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -46,38 +46,40 @@ selected interactively with the ``blueprint gui`` command or, if the GUI is not
 used, starts at the active cursor location and extends right and down for the
 requested width and height.
 
-Usage:
+Usage::
 
     blueprint <width> <height> [<depth>] [<name> [<phases>]] [<options>]
     blueprint gui [<name> [<phases>]] [<options>]
 
 Examples:
 
-blueprint gui
+``blueprint gui``
     Runs `gui/blueprint`, the interactive blueprint frontend, where all
     configuration for a ``blueprint`` command can be set visually and
     interactively.
 
-blueprint gui bedrooms dig -\-cursor 108,100,150
-    Starts ``gui/blueprint`` with some configuration options preset to custom
-    values.
-
-blueprint 30 40 bedrooms
+``blueprint 30 40 bedrooms``
     Generates blueprints for an area 30 tiles wide by 40 tiles tall, starting
-    from the active cursor on the current z-level. Output is written to the
-    ``bedrooms.csv`` file in the ``blueprints`` directory.
+    from the active cursor on the current z-level. Output is written to files
+    with names matching the pattern ``bedrooms-PHASE.csv`` in the ``blueprints``
+    directory.
+
+``blueprint 30 40 bedrooms dig --cursor 108,100,150``
+    Generates only the ``bedrooms-dig.csv`` file from the previous example, and
+    the blueprint start coordinate is set to a specific value instead of using
+    the in-game cursor position.
 
 Positional Parameters:
 
-:width:   Width of the area (in tiles) to translate.
-:height:  Height of the area (in tiles) to translate.
-:depth:   Number of z-levels to translate. Positive numbers go *up* from the
-          cursor and negative numbers go *down*. Defaults to 1 if not specified,
-          indicating that the blueprint should only include the current z-level.
-:name:    Base name for blueprint files created in the ``blueprints`` directory.
-          If no name is specified, "blueprint" is used by default. The string
-          must contain some characters other than numbers so the name won't be
-          confused with the optional ``depth`` parameter.
+:``width``:   Width of the area (in tiles) to translate.
+:``height``:  Height of the area (in tiles) to translate.
+:``depth``:   Number of z-levels to translate. Positive numbers go *up* from the
+    cursor and negative numbers go *down*. Defaults to 1 if not specified,
+    indicating that the blueprint should only include the current z-level.
+:``name``:    Base name for blueprint files created in the ``blueprints``
+    directory. If no name is specified, "blueprint" is used by default. The
+    string must contain some characters other than numbers so the name won't be
+    confused with the optional ``depth`` parameter.
 
 Phases:
 
@@ -85,21 +87,21 @@ If you want to generate blueprints only for specific phases, add their names to
 the commandline, anywhere after the blueprint base name. You can list multiple
 phases; just separate them with a space.
 
-:dig:    Generate quickfort ``#dig`` blueprints.
-:build:  Generate quickfort ``#build`` blueprints for constructions and
-         buildings.
-:place:  Generate quickfort ``#place`` blueprints for placing stockpiles.
-:query:  Generate quickfort ``#query`` blueprints for configuring rooms.
+:``dig``:    Generate quickfort ``#dig`` blueprints.
+:``build``:  Generate quickfort ``#build`` blueprints for constructions and
+    buildings.
+:``place``:  Generate quickfort ``#place`` blueprints for placing stockpiles.
+:``query``:  Generate quickfort ``#query`` blueprints for configuring rooms.
 
 If no phases are specified, all blueprints are created.
 
 Options:
 
-:-c, -\-cursor <x>,<y>,<z>:
+:``-c``, ``--cursor <x>,<y>,<z>``:
     Use the specified map coordinates instead of the current cursor position for
     the upper left corner of the blueprint range. If this option is specified,
     then an active game map cursor is not necessary.
-:-h, -\-help:
+:``-h``, ``--help``:
     Show command help text.
 
 .. _remotefortressreader:
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -37,22 +37,70 @@ For more information, see `the full Stonesense README <stonesense>`.
 
 blueprint
 =========
-Exports a portion of your fortress into QuickFort style blueprint files.
+The ``blueprint`` command exports the structure of a portion of your fortress in
+a blueprint file that you (or anyone else) can later play back with `quickfort`.
 
-Usage::
+Blueprints are ``.csv`` or ``.xlsx`` files created in the ``blueprints``
+subdirectory of your DF folder. The map area to turn into a blueprint is either
+selected interactively with the ``blueprint gui`` command or, if the GUI is not
+used, starts at the active cursor location and extends right and down for the
+requested width and height.
 
-    blueprint <x> <y> <z> <name> [dig] [build] [place] [query]
+Usage:
 
-Options (If only region and name are given, export all):
+    blueprint <width> <height> [<depth>] [<name> [<phases>]] [<options>]
+    blueprint gui [<name> [<phases>]] [<options>]
 
-:x,y,z:     Size of map area to export
-:name:      Name of export files
-:dig:       Export dig commands to "<name>-dig.csv"
-:build:     Export build commands to "<name>-build.csv"
-:place:     Export stockpile commands to "<name>-place.csv"
-:query:     Export query commands to "<name>-query.csv"
+Examples:
 
-Goes very well with `quickfort`, for re-importing.
+blueprint gui
+    Runs `gui/blueprint`, the interactive blueprint frontend, where all
+    configuration for a ``blueprint`` command can be set visually and
+    interactively.
+
+blueprint gui bedrooms dig -\-cursor 108,100,150
+    Starts ``gui/blueprint`` with some configuration options preset to custom
+    values.
+
+blueprint 30 40 bedrooms
+    Generates blueprints for an area 30 tiles wide by 40 tiles tall, starting
+    from the active cursor on the current z-level. Output is written to the
+    ``bedrooms.csv`` file in the ``blueprints`` directory.
+
+Positional Parameters:
+
+:width:   Width of the area (in tiles) to translate.
+:height:  Height of the area (in tiles) to translate.
+:depth:   Number of z-levels to translate. Positive numbers go *up* from the
+          cursor and negative numbers go *down*. Defaults to 1 if not specified,
+          indicating that the blueprint should only include the current z-level.
+:name:    Base name for blueprint files created in the ``blueprints`` directory.
+          If no name is specified, "blueprint" is used by default. The string
+          must contain some characters other than numbers so the name won't be
+          confused with the optional ``depth`` parameter.
+
+Phases:
+
+If you want to generate blueprints only for specific phases, add their names to
+the commandline, anywhere after the blueprint base name. You can list multiple
+phases; just separate them with a space.
+
+:dig:    Generate quickfort ``#dig`` blueprints.
+:build:  Generate quickfort ``#build`` blueprints for constructions and
+         buildings.
+:place:  Generate quickfort ``#place`` blueprints for placing stockpiles.
+:query:  Generate quickfort ``#query`` blueprints for configuring rooms.
+
+If no phases are specified, all blueprints are created.
+
+Options:
+
+:-c, -\-cursor <x>,<y>,<z>:
+    Use the specified map coordinates instead of the current cursor position for
+    the upper left corner of the blueprint range. If this option is specified,
+    then an active game map cursor is not necessary.
+:-h, -\-help:
+    Show command help text.
 
 .. _remotefortressreader:
 
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -8,9 +8,11 @@
 #include <algorithm>
 #include <sstream>
 
-#include <Console.h>
-#include <PluginManager.h>
+#include "Console.h"
+#include "DataDefs.h"
+#include "DataIdentity.h"
 #include "LuaTools.h"
+#include "PluginManager.h"
 #include "TileTypes.h"
 
 #include "modules/Buildings.h"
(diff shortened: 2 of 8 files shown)


myk002:
note that unit tests will fail until #1846 is merged

myk002:
> note that unit tests will fail until #1846 is merged

latest develop merged. tests should pass now

plugins/blueprint.cpp
```diff
@@ -27,22 +29,63 @@
...
+
+    // if not autodetecting, which phases to output
+    bool dig   = false;
+    bool build = false;
+    bool place = false;
+    bool query = false;
+
+    static struct_identity _identity;
+};
+static const struct_field_info blueprint_options_fields[] = {
```
lethosor, line 72: This is an interesting approach that I felt like I should comment on. This looks good for type-safety! Long-term, I think we should generate type definitions like these with codegen somehow instead of manually, since there could be failures down the road if these field definitions aren't kept in sync with the structure definition. (using `offsetof()` to calculate offsets should help guard against most issues... but as a counterpoint, I'm not sure if the field definitions need to be ordered by offset, so rearranging fields could potentially have unintended consequences, for example).

Am I correct in assuming that you didn't define this in [PluginStatics.cpp](https://github.com/DFHack/dfhack/blob/develop/library/PluginStatics.cpp) because the objects in question are short-lived and can't exist if the plugin is unloaded? (I was playing around with this and `opts` tended to end up at the same stack address every time!) I think this is safe, as long as you're sure that no Lua code retains references to `opts` (hopefully the garbage collector doesn't...). 

In the interests of avoiding duplication, you could also consider a macro like [`FLD()`](https://github.com/DFHack/dfhack/blob/41ff2c2e6c612814e09d03704f00bcdfb613db52/library/DataStaticsFields.cpp#L62), but the existing definition would admittedly be difficult to include.

lethosor (commented):
Looks good so far - I like the single path for command processing, and the use of Lua getopt. Still haven't played around with this in fortress mode, but from static review it looks good.

plugins/blueprint.cpp
myk002: +1 for generating the `struct_field_info`. As far as I could tell, this is the first time a marshal-able type has been manually defined in DFHack, so I guess the need for a generator just never came up before.

You are correct as to why I didn't put this in `PluginStatics`. The stack-allocated options struct doesn't live beyond the `blueprint` command execution so I figured I'd keep all the definitions together in the `blueprint.cpp` file instead of gaining a consistent identity object, but splitting the object definitions into `PluginStatics`.

Yeah, I manually expanded the #defines that are used in the generated `struct_field_info`s when I was figuring out how to do this. Not using the existing (but inaccessible) macros felt dirty, but copying the macros into `blueprint.cpp` felt somehow dirtier : p  I didn't really want to get into refactoring `DataStatics*` in this PR, but that might be the way to go longer term.

docs/Plugins.rst
```diff
@@ -37,22 +37,70 @@ For more information, see `the full Stonesense README <stonesense>`.
...
+Blueprints are ``.csv`` or ``.xlsx`` files created in the ``blueprints``
+subdirectory of your DF folder. The map area to turn into a blueprint is either
+selected interactively with the ``blueprint gui`` command or, if the GUI is not
+used, starts at the active cursor location and extends right and down for the
+requested width and height.
+
+Usage:
+
+    blueprint <width> <height> [<depth>] [<name> [<phases>]] [<options>]
+    blueprint gui [<name> [<phases>]] [<options>]
```
lethosor, line 52: This should probably be a code block so that it wraps better. Looks like this currently:
image


Suggested fix:
```suggestion
Usage::

    blueprint <width> <height> [<depth>] [<name> [<phases>]] [<options>]
    blueprint gui [<name> [<phases>]] [<options>]
```

docs/Plugins.rst
```diff
@@ -37,22 +37,70 @@ For more information, see `the full Stonesense README <stonesense>`.
...
+    configuration for a ``blueprint`` command can be set visually and
+    interactively.
 
-    blueprint <x> <y> <z> <name> [dig] [build] [place] [query]
+blueprint gui bedrooms dig -\-cursor 108,100,150
+    Starts ``gui/blueprint`` with some configuration options preset to custom
+    values.
 
-Options (If only region and name are given, export all):
+blueprint 30 40 bedrooms
```
lethosor, line 65: Perhaps an example of phases + options being used _without_ the `gui` argument would also be useful.

docs/Plugins.rst
```diff
@@ -37,22 +37,70 @@ For more information, see `the full Stonesense README <stonesense>`.
...
+If no phases are specified, all blueprints are created.
+
+Options:
 
-Goes very well with `quickfort`, for re-importing.
+:-c, -\-cursor <x>,<y>,<z>:
+    Use the specified map coordinates instead of the current cursor position for
+    the upper left corner of the blueprint range. If this option is specified,
+    then an active game map cursor is not necessary.
+:-h, -\-help:
```
lethosor, line 102: For completeness, since I suggested inline code for the examples earlier: it would work here too, but is more verbose (although it would avoid the need to escape the second `-`), and I don't think it's as necessary since it's consistent with how other parameters are formatted. Specifically, I think this looks good, but we really don't have a formatting standard for this, and 8 backticks are rather a lot:
```
:``-h``, ``--help``:
```

plugins/blueprint.cpp
```diff
@@ -592,179 +612,200 @@ command_result do_transform(DFCoord start, DFCoord end, string name, uint32_t ph
...
-    uint32_t option = 0;
-    if (parameters.size() == 4)
+
+    // start coordinates can come from either the commandline or the map cursor
+    DFCoord start(options.start);
+    if (options.start.x == -30000)
     {
-        option = DIG | BUILD | PLACE | QUERY;
+        int32_t x, y, z;
+        if (!Gui::getCursorCoords(x, y, z))
```
lethosor, line 771: Could this be simplified to set fields of `start` directly? `getCursorCoords()` does modify its parameters even on failure, but you're returning immediately in that case anyway.
```suggestion
        if (!Gui::getCursorCoords(start.x, start.y, start.z))
```

plugins/blueprint.cpp
```diff
@@ -592,179 +612,200 @@ command_result do_transform(DFCoord start, DFCoord end, string name, uint32_t ph
...
+    }
+
+    for (const string &param : parameters)
+        Lua::Push(L, param);
+
+    if (!Lua::SafeCall(out, L, parameters.size(), 0))
+        return false;
+
+    return true;
+}
```
lethosor, line 729: Is your goal with this function to launch `gui/blueprint` safely? If so, the poorly-named `Core::setHotkeyCmd()` should do the trick too. `autogems.cpp` and `hotkeys.cpp` use it, for instance. I do see that `do_gui()` introduces a 1-frame delay, though - I'm not sure what purpose that serves.

plugins/blueprint.cpp
```diff
@@ -562,27 +586,23 @@ string get_tile_query(df::building* b)
...
-command_result do_transform(DFCoord start, DFCoord end, string name, uint32_t phases, std::ostringstream &err)
+command_result do_transform(const DFCoord &start, const DFCoord &end,
+                            const blueprint_options &options,
+                            std::ostringstream &err)
 {
     ofstream dig, build, place, query;
 
-    std::string basename = "blueprints/" + name;
-
-#ifdef _WIN32
```
lethosor, line 579: Just wanted to confirm that this was intentional. I see you added the normalization in https://github.com/DFHack/dfhack/commit/c19fc1f34903f15e915fa730af6d9ed53f5fc93c#diff-ef1e8e7bfb935197babf6453e5f25e4a208e546ef64fee459c68e03c5df8ebcdR577-R580, presumably for parity with ` Filesystem::mkdir_recursive()`, and likely not a lot of people rely on putting blueprints in subfolders named with backslashes, but there is a slight change in behavior here: given `a/b\c`, the splitting logic (`find_last_of("/")`) would produce `a` as the directory name before passing it to mkdir_recursive. Not a huge deal, IMO.

plugins/blueprint.cpp
```diff
@@ -592,179 +612,200 @@ command_result do_transform(DFCoord start, DFCoord end, string name, uint32_t ph
         return CR_FAILURE;
     }
 
-    if (phases & QUERY)
+    if (options.auto_phase || options.query)
```
lethosor, line 615: I may have missed something, but it looks to me like every check for `auto_phase` is of the form `auto_phase ||  some_mode`. Would it make more sense to just enable every mode if no mode is specified (or whatever triggers `auto_phase` currently)?

lethosor (commented):
Looks great from a static review! I got sucked into details (e.g. documentation formatting) a bit, but those are minor things. Will try to test in more detail this weekend; let me know if you have any areas of concern in mind.

docs/Plugins.rst
myk002: actually, let me repurpose the gui example for this. I haven't implemented option parsing for `gui/quickfort` yet, and if we're going to cut a release after this merge, I don't want the docs to be misleading.

docs/Plugins.rst
myk002: done. I did all positional params as well for consistency.

plugins/blueprint.cpp
myk002: annoyingly, no (unless I do some casting). `start` is a `df::coord` which has `int16_t` fields. `getCursorCoords` takes `int32_t` params. I could cast, but I prefer to avoid that if possible. I've found that casually flirting with casting can cause strange and terrible bugs.

plugins/blueprint.cpp
myk002: Thank you for the pointer to `setHotkeyCmd()` I.. never would have thought to look there ; P

the idea behind the 1-frame delay (and it might not even achieve this goal) was to let the plugin call exit before I launched another related command. I haven't given any thought to re-entrancy, and I didn't want to create a code path that could conceivably call the blueprint plugin while the blueprint plugin was still processing an existing call. For example, if someone ran `blueprint gui --cursor=10,10,10` and immediately sent another `Enter` key to the df ui, it would immediately call blueprint again on the same frame. again, I haven't really analyzed the code behavior in this situation, but I figured I could head off any potential issues by just not making it possible.

plugins/blueprint.cpp
myk002: the normalization of name is now handled in the lua code, so it's guaranteed to be in proper format by the time it gets here - see the `parse_positionals()` function

plugins/blueprint.cpp
myk002: this is a forward-looking design decision. once I change how the blueprint data is buffered and output, only blueprints that actually have data will be written. the user will be able to force creation of otherwise empty blueprints (e.g. for getting a pretty-printed blank spreadsheet for editing purposes) by explicitly requesting the phase. so for that I'll need the original values of both auto_phase and the flag for the phase itself.

plugins/blueprint.cpp
myk002: I made the change, but I'm curious about what you think about the possibility of having two blueprint calls on the stack.

docs/Plugins.rst
myk002: done. thanks!

docs/Plugins.rst
```diff
@@ -37,22 +37,70 @@ For more information, see `the full Stonesense README <stonesense>`.
...
+requested width and height.
+
+Usage:
+
+    blueprint <width> <height> [<depth>] [<name> [<phases>]] [<options>]
+    blueprint gui [<name> [<phases>]] [<options>]
+
+Examples:
+
+blueprint gui
```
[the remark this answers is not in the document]
myk002: done.

myk002:
Need to update the unit tests after deleting do_gui. I'll get that tomorrow

myk002:
> Need to update the unit tests after deleting do_gui. I'll get that tomorrow

done

plugins/blueprint.cpp
myk002: to clarify this a bit, I don't think there's a specific path in `blueprint` or `gui/blueprint` that would cause a problem (though I can't guarantee that there will *never* be a problematic path). I'm just not sure about the plugin infrastructure and I'm wary of reentrancy in general in code that was not designed to be multithreaded.

lethosor: > the idea behind the 1-frame delay (and it might not even achieve this goal) was to let the plugin call exit before I launched another related command

I _thought_ `setHotkeyCmd()` handled this by queueing up the command to be executed on the next tick. Looking over it again, it definitely executes commands on a different thread, at least - it's intended to handle hotkeys bound with the `keybinding` command, which are detected in `SDL_PollEvent()`, where it isn't safe to invoke commands directly.

lethosor: I tried making `gui/blueprint` call a `blueprint` command immediately when run, and that seems to run consistently after the first call to `blueprint()` has returned. Seems that there's some delay, at least, even if not a full tick.

myk002: if it's executed in a different thread, reentrancy wouldn't be an issue.. but then we have the spectre of concurrency. Are there any safeguards against a hotkey command executing while an interactive command is running? Although C++-11 has thread-safe initialization of local statics, I've seen global static variables sprinkled around the code with no mutex guards.

I think this PR is probably good with its usage of `setHotkeyCmd()`, but I'll file a bug for an audit/migration of global statics.

myk002: > I tried making `gui/blueprint` call a `blueprint` command immediately when run, and that seems to run consistently after the first call to `blueprint()` has returned. Seems that there's some delay, at least, even if not a full tick.

that's reassuring. I'll still file the bug about global statics, though, since it still seems like a good idea to clean those up

plugins/blueprint.cpp
lethosor: Oh, thanks, I missed that. Casting references is just as dangerous as casting pointers, and if it compiled at all, it would end up clobbering memory since the function expects 4-byte parameters.

Maybe templating `getCursorCoords` or adding an overload for `df::coord` would be useful (but not required).

plugins/blueprint.cpp
lethosor: Oh, thanks! Missed that.

plugins/blueprint.cpp
lethosor: Makes sense - I almost asked about the absence of logic to handle that.

plugins/blueprint.cpp
lethosor: I _thought_ that hotkey commands ran with a CoreSuspender, but it looks like I might have been wrong about that. Maybe it's the responsibility of the command itself to do that. Maybe just moving the CoreSuspender creation earlier in blueprint() would give more complete thread-safety.

plugins/blueprint.cpp
myk002: ah, yeah, I can do that.
done.

plugins/blueprint.cpp
myk002: good idea. done.

plugins/blueprint.cpp
```diff
@@ -755,16 +756,12 @@ command_result blueprint(color_ostream &out, vector<string> &parameters)
     DFCoord start(options.start);
     if (options.start.x == -30000)
```
lethosor, line 757: Just noticed that `df::coord` has an [`isValid()` method](https://github.com/DFHack/dfhack/blob/41ff2c2e6c612814e09d03704f00bcdfb613db52/library/include/df/custom/coord.methods.inc#L6) that does this too. I don't know how widely it's used - I know there are a _lot_ of hardcoded checks for `-30000` too - but if you end up making other changes, this could be something to use as well.

(This PR is still on my list. Hopefully tomorrow!)

plugins/blueprint.cpp
```diff
@@ -592,179 +612,185 @@ command_result do_transform(DFCoord start, DFCoord end, string name, uint32_t ph
...
-    DFCoord end (x + stoi(parameters[0]), y + stoi(parameters[1]), z + stoi(parameters[2]));
-    uint32_t option = 0;
-    if (parameters.size() == 4)
+
+    // start coordinates can come from either the commandline or the map cursor
+    DFCoord start(options.start);
+    if (options.start.x == -30000)
     {
-        option = DIG | BUILD | PLACE | QUERY;
+        if (!Gui::getCursorCoords(options.start))
```
lethosor, line 759: I think both instances of `options.start` here should be `start`. As-is, `getCursorCoords()` is modifying `options.start`, but `start` is used later in the function. Specifically, the unmodified `start` fails the `isValidTilePos()` check below, so I always got the following error:
```
Invalid start position: -30000,-30000,-30000
```
until this change:
```diff
diff --git a/plugins/blueprint.cpp b/plugins/blueprint.cpp
index fc8ac5fa..2158f408 100644
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -754,9 +754,9 @@ command_result blueprint(color_ostream &out, vector<string> &parameters)
 
     // start coordinates can come from either the commandline or the map cursor
     DFCoord start(options.start);
-    if (options.start.x == -30000)
+    if (start.x == -30000)
     {
-        if (!Gui::getCursorCoords(options.start))
+        if (!Gui::getCursorCoords(start))
         {
             out.printerr("Can't get cursor coords! Make sure you specify the"
                     " --cursor parameter or have an active cursor in DF.\n");
```

plugins/lua/blueprint.lua
```diff
@@ -1,14 +1,161 @@
...
+               :format(args[argidx]))
+    end
+    argidx = argidx + 1
+    -- normalize paths to forward slashes
+    opts.name = name:gsub(package.config:sub(1,1), "/")
+
+    local auto_phase = true
+    local phase = args[argidx]
+    while phase do
+        if valid_phases[phase] then
```
lethosor, line 76: Should we do something if the argument isn't valid? I tried `blueprint 1 1 build foo` and it wrote four `build-*.csv` files, ignoring the `foo` argument. Both ``blueprint 1 1 build foo place` and ``blueprint 1 1 build place foo` write just `build-place.csv`. (I know I'm using `build` as the blueprint name, yeah... I wasn't actually confused, just wanted to see what would happen).

lethosor (commented):
Looks really good! A couple minor things I noticed from testing; not sure if you've already fixed them elsewhere.

plugins/blueprint.cpp
myk002: thank you for catching this. obviously, I did not test enough after migrating to the new `getCursorCoords` API : /

done.

plugins/lua/blueprint.lua
myk002: this makes sense. I'm not sure why I decided to just ignore invalid phases.

done, and updated tests for the behavior change.

lethosor (approved):

PR #1850, opened 2021-05-08
[buildingplan] make isPlanModeEnabled() consider enable_all

myk002:
this lets ``isPlanModeEnabled()`` return true when "enable all" is set. this fixes the `automaterial` plugin interface when building constructions and buildingplan's "enable all" setting is set.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -35,6 +35,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 - `buildingplan`: fixed an issue where planned constructions designated with DF's sizing keys (``umkh``) would sometimes be larger than requested
+- `buildingplan`: make ``isPlanModeEnabled()`` always return true if "enable all" is set. this fixes the `automaterial` plugin interface when building constructions and buildingplan's "enable all" setting is set.
 - `command-prompt`: fixed issues where overlays created by running certain commands (e.g. `gui/liquids`, `gui/teleport`) would not update the parent screen correctly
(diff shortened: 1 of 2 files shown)


docs/changelog.txt
```diff
@@ -35,6 +35,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 - `buildingplan`: fixed an issue where planned constructions designated with DF's sizing keys (``umkh``) would sometimes be larger than requested
+- `buildingplan`: make ``isPlanModeEnabled()`` always return true if "enable all" is set. this fixes the `automaterial` plugin interface when building constructions and buildingplan's "enable all" setting is set.
```
lethosor, line 38: From a user perspective, would "made constructions plannable if planning mode is enabled for all building types" (or "...if the 'enable all' setting is set" be an accurate description?

myk002: Yeah, I struggled a bit with how to phrase this. This change doesn't affect the buildingplan ui at all (it already used the correct detection method). It only affects other lua code that checks to see if plan mode is enabled for a particular building type. I believe automaterial is the only client of that call. How you phrased it would work just fine, I think.

lethosor: Hm, ok, I'd say that it's different from what I suggested, then (since I was thinking it fixed buildingplan itself). I'd probably either categorize this as an automaterial fix or a buildingplan fix that "fixes other plugins like automaterial".

(I'm fine dealing with this, by the way, just wanted to make sure I was describing it correctly - so thanks!)

myk002: Ack. Classifying it as an automaterial fix sgtm. Thanks for editing!

lethosor (approved):

PR #1858, opened 2021-05-25
[blueprint] report output filenames when creating blueprints

myk002:
also refactor so we can get this data from `gui/blueprint`, plus a bit of code hygiene updating.

### DIFF
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -685,11 +685,11 @@ static bool do_transform(const DFCoord &start, const DFCoord &end,
     return true;
 }
 
-static bool get_options(blueprint_options &opts,
+static bool get_options(color_ostream &out,
+                        blueprint_options &opts,
                         const vector<string> &parameters)
 {
     auto L = Lua::Core::State;
-    color_ostream_proxy out(Core::getInstance().getConsole());
     Lua::StackUnwinder top(L);
 
     if (!lua_checkstack(L, parameters.size() + 2) ||
@@ -711,10 +711,9 @@ static bool get_options(blueprint_options &opts,
     return true;
 }
 
-static void print_help()
+static void print_help(color_ostream &out)
 {
     auto L = Lua::Core::State;
-    color_ostream_proxy out(Core::getInstance().getConsole());
     Lua::StackUnwinder top(L);
 
     if (!lua_checkstack(L, 1) ||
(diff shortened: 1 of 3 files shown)


plugins/blueprint.cpp
```diff
@@ -789,8 +798,52 @@ command_result blueprint(color_ostream &out, vector<string> &parameters)
...
+        return 1;
+    }
+
+    return 0;
+}
+
+command_result blueprint(color_ostream &out, vector<string> &parameters)
+{
+    vector<string> files;
+    if (do_blueprint(Core::getInstance().getConsole(), parameters, files))
```
lethosor, line 836: This should technically use the existing output stream, for better behavior when run from other locations (command-prompt, dfhack-run, etc.)
```suggestion
    if (do_blueprint(out, parameters, files))
```

plugins/blueprint.cpp
```diff
@@ -789,8 +798,52 @@ command_result blueprint(color_ostream &out, vector<string> &parameters)
...
+    for (int i = 1; i <= argc; ++i)
+    {
+        const char *s = lua_tostring(L, i);
+        if (s == NULL)
+            luaL_error(L, "all parameters must be strings");
+        argv.push_back(s);
+    }
+
+    vector<string> files;
+    if (do_blueprint(Core::getInstance().getConsole(), argv, files))
```
lethosor, line 824: I think this output stream could use `Lua::GetOutput(L)` for the same reason. Note that this requires the Lua code to have been invoked with `SafeCall()`, and I wasn't able to easily trace the callers of `do_phase()` to figure out whether that's the case.

myk002: Will update the code, but just to reply to your comment, there might not be any callers of the functions that call `do_phase()`. That function group is purely for backwards compatibility with the previous blueprint Lua API. There are no callers in-tree, but there might be someone somewhere who calls them. I was just keeping compatibility on principle.

plugins/blueprint.cpp
myk002: Will fix, but calling getConsole is a paradigm that i've seen (and used) in many other places in the code. I'll be sure to use the passed-in `out` object or `Lua::getOutput()` in the future

lethosor: No problem. It may not be necessary in the Lua case, since a GUI script doesn't really have an output stream other than the console anyway, but if you have an `out` parameter, that's generally what you'd want to go with. I know there are several uses of `getConsole()` in other plugins and core code because passing `out` parameters around was either deemed too messy or just didn't make sense, so there are definitely exceptions.

myk002: done

plugins/blueprint.cpp
myk002: done

lethosor: That'd explain why I couldn't find any callers! Thanks for the update.

lethosor (approved):

PR #1859, opened 2021-05-26
quickfort ecosystem integration test harness

myk002:
and a few initial tests.

Depends on #1853, #1867, and DFHack/scripts#284

The idea here is to have a suite of .csv blueprints (only installed if `BUILD_TESTS` is ON) that get run through `quickfort`, executed with `dig-now` (and later extended with `orders-now` and `build-now` once those scripts are ready), and codified back into blueprint files via `blueprint`. The test harness would then compare the input to the output and fail on discrepancies. This validates behavior in both `quickfort` and `blueprint` and allows us to cover code that is difficult to run with unit tests.

Note that the integration tests are not quite side effect free. An unused area of the map is located, the test is run at that location (digging and building as required), and after the test is done, the tiles are converted back to hidden walls. However, the digging and re-walling process will remove any vein tiles in the area. This can have a slight negative effect if run during a "real" game. Do we want to add a flag that can prevent these types of tests from running? Or can we just declare that fortress-mode tests are not guaranteed to have a net-0 effect on a running game?


lethosor:
> Do we want to add a flag that can prevent these types of tests from running? Or can we just declare that fortress-mode tests are not guaranteed to have a net-0 effect on a running game?

I wouldn't be opposed to a flag, although in cases of a severe-enough bug, _anything_ in a test could end up changing or corrupting a save. I think our general recommendation should be not to save after running fortress mode tests.

test/quickfort/ecosystem.lua
```diff
@@ -0,0 +1,291 @@
...
+            if input_hash ~= output_hash or input_size ~= output_size then
+                -- show diff
+                local input, output =
+                    io.open(input_filepath, 'r'), io.open(output_filepath, 'r')
+                local input_lines, output_lines = {}, {}
+                for l in input:lines() do table.insert(input_lines, l) end
+                for l in output:lines() do table.insert(output_lines, l) end
+                input:close()
+                output:close()
+                expect.table_eq(input_lines, output_lines)
```
lethosor, line 284: This file-comparing logic _might_ be useful to move into `expect.lua`, although I don't expect (heh) to have many tests that compare files, so it can wait for now.

lethosor (approved):

PR #1861, opened 2021-05-27
[resume] invalidate cache when buildings are added programmatically

myk002:
It turns out option 5 from https://github.com/DFHack/scripts/pull/284#issuecomment-848815321 was equivalent to option 1, and they're both cheap and effective.  the `process_jobs` flag is handled internally by df on every frame, so it only stays true for one frame. this allows us to invalidate once and rescan only when required.

no changelog entry since this was never a user-visible problem until DFHack/scripts#284

### DIFF
--- a/plugins/resume.cpp
+++ b/plugins/resume.cpp
@@ -41,6 +41,7 @@ DFHACK_PLUGIN("resume");
 #define PLUGIN_VERSION 0.2
 
 REQUIRE_GLOBAL(gps);
+REQUIRE_GLOBAL(process_jobs);
 REQUIRE_GLOBAL(ui);
 REQUIRE_GLOBAL(world);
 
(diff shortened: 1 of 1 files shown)


lethosor:
> the `process_jobs` flag is handled internally by df on every frame, so it only stays true for one frame

I can't seem to reproduce this behavior when DF is paused:
```
[lua]# df.global.process_jobs=true
[lua]# ~df.global.process_jobs
true
[lua]# ~df.global.process_jobs
true
[lua]# ~df.global.process_jobs
true
[lua]# ~df.global.process_jobs  -- after advancing one tick with '.'
false
```

That said, I'm not _too_ concerned about the performance of `scan_for_suspended_buildings()`.

myk002:
interesting -- I reproduced your output at the lua commandline. When I instrumented the resume plugin to report how often the new if clause was hit, it output exactly once per run of `quickfort --cursor`

aaaand I just tested again, and could not reproduce my earlier results.  it rescans on every frame, as your output would indicate.  well, I agree that it's not something to fuss too much about. a linear scan through buildings while paused isn't that big of a deal

lethosor (approved):

PR #1863, opened 2021-06-02
[reveal] expose revflood core logic to lua as unhideFlood(pos)

myk002:
for reuse by the `dig-now` plugin (#1853)

Refactor is a straight cut-paste with minimal edits. This code could really stand some cleanup (unused vars, unnecessary use of the MapCache layer, forced allocation of all blocks even if they are not being unhidden, etc.), but that can come in a later PR.

questions:
- is `CoreSuspender` required when calling from lua?
- do we need any environment validity checking in the Lua entrypoint (like `Maps::isValid()`)? It doesn't cause any problems if it is called from, for example, the DF title screen. It just doesn't have any effect. Do we need errors to be thrown in that situation?


lethosor:
> * is `CoreSuspender` required when calling from lua?

My understanding is that it depends on how the Lua code was invoked. Lua scripts should already have a suspender acquired (through RunCoreQueryLoop) but Lua code invoked from plugins wouldn't necessarily. I checked a couple older plugins (rename, workflow, burrows), and I only saw one case where a C++ function exposed to Lua used a CoreSuspender (renameBurrow, which is apparently related to it being used as an event). In this case, I think you're fine just using the CoreSuspender in the `revflood` command entrypoint.

> * do we need any environment validity checking in the Lua entrypoint

Not crashing is a plus. Looks to me like `MapCache::testCoord()` fails gracefully on invalid coordinates, which is causing the no-op behavior currently. It's probably more effort than it's worth to pass an error message up the chain to Lua (I wouldn't be opposed to raising a Lua error if no map is loaded, but I don't feel like it's strictly necessary).

myk002:
> In this case, I think you're fine just using the CoreSuspender in the `revflood` command entrypoint.

done. Thanks!

> It's probably more effort than it's worth to pass an error message up the chain to Lua

I think I agree. an errorless noop in invalid situations seems like perfectly valid behavior.

lethosor (approved):

PR #1865, opened 2021-06-06
move fastdwarf's teleport routine to the Units module

myk002:
so it can be reused by `dig-now` and the `teleport` script (which has a much less sophisticated algorithm)


docs/changelog.txt
```diff
@@ -54,6 +54,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``utils.processArgsGetopt()``: now properly handles ``--`` like GNU ``getopt`` as a marker to treat all further parameters as non-options
 - ``utils.processArgsGetopt()``: now detects when required arguments to long-form options are missing
 - `xlsxreader`: added Lua class wrappers for the xlsxreader plugin API
+- added ``dfhack.units.teleport(unit, pos)``
```
lethosor, line 57: FYI, I usually put things like this under the `API` section, since they affect multiple languages

lethosor (approved):
Nice!

PR #1866, opened 2021-06-06
remove outdated info from plant plugin docs

myk002:
the `extirpate` and `immolate` commands were removed some time ago


lethosor (approved):
Good catch! Less important: should we also remove the source code in plants.cpp? It has been comment out for ~7 years at this point.

myk002:
Sure, I can do that in this PR. If we ever need the code again, it will be in the history.

PR #1867, opened 2021-06-07
[tiletypes] add --cursor and --quiet options to support scripted invocation

myk002:
only modifies `tiletypes-here` and `tiletypes-here-point`. The interactive basic `tiletypes` command doesn't really need these options.

I added these options so the `quickfort` ecosystem test harness (#1859) can run `tiletypes-here` commands without having to set the game cursor and without getting console spam.


plugins/lua/tiletypes.lua
```diff
@@ -0,0 +1,29 @@
+local _ENV = mkmodule('plugins.tiletypes')
+
+local utils = require('utils')
+
+local function parse_cursor(opts, arg)
```
lethosor, line 5: Can this be shared with `plugins/lua/blueprint.lua` somehow? Seems that the primary difference is the field name, but the function could likely be refactored to return a `df.coord`.

plugins/lua/tiletypes.lua
```diff
@@ -0,0 +1,29 @@
...
+    if not x then
+        qerror(('invalid argument for --cursor option: "%s"; expected format' ..
+                ' is "<x>,<y>,<z>", for example: "30,60,150"'):format(arg))
+    end
+    opts.cursor.x = tonumber(x)
+    opts.cursor.y = tonumber(y)
+    opts.cursor.z = tonumber(z)
+end
+
+function parse_commandline(opts, ...)
```
lethosor, line 16: Would it be worth trying to pull in a C++ argument-parsing module to avoid plugins delegating it to Lua? My main concern is complexity - overhead is probably insignificant. (Counterpoint: there are some benefits to having a consistent parser across tools, regardless of the language they're written in.)

plugins/tiletypes.cpp
```diff
@@ -53,6 +56,25 @@ using namespace df::enums;
...
+    // if set, then use this position instead of the active game cursor
+    df::coord cursor;
+
+    static struct_identity _identity;
+};
+static const struct_field_info tiletypes_options_fields[] = {
+    { struct_field_info::PRIMITIVE, "help",   offsetof(tiletypes_options, help),   &df::identity_traits<bool>::identity, 0, 0 },
+    { struct_field_info::PRIMITIVE, "quiet",  offsetof(tiletypes_options, quiet),  &df::identity_traits<bool>::identity, 0, 0 },
+    { struct_field_info::SUBSTRUCT, "cursor", offsetof(tiletypes_options, cursor), &df::coord::_identity,                0, 0 },
+    { struct_field_info::END }
```
lethosor, line 73: Reminds me of https://github.com/DFHack/dfhack/pull/1849#discussion_r630710706 - we'll want to keep options in mind to avoid defining these manually in the long term.

plugins/tiletypes.cpp
```diff
@@ -722,27 +745,40 @@ command_result executePaintJob(color_ostream &out)
...
+    }
+    else
+    {
+        int32_t x = 0, y = 0, z = 0;
+        if (!Gui::getCursorCoords(x,y,z))
+        {
+            out.printerr("Can't get cursor coords! Make sure you have a cursor active in DF or specify the --cursor option.\n");
+            return CR_FAILURE;
+        }
+        cursor = DFCoord(x, y, z);
```
lethosor, line 771: Possible simplification:

```c++
cursor = Gui::getCursorPos();
if (!cursor.isValid()) {
  // fail
}
```

plugins/tiletypes.cpp
```diff
@@ -722,27 +745,40 @@ command_result executePaintJob(color_ostream &out)
...
     {
         out.printerr("Map is not available!\n");
         return CR_FAILURE;
     }
+
+    uint32_t x_max = 0, y_max = 0, z_max = 0;
     Maps::getSize(x_max, y_max, z_max);
 
-    if (!Gui::getCursorCoords(x,y,z))
+    DFCoord cursor;
```
lethosor, line 758: I might use `df::coord` here just because it's the newer name (and maybe we should deprecate DFCoord...)

plugins/lua/tiletypes.lua
myk002: Is delegating arg parsing to Lua a bad thing? String parsing isn't exactly the strong suit of C++. In my experience, it's very verbose and very error-prone. There is toil with the `_identity` class and funky-looking code that calls Lua, true, and I can see how that could be off-putting, but the power and flexibility of Lua for arg parsing is a powerful draw. Since this use case is fairly well-defined, perhaps we could wrap it in a friendlier API?

plugins/lua/tiletypes.lua
myk002: yeah, let me put a common implementation in `utils.parseCoords()`. well, actually it might be better to use the implementation from `dig-now`, which supports "here" as a coordinate. That pulls in guidm, though, for the cursor position, which would be a circular dependency. Perhaps a new file for common argument parsing functions? `library/lua/argparse.lua`? And we can migrate the `processArgs*` functions there as well.

plugins/tiletypes.cpp
myk002: c++17 has some static reflection capabilities, such as iterating over fields. I have no experience with them, but the brief internet research I did indicates that we might be able to use it to simplify the identity struct generation. Is changing our target C++ standard an option?

I also wonder if we could do something in the current c++11 with templated functions (with specializations for primitive types). That's a little vague, but I could spend some time exploring that idea.

plugins/tiletypes.cpp
myk002: is it? from its usage in Maps, I had thought that DFCoord was the preferred spelling. I'll use `df::coord` in the future.

lethosor: In general, names in the `df::` namespace are newer than others (like `t_engraving` in Engravings.h). From a [blame](https://github.com/DFHack/dfhack/blame/3a7306eb03dc9504e7f2e0a5f0ad688d8c23013d/library/include/modules/Maps.h#L93), DFCoord appears to have been [replaced](https://github.com/DFHack/dfhack/commit/cc510a2c4b9887b885c939419bf67147ee997ac6) a while ago, but evidently the old name stuck around in some places (and maybe was even reintroduced in other new code). Hard to say. Since there are so few external plugins these days, we could probably stand to deprecate and eventually remove the old names for things now.

plugins/lua/tiletypes.lua
lethosor: Yeah, avoiding complexity is a good point, so that's another argument in favor besides the consistency-across-languages argument. If we want plugins to be able to easily modify their own custom structures from Lua, that led to my other comment at https://github.com/DFHack/dfhack/pull/1867#discussion_r661124795. This is fine now if we're willing to put in the manual effort to maintain the field definitions for those structures (just a struct of arguments in this case).

plugins/tiletypes.cpp
lethosor: We're unfortunately tied to the MSVC 2015 compiler on Windows, so even C++14 support is not great. (And with fancy C++ features come fancy error messages, slower compilation times, etc... although simplicity on our end would be good.)

I guess my vision is to have plugins be able to define their own types (in XML?) and have codegen produce metadata for them. It would be complicated to get everything linked in the right libraries, but we do something sort of similar with plugin-specific protobufs.

plugins/lua/tiletypes.lua
lethosor: We probably want to keep `processArgs*` in utils.lua (at least as stubs) for compatibility, so I'm not sure how best to resolve the circular import. What about lazily requiring `guidm` for the cursor position check (`require('gui.dwarfmode').getCursorPos()`)?

lethosor: (Open to also moving argument parsing to `argparse.lua` as long as scripts using `processArgs*` don't need to be updated)

plugins/tiletypes.cpp
myk002: Ack. For my part, I was using it because the Maps API used it, and because it was slightly shorter : )

plugins/lua/tiletypes.lua
myk002: done. this PR now depends on #1897

lethosor (approved):

PR #1868, opened 2021-06-07
ensure extents are populated for civzones and stockpiles

### LINKED ISSUE #1817: Improvements to dfhack.buildings.constructAbstract functionality has broken the code unintentionally in some cases
Improvement to the civ zone handling code has, unfortunately, broken it in some cases. The reason is that some civ zones require the "room" struct to be filled out regardless of whether there are any blocked tiles in it or not, while the code assumes the struct isn't needed if there are no holes.

I'm trying to use a script that worked fine with LNP 0.47.04-r04 but with 0.47.05 I run into trouble. The script is supposed to create gathering zones on trees that are to be picked.

The code in Buildings::checkFreeTiles does not set up any "room" data unless there are obstructions that causes the room to have forbidden/blocked tiles in it. Unfortunately, that logic doesn't seem to match how DF handles zones in all cases. When I set up a gathering zone, a pasture, or just a zone without specifying any functionality (leaving the meeting flag set, as it's set by default), the zone generated by DF has the room parameters filled out (I've made sure that no trees are inside the zones). Thus, it seems we have to identify the set of zones that do generate "room" data regardless of whether there are any obstructions or not, as well as those that leave this data unchanged when not needed (unless DF can handle that it's always specified even when redundant).

The reason my script worked earlier is that it had to post process the gathering zones to include the tree that always was in the middle of the zone, and thus there always was a blocked tile according to the earlier logic. The current logic, however, correctly determines that there are no tiles that block the zones, but uses the old logic to draw the conclusion that no "room" data is needed.

Edit: Observations from random save
Wagon, no room
Workshop, no room (but may not mean all of them)
Door, no room
Table, with & without room info (presumably defining room and "just" built)
Chair, no room
Armor stand, room info
bed, with & without

Edit 2:
It seems Civzones of the type ActivityZone always have to have the "room" component filled out. I suspect that requirement is specific for that kind of Civzone. In fact, I don't really understand what the other Civzones are for. Some, or quite possibly all of them, seem to be for adventure mode only. Some of them seem to be copies of fortress mode things that are NOT Civzones, such as stockpiles (which are stockpiles in fortress mode, not civzones), temples (which are activity zones rather than dedicated temple zones in fortress mode, although they're still civ zones), etc.

myk002:
fixes #1817

I hadn't run into this with `quickfort` since `quickfort` always initializes the room structure itself.

### DIFF
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -587,6 +587,18 @@ bool Buildings::getCorrectSize(df::coord2d &size, df::coord2d &center,
     }
 }
 
+static void init_extents(df::building_extents *ext, const df::coord &pos,
+                         const df::coord2d &size)
+{
+    ext->extents = new df::building_extents_type[size.x*size.y];
+    ext->x = pos.x;
+    ext->y = pos.y;
+    ext->width = size.x;
+    ext->height = size.y;
+
+    memset(ext->extents, 1, size.x*size.y);
+}
+
 bool Buildings::checkFreeTiles(df::coord pos, df::coord2d size,
                                df::building_extents *ext,
                                bool create_ext,
@@ -640,13 +652,7 @@ bool Buildings::checkFreeTiles(df::coord pos, df::coord2d size,
 
                 if (!ext->extents)
                 {
-                    ext->extents = new df::building_extents_type[size.x*size.y];
-                    ext->x = pos.x;
-                    ext->y = pos.y;
-                    ext->width = size.x;
-                    ext->height = size.y;
-
-                    memset(ext->extents, 1, size.x*size.y);
+                    init_extents(ext, pos, size);
                     etile = getExtentTile(*ext, tile);
                 }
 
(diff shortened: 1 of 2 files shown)


quietust:
If we're actually calling these "abstract" buildings, we should probably change that because `abstract_buildingst` is already a thing in Dwarf Fortress (used for keeping track of structures in worldgen sites like human towns, castles, and tombs).

myk002:
changed the PR title to reduce confusion, but the `Buildings::constructAbstract` API may need to be eventually renamed

library/modules/Buildings.cpp
```diff
@@ -670,10 +676,17 @@ std::pair<df::coord,df::coord2d> Buildings::getSize(df::building *bld)
     return std::pair<df::coord,df::coord2d>(pos, df::coord2d(bld->x2+1,bld->y2+1) - pos);
 }
 
-static bool checkBuildingTiles(df::building *bld, bool can_change)
+static bool checkBuildingTiles(df::building *bld, bool can_change,
+                               bool force_extents = false)
 {
     auto psize = Buildings::getSize(bld);
 
+    if (force_extents && !bld->room.extents)
```
lethosor, line 684: Given that only one caller sets `force_extents` currently, would it make sense to split this bit into a separate function? Seems a bit out of place in a `check` function to me. Granted, it is internal-only, and it looks like `checkFreeTiles()` can (or could?) modify buildings as well.

lethosor (approved):

library/modules/Buildings.cpp
myk002: I don't quite follow your meaning here. Are you suggesting moving this `if` statement to `constructAbstract()` or making it independently callable external to the `Buildings` module?

Some context that may or may not be relevant to your statement: the initialization code was already in `checkFreeTiles()`, so I'm not adding anything new there (though I agree that modification of the structure is not implied by the "check" name), I just refactored the existing code into a separate function so it's callable from more places. This `if` statement is here instead of in `constructAbstract` since we need to set the building size before allocating the extent.

lethosor: I guess I was advocating for moving the logic to a separate function (still internal) with a better name, but with context, it sounds like that isn't important. (Basically, it looked to me like a "checkX()" function was changing to "checkXAndMaybeDoY()", with only one caller relying on behavior Y, but I didn't realize you were just moving the Y logic around.)

PR #1869, opened 2021-06-10
[test] measure and display elapsed milliseconds for tests

myk002:
so we can keep tabs on how long tests take to run and optimize particularly inefficient tests.

This is becoming more relevant now that I'm starting to write more involved integration tests. It has already helped me get several `quickfort list` integration tests down from 300ms to 40ms.

With the other, higher-level changelog entries for the test harness, I'm not sure if this deserves its own changelog entry. WDYT?

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -426,7 +426,9 @@ local function run_test(test, status, counts)
     test.private.checks_ok = 0
     counts.tests = counts.tests + 1
     dfhack.internal.IN_TEST = true
+    local start_ms = dfhack.getTickCount()
     local ok, err = wrap_test(test.func)
+    local elapsed_ms = dfhack.getTickCount() - start_ms
     dfhack.internal.IN_TEST = false
     local passed = false
     if not ok then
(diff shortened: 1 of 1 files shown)


ci/test.lua
```diff
@@ -408,7 +412,8 @@ local function run_test(test, status, counts)
     elseif test.private.checks ~= test.private.checks_ok then
         dfhack.printerr('test failed: ' .. test.name)
     else
-        print(('test passed in %d ms: %s'):format(elapsed_ms, test.name))
+        local elapsed_str = get_elapsed_str(elapsed_ms)
+        print(('test passed in %s ms: %s'):format(elapsed_str, test.name))
```
lethosor, line 416: @BenLubar would this cause issues with how Buildmaster parses test output? It looks like this with this change:
```
test passed in 1 ms: structures/types_meta:bitfield
test passed in <1 ms: structures/types_meta:global
test passed in <1 ms: structures/types_meta:unit
test passed in <1 ms: structures/unions:unit_action_fields
test passed in <1 ms: structures/unions:unit_action_type
test passed in <1 ms: test:internal_in_test
test passed in 1 ms: structures/find:image_set
```

(I couldn't find relevant code in https://github.com/BenLubar/build-env or https://git.lubar.me/ben/dfhack-raft)

myk002: also the `Tests completed` line now looks like this:
```
Tests completed in 4198 ms:
```

lethosor: Let's find out!

lethosor (approved):

PR #1871, opened 2021-06-15
[quickfort] add files for quickfort list integration tests

myk002:
depends on #1859 for the CMakefiles modification that prevents the test blueprints from being installed unless `BUILD_TESTS` is on.

Sources for the .xlsx files are [here](https://docs.google.com/spreadsheets/d/19nQtVlqW9gYzr-ZEZ5NO4m9OTV6TOGMolwvCxzxXxp4/edit#gid=0) and [here](https://docs.google.com/spreadsheets/d/1g2c39_skAGQp6dEEBob61a-ZALC0oMObUhpxvsOT7HM/edit#gid=819413246)

I want to minimize the number of .xlsx files checked into the repo, but there need to be at least a few for verification purposes. These files will need new versions if I add additional blueprint modes to quickfort, but I have no plans to do so.


myk002:
actually, if we *really* want to avoid checking in .xlsx files, it might be possible to add xlsxio_xlsxwriter to the build (from the already-added xlsxio dependency), create a plugin wrapper, write a Lua wrapper around that, and construct the .xlsx files as part of the test. This would be a bit of work, but it's an option if you feel strongly about not having binary files in the repo.

lethosor:
(thought I had left a comment on this earlier...) No objection to these file sizes. If we had 10-20 of them in the 15KB range, I might start to be pickier, but as-is, more than half of our `*.cpp` source files are bigger than 15K (from some quick `du -h | sort -h` shenanigans, which could be inaccurate).

lethosor (approved):

PR #1876, opened 2021-06-18
EventManager refactor

cppcooper:
Not sure if this is a welcome refactor or not, but I found the implementation didn't fulfill my expectations of the interface. So I modified it some.
This also includes some documentation updates for plugin writing.

 Refactors EventManager
4e1a834
- handler freq now works on an individual handler basis (set to 0 to fire as frequently as possible)
- tick events no longer need to be re-registered after every eventful tick


expwnent:
I haven't touched DFHack in years but I did write this part of it so I'll try and take a look. The first potential problem that comes to mind is if extant scripts do require non-recurring events to fire at specific times (I assume you've already checked all the plugins for that). Not all scripts are written by DFHack devs so there's no way to know if people are relying on the old behavior. I forget how the plugin `eventful ` works. I didn't write that part.

I'd be pretty surprised if `diggingInvaders` still worked given how long it's been since I wrote it, and I'd be even more surprised if anyone's using it, but that's barely relevant to this PR.

cppcooper:
> Not all scripts are written by DFHack devs so there's no way to know if people are relying on the old behavior. I forget how the plugin `eventful ` works. I didn't write that part.

Eventful doesn't support TICK events from what I can see - which is a problem for another person. Beyond eventful I am pretty sure the EventManager isn't acessible to lua. So the only affected by this, are plugins with tick handlers whether merged with dfhack or not. Though if I messed up one of the functions, that would certainly break dependents.

expwnent:
You are right. `EventManager` is not accessible to lua scripts except through `eventful` as a proxy.

warmist:
If any questions arise about `eventful` i probably could answer them. IIRC tick was not exported to lua as there already was (is) another way to do that. However it's um... different. Namely [dfhack.timeout](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#core-interpreter-context)

library/modules/EventManager.cpp
```diff
@@ -75,8 +88,9 @@ int32_t DFHack::EventManager::registerTick(EventHandler handler, int32_t when, P
...
         }
     }
-    handler.freq = when;
-    tickQueue.insert(pair<int32_t, EventHandler>(handler.freq, handler));
+    handler.when = when;
+    tickQueue.insert(pair<int32_t, EventHandler>(handler.when, handler));
+    eventLastTick[handler.eventHandler] = -1;
     handlers[EventType::TICK].insert(pair<Plugin*,EventHandler>(plugin,handler));
     return when;
 }
```
cppcooper, line 96: Perhaps we shouldn't add tick listeners to the `handlers[TICK]` container. The idea here is that it doesn't provide anything useful to tick events, it is just another thing to clean up inside of `manageTickEvent` in fact.

library/modules/EventManager.cpp
```diff
@@ -353,14 +372,15 @@ static void manageTickEvent(color_ostream& out) {
...
+            ++iter;
             continue;
         }
-        a = handlers[EventType::TICK].erase(a);
+        iter = handlers[EventType::TICK].erase(iter);
         toRemove.erase(handle);
+        registerTick(handle, handle.freq, iter->first);
         if ( toRemove.empty() )
             break;
     }
```
cppcooper, line 386: What I am thinking is, that if we don't add to `handlers[TICK]` inside of `registerTick` then we can get both functionalities. Events which require re-registering a tick event handler, and regular listeners. If registered with the latter, we add it to the `handlers[TICK]` container, then in `manageTickEvent` we could, instead of iterating to remove a handle (which I think I may not have done correctly here) we could then iterate to re-queue it.

library/modules/EventManager.cpp
```diff
@@ -271,15 +296,15 @@ void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event
...
         }
-        for ( size_t a = 0; a < df::global::world->buildings.all.size(); a++ ) {
+        for ( size_t a = 0; a < df::global::world->buildings.all.size(); ++a ) {
             df::building* b = df::global::world->buildings.all[a];
             Buildings::updateBuildings(out, (void*)intptr_t(b->id));
             buildings.insert(b->id);
         }
         lastSyndromeTime = -1;
-        for ( size_t a = 0; a < df::global::world->units.all.size(); a++ ) {
+        for ( size_t a = 0; a < df::global::world->units.all.size(); ++a ) {
```
lethosor, line 305: In cases where the loop variable is a primitive value (a `size_t` in this case), `++a` is _not_ any faster than `a++`.

However, in nearly all of these cases that I see in this PR, the use of iterators (or loop variables) could be replaced with a range-based for loop, such as:

```diff
-        for ( size_t a = 0; a < df::global::world->units.all.size(); ++a ) {
-            df::unit* unit = df::global::world->units.all[a];
+        for ( df::unit *unit : df::global::world->units.all ) {
```

This also works with iterators (maybe with slight changes for iterating over maps). If we're rewriting loops, I would prefer this approach.

library/include/modules/EventManager.h
```diff
@@ -20,6 +20,7 @@ namespace DFHack {
             enum EventType {
                 TICK,
                 JOB_INITIATED,
+                JOB_STARTED, //has a worker
```
lethosor, line 23: New event types also need to be added to `plugins/lua/eventful.lua`.

library/modules/EventManager.cpp
cppcooper: I'll update the loops soon. I've been doing a lot of data migration and permissions may have gotten bonked on more than one repo. Should be able to finish this by the end of August.

cppcooper: Btw I just replaced all the post increments with find iirc. From my recollection pre is not always better but it is never worse, so I just use it exclusively and use those 3 brain cells for other tasks.

library/include/modules/EventManager.h
cppcooper: I added the event to that file and filled in the blanks with `plugins/eventful.cpp`
I haven't tested that it works, but the integration looks to be complete in the same way the other events have been integrated.

library/modules/EventManager.cpp
cppcooper: @lethosor You probably have an order to your reviews, but in case you miss this. What are your thoughts?
```
int32_t DFHack::EventManager::registerTick(EventHandler handler, int32_t when, Plugin* plugin, bool absolute) {
    if ( !absolute ) {
        df::world* world = df::global::world;
        if ( world ) {
            when += world->frame_counter;
        } else {
            if ( Once::doOnce("EventManager registerTick unhonored absolute=false") )
                Core::getInstance().getConsole().print("EventManager::registerTick: warning! absolute flag=false not honored.\n");
        }
    }
    handler.when = when;
    tickQueue.insert(pair<int32_t, EventHandler>(handler.when, handler));
    eventLastTick[handler.eventHandler] = -1;
    //this commented line ensures "Registered Ticks" are not added back to the queue after execution
    //handlers[EventType::TICK].insert(pair<Plugin*,EventHandler>(plugin,handler));
    return when;
}
```

plugins/diggingInvaders/diggingInvaders.cpp
```diff
@@ -372,8 +372,6 @@ void findAndAssignInvasionJob(color_ostream& out, void* tickTime) {
         clearDijkstra();
         return;
     }
-    EventManager::unregister(EventManager::EventType::TICK, findJobTickHandler, plugin_self);
-    EventManager::registerTick(findJobTickHandler, 1, plugin_self);
```
lethosor, line 376: Why was this removed? I see nowhere else where `findJobTickHandler` is registered, so I suspect this is disabling the handler entirely.

library/include/modules/EventManager.h
```diff
@@ -40,6 +41,7 @@ namespace DFHack {
             typedef void (*callback_t)(color_ostream&, void*); //called when the event happens
             callback_t eventHandler;
             int32_t freq;
+            int32_t when;
```
lethosor, line 44: Also asked on [Discord](https://discord.com/channels/793331351645323264/807443262704058408/881243513726074910), but I am curious what the purpose of this field is (is it `tick` or an alternative to `freq`? perhaps a comment would be useful)

library/modules/EventManager.cpp
lethosor: What feedback are you looking for here?

.gitignore
```diff
@@ -25,7 +25,7 @@ build/CMakeCache.txt
 build/cmake_install.cmake
 build/CMakeFiles
 build/data
-build/doc
+build/docs
```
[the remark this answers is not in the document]
cppcooper: All I know is until I fixed this typo and added the Testing directory, I was getting hundreds of untracked files every time I looked. Aside from scripts and configs, git shouldn't be seeing everything in build.

library/include/modules/EventManager.h
cppcooper: It's just the global tick count when it the event needs to trigger. iirc the code was dual purposing freq

library/modules/EventManager.cpp
cppcooper: Well, basically whether to provide both mechanisms or not, but as discovered with diggingInvaders it is somewhat needed. Even if to avoid refactoring diggingInvaders a bunch.

plugins/diggingInvaders/diggingInvaders.cpp
cppcooper: Indeed. I assumed it must be registered somewhere else otherwise it wouldn't ever be registered.. but obviously it gets called automaticalled, or manually.. in order to register initially. This seems like a good enough reason to provide both mechanisms. I'll put these lines back, and ensure `registerTick` doesn't result in re-queued events.

lethosor: I think the `unregister()` call ensured that the event was unregistered (before it was re-registered), if it was registered already. Not sure what would have happened with just one `registerTick` call previously, though.

If we can preserve the old behavior where (I assume) this didn't result in the same handler getting called an increasing number of times per tick, I agree that that would likely simplify things on the caller side.

cppcooper: This should function as it did before now.

`registerTick` gets called when the function executes at least once. `registerTick` will not add the handler to the list, but does queue the event. `manageTick` calls the event when the tick happens, removes it from the queue, adds it to a removal set.. then later after all the events have fired the events in the removal set get cross referenced against the handlers list, and if it is located it gets removed from the list and re-registered with `registerListener` (which **does** add event handlers to the handlers list). The flip side being that if no event handler is found the event isn't re-registered, as will be the case with `registerTick` tick events.

myk002:
This PR introduces a new event type: `JOB_STARTED`, right? Could you make it accessible via the `eventful` interface? Or is that planned for a future PR?

cppcooper:
> This PR introduces a new event type: `JOB_STARTED`, right? Could you make it accessible via the `eventful` interface? Or is that planned for a future PR?

It should be as of this commit
https://github.com/DFHack/dfhack/commit/7cbbf0392b6b2185d3bacf99abbed4b7966d34a1

cppcooper:
I do not know why `Build / docs` fails, it has a number of warnings about things I haven't touched (presumably the warnings are the error failing the build). So my only guess is something in CI changed along with develop and I should merge develop in.

myk002:
Could you add a changelog entry?

cppcooper:
> Could you add a changelog entry?

hmm, I thought I had added one. I'll take a look sometime (this week? I don't know)

myk002:
If you just write it here as a comment, I can add it to `changelog.txt` post-merge

cppcooper:
Sure.

- Adds new event type `JOB_STARTED` - triggered when a job first gains a worker
- Adds new event type `NEW_UNIT_ACTIVE` - triggered when a new unit appears on the active list
- Refactors use of `EventHandler::freq` - now indicates allowed frequency of event
- Adds `EventHandler::when` to supplant the old usage of `EventHandler::freq` - ie. indicates when an event will fire
-

library/modules/EventManager.cpp
```diff
@@ -380,20 +414,54 @@ static void manageJobInitiatedEvent(color_ostream& out) {
...
+            }
+        }
+        if (tick - eventLastTick[handler.eventHandler] >= handler.freq) {
+            eventLastTick[handler.eventHandler] = tick;
+            for(auto job : newly_started_jobs){
+                handler.eventHandler(out, (void*)job);
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
myk002: Ok, no hurry. We can merge this for the next release (which probably won't be too far away - consider this upcoming release a "practice" release for me)

myk002:
Looks good! I'll see if I can get this merged tomorrow.  I'll move the changelog entries to the new "Future" section for the next release after merging.

myk002:
Could you take a look at https://github.com/DFHack/dfhack/issues/2031 ?

If you're busy, I can temporarily revert this PR (and the spectate plugin that depends on it) until we figure that issue out

PR #1877, opened 2021-06-19
Show adjective in item prompt

wolfboyft:
Useful for high boot/low boot etc.
I think I got that in right...
The pcall is because sometimes the item type is, for example, a toy which doesn't have an adjective field, and it's made to error if you index a nonexistent field.


lethosor (commented):
Looks good!

lethosor (approved):

lethosor:
Added a changelog entry in 887362a6

PR #1879, opened 2021-06-21
Implements new plugin channel-safely

cppcooper:
This plugin, when enabled uses two tick events to track designation data. Tracked as groups of adjacent designations (could be considered a "dig project" if you will) which are used to determine when jobs are safe to complete.

~~The two tick events `onNewHour` and `onNewDay` are responsible for building the group data. Where every hour we check the map for new **active** designations, or any marked designations that happens to be swept along in the process. And every day we check the entire map for any designations at all. Both procedures then manage designations as determined appropriate.~~

Designation management comes down to just a few things.

1. does the tile have a designation priority, if not do nothing cause something probably went wrong
2. is the priority less than 6000 (allow marked designations with priority 6000/7000 to be ignored by the plugin)
3. is there group (which isn't done) above the tile, if so then it isn't safe to dig as the job above will become unsafe to dig

edit:

~~There is only one tick event, with an adjustable frequency.~~ Other events are onJobStart, onJobComplete, pause, unpause. The script only does one thing, so it does that one thing on each of these events to ensure no new designations slip by at any point. ~~The tick event is, I think, now redundant due to the 4 other triggers.. but I wasn't sure and was too lazy to both remove it or think about whether it actually is redundant.. pretty sure it is though.~~

The usage:
 ```
channel-safely <option> <args>
     You can use this plugin to manage your channel designations safely. It will\
     mark unsafe designations and unmark safe ones.
 
 channel-safely
     Manually runs the channel-safely management procedure to (un)mark
     designations as needed.
 
 channel-safely enable
     Enables the plugin to automatically run the management procedure. The plugin
     will run when a job starts/completes and on (un)pause/map load state changes.
 channel-safely enable cheats
     Enables the plugin to instantly dig permanently unsafe designations.
 
 channel-safely disable
     Disables the plugin. Will not automatically do anything.
 channel-safely disable cheats
     Disables instantly digging permanently unsafe designations.
```

### CODE REVIEW

**Comment by cppcooper** [comment]
```
 In file included from /usr/include/c++/4.8/memory:79:0,
                 from ../library/include/MiscUtils.h:34,
                 from ../library/include/PluginManager.h:30,
                 from ../plugins/channel-safely.cpp:7:
/usr/include/c++/4.8/functional: In instantiation of ‘struct std::_Bind_simple<void (*(unsigned int, unsigned int, unsigned int))(unsigned int&, unsigned int&, unsigned int&)>’:
/usr/include/c++/4.8/mutex:775:41:   required from ‘void std::call_once(std::once_flag&, _Callable&&, _Args&& ...) [with _Callable = void (&)(unsigned int&, unsigned int&, unsigned int&); _Args = {unsigned int&, unsigned int&, unsigned int&}]’
../plugins/channel-safely.cpp:220:59:   required from here
/usr/include/c++/4.8/functional:1697:61: error: no type named ‘type’ in ‘class std::result_of<void (*(unsigned int, unsigned int, unsigned int))(unsigned int&, unsigned int&, unsigned int&)>’
       typedef typename result_of<_Callable(_Args...)>::type result_type;
                                                             ^
/usr/include/c++/4.8/functional:1727:9: error: no type named ‘type’ in ‘class std::result_of<void (*(unsigned int, unsigned int, unsigned int))(unsigned int&, unsigned int&, unsigned int&)>’
         _M_invoke(_Index_tuple<_Indices...>)
         ^
```
heh, dope errors originating in the STL.. I guess I can't `std::call_once` /shrug

File: plugins/channel-safely.cpp [hunk 1]
```diff
@@ -0,0 +1,480 @@
...
+    if(enabled && World::isFortressMode()) {
+        switch (event) {
+            case SC_UNKNOWN:
+                break;
+            case SC_WORLD_LOADED:
+                break;
+            case SC_WORLD_UNLOADED:
+                break;
+            case SC_MAP_LOADED:
+                manager.manage_designations(out, true);
```
quietust, line 105: If it's only going to handle a single event type, this `switch` statement can probably be collapsed into a single `if` statement.

plugins/channel-safely.cpp
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+    namespace EM = EventManager;
+    if(!has_when<EM::EventHandler>::value){
+        EM::EventHandler tickHandler(onNewHour, hourTicks);
+        EM::registerTick(tickHandler, hourTicks, plugin_self);
+    }
+}
+
+void onNewDay(color_ostream &out, void* tick_ptr){
```
quietust, line 164: In general, formatting is rather inconsistent:
1. `){` versus `) {`
2. `if(` versus `if (`

plugins/channel-safely.cpp
```diff
@@ -0,0 +1,480 @@
...
+                world_pos.z = z;
+                add(world_pos, block);
+            }
+        }
+    }
+}
+
+void GroupData::add(df::coord world_pos, df::map_block* block) {
+    if(groups_map.find(world_pos) == groups_map.end()) {
+        df::coord neighbors[8];
```
quietust, line 382: Might be worth having a helper function to convert a `df::coord` into its neighbors, to avoid having to duplicate this segment from `manageNeighbours()` above

plugins/channel-safely.cpp
cppcooper: I figured the compiler would do that for me

plugins/channel-safely.cpp
cppcooper: huh?

plugins/channel-safely.cpp
```diff
@@ -0,0 +1,480 @@
...
+                }
+            }
+        }
+    }
+}
+
+void ChannelManager::manage_safety(color_ostream &out, df::map_block* block, const df::coord &local, const df::coord &tile, const df::coord &tile_above) {
+    df::tile_occupancy &tile_occupancy = block->occupancy[local.x][local.y];
+    for (df::block_square_event* event : block->block_events) {
+        switch (event->getType()) {
```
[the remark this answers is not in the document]
cppcooper: With a C cast, I presume it should still be a "valid" pointer

plugins/channel-safely.cpp
quietust: Line 164: `void onNewDay(color_ostream &out, void* tick_ptr){`, no space between the close-paren and the open-brace
Line 196: `void onComplete(color_ostream &out, void* job_ptr) {`, *has* space between close-paren and open-brace

Line 165: `if(enabled && World::isFortressMode()) {`, no space between "if" and open-paren
Line 168: `if ((tick_counter - last_tick_counter) >= dayTicks) {`, *has* space between "if" and open-paren

Both of those variances appear all over the place - it's not necessarily a problem, but it stood out to me so I decided to point it out.

cppcooper: Oh, yea. I haven't configured the auto-style rules.

cppcooper:
So this crashes `onComplete`, I believe when a group's last designation job is marked as done.
Reviewing the data structures, I can't see the reason for it. The location of the crash implies that it isn't df or dfhack related.. but I can't 100% rule that out especially since I can't find the exact cause so far. If anyone is bored over the next 12-48 hour, maybe take a crack at solving this?

plugins/channel-safely.cpp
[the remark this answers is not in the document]
cppcooper: Not sure it would simpler. Wouldn't this be the same as a dynamic cast? it needs to deduce the correct type? Isn't that slow? It's been a while since I decided to avoid cast functions unless built-in to the framework.

cppcooper:
I was going to submit a separate PR for the spectate plugin, but realized when I merged my new spectate branch into my local fork-dev that there were conflicts related to dig-now (from the develop branch), so this seems safest.

myk002:
> I was going to submit a separate PR for the spectate plugin, but realized when I merged my new spectate branch into my local fork-dev that there were conflicts related to dig-now (from the develop branch), so this seems safest.

Would rebasing this branch on upstream/develop help?

cppcooper:
> > I was going to submit a separate PR for the spectate plugin, but realized when I merged my new spectate branch into my local fork-dev that there were conflicts related to dig-now (from the develop branch), so this seems safest.
> 
> Would rebasing this branch on upstream/develop help?

I already resolved the conflicts when I merged. As a result of them existing at all though, I pushed to the remote and now this PR has two plugins.

lethosor (commented):
Sorry for the delays here. I'm hoping to get to this this week.

I still see a lot of commits that appear to be from `dig-now` under https://github.com/DFHack/dfhack/pull/1879/commits, and they persist even after I changed the base branch of this PR (which usually forces GitHub to recalculate which commits are actually new on your branch). It does appear that you pulled in an older version of dig-now, which I suppose was rebased when it was merged into develop. For instance, your branch has b9cf23eb326ca4de57e18b8e41b3e1ff710ea512, but the commit that got merged into develop was a4c86770a40eea2fbcc218bba7f27d23dcddff7a. Merging this PR as-is would result in ~50 duplicate commits to dig-now, which I would like to avoid.

Perhaps another rebase on develop would help? It might also be possible to rewrite this branch to _only_ include commits made to channel-safely (and the `Lua API.rst` changes too, assuming you want them). I am unfortunately not able to suggest much more that would clean up the history off the top of my head, but if you would like me to try my hand at rewriting this branch, I would be willing to do that (just be sure to be prepared to replace your branch locally if I do!).

cppcooper:
> Sorry for the delays here. I'm hoping to get to this this week.
> 
> I still see a lot of commits that appear to be from `dig-now` under https://github.com/DFHack/dfhack/pull/1879/commits, and they persist even after I changed the base branch of this PR (which usually forces GitHub to recalculate which commits are actually new on your branch). It does appear that you pulled in an older version of dig-now, which I suppose was rebased when it was merged into develop. For instance, your branch has [b9cf23e](https://github.com/DFHack/dfhack/commit/b9cf23eb326ca4de57e18b8e41b3e1ff710ea512), but the commit that got merged into develop was [a4c8677](https://github.com/DFHack/dfhack/commit/a4c86770a40eea2fbcc218bba7f27d23dcddff7a). Merging this PR as-is would result in ~50 duplicate commits to dig-now, which I would like to avoid.
> 
> Perhaps another rebase on develop would help? It might also be possible to rewrite this branch to _only_ include commits made to channel-safely (and the `Lua API.rst` changes too, assuming you want them). I am unfortunately not able to suggest much more that would clean up the history off the top of my head, but if you would like me to try my hand at rewriting this branch, I would be willing to do that (just be sure to be prepared to replace your branch locally if I do!).

Uhhg. I seem to recall @myk002 telling me this wouldn't happen.
I'll try to fix it. I should be able to rewind a bit, or something.. fucking git surgery sucks. I just hope I don't need to cherry pick anything.

lethosor:
If you're thinking of https://github.com/DFHack/dfhack/pull/1879#issuecomment-893611336, maybe rebasing again would help? The issue is that commits you were relying on (in dig-now) were replaced.

lethosor:
I generally recommend against reverting merges, and it looks like the revert you made in 0decc4a99aaac1576209b5913e7cfdeb77e852cc would end up removing a lot of things (docs/Support.rst, the new Lua readme, etc.). I'm thinking a rebase only is probably best.

Do you want me to try rebasing this branch? If you point me to the files you specifically wanted to change here, I can try.

cppcooper:
So I effectively ran ```git rebase develop --strategy-option=theirs``` but resolved manually.

lethosor:
Can you undo your changelog changes (and maybe add an entry for `channel-safely`)? Other than that, https://github.com/DFHack/dfhack/pull/1879/files looks a lot cleaner now - thanks!

Actually, I think the easiest approach might be to do an interactive rebase and just remove 7875795b398276ba022ff072f07b2bfc6e7549f1 from the list entirely - it looks identical to the overall diff I'm seeing for changelog.txt in this PR, and none of it is what we want (I think).

cppcooper:
```git rebase -p --onto 7875795^ 7875795```
well that's handy

**Comment by cppcooper** [comment]
```
../docs/_auto/news.rst:758: WARNING: undefined label: channel-safely
../docs/_auto/news-dev.rst:802: WARNING: undefined label: channel-safely
```
It is unclear how to resolve this. I suppose though that plugins need to be manually added to the docs.

**Comment by lethosor** [comment]
It doesn't look like there's any documentation for the links to go to. That should be in `docs/Plugins.rst`, and then for the link to work, you'll need to add a link target manually as described at https://docs.dfhack.org/en/latest/docs/Documentation.html#links:

```rst

.. _channel-safely:

channel-safely
==============

(details)
```

cppcooper:
I moved the files to a sub-directory, and I would push.. but I can't build dfhack on my machine. I posted the build errors in discord.

lethosor:
IIRC that was due to submodules being out of sync - looks like it compiled in CI, so that's good!

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,284 @@
...
+            EM::registerListener(EventType::JOB_COMPLETED, jobCompletionHandler, plugin_self);
+        }
+    }
+    // developer debug
+    else if (parameters.size() == 1 && parameters[0] == "debug") {
+        debug_out = &out;
+        ChannelManager::Get().debug();
+        debug_out = nullptr;
+    } else {
+        debug_out = nullptr;
```
lethosor, line 133: Suggestion: this could also print out the help message (as-is, `channel-safely` and `channel-safely help` both print nothing - we don't really _require_ plugins to handle those, but it is nice if they can)

plugins/channel-safely.cpp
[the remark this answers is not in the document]
ab9rf: Using `virtual_cast` helps to futureproof this code against the possibility that Toady changes something with the underlying type hierarchy, or that we find an error in our reverse-engineering. Both of these things happen from time to time.

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,269 @@
...
+#ifdef CS_DEBUG
+    debug_out = &out;
+    out.print("debugging enabled\n");
+#endif
+    if (enable && !enabled) {
+        using namespace EM::EventType;
+        EM::EventHandler tickHandler(onNewEvent, maxTickFreq);
+        EM::EventHandler jobStartHandler(onStart, 0);
+        EM::EventHandler jobCompletionHandler(onComplete, 0);
+        if (!has_when<EM::EventHandler>::value) {
```
cppcooper, line 156: I suppose I should remove this now. Might save the template somewhere for my own future reference though.

docs/Plugins.rst
```diff
@@ -1159,6 +1159,59 @@ Examples:
...
+    You can use this plugin to manage your channel designations safely. It will
+    mark unsafe designations and unmark safe ones.
+
+``channel-safely enable``
+    Enables the plugin to automatically run the management procedure. The plugin
+    can run on tick events and will run on (un)pause/job events and map load
+    state changes.
+``channel-safely enable cheats``
+    Enables the plugin to instantly dig permanently unsafe designations.
+``channel-safely enable ontick``
```
myk002, line 1201: The `ontick` handler was removed -- docs referring to it should be removed as well

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <PluginManager.h>
+#include <modules/EventManager.h>
+#include "channel-safely.h"
+
+using namespace DFHack;
+
+DFHACK_PLUGIN("channel-safely");
+DFHACK_PLUGIN_IS_ENABLED(enabled);
+REQUIRE_GLOBAL(world);
+//#define CS_DEBUG 1
```
myk002, line 16: instead of compile-time constants and undocumented "debug" commands, consider using DFHack's logging framework in `Debug.h`

plugins/channel-safely/channel-safely.h
```diff
@@ -0,0 +1,120 @@
...
+extern bool cheat_mode;
+extern void getNeighbours(const df::coord &tile, df::coord(&neighbours)[8]);
+extern void manageNeighbours(color_ostream &out, const df::coord &tile);
+extern void cancelJob(df::job* job);
+extern bool is_group_ready(const GroupData &groups, const GroupData::Group &group);
+extern bool is_group_occupied(const GroupData &groups, const GroupData::Group &group);
+extern bool safe_to_dig_down(const df::coord &tile);
+extern bool is_dig(df::job* job);
+extern bool is_channel(df::job* job);
+extern bool is_channel(df::tile_designation &designation);
```
myk002, line 120: needs final newline

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+    } else if (parameters.size() == 2 && parameters[0] == "enable" && parameters[1] == "cheats" && !cheat_mode) {
+        cheat_mode = true;
+        out.print("channel-safely: cheat mode enabled!\n");
+    }
+    // disable options
+    else if (parameters.size() == 1 && parameters[0] == "disable") {
+        return plugin_enable(out, false);
+    } else if (parameters.size() == 2 && parameters[0] == "disable" && parameters[1] == "cheats" && cheat_mode) {
+        cheat_mode = false;
+        out.print("channel-safely: ontick events disabled!\n");
```
myk002, line 85: should refer to cheats, not ontick events

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+        debug_out = &out;
+        ChannelManager::Get().debug();
+        debug_out = nullptr;
+    } else if (parameters.size() == 1 && parameters[0] == "help") {
+        return CR_WRONG_USAGE;
+    } else {
+        debug_out = nullptr;
+        out.color(color_value::COLOR_RED);
+        out.print("Invalid argument.\n\n");
+        out.color(color_value::COLOR_RESET);
```
myk002, line 98: you should use `out.printerr()` instead of manually changing the color

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+    debug_out = &out;
+#endif
+    if (debug_out) debug_out->print("onstatechange()\n");
+    if (enabled && World::isFortressMode() && Maps::IsValid()) {
+        switch (event) {
+            case SC_MAP_LOADED:
+                if (debug_out) debug_out->print("SC_MAP_LOADED\n");
+                ChannelManager::Get().manage_designations(out);
+                break;
+            case SC_PAUSED:
```
myk002, line 140: why are we running on pause events? what state could have changed from the last handled event?

plugins/channel-safely/channel-safely-plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+            ChannelManager::Get().mark_done(job->pos);
+            if (debug_out) debug_out->print("manage_safety()\n");
+            ChannelManager::Get().manage_safety(out, block, local, below, job->pos);
+            if (debug_out) debug_out->print("manageNeighbours()\n");
+            manageNeighbours(out, job->pos);
+        }
+    }
+    if (debug_out) debug_out->print("onJobComplete() - return\n");
+    debug_out = nullptr;
+}
```
myk002, line 211: needs EOF newline

plugins/channel-safely/channel-safely.cpp
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+void DigJobs::cancel_job(const df::coord &pos) {
+    auto iter = jobs.find(pos);
+    if (iter != jobs.end()) {
+        df::job* job = iter->second;
+        cancelJob(job);
+    }
+}
```
myk002, line 332: needs final newline

myk002 (changes requested):
haven't actually reviewed `channel-safely.cpp` or the header file yet, but I'm getting pulled away for a bit so I wanted to send what I had so far.

library/modules/Maps.cpp
```diff
@@ -118,8 +117,24 @@ void Maps::getSize (uint32_t& x, uint32_t& y, uint32_t& z)
...
+    y = int32_t(world->map.y_count_block);
+    z = int32_t(world->map.z_count_block);
+}
 
 // getter for map size in tiles
+void Maps::getTileSize (int32_t& x, int32_t& y, int32_t& z){
+    getSize(x, y, z);
+    x *= 16;
+    y *= 16;
+}
```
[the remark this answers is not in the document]
cppcooper: These take references, `ReadFeature` does not ie. you can't cast (I'd say something about rvalues or w/e but I struggle with those concepts). The existing methods should simply be removed because they are incorrect, but I'm trying to patch problems not create more. If you could tell me how to properly deprecate them that'd be great

library/modules/Maps.cpp
```diff
@@ -118,8 +117,24 @@ void Maps::getSize (uint32_t& x, uint32_t& y, uint32_t& z)
...
+void Maps::getSize (uint32_t& x, uint32_t& y, uint32_t& z)
+{
+    if (!IsValid())
+    {
+        x = y = z = 0;
+        return;
+    }
+    x = int32_t(world->map.x_count_block);
+    y = int32_t(world->map.y_count_block);
+    z = int32_t(world->map.z_count_block);
```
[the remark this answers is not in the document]
cppcooper: They shouldn't be there at all I suspect, I was probably testing something and forgot to revert.

docs/Plugins.rst
```diff
@@ -1159,6 +1159,59 @@ Examples:
...
         Turn off autolabor for wood cutting.
 
+.. _channel-safely:
+
+channel-safely
+==============
+Automatically manage channel designations to safely complete them.
+Channel-safely keeps dwarves safe by marking unsafe designations for
+later. A designation is unsafe if it is under another designation or
+over empty space. The latter situation will never be safe, but with
```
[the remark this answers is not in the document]
cppcooper: It's considered unsafe because the space below is empty, so when they dig themselves down they fall into the empty. If there are ways to prevent this it can't be easily ensured afaik.

I'm gonna go and add some comments throughout the source when I find some time.. but from what I remember "over empty space" essentially means they channeled into a channel or cavern, so there won't be a floor directly below them to fall onto which might already hurt them.. like bruise their feet.. or ego.. or something.

plugins/channel-safely/channel-safely-plugin.cpp
cppcooper: They may pause to designate an entire fortress. That's why I run on unpause. I run on pause, simply because that's a good time to run (the game is doing nothing so there are plenty of cpu cycles to go around). I can likely remove some more triggers, but I went with `better safe than sorry`

docs/Plugins.rst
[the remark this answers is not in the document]
myk002: That's what I don't get, though - dwarves don't stand on the tile they're channeling unless they approaching it from the tile directly above it. They normally stand to the side as they channel, don't they?

plugins/channel-safely/channel-safely-plugin.cpp
myk002: Unpause I totally agree with. Even theoretically, though, what new data could be found by the pause trigger? If nothing new was found by the now-removed tick events, why are we keeping pause?

library/modules/Maps.cpp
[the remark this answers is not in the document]
myk002: I think I'd do something like this:
```
void Maps::getSize (uint32_t& x, uint32_t& y, uint32_t& z)
{
    int32_t intx, inty, intz:
    getSize(intx, inty, intz);
    x = intx; y = inty; z = intz;
}
```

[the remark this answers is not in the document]
cppcooper: I get the inclination but why let the compiler add a second call? If it doesn't inline the function in the other it's that extra overhead for no good reason

plugins/channel-safely/channel-safely-plugin.cpp
cppcooper: I think actually that tick event was necessary. I forgot that the job events don't build groups, the only check/update with the job they have.

I'm going to finish going through and commenting.. but for now I think I'll be adding that tick event back (wups).. or maybe more likely once the commenting is done we have a discussion on what makes most sense.

docs/Plugins.rst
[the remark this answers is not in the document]
cppcooper: I don't think so, I can double check. Maybe you're thinking of situations where they can't stand on the tile they're going to channel so they have to dig it first then channel? Or I could be mistaken, but ime they fall if there is nothing below the tile so either way it'd be unsafe cause it's not worth checking where they are standing in my opinion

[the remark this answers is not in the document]
cppcooper: Glad I double checked. Guess I gotta eat my words. Dunno if I thought that because I had been seeing an edge case or what. Not sure I want to revise it to path find a path out.. but I suppose people would like that =/

docs/Lua API.rst
```diff
@@ -4009,7 +4009,7 @@ Lua plugin classes
 ``crng``
 ~~~~~~~~
 
-- ``init(id, df, dist)``: constructor
+- ``new(id, df, dist)``: constructor
```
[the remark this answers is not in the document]
cppcooper: mmm, I've got a bad habit of noticing problems in my code/documentation when I'm in the middle of working on something totally different with huge diffs that need to be sorted through for commits

plugins/channel-safely/channel-safely-plugin.cpp
cppcooper: Ok, so finished adding inline/function comments; just remembered I left the class comments unfinished (oops), they're less important though I think. Reorganized the source, separating the classes. It means more files, but I think it's easier to find things now.

So.. I do think the tick event was performing a critical operation in that it prevented the job events from lacking group data.. but with a 1 day frequency it certainly could have missed something in between.

I added
```cpp
void ChannelManager::manage_one(color_ostream &out, const df::coord &map_pos, df::map_block* block, bool manage_neighbours) {
    if (!groups.count(map_pos)) {
        build_groups();
    }
```
which get's called by the events, so that should avoid some problems.. but it might miss new designations above it.. now that I think about that..

But.. Now that's got a solution too.. I added a build call inside `has_groups_above` which also now checks if there is an untracked designation above it..

So with all this I think it makes 100% sense to remove the Pause trigger

docs/Plugins.rst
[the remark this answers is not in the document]
cppcooper: @myk002 so how would I detect this true unsafe over empty space case? (from a high level)

library/modules/Maps.cpp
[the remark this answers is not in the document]
cppcooper: I put the methods as inline inside the implementation file, and use them inside like you suggested in another comment thread. Hopefully the compiler will inline them inside the TU

library/modules/Maps.cpp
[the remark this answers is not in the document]
cppcooper: Like I commented in the other, I did this. But the inner call is defined as inline inside the .cpp file so hopefully it gets inlined

plugins/channel-safely/channel-safely-plugin.cpp
cppcooper: I'll try to take a look at it sometime soon.

docs/Plugins.rst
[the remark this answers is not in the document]
myk002: Ugh.. I had written a whole train of thought on my phone but didn't submit it and now it's gone. Let me see what I can reconstruct.

A lot of this will be duplicating what you already have, but let me see if I can form a chain of reasoning.

1. digging a channel is unsafe if it is likely to cause a dwarf to fall and get injured
2. a dwarf can only fall if there is empty space under the tile that is being channeled and a dwarf is on top of the tile when it is channeled
3. situations where a dwarf will be on top of the channeled tile:
    a) dwarf is channeling that tile and can only approach that tile from above OR was already standing on that tile when they took the job to dig it
    b) two channel designations are adjacent and a dwarf is standing on one designated tile while digging the other
    c) dwarf is pathing through the tile on the way to somewhere else
4. digging a channel is also undesirable (though not necessarily unsafe to any dwarves) if after digging the channel other designated tiles are no longer reachable

So, potential algorithm.

1. find groups of connected channeling designations (a "project)
2. put all designations in the project in marker mode that have project tiles directly above them (so when those tiles are channeled we don't have new empty space to contend with)
3. for unmarkered tiles that are over open space, marker them all, then unmarker a tile whose removal would not make any adjacent tile unreachable AND whose path does not involve a dwarf coming straight down at it. You can probably do more than one of these tiles at a time, but we'd have to think carefully about how.

also, we can set all passable tiles that are being channeled as a restricted traffic area to minimize other dwarves passing through at a bad time

test cases:

- channeling 10 z-levels straight down through rock should work. the tiles should be channeled strictly top to bottom, even if the tiles on middle levels are accessible to other miners
- channeling a 10x10 floor area over magma should slowly dig out the tiles one by one until they are all clear
- channeling 3 z-levels straight down through the ceiling of a cavern should stop at the last tile before the ceiling of the cavern is breached (since the path to that tile is straight down). The dwarf will become stuck.

how to calculate a path? df has a cache you can use, but how to test changes to that pathing? This I don't know.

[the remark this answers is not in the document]
cppcooper: > 2. ...on top of the tile when

This made me think of an easy partial solution whereby when a job is started the plugin spawns a tick event, probably a customized lambda, to check if the job's tile is occupied by a dwarf until the event handler for ticks is unregistered when the completed job event fires. During a tick even, when the job tile is occupied we pause the job somehow.. or cancel and requeue it(?).

I think then we could ignore pathing entirely.



> 2. put all designations in the project in marker mode that have project tiles directly above them (so when those tiles are channeled we don't have new empty space to contend with)
> 
> 3. for unmarkered tiles that are over open space, marker them all, then unmarker a tile whose removal would not make any adjacent tile unreachable AND whose approach path does not involve a dwarf coming straight down at it. You can probably do more than one of these tiles at a time, but we'd have to think carefully about how.

I think this would be a change from what I have (or did have, at least), but I think it would be a welcome one.

What I think it currently does is effectively wait until the top-most overlapping layer of designations are finished to enable any overlapping groups below them. And changing, I don't think I'd need to worry about groups.. but I dunno maybe it's better to wait for the group to finish to ensure paths above aren't ruined.. but then we've maybe made digging pits more difficult to complete due to project lifespan and the size of the hole. I can foresee trapped dwarves potentially, who must dig to escape.. might just be a fun! addition to keep?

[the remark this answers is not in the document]
cppcooper: I'm not sure when I'll get to this. I've been busy searching for an apartment and with work and uni. I foresee a large refactor, and perhaps reimplementation coming though

PR #1881, opened 2021-06-22
Fix label early key detection in scrolling logic

warmist:
Current logic is because @myk002 fixed it so label would allow other widgets control when label(s) are present. However that breaks label key detection for default scroll keys. This can be worked around by setting scrollkeys to empty.

TBH: label is quite complicated and used everwhere so i'm reluctant to touch it and would love for someone to look over if i'm correct.

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -442,7 +442,6 @@ function Label:onInput(keys)
                 v = -self.frame_body.height
             end
             self:scroll(v)
-            return false
         end
     end
     return check_text_keys(self, keys)


myk002 (approved):
Change looks good, and tested in `gui/quickfort` and `gui/gm-editor`. I believe this was my original intention when I realized that scroll keys needed to propagate beyond the Label. However, instead of removing the `return true` line, I changed it to `return false` instead of removing the line altogether as this PR does.

Since this actually fixes user-visible behavior, this PR could use a changelog entry.

lethosor:
+1 for a changelog entry. I put together https://github.com/DFHack/dfhack/commit/9b58fa45ff5ea3825c5271de317ec48923162395 based on the [discussion on Discord](https://discord.com/channels/793331351645323264/793331351645323267/856882510364540928), but can definitely change it.

PR #1882, opened 2021-06-23
Add getPlantAtTile function to Maps namespace

Bumber64:
Derived from disassembly of DF code. Returns a pointer to plant struct that owns the tile at position. Useful for finding a tree from one of its branches. Lua API support.

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -1923,6 +1923,13 @@ static int maps_getTileBiomeRgn(lua_State *L)
     return Lua::PushPosXY(L, Maps::getTileBiomeRgn(pos));
 }
 
+static int maps_getPlantAtCoords(lua_State *L)
+{
+    auto pos = CheckCoordXYZ(L, 1, true);
+    Lua::PushDFObject(L, Maps::getPlantAtCoords(pos));
+    return 1;
+}
+
 static const luaL_Reg dfhack_maps_funcs[] = {
     { "isValidTilePos", maps_isValidTilePos },
     { "isTileVisible", maps_isTileVisible },
@@ -1932,6 +1939,7 @@ static const luaL_Reg dfhack_maps_funcs[] = {
     { "getTileFlags", maps_getTileFlags },
     { "getRegionBiome", maps_getRegionBiome },
     { "getTileBiomeRgn", maps_getTileBiomeRgn },
+    { "getPlantAtCoords", maps_getPlantAtCoords },
     { NULL, NULL }
 };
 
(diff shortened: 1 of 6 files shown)


Bumber64:
changelog.txt probably needs to be updated.

library/LuaApi.cpp
```diff
@@ -1932,6 +1939,7 @@ static const luaL_Reg dfhack_maps_funcs[] = {
     { "getTileFlags", maps_getTileFlags },
     { "getRegionBiome", maps_getRegionBiome },
     { "getTileBiomeRgn", maps_getTileBiomeRgn },
+    { "getPlantAtCoords", maps_getPlantAtCoords },
```
lethosor, line 1942: Minor consistency thing: we don't really seem to use `Coords` in any other function names (though I know it was suggested as part of reverse-engineering on the forums). Perhaps `getPlantAtTile()` or `getPlantAt()` would be more consistent? I have a slight preference towards the former, for what it's worth.

library/modules/Maps.cpp
```diff
@@ -713,6 +715,50 @@ bool Maps::canStepBetween(df::coord pos1, df::coord pos2)
...
+{
+    if (x < 0 || x >= world->map.x_count || y < 0 || y >= world->map.y_count || !world->map.column_index)
+        return NULL;
+
+    df::map_block_column *mbc = world->map.column_index[(x / 48) * 3][(y / 48) * 3];
+    if (!mbc)
+        return NULL;
+
+    int32_t x_mod_48 = x % 48;
+    int32_t y_mod_48 = y % 48;
```
lethosor, line 731: Another more minor thing: would a more descriptive name like `region_x` work here? (Trying to use terminology from https://docs.dfhack.org/en/stable/docs/api/Maps.html, although I would still want to make it clear that these are coordinates _within_ a region, not _of_ a region).

These are only used once below, and I personally don't find `x_mod_48` more clear than just `(x % 48)`, so I'd mainly just suggest keeping this for readability. (If calculated with `x % 48` in the loop below, I'd expect most basic optimizations to result in the value just being calculated once anyway, so hand-optimizing may not be necessary here.)

lethosor (commented):
Yep, a changelog entry would be great - I'd probably put it in the "API" section, since it affects multiple languages. (Might need to create that section if it doesn't exist for the next version yet)

lethosor (approved):
Thanks for putting this together!

I'd usually recommend pushing to your fork from a local clone of the repo so that you don't have to modify individual files through the GitHub UI (assuming that's what you were doing). Happy to help if there was a part of this workflow giving you trouble.

PR #1883, opened 2021-06-23
Fix keybindings in the "Building Material" screen

arzyu:
Hi,
This PR improves the interactions of the "Building Material" screen.

The key bindings currently used in this interface are:
```
Move selector up/down - for item selection
Move view/cursor left/right - for column selection
```
There will be no problems with the default DF key bindings because "Move selector ..." and "Move view/cursor ..." both use Up/Down/Left/Right. However, if you use different key bindings for them, the problem arises.

This PR uniformly uses "Move selector ..." as key bindings for item/column selection:
```
Move selector up/down - for item selection
Move selector left/right - for column selection
```


myk002 (approved):

myk002:
code changes look good, though please add a changelog entry to `docs/changelog.txt` explaining the improvement. For example, you could add this line to the "Fixes" section:
```
`buildingplan`: ensure navigation keys work properly in the materials selection screen when alternate keybindings are used
```

arzyu:
@myk002 Thanks for reminding me. The change log has been added.

lethosor (approved):

PR #1885, opened 2021-06-23
new library string function: wrap()

myk002:
added to the string class like `startswith()` and `endswith()`

refactored and improved from the implementation in `quickfort`'s `dialog.lua`


lethosor (approved):
Looks good, and thanks for the tests!

PR #1886, opened 2021-06-23
[test] add quotes around printed unequal values

myk002:
so you can see confusing hidden trailing spaces


lethosor (approved):
I think we might as well add this to `expect.ne`, for parity if nothing else. For instance, `expect.ne('a ', 'a ')` could benefit from more clarity (although I'm not sure why that would be a useful test). The rest should generally only be used with numeric values anyway.

Edit: possible better example: empty strings passed to `expect.ne` produce this, which could definitely be more clear:
```
Check failed!  == 
```

PR #1887, opened 2021-06-23
[test] add delay_until function

myk002:
delays until a predicate returns true or the timeout (in frames) elapses


lethosor (approved):
Thanks! Not sure that these sorts of things necessarily belong in the changelog, although I appreciate the thought there - maybe they should just go in the "dev" changelog, if that's possible? Not a huge deal - I'll take a look as part of my pass for the next release.

PR #1889, opened 2021-06-28
support view unit mode for guidm.enterSidebarMode

myk002:
it's a top-level mode and should be supported.

### DIFF
--- a/test/library/gui/dwarfmode.lua
+++ b/test/library/gui/dwarfmode.lua
@@ -64,4 +64,10 @@ function test.enterSidebarMode()
     guidm.enterSidebarMode(df.ui_sidebar_mode.Default)
     expect.eq(df.ui_sidebar_mode.Default, df.global.ui.main.mode)
     expect.eq('dwarfmode/Default', dfhack.gui.getCurFocus(true))
+
+    -- verify that all supported modes lead where we say they'll go
+    for k,v in pairs(guidm.SIDEBAR_MODE_KEYS) do
+        guidm.enterSidebarMode(k)
+        expect.eq(k, df.global.ui.main.mode)
+    end
 end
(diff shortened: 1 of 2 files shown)


test/library/gui/dwarfmode.lua
```diff
@@ -64,4 +64,10 @@ function test.enterSidebarMode()
     guidm.enterSidebarMode(df.ui_sidebar_mode.Default)
     expect.eq(df.ui_sidebar_mode.Default, df.global.ui.main.mode)
     expect.eq('dwarfmode/Default', dfhack.gui.getCurFocus(true))
+
+    -- verify that all supported modes lead where we say they'll go
+    for k,v in pairs(guidm.SIDEBAR_MODE_KEYS) do
+        guidm.enterSidebarMode(k)
+        expect.eq(k, df.global.ui.main.mode)
+    end
```
lethosor, line 72: Should this switch back to `Default` when done? Or is that handled by ci/test.lua?

myk002: it's handled by `ci/test.lua`, but I added a mode switch back to default so `ci/test.lua` doesn't have to report that it's autocorrecting anything.

test/library/gui/dwarfmode.lua
```diff
@@ -64,4 +64,12 @@ function test.enterSidebarMode()
     guidm.enterSidebarMode(df.ui_sidebar_mode.Default)
     expect.eq(df.ui_sidebar_mode.Default, df.global.ui.main.mode)
     expect.eq('dwarfmode/Default', dfhack.gui.getCurFocus(true))
+
+    -- verify that all supported modes lead where we say they'll go
+    for k,v in pairs(guidm.SIDEBAR_MODE_KEYS) do
+        guidm.enterSidebarMode(k)
+        expect.eq(k, df.global.ui.main.mode)
```
lethosor, line 71: FYI, this is a situation where I'd use the `comment` argument to provide information about which iteration of the loop failed. It's fairly easy to determine if non-equal numbers get printed in this case, though. I dealt with a couple similar cases here: 1014c41d

myk002: Agree. Thanks for adding

PR #1891, opened 2021-06-29
[blueprint] normalize pathnames and handle dirs as basenames

myk002:
Depends on #1890 for the path normalization library function

uses the new path normalization library function and fixes handling of blueprint basenames that end in a slash (that is, basenames that are directories). For a basename of `dirname/` the code previously generated the following files:
```
blueprints/dirname/-dig.csv
blueprints/dirname/-build.csv
...
```
it now drops the dash and generates files like:
```
blueprints/dirname/dig.csv
blueprints/dirname/build.csv
...
```

### DIFF
--- a/plugins/blueprint.cpp
+++ b/plugins/blueprint.cpp
@@ -587,11 +587,20 @@ static string get_tile_query(df::building* b)
     return " ";
 }
 
+// can remove once we move to C++20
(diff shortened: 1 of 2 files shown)


lethosor (approved):
Verified manually that the failing tests pass with #1890 merged

PR #1894, opened 2021-06-30
[orders] fix crash on malformed order IDs

### LINKED ISSUE #1893: orders import DF crash
Version: DFHack version 0.47.05-r1 (release) on x86_64

orders import of an encrusted gem list results in crash to desktop.
Linked is the orders file testing with.
https://pastebin.com/7QTMkr9E

myk002:
Fixes #1893

This is similar to the fix in #1770, which caught errors during initial read, but this PR catches errors thrown during processing. This came up because a manually edited orders json file had a string ID instead of a numeric ID.

I believe this now covers all the places that can throw errors for import. I also believe export is safe since we control the input there and don't do string processing, but I can't say that for absolute certain.


lethosor (approved):
Thanks! (It'd maybe be nice to have more details than `Value is not convertible to Int.`, but that's a detail for another time)

myk002:
> Thanks! (It'd maybe be nice to have more details than `Value is not convertible to Int.`, but that's a detail for another time)

Yeah, I considered keeping context on which attribute was being accessed at all times such that the name would be accessible in the exception handler, but I didn't think it was worth it for such an edge case. If we get complaints, though, I have a solution designed.

PR #1898, opened 2021-07-03
migrate existing lua code to use new library fns

### LINKED ISSUE #1888: Move string.split() to core library
`string.split()` is currently [added to the `string` global](https://github.com/DFHack/scripts/blob/13fbb5a84d59d43bf68ce190726192484d52e00a/gui/load-screen.lua#L105) by `gui/load-screen`.  This affects any Lua code run after this script is run. There are several scripts that rely on this function being defined, and possibly more that are out of our control:
* https://github.com/DFHack/scripts/blob/c453a53bce292caea2e608ac86942dae4d92e73f/color-schemes.lua#L644
* https://github.com/DFHack/scripts/blob/c453a53bce292caea2e608ac86942dae4d92e73f/gui/settings-manager.lua#L35

Since this script is run in `dfhack.init-example`, this is not an issue in many cases. However, it did break Buildmaster in #1791 (after the test image was updated?) since that test environment uses an empty `dfhack.init` and does not run `gui/load-screen` at all, leaving `string.split()` undefined.

`gui/load-screen`'s implementation of `split()` fails to handle some edge cases, so I would recommend using the one defined in [utils.lua](https://github.com/DFHack/dfhack/blob/0297a5b1b8ea5492c99cbd214b37dcfd97f16bbe/library/lua/utils.lua#L485) instead (possibly replacing `split_string()` with a stub calling `string.split()`).

Another useful improvement would be to expose the `plain` argument of `string.find()`, e.g. as `string.split(str, sep, plain)`. `utils.split_string()` always interprets the separator as a pattern, which requires escaping characters like `.` in the separator (and this is the _only_ separator used in gui/load-screen, incidentally).

Inspired by #1885

myk002:
Fixes #1888 
Depends on #1897 

step 5 of #1888 

This PR migrates library and plugin lua code to use the new string and argparse library functions (with some minor auxiliary cleanup). Each change has been manually tested, with coverage of changed lines verified with luacov (except for stockflow. I verified that it works but for some reason luacov doesn't show coverage for that particular file).

### DIFF
--- a/plugins/lua/blueprint.lua
+++ b/plugins/lua/blueprint.lua
@@ -1,5 +1,6 @@
 local _ENV = mkmodule('plugins.blueprint')
 
+local argparse = require('argparse')
 local utils = require('utils')
 
 -- the info here is very basic and minimal, so hopefully we won't need to change
@@ -43,17 +44,11 @@ local valid_phase_list = {
 valid_phases = utils.invert(valid_phase_list)
 
 local function parse_cursor(opts, arg)
-    local _, _, x, y, z = arg:find('^(%d+),(%d+),(%d+)$')
-    if not x then
-        qerror(('invalid argument for --cursor option: "%s"; expected format' ..
-                ' is "<x>,<y>,<z>", for example: "30,60,150"'):format(arg))
-    end
+    local cursor = argparse.coords(arg)
     -- be careful not to replace struct members when called from C++, but also
     -- create the table as needed when called from lua
     if not opts.start then opts.start = {} end
-    opts.start.x = tonumber(x)
-    opts.start.y = tonumber(y)
-    opts.start.z = tonumber(z)
+    utils.assign(opts.start, cursor)
 end
 
 local function parse_positionals(opts, args, start_argidx)
@@ -93,7 +88,7 @@ local function process_args(opts, args)
         return
     end
 
-    return utils.processArgsGetopt(args, {
+    return argparse.processArgsGetopt(args, {
             {'c', 'cursor', hasArg=true,
              handler=function(optarg) parse_cursor(opts, optarg) end},
             {'h', 'help', handler=function() opts.help = true end},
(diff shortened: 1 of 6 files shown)


plugins/lua/blueprint.lua
```diff
@@ -43,17 +44,11 @@ local valid_phase_list = {
 valid_phases = utils.invert(valid_phase_list)
 
 local function parse_cursor(opts, arg)
-    local _, _, x, y, z = arg:find('^(%d+),(%d+),(%d+)$')
-    if not x then
-        qerror(('invalid argument for --cursor option: "%s"; expected format' ..
-                ' is "<x>,<y>,<z>", for example: "30,60,150"'):format(arg))
-    end
+    local cursor = argparse.coords(arg)
```
lethosor, line 47: I'm getting a test failure locally on this branch (when merged into develop @ f1ec222d9a6ec910af271731cdc71c0815d6678e) that I think is due to this change:
```
error: specified coordinates not on current map: "%s": 1,2,3

test errored: plugins/blueprint:parse_gui_commandline
```
Unfortunately it lacks a traceback, but there's only one test it could refer to:
```
    opts = {}
    b.parse_gui_commandline(opts, {'--cursor=1,2,3'})
    expect.table_eq({auto_phase=true, name='blueprint', start={x=1,y=2,z=3}},
                    opts)
```

myk002: I see why I missed this. `argparse.coords()` checks the validity of the coordinates, which will fail if a map is not loaded. the tests are marked as `mode=none` so a map is not guaranteed to be loaded, but it just so happened to be loaded every time I ran the tests.

CI would have caught this once I merged in the argparse PR, though, so this doesn't make me concerned that the test system is fragile or anything. Fixed.

Also fixed error string formatting error in argparse.

lethosor: Yep, I merged in the argparse PR (#1897) in an attempt to get CI to do this, but re-running CI didn't make it pick up the new commits from develop (only rebasing did, it seems).

Thanks for the formatting fix as well!

lethosor (approved):

PR #1903, opened 2021-07-11
[buildingplan] display which items are attached and which are pending

myk002:
This allows the player to determine which items are missing and need to be produced before a building can be built. It is most useful for buildings that take multiple items, such as dyer's workshops.

the information is visible in the `q`uery sidebar when examining a planned building


lethosor (approved):
Nice!

PR #1906, opened 2021-07-19
[orders] support reaction-specific item traits

myk002:
implemented for both import and export

uses the new test wrapper functionality to export current orders, run the test, clear orders, import the player's previous orders, and remove temp files. this prevents the tests from having visible side effects. The tests will reset the next order id, but this isn't visible to the player.

### DIFF
--- a/test/plugins/orders.lua
+++ b/test/plugins/orders.lua
@@ -1,7 +1,27 @@
 config.mode = 'fortress'
 
-TMP_FILE_NAME = 'tmp-test'
-TMP_FILE_PATH = ('dfhack-config/orders/%s.json'):format(TMP_FILE_NAME)
+local FILE_PATH_PATTERN = 'dfhack-config/orders/%s.json'
+
+local BACKUP_FILE_NAME = 'tmp-backup'
+local BACKUP_FILE_PATH = FILE_PATH_PATTERN:format(BACKUP_FILE_NAME)
+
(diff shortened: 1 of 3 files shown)


test/plugins/orders.lua
```diff
@@ -1,26 +1,70 @@
...
+    expect.eq(result, CR_OK)
+
+    -- ignore whitespace (otherwise the expected file content is impossible to
+    -- format properly in this file)
+    expect.eq(normalize_whitespace(expected_file_content),
+              normalize_whitespace(file_content))
 end
 
-function check_import_success(file_content)
+function check_import_success(file_content, comment, num_expected_orders)
```
lethosor, line 62: Minor thing: `comment` is unused. If it makes sense to include in failure messages, I suggest passing it through as the `comment` argument to the two `expect.eq` calls below to provide more context (or you could take the approach of `check_import_fail()` and create fancier error messages). Alternatively, the `comment` argument could be removed, given that test failure tracebacks _should_ be enough to isolate the cause of the failure now.

myk002: Fixed. Passed `comment` through to the `expect.eq` calls, which was my original intention. The full tracebacks have definitely superseded comments for identifying which test failed, but I still find comments useful for reminding myself what the heck I'm actually testing if it's not completely obvious.

lethosor (approved):

PR #1909, opened 2021-07-25
add autonick config file

Moth-Tolias:
example config file for autonick (https://github.com/DFHack/scripts/pull/315)


dfhack-config/autonick.txt
```diff
@@ -0,0 +1,13 @@
+# autonick config file
+# every line in this file that is not empty and does
+# not start with "#" will be used as a nickname by the
+# autonick script.
+
+Mouse
```
myk002, line 6: keep in mind that this file will actually get installed to the live dfhack-config/ dir, so if people just run the autonick script, they'll get a dwarf nicknamed "Funny Name for a Dwarf".

The example names in this file should probably be sane defaults (all the nicknames here other than "Funny Name for a Dwarf" are probably perfectly cromulent). or all commented out so autonick does nothing until the file is populated.

Personally I think I vote for a default file of 300 or so reasonable, easy-to-remember and possible-to-differentiate nicknames. Then without the user having to do any custom configuration, they can get some benefit from running autonick.

Moth-Tolias: i don't think i have anywhere near 300 names in my own lists, but 30 should be more than achievable.

Moth-Tolias:
here's a pared-down version of my own personal list with the more unusual names and "non-unisex" names removed. i've also grouped them into categories.

Moth-Tolias:
as it turns out the first search result for unisex names has 300 entries, so we might use that - any thoughts?

dfhack-config/autonick.txt
myk002: This looks great! Thanks!

there seem to be a few duplicates:
```
$ sort autonick.txt | uniq -d

Mouse
River
```

myk002:
> as it turns out the first search result for unisex names has 300 entries, so we might use that - any thoughts?

If they're reasonably dwarfy enough, sure. The kinds of names you have currently look great.

Moth-Tolias:
ah, reasonably dwarfy... i might have to trim the 300 list down again then :laughing: unfortunately most of the ones in the list are pretty human-y, but some might fit

Moth-Tolias:
a couple of the new names are repeats but i believe we should have over 100 now.

dfhack-config/autonick.txt
Moth-Tolias: whoops, i missed this comment, terribly sorry! i've removed the duplicates now. very handy trick, that!

dfhack-config/autonick.txt
```diff
@@ -0,0 +1,161 @@
...
+King
+Piper
+Poet
+Sailor
+Tanner
+Mason
+Miller
+Baker
+Archer
+Hunter
```
myk002, line 106: We should probably avoid nicknames that can be confused with professions. Otherwise "Mason" Urist, the miller could get confused, for example, with "Miller" Urist, the mason.

myk002 (approved):
Looks good! Thanks!
@lethosor for final review and merge

lethosor (approved):

Issue #1912, opened 2021-07-29
FilteredList:setFilter should work by default with choices consisting of tokens


TymurGubayev:
It is possible to make a multi-colored list by using choices like following.
```lua
text = {{text = 'none', pen = COLOR_RED}}
```

Unfortunately, this breaks the filter of a `ListBox` `with_filter`. It throws following
```
hack\lua\gui\widgets.lua:818: bad argument #1 to 'match' (string expected, got table)
```
https://github.com/DFHack/dfhack/blob/43d79052277d5b7cac3ffc5ee5f19d24953fa190/library/lua/gui/widgets.lua#L818

Now that I'm looking at surrounding code
https://github.com/DFHack/dfhack/blob/43d79052277d5b7cac3ffc5ee5f19d24953fa190/library/lua/gui/widgets.lua#L813-L822
I'm noticing one can provide a `search_key` with choices. I think the trivial choice -- just concatenating the texts -- should just be the default.

I.e.
```lua
local function set_search_keys(choices)
	for _, choice in ipairs(choices) do
		if not choice.search_key then
			local search_key = {}
			if type(choice.text) == 'table' then
				for _, token in ipairs(choice.text) do
					search_key[#search_key+1] = token.text or ''
				end
			end
			choice.search_key = table.concat(search_key, ' ')
		end
	end
end
```

If we throw `string.lower` in there we can have a case-insensitive search as well.

PR #1913, opened 2021-07-29
ListBox attributes controlling self:dismiss()

TymurGubayev:
`dismiss_on_select = false` makes it easier to implement toggleable options, where we don't want the list to disappear on selection. Like work order conditions' trait selection.

### DIFF
--- a/library/lua/gui/dialogs.lua
+++ b/library/lua/gui/dialogs.lua
@@ -150,8 +150,8 @@ ListBox.focus_path = 'ListBox'
 
 ListBox.ATTRS{
     with_filter = false,
-    dismiss_on_submit = true,
-    dismiss_on_submit2 = true,
+    dismiss_on_select = true,
+    dismiss_on_select2 = true,
     cursor_pen = DEFAULT_NIL,
     select_pen = DEFAULT_NIL,
     on_select = DEFAULT_NIL,
@@ -176,7 +176,7 @@ function ListBox:init(info)
     local on_submit2
     if self.select2_hint or self.on_select2 then
         on_submit2 = function(sel, obj)
-            if self.dismiss_on_submit2 then self:dismiss() end
+            if self.dismiss_on_select2 then self:dismiss() end
             if self.on_select2 then self.on_select2(sel, obj) end
             local cb = obj.on_select2
             if cb then cb(obj, sel) end
@@ -192,7 +192,7 @@ function ListBox:init(info)
             text_pen = spen,
             cursor_pen = cpen,
             on_submit = function(sel,obj)
-                if self.dismiss_on_submit then self:dismiss() end
+                if self.dismiss_on_select then self:dismiss() end
                 if self.on_select then self.on_select(sel, obj) end
                 local cb = obj.on_select or obj[2]
                 if cb then cb(obj, sel) end
--- a/library/lua/gui/dialogs.lua
+++ b/library/lua/gui/dialogs.lua
@@ -150,6 +150,8 @@ ListBox.focus_path = 'ListBox'
 
 ListBox.ATTRS{
     with_filter = false,
+    dismiss_on_submit = true,
+    dismiss_on_submit2 = true,
     cursor_pen = DEFAULT_NIL,
     select_pen = DEFAULT_NIL,
     on_select = DEFAULT_NIL,
@@ -174,7 +176,7 @@ function ListBox:init(info)
     local on_submit2
     if self.select2_hint or self.on_select2 then
         on_submit2 = function(sel, obj)
-            self:dismiss()
+            if self.dismiss_on_submit2 then self:dismiss() end
             if self.on_select2 then self.on_select2(sel, obj) end
             local cb = obj.on_select2
             if cb then cb(obj, sel) end
@@ -190,7 +192,7 @@ function ListBox:init(info)
             text_pen = spen,
             cursor_pen = cpen,
             on_submit = function(sel,obj)
-                self:dismiss()
+                if self.dismiss_on_submit then self:dismiss() end
                 if self.on_select then self.on_select(sel, obj) end
                 local cb = obj.on_select or obj[2]
                 if cb then cb(obj, sel) end
--- a/test/library/gui/dialogs.lua
+++ b/test/library/gui/dialogs.lua
@@ -1,5 +1,3 @@
-config.mode = 'fortress'
-
 local gui = require('gui')
 local function send_keys(...)
     local keys = {...}
--- a/test/library/gui/dialogs.lua
+++ b/test/library/gui/dialogs.lua
@@ -1,0 +1,172 @@
+config.mode = 'fortress'
+
+local gui = require('gui')
+local function send_keys(...)
+    local keys = {...}
+    for _,key in ipairs(keys) do
+        gui.simulateInput(dfhack.gui.getCurViewscreen(true), key)
+    end
+end
+
+local xtest = {} -- use to temporarily disable tests (change `function test.somename` to `function xtest.somename`)
+local wait = function()
+	--delay(30) -- enable for debugging the tests
+end
+
+local dialogs = require('gui.dialogs')
+
+function test.ListBox_opens_and_closes()
+	local before_scr = dfhack.gui.getCurViewscreen(true)
+	local choices = {{
+		text = 'ListBox_opens_and_closes'
+	}}
+	local lb = dialogs.ListBox({choices = choices})
+	expect.eq(before_scr, dfhack.gui.getCurViewscreen(true), "creating a ListBox object should not change CurViewscreen")
+	lb:show()
+	wait()
+	expect.ne(before_scr, dfhack.gui.getCurViewscreen(true), "ListBox:show should change CurViewscreen")
+	send_keys('LEAVESCREEN')
+	expect.eq(before_scr, dfhack.gui.getCurViewscreen(true), "Pressing LEAVESCREEN should return us to previous screen")
+end
+
+function test.ListBox_closes_on_select()
+	local before_scr = dfhack.gui.getCurViewscreen(true)
+
+	local args = {}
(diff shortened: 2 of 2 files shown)


library/lua/gui/dialogs.lua
```diff
@@ -150,6 +150,8 @@ ListBox.focus_path = 'ListBox'
 
 ListBox.ATTRS{
     with_filter = false,
+    dismiss_on_submit = true,
+    dismiss_on_submit2 = true,
```
myk002, line 154: from the perspective of the ListBox, shouldn't these be named `dismiss_on_select` and `dismiss_on_select2`?

TymurGubayev: I probably got confused because of how the Lua API docs mention on_submit/2, but they are actually not directly accessible anymore. I'll change it to (maybe) avoid further confusion.

myk002:
Looks good! Would you be ok with adding some unit tests for this? Not only would it help guarantee that the feature doesn't break when we change the class in the future, it would also help scripters see how the feature is intended to be used.

To create a unit test, add a file named `test/gui/dialogs.lua`. the content of the file should look like this:
```
config.mode = 'fortress'

local dialogs = require('gui.dialogs')

function test.ListBox_multi_select()
    -- add test here that instantiates a ListBox, sends keys to manipulate it, and verifies expected results
end
```

You can see examples of gui tests in the scripts repo, such as https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua

myk002:
I should clarify that I'm not asking for you to write tests for all of `gui/dialogs.lua`, or even all of the `ListBox` class. Just the specific feature that this PR adds. We'll flesh out the unit test coverage of existing code over time.

TymurGubayev:
> Looks good! Would you be ok with adding some unit tests for this? Not only would it help guarantee that the feature doesn't break when we change the class in the future, it would also help scripters see how the feature is intended to be used.
> 
> To create a unit test, add a file named `test/gui/dialogs.lua`. the content of the file should look like this:
> 
> ```
> config.mode = 'fortress'
> 
> local dialogs = require('gui.dialogs')
> 
> function test.ListBox_multi_select()
>     -- add test here that instantiates a ListBox, sends keys to manipulate it, and verifies expected results
> end
> ```
> 
> You can see examples of gui tests in the scripts repo, such as https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua

Sure, I can do that. But how do I actually run the tests to check they are working?

myk002:
> Sure, I can do that. But how do I actually run the tests to check they are working?

Yes, it would have been useful if I had told you that, wouldn't it ; )

first, change your cmake settings so BUILD_TESTS is ON, regenerate the build files, and run `ninja install` (or however you normally build and install dfhack).This will install the test harness and copy your new test file under `hack/scripts/test`.

then, at the `[DFHack]#` prompt, run `test` to run all tests or `test -tgui/dialogs` to just run the test you're writing. Note that when running all tests, the tests that require the game to be sitting at the title screen will fail (assuming you already have a fortress loaded). You can skip the "title mode" by specifying `test --modes=none,fortress`.

TymurGubayev:
> Yes, it would have been useful if I had told you that, wouldn't it ; )

That's what I thought!

> first, change your cmake settings so BUILD_TESTS is ON, regenerate the build files, and run `ninja install` (or however you normally build and install dfhack).This will install the test harness and copy your new test file from `<dfhack_src>/test/gui/dialogs.lua` to `<df_dir>/hack/scripts/test/gui/dialogs.lua`.

Yeah, I tried some time ago to actually build dfhack myself, and, well, normally I just use the LNP and may occasionally download it from the build server. Since I'm only changing Lua scripts, any chance there is a BUILD_TESTS-version lying around for me to just download?

myk002:
It's not difficult to do manually. Here's the "no build" version:

1. copy test.lua from `<dfhack_src>/ci/test.lua` (also available [online](https://github.com/DFHack/dfhack/blob/develop/ci/test.lua)) and copy it to `<df_dir>/hack/scripts/test.lua`
2. at the `[DFHack]#` prompt, run `test -d/path/to/the/test/dir/where/your/test/file/is -tdialogs`

TymurGubayev:
ok, got the test engine running. Immediate follow-up question: is there a sleep function or something to delay test execution so I can better debug it?

myk002:
`delay(frames)` is in the environment for tests. defined [here](https://github.com/DFHack/dfhack/blob/develop/ci/test.lua#L81) and [here](https://github.com/DFHack/dfhack/blob/develop/ci/test.lua#L301), example usage [here](https://github.com/DFHack/dfhack/blob/develop/ci/test.lua#L158).

myk002:
also see the related `delay_until(callback)` function, defined [here](https://github.com/DFHack/dfhack/blob/develop/ci/test.lua#L89), example usage [here](https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua#L59)

TymurGubayev:
So, I couldn't place it near https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua because it's in another project. So I placed the file in `test/library/gui/` instead.

test/library/gui/dialogs.lua
```diff
@@ -0,0 +1,172 @@
...
+
+    local args = {}
+    local mock_cb = mock.func()
+    local mock_cb2 = mock.func()
+    args.on_select = mock_cb
+    args.on_select2 = mock_cb2
+    args.dismiss_on_select = false
+    args.dismiss_on_select2 = false
+    args.choices = {{
+        text = 'ListBox_with_multi_select'
```
myk002, line 145: This PR gives us a low-level option for the behavior of the list, but I'm wondering now (seeing the tests does help me visualize the usage patterns) if we need some more functionality around it. This really isn't "multi-select" per se. It's just allowing multiple single `submit()` calls without forcing the user to re-open the ListBox, except with no visual changes to the dialog to indicate what is happening.

If the goal here is to provide a true multi-select interface, the visuals and behavior might need modification.  For example, it should somehow visually indicate that multiple items are selected (e.g. via a highlight that is distinguishable from the cursor or perhaps just an X character to the left of the list item). The user should also be told somehow that LEAVESCREEN is the only way out of the dialog if both `dismiss_on_select` options have been set to true.

What do you think? Do you already have some code that would use this that I could look at for context? Perhaps we can move some of that code into the library and encapsulate the behavior in a MultiSelectListBox subclass.

myk002:
> So, I couldn't place it near https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua because it's in another project. So I placed the file in `test/library/gui/` instead.

`test/gui` is fine. "library" is assumed.

test/library/gui/dialogs.lua
TymurGubayev: I've added a test demonstrating how I use it. The actual code where I use it belongs into the dfhack/scripts repository, probably in the materials.lua -- I'll get it there some other day.

I can make a MultiSelectListBox out of this, but there are more issues related to that - like FilteredList.setFilter not working by default anymore (see #1912). So I thought as the first step I'd propose the minimal possible change to achieve what I need :)

>  The user should also be told that LEAVESCREEN is the only way out of the dialog if both dismiss_on_select options have been set to true.

I'm not sure about that: pressing ESC to leave a screen is the natural way of doing it and works everywhere in DF. Additional advertising would only add visual clutter.

test/library/gui/dialogs.lua
```diff
@@ -0,0 +1,263 @@
+config.mode = 'fortress'
```
lethosor, line 1: I think this line can be removed - I removed it on my end, and the tests run without issue on the title screen.

test/library/gui/dialogs.lua
lethosor: I think I'm ok with leaving out `Esc` - we're already leaving out `Enter` and `Shift-Enter` in all ListBox instances.

lethosor:
> `test/gui` is fine. "library" is assumed.

The "library" prefix is consistent with the existing `test/library/gui/dwarfmode.lua`, so I'd prefer it. I guess a loose rule of thumb is that tests for `x/y.lua` live in `test/x/y.lua`, with any path components like "lua" omitted. (Whether "library" should also be omitted is a fair point, but it looks like we haven't.)

test/library/gui/dialogs.lua
TymurGubayev: If I remove it, the test might cause problems in some CI pipeline (or something, I don't know)

https://github.com/DFHack/dfhack/blob/e1f0e94f0f04729d5b4a0e4e70a8c2266d9b0928/ci/test.lua#L286-L287

https://github.com/DFHack/dfhack/blob/e1f0e94f0f04729d5b4a0e4e70a8c2266d9b0928/ci/test.lua#L46-L48

lethosor: Ok, well that is probably because fortress mode tests _don't run_ in CI currently. If it is failing in CI, that's something that should be fixed. Do you know where the error was occurring? I'll see if I can track it down, but if not, I am inclined to remove this line and see what happens.

lethosor: The line in question was added at the top of 70d088c763138ee3199080387e5bbfe68feeeae9 and hasn't been changed since, so I don't know how the lack of it could have triggered a CI failure. I'll try removing it on your branch.

TymurGubayev: > The line in question was added at the top of [70d088c](https://github.com/DFHack/dfhack/commit/70d088c763138ee3199080387e5bbfe68feeeae9) and hasn't been changed since, so I don't know how the lack of it could have triggered a CI failure. I'll try removing it on your branch.

sorry for confusion, I haven't actually experienced any problems, I just assumed something somewhere would go wrong.

But now that I think about it, you are right, a dialog window should work regardless of mode.

lethosor (approved):

PR #1915, opened 2021-08-06
Upgrade to Lua 5.3.6

lethosor:
I accidentally discovered [this bug](https://www.lua.org/bugs.html#5.3.4-1) in Lua 5.3.3 (which we use currently), and figured an upgrade would be worthwhile, even though this isn't very likely to affect us (given our limited use of `goto`s).

All changes are copied directly from `lua-5.3.6.tar.gz` upstream. The only exception is `luaconf.h`, where some changes needed to be applied from upstream and some of our own changes needed to be kept. I added markers around our changes to hopefully avoid confusion in the future.

My understanding is that "patch" releases of Lua don't break compatibility, but the number of changes here is a bit larger than I expected for 3 patch releases, so I think it's worthwhile to test a bit more and make sure our code doesn't break before merging this. I'm no expert on Lua internals, but a couple changes that stood out to me include:
* https://github.com/DFHack/dfhack/commit/e311e002c7890ad4340cca343ac74ad9a9035cd1#diff-aa1c6563f33582e1ecb3e2819a4a311d5563ac7216a371ee4f3a8b9e6416a15eL812
* https://github.com/DFHack/dfhack/commit/e311e002c7890ad4340cca343ac74ad9a9035cd1#diff-44c92af21289640b1608413cfd1aef7aa3403eda7986ba7955e7a35bf32d72c1L212 (probably won't break compatibility, but might affect how `pairs()` behaves)
* https://github.com/DFHack/dfhack/commit/e311e002c7890ad4340cca343ac74ad9a9035cd1#diff-16c407824315c94fdce687e5cb8a9179f9bae74acee50f7152e24773a6f069f9L658 (looks like this logic just moved around)


lethosor:
Found one regression so far:
```
[lua]# ~dfhack.lineedit()
>> f
f
./hack/lua/dfhack.lua:168: bad argument #1 to 'for iterator' (table expected, got string)
stack traceback:
    [C]: in function 'next'
    ./hack/lua/dfhack.lua:168: in function 'printall'
    ./hack/lua/dfhack.lua:537: in function <./hack/lua/dfhack.lua:535>
    [C]: in function 'dfhack.safecall'
    ./hack/lua/dfhack.lua:594: in function 'dfhack.interpreter'
    ../../code/dfhack/scripts/lua.lua:76: in local 'script_code'
    ./hack/lua/dfhack.lua:753: in function 'dfhack.run_script_with_env'
    (...tail calls...)
[lua]#
```

which has no issue with Lua 5.3.3:
```
[lua]# ~dfhack.lineedit()
>> f
f
[lua]# 
```

myk002:
> Found one regression so far:
> ```
> [lua]# ~dfhack.lineedit()
> >> f
> f
> ./hack/lua/dfhack.lua:168: bad argument #1 to 'for iterator' (table expected, got string)

which is the same as `~('f')`, right? it looks like this might have worked by chance in the past. even the signature of our `printall(table)` function assumed the input would be a table and not a string. It is useful for `printall` to handle non-tables, of course. We could fix this particular regression (and also `printall_recurse`?) by adding `if type(table) ~= 'table' then print(tostring(table)) return end` to the beginning of `printall`.

myk002 (approved):
From what I can tell, it looks like a safe upgrade. I think we can adapt to the change in behavior to `pairs`. The change to `tostring` *probably* doesn't affect us, but if we do find a script that breaks because of it, it also looks easy to adapt to -- just add a metatable with the `__tostring` field defined.

lethosor:
Hah, you're right, it was an issue to `printall()`. I only tested `lineedit()` and nothing else.

PR #1916, opened 2021-08-06
[Core] clear modstate on window focus

### LINKED ISSUE #1878: Keybindings sometimes act as if Alt is stuck down
As of commit 2fc5fbacb5bdf7183d61045a4786d58571081eec (though the problem showed up in earlier commits), I've noticed that some keybindings don't respect the Alt modifier. All of the keybindings I've tested are part of the default configuration.

For reference, I'm playing on 64-bit Linux, DF version 0.47.05

A list of problems I've noted:

* `Alt-I` (for gui/unit-info-viewer) will trigger when you hit `I` in the `v` screen, which conflicts with the inventory screen. Luckily DF also receives the input, so backing out of the DFHack screen will leave you looking at the inventory screen.
* `Alt-R` (for gui/room-list) will trigger when you hit `r` when making a bedroom out of a bed. As before, DF also receives the input, so backing out takes you to the expected room-building menu.
* `Alt-F1` (for hotkeys) will trigger on `F1`. Unlike the other two, DF won't receive the input, and so won't zoom to the location F1 is supposed to.

It seems that hitting `Alt` at all will suddenly require use of the Alt key again, and alt-tabbing away from the game brings the problem back again. Between the alt-tab and the extra `Alt` keypress, you can use these commands without `Alt` as often as you want.  My guess is that DFHack is seeing the Alt of the alt-tab and holding on to it when it shouldn't.

myk002:
Fixes #1878 (Keybindings sometimes act as if Alt is stuck down)

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## New Plugins
 - `dig-now`: instantly completes dig designations (including smoothing and carving tracks)
 
+## Fixes
+- Core: ``alt`` keydown state is now cleared when DF loses and regains focus, ensuring the ``alt`` modifier state is not stuck on for systems that don't send standard keyup events in response to ``alt-tab`` window manager events
+
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


lethosor (approved):
Thanks!

PR #1917, opened 2021-08-06
[lua] ensure printall, ~, and @ behave in lua 5.3.6

myk002:
prepares for the Lua upgrade in #1915

suggested fix for lua 5.3.6 change to how `pairs` works (or doesn't) with non-table input.

This ensures `!`, `^`, `@`, `~`, and plain `printall` all behave sanely with both table and non-table input.
```
[[lua]# !('hi')
hi
[lua]# ^('hi')
hi
[lua]# @('hi')
hi
[lua]# ~('hi')
hi
[lua]# printall('hi')
hi
[lua]# !({'hi'})
table: 0x7fffd0113680
[lua]# ^({'hi'})
table: 0x7fffd01137c0
1                       = hi
[lua]# @({'hi'})
table: 0x7fffd00927b0
1                        = hi
[lua]# ~({'hi'})
table: 0x7fffd00c2ad0
1                        = hi
[lua]# printall({'hi'})
1                        = hi
```

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -175,7 +175,7 @@ end
 
 function safe_pairs(t, iterator_fn)
     iterator_fn = iterator_fn or pairs
-    if (pcall(pairs, t)) then
+    if (pcall(iterator_fn, t)) then
         return _wrap_iterator(iterator_fn(t))
     else
         return function() end
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -176,15 +176,11 @@ local function print_element(k, v)
 end
 
 function printall(table)
-    if not safe_iterate(table, pairs, print_element) then
-        dfhack.println(tostring(table))
-    end
+    safe_iterate(table, pairs, print_element)
 end
 
 function printall_ipairs(table)
-    if not safe_iterate(table, ipairs, print_element) then
-        dfhack.println(tostring(table))
-    end
+    safe_iterate(table, ipairs, print_element)
 end
 
 local do_print_recurse
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -162,22 +162,28 @@ NEWLINE = "\n"
 COMMA = ","
 PERIOD = "."
 
+-- calls elem_cb(k, v) for each element of the table
+-- returns true if we iterated successfully, false if not
+local function safe_iterate(table, iterate_fn, elem_cb)
+    local function iterate()
+        for k,v in iterate_fn(table) do elem_cb(k, v) end
+    end
+    return pcall(iterate)
+end
+
+local function print_element(k, v)
+    print(string.format("%-23s\t = %s", tostring(k), tostring(v)))
+end
+
 function printall(table)
-    if type(table) ~= 'table' then print(tostring(table)) return end
-    local ok,f,t,k = pcall(pairs,table)
-    if ok then
-        for k,v in f,t,k do
-            print(string.format("%-23s\t = %s",tostring(k),tostring(v)))
-        end
+    if not safe_iterate(table, pairs, print_element) then
+        print(tostring(table))
     end
 end
 
 function printall_ipairs(table)
-    local ok,f,t,k = pcall(ipairs,table)
-    if ok then
-        for k,v in f,t,k do
-            print(string.format("%-23s\t = %s",tostring(k),tostring(v)))
-        end
+    if not safe_iterate(table, ipairs, print_element) then
+        print(tostring(table))
     end
 end
 
@@ -200,15 +206,9 @@ local fill_chars = {
 setmetatable(fill_chars, fill_chars)
 
 local function print_fields(value, seen, indent, prefix)
-    local ok,f,t,k = pcall(pairs,value)
-    if not ok then
-        dfhack.print(prefix)
-        dfhack.println('<Type doesn\'t support iteration with pairs>')
-        return 0
-    end
     local prev_value = "not a value"
     local repeated = 0
-    for k, v in f,t,k do
+    local print_field = function(k, v)
         -- Only show set values of bitfields
         if value._kind ~= "bitfield" or v then
             local continue = false
@@ -234,7 +234,10 @@ local function print_fields(value, seen, indent, prefix)
             end
         end
     end
-    if repeated > 0 then
+    if not safe_iterate(value, pairs, print_field) then
+        dfhack.print(prefix)
+        dfhack.println('<Type doesn\'t support iteration with pairs>')
+    elseif repeated > 0 then
         dfhack.println(prefix .. "<Repeated " .. repeated .. " times>")
     end
     return 0
@@ -535,11 +538,11 @@ function dfhack.interpreter(prompt,hfile,env)
         end,
         ['~'] = function(data)
             print(table.unpack(data,2,data.n))
-            if type(data[2]) == 'table' then printall(data[2]) end
+            printall(data[2])
         end,
         ['@'] = function(data)
             print(table.unpack(data,2,data.n))
-            if type(data[2]) == 'table' then printall_ipairs(data[2]) end
+            printall_ipairs(data[2])
         end,
         ['^'] = function(data)
             printall_recurse(data[2])
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -172,18 +172,18 @@ local function safe_iterate(table, iterate_fn, elem_cb)
 end
 
 local function print_element(k, v)
-    print(string.format("%-23s\t = %s", tostring(k), tostring(v)))
+    dfhack.println(string.format("%-23s\t = %s", tostring(k), tostring(v)))
 end
 
 function printall(table)
     if not safe_iterate(table, pairs, print_element) then
-        print(tostring(table))
+        dfhack.println(tostring(table))
     end
 end
 
 function printall_ipairs(table)
     if not safe_iterate(table, ipairs, print_element) then
-        print(tostring(table))
+        dfhack.println(tostring(table))
     end
 end
 
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -163,6 +163,7 @@ COMMA = ","
 PERIOD = "."
 
 function printall(table)
+    if type(table) ~= 'table' then print(tostring(table)) return end
     local ok,f,t,k = pcall(pairs,table)
     if ok then
         for k,v in f,t,k do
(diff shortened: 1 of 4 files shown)


library/lua/dfhack.lua
```diff
@@ -163,6 +163,7 @@ COMMA = ","
 PERIOD = "."
 
 function printall(table)
+    if type(table) ~= 'table' then print(tostring(table)) return end
```
lethosor, line 166: This unfortunately breaks for userdata: `~scr` now just prints `<viewscreen_titlest: 0x260c000>` instead of also printing fields. `pcall(pairs, table)` below was intended to allow any type where `pairs()` works.

I think https://github.com/DFHack/dfhack/commit/e311e002c7890ad4340cca343ac74ad9a9035cd1#diff-44c92af21289640b1608413cfd1aef7aa3403eda7986ba7955e7a35bf32d72c1L212 is likely what changed the behavior here - `pairs()` no longer throws an error on invalid objects, like `nil`, but `next()` does.

myk002: reworked to use a `safe_iterate()` function, which should allow all possible iterables to work.

`printall` and `printall_ipairs` now output a straight `tostring` of their input if iteration fails, but this has an undesirable effect when running, for example, `~('hi')`, which now prints out:
```
hi
hi
```
once by `print(table.unpack(data,2,data.n))` and once by `printall(data[2])`. What are your thoughts on how this should be handled? should `printall` go back to not printing anything for non-iterables?

lethosor: I was thinking something more along the lines of a drop-in replacement for `pairs()` would work, rather than something taking a callback, such as:
```lua
for k, v in safe_pairs(t) do
   ...
end
```

I suppose something like this could also be an option, although I can't think of use-cases other than `pairs()` and `ipairs()` off the top of my head:
```
for k, v in safe_iterate(pairs, t) do
   ...
end
```

I've found Lua iterators somewhat tricky to work with in the past, though, so your approach is easier to deal with. I'll see if I can wrangle something more similar to `pairs()` together.

myk002: If we write a drop-in replacement for pairs that just never fails, then how can we tell if something is non-iterable? Or is that the desired behavior - just do nothing instead of failing?

lethosor: Yep, `printall()` and `~` have historically silently printed nothing for non-iterable types. The couple other uses of `pcall(pairs)` also seem to be trying to iterate only if a type is iterable and doing nothing otherwise.

lethosor: For reference, here's a very basic implementation of what I had in mind that worked in Lua 5.3.3:
```lua
function safe_pairs(t)
    if (pcall(pairs, t)) then
        return pairs(t)
    else
        return function() end
    end
end
```

I haven't gotten this to work with 5.3.6 quite yet (the issue is that pairs() returns `next` as its first return value, and _that_ is what errors in 5.3.6)

lethosor: Ok, here's something I came up with that works as I was hoping in 5.3.6:
```lua
function _wrap_iterator(next_fn, ...)
    local wrapped_iter = function(...)
        local ret = {pcall(next_fn, ...)}
        local ok = table.remove(ret, 1)
        if ok then
            return table.unpack(ret)
        end
    end
    return wrapped_iter, ...
end

function safe_pairs(t)
    if (pcall(pairs, t)) then
        return _wrap_iterator(pairs(t))
    else
        return function() end
    end
end
```

`safe_pairs()` could be parameterized to take in `pairs` or `ipairs`, of course.

Example use:
```lua
function safe_printall(t)
    for k, v in safe_pairs(t) do
        print(('%s = %s'):format(k, v))
    end
end
```

I have somewhat mixed feelings about _how much_ longer it is than before. There's also some overhead from wrapping every call to `next_fn` (aka `next()`) with `pcall()`, although your implementation also has overhead. Unfortunately, looking at [this section of the Lua manual](https://www.lua.org/manual/5.3/manual.html#3.3.5), the iterator function itself must remain constant between iterations (`next` is equivalent to `f` in their example).

library/lua/dfhack.lua
```diff
@@ -162,21 +162,28 @@ NEWLINE = "\n"
...
+end
+
 function printall(table)
-    local ok,f,t,k = pcall(pairs,table)
-    if ok then
-        for k,v in f,t,k do
-            print(string.format("%-23s\t = %s",tostring(k),tostring(v)))
-        end
+    if not safe_iterate(table, pairs, print_element) then
+        dfhack.println(tostring(table))
```
lethosor, line 180: Just wanted to clarify that `printall()` previously didn't print anything if `pairs()` failed, so we should keep that behavior, and this appears to have changed it (I think). `~` in the Lua interpreter is a combination of `print()` and `printall()`, so I'm sometimes getting duplicate output with this change:

```
[DFHack]# lua ~_VERSION
Lua 5.3
Lua 5.3
```

myk002: yeah, I was attempting to "fix" its failure to output anything for non-iterables. I thought it was a deficiency that `printall()` didn't output anything for strings. I [noticed](https://github.com/DFHack/dfhack/pull/1917#discussion_r684580485) that it double-printed for strings when using `~` on the interactive prompt, but I wasn't sure how best to handle that. I'll change it back to the original behavior, though.

library/lua/dfhack.lua
myk002: I see the utility of `safe_pairs` as defined above for use in scripts. It is a good API-compatible drop-in replacement for `pairs`. However, I'm not sure it's better than the current implementation for use within `dfhack.lua`. First, I'm not sure exactly how much overhead `pcall` adds, but the current PR's `safe_iterate()` function calls `pcall()` once and `safe_pairs()` above calls `pcall()` O(N) times. That's a theoretical argument, though. I'll get some hard data on whether this matters when I can.

Second, though, regardless of whether we add `safe_pairs()` to `dfhack.lua` (and I think we should) we would still need a function like the current `safe_iterate()` for use in the `print_fields()` method, which needs to know *whether* an object was successfully iterated so it can report if iteration failed. And if we need the current, more processing-efficient implementation anyway, why not use it as the non-user-visible backend to `printall()`?

myk002:
What is an example of a userdata value that will cause iteration to fail? I tried `printall_recurse` on a userdata `uint32_t` but that just successfully output its value.

library/lua/dfhack.lua
TymurGubayev: >I have somewhat mixed feelings about how much longer it is than before 

I can help with that:
```lua
function safe_next(t, k)
    local ok, next_k, v = pcall(next, t, k)
    if ok then return next_k, v
    else return nil
    end
end

function safe_pairs (t)
    return safe_next, t, nil
end
```

lethosor:
`df.item._identity` is one that would fail:
```
[lua]# ~df.item._identity
userdata: 0x7ffff7f99d40
[lua]# pairs(df.item._identity)
(interactive):1: bad argument #1 to 'pairs' (table expected, got light userdata)
```

Note that it's technically light userdata, i.e. it doesn't have a metatable at all. It is certainly possible to create a userdata object from C with a metatable that doesn't have a `__pairs` metamethod, but I couldn't find an existing instance of that quickly.

myk002:
that's perfect. test coverage now 100%. Thanks!

test/library/misc.lua
```diff
@@ -0,0 +1,24 @@
+-- tests misc functions added by dfhack.lua
```
lethosor, line 1: I don't see any tests for a non-`pairs` iterator_fn, so maybe one could be useful.

If we want to test for the possible bug I identified above, I think we would need a type where pairs() errors and ipairs() is supported. I'm not aware of an existing one, but something like this could work:

```lua
local t = {1, 2}
setmetatable(t, {__pairs = function() error() end})
```

lethosor (commented):
I like the optional param to `safe_pairs()`! Sorry for the delay on this one - just a minor thing I noticed

test/library/misc.lua
lethosor: Here's a more complete test I came up with, and I also confirmed that the above is a bug:
```lua
function test.safe_pairs_ipairs()
    local t = {1, 2}
    setmetatable(t, {
        __pairs = function()
            expect.fail('pairs() should not be called')
        end,
    })

    local iterated = 0
    for k,v in safe_pairs(t, ipairs) do
        expect.eq(k, v)
        iterated = iterated + 1
    end
    expect.eq(#t, iterated)
end
```

library/lua/dfhack.lua
```diff
@@ -162,24 +162,49 @@ NEWLINE = "\n"
...
 
-function printall_ipairs(table)
-    local ok,f,t,k = pcall(ipairs,table)
-    if ok then
-        for k,v in f,t,k do
-            print(string.format("%-23s\t = %s",tostring(k),tostring(v)))
-        end
+function safe_pairs(t, iterator_fn)
+    iterator_fn = iterator_fn or pairs
+    if (pcall(pairs, t)) then
```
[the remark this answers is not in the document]
myk002: Ummm yes it should. Will fix.

library/lua/dfhack.lua
myk002: this is indeed shorter, but where does it actually call `pairs(t)`?

TymurGubayev: > this is indeed shorter, but where does it actually call `pairs(t)`?

Yeah, forgot about __pairs. How about this one then:
```lua
function safe_pairs (t)
    local ok, iter, t = pcall(pairs, t)
    function safe_iter(t, k)
        local ok, next_k, v = pcall(iter, t, k)
        if ok then return next_k, v
        else return nil
        end
    end
    return safe_iter, t, nil
end
```

TymurGubayev: Now that I actually look at the code side by side, my version is basically slightly refactored `_wrap_iterator` from above without vararg support (is it needed?). That's a pity.

library/lua/dfhack.lua
```diff
@@ -162,24 +162,49 @@ NEWLINE = "\n"
...
+    end
+    return pcall(iterate)
+end
+
+local function print_element(k, v)
+    dfhack.println(string.format("%-23s\t = %s", tostring(k), tostring(v)))
+end
+
+function printall(table)
+    safe_iterate(table, pairs, print_element)
```
lethosor, line 201: +1 for keeping this to avoid duplication with printall_ipairs

lethosor (approved):
Thanks for addressing this!

library/lua/dfhack.lua
lethosor: Vararg support was mostly me taking a shortcut, since we don't do anything with the parameters other than passing them through. Per https://www.lua.org/manual/5.3/manual.html#3.3.5, `next()` (`iter` in your example) is only ever called with 2 parameters, but it can return any number of values, which is why I collected all return values from calling it in my implementation. `pairs()` and `ipairs()` only take one value, but another iterator could potentially take more, which is why I made `_wrap_iterator()` take varargs.

PR #1921, opened 2021-08-10
[dreamfort] revise according to playtesting and feedback

myk002:
- Improvements to documentation:
  - Editing pass for clarity and correctness
  - Added info on common tasks that Dreamfort does not do (like manufacturing trade goods)
  - Added hyperlinks to screenshots and related info
  - Removed complicated instructions and warnings that are now handled automatically by the `prioritize` script
  - Added a reminder to have fun : ) (it's not all about spreadsheets and optimization!)
- Improvements to [automation orders](https://drive.google.com/drive/folders/196RS9d-tis7AoOw1t2EzLF6eRWqYwnEm):
  - Replaced the error-prone "do X when Y finishes" orders with proper item count-based conditions (depends on #1906)
  - Added glassmaking products
  - Added weapon/armor orders that create the best equipment possible based on the metals you have on hand
  - Orders are split into groups based on when you should import them so you gain the benefits incrementally without ever overloading your fort with automation tasks
- Improvements to [onMapLoad.init](https://drive.google.com/file/d/1BmUXzhu60gcU_QxeYGoiHpPkCAkXIFj8/view?usp=sharing) -- I uploaded this file separately from the [existing onMapLoad.init](https://drive.google.com/file/d/16JX21nSe0_t84BgoL7DJX1OVD2E6pJgH/view?usp=sharing) so users of the current release don't get error messages about missing scripts
  - Added configuration for the [new prioritize script](https://github.com/DFHack/scripts/pull/317)
  - Refined autobutcher settings
  - Banned booze from being cooked by default
  - Enabled and configured seedwatch for seed stock management
- new /setup blueprint that sets basic settings (players can change preferences and make further customizations after running this blueprint):
  - The manager, chief medical dwarf, broker, and bookkeeper noble roles are assigned to the first suggested dwarf. This is likely to be the expedition leader, but the game could suggest others if they have relevant skills.
  - Bookkeeping is set to the highest precision
  - Standing orders are set to:
    - only farmers harvest
    - gather refuse from outside (incl. vermin)
    - no autoloom (we'll be managing cloth production with automated orders)
  - A burrow named "Inside" is created (it's up to the player to define the area, though). An alert named "Siege" is also created and associated with the "Inside" burrow, intended for use in getting your civilians to safety during sieges.
  - Military uniforms get the following modifications:
    - all default uniforms set to replace clothing
    - all default uniforms get a leather cloak
    - the "Metal armor" uniform, the "metal armor" item is removed and replaced by "metal breastplate" and "metal mail shirt"
    - in the "Metal armor" uniform, the "metal legwear" item is removed and replaced by "metal greaves"
  - Hotkeys are created for the 8 most interesting levels. Only the names are set -- it's up to the player to adjust them to actual locations in the (H) menu since the blueprint can't know where the levels will eventually be built.
- Updated [embark suggestions](https://docs.google.com/spreadsheets/d/13PVZ2h3Mm3x_G1OXQvwKd7oIR2lK4A1Ahf6Om1kFigw/edit#gid=149144025) and [example embark profile](https://docs.google.com/spreadsheets/d/13PVZ2h3Mm3x_G1OXQvwKd7oIR2lK4A1Ahf6Om1kFigw/edit#gid=1727884387) based on playtesting results. In particular, Dreamfort now recommends raising geese for bones and leather instead of turkeys.
- Added a [list of useful professions](https://docs.google.com/spreadsheets/d/13PVZ2h3Mm3x_G1OXQvwKd7oIR2lK4A1Ahf6Om1kFigw/edit#gid=367414128) (collections of associated labors) to assign to your dwarves. [Uploaded corresponding files](https://drive.google.com/drive/folders/1C-mxqGWmSAQGY-QFuPAfuM_TKY5O7MG7) that can be used with DFHack manipulator to the [shared Dreamfort directory](https://drive.google.com/drive/folders/1iS90EEVqUkxTeZiiukVj1pLloZqabKuP).
- Improvements to the [checklist](https://docs.google.com/spreadsheets/d/13PVZ2h3Mm3x_G1OXQvwKd7oIR2lK4A1Ahf6Om1kFigw/edit#gid=1637828091) -- I kept this as a separate sheet for now so I don't confuse players trying to use the existing checklist with the currently released version of Dreamfort and DFHack. I'll overwrite the old checklist with this new one once we get closer to the next DFHack release:
  - Added links to onMapLoad.init and the new automation JSON files
  - Added line for the new /setup blueprint
  - Removed guidance that players should run the dig blueprints separately, but could run the (hidden) /dig_all_underground blueprint if conditions allow. Added guidance that players should run the (renamed and unhidden) /dig_all blueprint unless conditions disallow, with instructions on how to run the individual blueprints if necessary
  - added in calls to `prioritize ConstructBuilding` where especially useful
  - Added a "Plumbing" section where it would be good for the player to think about filling wells and finding/pumping magma
  - Added orders import lines where appropriate for importing the newly split automation orders
  - Moved building of the suites level earlier in the sequence, based on results of playtesting
- Surface level improvements:
  - The central stairwell has been changed from priority 7 to priority 6 so that miners who have hauling labors enabled will choose to dig rather than haul. Priority 6 is still low enough so that the miners will finish digging the current level before digging the stairs down to the next level.
  - Trees are now cut down at priority 1 so that woodcutters will voluntarily choose to woodcut without manual labor twiddling
  - starting cloth stockpile now also accepts refuse
  - starting food stockpile now allows all types of food/booze and also allows containers (barrels/pots)
  - on the roof, build order for the floor tiles around the access hatch is more carefully managed so that the floor tile to the north of the access hatch isn't built first, which would cause it to immediately collapse as it would not be properly supported
  - small clusters of traps are added to the north of the fortress to catch wildlife that path around the exterior walls
- Farming level improvements:
  - Stockpiles now take from starting stockpiles so even if the player forgets to remove the starting stockpiles, items will still be appropriately moved down to farming
- Industry level improvements:
  - The "miscliquids" stockpile (i.e. lye and quicklime) is now halved in size but now accepts containers
  - Added wood stockpile to indicate whether the stock of wood is depleted
  - Moved meltables stockpiles to the left to allow more room for magma furnaces/forges
- Services level improvements:
  - Moved the hospital well earlier in the build sequence to allow it to be used ASAP
  - Enabled animal training for the hospital zone so the dwarfvet plugin can use it as a veterinary hospital
  - Created inactive pond zones over the cistern tiles to assist players who fill the wells via bucket brigade. The inactive zones won't interfere with other strategies for filling the wells.
  - Redrew the cistern dig map and reworked the priorities used to account for the reduced priority of the central stairs


docs/changelog.txt
```diff
@@ -43,6 +43,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `tiletypes-here`, `tiletypes-here-point`: add --cursor and --quiet options to support non-interactive use cases
 
 ## Documentation
+- `quickfort`: Dreamfort blueprint set improvements: extensive revision based on playtesting and feedback. includes updated supporting ``onMapLoad.init`` file and automation orders JSON files. see full changelog at https://github.com/DFHack/dfhack/pull/1921
```
lethosor, line 46: Is this the right section? Or do you want it under "Misc Improvements", maybe? Or split? (Asking because it looks like a lot more than just documentation changes)

lethosor (approved):

docs/changelog.txt
myk002: I'm, um, not sure that ended up in the documentation section. I intended it for `Misc Improvements`. Slip of the arrow keys? Unintended consequence of a rebase? Dunno. But will fix.

What are your thoughts on where the supporting files are stored? Would the automation orders, sample `onMapLoad.init`, and sample professions files be better off distributed with DFHack? I designed them to be useful beyond just Dreamfort, and I'm starting to think they would benefit from version control, but that doesn't mean that they have to be part of the DFHack package.

If we do add them to the DFHack repo, I'm not suggesting that they be loaded automatically. The `onMapLoad.init` file is too opinionated for that, especially with the autobutcher settings. They could be added to an `extras` directory or something, though.

lethosor: It makes sense to me for them to be tracked. Unsure if there's a good existing location for them, though. `hack/examples` maybe? We have a bit of a special case with example raws currently, with `plugins/raw/*.txt` getting installed to `hack/raw/`, but this wouldn't really fall into the same category (just noting something similar we have set up already).

ab9rf: I like the idea of `hack/examples`.

myk002: That works for me, as long as we make it clear that they are actually ready-to-use, not just toy examples or syntax examples

myk002: Ok, I'll merge this PR as-is and address the additional files in a new PR

PR #1922, opened 2021-08-10
update autogems to respect workshop profiles

### LINKED ISSUE #1263: Autogems should respect workshop profile
I have two jeweler's workshops. One I specifically want used for gem cutting, and the other I want used for gem setting. If you have a manager, you can set the type of job each workshop is used for in the workshop profile. Specifically, I can disable gem cutting for one workshop, and disable gem setting for the other workshop. Even doing this though, autogems will still send cut gem jobs to both workshops. It would be nice if autogems respected this setting, or if there were another way to tell it to only send jobs to specific workshops.

ab9rf:
Add a test so that autogems will not assign cut gem jobs to workshops that are not allowed to do them.

Fixes #1263


lethosor (approved):
Thanks for handling this!

PR #1923, opened 2021-08-10
Fix out-of-bounds access in memscan.field_offset()

### LINKED ISSUE #1907: devel/export-dt-ini crashes DF on 64-bit Windows
Using DFHack 0.47.05-r2 on win64, running `devel/export-dt-ini` crashes DF immediately. There is nothing in `stderr.log` after `Invoking: devel/export-dt-ini`.

lethosor:
Fixes #1907

### DIFF
--- a/library/lua/memscan.lua
+++ b/library/lua/memscan.lua
@@ -310,7 +310,7 @@ function field_ref(handle,...)
 end
 
 function field_offset(type,...)
-    local tmp = df.new('intptr_t')  -- pointer to nullptr
+    local tmp = df.new('int8_t', df.sizeof(type))
     local _, haddr = df.sizeof(tmp)
     local handle = df.reinterpret_cast(type, tmp)
     local _, addr = df.sizeof(field_ref(handle,...))


ab9rf (approved):
My testing indicates that this resolves the issue on win64. I have not tested on win32 or on linux.

Also, I think I understand why this fixes the bug, and so I'm good to go with this. Mostly. :)

lethosor:
I'd also be curious if the INI you get on Win64 matches the one from https://github.com/Dwarf-Therapist/Dwarf-Therapist/blob/master/share/memory_layouts/windows/v0.47.05_graphics_win64.ini

(@cvuchener I also want to tag you on this in case you have input)

ab9rf:
> I'd also be curious if the INI you get on Win64 matches the one from https://github.com/Dwarf-Therapist/Dwarf-Therapist/blob/master/share/memory_layouts/windows/v0.47.05_graphics_win64.ini

There are two differences: the version line is different, and the generated file has an extra blank line at the bottom:
```
3c3
< version_name=v0.47.05 (graphics) win64
---
> version_name=v0.47.05 SDL win64
449d448
< 
```

lethosor:
The one with "SDL" is ours, so looks to me like it's missing a blank line relative to the one included in DT. But in any case, the important stuff is the same, so that's good!

https://www.githubstatus.com/ also claims that everything is up again, so I'll see if re-running CI works this time.

cvuchener:
These differences are expected. It's OK.

PR #1924, opened 2021-08-11
Documentation: Compilation instructions update for Apple Silicon

chrismdp:


docs/Compile.rst
```diff
@@ -338,6 +338,33 @@ from your GCC version and distribute that too, it will fail on older OS X
...
+
+In your terminal, ensure you have your path set to the correct homebrew in
+addition to the normal ``CC`` and ``CXX`` flags above::
+
+  export PATH=/usr/local/bin:$PATH
+
+In order to ignore certain warnings present in later versions of ``gcc``,
+ensure you pass the following flag to ``cmake``::
+
+  -DCMAKE_CXX_FLAGS="-fpermissive"
```
lethosor, line 362: Do you have examples of these warnings? What version of GCC are you using? DFHack compiles cleanly for me on Linux with GCC 10.3 without `-fpermissive`

lethosor: GCC 11.1 also produces a clean build for me, with all CMake options turned on that I think build extra DFHack code (BUILD_STONESENSE, BUILD_DEV_PLUGINS, BUILD_SKELETON).

docs/Compile.rst
```diff
@@ -338,6 +338,33 @@ from your GCC version and distribute that too, it will fail on older OS X
...
 recommended to use GCC 4.8 or 7.
 
+Notes for M1 users
+------------------
+
+Alongside the above, you will need to follow these additional steps to get it
+running on Apple silicon.
+
+Install an x86 copy of ``homebrew`` alongside your existing one. `This
+stackoverflow answer <https://stackoverflow.com/a/64951025>` describes the
```
lethosor, line 348: I believe you need a `_` or `__` after the link here. I usually go with the latter because it avoids issues with using the same link more than once.

```suggestion
stackoverflow answer <https://stackoverflow.com/a/64951025>`__ describes the
```

(Since it seems like you might be familiar with RST: we've [configured](https://github.com/DFHack/dfhack/blob/0aeb10784431db91c0e0186829fd627c2134f1e2/conf.py#L281) our default role as `ref`, so single backticks without an extra prefix or suffix always point to named anchors internal to our docs)

docs/Compile.rst
chrismdp: Sure:

```
$ cmake .. -G Ninja -DBUILD_DEV_PLUGINS=ON -DBUILD_SKELETON=ON -DBUILD_STONESENSE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=~/code/df/LMP-0.47.05-dfhack-r1/DF\ v0.47.05 -DDFHACK_BUILD_ARCH=64
$ ninja install -j3
[267/664] Building CXX object plugins/embark-assistant/CMakeFiles/embark-assistant.dir/embark-assistant.cpp.o
FAILED: plugins/embark-assistant/CMakeFiles/embark-assistant.dir/embark-assistant.cpp.o
/usr/local/bin/g++-11 -DDFHACK64 -DLINUX_BUILD -DLUA_BUILD_AS_DLL -DPROTOBUF_USE_DLLS -D_DARWIN -D_GLIBCXX_USE_C99 -D_GLIBCXX_USE_CXX11_ABI=0 -Dembark_assistant_EXPORTS -I/Users/chris/code/dfhack/depends/protobuf -I/Users/chris/code/dfhack/depends/lua/include -I/Users/chris/code/dfhack/depends/md5 -I/Users/chris/code/dfhack/depends/tinyxml -I/Users/chris/code/dfhack/depends/lodepng -I/Users/chris/code/dfhack/depends/tthread -I/Users/chris/code/dfhack/depends/clsocket/src -I/Users/chris/code/dfhack/depends/xlsxio/include -I/Users/chris/code/dfhack/library/include -I/Users/chris/code/dfhack/library/proto -I/Users/chris/code/dfhack/plugins/proto -I/Users/chris/code/dfhack/library/depends/xgetopt -fvisibility=hidden -mtune=generic -m64 -mno-avx -O2 -g  -g -Wall -Wno-unused-variable -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk -fPIC -std=c++11 -MD -MT plugins/embark-assistant/CMakeFiles/embark-assistant.dir/embark-assistant.cpp.o -MF plugins/embark-assistant/CMakeFiles/embark-assistant.dir/embark-assistant.cpp.o.d -o plugins/embark-assistant/CMakeFiles/embark-assistant.dir/embark-assistant.cpp.o -c /Users/chris/code/dfhack/plugins/embark-assistant/embark-assistant.cpp
In file included from /Users/chris/code/dfhack/plugins/embark-assistant/embark-assistant.cpp:10:
/Users/chris/code/dfhack/plugins/embark-assistant/../uicommon.h: In function 'std::string int_to_string(int)':
/Users/chris/code/dfhack/plugins/embark-assistant/../uicommon.h:163:59: error: taking address of rvalue [-fpermissive]
  163 |     return static_cast<ostringstream*>( &(ostringstream() << n) )->str();
      |                                          ~~~~~~~~~~~~~~~~~^~~~~
[269/664] Building CXX object plugins/embark-assistant/CMakeFiles/embark-assistant.dir/finder_ui.cpp.o
```

lethosor: Thanks! Was this just with a debug build or also with a release build?

If this is the only issue, we should probably just change `int_to_string` and avoid the need to change compiler flags. To me from a glance, it looks unnecessarily complicated.

ab9rf: Rewriting it as `return static_cast<ostringstream&>( (ostringstream() << n) ).str();` should avoid this problem. The current code appears to be a standards violation, although apparently neither gcc nor MSVC seem to mind for some reason.

See also [this discussion on stackoverflow](https://stackoverflow.com/questions/8287188/stdostringstream-printing-the-address-of-the-c-string-instead-of-its-content).

lethosor: What's the purpose of the `.flush()`? It was absent before, and on my end (Linux with GCC 10) it seems to compile and behave properly with just `return static_cast<ostringstream&>( (ostringstream() << n) ).str();`.

Also worth noting that the issue seems specific to GCC 11 (or maybe something close) on macOS - GCC 11 on Linux didn't produce this error for me.

ab9rf: > What's the purpose of the `.flush()`? It was absent before, and on my end (Linux with GCC 10) it seems to compile and behave properly with just `return static_cast<ostringstream&>( (ostringstream() << n) ).str();`.
> 
> Also worth noting that the issue seems specific to GCC 11 (or maybe something close) on macOS - GCC 11 on Linux didn't produce this error for me.

It's possibly necessary to make sure that the compilers selects the correct overload. See the stackoverflow article I linked above. I'm not sure it's required in this situation. I suspect C++ will choose the correct overload without it, but `ostream` has _so_ many overloads that it can be difficult to predict what the compiler will do. In any case, it's fairly harmless in this situation.

lethosor: Yeah, I admittedly haven't tried it without flush() on other platforms (who knows how MSVC would behave). I'll merge this and spin up another PR to test out a replacement.

lethosor (approved):

docs/Compile.rst
lethosor: Interesting - I am able to reproduce the compilation error on Godbolt with GCC 11.2, but _not_ GCC 11.1. I'm using GCC 11.1 locally, so I guess something changed between the two. Unusual for a minor release, but since homebrew tends to include the latest release of everything, that makes sense.

ab9rf: > Yeah, I admittedly haven't tried it without flush() on other platforms (who knows how MSVC would behave). I'll merge this and spin up another PR to test out a replacement.

I've tested it both ways on MSVC and it works either way.

PR #1926, opened 2021-08-20
Handle projectiles in Items module functions

myk002:
This allows `moveTo*()` and `remove()` functions to work with projectiles.

This solves the issue with the `build-now` script in DFHack/scripts#310 when an extra item remains in the game world when a construction is built on a tile that has no adjacent walkable tiles.

### DIFF
--- a/library/include/MiscUtils.h
+++ b/library/include/MiscUtils.h
@@ -310,6 +310,27 @@ Link *linked_list_insert_after(Link *pos, Link *link)
     return link;
 }
 
+/**
+ * Returns true if the item with id idToRemove was found, deleted, and removed
+ * from the list. Otherwise returns false.
+ */
+template<typename Link>
+bool linked_list_remove(Link *head, int32_t idToRemove)
+{
+    for (Link *link = head; link; link = link->next)
+    {
+        if (!link->item || link->item->id != idToRemove)
+            continue;
+
+        link->prev->next = link->next;
+        if (link->next)
+            link->next->prev = link->prev;
+        delete(link);
+        return true;
+    }
+    return false;
+}
+
 template<typename T>
 inline typename T::mapped_type map_find(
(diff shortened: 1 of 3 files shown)


library/modules/Items.cpp
```diff
@@ -832,6 +831,26 @@ static bool detachItem(MapExtras::MapCache &mc, df::item *item)
...
+        df::proj_list_link *link = world->proj_list.next;
+        for (; link; link = link->next)
+        {
+            if (link->item->id != proj_id)
+                continue;
+
+            link->prev->next = link->next;
+            if (link->next)
+                link->next->prev = link->prev;
+            delete(link);
```
lethosor, line 847: I found `linked_list_append()` and `linked_list_insert_after()` in `MiscUtils.h` - maybe we could also use a `linked_list_remove()`?

lethosor (approved):

library/modules/Items.cpp
myk002: done.

library/include/MiscUtils.h
```diff
@@ -310,6 +310,27 @@ Link *linked_list_insert_after(Link *pos, Link *link)
     return link;
 }
 
+/**
+ * Returns true if the item with id idToRemove was found, deleted, and removed
+ * from the list. Otherwise returns false.
+ */
+template<typename Link>
+bool linked_list_remove(Link *head, int32_t idToRemove)
```
lethosor, line 318: Ah, I didn't consider that this is difficult to implement generically. I think it's fine to limit it to the `id` field for now (given that the sole caller works with that) - if we need it to be more generic, the pointer-to-member approach used in [this overload of linear_index()](https://github.com/DFHack/dfhack/blob/5750bfff9b2275733df207f037bd8a04e2c2139f/library/include/MiscUtils.h#L160-L167) could work.

PR #1929, opened 2021-08-21
Fix compilation error on macOS with GCC 11.2

lethosor:
This should make `-fpermissive` (from #1924) unnecessary, so I also removed those instructions.

@chrismdp could you confirm that this works for you without `-fpermissive`?

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -338,6 +338,8 @@ from your GCC version and distribute that too, it will fail on older OS X
 versions.) For this reason, if you plan on distributing DFHack, it is highly
 recommended to use GCC 4.8 or 7.
 
+.. _osx-m1-notes:
+
 Notes for M1 users
 ------------------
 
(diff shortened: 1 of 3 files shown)


lethosor:
From https://github.com/DFHack/dfhack/pull/1924#discussion_r693378961, turns out the difference between GCC 11.1 and 11.2 is relevant here.

Experimenting on godbolt.org a bit with x86 GCC 11.1 (`-Wl,-z,defs -O3 -DNDEBUG -fPIC -std=c++11`) suggests that the old and new implementations are pretty much the same.

```c++
#include <sstream>
#include <string>

using namespace std;

std::string int_to_string(const int n)
{
    std::ostringstream ss;
    ss << n;
    return ss.str();
}

std::string int_to_string_old(const int n)
{
    return static_cast<ostringstream*>( &(ostringstream() << n) )->str();
}
```

ab9rf (approved):
I endorse this implementation. Terseness is not always a virtue.

chrismdp:
Sorry for the delay: can confirm the build now passes without `-fpermissive`. I now get one warning (which may also be present on other platforms):

```
[186/336] Building CXX object plugins/CMakeFiles/buildingplan.dir/buildingplan.cpp.o
/Users/chris/code/dfhack/plugins/buildingplan.cpp: In static member function 'static bool buildingplan_query_hook::is_filter_satisfied(df::building*, int)':
/Users/chris/code/dfhack/plugins/buildingplan.cpp:505:51: warning: comparison of integer expressions of different signedness: 'std::vector<df::job_item*>::size_type' {aka 'long unsigned int'} and 'int' [-Wsign-compare]
  505 |                 || bld->jobs[0]->job_items.size() <= filter_idx)
      |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
```

lethosor:
Hm, that's not present for me with GCC 11.1 either, but it should be easy enough to fix. It does show up in [Buildmaster logs](https://buildmaster.lubar.me/executions/execution-details?executionId=933447), so it's probably something that only shows up in Debug or RelWithDebInfo builds (due to different warnings being enabled for those).

ab9rf (commented):
Wouldn't it be more appropriate to retype `filter_idx` to an unsigned type?

lethosor:
Perhaps. I am a little wary of making that change due to the possibility of underflow, though, and I haven't screened all uses of `filter_idx` to make sure that they would handle underflow correctly.

chrismdp:
@lethosor thanks! 👏🏼

PR #1932, opened 2021-08-24
[test] add pattern matching expect.find()

myk002:
Adding a pattern-matching function to `expect` since I got tired of writing tests like `expect.true_(mock_print.call_args[1][1]:find('^Prioritized 2'))`. This works fine when the test passes, but when it fails it doesn't give you any information about why it failed. The new `expect.find()` function will report the pattern and the string it failed to match.

I settled on `find()` since the string method of the same name has the same semantics.

Other options might be: `str_find()` since it fails on non-string or `match()` (but that might imply that the pattern needs to match the entire string).

No changelog entry for PRs like this, right?

### DIFF
--- a/test/library/test_util/expect_unit.lua
+++ b/test/library/test_util/expect_unit.lua
@@ -1,18 +1,18 @@
 local expect_raw = require('test_util.expect')
 
 function test.find()
-    expect.true_(expect_raw.find('a ', 'a str', 'a comment'))
+    expect.true_(expect_raw.str_find('a ', 'a str', 'a comment'))
 
-    local ok, comment, msg = expect_raw.find('ab', 'a str', 'a comment')
+    local ok, comment, msg = expect_raw.str_find('ab', 'a str', 'a comment')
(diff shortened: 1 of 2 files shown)


lethosor (approved):
+1 for more descriptive failures, and agreed that it doesn't need to be in the changelog.

I might lean towards `str_find` for the name, since we already have `table_eq` and `error_match` that are specialized for other types.

myk002:
> +1 for more descriptive failures, and agreed that it doesn't need to be in the changelog.
> 
> I might lean towards `str_find` for the name, since we already have `table_eq` and `error_match` that are specialized for other types.

done

test/library/test_util/expect_unit.lua
```diff
@@ -1,6 +1,6 @@
 local expect_raw = require('test_util.expect')
 
-function test.find()
+function test.str_find()
```
lethosor, line 3: You just beat me to it :)

PR #1938, opened 2021-08-26
autochop estimates [WIP]

### LINKED ISSUE #876: Make autochop estimate the number of logs you get from a tree
Instead of periodically designating all trees in the selected burrows, designate enough trees to produce enough wood to get to `max_logs`, or all of them if there aren't enough.

Moth-Tolias:
implements/resolves #876. marked as draft because we still need to complile and test, and also figure out a way to do the second part of #876, which is to prioritise older / higher yield trees.

### DIFF
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -18,6 +18,8 @@
 #include "df/map_block.h"
 #include "df/material.h"
 #include "df/plant.h"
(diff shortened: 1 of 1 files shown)


Moth-Tolias:
buggered up the repo somehow. trying to fix it.

PR #1939, opened 2021-08-26
update Job.cpp to use the DF-provided job cancel method

### LINKED ISSUE #1935: autochop may not be cancelling FellTree jobs correctly
In a fairly long-running fort (~6 years) I had a problem with labormanager allocating a ridiculous number of dwarves to treecutting. On investigation, this was because there were around 600 `FellTree` jobs on the job posting list that were active but not assigned to any dwarf. These jobs do not appear in the jobs screen (which I determined is because they have job flag `anon_2` set, and the job screen doesn't display jobs on the posting list with `anon_2` set unless they are assigned to a unit) and are apparently not assignable. It is my best guess that these "orphan" jobs are the result of tree cutting designations that were cancelled by `autochop`.

Under normal circumstances, when a tree is marked for chopping, the game creates a job posting (with `anon_2` set). This posting is then removed entirely when the job is assigned to a dwarf or when the designation is cancelled. Once the job is assigned to a dwarf, the job appears only on the job linked list. 

However, when `autochop` cancels a tree chop designation that has not yet been assigned to a dwarf, it seems it leaves the posting behind in at least some circumstances. The game doesn't try to assign these "orphan" postings to dwarves (for reasons I do not as yet know, haven't found that code yet, and I could not identify any obvious differences between the unassignable orphans and jobs that were created through the game interface and which could and eventually were assigned to units), but they're still on the posting list, which both makes labormanager malfunction and probably also causes tick lag (the posting list is enumerated frequently, and so cruft on it will presumably slow the game down).

I did find that I could make them go away by changing them to `Clean` jobs, at which point the next process-jobs run made them disappear.

My conclusion is that there's something not quite right about our job cancel logic, presumably in `Job::removeJob`. However, I don't have an immediate solution.

ab9rf:
Use DF's provided job cancel function instead of the hackadoo workaround we are currently using.

I've tested this fairly extensively and it works as least as well as our old approach.

Should close #1935.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -46,6 +46,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `EventManager`: add new event type ``NEW_UNIT_ACTIVE``, triggered when a new unit appears on the active list
 - `EventManager`: now each registered handler for an event can have its own frequency instead of all handlers using the lowest frequency of all handlers
 - `stocks`: allow search terms to match the full item label, even when the label is truncated for length
+- `dfhack-examples-guide`: add mugs to ``basic`` manager orders
 
 ## Documentation
 - Add more examples to the plugin skeleton files so they are more informative for a newbie
--- a/docs/guides/examples-guide.rst
+++ b/docs/guides/examples-guide.rst
@@ -63,7 +63,7 @@ This collection of orders handles basic fort necessities:
 - prepared meals and food products (and by-products like oil)
 - booze/mead
 - thread/cloth/dye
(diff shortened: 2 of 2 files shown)


lethosor:
Have you been able to verify whether this addresses #1935?

ab9rf:
I played a test fort with autochop enabled for several seasons and did not see the same job posting anomalies, or indeed any anomalies

myk002:
Could you resolve the merge conflicts and add a changelog entry (probably in the "Internals" section)?

PR #1953, opened 2021-09-04
Cleanup Kitchen module (and the seedwatch plugin which uses it)

quietust:
I started this cleanup over a year ago (with the intent to eventually get rid of the Kitchen module entirely, since there's literally only one plugin that uses it), and then I forgot about it.

May as well get it merged as-is, assuming everything still actually works.

### DIFF
--- a/library/include/modules/Kitchen.h
+++ b/library/include/modules/Kitchen.h
@@ -43,12 +43,6 @@ namespace DFHack
 {
 namespace Kitchen
 {
-const unsigned int seedLimit = 400; // a limit on the limits which can be placed on seeds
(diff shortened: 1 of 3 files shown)


myk002:
is there any change in behavior, or is this just a pure code refactor? Would it be possible to create a `fortress`-mode functional test to verify that it works as expected? The test could use createitem to generate seeds over the threshold and check to ensure the cookability is set correctly (then it should destroy the created seeds to avoid borking the game it is run in).

quietust:
It is indeed intended to be purely a refactor with no actual changes in behavior.
I'm sure _somebody_ could write functional tests for it, but at the moment that's probably not going to be me.

PR #1955, opened 2021-09-05
remove fortplan and unify buildingplan

### LINKED ISSUE #1644: turn down fortplan and digfort
The quickfort script is positioned to replace fortplan and digfort, but we should have a proper plan for deprecating and removing them so their current users aren't caught by surprise. Here is a possible plan (open for discussion and changes):

1) do nothing until quickfort has been "proven" for at least one release and no significant bugs are open against it
2) advertise that fortplan and digfort are deprecated:
- output a deprecation warning when they are run, pointing people to quickfort
- update their documentation with the deprecation warning and pointer to quickfort
- add a deprecation warning and pointer to quickfort to their source files
3) after a good amount of time (one year?), remove fortplan and digfort source but leave documentation stubs that announce what has happened and provide a pointer to quickfort
4) after more time (or maybe never), remove the documentation stubs for fortplan and digfort

edit: step 4 is not necessary. we have Removed.rst for this purpose

myk002:
Fixes #1644 

Removes fortplan code but leaves a doc stub reminding people what happened and how to migrate to `quickfort`

This PR also un-libifies `buildingplan` to prevent future misuse (starting buildingplan from two separate plugins just runs two completely separate engines with separate state, one of which is completely inaccessible from the UI).


myk002:
I'm ignoring doc build errors since there doesn't seem to be a clean way to get both this PR and DFHack/scripts#332 to pass the check at the same time.

myk002:
All build checks pass (including all targets on the buildmaster). Merging.

PR #1958, opened 2021-09-07
Enable -Wall and -Werror for all build types (GCC)

lethosor:
Background: when PRs introduce warnings, they are only flagged by Buildmaster, and often go unnoticed until PRs are merged. Sometimes these warnings turn out to be important. In some cases, they are just signed/unsigned comparison issues that wouldn't cause issues in practice, but if we enforce a warning-free build, we will have a better chance of catching problematic warnings sooner. Some recent examples: a083a7f67031504758fac53127b41989fa2db3ce, 1d9f12fe3fc7d2844a78c427362a9021e2888e7f, 572b6ebab5aa3e6fea24a5fef57b2db2f129207c.

Due to our past efforts to make the Buildmaster build warning-free, all of our code currently compiles with `-Werror` (i.e. there weren't any warnings introduced by enabling `-Wall` for release builds). I'm planning to experiment with more strict flags (e.g. adding `-Wextra` and removing `-Wno-unused-variable`) to see how well those work.

Note: this doesn't change MSVC builds (yet?)


myk002 (approved):
builds cleanly on Gentoo x86_64 with gcc-10.3.0

warmist (approved):
Looks good to me

ab9rf:
I'm going to look to see what is needed, if anything, to make this change apply to MSVC builds. Sorry I didn't get to this sooner.

lethosor:
I took a look at https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-160, and I think `/WX` along with something between `/W2` and `/W4` would do the trick. If `/W4` is more similar to GCC's `-Wextra`, we might want to go with `/W3` or lower for now. (`-Wextra` produces a *lot* of warnings, and I've only made a bit of progress towards getting that to compile cleanly: https://github.com/lethosor/dfhack/tree/gcc-wextra)

There is some potential for devs using GCC to be inconvenienced if there are warnings (or errors) that only MSVC throws, but the opposite is true now that the GCC flags have been updated. I'm not really familiar with what sorts of things slip through the cracks in just one compiler. In general, I'd be wary of making one compiler's flags significantly more strict than the other's (e.g. if `-Wextra` and `/W4` are roughly equivalent, we should switch to them at the same time).

ncfavier:
With this change, I can't build external plugins that have warnings anymore, for example dfplex:

```
/build/source/plugins/dfplex/command.cpp: In function 'void apply_special_case(Client*, std::set<df::enums::int>
/build/source/plugins/dfplex/command.cpp:708:35: error: unused variable 'vs_civ' [-Werror=unused-variable]
  708 |         df::viewscreen_civlistst* vs_civ = static_cast<df::viewscreen_civlistst*>(vs);
      |                                   ^~~~~~
/build/source/plugins/dfplex/command.cpp: In function 'void apply_command(std::set<df::enums::interface_key::in>
/build/source/plugins/dfplex/command.cpp:853:16: error: unused variable 'at_root' [-Werror=unused-variable]
  853 |     const bool at_root = is_at_root();
      |                ^~~~~~~
[...]
```

Is there an easy workaround?

ncfavier:
Would adding `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error")` at the end of `plugins/CMakeLists.txt` before we import any external code be acceptable?

myk002:
External plugins should all be in the `plugins/external` directory. In `plugins/external/CMakeLists.txt`, you can add a line like the following:
`set_target_properties(dfplex PROPERTIES COMPILE_FLAGS "-Wno-unused-variable)`

myk002:
you can see how we do this for some dependencies in `depends/CMakeLists.txt`

ab9rf:
I just tried adding `/WX` and `/W4` for MSVC builds but this yielded 269 build errors which is far too many to deal with right now.

Subbing out `/W3` for `/W4` yielded a clean build. Recommend the following change, in `CMakeLists.txt`:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2e6f5c835..7c2cb36fc 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)

 if(MSVC)
+    # increase warning level and cause warnings to be treated as errors
+    add_definitions("/WX")
+    add_definitions("/W3")
+
     # disable C4819 code-page warning
     add_definitions("/wd4819")
```

Later, when I have more time, I'll capture the `/W4` error list and we can look at it to see if we even want to try to mitigate them.

PR #1972, opened 2021-09-25
[argparse] add support for standalone long-form params in processArgsGetopt()

myk002:
Implement handling for long form parameters that are not an alias for a short form parameter. For example, you can now have a parameter like `--longparam` without needing to have an equivalent one-letter `-l` param.

This is useful for esoteric parameters that you don't want to waste a precious one-letter param on.

### DIFF
--- a/library/lua/argparse.lua
+++ b/library/lua/argparse.lua
@@ -80,6 +80,7 @@ end
 -- it will be passed to the handler function.
 -- longOptionAlias is optional.
 -- hasArg defaults to false.
+-- to have an option that has only a long form, pass '' as the shortOptionName.
 --
 -- example usage:
 --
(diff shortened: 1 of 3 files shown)


library/lua/argparse.lua
```diff
@@ -80,6 +80,7 @@ end
 -- it will be passed to the handler function.
 -- longOptionAlias is optional.
 -- hasArg defaults to false.
+-- to have an option that has only a long form, pass '' as the shortOptionName.
```
lethosor, line 83: I might suggest allowing or supporting `nil` as the sentinel here (with the caveat that `#optionAction` would behave unpredictably, but it looks like you're only accessing `optionAction[1]` and other fields directly)

lethosor (approved):

library/lua/argparse.lua
myk002: good idea. done.

PR #1973, opened 2021-09-29
WIP: Fix ipairs() for enum attrs

### LINKED ISSUE #1860: ipairs() called on enum attr tables never ends
To reproduce:
```
[lua]# @df.item_type
...
90                     	 = BRANCH
[lua]# @df.item_type.attrs
...
111655                 	 = <item_type._attr_entry_type: 0x7ffff7dcfca0>
...
```

This can be stopped with `kill-lua` if run with `dfhack-run` (since printall_ipairs() is pure Lua).

Reported by @wolfboyft on Discord

lethosor:
Fixes #1860

(Note: I had put this together but evidently never finished it - need to investigate what is missing)

Edit: looks like I didn't make the test pass - that sounds familiar to me now.

### DIFF
--- a/library/LuaWrapper.cpp
+++ b/library/LuaWrapper.cpp
@@ -1534,6 +1534,25 @@ static void FillEnumKeys(lua_State *state, int ix_meta, int ftable, enum_identit
         lua_pushcclosure(state, meta_enum_attr_index, 3);
 
         freeze_table(state, false, (eid->getFullName()+".attrs").c_str());
+        int ix_attrs = lua_gettop(state);
+
+        // add metamethods to metatable of .attrs
+        if (lua_getmetatable(state, ix_attrs))
+        {
+            int ix_attrs_meta = lua_gettop(state);
+
+            if (eid->getFirstItem() <= eid->getLastItem())
+            {
+                lua_pushvalue(state, ix_attrs);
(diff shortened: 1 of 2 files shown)


myk002:
this does address a legitimate issue. do you think you will be able to get this done within the release cycle?

lethosor:
Looking at this again. The Lua layer here is a bit complicated, so no promises.

lethosor:
I looked into this some. Turns out this is broken because I'm calling an existing ipairs() implementation that uses `rawget()`:
https://github.com/DFHack/dfhack/blob/f290b1c804f6207afc786b3a3ca83e8cd87c44d9/library/LuaWrapper.cpp#L1343

but the `attrs` table is not actually populated, and instead uses the `__index` method to return values, which `rawget()` ignores:
https://github.com/DFHack/dfhack/blob/f290b1c804f6207afc786b3a3ca83e8cd87c44d9/library/LuaWrapper.cpp#L1534-L1536

so a solution would likely involve a custom `ipairs()` implementation or changing `attrs` so that it is populated for all valid enum items.

By the way, the reason for this implementation is probably because indexing `attrs` with an invalid index returns a table of the default attribute values. So I am leaning towards the former as a solution (a custom `ipairs()` instead of reusing `wtype_ipairs()`).

ab9rf:
This is probably superseded by #5966

PR #1974, opened 2021-09-29
test.lua: Dump some DF globals

lethosor:
This should help identify differences in environments, e.g. what the screen resolution is in `PRINT_MODE:TEXT`. The `scripts/gui/blueprint:render_labels` test currently fails in an 80x25 resolution, which implies that our CI is using something else.


lethosor:
CI is using 80x24. `scripts/gui/blueprint:render_labels` just isn't running in CI because we don't have a save to run fortress mode tests in (related: #1840)

PR #1982, opened 2021-10-04
[blueprint] implement zone phase

myk002:
part of milestone 4 of #1842 

implements "zone" phase that discovers and records activity zones, including custom configuration for ponds, gathering, and hospitals.

ecosystem integration tests for zones added.

this PR also fixes the docs for the hospital zone config, which mistakenly listed `powder` instead of `plaster`. The variable reference was fixed in DFHack/scripts#346


myk002:
I note that `Buildings::findCivzoneAt()` is inefficiently written, and is slowing down `blueprint` execution. I'll take a look at caching civzone lookups in a separate PR.

PR #1984, opened 2021-10-08
cache zones so Buildings::findCivzonesAt() is O(1) instead of O(N)

myk002:
tl;dr: This PR allows `Buildings::findCivzonesAt()` to execute in amortized constant time. The memory cost is proportional to the combined area of all civzones, but is not significant compared to the total memory usage of DF+DFHack.

Long version:
The `blueprint` plugin recently acquired the ability to create `#zone` blueprints to record the layout of civzones in a map area. When scanning for zones, `blueprint` will call `Buildings::findCivzonesAt()` for each tile in the transcription area. Each call to `findCivzonesAt()` has to scan every civzone in the game since there is no way to directly look up which civzones cover which tiles. This isn't so slow when the number of civzones is very small, but as the player allocates more civzones, the algorithm starts to crawl. With thousands of civzones, `blueprint` runtime goes from 200 milliseconds to more than 2 minutes! This also affects `quickfort` since it scans tiles for civzones in `#query` and `#zone` modes.

We can get our runtime back to 200 milliseconds by caching civzone locations in a hashmap so lookups are O(1) instead of O(N) in the total number of civzones. We do something similar for buildings so that `Buildings::findAtTile()` can return in constant time instead of linearly scanning through all buildings in the game. However, there are a few critical differences between buildings and civzones that make civzones more difficult to cache:

1. multiple civzones can overlap the same tile
2. civzone boundaries can be mutated without destroying and recreating the civzone
3. tiles have a flag that indicate whether there is a building covering the tile. there is no such flag for civzones.

We can address (1) by hashing a position to a vector of IDs (not a set, since we want to maintain the order so we know which is the "top" zone) instead of just an ID like the building cache does. We will have to do a linear walk of the vector, but we expect that vector to be much much smaller than the list of all civzones.

(2) is trickier. If civzones could be dynamically expanded, a cache would be useless since we'd always need to double check to see if a tile is newly covered by an existing civzone. Luckily, civzones can only be dynamically shrunk: tiles can be removed from a civzone, but new tiles cannot be added to an existing civzone. We can scan newly created civzones to cache their tiles, and when there is a cache hit, we can just verify that the civzones that we *think* cover that tile still cover that tile. If it turns out that a civzone no longer covers that tile, the cache can be updated at lookup time to reflect the new knowledge.

(3) is the kicker. Tiles have a flag that tracks whether any building covers the tile. If it is false, there is no need to scan the buildings list, since you already know you won't find anything. This allows the building cache to be best-effort without much loss in performance. Specifically, the building cache is updated by the `EventManager` module every 100 ticks. If a call is made to `findAtTile()` before the cache is updated, and the tile flag is true but the cache does not contain the target building ID, you *know* you need to do a linear scan to look for the building. If the flag is false, you can just skip it and not pay any performance penalty.

If we were to take the same approach for civzones, however, we would not be able to differentiate the absence of a civzone from a missing cache entry, so we would *always* need to do a full linear scan. And even if we get a cache hit, we can't be sure that our vector of ids is complete and we'd need to do a scan there too. In order to guarantee an O(1) lookup, civzones need an authoritative cache that guarantees a cache hit means all civzone ids that cover the tile are known, and a cache miss means there are exactly 0 civzones that cover the tile.

My first attempt at an authoritative cache involved forcing `EventManager` to fire `BUILDING` events immediately whenever a civzone was created. I tried two strategies to get this to work. First, I created a `forceEvent(etype)` API for `EventManager` and called it from places where buildings are created (e.g. `Buildings::constructAbstract()`). This caused a lot of events to fire when a large number of buildlings were created, which was inefficient. Moreover, it didn't catch civzones created manually via the DF UI.

My second attempt was to schedule the `BUILDING` event to fire on every tick. This was also inefficient, and also didn't catch the case where a civzone was created and then queried *without unpausing* (i.e. between ticks), which is a common case when running `quickfort`.

I concluded that `EventManager` was not suitable for my needs. However, I took note of how `EventManager` scans for new buildings, and I realized that if I could run that algorithm whenever `findCivzonesAt()` is called, then I could guarantee that all new civzones would be known to the cache before we attempted a lookup. I could also still use the `EventManager` events to keep the cache kinda up to date so `findCivzonesAt()` wouldn't have too much catch-up to do at any one time.

So the final implementation here is a degenerate form of the `EventManager` algorithm that just scans for new civzones. As mentioned in (2), we don't care about civzone deletion since we'll pick that up when we look at the tiles that it used to cover.

Testing has shown the code in this PR to work as advertised, but my understanding of zones may be incomplete. Feedback is welcomed.

Incidentally, I think there is a bug in the interaction between the `EventManager` `BUILDING` event and the building cache. If a planned-but-as-yet-unconstructed building is canceled and another building is built in the same spot, and both actions happen between `EventManager` `BUILDING` events, I believe the building cache will become inconsistent. This happens because `EventManager` reports both removed and newly built buildings in the same event, but it reports the new ones before the removed ones. This causes the building cache to ignore the new building (since it thinks it already has a building in that spot cached), but then it clears the cache at that position when it gets notified of the removed building. I believe we can fix this by reporting removed buildings before new ones in the `BUILDING` event, but I haven't investigated this fully yet, and I don't know if anything relies on the current ordering.

Also, we could modify the building cache to update itself in `findAtTile()` like this PR does for civzones. This would allow the building cache to be useful for buildings that have been created since the last `BUILDING` event.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `quickfort-library-guide`: added pump stack blueprints (with step-by-step usage guide) to the quickfort blueprint library
 - `quickfort`: Dreamfort blueprint set improvements: added iron and flux stock level indicators on the industry level
 
+## API
+- ``Buildings::findCivzonesAt()``: lookups now complete in constant time instead of linearly scanning through all civzones in the game
+
 ## Lua
 - ``argparse.processArgsGetopt()``: you can now have long form parameters that are not an alias for a short form parameter. For example, you can now have a parameter like ``--longparam`` without needing to have an equivalent one-letter ``-l`` param.
 - ``dwarfmode.enterSidebarMode()``: ``df.ui_sidebar_mode.DesignateMine`` is now a suported target sidebar mode
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -102,6 +102,7 @@ struct CoordHash {
 };
 
 static unordered_map<df::coord, int32_t, CoordHash> locationToBuilding;
+static unordered_map<df::coord, vector<int32_t>, CoordHash> locationToCivzones;
 
 static df::building_extents_type *getExtentTile(df::building_extents &extent, df::coord2d tile)
 {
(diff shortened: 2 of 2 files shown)


myk002:
> > For this test, I used `quickfort` to create 27648 zones over 9216 tiles (3 zones per tile).
> 
> Now I'm even more curious - with that many zones, what happens to the game's framerate when you hit `i`?

I finally got a chance to test this. The answer: absolutely nothing. FPS before creating zones (both in and out of `i` mode, both paused and unpaused): 100/50. FPS after creating zones (both in and out of `i` mode, both paused and unpaused): 100/50. DF doesn't seem to care that there are a lot of zones.

myk002:
I've been running this PR for some time now and haven't seen any issues. It gets tested regularly with the quickfort ecosystem tests, and there is a significant improvement in the performance of `quickfort` and `blueprint`. I'll merge once all build checks complete. If there are any complaints we can back it out.

PR #1989, opened 2021-11-07
hive-color.h

Vettlingr:
Hives are normally shown as yellow, regardless of tool material. As with nestboxes, this should change how the games show hives, making it possible to see what material the hives are made of.

Needs testing.

### DIFF
--- a/plugins/tweak/tweak.cpp
+++ b/plugins/tweak/tweak.cpp
@@ -103,6 +103,7 @@
 #include "tweaks/military-assign.h"
 #include "tweaks/pausing-fps-counter.h"
 #include "tweaks/nestbox-color.h"
+#include "tweaks/hive-color.h"
 #include "tweaks/shift-8-scroll.h"
 #include "tweaks/stable-cursor.h"
 #include "tweaks/stone-status-all.h"
@@ -237,6 +238,8 @@ DFhackCExport command_result plugin_init (color_ostream &out, std::vector <Plugi
         "    Allows assigning more than 3 wheelbarrows to a stockpile\n"
         "  tweak nestbox-color [disable]\n"
         "    Makes built nestboxes use the color of their material\n"
+        "  tweak hive-color [disable]\n"
+        "    Makes built hives use the color of their material\n"
         "  tweak military-color-assigned [disable]\n"
         "    Color squad candidates already assigned to other squads in brown/green\n"
         "    to make them stand out more in the list.\n"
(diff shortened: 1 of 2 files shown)


quietust:
This change looks like it will also affect hives which are queued for construction - the original version of this vmethod checks whether `getBuildStage()` is equal to 0 or 1 in order to decide between these.

A better way of fixing this would be to override the "getNameColor" vmethod (which sets `gps.screenf+`gps.screenb`+`gps.screenbright` to the desired material color (which should _also_ be done for the Nest Box tweak).

Vettlingr:
What getbuildStage() does not seem to be an issue for the nest-box tweak. Its graphical representation shows up as it should while under construction. Not sure why, but I think it has to do with how df handles colors. Wether or not this also applies to hives is a matter of testing though. 

I'm not sure on where to but in the getBuildStage()==1 to address this issue as I have no idea how vmethod works :S

Vettlingr:
I'm guessing by intuition     --    if (flags.bits.exists) and building.getBuildingStage()==1 then

But its a shot in the dark

Vettlingr:
Hi again. It might be an idea to do the same for cage traps, weapon traps and stone-fall traps?

lethosor:
I'm not quite sure why that would be desirable - the color of traps is already used to distinguish between the different types of traps.

Vettlingr:
You can change the graphics of the trap tile with TWBT through certain override lines.
That is the only reason.

myk002 (changes requested):
Have you been able to test this yet? Does the code need a getBuildStage() check? Your intuition is correct, I think, on where to put it.

Also, could you merge latest develop and add a changelog entry for this tweak?

Vettlingr:
I have not been able to test this yet, and I am unsure how it can be tested without compiling. I cannot compile properly on my mac for the moment, so I am just relying on the Nest-box code to be transferable to hives for that reason. Hence I don't have any interest on diverging from the already proven solution.
If quietust have certain preferences that deviate from the nest-box colour code, I would suggest him implementing that himself since I haven't figured out how to test this code yet.

myk002:
Where are you getting stuck on compiling? Do the instructions at https://docs.dfhack.org/en/stable/docs/Compile.html#macos need updating?

lethosor:
+1. We don't have a lot of people compiling on macOS, so we rely on those that do to help keep the instructions up-to-date. There isn't a way to test this other than compiling DFHack.

PR #1990, opened 2021-11-07
Added hive-color hook

Vettlingr:
For the new file hive-color.h similar to nest-box-color.h


lethosor:
This should be combined with #1989

PR #1991, opened 2021-11-08
Submitting first version of Enlinker to try to get feedback

UnknowableCoder:
After a few weeks of posting mostly idiotic questions out on the DFHack Discord, I finally managed to implement what I had long since wanted in DF: an alternative logic system that allows the user to set a threshold (and optional delay) for building activation and de-activation, providing endless new exciting possibilities for dwarfputing in general, on one hand, but also massively simplifying relatively simple things such as e. g. opening and closing a fortress "door" made of both bridges and floodgates through a single lever. Given my chronic lack of inspiration for good names, I call this Enlinker, but I'm sure we could figure out something better... 

Along the way, I decided it would be a good idea to store this information such that it would also work in Adventure Mode, so one could build e. g. a trap-filled maze for others to explore or an alternative entrance to a fortress that an adventurer could also use. This led me to develop some sort of system (still mostly experimental) that could load and save information based on the site-id, as suggested by the helpful folks over on the Discord.

The bulk of the user interaction will be done through a GUI script that I will be also submitting to consideration through the DFHack scripts repo, but here I present the C++ code that implements most of the logic.

It's structured as follows:

- `building_activation.h`: provides a generic way to check, set or toggle building activation for (almost?) all of the DF building types that can be activated. 
- `loctools.h`: holds a few useful functions for getting the current site, or a list of nearby sites
- `nysave.h`: implements part of the logic that enables location-dependent information to be stored
- `enlink.cpp`: implements the alternative logic system itself and the rest of the location-dependent information (which is tied to the `plugin_onupdate` callback as the adventurer location needs to be queried.)

I am well aware that parts of the code are done in a slightly obnoxious way and needs to be polished (especially in what regards the lack of separation between `nysave,h` and `enlink.cpp`), but it was suggested in the DFHack Discord that I start this pull request so I could get some feedback, which is why I am presenting this in its current state. Perhaps, as also discussed there, part of the functionality (namely what I am doing to save and load location-dependent data) can be added to DFHack itself with a more general-purpose API.

As a final note, the alternative logic system is currently highly unbalanced, as the links between buildings are created instantaneously and without any interaction from the dwarves. This means, for instance, one can link a lever to a floodgate submerged in magma halfway across the map instantaneously and not need to worry about magma-safe mechanisms. Personally, this does not concern me too much, but I am sure most players would disagree...


plugins/CMakeLists.txt
```diff
@@ -1,197 +1,198 @@
-include(Plugins.cmake)
```
myk002, line 1: might want to undo the line ending changes so this diff is actually readable...

plugins/enlink/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+            # add any extra linux libraries here
+            ${PROJECT_LIBS}
+        )
+    endif()
+else()
+    set(PROJECT_LIBS
+        # add any extra windows libraries here
+        ${PROJECT_LIBS}
+    )
+endif()
```
myk002, line 36: this can all be removed

plugins/enlink/building_activation.h
```diff
@@ -0,0 +1,325 @@
+#ifndef UNKNBLCDR_BUILDING_ACTIVATION
```
myk002, line 1: our current headers prefer `#pragma once`

myk002 (commented):
This looks like a useful tool. I think the reason it hasn't gotten any attention is because the PR is just too long and we have to infer the design from the implementation. It might be better to start with a design doc in a bug, something like #1842. Once we discuss the feature set and the design, it will be much easier to digest the PRs (and also then we can do the PRs feature by feature instead of the whole thing at once)

cppcooper (changes requested):
I want to read this, and review it.. but I don't want to build and maintain a mental model of the entire code. I request copious comments.

UnknowableCoder:
My apologies for the long silence in regards to this. Real-life commitments have been keeping me away, and, to be honest, I kept procrastinating (haven't touched DF for months!).
I will work on cleaning up whatever went wrong with the spurious changes (I really, really messed up there) and, in general, improving and commenting the code, so it can better be discussed and/or reviewed.
Again, my apologies for the delay. I will try to get everything done as soon as I can.

myk002:
closing due to lack of docs; please reopen if you have time to finish this project

PR #1993, opened 2021-11-25
Remove getOuterContainerRef from Lua API

Bumber64:
Don't think this works properly without allocating a new specific_ref. More trouble that it's worth.


warmist:
tbh i'm not sure why this does not work.

Bumber64:
> tbh i'm not sure why this does not work.

I meant to push and squash to the existing request, so I closed it immediately, which probably caused the check to abort or something.

PR #1995, opened 2021-12-17
Add dfhack.gui.getCursorPos() to Lua API

Bumber64:
Equivalent to ``xyz2pos(pos2xyz(df.global.cursor))``

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -1076,6 +1076,10 @@ Announcements
 Other
 ~~~~~
 
+* ``dfhack.gui.getCursorPos()``
+
+  Equivalent to ``xyz2pos(pos2xyz(df.global.cursor))``
+
 * ``dfhack.gui.getDepthAt(x, y)``
 
   Returns the distance from the z-level of the tile at map coordinates (x, y) to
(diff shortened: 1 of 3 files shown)


myk002:
How is this different from guidm.getCursorPos()?
https://github.com/DFHack/dfhack/blob/e813301d168c1d3d27e2dad1fbe8edf0c2a381d5/library/lua/gui/dwarfmode.lua#L93

lethosor:
The only difference I see is the behavior when `x == -30000`.

However, I'm inclined to say that the existing function is sufficient for the purposes of most scripts. `copyall(df.global.cursor)` should be sufficient for anything else that needs a Lua table. `copyall()` would convert any DF object to a Lua table, so I'm not really sure that we need a special-purpose function to do the same just for the `cursor` global.

Bumber64:
> How is this different from guidm.getCursorPos()?
> 
> https://github.com/DFHack/dfhack/blob/e813301d168c1d3d27e2dad1fbe8edf0c2a381d5/library/lua/gui/dwarfmode.lua#L93

I assume that requires an import or something, because I can't just use it from DFHack's Lua interpreter.

If ``copyall(df.global.cursor)`` works, maybe the examples in "Lua API.rst" should use it instead? It's certainly less typing than any other option.

myk002:
> 



> I assume that requires an import or something, because I can't just use it from DFHack's Lua interpreter.

Yes, sorry, the full example would be:
```
local guidm = require('gui.dwarfmode')
local cursor = guidm.getCursorPos()
```

PR #1996, opened 2021-12-18
Use copyall instead of nesting pos2xyz and xyz2pos

Bumber64:


lethosor (approved):
Thanks! Looks like it was my mistake originally

For reference: followup to #1995

PR #2005, opened 2022-02-04
avoid treating libexpat build warnings as errors

myk002:
The [-Werror](https://github.com/DFHack/dfhack/commit/8d551c7a7cbf05a65df498c467efcfb4efcbd4fe#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) in our top-level `CMakeLists.txt` is propagated to the dependencies that we include directly as subdirectories. This will cause warnings in dependency code to halt the DFHack build. I've been running into warnings in libexpat that are preventing me from building.

libexpat has a EXPAT_WARNINGS_AS_ERRORS cmake setting, but toggling the setting makes no difference since our top-level `-Werror` is always in effect. This change effectively re-overrides the setting to be treated as always OFF.

Unless we want to be on the hook for solving libexpat build warnings, I believe this way will avoid unnecessary build breakages for users.


lethosor:
For consistency (and personal preference) I would prefer to disable individual warnings as needed. We're doing that for a couple dependencies already:
https://github.com/DFHack/dfhack/blob/57b4a419b3541c888ca0fdf58a2089143fa85edf/depends/CMakeLists.txt#L16-L18
https://github.com/DFHack/dfhack/blob/57b4a419b3541c888ca0fdf58a2089143fa85edf/depends/CMakeLists.txt#L67-L70

Out of curiosity, what compiler are you using? Apparently I haven't built DFHack since upgrading GCC to 11.2, because I just ran into a couple `maybe-uninitialized` warnings in libexpat myself. I guess we should upgrade our max GCC version in CI to catch this.

myk002:
done.

yeah, I started seeing the warnings after upgrading to gcc-11.2 (x86_64-pc-linux-gnu-11.2.0)

Issue #1998, opened 2022-01-10
feature request: quickfort: allow x,y,z placement offsets in meta blueprints


klenze:
For some types of blueprints (pump stacks, spiral train tracks), it would be helpful to apply blueprints in meta mode with a configurable offset. 
z offsets can technically be specified, but only as increments and decrements. 

Assuming I have painstakingly defined eight levels of a 3x3 spiral track, I could them do (using fantasy `offset` syntax)
```
#meta label(spiral32)
/spiral offset(0;0;0)
/spiral offset(0;0;8)
/spiral offset(0;0;16)
/spiral offset(0;0;24)
```

In a similar vein, bedrooms could be defined once as a prototype and then instantiated repeatedly, allowing the user to easily update their design.

I do not think that such a feature would necessarily lead to range loop expressions in meta blueprints, then parametric blueprints and eventually Turing-complete blueprints. (-:

myk002:
This looks doable, and the general syntax looks like it should be forward compatible with other potential features like rotation and repetition (and Turing completeness).

Are z offsets necessary? Meta blueprints already have `#<` and `#>` for going up and down z-levels. For example, the [Dreamfort apartments meta blueprint](https://docs.google.com/spreadsheets/d/1QNHORq6YmYfuVVMP5yGAFCQluary_JbgZ-UXACqKs9g/edit?usp=drivesdk). How about just `offset(X;Y)`?

I think I'm happy with `offset` as the keyword. `translate` might be more technically accurate, but `translate` is longer than `offset` and some people might not be aware of the mathematical meaning of the term `translate`.

To make sure I understand what is being asked for here, could you give me a blueprint for `/spiral` and a reference blueprint for how you expect it to be applied to the map? I can then use those blueprints directly in quickfort's automated regression tests.

klenze:
> Are z offsets necessary? Meta blueprints already have `#<` and `#>` for going up and down z-levels. For example, the [Dreamfort apartments meta blueprint](https://docs.google.com/spreadsheets/d/1QNHORq6YmYfuVVMP5yGAFCQluary_JbgZ-UXACqKs9g/edit?usp=drivesdk). How about just `offset(X;Y)`?
> 
This would make my above meta a bit longer:

```
#meta label(spiral32)
/spiral
#>
#>
#>
#>
#>
#>
#>
#>
/spiral
#>
#>
#>
#>
#>
#>
#>
#>
/spiral
#>
#>
#>
#>
#>
#>
#>
[...]

PR #2007, opened 2022-02-07
[quickfort] add docs for extended syntax for z-level changes in quickfort blueprints

myk002:
#1998

Documentation for the new extended syntax for jumping up or down multiple z-levels in quickfort blueprints in DFHack/scripts#354


myk002:
buildmaster failed due to timeout, but this is a doc-only change, and the docs and build checks were successful, so merging anyway.

PR #2011, opened 2022-02-18
[test] make a (shallow) copy of table params when storing call args

myk002:
this improves test introspection capabilities since the table is snapshotted at the time of the mock call. A deep copy would be ideal, but this should cover most use cases.


myk002:
ha, and the errors this surfaced are entirely legit. the tests are using the incorrect call context. will fix in the scripts repo

Issue #1743, opened 2021-01-03
Update Windows build docs for VS2022


lethosor:
Specifically, our build scripts work with VS2015, but VS2017-19 require different generator names and architecture flags passed to CMake. See https://github.com/DFHack/dfhack/issues/1518#issuecomment-753334092 and https://github.com/DFHack/dfhack/issues/1715#issuecomment-735369867 for more details.

zishiwu123:
**Disclaimer**
This info pertains to VS2015 build tools and not VS2017-19, but I think it can help save time and frustration for new people trying to build `dfhack\build\win64` on Windows. I have marked my suggestions as bullet points.

**VS2015 Build Tool Error**
The [VS2015 build docs](https://docs.dfhack.org/en/stable/docs/Compile.html#microsoft-visual-studio-2015) contains [this link to download VS2015 build tools](https://visualstudio.microsoft.com/vs/older-downloads/#microsoft-build-tools-2015-update-3) without downloading the full Visual Studio IDE. When I tried downloading and installing just the `Microsoft Build Tools 2015 Update 3`, the installation wizard said there was an error due to missing dependencies but did not specify what the dependencies were.

It is possible that the dependencies error could have been prevented by first installing the `Microsoft Visual C++ 2015 Redistributable Update 3` from that link before installing the build tools.
- I have not tried this approach but I can do it if you think it would be helpful.

**My Solution**
Instead, I logged into my Windows account and used [this link](https://my.visualstudio.com/Downloads?q=visual%20studio%202015&wt.mc_id=o~msft~vscom~older-downloads) which allows you to download different version of Visual Studio and its toolchain. I downloaded `Visual C++ Build Tools for Visual Studio 2015 with Update 3` by clicking on the Download button and then the DVD dropdown option.
![image](https://user-images.githubusercontent.com/17225889/158082969-1f539c0f-9a6f-4552-9f9b-6617c90142d0.png)

This put an ISO file in my Downloads folder. I clicked the ISO file, it opened up a folder that contained a `VisualCppBuildTools_Full` installer, a security catalog file, and a `packages` folder which contained all the dependencies. I had no errors installing the VS2015 build tools by clicking the installer and using the default options.
![image](https://user-images.githubusercontent.com/17225889/158083205-494ec2ae-439f-44dd-95dc-7791f6c61828.png)

Here's the contents of the `packages` folder in case anyone is interested about what dependencies MSBuild v14.0 requires. These include:
[...]

Issue #1518, opened 2020-03-10
Does anybody know why I cant compile 047.04.-alpha?


LordBSD:
λ  cmake ..                                                                                    
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.                    
CMake Error at CMakeLists.txt:58 (message):                                                    
  MSVC 2015 or 2017 is required                                                                
                                                                                               
                                                                                               
CMake Warning (dev) at CMakeLists.txt:376 (find_package):                                      
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.                   
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy                  
  command to set the policy and suppress this warning.                                         
                                                                                               
  CMake variable ZLIB_ROOT is set to:                                                          
                                                                                               
    C:/Users/jdion/Downloads/dfhack/build/depends/zlib/                                        
                                                                                               
  For compatibility, CMake is ignoring the variable.                                           
This warning is for project developers.  Use -Wno-dev to suppress it.                          
                                                                                               
-- JsonCpp Version: 1.8.4                                                                      
-- Looking for C++ include clocale                                                             
-- Looking for C++ include clocale - found                                                     
-- Looking for localeconv                                                                      
-- Looking for localeconv - found                                                              
-- Looking for C++ include sys/types.h                                                         
-- Looking for C++ include sys/types.h - found                                                 
-- Looking for C++ include stdint.h                                                            
-- Looking for C++ include stdint.h - found                                                    
-- Looking for C++ include stddef.h                                                            
-- Looking for C++ include stddef.h - found                                                    
-- Check size of lconv                                                                         
-- Check size of lconv - done                                                                  
-- Performing Test HAVE_DECIMAL_POINT                                                          
-- Performing Test HAVE_DECIMAL_POINT - Success                                                
-- Configuring incomplete, errors occurred!                                                    
See also "C:/Users/jdion/Downloads/dfhack/build/CMakeFiles/CMakeOutput.log".                   
See also "C:/Users/jdion/Downloads/dfhack/build/CMakeFiles/CMakeError.log".                    
C:\Users\jdion\Downloads\dfhack\build [develop ≡ +0 ~1 -0 !]                                   
λ             

Why is it saying   MSVC 2015 or 2017 is required if I have visual studio 2019? Thanks I appreciate you guys always helping me as im still learning.

lethosor:
[...]

Issue #1480, opened 2020-01-04
Compile.rst: The Windows handling has changed...


PatrikLundell:
I read the document at the address "https://dfhack.readthedocs.io/en/stable/docs/Compile.html", which I hope contains the current version.

- Chocolatey has changed its layout a little, so you'd get to the installation via an "Install Now" button at the top of the page. This is a minor change.
- MSVC is a different issue, though. Things have changed since I installed it the last time, and apparently XP support has been removed since then. I first tried 2017, and it didn't work, then 2015, and it didn't work. Removed everything and installed 2015 (and SDK too, for good measure) since that's what works on the old computer. The first problem is with <dfhack>\build\win64\generate-MSVC-gui.bat, which claims there is no C or CC compiler. Eventually I tried to see if MSVC could make sense of <dfhack>\build\win64\VC2015\CMakeFiles\3.16.2\VCTargetsPath.vcxproj (because the error messages seemed to be caused by two programs down here not compiling, among red herring messages of missing registry keys that don't exist on the old system either). MSVC then claimed XP support was needed, and offered to install it (which I accepted). I have no idea how you can get the original install to install that part. After that those programs compiled and the ...-gui bat file worked as expected.
However, Microsoft must have made other changes as well, possibly with default switches, because trying to compile DFHack both through MSVC itself (using the generated .sln file), and through build-debug.bat results in a massive warning and error output. As far as I can see, there seems to be two different warnings (237 in total) of the type:

 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\xlocale(341): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc (compiling source file C:\dfhack\depends\protobuf\google\protobuf\descriptor.pb.cc) [C:\dfhack\build\win64\VC2015\depends\protobuf\protobuf.vcxproj]

and

C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\exception(359): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc (compiling source file C:\dfhack\depends\protobuf\google\protobuf\io\gzip_stream.cc) [C:\dfhack\build\win64\VC2015\depends\protobuf\protobuf.vcxproj]

with the error messages (7 of them) of the type:

[...]

PR #2032, opened 2022-03-15
update Windows build docs for Visual C++ 2015

zishiwu123:
Related to #1743 but for Visual C++ 2015 build tools, not Visual Studio IDE 2017-19. Enhancement to Windows build documentation.

**Before PR**
Windows build docs for Visual C++ 2015 previously instructed users who wanted to only download build tools to use [this link](https://visualstudio.microsoft.com/vs/older-downloads/#microsoft-build-tools-2015-update-3). However, downloading Microsoft Build Tools 2015 Update 3 option from that link is not enough. The installer wizard will fail and display and error message about missing dependencies without explicitly telling the user what was missing.

**After PR**
[Old link](https://visualstudio.microsoft.com/vs/older-downloads/#microsoft-build-tools-2015-update-3) has been replaced with a [new link](https://my.visualstudio.com/Downloads?q=visual%20studio%202015&wt.mc_id=o~msft~vscom~older-downloads) to download the Visual C++ Build Tools for Visual Studio 2015 with Update 3. The user only needs to login to their Microsoft account to access the downloads. The [Windows build tools documetation](https://github.com/DFHack/dfhack/blob/develop/docs/Compile.rst#microsoft-visual-studio-2015) has also been updated with instructions for users to download the build tools ISO file from new link because:
- The installer wizard inside the ISO file contains all the dependencies required by MSBuild v14.0, the version required by DFHack. Dependencies include:
  - Microsoft .NET Framework 4.6.1 Developer Pack
  - Microsoft Visual C++ 2015 Redistributable (x64) - 14.0.24210
  - Windows 10 Universal SDK - 10.0.10240
  - Windows 8.1 SDK
- The installer wizard is straightforward as the user only needs to stick to the default options and click next.
- Users who prefer getting the build tools as part of the Visual Studio IDE installations can still benefit from checking the dependencies listed in the build tools ISO file. It informs them what optional tools must be installed with the IDE to ensure that MSBuild v14.0 has all its dependencies met.

**Notes**
- I was unable to build DFHack for win64 until I changed the `build\win64\generate-MSVC-<options>.bat` files to use `v140` instead of `v140_xp` as [mentioned here](https://github.com/DFHack/dfhack/issues/1518#issuecomment-620938952). Then I was able to run `generate-MSVC-minimal.bat` and build DFHack.
- I was able to build DFHack for win32 using `v140_xp`.
- I found it easier to download just the build tools (v140) rather than download the Visual Studio IDE + build tools together. The build tools only was the only setup that worked for me, but [some people](https://github.com/DFHack/dfhack/issues/1518#issuecomment-622132767) have reported better luck with installing Visual Studio IDE + build tools together. Here are the 3 setups listed below:
  - Visual C++ 2015 build tools only
  - VS 2015 IDE + build tools (got ISO file for it from same link as the build tools only)
  - VS 2017 IDE + build tools (got from [here](https://my.visualstudio.com/Downloads?q=visual%20studio%202017&wt.mc_id=o~msft~vscom~older-downloads))
- The VS 2015 IDE installer also came in the form of an ISO file, but it was very slow to install and even slower to remove. It installed random things into my computer for mobile development and SQL Server, which I had to manually remove some folders because the installer wizard did not delete these folders even though I specified everything to be removed. It might be because I clicked the Windows 8.1 SDK option, which came with mobile app development tools. I saw that was a dependencies in the build tools only ISO file; hence why I clicked that option.
- The VS 2017 IDE installer downloaded MSBuild v15.0 which did not allow me to build because DFHack requires MSBuild v14.0. Here is the path to the MSBuild integrated with VS 2017 IDE: `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin`.

**Suggestions**:
- Update the `build\win64\generate-MSVC-<options>.bat` files to use `v140` instead of `v140_xp` as this error has already been reported in #1518. Not sure how many users of DFHack are running Windows XP as their main gaming platform. But I don't think there are enough that we need to have backwards compatibility with Windows XP at the cost of not being able to build for Windows 10. This opinion only applies to win64.
- [Some people](https://github.com/DFHack/dfhack/issues/1518#issuecomment-753334092) and are able to install newer versions of Visual Studio IDE (e.g. 2019) along with older build tools (MSBuild v14.0 which corresponds to the `v140` flag). But if someone does not already have VS2015 installed as [mentioned here](https://github.com/DFHack/dfhack/issues/1743#issuecomment-1068496970), I would advise them to install just the build tools like I did.

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -485,11 +485,54 @@ other versions won't work against Dwarf Fortress due to ABI and STL incompatibil
 
 You can install Visual Studio 2015_ or 2017_ Community edition for free, which
 include all the features needed by DFHack. You can also download just the
-`build tools`_ if you aren't going to use Visual Studio to edit code.
+Visual C++ 2015 `build tools`_ if you aren't going to use Visual Studio to edit code.
+
+Option 1: Build Tools Only
+++++++++++++++++++++++++++
+Click `build tools`_ and you will be prompted to login to your Microsoft account.
+Then you should be redirected to a page with various download options with 2015
+in their name. If this redirect doesn't occur, just copy, paste, and enter the
+download link again and you should see the options. You need to get:
+Visual C++ Build Tools for Visual Studio 2015 with Update 3.
+Click the download button next to it and a dropdown of download formats will appear.
+Select the DVD format to download an ISO file. When the donwload is complete,
+click on the ISO file and a folder will popup with the following contents:
+
+* packages (folder)
+* VCPlusPlusBuildTools2015Update3_x64_Files.cat
+* VisualCppBuildTools_Full.exe
+
+The packages folder contains the dependencies that are required by the build tools.
+These include:
+
+* Microsoft .NET Framework 4.6.1 Developer Pack
+* Microsoft Visual C++ 2015 Redistributable (x64) - 14.0.24210
+* Windows 10 Universal SDK - 10.0.10240
+* Windows 8.1 SDK
+
+Click VisualCppBuildTools_Full.exe and use the default options provided by the installer
+wizard that appears. After the installation is completed, add the path where MSBuild.exe
+was installed to your PATH environment variable. The path should be:
+
+* ``C:\Program Files (x86)\MSBuild\14.0\Bin``
+
+Option 2: IDE + Build Tools
++++++++++++++++++++++++++++
+Click Visual Studio 2015_ or 2017_ to download an installer wizard that will prompt you
+to select the optional tools you want to download alongside the IDE. If you are not sure
+what options to check, it can be helpful to download the `build tools`_ and open the
+ISO file to see the required dependencies in the packages folder. For Visual Studio 2017,
+it is important to select the v140 build tools option. DFHack requires v140 and not
+the v141 version that comes in a bundle when you click the Desktop Development with C++
+option from the installer.
+
+.. warning::
+  Do not run the build tools installer together with the Visual Studio installer as they
+  will have conflicts with each other by trying to install the same dependencies.
 
 .. _2015: https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2015-and-other-products
 .. _2017: https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=15
-.. _build tools: https://visualstudio.microsoft.com/vs/older-downloads/#microsoft-build-tools-2015-update-3
+.. _build tools: https://my.visualstudio.com/Downloads?q=visual%20studio%202015&wt.mc_id=o~msft~vscom~older-downloads
 
 Additional dependencies: installing with the Chocolatey Package Manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -94,6 +94,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``argparse.processArgsGetopt()``: you can now have long form parameters that are not an alias for a short form parameter. For example, you can now have a parameter like ``--longparam`` without needing to have an equivalent one-letter ``-l`` param.
 - ``dwarfmode.enterSidebarMode()``: ``df.ui_sidebar_mode.DesignateMine`` is now a suported target sidebar mode
 
+## Documentation
+- Updated download link and installation instructions for Visual C++ 2015 build tools on Windows
+
 ## Removed
 - `fortplan`: please use `quickfort` instead
 
--- a/docs/Authors.rst
+++ b/docs/Authors.rst
@@ -193,4 +193,5 @@ Will Rogers             wjrogers
 ZechyW                  ZechyW
 Zhentar                 Zhentar
 zilpin                  zilpin
+Zishi Wu                zishiwu123
 ======================= ======================= ===========================


ab9rf:
I'm going to reiterate that we should probably direct people to use VS 2019 or VS 2022, rather than VS 2017.

ab9rf (approved):
while this doesn't cover vs2019 or vs2022, it looks good for what it does cover

zishiwu123:
I spent half my weekend trying to build DFHack on Windows. I'd motivated to improve the Windows build tools documentation so that developers using Windows to build DFHacks will not have to suffer so much.

docs/changelog.txt
```diff
@@ -94,6 +94,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``argparse.processArgsGetopt()``: you can now have long form parameters that are not an alias for a short form parameter. For example, you can now have a parameter like ``--longparam`` without needing to have an equivalent one-letter ``-l`` param.
 - ``dwarfmode.enterSidebarMode()``: ``df.ui_sidebar_mode.DesignateMine`` is now a suported target sidebar mode
 
+## Documentation
```
myk002, line 97: Make sure you add this entry to the "Future" section above. Right now it's in the release notes for the last release (0.47.05-r4)

docs/Compile.rst
```diff
@@ -485,11 +485,54 @@ other versions won't work against Dwarf Fortress due to ABI and STL incompatibil
 
 You can install Visual Studio 2015_ or 2017_ Community edition for free, which
 include all the features needed by DFHack. You can also download just the
-`build tools`_ if you aren't going to use Visual Studio to edit code.
+Visual C++ 2015 `build tools`_ if you aren't going to use Visual Studio to edit code.
+
+Option 1: Build Tools Only
+++++++++++++++++++++++++++
```
myk002, line 491: To fix the doc build warnings, try an underline style that hasn't been used in the file yet, like backticks (```````````)

myk002:
Thank you for this PR! It does make things much clearer!

ab9rf:
We can't use `v140_xp` instead of `v140` because that'll result in a binary that isn't ABI-compatible with Dwarf Fortress. DFHack has to be built with a compiler toolchain that is ABI-compatible with the one Toady uses, and since he has used the `v140_xp` toolchain for all recent builds, we have to as well.

You might be able to get DFHack to compile with `v140` but I'm not sanguine that it'll actually work when built that way.

ab9rf:
An additional note: the `v141_xp` toolchain in VS2017 is reportedly ABI-compatible with the `v140_xp` toolchain of VS2015, and can probably be used to build DFHack successfully. This has not, however, been tested, to the best of my knowledge.

zishiwu123:
> We can't use `v140` instead of `v140_xp` because that'll result in a binary that isn't ABI-compatible with Dwarf Fortress. DFHack has to be built with a compiler toolchain that is ABI-compatible with the one Toady uses to build Dwarf Fortress itself, and since he has used the `v140_xp` toolchain for all recent builds, we have to as well.
> 
> You might be able to get DFHack to compile with `v140` but I'm not sanguine that it'll actually work when built that way. `v140` and `v140_xp` have incompatible linkers.
> 
> Please confirm that the builds you created using this approach actually work with Dwarf Fortress.

Thank you for the explanation @ab9rf. How would you like me to test my build of DFHack with the `v140` toolchain? I am new to the game, a lurking fan who decided to finally try it out, so I am not familiar with all the features. So far I have created a new world and used the embark assistant plugin. I searched for 4x4 tiles with streams+ level rivers, dense level forest, and deep metals. The plugin found and highlighted matching areas.

Are there any other plugins or commands you would like me to test?

ab9rf:
> Are there any other plugins or commands you would like me to test?

It's promising that you got that far without a crash, but I would like to see multiple fort-mode plugins tested, including those that create, modify, and destroy units and items. Also test as many plugins and scripts as possible which involve vmethod overrides, as those are highly likely to crash if there is an ABI incompatibility.

I have spent some time researching what the difference between `v140` and `v140_xp` and have not been able to find a clear statement as to what is different other than the linker. It's entirely possible that the compilers are the same and the only difference is in runtime linkage; if this is true, it should have little or no effect on DFHack itself because the parts of DFHack that have to be ABI compatible are only built to DLL and thus are not linked to a standalone executable.

Perhaps @quietust or one of our other developers knows more about the specifics of ABI compatibility of these different toolsets.

zishiwu123:
The force push was because I accidentally typed the characters **`:** to the end of my commit message and wanted to remove those. I wasn't used to the key bindings on the default editor that comes with git for windows.
- fixed doc errors / moved changelog entry to future section`:

Question: for pull requests that have changes suggested by reviewers, do you prefer that: I place them in separate commits no matter how small the changes are? As opposed to running `git commit --amend` and then `git push --force` to overwrite a pushed commit?

I read the first option is safer than the second. Also I read from [this article](https://cloudfour.com/thinks/squashing-your-pull-requests/) that the commits in a PR can be squashed prior to merging.

myk002:
I personally prefer squash and merge, keeping the related changes from one PR in one commit and setting the commit timestamp to the time of merge. This has the side benefit of not cluttering the main branch with merge commits.

If there are commits that I think should be kept separate, I use rebase and merge.

tl;dr we try to avoid a lot of merge commits in the main branch, but handle the commits in your PR however you like.

myk002 (approved):

myk002:
@ab9rf should anything about v140 vs v140_xp be included in this PR, or is the text as written good for merge?

zishiwu123:
@myk002 My intent in bringing up the `v140` vs `v140_xp` was not to add changes in the PR regarding this. I only wanted to bring this to your attention again since it was mentioned in #1518.

I think this PR should only focus on updating the Windows build documentation, specifically on how to install build tools. I think other matters such `v140` vs `v140_xp` should be opened in a separate issue so we don't clutter the scope of this PR.

myk002:
Looking through the issues list, this PR is also relevant to #1480

ab9rf:
> @ab9rf should anything about v140 vs v140_xp be included in this PR, or is the text as written good for merge?

We need better information on the difference between `v140` and `v140_xp` before we make a recommendation in this area. Until we have that information, I don't want to put anything into the documentation on this issue.

myk002:
> We need better information on the difference between `v140` and `v140_xp` before we make a recommendation in this area. Until we have that information, I don't want to put anything into the documentation on this issue.

Ack. I'll merge this PR, then, and we can update the recommendation on the build chain once you're research is complete.

ab9rf:
I've completed a build using both `v140_xp` and `v140`. While there are differences between the resulting DLLs, as far as I can tell they are immaterial. Some data areas are shuffled, but all the same data is present, and the code generation appears identical. As DFHack (and probably also Dwarf Fortress itself) doesn't run reliably on 64-bit XP even when built with `v140_xp`, due to a bug in the MSFT-provided libraries used with that build chain that MSFT has never fixed, there should be no harm in using `v140` instead of `v140_xp` as the build toolchain, at least on 64-bit platforms. However, we should continue to do release builds using `v140_xp`, especially for the 32-bit version, which still works fine on XP.

myk002:
@ab9rf could you add that into the build docs? Maybe also a comment would be useful in https://github.com/DFHack/dfhack/blob/develop/build/win64/generate-MSVC-all.bat

ab9rf:
> @ab9rf could you add that into the build docs? Maybe also a comment would be useful in https://github.com/DFHack/dfhack/blob/develop/build/win64/generate-MSVC-all.bat

#2034

PR #2033, opened 2022-03-16
autofarm code cleanup

ab9rf:
This is a minor cleanup of autofarm, mostly to eliminate some code smells (like `using std`) and to be more consistent with modern C++ coding standards

### DIFF
--- a/plugins/autofarm.cpp
+++ b/plugins/autofarm.cpp
@@ -26,19 +26,13 @@
 
 #include <queue>
 
-using std::vector;
-using std::string;
-using std::map;
-using std::set;
-using std::queue;
-using std::endl;
 using namespace DFHack;
 using namespace df::enums;
 
 using df::global::world;
 using df::global::ui;
 
-static command_result autofarm(color_ostream &out, vector <string> & parameters);
+static command_result autofarm(color_ostream& out, std::vector<std::string>& parameters);
 
 DFHACK_PLUGIN("autofarm");
 
@@ -55,10 +49,10 @@ const char *usage = (
 
 class AutoFarm {
 private:
-    map<int, int> thresholds;
+    std::map<int, int> thresholds;
     int defaultThreshold = 50;
 
-    map<int, int> lastCounts;
+    std::map<int, int> lastCounts;
 
 public:
     void initialize()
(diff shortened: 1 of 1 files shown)


plugins/autofarm.cpp
```diff
@@ -168,41 +162,37 @@ class AutoFarm {
...
+        const uint32_t bad_flags{
+#define F(x) (df::item_flags::Mask::mask_##x)
+        F(dump) | F(forbid) | F(garbage_collect) |
+        F(hostile) | F(on_fire) | F(rotten) | F(trader) |
+        F(in_building) | F(construction) | F(artifact)
 #undef F
+        };
 
-        for (auto ii : world->items.other[df::items_other_id::SEEDS])
+        for (auto& ii : world->items.other[df::items_other_id::SEEDS])
```
myk002, line 175: Is there a benefit to using a reference to a pointer here?

plugins/autofarm.cpp
```diff
@@ -213,11 +203,11 @@ class AutoFarm {
             farm->plant_id[season] = new_plant_id;
             out << "autofarm: changing farm #" << farm->id <<
                 " from " << get_plant_name(old_plant_id) <<
-                " to " << get_plant_name(new_plant_id) << endl;
+                " to " << get_plant_name(new_plant_id) << '\n';
```
myk002, line 206: I think it's functionally equivalent, but isn't the standard to use `std::endl` in this context?

plugins/autofarm.cpp
```diff
@@ -213,11 +203,11 @@ class AutoFarm {
...
         }
     }
 
-    void set_farms(color_ostream& out, set<int> plants, vector<df::building_farmplotst*> farms)
+    void set_farms(color_ostream& out, std::set<int>& plants, std::vector<df::building_farmplotst*>& farms)
```
myk002, line 210: plants could be made const. I think farms can too. That way the change to passing by reference can guarantee no side effects.

plugins/autofarm.cpp
```diff
@@ -281,44 +268,33 @@ class AutoFarm {
...
-            if (i &&
-                (i->flags.whole & bad_flags.whole) == 0 &&
-                plantable_plants.count(i->mat_index) > 0)
-            {
-                lastCounts[i->mat_index] += i->stack_size;
-            }
-        }
+        };
 
+        for (auto& i : world->items.other[df::items_other_id::PLANT])
```
myk002, line 290: Elegant!

plugins/autofarm.cpp
```diff
@@ -281,44 +268,33 @@ class AutoFarm {
...
+            count(i);
+        for (auto& i : world->items.other[df::items_other_id::PLANT_GROWTH])
```
myk002, line 292: This is a pointer type, right? No need for the &, I think

ab9rf: but also harmless

plugins/autofarm.cpp
ab9rf: `endl` forces an output flush, there's no reason to do that here

plugins/autofarm.cpp
ab9rf: it annoyed me that i had repeated code. i'm a big fan of DRY

plugins/autofarm.cpp
ab9rf: not really, but at the same time there's no real cost to it

myk002: Fair enough

myk002 (approved):

PR #2034, opened 2022-03-16
visual studio 2022/2019 build instructions

ab9rf:
update instructions to recommend using VS 2022 or VS 2019 with downlevel tool support

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -542,6 +542,7 @@ if the 2015 toolchain is installed.
 
 .. _2022: https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&channel=Release&version=VS2022&source=VSLandingPage&cid=2030&passive=false
 .. _2019: https://my.visualstudio.com/Downloads?q=visual%20studio%202019&wt.mc_id=o~msft~vscom~older-downloads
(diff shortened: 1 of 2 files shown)


myk002:
Should we mark this as fixing #1743 (and #1480)?

docs/Compile.rst
```diff
@@ -469,7 +469,8 @@ Dependencies
 ------------
 You will need the following:
 
-* Microsoft Visual C++ 2015 or 2017
+* Microsoft Visual C++ 2022, 2019, 2017, or 2015 (optional)
```
myk002, line 472: Instead of "optional", how about "if editing code"?

myk002 (approved):

myk002:
Needs changelog entry

docs/Compile.rst
ab9rf: it's a perfectly reasonable option to use VS Code to edit code and let VS Code call the build tools. the integration isn't even all that bad.

myk002: Ah, ok. I wasn't aware that VS Code and Visual C++ were different products. The last time I did Windows programming, there was just Visual Studio, and Visual C++ just plugged into that.

myk002:
Does this PR fully cover the "our build scripts work with VS2015, but VS2017-19 require different generator names and architecture flags passed to CMake" concern from #1743?

PR #2037, opened 2022-03-19
 Added (chain) for [CHAIN_METAL_TEXT] armours in library/gui/materials.lua used by the gui/create-item script-- again (oops)

wolfboyft:
Hopefully I'm not messing anything up this time.
Is the changelog entry alright?

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -49,6 +49,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `EventManager`: now each registered handler for an event can have its own frequency instead of all handlers using the lowest frequency of all handlers
 - `stocks`: allow search terms to match the full item label, even when the label is truncated for length
 - `dfhack-examples-guide`: add mugs to ``basic`` manager orders
+- `gui/create-item`: Added "(chain)" annotation text for armours with the [CHAIN_METAL_TEXT] flag set
 
 ## Documentation
 - Add more examples to the plugin skeleton files so they are more informative for a newbie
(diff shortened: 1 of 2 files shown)


docs/changelog.txt
```diff
@@ -1361,6 +1361,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Lua
 - Added a new ``dfhack.console`` API
 - API can now wrap functions with 12 or 13 parameters
+- Added (chain) for [CHAIN_METAL_TEXT] armours ``library/lua/gui/materials.lua`` (affects the ``gui/create-item`` script)
```
myk002, line 1364: A few suggestions for changelog entries:

- You should add changelog entries to the "Future" version at the top of the file. This section was for version 0.44.02-alpha1.
- Lead with the script that will show the change, and put the script name in single backticks, not double, so a hyperlink to the script is generated by the doc parser.
- Follow the script/plugin name with a colon. This is how the doc parser groups related entries.
- Unless the change is completely internal, describe the change in behavior that the user will see. End users are the ones reading this changelog.

Example:
```
- `gui/create-item`: Added "(chain)" annotation text for armours with the [CHAIN_METAL_TEXT] flag set

File: library/lua/gui/materials.lua [hunk 2]
```diff
@@ -314,13 +314,10 @@ function ItemTypeDialog(args)
...
+                local success, chain = pcall(function() return def.props.flags.CHAIN_METAL_TEXT end)
+                local text = success and chain and " (chain) " or " "
                 local success, adjective = pcall(function() return def.adjective end)
-                if success and adjective ~= "" then
-                    text = " " .. adjective .. " " .. def.name
-                else
-                    text = " " .. def.name
-                end
+                text = success and adjective ~= "" and text .. adjective .. " " .. def.name or text .. def.name
                 if not filter or filter(itype,subtype,def) then
```
myk002, line 321: This... might be taking it to far : ) the if ... else format would be clearer here.

myk002 (changes requested):

wolfboyft:
Bottom of the changelog section's the right place I assume (if there is one)
And yeah I was thinking that while writing it. Made it clearer but I suppose it doesn't matter a whole lot.

myk002 (approved):

PR #2038, opened 2022-03-19
Added custom-raw-tokens utility

wolfboyft:
So... documentation, how should I go about doing that? And should I be using mkmodule etc for what is currently just returning a function?


myk002:
> So... documentation, how should I go about doing that? And should I be using mkmodule etc for what is currently just returning a function?

I think docs should go in https://github.com/DFHack/dfhack/blob/develop/docs/Lua%20API.rst#lua-modules

I'm looking forward to reading the docs, actually, since I'm not really sure about who is using this and why. I've never done any raw hacking myself. I'd like to see some example workflows.

Re: module, yes, if it's going in the DFHack Lua library, it should take the same form as the other modules. You probably already know how to do it, but if not, see https://github.com/DFHack/dfhack/blob/develop/library/lua/argparse.lua for a (relatively) short and simple example.

wolfboyft:
Well, the motivating example was one of my mods, and a very clear example of its use is here: https://github.com/wolfboyft/guns/blob/bc1c73af36e034b31de9ca06b73d330e9de27297/projectileManager.lua#L36

I imagine seeing the capacity to use a module like this in the docs would give ideas anyway, that happened a lot for me at least.
A concrete example I was thinking of would be a reaction that [has a chance to] makes the person who made it go insane. I know a lot of mods revolve around weird dwarven science, so...
Or maybe food that, when eaten, decreases stress by 100 (like in real life lol). Though I'm not sure I would have an easy time making such an example since there's no events for that. Eventful's going [to be going] through an overhaul, right?
I'm of course going to put more concrete examples in the docs.

RE caching, I would say we don't know what sort of on-the-cusp-of-too-laggy forts people might want to use this in. A whole huge creature raw with custom raw data at the end being iterated with string finding functions over many creatures would probably benefit from the optimisation. Besides, there's no reason to go back on optimisations if they're clear enough, right? Or is it the fact that different arguments don't return different values that makes it questionable? And huh, manager feels appropriate to me since it has a few extra features than lookup, but I guess so maybe (or I guess so, sure. Do you 100% want me to change it?).

And okay, I'll just have the main lookup function called something like lookUp and do the _ENV business.

RE changelogs, is
```
# Future
## Lua
- Added ``customRawData`` utility to Lua library.
```
alright?

Also, should I maybe use a table for the eventful onUnload function to "bury" it more for elsewhere?

wolfboyft:
...I wonder if I could metatable the returned value to be callable. customRawData(xyz) was really nice IMO.

wolfboyft:
I will make it so that the raw subtype or an instance of that subtype can both be passed to getTag.

myk002:
> ...I wonder if I could metatable the returned value to be callable. customRawData(xyz) was really nice IMO.

You can also alias functions with a local var for the same level of convenience. I do this with the log functions in quickfort, for example. 

https://github.com/DFHack/scripts/blame/5472fff868219e504d047d2401db34def3d241cf/internal/quickfort/dig.lua#L20

myk002:
Make sure you merge from upstream/develop, not your local branches. That will keep the changes from your other PR from appearing in this PR

myk002:
Regarding caching, what is the performance increase over, say, 10 seconds? How many lookups can you make in 10 seconds? Then remove the cache and try that many lookups again. If the difference is less than, say, 5%, then the cache probably isn't worth the code complexity or memory storage.

The cache could very well be worth it. It's just good practice to test it first and not assume.

wolfboyft:
6131073 with caching
34890 without caching
using:
local customRawData = require("customRawData")

```
local time = os.time()
local count = 0
while os.time()-time < 30 do
	customRawData.getCreatureTag(dfhack.gui.getSelectedUnit(), "ABSENT_TAG")
	count = count + 1
end
print(count)
```
on a single human in arena mode

Much more than 5%! So I guess that means caching should stay

wolfboyft:
I'm thinking this is ready.

myk002:
Be sure to
```
git fetch upstream
git merge upstream/develop
git push
```
on this branch -- I still see some extraneous commits

docs/Lua API.rst
```diff
@@ -2984,6 +2984,66 @@ Predefined instance methods:
...
+
+* ``customRawData.getTag(typeDefinition, tag[, ...])``
+
+  Where ``typeDefinition`` is a raw struct as seen in ``df.global.world.raws``, e.g:
+  ``dfhack.gui.getSelectedItem().subtype``, ``tag`` is the name of the custom tag you
+  want to get, and the remaining arguments correspond to the arguments in the tag as
+  booleans denoting whether the argument in the tag is a string or a number. ``true``
+  for string, false (or absent) for number. It will then be returned using single or
+  multiple return values. If the tag is not present, the result is false, if it is
+  present but has no arguments, the result is true.
```
myk002, line 3000: will the caller always know whether to expect strings or numbers for a particular tag? can we always return numbers if they are numbers and let the caller convert them to strings if desired? alternately, can we just always return strings and let the caller convert the numbers if they know to expect numbers? either way it would solve the issue of the cache returning unexpected types.

docs/Lua API.rst
```diff
@@ -2984,6 +2984,66 @@ Predefined instance methods:
...
+
+* ``customRawData.getRaceCasteTag(raceDefinition, casteNumber, tag[, ...])``
+
+  Gets tag from a race or caste instead if appplicable. Everything with getTag applies to this.
```
myk002, line 3004: what is the type of `raceDefinition`? why do we need a separate method for race caste tags? Can this be handled by `getTag()`?

docs/changelog.txt
```diff
@@ -62,6 +62,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - Lua wrappers for functions reverse-engineered from ambushing unit code: ``isHidden(unit)``, ``isFortControlled(unit)``, ``getOuterContainerRef(unit)``, ``getOuterContainerRef(item)``
+- Added ``customRawData`` utility to Lua library.
```
myk002, line 65: `customRawData` should be surrounded in single quotes, not double, to link to the documentation. also maybe add a note about what the module does.

myk002 (changes requested):

docs/Lua API.rst
```diff
@@ -2984,6 +2984,66 @@ Predefined instance methods:
...
+    -- (per unit every tick)
+    local healAmount = 0
+    for _, entry in ipairs(unit.inventory) do
+        if entry.mode == 2 then -- Worn
+            healAmount = healAmount + (customRawData.getTag(entry.item.subtype, "HEAL_AMOUNT") or 0)
+        end
+    end
+    unit.body.blood_count = math.min(unit.body.blood_max, unit.body.blood_count + healAmount)
+
+This also defines a function it uses at ``eventful.onUnload.clearExtractedCustomRawData``
```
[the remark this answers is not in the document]
wolfboyft: The rationale behind it was to warn users of an exposed function but I will replace it with your text and use a table instead with a comment explaining why.

docs/Lua API.rst
```diff
@@ -2984,6 +2984,66 @@ Predefined instance methods:
 
 To avoid confusion, these methods cannot be redefined.
 
+customRawData
```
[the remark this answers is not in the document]
wolfboyft: Oh, sorry. And how about `custom-raw-tags`? That's hyphenated, and explains that it's for custom and not vanilla tags. Though, data... nah, tags is fine. Advanced raw constructs (if there are any left) are still tags.

docs/Lua API.rst
```diff
@@ -2984,6 +2984,66 @@ Predefined instance methods:
...
+=============
+
+A utility allowing for custom tags to be added to the raw definitions for modding purposes:
+
+* ``customRawData.getTag(typeDefinition, tag[, ...])``
+
+  Where ``typeDefinition`` is a raw struct as seen in ``df.global.world.raws``, e.g:
+  ``dfhack.gui.getSelectedItem().subtype``, ``tag`` is the name of the custom tag you
```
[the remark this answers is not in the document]
wolfboyft: For most it's a pointer, not an int. E.g. https://github.com/DFHack/df-structures/blob/master/df.items.xml#L1224. The way it's set up currently is that subtypes are acquired manually for what you want to specifically read from. A raw struct is a type definition with raws and all that, I guess. Maybe just calling them type definitions

docs/Lua API.rst
wolfboyft: Hmm... sure, okay. I'll remove the bool business.

docs/Lua API.rst
wolfboyft: No, it's separate behaviour. Or would you rather the caste business be merged in with if statements?

docs/Lua API.rst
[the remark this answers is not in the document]
wolfboyft: Well, I can't use `custom-raws-tags` in the examples, though, since syntax.

docs/Lua API.rst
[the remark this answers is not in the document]
wolfboyft: They're called raw structures once here https://github.com/DFHack/dfhack/blob/develop/docs/changelog.txt#L1162, though

wolfboyft:
Wait, let me check if duplicate raw tags only read the last. I can easily just hold an args variable until the end of the raw strings or if another instance is found if so.

docs/Lua API.rst
```diff
@@ -2984,6 +2984,63 @@ Predefined instance methods:
...
+    -- (per unit every tick)
+    local healAmount = 0
+    for _, entry in ipairs(unit.inventory) do
+        if entry.mode == 2 then -- Worn
+            healAmount = healAmount + tonumber((customRawTags.getTag(entry.item.subtype, "HEAL_AMOUNT")) or 0)
+        end
+    end
+    unit.body.blood_count = math.min(unit.body.blood_max, unit.body.blood_count + healAmount)
+
+The cache is cleared when a map is unloaded.
```
myk002, line 3042: since we removed references to the cache above, and since the cache no longer has potentially unexpected side effects, maybe remove this note too.

docs/Lua API.rst
```diff
@@ -2984,6 +2984,63 @@ Predefined instance methods:
 
 To avoid confusion, these methods cannot be redefined.
 
+.. _custom-raw-tags:
+
+custom-raw-tags
+===============
+
+A module for reading custom tags added to the raws by mods.
```
myk002, line 2992: Sorry for my ignorance on this, but can this module be used to read vanilla (non-mod-added) tags in the raws? Is it useful to be able to read vanilla tags? Do vanilla tags even exist?

myk002:
> > what is the type of raceDefinition? why do we need a separate method for race caste tags? Can this be handled by getTag()?

> No, it's separate behaviour. Or would you rather the caste business be merged in with if statements?

They both appear to look up tags. What would a user want? Would a user want to call a different function for a different source type? Is it useful to differentiate?

myk002:
> > Isn't the return value from Gui::getSelectedItem() an item pointer, and subtype is an int16?

> For most it's a pointer, not an int. E.g. https://github.com/DFHack/df-structures/blob/master/df.items.xml#L1224

What happens if the particular item that the caller is processing returns an int instead of pointer?

wolfboyft:
> > > what is the type of raceDefinition? why do we need a separate method for race caste tags? Can this be handled by getTag()?
> 
> > No, it's separate behaviour. Or would you rather the caste business be merged in with if statements?
> 
> They both appear to look up tags. What would a user want? Would a user want to call a different function for a different source type? Is it useful to differentiate?

Well, one takes an argument for a caste, and the other uses no caste. The function that takes no caste is the one that properly handles the behaviour around caste tags. I mean, it really seems fine to me, I have to say...

docs/Lua API.rst
wolfboyft: It can actually read vanilla tags, I've been using CASTE_NAME for a lot of my testing. Is it useful (otherwise) to read vanilla tags? No, I highly doubt it. The "raw struct" should have all the information already outside of its raw strings. And I don't really understand the question "do vanilla tags even exist?" lol, vanilla tags are ones that are used by the game and don't throw errors.

wolfboyft: I'm not really sure if there was a change you wanted me to make, but, `A module for reading custom tags added to the raws by mods. Can also read vanilla tags if requested.` is this more on-point?

myk002:
> > > > what is the type of raceDefinition? why do we need a separate method for race caste tags? Can this be handled by getTag()?
> > 
> > 
> > > No, it's separate behaviour. Or would you rather the caste business be merged in with if statements?
> > 
> > 
> > They both appear to look up tags. What would a user want? Would a user want to call a different function for a different source type? Is it useful to differentiate?
> 
> Well, one takes an argument for a caste, and the other uses no caste. The function that takes no caste is the one that properly handles the behaviour around caste tags. I mean, it really seems fine to me, I have to say...

I'm not making veiled suggestions - I'm asking these questions because I really don't know anything about raws or modding. I'm just trying to make sure the API is user-focused and that we're not missing anything we might regret not catching later.

wolfboyft:
ok, ok. So, basically, when, in a creature definition, a caste is selected, its tags take priority over tags defined when no caste is selected. (actually i will also need to test if SELECT_CASTE:ALL is different to having put no caste tags. it's probably fine but ust to be sure i'm imitating vanilla behaviour...). so if somebody wanted to get a custom tag to do with a unit, they would pretty much always want to have it respect caste tag priority. so they'd likely use getUnitTag, but getRaceCasteTag exists in case they have those data but not the unit, maybe during extreme optimisation or a virtual "what if" environment where the unit does not exist (perhaps before creating one). getRaceCasteTag takes the same sort of arguments as getTag which is right to be exposed if getTag is exposed. Because the architecture of the standard getTag is so different, they are separate functions. So it's user-focused and consistent there. But I am concerned with the whole users-having-to-get-raw-structs-manually-from-numbers thing, I think passing an item or an entity to getTag probably should be made to work automatically.
I hope I am not missing your points here.

Okay, let's consider use cases:
1 Want to get tag from a simple type definition like an entity type definition (needs struct)
2 Want to get tag from something simple like an entity (needs instance of struct)
3 Want to get tag from race-caste definition and number (needs struct)
4 Want to get tag from unit's race and caste (needs instance of struct)
5 Want to get tag from an item's subtype (needs instance of struct)
6 Want to get tag from an item's material (needs instance of struct)
Currently their solutions are:
1 getTag
2 getTag with manual indexing from df.global.world.raws or just indexing the object if it's something like item.subtype
3 getRaceCasteTag
4 getUnitTag
5 getTag with manual indexing of item
6 getTag with a different manual indexing of item
(reminds me, i think there is a material equivalent to creature variations which might not be already expanded in the raw strings for a raw struct, so im going to check. i will check all the xyz token wiki pages)

and that's about it for differentiable use cases that I am aware of right now. I'm going to do further research on the wiki.

Since we're now considering different function signatures on the same function name maybee we should switch to C++ (though it is possible in lua with hackery). I reckon I can  do that, though in a project-specific way i dont know.

Actually, coming to think of it, I could just know the maximum number of arguments we are going to use, then expose one getTag function with args a b c d e f g etc and then do type checks etc on that and then call internal functions like getTagSimple etc. For user-focusedness I suppose passing an item could just automatically do its subtype (so long as we say what all the automatic indexing is), after all, passing a unit wouldn't deal with its inventory items or its civ. In the docs the function could be written as getTag(argument, configuration, number, one) blah blah getTag(argumentConfiguration, two) blah blah

OK that's a lot of text, sorry.
TL;DR: 1 automatic-arguments function that you can pass raw structs or instances of them in the world into that then calls other internal functions for the behaviour, and reading wiki raws to check for missed concepts.

wolfboyft:
I've come up with the following function signatures and will begin writing after checking every single raw tag on the wiki for things we need to be aware of.
getTag(unit, tag): wraps getTag(raceDefinition, casteNumber, tag)
getTag(raceDefinition, casteNameString, tag): wraps getTag(raceDefinition, casteNumber, casteString, tag)
getTag(raceDefinition, casteNumber, tag): wraps getTag(raceDefinition, casteNumber, casteString, tag)
local getTag(raceDefinition, casteNumber, casteString, tag): core
getTag(simpleInstance, tag): wraps getTag(simpleTypeDefinition, tag)
getTag(simpleTypeDefinition, tag): core

docs/Lua API.rst
myk002: I meant, are tags just a mod thing or are they used by vanilla as well, but I understand from your response that tags are common in the raws, it's just that the information from those tags are already accessible in c++ and lua via the generated struct definitions, so it's not likely that anyone would need to use this module to read vanilla tags.  No need to add anything about vanilla tags to the docs, then.

myk002:
I think that API is much clearer (and usable) than what we started out with. Thank you for putting up with my prodding!

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,243 @@
+--[[
+custom-raw-s
```
myk002, line 2: typo?

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,243 @@
...
+Allows for reading custom tokens added to raws by mods
+by Tachytaenius (wolfboyft)
+
+Yes, custom raw tokens do quietly print errors into the error log but the error log gets filled with garbage anyway
+]]
+
+local _ENV = mkmodule("custom-raw-tokens")
+
+local eventful = require("plugins.eventful")
```
myk002, line 11: @lethosor what do you think about this? is it acceptable to depend on a plugin in a core library? should `eventful` (and other "core-like" plugins like `debug`) be moved into the core lib to avoid the architecture inversion?

alternately, we could use `sc-script` to call a script (or this module directly?) with specific parameters (e.g. `--reset`) on UNLOAD events. that just uses core functionality instead of `eventful`, but is a little convoluted for this simple use case.

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,243 @@
...
+local utils = require("utils")
+
+local customRawTokensCache = {}
+eventful.onUnload["custom-raw-tokens"] = function()
+    customRawTokensCache = {}
+end
+
+-- Not state, defined later:
+local ensureTable, doTable, getSubtype, rawStringsFieldNames
+local getTokenCore, getRaceCasteTokenCore
```
myk002, line 21: this is not a common approach, and I don't think I can recommend it. It is clearer (and more common) to define your locals (vars and functions) above the code that uses them.

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,243 @@
...
+Function signatures:
+getToken(rawStruct, token)
+getToken(rawStructInstance, token)
+getToken(raceDefinition, casteNumber, token)
+getToken(raceDefinition, casteString, token)
+]]
+function getToken(a, b, c)
+    -- Argument processing
+    assert(type(a) == "userdata", "Expected userdata for argument 1 to getToken")
+    if df.is_instance(df.creature_raw, a) then
```
myk002, line 33: factoring the bodies of the if clauses out into separate local functions would make this function much easier to read

myk002 (changes requested):

wolfboyft:
How about something to clean up the error log? Maybe a registry of valid tokens for which the errors are removed? Do we have access to errorlog.txt?

myk002:
> How about something to clean up the error log? Maybe a registry of valid tokens for which the errors are removed? Do we have access to errorlog.txt?

isn't that output by the DF code itself? I'm not aware of any hooks we have into how that file is written.

myk002 (commented):
overall, looks good. I'd just wait for @lethosor's opinion on whether eventful is a cromulent dependency

myk002 (approved):

myk002 (approved):
It's this PR still WIP? If not, the title should be updated

docs/Lua API.rst
```diff
@@ -2984,6 +2984,80 @@ Predefined instance methods:
 
 To avoid confusion, these methods cannot be redefined.
 
+.. _custom-raw-tokens:
+
+custom-raw-tokens
+===============
```
myk002, line 2990: This is causing the build error due docs: home/runner/work/dfhack/dfhack/docs/Lua API.rst:2990: WARNING: Title underline too short.

wolfboyft:
> It's this PR still WIP? If not, the title should be updated

Well, it may be for some really really specific cases a little bit "if someone asks". But that's probably not enough for the core library, is it, so I'm keeping it WIP while I take a break and study the raws on the wiki at my own pace.

wolfboyft:
Any more features can be added if asked. Point to me as the maintainer and I'll see what I can do.

wolfboyft:
Submitting for early code review. I am going to replace the on map load with something that will go on the next frame after on world load (on world load being when valid tokens are registered). Also, "*** Error(s) found in the file" lines will still exist even if there's nothing after them, should I deal with that?

wolfboyft:
I think that's everything I had planned besides the errors found in idea

wolfboyft:
As discussed on discord, let's roll it back then. How should that be done? Should I undo and force push?

myk002:
> As discussed on discord, let's roll it back then. How should that be done? Should I undo and force push?

No need to force push. The PR will get squashed into a single, clean commit when it is merged. Commit as much as you like!

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,375 @@
...
+            end
+        elseif thisCasteActive and token == rawStringToken then
+            currentTokenIterator = iter
+        end
+    end
+    if currentTokenIterator then
+        return doToken(thisRaceDefCache, token, currentTokenIterator)
+    end
+    thisRaceDefCacheCaste[token] = false
+    if casteNumber ~= -1 then
```
myk002, line 155: Can simplify to:
```
if casteNumber == -1 then return false end
return getRaceCasteTokenCore(raceDefinition, -1, token)
```
(Plus existing comments)

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,375 @@
...
+        if type(tokenData) == "table" then
+            return table.unpack(tokenData)
+        else
+            return tokenData
+        end
+    end
+    
+    -- Get data anew
+    local success, type = pcall(function() return typeDefinition._type end)
+    local rawStrings = typeDefinition[rawStringsFieldNames[type]]
```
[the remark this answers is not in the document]
wolfboyft: What should I say instead?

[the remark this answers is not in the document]
myk002: have you tested what happens when `pcall` returns false? if it works, then this is fine, but if you get stacktrace, then you probably need to test the value of `success` and throw a more descriptive error message before you use `type` to (indirectly) index `typeDefinition`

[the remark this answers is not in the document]
wolfboyft: Oh, yeah, it's absolutely fine, I made it to deal with putting in items like blocks.

wolfboyft:
Should/could this be moved to modtools? Also we really need to find another reviewer...

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,372 @@
...
+            end
+        end
+        assert(growthNumber, "Invalid argument 2 to getToken, growth name \"" .. growthString .. "\" not found")
+        assert(type(c) == "string", "Invalid argument 3 to getToken, must be a string")
+        token = c
+    end
+    return getPlantGrowthTokenCore(plantDefinition, growthNumber, token)
+end
+
+local function getTokenArg1Else(userdata, b, c)
```
lethosor, line 281: Is `c` used? (a more descriptive name would make it easier to tell)

lethosor (commented):
I think this looks good too. I'm not much of a modder, but I have no issues with the API as it currently stands.

lethosor:
pre-commit.ci autofix

docs/Lua API.rst
```diff
@@ -3192,6 +3192,80 @@ Predefined instance methods:
...
+
+custom-raw-tokens
+=================
+
+A module for reading custom tokens added to the raws by mods.
+
+* ``customRawTokens.getToken(typeDefinition, token)``
+
+  Where ``typeDefinition`` is a type definition struct as seen in ``df.global.world.raws``
+  (e.g.: ``dfhack.gui.getSelectedItem().subtype``) and ``token`` is the name of the custom token
```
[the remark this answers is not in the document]
wolfboyft: "I can't find any item type of the 10 I checked with a subtype field" I don't understand this clause specifically, could you rephrase?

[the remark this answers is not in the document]
lethosor: Sure. I selected ~10 different items in-game, and for all of them, `dfhack.gui.getSelectedItem().subtype` gave me the following error (or similar):

```
(interactive):1: Cannot read field item_anvilst.subtype: not found.
stack traceback:
	[C]: in metamethod '__index'
	(interactive):1: in main chunk
	[C]: in function 'safecall'
	./hack/lua/dfhack.lua:612: in function 'dfhack.interpreter'
	../../code/dfhack/scripts/lua.lua:76: in local 'script_code'
	./hack/lua/dfhack.lua:775: in function 'dfhack.run_script_with_env'
	(...tail calls...)
```

so I am not sure what the example is intended to mean.

I think perhaps a better example could be
```suggestion
  Where ``typeDefinition`` is a type definition struct as seen in ``df.global.world.raws``
  (e.g.: an instance of ``df.creature_raw``) and ``token`` is the name of the custom token
```
if this is accurate.

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,372 @@
...
+        -- Assume raw struct *is* argument 1
+        rawStruct = userdata
+    end
+    if not rawStruct then return false end
+    assert(type(b) == "string", "Invalid argument 2 to getToken, must be a string")
+    token = b
+    return getTokenCore(rawStruct, token)
+end
+
+function getToken(a, b, c)
```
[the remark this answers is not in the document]
wolfboyft: Would it be OK to reject this one? Making the changes would be very error-prone, at least for me

docs/Lua API.rst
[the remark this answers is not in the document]
wolfboyft: Oh, it's for stuff like weapons and armour and the like. Things with raw definitions. Barrels, anvils, blocks etc don't have any such thing.

library/lua/custom-raw-tokens.lua
```diff
@@ -0,0 +1,372 @@
...
+local _ENV = mkmodule("custom-raw-tokens")
+
+local customRawTokensCache = {}
+dfhack.onStateChange.customRawTokens = function(code)
+    if code == SC_WORLD_UNLOADED then
+        customRawTokensCache = {}
+    end
+end
+
+local function ensureTable(tableToHoldIn, key)
```
myk002, line 20: I had similar code in quickfort. See #2118

library/lua/custom-raw-tokens.lua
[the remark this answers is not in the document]
myk002: If you ever find yourself saying something like this, consider adding unit tests. That way you can be sure the code still works after changes like this

library/lua/custom-raw-tokens.lua
myk002: #2118 is now merged -- you can use the new global `ensure_key(tableToHoldIn, key)` here

myk002 (approved):

myk002:
Looks good! Give me the ok to merge when you're comfortable with it.

wolfboyft:
All clear on my end. And, I tested it, looks fine with the args changes.

PR #2044, opened 2022-03-23
EventManager refactoring

### LINKED ISSUE #2031: EventManager can miss sending events to handlers
If a handler is registered to an event
And the requested frequency is relatively prime to an existing handler on that event
And there is state kept between invocations of the event triggering code
Then a handler can miss events that it should receive

An example is `manageNewUnitActiveEvent` at https://github.com/DFHack/dfhack/blob/07e96a6a1ca26205585f149b4eb61dada16aa3a5/library/modules/EventManager.cpp#L608

It will check for new units and send events to handlers whose "cool down" timer has expired. Handlers that have had an event fire too recently (according to their freq value) will not get an event fired. This is working as intended.

However, when `manageNewUnitActiveEvent` finishes looping through the handlers, it marks all currently active units as processed. Therefore, the handlers that were not sent an event for the detected new units will never get an event for those units.

In order to continue to support different freq values for different handlers, this issue will need to be solved. Otherwise, we should revert back to the previous behavior of all handlers effectively having the lowest requested freq value.

cppcooper:
~~As in the title, this fixes #2031 and also as (currently) mentioned, this is a work in progress. I wouldn't have opened the PR if not for the commits that I also wouldn't have made if not for needing to checkout master and build/install.~~

~~One major change so far is that the `BUILDING` event now sends destroyed buildings as negative id values, meaning they need to be handled appropriately by any users (of which I only found one `outside-only.lua` via eventful; it doesn't even use the id values). now comes with component events (created/destroyed) that can be used instead of it.~~

edit: so this PR did fix #2031 orginally, and still does.. but #2031 has "officially" been fixed by reverting #1876 [for the most part]. So this [effectively] re-implements #1876, but with fixes for #2031. As of writing this edit, there are some other nice changes such as to the function lists which needed manual synchronization with the enum event order. Now the order is automatically generated for these function arrays and matches that of the EventType enums. A positive side of that particular change is that the compiler errors when new events are added but not integrated into switches used to generate these arrays (there are 2, one in EventManager.cpp and one in eventful.cpp)


library/modules/EventManager.cpp
```diff
@@ -183,26 +188,93 @@ static const eventManager_t eventManager[] = {
     manageInteractionEvent,
 };
 
+template<typename T>
+class event_tracker { //todo: use inheritance? stl seems to use variadics, so it's unclear how well that would actually work
+private:
+    std::unordered_map<T, int32_t> seen;
```
cppcooper, line 194: A variant with a multimap `seen` might be good (thinking repeat jobs).

library/modules/EventManager.cpp
```diff
@@ -479,17 +596,82 @@ TODO: consider checking item creation / experience gain just in case
...
+    //if it happened within a tick, must have been cancelled by the user or a plugin: not completed
+    if(tick > fn_last_tick) {
+        // we're going to need to check if started jobs are still on this list, so we copy the IDs to avoid a linear search into this list for each started_job
+        std::unordered_map<int32_t, df::job*> current_jobs;
+        for (df::job_list_link* link = df::global::world->jobs.list.next; link != NULL; link = link->next) {
+            df::job* job = link->item;
+            current_jobs.emplace(job->id, job);
+        }
+        // iterate the started jobs list
+        for (auto &iter: startedJobs) {
```
cppcooper, line 614: Seems as though `startedJobs` is insufficient for this task. Testing shows this manager doesn't work correctly and this is the only thing that makes a lot of sense.

library/modules/EventManager.cpp
```diff
@@ -434,38 +539,50 @@ static void manageJobInitiatedEvent(color_ostream& out) {
...
+        oldest_last_tick = oldest_last_tick < last_tick ? oldest_last_tick : last_tick;
         // make sure the frequency of this handler is obeyed
-        if (tick - eventLastTick[handler.eventHandler] >= handler.freq) {
+        if (tick - last_tick >= handler.freq) {
             eventLastTick[handler.eventHandler] = tick;
-            // send the handler the new jobs
-            for(auto job : newly_started_jobs){
-                handler.eventHandler(out, (void*)job);
+            // send the handler all the new jobs since it last fired
+            auto jter = startedJobs.upper_bound(last_tick);
```
cppcooper, line 570: Late listener registrations can result in a flood of old events from before the listener was registered. I'd comment in the register function, but it isn't in the diff.

library/modules/EventManager.cpp
```diff
@@ -479,17 +596,82 @@ TODO: consider checking item creation / experience gain just in case
...
+                if (started_job->completion_timer != 0)
+                    continue;
+                if (valid_df_job->completion_timer != -1)
+                    continue;
+
+                //still false positive if cancelled at EXACTLY the right time, but experiments show this doesn't happen
+
+                // the job won't be added if it already exists - NOTE: this means a repeat job can't be seen as completed again until it gets cleaned from the list
+                // This is probably going to be a huge problem as it means there is room for an interaction issue between plugins/scripts
+                if (completedJobIDs.emplace(id).second) {
```
cppcooper, line 634: As the comments suggest, this probably should be removed.

library/modules/EventManager.cpp
cppcooper: I haven't specifically tried to test repeat jobs, but I don't think any changes are necessary at this point having removed the id from checks. As is completed jobs only have one tick (the tick they can be detected completed on) to be recognized. Since the method of doing this is the same as the old way, it should work if it did work. Maybe this could be useful somewhere else though..?

library/modules/EventManager.cpp
```diff
@@ -703,63 +898,160 @@ static void manageBuildingEvent(color_ostream& out) {
...
-                df::construction &construction = (*constru_iter).second;
-                if (df::construction::find(construction.pos) != NULL) {
-                    ++constru_iter;
-                    continue;
-                }
-                //construction removed
-                //out.print("Removed construction (%d,%d,%d)\n", construction.pos.x,construction.pos.y,construction.pos.z);
-                handler.eventHandler(out, (void*) &construction);
-                constru_iter = constructions.erase(constru_iter);
-            }
```
cppcooper, line 762: @expwnent Hoping you can tell me how this is able to work. If the construction object isn't available in DF anymore and we haven't cloned it like we do with `df::job*` in the job events.. then how is it still valid? Can it be expected to stay valid long term?

myk002: The construction being sent to the handler is the copy we have in our local map (`constructions`)

myk002: It might be worth renaming our local caches to make this more clear. E.g. rename `constructions` to `constructions_cache` or something

library/modules/EventManager.cpp
myk002: This could be desired behavior. Maybe we should have a flag on the register function to select whether the backlog is flushed or not

cppcooper: sure, but I could add another param with a default value of off so nothing breaks.

cppcooper:
So according to the tests I've run, everything I've touched is working as we would hope it to. I haven't performed extensive testing but all the events appear to be firing as they had before from what I can tell, and the backlogs are sent to delayed listeners.

What's left, appear to be spaghetti events and I dunno how long they'll take me just to figure out. So far the toughest has taken maybe a couple hours total.

Here's a list of the remaining

1. `updateReportToRelevantUnits`
2. `manageReportEvent`
3. `getWound`
4. `manageUnitAttackEvent`
5. `getVerb`
6. `getAttacker`
7. `gatherRelevantUnits`
8. `manageInteractionEvent`

library/modules/EventManager.cpp
```diff
@@ -703,293 +913,426 @@ static void manageBuildingEvent(color_ostream& out) {
...
+            eventLastTick[handler] = tick;
+            // send all new reports since it last fired
+            auto iter = newReports.upper_bound(last_tick);
+            for(;iter != newReports.end();) {
+                if (valid_reports.count(iter->second)) {
+                    handler.eventHandler(out, (void*) intptr_t(iter->second));
+                    ++iter;
+                } else {
+                    iter = newReports.erase(iter->second);
+                }
```
cppcooper, line 1293: It's a bit unclear how useful this actually is. Users would still need to find the report from the id. So I think this is a bit redundant. Worst case scenario, we've taken away useful information. Best case we've prevented a user from erroneously assuming the id exists

myk002 (commented):
I'm glad this is getting better organization, testing, and data structures, but it's a lot of code to review at once. Is there anything in particular that you'd like feedback on?

one thing I can suggest from personal experience: could you guarantee that the *destroyed events fire before the associated *created events? sometimes the same IDs are reused, and it's better to give callers the opportunity to clear out old data before new data is added. For example, our current building cache in the `Buildings` module has a bug that is triggered by the current (checked-in) behavior of sending created events before destroyed events.

cppcooper:
> it's a lot of code to review at once.

Indeed it is, sorry. It'd probably be best to checkout the branch and read it in your IDE than suffer through the massive diffs. I did comment most stuff though, so there's that. The last few managers were too complex for me to take a crack at rewriting their update code for optimization.. so hopefully they are already optimized ([X] to doubt).

> Is there anything in particular that you'd like feedback on?

Not really no. Just the regular review process. Requested it simply to notify you that this is done.

> could you guarantee that the *destroyed events fire before the associated *created events?

What exactly do you mean? Afaik/iirc, `manageEvents` iterates the events according to the `EventType` enum order, and I placed the events in created/destroyed order. Though I separated the created/destroyed into new events I retained the old one.. so until they are fully removed it can't be guaranteed for the separated versions.

cppcooper:
> #2056 has been merged. Could you merge develop into this branch and overwrite the changes in EventManager.h and EventManager.cpp?

As soon as I get the chance

library/modules/EventManager.cpp
cppcooper: `bool backlog = false` was added to the signature in the header

cppcooper:
As far as I am aware I've done everything necessary to fix #2031 in this PR.

Quite a bit of refactoring, so lots of code to review. If there are comments needed somewhere in the diffs lmk, I have an awful memory so it'll help me too when/if other things are reviewed and I've forgotten everything about the PR.

library/include/modules/EventManager.h
```diff
@@ -28,8 +29,12 @@ namespace DFHack {
                 UNIT_NEW_ACTIVE,
                 UNIT_DEATH,
                 ITEM_CREATED,
-                BUILDING,
-                CONSTRUCTION,
+                BUILDING, // todo: deprecate this event
+                BUILDING_CREATED,
+                BUILDING_DESTROYED,
```
myk002, line 34: I suggest putting the destroyed/removed events above the related created/added events. From experience it's important to know what's gone before learning what's new.

myk002 (commented):
I think I'll have to review the rest on an actual computer. Phone is not adequate.

library/include/modules/EventManager.h
```diff
@@ -28,8 +29,12 @@ namespace DFHack {
                 UNIT_NEW_ACTIVE,
                 UNIT_DEATH,
                 ITEM_CREATED,
-                BUILDING,
-                CONSTRUCTION,
+                BUILDING, // todo: deprecate this event
```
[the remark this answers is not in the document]
cppcooper: Eventful still relies on the `BUILDING` event
```cpp
static void ev_mng_building(color_ostream& out, void* ptr)
{
    int32_t id=(int32_t)(intptr_t)ptr;
    onBuildingCreatedDestroyed(out, id);
}
```
The same is true for `CONSTRUCTION`

Ideally we'll add the new events to eventful, update scripts to use the new split events, and remove the originals.

library/include/modules/EventManager.h
cppcooper: I just retained the old behaviour of finding new before then finding missing

cppcooper: Ok so I just re-ordered them. I didn't remove the deprecation todo's though, because the created/destroyed lists need to be processed before the events are sent; it was done this way to avoid looping the handler's twice and having one where the last tick is updated and the other not...  I mean it's really just a readability thing, so I can change that if you request it.

So bottom line currently if I put the split events above the 2-in-1 events the split ones would technically work better, but as is they're at least consistent with the 2-in-1 versions. Which don't really care which events are sent first since both lists are finalized by the time the events start sending.

I'm going to hold off on pushing until the CI changes are done. I can if you really want me to though.

library/include/modules/EventManager.h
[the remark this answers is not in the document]
myk002: I mean the handler array in eventful is keyed of the values of this enum. By adding additional event types, the array becomes misaligned. This is what caused the crash in #2059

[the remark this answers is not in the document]
cppcooper: Oh, this will cause a new 2059 then. I'll dig into eventful to make sure it doesn't for these new events

plugins/devel/event-testing.cpp
```diff
@@ -0,0 +1,164 @@
+#include "Core.h"
```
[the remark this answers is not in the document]
myk002: I meant more a `devel/eventful.lua` script that just registers handlers for everything and echoes to the screen when the callbacks are called. This would at least allow you to manually verify that the event detection -> event triggering -> Lua marshalling chain basically functions for all event types. I would expect that kind of test to have been run before the code is merged. Some of the types we're working with here are not well understood, and we're making a significant (and risky) change to the API in that we're allowing events to fire in a tick other than the one where the event was detected.

Writing automated regression tests that trigger all the events would be very challenging, but it might be possible for a few event types. It could be done for building creation pretty easily, for example. Quickfort even has an API that you can use to create and destroy buildings with a single call.

library/include/modules/EventManager.h
myk002: Just to be clear, this was the ordering I was suggesting:
BUILDING_DESTROYED
BUILDING_CREATED
BUILDING
CONSTRUCTION_REMOVED
CONSTRUCTION_ADDED
CONSTRUCTION

This will allow us to avoid bugs like the one currently in the buildings cache - described at #1984 (at the bottom of the description comment). Once the split events are available, I'll migrate the buildings cache to use them, but they will only be useful if destroyed events come before created events.

plugins/devel/event-testing.cpp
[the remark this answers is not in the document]
cppcooper: ah, ok a devel/ script definitely doesn't seem impossible. I'll try to write one before the end of the month, since it's lua I'm less keen

library/include/modules/EventManager.h
cppcooper: Interesting bug. I'll drop the consistency since it sounds like putting the split events first allows you (and others?) to transition to them better.

myk002: Gah, sorry, to be even clearer, it's just the relative positions of 
BUILDING_DESTROYED
BUILDING_CREATED
that really matter here. The original BUILDING event can go before or after those two. Or even between. As long as BUILDING_DESTROYED goes before BUILDING_CREATED

cppcooper: If `BUILDING` came first then the order of the other two wouldn't matter (literally). Their processing would be redundant so nothing would change in the containers holding the events. I've put the destroyed first for each, then the created, then the old events

library/include/modules/EventManager.h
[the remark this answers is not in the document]
cppcooper: Inside EventManager.cpp I've taken to dynamically building the event manager array with this loop, and a switch function to return the event managers.

```cpp
std::array<eventManager_t,EventType::EVENT_MAX> compileManagerArray() {
    std::array<eventManager_t, EventType::EVENT_MAX> managers{};
    auto t = (EventType::EventType) 0;
    while (t < EventType::EVENT_MAX) {
        managers[t] = getManager(t);
        t = (EventType::EventType) int(t + 1);
    }
    return managers;
}
```

When I look at eventful.cpp I'll try to do something similar. There is still something to update when event types are added/removed, but the order doesn't need to be synced manually anymore.

library/modules/EventManager.cpp
```diff
@@ -173,28 +173,88 @@ static void manageInteractionEvent(color_ostream& out);
...
+        case EventType::REPORT:
+            return manageReportEvent;
+        case EventType::UNIT_ATTACK:
+            return manageUnitAttackEvent;
+        case EventType::UNLOAD:
+            return manageUnloadEvent;
+        case EventType::INTERACTION:
+            return manageInteractionEvent;
+        case EventType::EVENT_MAX:
+            return nullptr;
```
myk002, line 220: Maybe add a comment here that you are deliberately not including a default case so that the compiler will warn/error on missing enums

cppcooper: sounds like a plan

cppcooper: done

cppcooper: > Also in eventful.cpp

Do you think it might be more beneficial to actually include in eventful? (a default case)

cppcooper:
I was thinking about unit testing, and I had this idea. So I wrote it out.. not sure how well it would work. It would at least simplify the code generation.
```cpp
/* EVENT_TEST and the following are intended to aid
 * in the creation of a unit test for events...
 * it can't help create testing conditions...
 * but it can help ease testing
 */
#define EVENT_TEST(ev,type,data_type)                   \
    data_type q;                                        \
    void on_##ev(color_ostream&, void* d);              \
    EM::EventHandler ev(on_##ev,0);                     \
                                                        \
    void on_##ev(color_ostream&, void* d) {             \
        static bool run_once = false;                   \
        if(!run_once) {                                 \
            run_once = true;                            \
            q = (data_type)d;                           \
            EM::unregister(type, ev, plugin_self);      \
        }                                               \
    }                                                   \
                                                        \
    void on_##ev##2(color_ostream&, void* d);           \
    EM::EventHandler ev##2(on_##ev##2,1200);            \
                                                        \
    void on_##ev##2(color_ostream&, void* d) {          \
        static bool run_once = false;                   \
        if(!run_once) {                                 \
            run_once = true;                            \
            assert( q == (data_type)d );                \
            EM::unregister(type, ev##2, plugin_self);   \
        }                                               \
    }                                                   \
                                                        \
    void register_##ev(){                               \
        EM::registerListener(type, ev, plugin_self);    \
    }

EVENT_TEST(job_initiated,Event::JOB_INITIATED,df::job*)
```

library/modules/EventManager.cpp
myk002: now that the construction of the array is "safe", we *could* put a default case in, but why would we ever want an event to *not* be handled by eventful? is there such a thing as a C++-only event?

myk002:
That would be useful for a C++ unit test, but the problem is that we don't have any infrastructure for running C++ unit tests. All the C++ code is exercised over the Lua interface layer by unit tests written in Lua : /

myk002:
Here's a client you can use for eventful testing:
```
-- Testing script for eventful events

--[====[
devel/eventful-client
=====================

Usage::

    devel/eventful-client help
    devel/eventful-client add <event type> <frequency>
    devel/eventful-client add all <frequency>
    devel/eventful-client list
    devel/eventful-client clear

:help:  shows this help text and a list of valid event types
:add:   add a handler for the named event type at the requested tick frequency
:list:  lists active handlers and their frequencies
:clear: unregisters all handlers

Note this script does not handle the eventful reaction or workshop events.
]====]

local eventful = require('plugins.eventful')

local utils = require('utils')

rng = rng or dfhack.random.new()

handlers = handlers or {}

local function get_event_fn_name(event_type_name)
    local fn_name = 'on'
    for word in event_type_name:gmatch('[^_]+') do
        fn_name = fn_name .. word:sub(1,1) .. word:sub(2):lower()
    end
    if eventful[fn_name] then return fn_name end
    fn_name = fn_name .. 'CreatedDestroyed'
    if eventful[fn_name] then return fn_name end
    return nil
end

local function get_event_registry()
    local registry = {}
    local event_list = utils.invert(eventful.eventType)
    for i=0,#event_list do
        local event_type_name = event_list[i]
        if event_type_name == 'EVENT_MAX' then
            return registry
        end
        registry[i+1] = {etype=event_type_name,
                         fn=get_event_fn_name(event_type_name)}
    end
end

local function help()
    print(dfhack.script_help())
    print()
    print('Event types (and the eventful event functions they map to):')
    for _,v in ipairs(get_event_registry()) do
        print(('  %s -> %s'):format(v.etype, v.fn))
    end
end

local function make_handler_fn(registry_entry, freq)
    return function()
        print(('eventful-client: %s received %s event (freq=%d)')
              :format(os.date("%X"), registry_entry.etype, freq))
    end
end

-- adds a new handler for the specified event type. uses a unique handler name
-- each time so multiple handlers can be registered for the same event.
local function add_one(registry_entry, freq)
    if not registry_entry.fn then
        print(('no eventful event function for %s events; skipping')
              :format(registry_entry.etype))
        return
    end
    eventful.enableEvent(eventful.eventType[registry_entry.etype], freq)
    local handler_name = 'eventful-client.'..tostring(rng:random())
    local handler = make_handler_fn(registry_entry, freq)
    print(('eventful-client registering new %s handler at freq %d: %s')
          :format(registry_entry.etype, freq, handler_name))
    eventful[registry_entry.fn][handler_name] = handler
    handlers[handler_name] = freq
end

local function add(spec, freq)
    for _,v in ipairs(get_event_registry()) do
        if spec == 'all' or spec == v.etype then
            add_one(v, freq)
        end
    end
end

local function iterate_handlers(fn)
    local count = 0
    for _,v in ipairs(get_event_registry()) do
        if not eventful[v.fn] then goto continue end
        for k in pairs(eventful[v.fn]) do
            if k:startswith('eventful-client.') then
                count = count + 1
                fn(v, k)
            end
        end
        ::continue::
    end
    return count
end

local function list_fn(registry_entry, handler_name)
    print(('  %s -> %s (freq=%s)')
          :format(registry_entry.etype, handler_name,
                  handlers[handler_name] or 'unknown'))
end

local function list()
    print('Active handlers:')
    if 0 == iterate_handlers(list_fn) then
        print('  None')
    end
end

local function clear_fn(registry_entry, handler_name)
    list_fn(registry_entry, handler_name)
    eventful[registry_entry.fn][handler_name] = nil
    handlers[handler_name] = nil
end

local function clear()
    print('Clearing handlers:')
    iterate_handlers(clear_fn)
end

local action_switch = {
    add=add,
    list=list,
    clear=clear,
}
setmetatable(action_switch, {__index=function() return help end})

local args = {...}
local action = table.remove(args, 1) or 'help'

action_switch[action](table.unpack(args))
```

library/modules/EventManager.cpp
cppcooper: if we always want events to be exported to lua, then it would make more sense to just do what eventful does in the event manager over having a plugin provide that

myk002:
From playing with the client, it looks like eventful is lowest-frequency based, so if multiple (lua) listeners are registered for an event, they will all fire at the minimum of all requested frequencies for that event.

cppcooper:
> From playing with the client, it looks like eventful is lowest-frequency based, so if multiple (lua) listeners are registered for an event, they will all fire at the minimum of all requested frequencies for that event.

This just furthers my point above:
> if we always want events to be exported to lua, then it would make more sense to just do what eventful does in the event manager over having a plugin provide that

myk002:
also, `devel/eventful-client` doesn't find the new unit handler function since it is named differently from the related event. I assume you'll want to fix that so I didn't special-case it in the client code. It will pick it up once you rename the function from `onUnitNewActive` to `onNewUnitActive` (or was it the other way around... is is the event type name that needs changing? well, anyway, once they line up, the script will find it)

myk002:
> > if we always want events to be exported to lua, then it would make more sense to just do what eventful does in the event manager over having a plugin provide that

I'm not so sure -- I think there is value in not having EventManager have to think about Lua. That's too much complexity in one place. There'd still have to be a Lua translation layer that maps EventManager constructs onto Lua constructs. I think that separating that out into a different unit is the right move. Whether that unit is in core or in a plugin is a different discussion, but I'm not sure if it would change the algorithm much. It would have the same API access to EventManager.

PR #2051, opened 2022-03-26
[manipulator] Tweak colors to make cursor easier to locate

softmoth:
Use standout colors instead of just a darker foreground to indicate cursor position.

I spend half my time in manipulator jiggling the cursor around trying to spot it. This fixes that for me.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -52,6 +52,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `gui/create-item`: Added "(chain)" annotation text for armours with the [CHAIN_METAL_TEXT] flag set
 - DFHack log messages now have configurable headers (e.g. timestamp, origin plugin name, etc.) via the ``debugfilter`` command of the `debug` plugin
 - Script execution log messages (e.g. "Loading script: dfhack_extras.init" can now be controlled with the ``debugfilter`` command. To hide the messages, add this line to your ``dfhack.init`` file: ``debugfilter set Warning core script``
+- `manipulator`: Tweak colors to make the cursor easier to locate
 
 ## Documentation
 - Add more examples to the plugin skeleton files so they are more informative for a newbie
--- a/plugins/manipulator.cpp
+++ b/plugins/manipulator.cpp
@@ -2083,7 +2083,10 @@ void viewscreen_unitlaborsst::render()
                 }
             }
             else
+            {
+                fg = COLOR_WHITE;
                 bg = COLOR_CYAN;
+            }
             if ((col_offset == sel_column) && (row_offset == sel_row))
             {
                 fg = is_labor_set ? COLOR_WHITE : COLOR_GREY;
--- a/plugins/manipulator.cpp
+++ b/plugins/manipulator.cpp
@@ -2052,11 +2052,10 @@ void viewscreen_unitlaborsst::render()
         for (int col = 0; col < col_widths[DISP_COLUMN_LABORS]; col++)
         {
             int col_offset = col + first_column;
-            fg = 15;
-            bg = 0;
+            bool is_labor_set = columns[col_offset].labor != unit_labor::NONE;
+            fg = COLOR_WHITE;
+            bg = COLOR_BLACK;
             uint8_t c = 0xFA;
-            if ((col_offset == sel_column) && (row_offset == sel_row))
-                fg = 9;
             if (columns[col_offset].skill != job_skill::NONE)
             {
                 df::unit_skill *skill = NULL;
@@ -2072,17 +2071,22 @@ void viewscreen_unitlaborsst::render()
                 else
                     c = '-';
             }
-            if (columns[col_offset].labor != unit_labor::NONE)
+            if (is_labor_set)
             {
                 if (unit->status.labors[columns[col_offset].labor])
                 {
-                    bg = 7;
+                    bg = COLOR_GREY;
                     if (columns[col_offset].skill == job_skill::NONE)
                         c = 0xF9;
                 }
             }
             else
-                bg = 3;
+                bg = COLOR_CYAN;
+            if ((col_offset == sel_column) && (row_offset == sel_row))
+            {
+                fg = is_labor_set ? COLOR_BLUE : COLOR_GREY;
+                bg = COLOR_WHITE;
+            }
             Screen::paintTile(Screen::Pen(c, fg, bg), col_offsets[DISP_COLUMN_LABORS] + col, 4 + row);
         }
     }
--- a/plugins/manipulator.cpp
+++ b/plugins/manipulator.cpp
@@ -2052,7 +2052,7 @@ void viewscreen_unitlaborsst::render()
         for (int col = 0; col < col_widths[DISP_COLUMN_LABORS]; col++)
         {
             int col_offset = col + first_column;
-            bool is_labor_set = columns[col_offset].labor != unit_labor::NONE;
+            bool is_labor_set = false;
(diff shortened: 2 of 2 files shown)


myk002 (approved):
I'll need to test to see what this looks like, but the code changes look good

myk002 (approved):

myk002:
No need to redo the commits and force push, by the way. I can squash all commits in this PR into a single commit in the upstream branch. It doesn't matter how many commits are in the PR.

softmoth:
Thanks for your help. I've pushed another fix for logic I'd messed up. I'm complete with this now.

myk002:
I think there's still a bug in here. I started two instances of DF. The top one is the current develop branch. The bottom one is built from your PR. For some fields, I see the intended change: https://drive.google.com/file/d/1oKVbRMstsVlWPFrwB_P4wHr10G2-ln2Q/view?usp=sharing

but for other fields, the blue highlight just disappears. In this example, I have the 'A' in the upper left corner selected in both windows: https://drive.google.com/file/d/1cJf0BXn9ilY_HHEmxTLABW6RfI1sM2Gx/view?usp=sharing

softmoth:
Thanks for checking so thoroughly. I think this is actually showing the correct behavior, and why this change was needed. If I understand right, the top dwarf is NOT set as a miner, even though they have legendary skill at it. So it should be greyed out.

But if it's confusing you, probably it will confuse others. I was about 50/50 for using a colored background, like the LIGHTBLUE option I just pushed. I suspect it may be a better choice:

Labor OFF
labor off (black on white)

Labor ON
Labor ON (black on blue)

lethosor:
What if we used a dark blue background for highlighted + disabled, and the old white background + blue text for highlighted + enabled? 

I'm not sure how to indicate clearly which labors are enabled if we change the background color of enabled labors. If we use two different background colors for highlighted+enabled and highlighted+disabled, we have a different problem: you can tell which labor is highlighted, but you have to jiggle the cursor to tell whether it's enabled or not (until you memorize the colors after doing this enough). My initial guess from looking at your screenshots in your last comment, without looking at the captions, was that the _second_ one is highlighted but disabled (or "OFF").

I personally have adjusted my blue colors to make them easier to see. Regardless, manipulator also highlights both the row and column that are selected, which should help people spot the rough area of the selected cell. Making things like this stand out when colors are customizable is tricky...

lethosor:
Perhaps a better way of phrasing it: the most visually obvious indicator that a labor is enabled is currently its white background color. If we change the background color of highlighted labors (e.g. to blue), it's less obvious whether the highlighted labor is enabled or disabled.

softmoth:
I offer one more iteration that I'd not tried yet, and I think it works well. I might have to use it a while to see, but it seems obvious now that I've tried it and I think it might be the best compromise:

Labor OFF
shot-20220328 4

Labor ON
shot-20220328 5

It's still easy enough to spot the cursor in the field of cells, and the ON labors are strictly more emphasized than the OFF labors (rather than the confusing "is blue ON because it's 'special' or 'more colorful', or is white ON because it's brighter?").

Obviously I'm glossing over the fact that the black-on-grey cursor and white-on-grey non-selected ON labor could be confused. But with the fg color being black (off) and white (on), I think it's about as legible as any combo could be.

Anyways, I do hope something along these lines is acceptable, as I find any of them a drastic improvement and can't imagine going back. But I get it if you find it a step in the wrong direction.

myk002:
UX design is hard : (
You were right, of course, that the miner's mining labor was off in my screenshot and I was incorrect about the conclusions I drew. Let me play around with your latest version

lethosor:
> Labor OFF shot-20220328 4
> 
> Labor ON shot-20220328 5

Sorry to continue disagreeing, but I find the difference between a white and light grey background to be more subtle than the original implementation, especially since the text color is still the same. Maybe a blue text color would work? So in Myk's format:

|                   | No cursor      | Cursor                     |
|-------------------|----------------|----------------------------|
| Labor enabled     | Black on Gray  | Blue on Gray              |
| Labor not enabled | White on Black | Black/Gray on Blue |

myk002:
That looks workable. I played with the colors a little bit, and this is what I ended up with for that scheme:

|                   | No cursor         | Cursor                |
|-------------------|-------------------|-----------------------|
| Labor enabled     | fg=BLACK bg=WHITE | fg=LIGHTBLUE bg=GREY  |
| Labor not enabled | fg=WHITE bg=BLACK | fg=BLACK bg=LIGHTCYAN |

and here's the diff:
```
@@ -2076,7 +2076,8 @@ void viewscreen_unitlaborsst::render()
                 if (unit->status.labors[columns[col_offset].labor])
                 {
                     is_labor_set = true;
-                    bg = COLOR_GREY;
+                    fg = COLOR_BLACK;
+                    bg = COLOR_WHITE;
                     if (columns[col_offset].skill == job_skill::NONE)
                         c = 0xF9;
                 }
@@ -2085,8 +2086,8 @@ void viewscreen_unitlaborsst::render()
                 bg = COLOR_CYAN;
             if ((col_offset == sel_column) && (row_offset == sel_row))
             {
-                bg = is_labor_set ? COLOR_LIGHTBLUE : COLOR_WHITE;
-                fg = COLOR_BLACK;
+                fg = is_labor_set ? COLOR_LIGHTBLUE : COLOR_BLACK;
+                bg = is_labor_set ? COLOR_GREY : COLOR_LIGHTCYAN;
```

softmoth:
I offer two others. If these don't seem clearly better than what's there, maybe this issue should be tabled. I can patch it myself, as it's important to me but I understand the visuals are subjective.

Labor ON:
shot-20220329

Labor OFF:
shot-20220329 1

To me, any of the proposed changes is an improvement, but I feel I've taken up too much of your time already. Thanks again for your attention, and let me know if I can help further.

lethosor:
I'm finding the implementation in [0ea3aa2](https://github.com/DFHack/dfhack/pull/2051/commits/0ea3aa256cbdcadcac791398b813ca2e0320be64) hard to see in some spots, but it also doesn't line up with your last screenshot, so I'm not sure if that's intentional:

image

compared to `develop`'s appearance:

image


I guess my rule of thumb is: when an existing cell is highlighted, either its foreground or background color should change - not both. I'm no UI designer, so it's just based on what I think is most "discoverable".

I might try combining some of these approaches to see if I can suggest something as well.

lethosor:
All right, https://github.com/lethosor/dfhack/commit/cc501aa4c6d48ee2a71b11324eaf941dafbe7286 is my suggestion. I sort of violated my own rule by changing between black and blue as the background color for non-vs-highlighted cells. They are similar in intensity, so my hope is that they are both easy to interpret as "disabled". I also switched to light green as the text color for highlighted cells. It seemed more visible than yellow to me, but light cyan would probably be just as visible.

Worth noting, again, that my color palette may differ from others.

Here are some screenshots:

<details>

image

image

image

image

image

</details>


I also looked into blinking the cursor. I wasn't a huge fan of it myself, but blinking between the normal and highlighted colors is relatively unambiguous. (I just tossed in a `&& (enabler->last_tick / 500 % 2)` in a conditional, if you want to try it yourself.)

myk002:
I'm happy with this latest iteration. it looks and feels natural.

myk002:
it looks like I merged a conflict in `changelog.txt`. let me just resolve that conflict. you might have to pull this branch if you make further local changes.

softmoth:
I rebased to dfhack/develop, so there's no conflict.

lethosor:
I like this too. The foreground color staying the same when a cell is highlighted is a nice touch.

PR #2053, opened 2022-03-27
Typo Fix in dig.cpp

DwarvenM:
Changed line 1277 to say "This is a vein. Use digv instead!/n" instead of "This is a vein. Use vdig instead!/n"


myk002 (approved):

myk002:
Note to self: add changelog entry after merge

PR #1918, opened 2021-08-07
Adds new plugin: Spectate

cppcooper:
Relies on [PR#1876](https://github.com/DFHack/dfhack/pull/1876) - EventManager refactor

Adds a simple plugin for adding a spectate mode

PR #2056, opened 2022-03-29
Fix 2031 by revert

### LINKED ISSUE #2031: EventManager can miss sending events to handlers

cppcooper:
Fixes #2031 by way of `git checkout master -- library/modules/EventManager.cpp` then re-adding the new event types.

### DIFF
--- a/library/include/modules/EventManager.h
+++ b/library/include/modules/EventManager.h
@@ -22,7 +22,7 @@ namespace DFHack {
                 JOB_INITIATED,
                 JOB_STARTED, //has a worker
                 JOB_COMPLETED,
-                NEW_UNIT_ACTIVE,
+                UNIT_NEW_ACTIVE,
                 UNIT_DEATH,
                 ITEM_CREATED,
                 BUILDING,
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -57,24 +57,11 @@ static multimap<int32_t, EventHandler> tickQueue;
 
 //TODO: consider unordered_map of pairs, or unordered_map of unordered_set, or whatever
 static multimap<Plugin*, EventHandler> handlers[EventType::EVENT_MAX];
-static map<EventHandler::callback_t,int32_t> eventLastTick;
+static int32_t eventLastTick[EventType::EVENT_MAX];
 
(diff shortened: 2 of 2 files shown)


myk002:
does this keep the new behavior (`freq` per handler) for tick events? or is that too complicated to add back without changing the other events too?

cppcooper:
This keeps the old behaviour. All I did was literally checkout the file from master and add the additional event types. No refactoring was done (aside from renaming the `NEW_UNIT_ACTIVE` => `UNIT_NEW_ACTIVE` so that the naming scheme was consistent with the existing `UNIT_DEATH`)

myk002:
Isn't the tick freq configurability required by the spectate plugin? https://github.com/DFHack/dfhack/pull/1918/files#diff-57b4639edbce45b51d81bfb4f7592072c1ebb3e469991755a1b5730b0740179dR67

cppcooper:
Tick Events use a special queue (unordered_multimap iirc). They calculate when they need to fire and then enqueue with that as the key. I might be getting details wrong but I'm 99% sure that the tick events can be any freq and it'll be obeyed.

The other events in that plugin don't use non-zero frequencies.

#1918 relied on #1876 because of the event types it uses. It needs to look at jobs when they are assigned a worker, not just when they are created. I can see how confusion would arise, I play it a bit loose with my PR contents

PR #2057, opened 2022-03-29
Improves support for dfhack as a sub-project

cppcooper:
Replaces `CMAKE_SOURCE_DIR` with `dfhack_SOURCE_DIR` in the cmake config chain

### DIFF
--- a/CMake/DownloadFile.cmake
+++ b/CMake/DownloadFile.cmake
@@ -11,7 +11,7 @@ endfunction()
 
 function(search_downloads FILE_MD5 VAR)
     set(${VAR} "" PARENT_SCOPE)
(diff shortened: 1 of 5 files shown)


lethosor (approved):

myk002:
Interesting build errors. BUILD_TESTS appears to not be processed? And the windows build don't seem to work at all

cppcooper:
> Interesting build errors. BUILD_TESTS appears to not be processed? And the windows build don't seem to work at all

Not sure what to do..
```
Z:/home/buildmaster/work/_E0/src/plugins/devel/check-structures-sanity.cpp(775): error C2589: '(': illegal token on right side of '::'
```
because this file doesn't exist. I'll have to load my windows VM when I have a chance.

PR #2058, opened 2022-03-29
Improves support for dfhack as a sub-project

cppcooper:


myk002 (approved):
Approving because of lethosor's earlier approval and because it passes ci

cppcooper:
CMAKE_SOURCE_DIR points to the top level of a cmake tree

cppcooper:
`project(dfhack)` we believe is what defines `dfhack_SOURCE_DIR` so unless a lower cmake file uses that project name there shouldn't be an issue. If it isn't defined by `project()` then I don't know where it comes from.

Why we need it is to avoid relative paths I suspect. It might be harder to read `../../depends/whatever.wtf` and annoying to write the cmake files needing to think about relativity

PR #2059, opened 2022-03-29
fix crash in eventful due to misaligned fn map

myk002:
the EventManager got a new event type for new active units, but the handler for the event was not added to the array of handlers in `eventful` so the wrong handlers were being called for some events. In particular, the onUnload event was being sent to the wrong handler and was causing a crash when a map was unloaded.

### DIFF
--- a/library/include/modules/EventManager.h
+++ b/library/include/modules/EventManager.h
@@ -17,6 +17,9 @@
 namespace DFHack {
     namespace EventManager {
         namespace EventType {
+            // NOTICE: keep this list synchronized with the eventHandlers array
+            // in plugins/eventful.cpp or else events will go to the wrong
+            // handlers.
             enum EventType {
                 TICK,
                 JOB_INITIATED,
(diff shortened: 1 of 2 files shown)


lethosor:
To understand the severity: was this issue introduced in r4 or in #2056?

myk002:
> To understand the severity: was this issue introduced in r4 or in #2056?

Dev issue only. It was introduced in #1876, merged after r4 was released. The release process would have caught this issue (crash on save).

lethosor:
Thanks! Sounds like something I'd use the `develop only` label for (at least if it's also labeled bug/crash)

cppcooper:
> > To understand the severity: was this issue introduced in r4 or in #2056?
> 
> Dev issue only. It was introduced in #1876, merged after r4 was released. The release process would have caught this issue (crash on save).

actually he was right, #1876 updated eventful I'm pretty sure. I didn't think to update it in #2056 though after changing the enum name to be consistent with the other unit events.

myk002:
> > > To understand the severity: was this issue introduced in r4 or in #2056?
> > 
> > 
> > Dev issue only. It was introduced in #1876, merged after r4 was released. The release process would have caught this issue (crash on save).
> 
> actually he was right, #1876 updated eventful I'm pretty sure. I didn't think to update it in #2056 though after changing the enum name to be consistent with the other unit events.

The array in eventful was updated for the new job started event, but not the new new unit event. So it ended up off by one instead of off by two.

PR #2060, opened 2022-03-29
[core] Debug logging facelift

### LINKED ISSUE #2045: Make script execution logging verbosity configurable
So was wondering if Script logging could be optional so that it doesn't end up mass printing logs the in a game session mostly with OnMapLoad, SC_WORLD_LOADED and any onLoad stuff as for folks who constantly shuffle between saves or save scum this would start to get annoying quick on a causal playthrough, otherwise it's seems like neat option for debugging.

myk002:
Fixes #2045 

Init script execution logging will now be configurable via the `debugfilter` command. To turn off script execution logging, add the following line to your `dfhack.init` file:
```
debugfilter set Warning core script
```

This PR, however, is mostly about allowing debug log output to blend seamlessly into our existing messages, and there are a number of policy questions to be answered. Since none of our code really uses the debug log yet, changing things won't break anything, but we do need to decide how things should work going forward, as adoption of the logging framework should increase in the future.

Before this change, log messages output through the debug logging framework had the following format:
`12:27:22.306:t1:core:script: Loading script: dfhack.init`

This is very verbose compared to the line I intended to make filterable (just the message `Loading script: dfhack.init`). If we want the "verbosity-controllable" messages to blend in with our current message output, we'd need to get rid of that long ugly header. However, there will still be cases where the header would be useful, so we can't just delete it wholesale.

Therefore, I made the elements of the header individually controllable with flags (all defaulting to false), and added a new `header` command to the `debugfilter` plugin as a config interface.

All "normal" output messages are currently gray, and I wanted to make INFO messages blend in with those messages. Debug log messages are color coded according to their verbosity. Before this change, INFO messages were cyan and TRACE messages were gray. I changed INFO to gray (again, in order to blend in with "normal" messages) and I changed TRACE messages to brown, which I thought was suitably "tracey".

Finally, the debug log code has the capability to excise all messages below a certain threshold from the binary. It was set to remove all messages below INFO if you compile with NDEBUG and allow all messages otherwise. However, there is no build configuration that doesn't define NDEBUG, so it would effectively make TRACE and DEBUG messages useless. I moved the lower bound to DEBUG so users could potentially see DEBUG level messages to add to bug reports. We could move this down to TRACE as well (for the same reasons), but I thought for now I'd reserve that for devs who do somehow manage to compile with NDEBUG not set.

Also doc updates for the Debug API.

### DIFF
--- a/library/Debug.cpp
+++ b/library/Debug.cpp
@@ -80,11 +80,11 @@ static color_value selectColor(const DebugCategory::level msgLevel)
 {
     switch(msgLevel) {
     case DebugCategory::LTRACE:
-        return COLOR_GREY;
+        return COLOR_BROWN;
     case DebugCategory::LDEBUG:
         return COLOR_LIGHTBLUE;
     case DebugCategory::LINFO:
-        return COLOR_CYAN;
+        return COLOR_GREY;
     case DebugCategory::LWARNING:
         return COLOR_YELLOW;
     case DebugCategory::LERROR:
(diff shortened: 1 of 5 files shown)


library/Debug.cpp
```diff
@@ -80,11 +80,11 @@ static color_value selectColor(const DebugCategory::level msgLevel)
 {
     switch(msgLevel) {
     case DebugCategory::LTRACE:
-        return COLOR_GREY;
+        return COLOR_BROWN;
     case DebugCategory::LDEBUG:
         return COLOR_LIGHTBLUE;
     case DebugCategory::LINFO:
-        return COLOR_CYAN;
+        return COLOR_GREY;
```
lethosor, line 87: I use a light terminal theme (weird, I know), so I would somewhat prefer COLOR_RESET:

image

over COLOR_GREY:

image

...although I can still read either. I'd definitely like to change the COLOR_WHITE below to COLOR_RESET, for obvious reasons.

Maybe "INFO" should just fall through to the default color. Seems like a pretty "normal" type of message in general, to me.

lethosor (approved):
> there is no build configuration that doesn't define NDEBUG

I think `Debug` does? It's not supported on Windows (we had a recent Discord conversation on this), but I vaguely remember seeing `assert()`s trigger on macOS in `Debug` mode years ago. It's possible we have changed things since then.

Anyway, I think stripping out TRACE messages at compile-time and leaving DEBUG messages in sounds perfectly reasonable.

LGTM, and thanks for the various typo/etc. fixes!

library/Debug.cpp
myk002: done

PR #2061, opened 2022-03-30
git-describe: fail earlier if git fails to run or doesn't exist

lethosor:
In cases where the cached `GIT_EXECUTABLE` does not exist (don't ask...), `git-describe.cmake` would fail silently, resulting in a `git-describe.h` with missing definitions that caused `DFHackVersion.cpp` to fail to compile.

This change produces a more descriptive error earlier in the build process.


myk002:
passes CI (though I know the effect of this change is more felt in local builds); merging

PR #2064, opened 2022-03-31
only designate enough trees to reach max_logs

### LINKED ISSUE #876: Make autochop estimate the number of logs you get from a tree

Moth-Tolias:
redo of https://github.com/DFHack/dfhack/pull/1938 because... reasons.

only half implements https://github.com/DFHack/dfhack/issues/876 - i have zero idea about how to go about designating the largest / oldest / trees that won't grow anymore, so if anyone has any ideas on that, please let me know! - but the first part, only designating the amount of trees required to get to max_logs, is here and worked in my testing seven months ago.

Edit from myk002: Fixes #876

### DIFF
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -340,7 +340,7 @@ static int do_chop_designation(bool chop, bool count_only, int *skipped = nullpt
 
     //designate
     multimap<int, int>::iterator itr;
-    for (size_t itr = trees_by_size.begin(); itr != trees_by_size.end(); itr++)
+    for (itr = trees_by_size.begin(); itr != trees_by_size.end(); itr++)
     {
         const df::plant *plant = world->plants.all[itr->second];
 
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -47,6 +47,8 @@ DFHACK_PLUGIN("autochop");
 REQUIRE_GLOBAL(world);
 REQUIRE_GLOBAL(ui);
 
+static int get_log_count();
+
 static bool autochop_enabled = false;
 static int min_logs, max_logs;
 static const int LOG_CAP_MAX = 99999;
--- a/plugins/autochop.cpp
+++ b/plugins/autochop.cpp
@@ -307,6 +307,41 @@ static int estimate_logs(const df::plant *plant)
     return trunks;
 }
 
+static int get_log_count()
+{
+    std::vector<df::item*> &items = world->items.other[items_other_id::IN_PLAY];
+
+    // Pre-compute a bitmask with the bad flags
+    df::item_flags bad_flags;
+    bad_flags.whole = 0;
+
+#define F(x) bad_flags.bits.x = true;
+    F(dump); F(forbid); F(garbage_collect);
+    F(hostile); F(on_fire); F(rotten); F(trader);
+    F(in_building); F(construction); F(artifact);
+    F(spider_web); F(owned); F(in_job);
+#undef F
+
+    size_t valid_count = 0;
+    for (size_t i = 0; i < items.size(); i++)
+    {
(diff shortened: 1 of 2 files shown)


plugins/autochop.cpp
```diff
@@ -415,6 +378,71 @@ static int get_log_count()
     return valid_count;
 }
 
+static int do_chop_designation(bool chop, bool count_only, int *skipped = nullptr)
```
myk002, line 381: Would you mind moving the function back to its original location so the diff is cleaner? It's hard to see what has changed.

myk002:
It occurs to me that you could approximate the "oldest trees first" requirement by ordering the trees by size before you iterate through them and start marking for chopping.

plugins/autochop.cpp
Moth-Tolias: done

plugins/autochop.cpp
```diff
@@ -288,18 +292,41 @@ static bool skip_plant(const df::plant * plant, bool *restricted)
...
     }
+
     for (size_t i = 0; i < world->plants.all.size(); i++)
     {
         const df::plant *plant = world->plants.all[i];
-
         bool restricted = false;
+
+        if ((estimated_yield >= max_logs) && chop)
+            continue;
```
myk002, line 328: why continue and not break?

myk002:
overall, looks good (that raw pointer dereferencing makes me nervous, but I realize that's just how the data structure is laid out and I don't think there's any better way to do it)

what do you think about using a multimap to iterate through the trees in order of expected yield?

also needs a changelog entry.

Moth-Tolias:
> what do you think about using a multimap to iterate through the trees in order of expected yield?

sounds good to me. looking up how to use multimaps now.

> also needs a changelog entry.

right on.

plugins/autochop.cpp
```diff
@@ -288,18 +292,40 @@ static bool skip_plant(const df::plant * plant, bool *restricted)
...
+    }
+
+    return trunks;
+}
+
 static int do_chop_designation(bool chop, bool count_only, int *skipped = nullptr)
 {
     int count = 0;
+    int estimated_yield = get_log_count();
+    multimap<int, int> trees_by_size;
```
myk002, line 316: Mapping int to `df::plant *` would simplify lookups

plugins/autochop.cpp
```diff
@@ -288,18 +292,40 @@ static bool skip_plant(const df::plant * plant, bool *restricted)
...
+
     if (skipped)
     {
         *skipped = 0;
     }
+
+    //get trees
     for (size_t i = 0; i < world->plants.all.size(); i++)
```
myk002, line 324: Since you don't really need the index, you can simplify this to:
`for (auto plant : world->plants.all)`

plugins/autochop.cpp
```diff
@@ -309,6 +335,18 @@ static int do_chop_designation(bool chop, bool count_only, int *skipped = nullpt
             continue;
         }
 
+        trees_by_size.insert(pair<int, int>(estimate_logs(plant), i));
+    }
+
+    //designate
+    multimap<int, int>::iterator itr;
+    for (itr = trees_by_size.begin(); itr != trees_by_size.end(); itr++)
```
myk002, line 343: You can simplify this to:
`for (auto & entry : trees_by_size)`
And then get the plant via:
`df::plant * plant = entry.second;`

myk002 (changes requested):

myk002 (approved):

myk002:
I'd say this fully fixes #876

PR #2074, opened 2022-04-03
[tweak/stable-cursor] Fix for when default menu is off

softmoth:
When the default menu is off, the viewport shifts by 16 (half the menu
width), but getViewportPos() does not report that until after the
interpose vmethod is over. So the view is always off by 16 in the x
direction.

The check_menu() function is a simplified version of what is in
Gui::getDwarfmodeViewDims().

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -41,6 +41,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `autofarm` removed restriction on only planting 'discovered' plants
 - `luasocket` (and others): return correct status code when closing socket connections
+- `tweak`: stable-cursor now works when the default menu is hidden
 
 ## Misc Improvements
 - `dig-now`: handle fortification carving
--- a/plugins/tweak/tweaks/stable-cursor.h
+++ b/plugins/tweak/tweaks/stable-cursor.h
@@ -11,6 +11,7 @@ using df::global::ui_build_selector;
 using df::global::ui_menu_width;
 
 static df::coord last_view, last_cursor;
+const int16_t menu_shift = Gui::MENU_WIDTH / 2 + 1;
 
 struct stable_cursor_hook : df::viewscreen_dwarfmodest
 {
@@ -32,6 +33,11 @@ struct stable_cursor_hook : df::viewscreen_dwarfmodest
         }
     }
 
+    bool check_menu()
+    {
+        return ui_menu_width && (*ui_menu_width)[0] < (*ui_menu_width)[1];
+    }
+
     DEFINE_VMETHOD_INTERPOSE(void, feed, (set<df::interface_key> *input))
     {
         bool was_default = check_default();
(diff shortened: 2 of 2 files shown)


softmoth:
This fixes the problem everywhere except at the right margin, it seems. Unfortunately getDwarfmodeViewDims() does not update its info after the map changes, I think it's cached as part of the hook mechanism. But I think I can fix it using a ScreenSize() call and manually adjusting for the menu and area width (essentially re-implementing portions of getDwarfmodeViewDims() to avoid stale cached results). I'll update this issue once I've worked more along these lines.

softmoth:
This fixes it everywhere, except for switching into the build menu (`b`). The logic between `is_default` and *is the menu showing* is using some wrong assumptions in this case.

I will try to figure out what is going on with that.

I'd appreciate anyone testing this, though, to ensure that it really does work as expected. I've tested it with 80x24 and my full screen of 140 tiles wide.

softmoth:
OK, it is now working in all circumstances as far as I can tell. Also, I was able to use Gui::getDwarfmodeViewDims() after all; even though the screen locations don't update during the tweak hook, there's enough info in there to construct the needed dimensions, and it's certainly less extra brittle code than repeating chunks of it. This lets me get rid of my check_menu() and check_area() functions.

There are a few more fixes I would like to make, but they are beyond the scope of this fix:

- When opening a sidebar in a new spot, where stable-cursor doesn't apply, the game shifts the cursor over 16 tiles from where it should be when the menu is forced onto the screen. The view is centered correctly, but the cursor is off to the left rather than in the middle of the view. This is very annoying.
- When the `last_cursor` position is in the extreme left or right 16 tiles of the map, the cursor is not visible when the sidebar first opens. Because the viewport has shrunk by 16 tiles on each side, the cursor is hiding in the now-offscreen strip. Only once you move the cursor does the viewport jump over to reveal the cursor. Instead, the stable-cursor tweak should shift the view over to show the edge of the map instantly.

And one more feature I would like to add:

- It would be nice to have a sloppy version of the stable cursor which keeps the cursor location even if the view has moved a little bit. I think if it's within 20 or 30 tiles of the previous location, the old cursor location is still beneficial to reuse.

softmoth:
Gah. I just realized that this *whole thing* is due to TWBT. If I remove `twbt.plug.so`, *the viewport doesn't shift at all* when the forced menu pops in and out.

SO... I think the right approach is to probably to patch TWBT rather than dfhack. I know TWBT hasn't seen any maintenance for a while. I'm open to feedback on how best to proceed if anyone has ideas.

myk002:
> Gah. I just realized that this _whole thing_ is due to TWBT. If I remove `twbt.plug.so`, _the viewport doesn't shift at all_ when the forced menu pops in and out.
> 
> SO... I think the right approach is to probably to patch TWBT rather than dfhack. I know TWBT hasn't seen any maintenance for a while. I'm open to feedback on how best to proceed if anyone has ideas.

That must be annoying to discover after putting all this time into the PR : /

Twbt is getting recent changes [here](https://github.com/thurin/df-twbt). I bet that if you submit a patch then thurin would get it merged.

softmoth:
Thanks. That's encouraging if thurin is still taking patches. And it wasn't a complete waste, as I have a much better feel for some parts of the code.

For my own reference, I believe the culprit is [thurin/df-twbt/renderer.hpp#L174](https://github.com/thurin/df-twbt/blob/master/renderer.hpp#L174) (also `gmenu_w` in line 150).

I'm going to close this issue. Any further developments along these lines deserve a fresh start.

lethosor:
TWBT does hook into `Gui::getDwarfmodeViewDims()` when it's enabled, so I would naively expect an approach that centers the map around the center of `dims.map_*` to work regardless of the state of TWBT. I wasn't closely following the work you were doing here, though.

softmoth:
@lethosor Thank you. It's actually the map centering that causes the problem here. When the menu pops into existence, TWBT shifts the viewport to keep the old center in place (shrinks the viewport from both sides), but vanilla DF leaves the top-left corner in place (shrinks the viewport from the right). This is the source of several bugs when TWBT is on, if I am following everything right.

softmoth:
TWBT patch fixes all of this: https://github.com/thurin/df-twbt/pull/3

Issue #2075, opened 2022-04-06
Add pre-commit


lethosor:
[pre-commit](https://pre-commit.com/) is a tool that provides basic linting functionality out of the box, and is relatively easy to set up locally so that it runs on every commit. There is a complementary [CI service](https://pre-commit.ci/) that can automate these checks on PRs. We could also integrate formatters to address #1535, eventually.

(Opening an issue to cross-link PRs in our different repos.)

ab9rf:
done for dfhack, df-structures, and scripts; possibly still open for other repositories

PR #2077, opened 2022-04-07
pre-commit config

lethosor:
Implements #2075 for this repo

Excluding `depends/` because hooks were touching lua/protobuf code, which we should generally not be touching so that diffs with upstream stay clean.


myk002:
I note that it's uglifying the JSON files something awful. Can we retain the indentation? Also, the files are generated by the orders plugin. We should modify the plugin to create files in a format that pre-commit approved of.

myk002:
Also, I'm not sure how it is used, but modifying plugins/raw/entity_default.diff might make it unusable

myk002:
I'll merge the blueprint changes back into the Google drive spreadsheets

lethosor:
I'm trying to disable the tab check for those JSON files. No luck so far. Good point on the diffs too.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+  rev: 0.14.2
+  hooks:
+    - id: check-github-workflows
+- repo: https://github.com/Lucas-C/pre-commit-hooks
+  rev: v1.1.13
+  hooks:
+    - id: forbid-tabs
+    - id: remove-tabs
+# specific to dfhack:
+exclude: '^(depends/)'
```
myk002, line 31: Would `exclude_types = [json]` do the trick?

lethosor: Supposedly, according to https://pre-commit.com/#pre-commit-configyaml---hooks, we can specify `exclude_types` on a per-hook basis. I tried adding `exclude_types = [json]` to the `*-tabs` hooks above and it did not appear to change their behavior at all. Maybe it is a responsibility of the hook to support that, though.

Honestly, I might tend towards excluding JSON files in the specific problematic folder from all hooks. In this case, `data/examples/orders/*.json` are the only files giving me trouble, and they seem to be generated and likely won't be hand-edited in the future, so style is not really important.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+  rev: v1.1.13
+  hooks:
+    - id: forbid-tabs
+      exclude_types:
+        - json
+    - id: remove-tabs
+      exclude_types:
+        - json
+# specific to dfhack:
+exclude: '^(depends/|data/examples/.*\.json$)'
```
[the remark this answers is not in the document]
lethosor: Oh, good point, changing diffs is a bad idea

myk002 (approved):
lgtm, as long as the tabs -> spaces are ok in `plugins/raw/*.txt`

lethosor:
Yeah, DF seems to handle spaces in raw files (tested with `creature_amphibians.txt`)

PR #2080, opened 2022-04-07
[tweak/stable-cursor] Keep stable cursor when viewport is near enough

softmoth:
Allow the viewport to move a bit and still keep the cursor location. In addition to "enable" and "disable", the distance is now configurable with "tweak stable-cursor 20" (default is 10 tiles).

I chose a relatively small default in case people are using the 80x24 screen, in which case moving 10 tiles up is already near the edge of the screen. This still allows one viewport move (view moves 10 tiles at a time), and I think it feels like a nice compromise.

On a larger viewport, `tweak stable-cursor 30` feels natural to me.

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -442,7 +442,8 @@ Subcommands that persist until disabled or DF quits:
 :nestbox-color:         Fixes the color of built nestboxes
 :reaction-gloves:       Fixes reactions to produce gloves in sets with correct handedness (:bug:`6273`)
 :shift-8-scroll:        Gives Shift-8 (or :kbd:`*`) priority when scrolling menus, instead of scrolling the map
-:stable-cursor:         Saves the exact cursor position between t/q/k/d/b/etc menus of fortress mode.
+:stable-cursor:         Saves the exact cursor position between t/q/k/d/b/etc menus of fortress mode, if the
+                        map view is near enough to its previous position.
 :stone-status-all:      Adds an option to toggle the economic status of all stones
 :title-start-rename:    Adds a safe rename option to the title screen "Start Playing" menu
 :tradereq-pet-gender:   Displays pet genders on the trade request screen
--- a/plugins/tweak/tweaks/stable-cursor.h
+++ b/plugins/tweak/tweaks/stable-cursor.h
@@ -32,6 +32,26 @@ struct stable_cursor_hook : df::viewscreen_dwarfmodest
         }
     }
 
+    bool check_viewport_near_enough()
+    {
+        df::coord view = Gui::getViewportPos();
+        auto dims = Gui::getDwarfmodeViewDims();
+        int view_size = min(dims.map_x2 - dims.map_x1, dims.map_y2 - dims.map_y1);
+        // Heuristic to keep the cursor position if the view is still in the
+        // part of the map that was already visible
+        int near_enough_distance = min(30, view_size / 2);
+
+        return
+            // Is the viewport near enough to the old viewport?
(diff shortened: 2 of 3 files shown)


myk002:
Instead of making it configurable, would it be better to make the wiggle room dependent on current screen size?

softmoth:
Maybe it does. I can try coding that up. I wonder if it violates the principle of least surprise, but probably it won't feel that way if it naturally tracks screen size. Let's see.

plugins/tweak/tweaks/stable-cursor.h
```diff
@@ -32,6 +34,14 @@ struct stable_cursor_hook : df::viewscreen_dwarfmodest
         }
     }
 
+    bool check_viewport_near_enough()
+    {
+        df::coord view = Gui::getViewportPos();
+        return abs(view.x - last_view.x) <= near_enough_distance &&
+            abs(view.y - last_view.y) <= near_enough_distance &&
+            view.z == last_view.z;
```
myk002, line 42: should the z-level matter? Wouldn't it still be valuable to keep the cursor position even if the z-level has changed?

softmoth: In my view it's not helpful; on another z level, what the cursor was referencing is not visible, so there's no reason to place the cursor there. Maybe for people who order their fort with one workshop directly over another or something it would be desirable? I could see it either way I guess.

myk002: Yeah, I could go either way as well. If keeping the cursor on a building (or any specific tile) is the only purpose, then yeah changing z-levels "dirties" the cursor. If the user is expecting the cursor to appear on the same part of the screen when moving z-levels, then it might be unexpected that the cursor jumps back to the middle of the screen.

That said, before I thought to look at this carefully, I never noticed what the stable-cursor tweak did in this situation. How often does a user go into a cursor mode, exit the cursor mode, switch z-levels, and re-enter a cursor mode? Probably not often. If they're switching z-levels with a cursor, they're probably going to stay in the same cursor-having mode while they switch z-levels. tl;dr what you have is fine : )

softmoth:
OK, I made it auto-resize to the smaller of 30 or half the viewport size. It seems to work well.

I also added a check to ensure the cursor position is somewhere within the current view. If the cursor was already near the edge of the view, and then the view moves a little further, it can easily be off screen. Then the view jumps over to the cursor. This can still happen even if the cursor is on the screen, but just close to the edge of the viewport. It's mildly annoying but also useful, and I mainly run into it when testing  but not when just playing with this on. So I think in practice this is a nice addition.

myk002 (approved):

docs/Plugins.rst
```diff
@@ -442,7 +442,8 @@ Subcommands that persist until disabled or DF quits:
 :nestbox-color:         Fixes the color of built nestboxes
 :reaction-gloves:       Fixes reactions to produce gloves in sets with correct handedness (:bug:`6273`)
 :shift-8-scroll:        Gives Shift-8 (or :kbd:`*`) priority when scrolling menus, instead of scrolling the map
-:stable-cursor:         Saves the exact cursor position between t/q/k/d/b/etc menus of fortress mode.
+:stable-cursor:         Saves the exact cursor position between t/q/k/d/b/etc menus of fortress mode, if the
+                        map view is near enough (default: 10 tiles) to its previous position.
```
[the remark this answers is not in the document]
softmoth: Thanks, done.

myk002 (approved):

PR #2082, opened 2022-04-08
Add string:escape_pattern() utility function

myk002:
add `string:escape_pattern()` to dfhack.lua

stolen from `devel/query.lua` since it was such a nice implementation. will migrate scripts to use the common implementation later.

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -2685,7 +2685,7 @@ functions. These are invoked just like standard string functions, e.g.::
   kept in place. If a single word is longer than width, it is split over
   multiple lines. If width is not specified, 72 is used.
 
-* ``string:escape()``
+* ``string:escape_pattern()``
 
   Escapes regex special chars in a string. E.g. ``'a+b'`` -> ``'a%+b'``.
 
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -2644,6 +2644,51 @@ environment by the mandatory init file dfhack.lua:
   Walks a sequence of dereferences, which may be represented by numbers or strings.
   Returns *nil* if any of obj or indices is *nil*, or a numeric index is out of array bounds.
 
+.. _lua-string:
+
+String class extentions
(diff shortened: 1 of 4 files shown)


lethosor (commented):
I think we should name it something like `regex_escape` or `escape_pattern`. You could also want to backslash-escape characters that interfere with parsing strings, e.g. transform `a"b` to `a\"b`.

LGTM otherwise. I'm curious if pre-commit.ci will pass if you rebase

myk002:
Fair enough. I'm fine with either of those two naming options. Maybe `escape_pattern` is more in line with Lua lingo, so I'll go with that.

I'm not so sure about adding backslash escaping to the same function. That would materially change the pattern sent to `string.find`, for example. Another similar function could be written, though. Which scripts/code needs to escape quotes?

Alternately, the structure on the implementation makes it pretty easy to escape anything with anything. It could be a generic escape function, taking optional params that default to the pattern case but can be used more broadly.

`string:escape(chars_to_escape, char_to_escape_with)`

lethosor:
I'm not proposing combining this function with backslash-escaping, but rather suggesting that there are two separate operations, "escape string" and "escape regex", which someone could think of when seeing `escape()`.

Maybe escaping quotes isn't a very common use-case, but there are other types of escaping as well (URL-escaping, for one, which Javascript's `escape()` implements). My point is really that I don't think escaping patterns is common enough to warrant the `escape()` name. Maybe it is in Lua since some string-matching functions take patterns by default (but I thought that most had options to treat their parameters as "plain" text, not patterns? I could be wrong.).

Re pre-commit, that's my mistake - I thought pre-commit was merged in all repos, but forgot that #2077 isn't merged yet.

myk002:
Ah, ack. `escape_pattern` it is then. Fwiw I did check before I wrote this PR that it wasn't so easy as passing a `plain` param. From what I can tell, only `string: find()` supports that. All the others do not.

PR #2083, opened 2022-04-08
automatically manage sidebar modes in MenuOverlay

myk002:
Refactor sidebar mode management code from `gui/blueprint` (and the new `gui/quickfort` that isn't merged yet) into the common library class.

Modifies the low-level `Screen:show()` function to handle `onAboutToShow()` dismissing a window. This allows a `MenuOverlay` subclass to reinvoke itself (or get run twice by the user) without having to manually track whether it already has a window shown so it can dismiss it before showing the new one. `gui/blueprint` does this, and now all that code can be removed.

There are many subclasses of MenuOverlay that can take advantage of the new code. I'll migrate them after this change is approved and merged.

### DIFF
--- a/library/lua/gui.lua
+++ b/library/lua/gui.lua
@@ -537,8 +537,11 @@ function Screen:show(parent)
     if self._native then
         error("This screen is already on display")
     end
+    self:onAboutToShow(parent or dfhack.gui.getCurViewscreen(true))
+
+    -- if we're autodetecting the parent, refresh the parent handle in case the
+    -- original parent has been dismissed by onAboutToShow()'s actions
     parent = parent or dfhack.gui.getCurViewscreen(true)
-    self:onAboutToShow(parent)
     if not dscreen.show(self, parent.child) then
         error('Could not show screen')
     end
--- a/library/lua/gui/dwarfmode.lua
+++ b/library/lua/gui/dwarfmode.lua
@@ -426,8 +426,6 @@ function DwarfOverlay:simulateCursorMovement(keys, anchor)
 end
 
 function DwarfOverlay:onAboutToShow(parent)
-    DwarfOverlay.super.onAboutToShow(self, parent)
-
     if not df.viewscreen_dwarfmodest:is_instance(parent) then
         error("This screen requires the main dwarfmode view")
     end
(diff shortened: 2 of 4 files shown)


library/lua/gui/dwarfmode.lua
```diff
@@ -426,6 +426,8 @@ function DwarfOverlay:simulateCursorMovement(keys, anchor)
 end
 
 function DwarfOverlay:onAboutToShow(parent)
+    DwarfOverlay.super.onAboutToShow(self, parent)
```
myk002, line 429: this currently calls an empty function, but something *could* be there in the future.

library/lua/gui/dwarfmode.lua
```diff
@@ -436,21 +438,51 @@ MenuOverlay = defclass(MenuOverlay, DwarfOverlay)
...
-    MenuOverlay.super.onAboutToShow(self,below)
+function MenuOverlay:onAboutToShow(parent)
+    if not dfhack.isMapLoaded() then
+        qerror('Please load a fortress map.')
+    end
+
+    self.saved_sidebar_mode = df.global.ui.main.mode
+    -- if we can't get back to the saved sidebar mode via enterSidebarMode
+    -- then fall back to the default mode
+    if dfhack.gui.getCurFocus(true):find('^dfhack/')
```
softmoth, line 460: Looking at `enterSidebarMode()`, it doesn't appear that it cares about `^dfhack/` as long as it can Escape out of the mode a few times and get back to `dwarfmode/Default`. Is there a particular case you're handling here? Should it be documented, or perhaps this test should be factored out into a `canEnterSidebar()` test function?

softmoth (commented):
Looks good to me. I have a few small questions but they're not essential.

library/lua/gui/dwarfmode.lua
myk002: I'll add a comment. the case we're handling here is that if a `dfhack` window is on the screen, then we can't trust that the current sidebar mode was set by the user. it could be a previous `MenuOverlay` subclass that is currently being shown, and it has set the sidebar mode for its own purposes.

library/lua/gui.lua
```diff
@@ -537,8 +537,11 @@ function Screen:show(parent)
     if self._native then
         error("This screen is already on display")
     end
+    self:onAboutToShow(parent or dfhack.gui.getCurViewscreen(true))
```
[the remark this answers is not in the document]
myk002: I had originally written it so that `onAboutToShow()` can optionally return a new parent, but I saw that as too big of a risk. It would be too easy for `onAboutToShow` overrides to forget to do this, or worse, have a superclass method that changes the parent, and the subclass override that calls the superclass implementation but forgets to propagate the new parent. In the end, I figured that if a client knows what they're doing and passes a `parent` to `onShow`, then use that. otherwise autocorrect whenever possible.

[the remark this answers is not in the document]
myk002: I can document this in Lua API.rst

softmoth (approved):

library/lua/gui/dwarfmode.lua
```diff
@@ -436,21 +438,61 @@ MenuOverlay = defclass(MenuOverlay, DwarfOverlay)
...
 
 function MenuOverlay:computeFrame(parent_rect)
     return self.df_layout.menu, gui.inset_frame(self.df_layout.menu, self.frame_inset)
 end
 
-function MenuOverlay:onAboutToShow(below)
-    MenuOverlay.super.onAboutToShow(self,below)
+function MenuOverlay:onAboutToShow(parent)
+    if not dfhack.isMapLoaded() then
+        qerror('Please load a fortress map.')
```
lethosor, line 454: Since this is a library function that devs are calling, this line ought to use `error()` to produce a backtrace, or the check should be removed and left to the calling script. (If it's too dangerous to proceed, we should use `error()`.)

library/lua/gui/dwarfmode.lua
```diff
@@ -436,21 +438,61 @@ MenuOverlay = defclass(MenuOverlay, DwarfOverlay)
...
+    --    sidebar mode for its own purposes.
+    if not SIDEBAR_MODE_KEYS[self.saved_sidebar_mode]
+            or dfhack.gui.getCurFocus(true):find('^dfhack/') then
+        self.saved_sidebar_mode = df.ui_sidebar_mode.Default
+    end
+    enterSidebarMode(df.ui_sidebar_mode.Default)
+
+    -- refresh parent since the original parent may have just been dismissed
+    parent = dfhack.gui.getCurViewscreen(true)
+    MenuOverlay.super.onAboutToShow(self, parent)
```
lethosor, line 475: Your comment above says
```
    -- if sidebar_mode is set, we will enter the specified sidebar mode on show
    -- and restore the previous sidebar mode on dismiss. otherwise it is up to
    -- the caller to ensure we are in a sidebar mode where the menu is visible
```
but this is unconditionally modifying `ui.main.mode`, potentially changing behavior unexpectedly for clients (not 100% sure if it is a "breaking" change). Could this entire block be moved under the check for `self.sidebar_mode`? I think clients could still opt-in by setting `sidebar_mode = Default` where the subclass of MenuOverlay is defined.

library/lua/gui/dwarfmode.lua
myk002: Since the sidebar is not relevant in non-fort modes, I'd go with error here

library/lua/gui/dwarfmode.lua
myk002: You're absolutely right. I'll solidify the logic on this before I merge.

myk002: done

PR #2084, opened 2022-04-08
ensure ListBoxes always have enough height

myk002:
Currently, if a filtered ListBox has a filter that doesn't match any items, it will shrink the dialog box so small that users won't be able to see the message that says that nothing matches the filter.

This PR modifies the minimum dialog box height to take the error message into account.

There appeared to be a typo in the call to `super`. The wrong class name was used. This apparently gave the same result as the correct class name, so that wasn't much of an issue. I fixed it anyway, though.

### DIFF
--- a/library/lua/gui/dialogs.lua
+++ b/library/lua/gui/dialogs.lua
@@ -215,7 +215,7 @@ function ListBox:getWantedFrameSize()
     local mw, mh = ListBox.super.getWantedFrameSize(self)
     local list = self.subviews.list
     list.frame.t = mh+1
-    return math.max(mw, list:getContentWidth()), mh+3+math.min(20,list:getContentHeight())
+    return math.max(mw, list:getContentWidth()), mh+3+math.min(18,list:getContentHeight())
 end
 
 function ListBox:onInput(keys)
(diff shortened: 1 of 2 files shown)


library/lua/gui/dialogs.lua
```diff
@@ -212,10 +212,10 @@ function ListBox:onRenderFrame(dc,rect)
 end
 
 function ListBox:getWantedFrameSize()
-    local mw, mh = InputBox.super.getWantedFrameSize(self)
+    local mw, mh = ListBox.super.getWantedFrameSize(self)
     local list = self.subviews.list
     list.frame.t = mh+1
-    return math.max(mw, list:getContentWidth()), mh+1+math.min(20,list:getContentHeight())
+    return math.max(mw, list:getContentWidth()), mh+3+math.min(20,list:getContentHeight())
```
softmoth, line 218: Why not `mh + 3 + math.min(18, list:getContentHeight())`, or is that max size not critical?

softmoth (changes requested):

library/lua/gui/dialogs.lua
myk002: It's *probably* not critical, but there could, in theory, be a viewscreen that depends on the maximum size of the list being within a certain bounds.  I'll adjust it. Thanks!

softmoth (approved):

PR #2085, opened 2022-04-08
Handle punctuation in FilteredList filter matching

myk002:
Allows punctuation to be typed into a filter and allows the filter to match keys with punctuation.

This came up in the quickfort GUI redesign. I couldn't match labels that users would naturally type in starting with a `/`. I also couldn't match a `-n` if they copied the text from the `quickfort list` output.

This PR extends the FilteredList implementation by allowing punctuation to be used just like any other key.


lethosor (approved):
Assuming this has been tested

PR #2089, opened 2022-04-09
Document and update argparse module

myk002:
- moves existing documentation from in-file to `Lua API.rst`
- significantly extends documentation for `processArgs()` and `processArgsGetopt()`, hopefully making them more accessible to new developers
- adds unit tests for `argparse.processArgs()`
- use new string fns in `processArgs()` (e.g. `string:startswith()`) instead of using longhand code
- allow `processArgs()` to recognize options with two dashes instead of just one to make transition to `processArgsGetopt()` easier if devs want to upgrade from one to the other

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -2855,6 +2855,9 @@ parameters.
     yourscript --optName [ list of [ nested values ] [ in square brackets ] ]
     yourscript --optName \--value
 
+  Note that ``processArgs`` does not support non-option ("positional")
+  parameters. They are supported by ``processArgsGetopt`` (see below).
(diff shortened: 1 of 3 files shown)


softmoth (commented):
Looks really great! Positional parameters are documented for `processArgsGetopt`, but not really for `processArgs`. They're a reason to use processArgsGetopt, but what happens to them if you're using processArgs? Do they just get skipped and ignored? Throw an error?

myk002:
they throw an error. the fall down to the last else clause and error out, I believe

softmoth (approved):

PR #2091, opened 2022-04-11
[CMake] Include common setup to facilitate 3rd party plugins

softmoth:
This makes no functional changes. The diff is purely moving some text from CMakeLists.txt to CMake/Common.cmake, and including that file.

The TWBT plugin hard-codes duplicate DFHack version info and compiler settings. This is confusing and brittle. And easy(?) to fix by putting common definitions in a separate include file that can be used by third party plugins from their own CMakeLists.txt (see https://github.com/thurin/df-twbt/pull/4).

The common portions are:
- CMAKE_GENERATOR check
- CMAKE_VERSION policy fixes
- CMAKE_CONFIGURATION_TYPE (Release/RelWithDebInfo)
- C/C++ Compiler
  - Ensure version is compatible with DF binaries
  - CXX standard
  - Build arch & cross-compiling
  - Compiler flags
- DFHack Version definition
- Installation destinations

I am a CMake novice. There may well be a better way to do this, but this does seem to work, and so it's a place to start the conversation at least.

### DIFF
--- a/CMake/Common.cmake
+++ b/CMake/Common.cmake
@@ -1,0 +1,199 @@
+
+if("${CMAKE_GENERATOR}" STREQUAL Ninja)
+    if("${CMAKE_VERSION}" VERSION_LESS 3.9)
+        message(WARNING "You are using an old version of CMake (${CMAKE_VERSION}) with Ninja. This may result in ninja errors - see docs/Compile.rst for more details. Upgrading your CMake version is recommended.")
+    endif()
+endif()
+
+if(NOT("${CMAKE_VERSION}" VERSION_LESS 3.12))
+    # make ZLIB_ROOT work in CMake >= 3.12
+    # https://cmake.org/cmake/help/git-stage/policy/CMP0074.html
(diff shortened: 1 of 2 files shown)


softmoth:
After discussing this on Discord, I think this approach isn't necessary if 3rd party plugins are assumed to be in `dfhack/my-fancy-plugin`. When in a subdirectory they already have access to the parent's cmake goodies, I'm told. I'll work on an example of that approach in df-twbt, and if it goes well then probably it's best to leave this alone.

That's not to say that the CMakeLists.txt couldn't be organized into various include files or modules, etc. But it's probably not necessary.

cppcooper:
Actually, I think that common file is a useful addition just for the flexibility. I also made my twbt PR rely on it =/

lethosor:
From our discusson on Discord, if plugins are always in `dfhack/plugins`, they will inherit the necessary variables without needing to include anything. Let's continue the discussion at https://github.com/thurin/df-twbt/pull/2 to decide what TWBT actually needs to support.

PR #2092, opened 2022-04-12
handle number-indexed lua maps in safe_index

myk002:
`safe_index` purports to make it safe to walk through a sequence of dereferences, but it fails to handle two cases:
1) the index is numeric and is the upper bound of a lua list
2) the index is numeric and is the index into a sparse lua map

the first issue occurs because `safe_index` assumes it is looking at a C++ vector when it gets numeric indices. in the case of C++, it correctly bails if index is < 0 or >= `#obj`.  however, for lua lists, this check doesn't make sense. first of all it's fine if you index to a location outside of the list (the next check will correctly return nil anyway). second, lua lists range from 1 to #, not 0 to # - 1.

The second issue occurs because to check the range of the list, it uses the `#` operator. this always returns 0 if the map is sparse and doesn't contain an element at `[1]`.

both issues can be fixed by only doing the range check if `type(obj)` is userdata.


lethosor (approved):
Seems that it doesn't really handle invalid string indices for userdata, but this is an improvement

PR #2093, opened 2022-04-13
Adding cprobe capability to cursecheck

### LINKED ISSUE #2087: Add cprobe capabilities to cursecheck
Hi,

I was using cursecheck and got the name of the cursed create, but still can't find it, so I through it would be great to have some of the cprobe capabilities on that plugin so I could get the creature ID and teleport it under my cursor.

I was checking the code and both plugins loop trough all creatures, so what I am suggesting is adding 1 more parameter on cursecheck to print that line I extracted from probe:
`out.print("Creature %d, race %d (%x), civ %d (%x)\n", unit->id, unit->race, unit->race, unit->civ_id, unit->civ_id);`

You think this would be possible? Do you see any problem with it?
If you need contribution on the coding, I am available, but would need a few directions on any info on how to contribute.

Thanks in advance.

GuilhermeAbraham:
This PR Closes #2087.

I added a new parameter `cprobe` to display part of cprobe output for the unit being evaluated by cursecheck. 

Also, added this parameter to the `help` and to the documentation.
Would appreciate feedback on those description texts.

Thanks in advance.

### DIFF
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -182,8 +182,7 @@ Options:
 
 :detail:      Print full name, date of birth, date of curse and some status
               info (some vampires might use fake identities in-game, though).
-:cprobe:      Print some information from creature probe plugin, displaying
-              the creature and race IDs.
+:ids:         Print the creature and race IDs.
 :nick:        Set the type of curse as nickname (does not always show up
               in-game, some vamps don't like nicknames).
 :all:         Include dead and passive cursed creatures (can result in a quite
--- a/docs/Plugins.rst
+++ b/docs/Plugins.rst
@@ -182,6 +182,8 @@ Options:
 
 :detail:      Print full name, date of birth, date of curse and some status
               info (some vampires might use fake identities in-game, though).
+:cprobe:      Print some information from creature probe plugin, displaying
+              the creature and race IDs.
 :nick:        Set the type of curse as nickname (does not always show up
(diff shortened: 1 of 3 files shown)


docs/Plugins.rst
```diff
@@ -182,6 +182,8 @@ Options:
 
 :detail:      Print full name, date of birth, date of curse and some status
               info (some vampires might use fake identities in-game, though).
+:cprobe:      Print some information from creature probe plugin, displaying
```
lethosor, line 185: I think we should name this `id` or `ids`, because that's more reflective what is being printed (and is somewhat more consistent with the other existing options here). `cprobe` is a tool that some output has been copied from, but people using `cursecheck` may not be familiar with it.

Other than that, this looks like it should work to me!

myk002:
Could you add a line to `docs/changelog.txt`? Probably in the [Misc Improvements](https://github.com/DFHack/dfhack/blob/develop/docs/changelog.txt#L45) section

GuilhermeAbraham:
Applied the given suggestions and added an entry to changelog.

I was a little lost if it is only on changelog, because there is the history and news file too.

Let me know if I missed anything and if you have any other suggestion. Thanks in advance.

myk002 (approved):

myk002:
I fixed up a few minor issues in the changelog entry, but this looks great! Thanks for the PR!

PR #2094, opened 2022-04-13
Adds an onUnitAction event

Putnam3145:
I've had an [onUnitAction event](https://github.com/Putnam3145/SPARKING-Dwarf-Fortress-DBZ-mod/blob/master/raw/scripts/modtools/putnam_events.lua) that I've wanted to move to C++ since, oh, 2015. I only got DFHack compiled recently, so here it is.

Tested to work with my mod that actually uses this, but that's just on windows.


library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
         //TODO: deduce attacker from latest defend event first
     }
 }
+
+static void manageActionEvent(color_ostream& out) {
+    if (!df::global::world)
+        return;
+    multimap<Plugin*,EventHandler> copy(handlers[EventType::UNIT_ACTION].begin(), handlers[EventType::UNIT_ACTION].end());
+    for ( size_t a = 0; a < df::global::world->units.all.size(); a++ ) {
+        df::unit* unit = df::global::world->units.all[a];
```
myk002, line 1270: Since you don't use the index for anything other than getting the unit out of the vector, consider
 `for (auto unit : df::global::world->units.all)`

library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
...
+        if ( Units::isActive(unit) ) {
```
myk002, line 1271: To reduce indentation, how about
```
if (!Units::isActive(unit)) {
    unitToKnownActions.erase(unit->id);
    continue;
}
```
This way you don't need an else clause

library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
...
+            auto knownActions = &unitToKnownActions[unit->id];
```
myk002, line 1272: consider using a reference instead of a pointer:
`auto & knownActions = unitToKnownActions[unit->id];`

plugins/eventful.cpp
```diff
@@ -220,6 +223,10 @@ static void ev_mng_interaction(color_ostream& out, void* ptr) {
     EventManager::InteractionData* data = (EventManager::InteractionData*)ptr;
     onInteraction(out, data->attackVerb, data->defendVerb, data->attacker, data->defender, data->attackReport, data->defendReport);
 }
+static void ev_mng_action(color_ostream& out, void* ptr) {
```
myk002, line 226: I suggest naming this `ev_mng_unitAction` to match the event type name

library/include/modules/EventManager.h
```diff
@@ -95,6 +96,12 @@ namespace DFHack {
             int32_t defendReport;
         };
 
+        struct ActionData {
+            int32_t unitId;
+            df::unit_action* action;
+            int32_t actionId;
```
myk002, line 102: If `actionId` is a field of `action`, why pass the id as a separate field to the handler?

library/modules/EventManager.cpp
```diff
@@ -200,6 +203,9 @@ static int32_t reportToRelevantUnitsTime = -1;
 //interaction
 static int32_t lastReportInteraction;
 
+//unit action
+static std::map<int32_t,std::vector<int32_t> > unitToKnownActions;
```
myk002, line 207: Since we don't iterate through this structure, an `unordered_map` might bea better choice here.

library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
...
+                        knownActions->push_back(action->id);
+                        for ( auto b = copy.begin(); b != copy.end(); b++ ) {
+                            EventHandler handle = (*b).second;
+                            ActionData data = {unit->id, action, action->id};
+                            handle.eventHandler(out, (void*)&data);
+                        }
+                    }
+                } else {
+                    auto newEnd = std::remove(knownActions->begin(), knownActions->end(), action->id);
+                    knownActions->erase(newEnd, knownActions->end());
```
myk002, line 1285: Could you explain what is happening here? Is an action type of None invalidating all more recent actions?

myk002:
Also be aware that #2044 is in flight. @cppcooper could you review to see how to merge this new event?

library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
...
+        return;
+    multimap<Plugin*,EventHandler> copy(handlers[EventType::UNIT_ACTION].begin(), handlers[EventType::UNIT_ACTION].end());
+    for ( size_t a = 0; a < df::global::world->units.all.size(); a++ ) {
+        df::unit* unit = df::global::world->units.all[a];
+        if ( Units::isActive(unit) ) {
+            auto knownActions = &unitToKnownActions[unit->id];
+            for ( df::unit_action* action : unit->actions ) {
+                if ( action->type != df::unit_action_type::None) {
+                    if ( std::find(knownActions->begin(), knownActions->end(), action->id) == knownActions->end() ) {
+                        knownActions->push_back(action->id);
```
cppcooper, line 1276: So it seems like the data structure is only being used to keep track of events already seen. `event_tracker` in #2044 will take care of this. But.. see my suggestion above in the meanwhile. Which will reduce what needs to be done in #2044 when merging these changes into it.

library/modules/EventManager.cpp
cppcooper: > Since we don't iterate through this structure, an `unordered_map` might bea better choice here.

True.

But an even better idea is to add a hash for the data we actually care about, since they are trivial to construct, and change it to `std::unordered_set<ActionData>`. This way we avoid whatever kind of search `std::find` is doing (I would presume linear, since it can't know if it's sorted.. but maybe that's a requirement else UB). It would also remove at least one indent level.

cppcooper (commented):
> Also be aware that https://github.com/DFHack/dfhack/pull/2044 is in flight. @cppcooper could you review to see how to merge this new event?

So I've reviewed this now.

myk002:
Could you merge in latest develop and add the event functions to the new arrays added by #2097 ?

library/modules/EventManager.cpp
Putnam3145: It's the [erase-removed idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom), though perhaps more verbose than it needs to be

Putnam3145:
Yeah, I was just gonna wait until the rework stuff is merged then adapt, I'll do all the stuff now.

library/modules/EventManager.cpp
Putnam3145: Nah, I generally prefer things be structured in the way mentioned for stuff like this, I just sort of forget to do it

library/modules/EventManager.cpp
```diff
@@ -1254,3 +1261,32 @@ static void manageInteractionEvent(color_ostream& out) {
...
+                        for ( auto b = copy.begin(); b != copy.end(); b++ ) {
+                            EventHandler handle = (*b).second;
+                            ActionData data = {unit->id, action, action->id};
+                            handle.eventHandler(out, (void*)&data);
```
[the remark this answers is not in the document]
cppcooper: The only one that need apply to this PR would be 1. With 2, it would just introduce an unused variable.

library/include/modules/EventManager.h
Putnam3145: Midunderstanding of the comment `//it has to keep the id of an item because the item itself may have been deallocated`, most likely; I didn't quite put together that this is for inventory_item in particular

library/modules/EventManager.cpp
myk002: I see. So it's only removing a single element. Does order matter for these actions, or would we be better off using a set instead of a vector?

library/modules/EventManager.cpp
```diff
@@ -235,6 +239,9 @@ static int32_t reportToRelevantUnitsTime = -1;
 //interaction
 static int32_t lastReportInteraction;
 
+//unit action
+static std::unordered_map<int32_t,std::vector<int32_t> > unitToKnownActions;
```
myk002, line 243: It looks like this should be a `std::unordered_map<int32_t,std::unordered_set<int32_t>>` since there doesn't seem to be a reason to pay the cost of iterating through the vector on every lookup, or am I missing something?

myk002 (changes requested):
Apart from the one question on the data structure, looks great! Could you merge latest develop one more time and do a quick verification with `devel/eventful-client` that the events come out the way you expect?

library/modules/EventManager.cpp
cppcooper: Your map<set> is a fine approach, but I still think using the action event data alone would be best. `std::unordered_set<ActionData>` with a hash function that hashes on the unit id and the action id. Maybe I'm missing something, but that looks like it'll give the same guarantees with the fastest lookup times

myk002: Agreed. That would be cleaner.

myk002: @Putnam3145 are you ok with this change?

myk002 (changes requested):

Putnam3145:
Any and all, yes, sorry, uhh, I get distracted by things and I maintain an active-enough open source repository that github notifications are completely useless

ab9rf:
is this still viable?

ab9rf:
This has been abandoned for three years, there is not a pressing demand for it, and it will need a substantial rework to adapt to subsequent changes to the code. As such, I'm closing it unmerged.

PR #2095, opened 2022-04-14
Adds plugins/external/ and auto-populates the cmake within upon creation

cppcooper:
A simple sub-directory to add auto-ignored plugins


lethosor:
pre-commit.ci autofix

(testing)

myk002:
Won't we also need an add_subdirectory(external) in the plugins CMakeLists.txt file?

lethosor:
Unfortunately it looks like we ended up doing this simultaneusly.

My implementation at #2096 ends up creating the CMakeLists.txt automatically, since it's ignored. Otherwise, the only difference is where we decided to put the gitignore rule. I went for the top-level CMakeLists.txt for consistency with how the old system was set up, but your approach is valid too.

cppcooper:
> Won't we also need an add_subdirectory(external) in the plugins CMakeLists.txt file?

Yes, I'm working on it now.

```cmake
# build the plugins
message("before")
if(BUILD_PLUGINS)
    add_subdirectory(plugins)
    macro(SUBDIRLIST result curdir)
        file(GLOB children ABSOLUTE ${curdir}/ ${curdir}/*/)
        message("here")
        message("${children}")
        set(dirlist "")
        foreach(child ${children})
            if(IS_DIRECTORY ${curdir}/${child})
                list(APPEND dirlist ${child})
            endif()
        endforeach()
        SET(${result} ${dirlist})
    endmacro()

    SUBDIRLIST(SUBDIRS plugins/external)

    message("foreach")
    foreach(subdir ${SUBDIRS})
        message("plugins/external/${subdir}")
        add_subdirectory()
    endforeach()
endif()
message("after")
```

nearly figured it out. I would have `add_subdirectory(plugins/external)` but that would have required a cmake file for it, which then people would need to change and so it would need to be untracked

cppcooper:
> ends up creating the CMakeLists.txt automatically, since it's ignored

ahh, much better

lethosor:
The loop approach is a cool idea, but less flexible, because it doesn't allow devs to disable plugins that they don't want to build. Having to add (or comment out) an `add_subdirectory()` call is manual, but should generally be pretty quick (compared to deleting/recreating/moving a plugin's folder around to enable/disable building it).

cppcooper:
> The loop approach is a cool idea, but less flexible, because it doesn't allow devs to disable plugins that they don't want to build.

Fair point. I pushed it nonetheless. Though I think this only triggers in a situation where they're building all the plugins anyway.. so maybe it doesn't matter?

myk002:
I merged #2096 as a baseline so we could consider this PR relative to those changes. Could you merge?

cppcooper:
I actually already merged his branch in, I'm working out the generation now. I technically got it working.. but it was with absolute paths and now I'm trying to fix that.

cppcooper:
So it works now, I tested earlier by generating with twbt inside

plugins/CMakeLists.txt
```diff
@@ -189,6 +189,17 @@ if(BUILD_SKELETON)
...
+macro(SUBDIRLIST result subdir)
+    file(GLOB children ABSOLUTE ${subdir}/ ${subdir}/*/)
+    set(dirlist "")
+    foreach(child ${children})
+        if(IS_DIRECTORY ${child})
+            file(RELATIVE_PATH child ${CMAKE_CURRENT_SOURCE_DIR}/${subdir} ${child})
+            list(APPEND dirlist ${child})
+        endif()
+    endforeach()
+    SET(${result} ${dirlist})
```
lethosor, line 201: our convention is lowercase for function/macro calls
```suggestion
    set(${result} ${dirlist})
```

plugins/CMakeLists.txt
```diff
@@ -204,17 +215,22 @@ endif()
...
-    file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/external/CMakeLists.txt"
+    set(content_str
 "# Add external plugins here - this file is ignored by git
 
 # Recommended: use add_subdirectory() for folders that you have created within
 # this folder, or dfhack_plugin() for single files that you have added here.
 
 # See the end of /plugins/CMakeLists.txt for more details.
 ")
+    SUBDIRLIST(SUBDIRS external)
```
lethosor, line 226:
```suggestion
    subdirlist(SUBDIRS external)
```
(probably requires adjusting the macro definition too)

plugins/CMakeLists.txt
```diff
@@ -204,17 +215,22 @@ endif()
...
+    foreach(subdir ${SUBDIRS})
+        set(content_str "${content_str}\nadd_subdirectory(${subdir})")
```
lethosor, line 228: Maybe `\n` at the end instead?

cppcooper: This and the others are done

myk002 (approved):

PR #2097, opened 2022-04-14
EventManager/Eventful - Generate event arrays

cppcooper:
A PR for just the auto-generated arrays part so we can avoid manual synchronization and sneaky problems arising.

### DIFF
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -194,30 +194,6 @@ std::array<eventManager_t,EventType::EVENT_MAX> compileManagerArray() {
     return managers;
 }
 
-// declaration moved to inside manageEvents
-//static const eventManager_t eventManager[] = {
-//        manageTickEvent,
-//        manageJobInitiatedEvent,
-//        manageJobStartedEvent,
-//        manageJobCompletedEvent,
-//        manageUnitNewActiveEvent,
-//        manageUnitDeathEvent,
-//        manageItemCreationEvent,
-//        manageBuildingEvent,
-//        manageBuildingCreatedEvent,
-//        manageBuildingDestroyedEvent,
-//        manageConstructionEvent,
-//        manageConstructionAddedEvent,
-//        manageConstructionRemovedEvent,
-//        manageSyndromeEvent,
-//        manageInvasionEvent,
-//        manageInventoryChangeEvent,
-//        manageReportEvent,
-//        manageUnitAttackEvent,
-//        manageUnloadEvent,
-//        manageInteractionEvent,
-//};
-
 //job initiated
 static int32_t lastJobId = -1;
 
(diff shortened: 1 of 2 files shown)


myk002:
legitimate build error on windows:
`library/modules/EventManager.cpp(186): error C2027: use of undefined type 'std::array<eventManager_t,16>'`

cppcooper:
Oops. I forgot the include

library/modules/EventManager.cpp
```diff
@@ -140,24 +141,82 @@ static void manageInteractionEvent(color_ostream& out);
...
+    std::array<eventManager_t, EventType::EVENT_MAX> managers{};
+    auto t = (EventType::EventType) 0;
+    while (t < EventType::EVENT_MAX) {
+        managers[t] = getManager(t);
+        t = (EventType::EventType) int(t + 1);
+    }
+    return managers;
+}
+
+// declaration moved to inside manageEvents
```
myk002, line 197: No need to keep this commented. We can just remove it

myk002:
Overall, looks good. Have you sanity checked that appropriate events get triggered with `devel/eventful-client`?

cppcooper:
> appropriate events get triggered with `devel/eventful-client`

I checked a `JOB_INITIATED` event :shrug: and that's what it said in the console. But also I'm not sure how it would trigger the wrong events based on how the array is built. It iterates the enum, then uses a switch to pick the function corresponding to that enum value so I'd have to type the wrong function in the switch for the wrong events to be triggered.

myk002:
> so I'd have to type the wrong function in the switch for the wrong events to be triggered.

A good rule to program by: code that isn't tested is broken. Just because it looks obviously right doesn't mean that it is. Always challenge your assumptions. No one person can understand everything in a complex system, and DF (and DFHack) is a complex system. Corner cases can surprise you.

myk002 (approved):

myk002:
I'm not sure if you have a usual process for this already, but I've found that the best way to continue development after splitting off some code like this is to create a diff patch between your dev branches and then apply the diff to a new branch off develop once the split-off code is merged. E.g:
```
git diff fix-2301 generate-event-arrays >/tmp/em.diff
git checkout develop; git fetch upstream; git merge upstream/develop
git checkout -b new-EventManager
git apply /tmp/em.diff
```

cppcooper:
> I'm not sure if you have a usual process for this already, but I've found that the best way to continue development after splitting off some code like this is to create a diff patch between your dev branches and then apply the diff to a new branch off develop once the split-off code is merged. E.g:
> 
> ```
> git diff fix-2301 generate-event-arrays >/tmp/em.diff
> git checkout develop; git fetch upstream; git merge upstream/develop
> git checkout -b new-EventManager
> git apply /tmp/em.diff
> ```

What's wrong with merge? or rebase

myk002:
If that works, great. I find it gives me a lot of merge conflicts when I do it that way. Creating a diff and repatching allows you to only care about what the text looks like and not have to care about the commit chain that got it there.

Issue #2098, opened 2022-04-14
Unusable bool_distribution?


cppcooper:
https://github.com/DFHack/dfhack/blob/5f3d5bbcd5f56f238457ec4d6b5398af8243a631/plugins/lua/cxxrandom.lua#L148-L163

I was just reading an old PR, for some other reason, when I came across these lines that I wrote. They look like bool_distribution from the lua interface will be totally unusable. Needs to be tested, and very likely fixed.

PR #2099, opened 2022-04-15
Adds cxxrandom unit test and fixes interface problems

cppcooper:
Fixes a mistake in the `bool_distribution` and adds a unit test

Resolves https://github.com/DFHack/dfhack/issues/2098

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -4215,7 +4215,7 @@ Native functions (exported to Lua)
 
   adds a number to the sequence
 
-- ``ShuffleSequence(rngID, seqID)``
+- ``ShuffleSequence(seqID, rngID)``
 
   shuffles the number sequence
 
@@ -4278,7 +4278,7 @@ Lua plugin classes
 ``bool_distribution``
 ~~~~~~~~~~~~~~~~~~~~~
 
-- ``init(min, max)``: constructor
+- ``init(chance)``: constructor
 - ``next(id)``: returns next boolean in the distribution
 
   - ``id``: engine ID to pass to native function
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -4291,6 +4291,41 @@ Lua plugin classes
 - ``shuffle()``: shuffles the sequence of numbers
 - ``next()``: returns next number in the sequence
 
+Usage
+-----
+
+The basic idea is you create a number distribution which you generate random numbers along. The C++ relies
+on engines keeping state information to determine the next number along the distribution.
+You're welcome to try and (ab)use this knowledge for your RNG purposes.
+
+Example::
+
+    local rng = require('plugins.cxxrandom')
+    local norm_dist = rng.normal_distribution(6820,116) // avg, stddev
+    local engID = rng.MakeNewEngine(0)
+    -- somewhat reminiscent of the C++ syntax
+    print(norm_dist:next(engID))
+
+    -- a bit more streamlined
+    local cleanup = true --delete engine on cleanup
+    local number_generator = rng.crng:new(engID, cleanup, norm_dist)
+    print(number_generator:next())
+
+    -- simplified
+    print(rng.rollNormal(engID,6820,116))
+
+The number sequences are much simpler. They're intended for where you need to randomly generate an index, perhaps in a loop for an array. You technically don't need an engine to use it, if you don't mind never shuffling.
+
+Example::
+
+    local rng = require('plugins.cxxrandom')
+    local g = rng.crng:new(rng.MakeNewEngine(0), true, rng.num_sequence:new(0,table_size))
+    g:shuffle()
+    for _ = 1, table_size do
+        func(array[g:next()])
+    end
+
+
 dig-now
 =======
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,7 +39,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Removed
 
 ## Fixes
-- `cxxrandom`: fixed problems with interface and updated docs
+- `cxxrandom`: fixed exception when calling ``bool_distribution``
+- `cxxrandom`: fixed id order for ShuffleSequence, but adds code to detect which parameter is which so each id is used correctly. 16000 limit before things get weird (previous was 16 bits)
 - `autofarm` removed restriction on only planting 'discovered' plants
 - `luasocket` (and others): return correct status code when closing socket connections
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,6 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Removed
 
 ## Fixes
+- `cxxrandom`: fixed problems with interface and updated docs
 - `autofarm` removed restriction on only planting 'discovered' plants
 - `luasocket` (and others): return correct status code when closing socket connections
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -61,6 +61,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `tweak` stable-cursor: Keep the cursor stable even when the viewport moves a small amount
 
 ## Documentation
+- `cxxrandom`: added usage examples
 - Add more examples to the plugin skeleton files so they are more informative for a newbie
 - Lua API.rst added: ``isHidden(unit)``, ``isFortControlled(unit)``, ``getOuterContainerRef(unit)``, ``getOuterContainerRef(item)``
 - Update download link and installation instructions for Visual C++ 2015 build tools on Windows
--- a/plugins/cxxrandom.cpp
+++ b/plugins/cxxrandom.cpp
@@ -50,7 +50,7 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
     return CR_OK;
 }
 
-#define EK_ID_BASE (1l << 40)
+#define EK_ID_BASE (1ll << 40)
 
 class EnginesKeeper
 {
@@ -64,9 +64,11 @@ public:
         return instance;
     }
     uint64_t NewEngine( uint64_t seed ) {
+        auto id = ++id_counter;
+        CHECK_INVALID_ARGUMENT(m_engines.count(id) == 0);
         std::mt19937_64 engine( seed != 0 ? seed : std::chrono::system_clock::now().time_since_epoch().count() );
-        m_engines[++id_counter] = engine;
-        return id_counter;
+        m_engines[id] = engine;
+        return id;
     }
     void DestroyEngine( uint64_t id ) {
         m_engines.erase( id );
@@ -151,7 +153,7 @@ public:
 class SequenceKeeper
 {
 private:
-    SequenceKeeper() {}
+    SequenceKeeper() = default;
     std::unordered_map<uint64_t, NumberSequence> m_sequences;
     uint64_t id_counter = SK_ID_BASE;
 public:
@@ -160,12 +162,16 @@ public:
         return instance;
     }
     uint64_t MakeNumSequence( int64_t start, int64_t end ) {
-        m_sequences[++id_counter] = NumberSequence(start, end);
-        return id_counter;
+        auto id = ++id_counter;
+        CHECK_INVALID_ARGUMENT(m_sequences.count(id) == 0);
+        m_sequences[id] = NumberSequence(start, end);
+        return id;
     }
     uint64_t MakeNumSequence() {
-        m_sequences[++id_counter] = NumberSequence();
-        return id_counter;
+        auto id = ++id_counter;
+        CHECK_INVALID_ARGUMENT(m_sequences.count(id) == 0);
+        m_sequences[id] = NumberSequence();
+        return id;
     }
     void DestroySequence( uint64_t seqID ) {
         m_sequences.erase(seqID);
--- a/plugins/cxxrandom.cpp
+++ b/plugins/cxxrandom.cpp
@@ -50,67 +50,67 @@ DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_chan
     return CR_OK;
 }
 
-#define EK_ID_BASE 16000
+#define EK_ID_BASE (1l << 40)
 
 class EnginesKeeper
 {
 private:
-    EnginesKeeper() {}
-    std::unordered_map<uint16_t, std::mt19937_64> m_engines;
-    uint16_t id_counter = EK_ID_BASE;
+    EnginesKeeper() = default;
+    std::unordered_map<uint64_t, std::mt19937_64> m_engines;
+    uint64_t id_counter = EK_ID_BASE;
 public:
     static EnginesKeeper& Instance() {
         static EnginesKeeper instance;
         return instance;
     }
-    uint16_t NewEngine( uint64_t seed ) {
+    uint64_t NewEngine( uint64_t seed ) {
         std::mt19937_64 engine( seed != 0 ? seed : std::chrono::system_clock::now().time_since_epoch().count() );
         m_engines[++id_counter] = engine;
         return id_counter;
     }
-    void DestroyEngine( uint16_t id ) {
+    void DestroyEngine( uint64_t id ) {
         m_engines.erase( id );
     }
-    void NewSeed( uint16_t id, uint64_t seed ) {
+    void NewSeed( uint64_t id, uint64_t seed ) {
         CHECK_INVALID_ARGUMENT( m_engines.find( id ) != m_engines.end() );
         m_engines[id].seed( seed != 0 ? seed : std::chrono::system_clock::now().time_since_epoch().count() );
     }
-    std::mt19937_64& RNG( uint16_t id ) {
+    std::mt19937_64& RNG( uint64_t id ) {
         CHECK_INVALID_ARGUMENT( m_engines.find( id ) != m_engines.end() );
         return m_engines[id];
     }
 };
 
 
-uint16_t GenerateEngine( uint64_t seed ) {
+uint64_t GenerateEngine( uint64_t seed ) {
     return EnginesKeeper::Instance().NewEngine( seed );
 }
 
-void DestroyEngine( uint16_t id ) {
+void DestroyEngine( uint64_t id ) {
     EnginesKeeper::Instance().DestroyEngine( id );
 }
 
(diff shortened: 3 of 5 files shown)


docs/Lua API.rst
```diff
@@ -4215,7 +4215,7 @@ Native functions (exported to Lua)
 
   adds a number to the sequence
 
-- ``ShuffleSequence(rngID, seqID)``
+- ``ShuffleSequence(seqID, rngID)``
```
myk002, line 4218: since this is an established API, we should avoid changing it, even if the new parameter order would look more consistent. There is no need to break existing clients. Changing APIs is the kind of thing that frustrates contributors, and should not be done lightly.

On the other hand, if you could make either order of parameters work (so as not to break existing clients), we could change the "preferred" ordering here without problems.

docs/changelog.txt
```diff
@@ -39,6 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Removed
 
 ## Fixes
+- `cxxrandom`: fixed problems with interface and updated docs
```
myk002, line 42: since this change is specific to `bool_distribution`, it would be more useful to readers if we name it directly here. also, the changelog entry should describe exactly what was fixed. For example:
```
- `cxxrandom`: fixed exception when calling ``bool_distribution``
```
and unless the PR contains significant additions to documentation (like entire new sections or examples), we don't need to mention doc updates here

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
+local rng = require('plugins.cxxrandom')
+
+function test.cxxrandom_distributions()
+    rng.normal_distribution:new(0,5)
```
myk002, line 4: I don't see this invocation style documented in the `cxxrandom` docs. Does this return an allocated object or an engine id? Can we test for the return value to be non-null? Does the returned object need to be destroyed/freed?

an end-to-end example of usage like this would be useful to have in the docs.

docs/Lua API.rst
cppcooper: > On the other hand, if you could make either order of parameters work (so as not to break existing clients), we could change the "preferred" ordering here without problems.

done

test/plugins/cxxrandom.lua
cppcooper: it's a lua "object"
```lua
normal_distribution = {}
function normal_distribution:new(avg, stddev)
    local o = {}
    self.__index = self
    if type(avg) ~= 'number' or type(stddev) ~= 'number' then
        error("Invalid arguments in normal_distribution construction. Average and standard deviation must be numbers.")
    end
    o.average = avg
    o.std_deviation = stddev
    setmetatable(o,self)
    return o
end
```

cppcooper:
Resolves #2098

docs/Lua API.rst
cppcooper: For clarity, I meant it will try to detect which type of id each argument is and swap them if it thinks they're in reverse. The downside: things get weird [buggy] before filling in all 16bits of the id space(s)

docs/changelog.txt
cppcooper: So I guess I should probably add something about the usage examples I added.

myk002: 👍

myk002:
> Resolves #2098

This has to go in the description text in order for the referenced bug to get autoclosed

cppcooper:
> > Resolves #2098
> 
> This has to go in the description text in order for the referenced bug to get autoclosed

Oh oops. And I just closed the issue thinking this was merged. heh

cppcooper:
I added the new example in docs to the changelog, if there's anything else lmk

plugins/lua/cxxrandom.lua
```diff
@@ -151,7 +151,7 @@ bool_distribution = {}
 function bool_distribution:new(chance)
     local o = {}
     self.__index = self
-    if type(min) ~= 'number' or type(max) ~= 'number' then
+    if type(chance) ~= 'number' then
         error("Invalid arguments in bool_distribution construction. min and max must be numbers.")
```
cppcooper, line 155: Update the message.

plugins/cxxrandom.cpp
```diff
@@ -137,118 +123,100 @@ class NumberSequence
...
+    void AddToSequence(uint16_t seqID, int64_t num ) {
+        CHECK_INVALID_ARGUMENT(m_sequences.find(seqID) != m_sequences.end());
+        m_sequences[seqID].Add(num);
     }
-    void Shuffle( uint16_t id, uint16_t rng_id )
-    {
-        CHECK_INVALID_ARGUMENT( m_sequences.find( id ) != m_sequences.end() );
-        m_sequences[id].Shuffle( rng_id );
+    void Shuffle(uint16_t seqID, uint16_t engID ) {
+        uint16_t sid = seqID >= SK_ID_BASE ? seqID : engID;
```
myk002, line 178: This is fairly brittle. Since the ids are opaque to Lua, how about making them different types? Then you can figure out which is which by resolving the type sent to you from Lua.

plugins/lua/cxxrandom.lua
myk002: Yeah, was just about to comment about that

plugins/cxxrandom.cpp
```diff
@@ -137,118 +123,100 @@ class NumberSequence
...
+    static SequenceKeeper& Instance() {
         static SequenceKeeper instance;
         return instance;
     }
-    uint16_t MakeNumSequence( int64_t start, int64_t end )
-    {
-        m_sequences[++counter] = NumberSequence( start, end );
-        return counter;
+    uint16_t MakeNumSequence( int64_t start, int64_t end ) {
+        m_sequences[++id_counter] = NumberSequence(start, end);
```
myk002, line 163: Shouldn't there be some collision detection here? Ids are only 16 bits after all.  Also for NewEngine above

plugins/lua/cxxrandom.lua
```diff
@@ -208,7 +208,7 @@ function num_sequence:shuffle()
     if self.rngID == 'nil' then
         error("Add num_sequence object to crng as distribution, before attempting to shuffle.")
     end
-    ShuffleSequence(self.rngID, self.seqID)
+    ShuffleSequence(self.seqID, self.rngID)
```
myk002, line 211: Why do these classes not use the DFHack class conventions? E.g. `defclass`

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+function test.cxxrandom_seed()
+    local nd = rng.normal_distribution:new(0,500000)
+    local e1 = rng.MakeNewEngine(1)
+    local e2 = rng.MakeNewEngine(1)
+    local e3 = rng.MakeNewEngine(2)
+    local g1 = rng.crng:new(e1, true, nd)
+    local g2 = rng.crng:new(e2, true, nd)
+    local g3 = rng.crng:new(e3, true, nd)
+    expect.eq(g1:next(), g2:next())
+    expect.ne(g1:next(), g3:next())
```
myk002, line 21: This is not guaranteed to pass, unfortunately. I don't think you can test non-equality here

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+end
+
+function test.cxxrandom_ranges()
+    local e1 = rng.MakeNewEngine(1)
+    local g1 = rng.crng:new(e1, true, rng.normal_distribution:new(0,1))
+    local g2 = rng.crng:new(e1, true, rng.real_distribution:new(-5,5))
+    local g3 = rng.crng:new(e1, true, rng.int_distribution:new(-5,5))
+    local g4 = rng.crng:new(e1, true, rng.num_sequence:new(-5,5))
+    for i = 1, 10000 do
```
myk002, line 30: I think it's fine to initialize the engine with a nil seed and don't loop here. Unit tests should not be looping.

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+        local a = g1:next()
+        local b = g2:next()
+        local c = g3:next()
+        local d = g4:next()
+        expect.ge(a, -5)
```
myk002, line 35: Is there a deterministic limit to how high/low the normal distribution can get?

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+        expect.ge(c, -5)
+        expect.ge(d, -5)
+        expect.le(a, 5)
+        expect.le(b, 5)
+        expect.le(c, 5)
+        expect.le(d, 5)
+    end
+    local gb = rng.crng:new(e1, true, rng.bool_distribution:new(0.00000000001))
+    for i = 1, 10 do
+        expect.false_(gb:next())
```
myk002, line 46: We can test that the answer is either == true or == false (i.e. that it is a boolean), but we cannot guarantee the exact value.

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+    end
+end
+
+function test.cxxrandom_exports()
+    local id = rng.GenerateEngine(0)
+    rng.NewSeed(id, 2022)
+    expect.ge(rng.rollInt(id, 0, 1000), 0)
+    expect.ge(rng.rollDouble(id, 0, 1), 0)
+    expect.ge(rng.rollNormal(id, 5, 1), 0)
```
myk002, line 55: Again, I'm not sure what the guarantees are around normal distributions. Is this guaranteed?

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+    expect.true_(rng.rollBool(id, 0.9999999999))
```
myk002, line 56: This one is definitely not guaranteed.

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,65 @@
...
+    local sid = rng.MakeNumSequence(0,8)
+    rng.AddToSequence(sid, 9)
+    rng.ShuffleSequence(sid, id)
+    for i = 1, 10 do
+        expect.ge(rng.NextInSequence(sid), 0)
```
myk002, line 61: You can also verify that they're less than 10

myk002 (changes requested):

plugins/cxxrandom.cpp
cppcooper: The types are sent to lua, that's the only way lua users can get an id. I could use different integer types, but I don't see how lua would retain such information. I think I'd be better off just uptyping to 32 or 64 bits and increasing the bins for the two id types. If it is a problem that is, as is they have acess to several thousand ids which granted isn't a tonne of they just leave df open for a few months and were to regularly use RNG

plugins/cxxrandom.cpp
cppcooper: I don't see it as a big concern. The ids would overlap the other's range before they'd collide with themselves. In either case though, I don't see much that can be done without some overboard complexity for the problem.

plugins/lua/cxxrandom.lua
cppcooper: Dunno what that is.

test/plugins/cxxrandom.lua
cppcooper: It passes. The tests are all seeded.

test/plugins/cxxrandom.lua
cppcooper: I'd have to dig into the c11 standard to find out what they say about such things. But from a mathmatical standpoint, it'd be pretty wild to be five sigma out: https://cdn-images-1.medium.com/max/2600/1*IdGgdrY_n_9_YfkaCh-dag.png

But also, all the tests are seeded.

test/plugins/cxxrandom.lua
cppcooper: Seeded

test/plugins/cxxrandom.lua
cppcooper: All seeded lol

test/plugins/cxxrandom.lua
cppcooper: I suppose. These tests are kind of just "sanity" checks to make sure someone modifying things hasn't severely broken how everything works.

plugins/cxxrandom.cpp
myk002: I meant wrap the ids in a type that has an `_identity`. Then you can use `virtual_cast` to figure out what you have

test/plugins/cxxrandom.lua
cppcooper: What part? But again seeded, so if it passes once and all that

plugins/cxxrandom.cpp
cppcooper: Sounds like a decent plan, no idea how to implement that though.

myk002: Example: https://github.com/DFHack/dfhack/blob/8eb2831b7ed1b3062f9fde1b541da55b89bbf6ac/plugins/blueprint.cpp#L93

plugins/cxxrandom.cpp
myk002: How about just testing whether the id already exists in the map and erroring out if it's there?

plugins/lua/cxxrandom.lua
myk002: Docs: https://docs.dfhack.org/en/stable/docs/Lua%20API.html#class

Example: https://github.com/DFHack/dfhack/blob/8eb2831b7ed1b3062f9fde1b541da55b89bbf6ac/library/lua/gui.lua#L145

test/plugins/cxxrandom.lua
myk002: That's a good point. Could you add a comment saying the tests pass with the given seed, just in case the underlying implementation changes (or differs across platforms) and the tests start failing in the future?

test/plugins/cxxrandom.lua
myk002: : )

plugins/lua/cxxrandom.lua
myk002: It's why I was confused about the `:new()` style the code uses. DFHack classes are normally instantiated like `local obj = My class{namedparam=val, another=val}`

plugins/cxxrandom.cpp
cppcooper: Sure, doable. Simple safety check. It'll make more sense if I also increase the int sizes and or do the identity thingy

test/plugins/cxxrandom.lua
cppcooper: Can do. Will do.

plugins/lua/cxxrandom.lua
cppcooper: I'm like 90% sure I just looked up how to make lua objects and followed whatever SO the search led me to.

myk002: Well, no need to change it now. Future classes should use the DFHack class structure, tho.

myk002 (approved):
Looks good! I'd still like to see the collision detection implemented for completeness, even though "it's never gonna happen". Not a deal breaker, though

cppcooper:
> Looks good! I'd still like to see the collision detection implemented for completeness, even though "it's never gonna happen". Not a deal breaker, though

Well I wouldn't say never. Someone determined to make it happen could make it happen. But yea, practically speaking half of the 64 bit space is pretttty large

myk002:
Merge latest develop to fix the docs error

test/plugins/cxxrandom.lua
myk002: Oh, this is still open - yeah, this test is taking more than 200x the time of other tests to run. I think it would be best not to loop here

myk002 (changes requested):
The tests need unlooping, but otherwise looks good.

cppcooper:
> The tests need unlooping, but otherwise looks good.

The for loops?

myk002:
> > The tests need unlooping, but otherwise looks good.
> 
> The for loops? Why?

The comment I left before on line 30 of the test file seems to be hidden for some reason. Let me rewrite it here:

I think it's fine to initialize the engine with a nil seed and don't loop here. Unit tests should not be looping. They should be light and fast.

Just test once. If it ever fails it will show up eventually.

test/plugins/cxxrandom.lua
cppcooper: ah this one. yea hidden. I saw it once upon a time, then I thought it was resolved in your batch of resolving

cppcooper:
> Just test once. If it ever fails it will show up eventually.
How about like.. 10 times? instead of 10,000. As to shorten the "eventually"

myk002:
> > Just test once. If it ever fails it will show up eventually.
> 
> How about like.. 10 times? instead of 10,000. As to shorten the "eventually"

I think just once is fine, but if 10 times allows the test to pass in 1ms or less, then it's ok

myk002:
You can see how long each test takes to pass in the output for the test step in the automated builds (not the buildmaster ones)

test/plugins/cxxrandom.lua
cppcooper: reduced to 10, and it is now <1ms

test/plugins/cxxrandom.lua
```diff
@@ -0,0 +1,75 @@
...
+function test.cxxrandom_seed()
+    local nd = rng.normal_distribution:new(0,500000)
+    local e1 = rng.MakeNewEngine(1)
+    local e2 = rng.MakeNewEngine(1)
+    local e3 = rng.MakeNewEngine(2)
+    local g1 = rng.crng:new(e1, true, nd)
+    local g2 = rng.crng:new(e2, true, nd)
+    local g3 = rng.crng:new(e3, true, nd)
+    expect.eq(g1:next(), g2:next())
+    expect.ne(g1:next(), g3:next())
```
myk002, line 29: I think this would pass even if all three engines started with the same seed. This compares the *second* output of `g1` with the *first* output of `g3`. I think what is intended here is that we ensure the first output of the two engines are different. Storing the first output of `g1` in a local and using that in the comparisons would fix it, I think.

cppcooper: Oops! Nice catch. I wrote these tests perhaps a bit too quickly

myk002 (approved):

PR #2100, opened 2022-04-15
default unset `frame_inset` values to 0

TymurGubayev:
This change allows writing `frame_inset = {r=1}` instead of `frame_inset = {l=0, r=1, t=0, b=0}`

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -3689,6 +3689,8 @@ Base of all the widgets. Inherits from View and has the following attributes:
   :x: left/right margin, if ``l`` and/or ``r`` are omitted.
   :y: top/bottom margin, if ``t`` and/or ``b`` are omitted.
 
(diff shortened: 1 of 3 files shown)


myk002 (approved):
Looks good. Please add a changelog entry (in the Lua section)

myk002:
Oh, also, could you add a note that unspecified fields default to 0 in the `frame_inset` docs at https://docs.dfhack.org/en/stable/docs/Lua%20API.html#widget-class ?

TymurGubayev:
> Please add a changelog entry (in the Lua section)

did that

> Oh, also, could you add a note that unspecified fields default to 0 in the `frame_inset` docs at https://docs.dfhack.org/en/stable/docs/Lua%20API.html#widget-class ? (in `docs/Lua API.rst`)

and that.

PR #2101, opened 2022-04-15
WIP: add scroll icons to Label widget

TymurGubayev:
It's an opt-out. The icons are rendered in the right-most column of the 1st and last row. They are only rendered when text can actually be scrolled in the corresponding direction.

WIP: Currently, the icons might overlay text characters, there is no mechanism preventing it

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -391,6 +391,18 @@ function render_text(obj,dc,x0,y0,pen,dpen,disabled)
     obj.text_width = width
 end
 
+function render_scroll_icons(obj,dc)
+    if obj:getTextHeight() > obj.frame_body.height then
+        if obj.start_line_num ~= 1 then
+            dc:seek(dc.width-1, 0):string(obj.up_arrow_icon, obj.scroll_icon_pen)
+        end
+        local last_visible_line = obj.start_line_num + obj.frame_body.height - 1
+        if last_visible_line < obj:getTextHeight() then
+            dc:seek(dc.width-1, dc.height-1):string(obj.down_arrow_icon, obj.scroll_icon_pen)
+        end
+    end
+end
+
 function check_text_keys(self, keys)
     if self.text_active then
         for _,item in ipairs(self.text_active) do
@@ -415,6 +427,11 @@ Label.ATTRS{
     on_click = DEFAULT_NIL,
     on_rclick = DEFAULT_NIL,
     scroll_keys = STANDARDSCROLL,
+    --
+    show_scroll_icons = true,
+    up_arrow_icon = string.char(24),
+    down_arrow_icon = string.char(25),
+    scroll_icon_pen = COLOR_LIGHTCYAN,
 }
 
 function Label:init(args)
@@ -463,6 +480,9 @@ function Label:onRenderBody(dc)
         text_pen = self.text_hpen
     end
     render_text(self,dc,0,0,text_pen,self.text_dpen,is_disabled(self))
+    if self.show_scroll_icons then
+        render_scroll_icons(self,dc)
+    end
 end
 
 function Label:scroll(nlines)


library/lua/gui/widgets.lua
```diff
@@ -391,6 +391,18 @@ function render_text(obj,dc,x0,y0,pen,dpen,disabled)
     obj.text_width = width
 end
 
+function render_scroll_icons(obj,dc)
+    if obj:getTextHeight() > obj.frame_body.height then
+        if obj.start_line_num ~= 1 then
+            dc:seek(dc.width-1, 0):string(obj.up_arrow_icon, obj.scroll_icon_pen)
```
TymurGubayev, line 397: We want to set `frame_inset.r` to at least `1` to have an empty column, but the painter object `dc` can't draw there: after `dc:seek(dc.width, 0)` the `dc:isValidPos()` is `false`.

library/lua/gui/widgets.lua
```diff
@@ -415,6 +427,11 @@ Label.ATTRS{
     on_click = DEFAULT_NIL,
     on_rclick = DEFAULT_NIL,
     scroll_keys = STANDARDSCROLL,
+    --
+    show_scroll_icons = true,
```
myk002, line 431: maybe default this to `DEFAULT_NIL` and allow it to autodetect whether the content height exceeds the body height

myk002: and if the caller wants to force them on or off, they can set `show_scroll_icons` to a boolean

myk002:
once you have the code/behavior settled, docs for the new Label attributes go [here](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#label-class)

library/lua/gui/widgets.lua
TymurGubayev: The actual icons are only shown when there is a need for them. I.e. they won't necessarily be visible even with `show_scroll_icons = true` right now. What it will do is potentially set that `frame_inset.r` to `1`. I think forcing the arrows to be always visible will be confusing.

Maybe `DEFAULT_NIL` could set the boolean value on widget creation but before adjusting the `frame_inset.r`?

myk002: fair enough -- defaulting to true is fine if the behavior is still dynamic.

On that note, the length of the text is dynamic too, so we have to be careful about just doing things on init. I *believe* the vertical height of the text can only change when `setText()` is called since dynamic text can't add lines -- they can only change the contents of existing lines. This is because newlines returned from text-generating functions don't cause the rendered text to go to the next line. I'm on the fence on whether this is a bug or intended behavior, but it's the current implementation.

There should also be a change when you can no longer scroll in a particular direction. Like maybe changing the icon pen from COLOR_CYAN to something darker (or gray?)

TymurGubayev: >There should also be visual feedback when you can no longer scroll in a particular direction. Like maybe changing the icon pen from COLOR_CYAN to something darker (or gray?)

Currently, the change is the icon disappears. In other words, a scroll icon is visible if and only if it is possible to scroll in indicated direction.

TymurGubayev: I've made some changes.

* since `render_scroll_icons` only works with a label, it's now a class function
* `show_scroll_icons` has 4 possible values: `false` for no icons, `left` for icons on the first column on the left (also ensuring `frame_inset.l >= 1`), `right` - last column on the right, `DEFAULT_NIL` - same as `right` if text height greater than `frame_body.height`, else same as `false`.

We can't call `Label:update_scroll_inset` from `Label:setText`, because `self.frame_body` isn't calculated at the time it's called (this is only needed for the `DEFAULT_NIL` logic).

library/lua/gui/widgets.lua
```diff
@@ -433,6 +438,47 @@ function Label:setText(text)
...
+            l = 1
+        elseif r <= 0 then
+            r = 1
+        end
+        self.frame_inset = {l=l,t=t,r=r,b=b}
+    end
+end
+
+function Label:render_scroll_icons(dc,rect)
+    if self:getTextHeight() > self.frame_body.height then
```
myk002, line 465: It might be clearer if this if test were in the calling function (onRenderFrame)

myk002:
This is nearly ready for merge. @TymurGubayev do you have time to address the last few comments?

library/lua/gui/widgets.lua
TymurGubayev: did that

library/lua/gui/widgets.lua
```diff
@@ -433,6 +438,47 @@ function Label:setText(text)
...
+        end
+        local last_visible_line = self.start_line_num + self.frame_body.height - 1
+        if last_visible_line < self:getTextHeight() then
+            dc:seek(x, dc.height-1):char(self.down_arrow_icon, self.scroll_icon_pen)
+        end
+    end
+end
+
+function Label:postComputeFrame()
+    if self._text_dirty then
```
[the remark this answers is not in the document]
TymurGubayev: did that too. I forgot the first rule of optimization: don't optimize.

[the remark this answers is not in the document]
TymurGubayev: Also, the scroll doesn't work with MenuOverlay (at least the way how I tested it), because arrow keys move the cursor on the map... But that is out of scope of this pull request.

docs/Lua API.rst
```diff
@@ -3782,6 +3782,13 @@ It has the following attributes:
 :auto_width: Sets self.frame.w from the text width.
 :on_click: A callback called when the label is clicked (optional)
 :on_rclick: A callback called when the label is right-clicked (optional)
+:show_scroll_icons: Controls scroll icons' behaviour: ``false`` for no icons, ``'right'`` or ``'left'`` for
+     icons next to the text in an additional column (``frame_inset`` is adjusted to have ``.r`` or ``.l`` greater than ``0``),
+     ``nil`` same as ``'right'`` but changes ``frame_inset`` only if a scroll icon is actually necessary
+     (if ``getTextHeight()`` is greater than ``frame_body.height``). Default is ``nil``.
+:up_arrow_icon: The symbol for scroll up arrow.
+:down_arrow_icon: The symbol for scroll down arrow.
+:scroll_icon_pen: Specifies the pen for scroll icons.
```
myk002, line 3791: it would be useful to note what the defaults are for `up_arrow_icon`, `down_arrow_icon`, and `scroll_icon_pen`.

library/lua/gui/widgets.lua
```diff
@@ -413,6 +413,11 @@ Label.ATTRS{
     on_click = DEFAULT_NIL,
     on_rclick = DEFAULT_NIL,
     scroll_keys = STANDARDSCROLL,
+    --
```
myk002, line 416: the separator is fine, but maybe put it above the `scroll_keys` line to keep all the scroll params together.

myk002 (commented):
this looks beautiful and clean. thank you very much for this!

docs/Lua API.rst
TymurGubayev: did that.

library/lua/gui/widgets.lua
TymurGubayev: did that.

The parameter was also missing from the docs - I've added it.

myk002 (approved):
Awesome! And thanks for documenting `scroll_keys`.

PR #2106, opened 2022-04-19
EventManager types become hashable

cppcooper:
The event manager's event data types now have hash functions. One is even put to use, for the constructions event manager. The rest don't get put to use in this PR, as they don't have existing behaviour that can be improve by a switch to an unordered container.

~~This is also branched from PR #2105~~


myk002 (approved):
New commits look good. Once #2105 is merged I'll rebase this an merge on top

PR #2107, opened 2022-04-19
[WIP] Refactors EventManager to support sending on multiple event frequencies

cppcooper:
This adds code for saving event data for future ticks, and the updated callbacks sending the appropriate saved data.

edited: undo of formatting and force pushed, description changed accordingly

### DIFF
--- a/library/include/modules/EventManager.h
+++ b/library/include/modules/EventManager.h
@@ -90,7 +90,7 @@ namespace DFHack {
             std::shared_ptr<InventoryItem> item_old;
             std::shared_ptr<InventoryItem> item_new;
             InventoryChangeData() = default;
-            InventoryChangeData(int32_t id_in, InventoryItem* old_in, InventoryItem* new_in) :
+            InventoryChangeData(int32_t id_in, std::shared_ptr<InventoryItem> &old_in, std::shared_ptr<InventoryItem> &new_in) :
                     unitId(id_in),
                     item_old(old_in),
                     item_new(new_in) {}
--- a/library/include/modules/EventManager.h
+++ b/library/include/modules/EventManager.h
@@ -226,8 +226,12 @@ namespace std {
 }
 
 namespace df{
-    inline bool operator==(const df::construction &A, const df::construction &B){
-        return A.pos == B.pos;
+    inline bool operator==(const df::construction &A, const df::construction &B) {
+        return A.pos == B.pos &&
+               A.flags.whole == B.flags.whole &&
+               A.item_type == B.item_type &&
+               A.mat_type == B.mat_type &&
+               A.item_subtype == B.item_subtype;
     }
 }
 
--- a/library/include/modules/EventManager.h
+++ b/library/include/modules/EventManager.h
@@ -29,8 +29,12 @@ namespace DFHack {
                 UNIT_NEW_ACTIVE,
                 UNIT_DEATH,
                 ITEM_CREATED,
-                BUILDING,
-                CONSTRUCTION,
+                BUILDING_DESTROYED,
+                BUILDING_CREATED,
+                BUILDING, // todo: deprecate this event
+                CONSTRUCTION_REMOVED,
+                CONSTRUCTION_ADDED,
+                CONSTRUCTION, // todo: deprecate this event
                 SYNDROME,
                 INVASION,
                 INVENTORY_CHANGE,
@@ -46,9 +50,11 @@ namespace DFHack {
             typedef void (*callback_t)(color_ostream&, void*); //called when the event happens
             callback_t eventHandler;
             int32_t freq; //how often event is allowed to fire (in ticks) use 0 to always fire when possible
+            int32_t when = -1; //when to fire event (global tick count)
 
-            EventHandler(callback_t eventHandlerIn, int32_t freqIn): eventHandler(eventHandlerIn), freq(freqIn) {
-            }
+            EventHandler(callback_t eventHandlerIn, int32_t freqIn) :
+                    eventHandler(eventHandlerIn),
+                    freq(freqIn) {}
 
             bool operator==(const EventHandler& handle) const {
                 return eventHandler == handle.eventHandler && freq == handle.freq;
@@ -61,7 +67,9 @@ namespace DFHack {
         struct SyndromeData {
             int32_t unitId;
             int32_t syndromeIndex;
-            SyndromeData(int32_t unitId_in, int32_t syndromeIndex_in): unitId(unitId_in), syndromeIndex(syndromeIndex_in) {}
+            SyndromeData(int32_t unitId_in, int32_t syndromeIndex_in) :
+                    unitId(unitId_in),
+                    syndromeIndex(syndromeIndex_in) {}
             bool operator==(const SyndromeData &other) const {
                 return unitId == other.unitId && syndromeIndex == other.syndromeIndex;
             }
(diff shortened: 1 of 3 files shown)


plugins/eventful.cpp
```diff
@@ -112,32 +112,32 @@ DEFINE_LUA_EVENT_NH_0(onUnload);
 DEFINE_LUA_EVENT_NH_6(onInteraction, std::string, std::string, int32_t, int32_t, int32_t, int32_t);
 
 DFHACK_PLUGIN_LUA_EVENTS {
-    DFHACK_LUA_EVENT(onWorkshopFillSidebarMenu),
```
myk002, line 115: Please undo unnecessary formatting changes. It's hiding the important parts of the diff

myk002:
> didn't seem like something to open a PR over but I've been wrong before..

I can see already that it would have been better in separate PRs. Always separate refactoring/reformatting PRs from business logic changes. Otherwise the review feedback gets muddled.

cppcooper:
> it would have been better in separate PRs

Alright, I resorted things back to how they were.

library/modules/EventManager.cpp
```diff
@@ -58,11 +58,29 @@ static multimap<int32_t, EventHandler> tickQueue;
 
 //TODO: consider unordered_map of pairs, or unordered_map of unordered_set, or whatever
 static multimap<Plugin*, EventHandler> handlers[EventType::EVENT_MAX];
-static int32_t eventLastTick[EventType::EVENT_MAX];
+static std::unordered_map<EventHandler,int32_t> eventLastTick;
 
 static const int32_t ticksPerYear = 403200;
 
-void DFHack::EventManager::registerListener(EventType::EventType e, EventHandler handler, Plugin* plugin) {
+void enqueueTickEvent(EventHandler &handler){
```
cppcooper, line 65: Add a comment explaining this function is only used inside the file

library/modules/EventManager.cpp
```diff
@@ -104,35 +124,47 @@ void DFHack::EventManager::unregister(EventType::EventType e, EventHandler handl
...
         i = handlers[e].erase(i);
+        eventLastTick.erase(handler);
         if ( e == EventType::TICK )
             removeFromTickQueue(handler);
     }
 }
 
 void DFHack::EventManager::unregisterAll(Plugin* plugin) {
-    for ( auto i = handlers[EventType::TICK].find(plugin); i != handlers[EventType::TICK].end(); i++ ) {
+    for ( auto i = handlers[EventType::TICK].find(plugin); i != handlers[EventType::TICK].end(); ++i ) {
```
cppcooper, line 134: I've done this for any loops using iterators, as my understanding is `iter++` requires extra superfluous operations, a copy if I recall correctly.

myk002:
I'm still a little wary of this change, especially with how it changes the guarantees of consistency (i.e. objects referred to in the events actually still existing). My concern is that any issues with this change won't show up until after a release, and then there could be a fire drill in getting a fixed release out and updating all the starter packs.

I think we need to give ourselves some better mitigation options. Something that players can do themselves to fix things if the new behavior causes issues for them.

Could the behavior be toggled with a command players can run from dfhack.init? We can merge with the flag defaulting to false, then ask people who use mods that use EventManager to test. Then we can default the flag to true, then after a few releases with no reported issues, we can remove the old behavior entirely. How does that sound?

cppcooper:
Sounds like a simple way to avoid problems. Maybe I could additionally spend a bit of time and make comments about known life cycles in the managers.

myk002:
That would help a lot, I think. Thanks!

cppcooper:
Ok so I didn't put the comments in the managers, I thought it would make more sense to centralize the information to one location. So I put it with the `event_tracker` declarations

library/modules/EventManager.cpp
```diff
@@ -405,51 +578,92 @@ static void manageJobInitiatedEvent(color_ostream& out) {
...
-        return;
+    int32_t tick = df::global::world->frame_counter;
+    int32_t oldest_last_tick = (uint16_t)-1;
+
+    // update new jobs list
+    for ( df::job_list_link* link = df::global::world->jobs.list.next; link != nullptr; link = link->next ) {
+        df::job* job = link->item;
+        // the jobs must come after the last known job id
+        if (job && job->id > lastJobId) {
+            newJobs.emplace(tick, Job::cloneJobStruct(link->item, true));
```
myk002, line 589: Your recent comments about lifecycle management made me look at this a little closer. By cloning the job struct, doesn't this prevent handlers from modifying the job? Or if they do want to modify, do they now need to look the job up themselves? This might hold for any objects that we now make a copy of for lifecycle management purposes.

cppcooper: > doesn't this prevent handlers from modifying the job? Or if they do want to modify, do they now need to look the job up themselves?

Yea, if they want to modify the job they would need to look it up. If I go about optimizing all the job* stuff then maybe I can find a way to retain sending the "true" job* when possible for the first two job event types (ie. not completed).

First optimization I think would be to always send the true-job* when the event's tick age is 0. As for the rest.. I think I would need to search for the true-job* from the id.. so ig a whole bunch of binary searches. Maybe I could treat repeating jobs a bit differently and cache them(?) probably safer to just search them up as well.

The completed event type would still need to use cloned-job* though, for [I think] obvious reasons.

> This might hold for any objects that we now make a copy of for lifecycle management purposes.

I haven't taken to managing any other objects. Everything else is purely ids at the root of it. Except the constructions, those structs don't have ids in them (that I could see)

cppcooper: Aside from listeners running every tick, and that 0-age edge case.. I think it might be best to just leave the task of finding a modifiable job* to the listener. Though I certainly don't love the idea of sending different sets of pointers to different listeners based on their tick frequency and/or happenstance (0-age).

myk002: If this were a new API, I might agree, but I think we have to provide the same guarantees as the existing implementation in order to avoid breaking things

cppcooper: I shall start by changing the type in the declaration.. let's see. Then 6 hours later we find ourselves mid refactor of all the scanning code.

cppcooper:
So at some point in revising towards an ability to throw out stale id references I began refactoring various managers to share scan loops. Might be a week or two.. I guess we'll see. I'm knee deep in the code for all the report related stuff.
image

I just need to finish transposing some intricately detailed code to these new scan functions, spot as many optimization zones as possible, then simplify the logic and expressions.. it's probably 400 lines of code to move

cppcooper:
> just need to finish transposing

So that's done. I'll finish off the todos next.

I deal with some stale id's, but I haven't touched most of the event data yet. There is also some code I think I might have some luck simplifying.. and some comments I need to add.

Afterwards I'll try to find an easy way to split this into two PR's. One with the scan centralization, and another with the data structure changes and changes to the managers.

Speaking of the managers, I changed them all to `send.*Events`. Mostly to keep track, but also I think it makes more sense not sure whether I should change them back or not.

myk002:
Let's take this up again after discussion in #2190

library/include/modules/EventManager.h
```diff
@@ -29,8 +28,12 @@ namespace DFHack {
...
                 UNIT_DEATH,
                 ITEM_CREATED,
-                BUILDING,
-                CONSTRUCTION,
+                BUILDING_DESTROYED,
+                BUILDING_CREATED,
+                BUILDING, // todo: deprecate this event
+                CONSTRUCTION_REMOVED,
+                CONSTRUCTION_ADDED,
+                CONSTRUCTION, // todo: deprecate this event
```
cppcooper, line 36: This can be made into a separate PR.

library/modules/EventManager.cpp
```diff
@@ -193,1087 +240,1564 @@ std::array<eventManager_t,EventType::EVENT_MAX> compileManagerArray() {
...
-        nextItem = *df::global::item_next_id;
-        nextBuilding = *df::global::building_next_id;
+        // if there's a new invasion we'll add it
+        if (!df::global::ui)
+            return;
+        invasions.emplace(tick, nextInvasion);
         nextInvasion = df::global::ui->invasions.next_id;
-        lastJobId = -1 + *df::global::job_next_id;
+    }
+    inline void clear() {
```
cppcooper, line 473: EventManager is basically intended to centralize scanning data structures for changes, and centralizing inside the implementation is just a good next step to that. So this can be turned into a separate PR. Though this will need rewriting as this is intended to also prepare sending on multiple tick frequencies.

library/modules/EventManager.cpp
```diff
@@ -193,1087 +240,1564 @@ std::array<eventManager_t,EventType::EVENT_MAX> compileManagerArray() {
     return managers;
 }
 
+template<typename T, bool has_duplicates = false>
+class event_tracker {
+private:
+    std::unordered_map<T, int32_t> seen;
+    std::multimap<int32_t, T> history;
+public:
```
cppcooper, line 248: Sending events on multiple tick frequencies will require a separate PR as well, as it changes the code most substantially, and myk requested a full writeup.. both of which warrant a completely separate PR for these changes.

library/modules/EventManager.cpp
```diff
@@ -58,11 +64,30 @@ static multimap<int32_t, EventHandler> tickQueue;
...
+}
+
+void DFHack::EventManager::registerListener(EventType::EventType e, EventHandler handler, Plugin* plugin, bool backlog) {
+    if(e == EventType::TICK){
+        enqueueTickEvent(handler);
+    }
+    int32_t tick = backlog ? -2 : (df::global::world ? df::global::world->frame_counter : -1);
+    eventLastTick[handler] = tick; // received ranges.. // backlog: -1,n // !world: 0,n // world: tick,n
     handlers[e].insert(pair<Plugin*, EventHandler>(plugin, handler));
 }
```
cppcooper, line 92: Again, this can be a separate PR. This allows registering tick event listeners (instead of constantly re-queuing your tick events). Many of my onUpdate plugin functions could be replaced with this.

library/modules/EventManager.cpp
myk002: Specifically, what I'm looking for is justification for the change. I want to see how the added complexity makes real world use cases possible. How would current users and actual planned future users benefit?

The usual behavior of an event trigger framework is to notify the listeners when an event occurs. Our current implementation is closer to this behavior than the proposed one, which is more of a watermarked pub/sub architecture. Pub/sub can certainly be useful, but it's not obvious to me how it's useful for DFHack's use case.

PR #2108, opened 2022-04-19
[tweak] fast-trade: Fix Shift+UP to select current item

softmoth:
The DOWN and UP keys should both work the same way: toggle the item on
the *current* line, then move the cursor (down or up) to the next item.

While here, since the logic is the same when moving goods to the depot
and during the trade, it is cleaner to pull it out into a common
definition.

### DIFF
--- a/plugins/tweak/tweaks/fast-trade.h
+++ b/plugins/tweak/tweaks/fast-trade.h
@@ -1,59 +1,44 @@
 using namespace std;
 
+static df::interface_key fast_trade_move_key(const set<df::interface_key> *input)
+{
+    return input->count(interface_key::CURSOR_DOWN_FAST)
+         ? interface_key::STANDARDSCROLL_DOWN
+         : input->count(interface_key::CURSOR_UP_FAST)
+         ? interface_key::STANDARDSCROLL_UP
+         : interface_key::NONE;
+}
+
+#define DEFINE_FAST_TRADE_INTERPOSE(is_trading)                             \
+    DEFINE_VMETHOD_INTERPOSE(void, feed, (set<df::interface_key> *input))   \
+    {                                                                       \
+        df::interface_key move_key;                                         \
+        if ((is_trading) && (move_key = fast_trade_move_key(input)))        \
+        {                                                                   \
+            set<df::interface_key> tmp;                                     \
+            tmp.insert(interface_key::SELECT);                              \
+            INTERPOSE_NEXT(feed)(&tmp);                                     \
+                                                                            \
+            tmp.clear();                                                    \
+            tmp.insert(move_key);                                           \
+            INTERPOSE_NEXT(feed)(&tmp);                                     \
+        }                                                                   \
+        else                                                                \
+            INTERPOSE_NEXT(feed)(input);                                    \
+    }
+
+
 struct fast_trade_assign_hook : df::viewscreen_layer_assigntradest {
     typedef df::viewscreen_layer_assigntradest interpose_base;
 
-    DEFINE_VMETHOD_INTERPOSE(void, feed, (set<df::interface_key> *input))
-    {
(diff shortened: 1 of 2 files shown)


lethosor:
I believe the design goal here was that shift-up would be the inverse of shift-down, i.e. pressing one followed by the other would have no net effect. Not sure if this was detailed in a thread on GitHub or just on IRC at the time.

softmoth:
I get that, but it's the wrong decision, in my opinion. It's aggressively bad when, e.g., I page down and then want to select items from the bottom up. It breaks every UI expectation I have from every other program in existence.

softmoth:
I expect the resolution to this difference of opinion is above my pay grade. I think I've made my argument, and I think it will take people a very short time to retrain their fingers to just let go of Shift for one tap when you've gone too far (in *either* direction, now, the same instinct will produce the same results). I encourage you to try it. But if the community feels that our commitment to the prior behavior is more valuable than switching to the **obvious** improvement, I'll let it go. :zany_face:

plugins/tweak/tweaks/fast-trade.h
```diff
@@ -1,59 +1,44 @@
...
+
+#define DEFINE_FAST_TRADE_INTERPOSE(is_trading)                             \
+    DEFINE_VMETHOD_INTERPOSE(void, feed, (set<df::interface_key> *input))   \
+    {                                                                       \
+        df::interface_key move_key;                                         \
+        if ((is_trading) && (move_key = fast_trade_move_key(input)))        \
+        {                                                                   \
+            set<df::interface_key> tmp;                                     \
+            tmp.insert(interface_key::SELECT);                              \
+            INTERPOSE_NEXT(feed)(&tmp);                                     \
```
myk002, line 20: This appears to lose the extra SELECT when `in_edit_count` is true.

plugins/tweak/tweaks/fast-trade.h
```diff
@@ -1,59 +1,44 @@
 using namespace std;
 
+static df::interface_key fast_trade_move_key(const set<df::interface_key> *input)
+{
+    return input->count(interface_key::CURSOR_DOWN_FAST)
```
myk002, line 5: I can see what this is doing, but I have to think. That means the code is more clever than it has to be. Please use plain if statements for the sake of readability.

myk002 (changes requested):
I can see reasons to keep the behavior as it is, and I can see reasons to change like this PR proposes.

The existing behavior treats shift up as "undo", so shift down down up is the same as shift down. This is desirable if you're selecting a bunch of items and went one too far and have to backtrack. However, if you're scanning through a list of items, and some are selected and some aren't, and you want to deselect some specific items (say because they are too expensive), it is handy to have both shift up and shift down toggle the current item and then move in the indicated direction.

This PR changes shift up to be an alternate form of selection, allowing you to select items going up just as easily as selecting items going down. However, it makes the "one too many" case a bit more complicated. After shift down down, undoing the last item would be up shift up (or up shift down), which is a more complicated sequence.

Both implementations make sense. Which one makes *more* sense?

Personally, I encounter both use cases. I take advantage of the current, "undo" behavior during initial selection, but I would really appreciate the proposed "toggle then move" behavior when fine-tuning my trade selections later. During fine tuning, the current "move then toggle" behavior trips me up every darn time.

I'll have to try this PR out and see whether I miss the old behavior more than I appreciate the new behavior.

myk002:
After playing a game with this enabled, I think I'm going to vote against merging. The existing "reverse to undo" behavior appears to be useful more often (at least in my testing) than the proposed "select current then move" behavior. I find myself leaving a mess of toggled items with the new behavior that are harder to clean up than with the old behavior. I don't think this is something I could get used to.

softmoth:
OK, no worries. Thanks for giving it a run. Let's close this, I don't think we'll get more feedback by leaving it open....

myk002:
Sgtm. Thanks!

Issue #2036, opened 2022-03-17
C++ code coverage support


myk002:
DFHack has support for measuring coverage of Lua code, but has nothing for the C++ code. As we prepare DFHack for compatibility testing with the Steam version of DF, we'll need tooling to verify whether code is working properly on the new target.

I don't have enough knowledge for how to do this on Windows, but on platforms where we build with gcc (Linux and OSX), we can compile with `-O0 --coverage` to allow for `gcov` to generate coverage reports.

The workflow will be something like:
```
$ cd <dfhack src tree>/build
$ ccmake .. (and set COVERAGE to ON)
$ ninja install
$ dfhack
DFHack # test -- die (this will run all tests of all three types: 'none', 'title', and 'fortress')
$ gcov (with options and file names to generate the annotated output)
```

I am unsure if the way dfhack causes the process to exit will allow the coverage data to be flushed to disk. If not, we might be able to manually call `__gcov_flush()` before exiting the process.

Once this basic workflow is functional, we can integrate it into Github Actions. I found https://docs.codecov.com/docs, which is a paid product, but can be used for free by open source projects. I'm still looking into this.

ab9rf:
Code coverage for C++ in MSVC is only available in Enterprise Edition, and the cheapest legitimate source I can find for VSEE is $1100 (there are keys available for less, but I'm dubious of their provenance).

myk002:
Ugh, yeah, Windows-based tools seem very not free. Here are some other, non-free options: https://stackify.com/code-coverage-tools/

This probably means that we'll only get coverage measurement on *nix, but the tests themselves should run on Windows just fine. That will be good enough to give us confidence that things are working, I think.

PR #2109, opened 2022-04-21
unit tests: Add CTest support, and a trivial first unit test

softmoth:
If BUILD_TESTS=ON:
- Adds a 'test' target for ninja
- Adds a library/MiscUtils.test unit test executable

This uses the CTest tool that is part of CMake. Hopefully it might provide a basis for unit tests. I tried to do the simplest thing that works for this first commit. CMake has features to integrate Boost.Test or Google Test frameworks, but I'm not sure if that's the direction to go in.

## Questions
- CTest creates a `BUILD_TESTING` option (on by default) that users can turn off to disable tests. This makes an awkward clash with our `BUILD_TESTS` option which is meant to do the same thing (governs our Lua integration tests). How should this be handled? I propose removing `BUILD_TESTS` and using the CMake-standard `BUILD_TESTING` option for both. If there's a way to make a hidden CMake option `BUILD_TESTS` that sets `BUILD_TESTING` for backwards compatibility, that should be added as well.

## Flaws in this Draft
- Dependencies are broken. This shows up on Windows builds.
- Dependencies are broken. The automatically-generated `test` target does not depend on executables added with `add_test`. This is an obvious requirement, so I must be calling things incorrectly.
- Missing documentation. If this PR solidifies into something that can be merged, I'll add docs.

## Follow-up improvements needed
- Integration with CI and github commit checks.
- Writing unit tests by hand is not a long-term solution; we'll want to [use Boost.Test or Google Test](https://docs.microsoft.com/en-us/visualstudio/test/how-to-use-ctest-for-cpp?view=vs-2022#to-run-tests) or something. 

What do you think?


softmoth:
I forgot to mention: a problem I'm having is that `ninja test` will try to run even if `ninja` hasn't completed yet. I can't figure out how to make it build first before it runs `ctest`. The docs for `add_dependency()` make it sound like it's not helpful, but I don't know what else to use.

myk002:
This definitely has potential. The cpp code is very ripe for testing.

CTest seems perfectly fine for running the tests, but I think we need a better framework for *writing* the tests. How difficult would it be to integrate something like Google Test and Google Mock? https://chromium.googlesource.com/external/github.com/google/googletest/+/refs/tags/release-1.8.0/googlemock/docs/ForDummies.md

softmoth:
> How difficult would it be to integrate something like Google Test

CTest has built-in support for Boost.Test and Google Test. I thought I could work on that once I get a bare-bones first test or two going.

myk002:
Copied from discord so the questions don't get lost:

So I guess there's a number of design decisions we need to make around cpp tests. Should they be 1 binary per test or compiled into a monolithic suite? Should we even have ctest integration, or should the test interface be a plugin (to support integration tests)? Should we have both, one for pure unit tests and the other for integration tests? Is it better/easier to test all cpp code from Lua?

Also, how will we measure and report cpp code coverage in each of the above scenarios? I have some initial thoughts on this at #2036

myk002:
To answer your question in the description, it should be fine to replace BUILD_TESTS with BUILD_TESTING, but they'd need to exist in parallel at least for a while until CI is migrated. What do you think about my design questions in the previous comment?

softmoth:
I'm not sure we need to decide up front whether it's monolithic or not. If we're using gtest + CTest, I think we can combine as many tests as makes sense into feature-specific or module-specific test binaries, for example. CTest will run whatever tests we make and automatically aggregate individual cases within.

IMO integration tests should be separate, done via lua. By "integration tests" I mean tests that run Dwarf_Fortress and exercise the entire system. That's already working great with lua, no need to do it in C++ as well. But for testing isolated C++ components, and for enabling test-driven design and safer refactoring, etc., unit tests  in C++ are useful.

myk002:
> I'm not sure we need to decide up front whether it's monolithic or not. If we're using gtest + CTest, I think we can combine as many tests as makes sense into feature-specific or module-specific test binaries, for example. CTest will run whatever tests we make and automatically enumerate individual cases within.
> 
> IMO integration tests should be separate, done via lua. By "integration tests" I mean tests that run Dwarf_Fortress and exercise the entire system. That's already working great with lua, no need to do it in C++ as well. But for testing isolated C++ components, and for enabling test-driven design and safer refactoring, etc., unit tests in C++ are useful.

SGTM. and I agree about using Lua for integration tests. I just wanted to know if you had any specific ideas/plans in that area.

CMakeLists.txt
```diff
@@ -497,14 +497,36 @@ if(BUILD_DOCS)
...
-option(BUILD_TESTS "Include tests (currently just installs Lua tests into the scripts folder)" OFF)
-if(BUILD_TESTS)
-    if(EXISTS "${dfhack_SOURCE_DIR}/test/scripts")
-        message(SEND_ERROR "test/scripts must not exist in the dfhack repo since it would conflict with the tests installed from the scripts repo.")
+# Testing with CTest
+include(CTest)
+
+include(CMakeDependentOption)
+cmake_dependent_option(
+    BUILD_TEST_SCRIPTS "Install integration tests in hack/scripts/test" OFF
```
myk002, line 505: perhaps `BUILD_TESTING_LUA` instead of `BUILD_TEST_SCRIPTS`. `BUILD_TEST_SCRIPTS` implies (at least to me) that only the scripts are being tested. The Lua-based testing is more general than that.

CMakeLists.txt
```diff
@@ -497,14 +497,36 @@ if(BUILD_DOCS)
...
+mark_as_advanced(FORCE BUILD_TESTS)
+
+# Handle deprecated BUILD_TESTS option
+option(BUILD_TESTS "Deprecated option; please use BUILD_TEST_SCRIPTS=ON" OFF)
+if(BUILD_TESTING AND BUILD_TESTS)
+    set(BUILD_TEST_SCRIPTS ON FORCE)
+endif()
+
+if(BUILD_TESTING)
+    if(BUILD_TEST_SCRIPTS)
```
myk002, line 516: I think the Lua test option should be independent of the c++ test option. If `BUILD_TESTING` is `OFF`, shouldn't the lua tests still be installed?

library/CMakeLists.txt
```diff
@@ -82,6 +82,18 @@ set(MAIN_SOURCES
     RemoteTools.cpp
 )
 
+if(BUILD_TESTING)
+    # TODO Make a function or macro for this
+    add_executable(MiscUtils.test MiscUtils.test.cpp)
+    target_link_libraries(MiscUtils.test dfhack)
```
myk002, line 88: is this necessary? what dependencies does MiscUtils have? If it does have dependencies, shouldn't they be mocked out for the unit tests? How hard is it to just build MiscUtils.cpp instead of the entire dfhack library?

library/CMakeLists.txt
```diff
@@ -82,6 +82,18 @@ set(MAIN_SOURCES
...
+    add_test(NAME MiscUtils.test COMMAND MiscUtils.test)
+
+    # How to get `test` to ensure everything is up to date before running
+    # tests? This add_dependencies() fails with:
+    #   Cannot add target-level dependencies to non-existent target "test".
+    #add_dependencies(test MiscUtils.test)
```
myk002, line 94: when does the `test` target get defined? can we include things in a different order in the top level CMakeLists.txt file to fix this?

library/CMakeLists.txt
```diff
@@ -413,6 +425,8 @@ if(APPLE)
     target_link_libraries(dfhack ncurses)
     set_target_properties(dfhack PROPERTIES VERSION 1.0.0)
     set_target_properties(dfhack PROPERTIES SOVERSION 1.0.0)
+elseif(UNIX)
+    target_link_libraries(dfhack SDL)
```
myk002, line 429: what prompted this change? do the tests not run without it?

myk002 (changes requested):

myk002:
bumping from project for now since the review isn't active. Once you're ready to take this up again, we can look into getting it merged!

lethosor:
I might also be interested in taking a stab at this if that would be helpful

PR #2113, opened 2022-04-25
Add functions reverse-engineered from announcement code

Bumber64:
Added autoDFAnnouncement, which takes a report_init and a string, and handles them like DF does. (E.g., handling repeats and ignoring those in unrevealed map.)

Added pauseRecenter, which recenters on an xyz coord and optionally pauses, while respecting pause_zoom_no_interface_ms.

~~Added recenterViewscreen, which recenters on an xyz coord using a report zoom style (Generic, Item, or Unit. Existing revealInDwarfmodeMap function names "Unit" style "bool center". Generic style ignores coords and just enforces valid view bounds.)~~ Just replace revealInDwarfmodeMap with new code.

~~Added parseReportString, which parses a string using '&' as a control character (&r as two newlines, && as &) and cuts to a certain length w/o splitting words.~~ Reduced this to a utility function, unlikely to be needed outside gui.cpp.

~~Changes can be made to existing functions based on information gained. Will address in separate PR.~~ Might as well address here.


docs/Lua API.rst
```diff
@@ -1004,6 +1004,22 @@ Fortress mode
   Resets the fortress mode sidebar menus and cursors to their default state. If
   ``pause`` is true, also pauses the game.
 
+* ``dfhack.gui.pauseRecenter(pos[,pause])``
+  ``dfhack.gui.pauseRecenter(x,y,z[,pause])``
+
+  Same as ``resetDwarfmodeView``, but also recenter if ``x`` isn't ``-30000``,
+  and respects RECENTER_INTERFACE_SHUTDOWN_MS in DF's init.txt.
```
myk002, line 1011: What does the `RECENTER_INTERFACE_SHUTDOWN_MS` setting do?

myk002, line 1011: What does the `RECENTER_INTERFACE_SHUTDOWN_MS` setting do? I'm asking because I don't know, but you should probably explain in the docs because others might not know too

docs/Lua API.rst
```diff
@@ -1004,6 +1004,22 @@ Fortress mode
...
+
+* ``dfhack.gui.recenterViewscreen(pos[,zoom])``
+  ``dfhack.gui.recenterViewscreen(x,y,z[,zoom])``
+  ``dfhack.gui.recenterViewscreen([zoom])``
+
+  Recenter the view on a position using a specific zoom type. If no position is
+  given, recenter on ``df.global.cursor``. Zoom types are ``df.report_zoom_type``
+  (0 = Generic, 1 = Item, 2 = Unit), where Generic skips recentering and
+  enforces valid view bounds (the same as x = -30000.)
+  Default zoom type is Item.
```
myk002, line 1021: How do `Item` and `Unit` behave?

myk002, line 1021: How do `Item` and `Unit` behave?

docs/Lua API.rst
```diff
@@ -1073,6 +1089,14 @@ Announcements
   Uses the type to look up options from announcements.txt, and calls the above
   operations accordingly. The units are used to call ``addCombatReportAuto``.
 
+* ``dfhack.gui.autoDFAnnouncement(report,text[,log_failures])``
+  ``dfhack.gui.autoDFAnnouncement(type,pos,text,color[,is_bright,unit1,unit2,is_sparring,log_failures])``
+
+  Takes a ``df.report_init`` and a string and processes them just like DF does.
+  Sometimes this means the announcement won't occur. Set ``log_failures`` to ``true`` to
+  log the reason why to the dfhack console (e.g., unrevealed map or wrong gamemode.)
+  Can also be built from parameters instead of a ``report_init``.
```
myk002, line 1098: Can we assume that if someone is calling this, they know what the types and purpose of all the parameters are? Do we need any more details here?

myk002, line 1098: Can we assume that if someone is calling this, they know what the types and purpose of all the parameters are?

docs/Lua API.rst
Bumber64: DF's init.txt says:
"This controls the number of milliseconds that must pass before input works again after the view recenters on an event in dwarf mode."
Not sure why we'd need to document DF's features, but I guess it wouldn't hurt.

docs/Lua API.rst
Bumber64: The functions above explain some of them. I'll explain is_sparring and give a link to the report_init struct definition.

docs/Lua API.rst
myk002: A link to existing df docs would be fine too

docs/Lua API.rst
myk002: A link to existing docs would work here too if the behavior is already documented somewhere

myk002 (commented):
Waiting on resolution in DFHack/df-structures#454 about sparring vs. not_sparring

Bumber64:
I'm not entirely sure "recenterViewscreen" uses "report_zoom_type", but it's the closest enum I could find that makes sense.

From what I can tell, zoom_type1 and zoom_type2 in reports are always left at 0 (Generic) and are ignored by the game. They aren't used during the initial announcement, which calls pauseRecenter, which always calls recenterViewscreen with value 1 (which corresponds to report_zoom_type::Item if that's the actual struct.) Pressing 'z' in game when highlighting a report isn't affected by changing these either. So "report_zoom_type" isn't actually used by DF reports (anymore.)

I'm not sure how the report_zoom_type enum was determined, but I would think "Unit" and "Item" would be swapped, since following a unit in game uses the lazy style that doesn't put them in the exact center. For that reason I'm suspicious of it.

library/LuaApi.cpp
```diff
@@ -1505,7 +1507,127 @@ static const LuaWrapper::FunctionReg dfhack_gui_module[] = {
...
+        else
+            rv = Gui::autoDFAnnouncement(*r, message);
+    }
+    else
+    {
+        df::coord pos;
+        int color;
+        bool bright, is_sparring, log_failures;
+        df::unit *unit1, *unit2;
+
```
myk002, line 1530: it looks like you need to initialize these to make the compiler happy. this is tripping `-Werror`.

library/modules/Gui.cpp
```diff
@@ -1360,6 +1362,137 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
     fseed.close();
 }
 
+bool Gui::parseReportString(std::vector<std::string> &out, const std::string &str, size_t line_length)
+{   // out vector will contain strings cut to line_length, avoiding cutting up words
+    // Reverse-engineered from DF announcement code, fixes applied
```
myk002, line 1367: out of curiosity, what fixes were made?

docs/Lua API.rst
```diff
@@ -1073,6 +1088,15 @@ Announcements
   Uses the type to look up options from announcements.txt, and calls the above
   operations accordingly. The units are used to call ``addCombatReportAuto``.
 
+* ``dfhack.gui.autoDFAnnouncement(report,text[,log_failures])``
```
myk002, line 1091: could you link to where the `&` escape semantics are documented?

library/modules/Gui.cpp
```diff
@@ -1360,6 +1362,137 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
     fseed.close();
 }
 
+bool Gui::parseReportString(std::vector<std::string> &out, const std::string &str, size_t line_length)
```
myk002, line 1365: Could this function re-use `word_wrap()` in https://github.com/DFHack/dfhack/blob/develop/library/MiscUtils.cpp#L171 ?
Like, could we pre-process the escape chars and send the strings, line by line, to `word_wrap()` for the actual wrapping?

myk002 (commented):
not quite done reviewing the code, but I wanted to send you what I had so far so we could discuss.

library/modules/Gui.cpp
Bumber64: The issue with cutting the string at the wrong index, described in the notes of this bug report:
http://www.bay12games.com/dwarves/mantisbt/view.php?id=11813#c41165

Bumber64: The unfixed code is posted on the Bay12 Forums:
http://www.bay12forums.com/smf/index.php?topic=98412.msg8324640#msg8324640

docs/Lua API.rst
Bumber64: Those are from reverse engineering the parsing function. I've added the info here now.

docs/Lua API.rst
```diff
@@ -1073,6 +1088,15 @@ Announcements
...
+  ``dfhack.gui.autoDFAnnouncement(type,pos,text,color[,is_bright,unit1,unit2,is_sparring,log_failures])``
+
+  Takes a ``df.report_init`` (see: https://github.com/DFHack/df-structures/blob/master/df.announcements.xml#L451)
```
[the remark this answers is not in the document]
Bumber64: Sure.

docs/Lua API.rst
```diff
@@ -1073,6 +1088,15 @@ Announcements
...
+  and a string and processes them just like DF does. Sometimes this means the announcement won't occur.
+  Set ``log_failures`` to ``true`` to log the reason why to the dfhack console (e.g., unrevealed map or wrong gamemode.)
```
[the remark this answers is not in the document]
Bumber64: It's new for DFHack, along with the return values for the function (which I should document.) I'll look into using debugfilter.

library/include/modules/Gui.h
```diff
@@ -124,13 +127,25 @@ namespace DFHack
 
         // Show an announcement with effects determined by announcements.txt
         DFHACK_EXPORT void showAutoAnnouncement(df::announcement_type type, df::coord pos, std::string message, int color = 7, bool bright = true, df::unit *unit1 = NULL, df::unit *unit2 = NULL);
+        
+        // Process an announcement exactly like DF would, which might result in no announcement
+        DFHACK_EXPORT int autoDFAnnouncement(df::report_init r, std::string message);
+        DFHACK_EXPORT int autoDFAnnouncement(df::report_init r, std::string message, bool log_failures);
+        DFHACK_EXPORT int autoDFAnnouncement(df::announcement_type type, df::coord pos, std::string message, int color = 7, bool bright = true, df::unit *unit1 = NULL, df::unit *unit2 = NULL, bool sparring = false, bool log_failures = false);
```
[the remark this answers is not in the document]
lethosor: I would add a couple line breaks to get it down to roughly the same length as the surrounding declarations. No formal guidelines at the moment.

docs/Lua API.rst
```diff
@@ -1004,6 +1004,21 @@ Fortress mode
...
+
+  Same as ``resetDwarfmodeView``, but also recenter if ``x`` isn't ``-30000``, and respects
+  RECENTER_INTERFACE_SHUTDOWN_MS (the delay before input is recognized when a recenter occurs) in DF's init.txt.
+
+* ``dfhack.gui.recenterViewscreen(pos[,zoom])``
+  ``dfhack.gui.recenterViewscreen(x,y,z[,zoom])``
+  ``dfhack.gui.recenterViewscreen([zoom])``
+
+  Recenter the view on a position using a specific zoom type. If no position is given,
+  recenter on ``df.global.cursor``. Zoom types are ``df.report_zoom_type`` (0 = Generic, 1 = Item, 2 = Unit),
```
lethosor, line 1018: IMO we should not be listing the raw numeric values of enum items in the docs - the point of the names is that they are more stable than the raw values, so we should encourage people to pass in parameters by name by not providing numbers in the docs.

(Ideally the explanation of what the values mean would be in docs for this specific enum, but we don't have that yet, so I'm fine with it here instead.)

docs/changelog.txt
```diff
@@ -42,10 +42,13 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Misc Improvements
 
 ## Documentation
+- ``Lua API.rst``:  added ``autoDFAnnouncement``, ``pauseRecenter``, ``recenterViewscreen``
 
 ## API
+- add functions reverse-engineered from announcement code: ``Gui::parseReportString``, ``Gui::autoDFAnnouncement``, ``Gui::pauseRecenter``, ``Gui::recenterViewscreen``
```
lethosor, line 48: I probably should have chimed in on your last PR, but I think the only changelog entry we need is this one. Documentation in particular is implied - we don't need to mention in the changelog that a new function was added _and_ that it was documented at the same time. I'm not as picky about the "Lua" section, but historically I have left it out (i.e. I've only considered listing additions to the API to be needed in the "API" section).

library/modules/Gui.cpp
```diff
@@ -1574,6 +1707,239 @@ void Gui::showAutoAnnouncement(
...
+{   // Reverse-engineered from DF announcement code
+
+    if (!world->unk_26a9a8) // TODO: world->show_announcements
+        return 1;
+
+    df::announcement_flags a_flags;
+    if (is_valid_enum_item(r.type))
+        a_flags = df::global::d_init->announcements.flags[r.type];
+    else
+        return 2;
```
lethosor, line 1720: Can these return codes be made an enum instead of magic numbers? Such an enum should probably be defined in `Gui.h` because `LuaApi.cpp` also needs it, and any callers would benefit from it too.

I wouldn't worry too much about exposing the enum to Lua yet, but if you want to, you could put it in df-structures as `report_failure_reason` or something along those lines. If it's an enum that DF doesn't use (I'm not sure if you added these codes yourself or if they were part of the reverse-engineered code), I would follow our existing pattern of prefixing it with `dfhack_`.

library/modules/Gui.cpp
```diff
@@ -1574,6 +1707,239 @@ void Gui::showAutoAnnouncement(
...
+    int rv = autoDFAnnouncement(r, message);
+
+    if (log_failures)
+    {
+        switch (rv)
+        {
+            case 0:
+                break; // success
+            case 1:
+                Core::print("Skipped an announcement because world->show_announcements is false:\n%s\n", message.c_str());
```
lethosor, line 1888: ...and if you put the enum in df-structures, you could make an attribute for these descriptions, which would make them accessible directly to C++ and Lua without needing to write an extra function.

docs/Lua API.rst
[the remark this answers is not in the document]
lethosor: I also want to point out that, either way, this link is not stable because right line numbers on `master` can change. And if we use a permalink to a specific commit, it could also become out of date by linking to an old/inaccurate definition. Not too sure what the best solution is, since we don't have docs for DF types, so I guess this is ok for now. At least the filename should be stable for the time being.

library/modules/Gui.cpp
Bumber64: I added them so dfhack scripters could know why their announcements aren't displaying. ``debugfilter`` can handle this, and we can return boolean if the announcement printed or not.

The remaining use case is that somebody wants to do something different based on why it failed, which is niche enough that they can write their own function.

lethosor: I know that's what the numbers are for - my point is that they ought to be in an enum instead of just being hardcoded like this. If you change `2` here, you have to change it in at least one other place. (I misinterpreted what was going on in LuaApi.cpp earlier - looks like this is only really used in Gui.cpp at the moment, so the enum could be internal to that file.)

Also, in terms of API design, I really think the numbers should be possible to interpret from code. Right now, there is an overload of `autoDFAnnouncement()` that displays an announcement, optionally prints an error message, and returns. I think the responsibility of printing an error message should be left to the caller, which knows where it should go. Maybe the caller wants to print it to a file, or display it to the screen, or not print it immediately. `debugfilter` can't help with any of those use-cases, and we shouldn't have to worry about supporting all of those use-cases in every function that can fail with an error.

myk002: If the debug messages are output inline (where the error occurs) instead of here, there may not be a need for an enum at all.

lethosor: Sure, I think that's also fine. I guess I was thinking of them as more "errors" than "diagnostics". If they're not deemed important enough to pass up to the caller, that's fine by me - but the status codes _are_ being passed to the caller currently (even to Lua), which is why I thought they might be important.

So I guess to summarize my opinions here:
- If the status codes are returned to the caller, translation to messages should be done by the caller, if desired
- If the status codes aren't returned, I'd be fine with printing messages inline like you said
- If status codes appear more than once, they ought to be an enum (at least within Gui.cpp) to improve readability

library/modules/Gui.cpp
lethosor: Given https://github.com/DFHack/dfhack/pull/2113#discussion_r867597065, I would probably shy away from this unless the messages are "useful" enough for the caller to have access to them

docs/Lua API.rst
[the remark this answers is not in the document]
Bumber64: Having difficulty getting it to work.
```
#include "Debug.h"
...
DBG_DECLARE(core,gui,DebugCategory::LINFO);
...
DEBUG(gui, out).print("Test");
```
```
E0135	namespace "DFHack" has no member "debug_gui"	dfhack	C:\Users\Ryan\dfhack\library\modules\Gui.cpp
C2039	'debug_gui': is not a member of 'DFHack'	dfhack	C:\Users\Ryan\dfhack\library\modules\Gui.cpp
C2065	'out': undeclared identifier	dfhack	C:\Users\Ryan\dfhack\library\modules\Gui.cpp
C2228	left of '.print' must have class/struct/union	dfhack	C:\Users\Ryan\dfhack\library\modules\Gui.cpp
```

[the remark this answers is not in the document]
myk002: The DBG_DECLARE has to be inside of a namespace DFHack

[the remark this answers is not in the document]
myk002: E.g. https://github.com/DFHack/dfhack/blob/develop/plugins/debug.cpp#L42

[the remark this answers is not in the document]
Bumber64: Still doesn't recognize `out`, which needs to be a `color_ostream&`.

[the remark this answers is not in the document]
myk002: Is there an `out` accessible at the site if the call? If you don't have an `out`, you can actually omit the parameter and the logging framework will use the standard console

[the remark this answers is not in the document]
Bumber64: I actually tried omitting it, but I didn't put a `\n` after Test, so I thought it wasn't working when it didn't print. Thanks.

library/modules/Gui.cpp
Bumber64: I tried it and ran into an issue. Testing with this string:
```
"The quick brown fox jumps over                                                   the lazy dog!"
```
DF's code gives this result:
```
"The quick brown fox jumps over                                           "
"the lazy dog!"
```
Parsing then using `word_wrap` with `ignore_space` as `false`:
```
"The quick brown fox jumps over                                           "
"       the lazy dog!"
```
Parsing then using `word_wrap` with `ignore_space` as `true`:
```
"The quick brown fox jumps over the lazy dog!"
```
Not sure if I should just accept the `true` result or manually trim space from the output strings with `false` result.

myk002: @softmoth is this behavior from word_wrap expected? @Bumber64 how does the output differ from what you need/were expecting?

Bumber64: Exceeding `line_length` in characters (including spaces) should start a new line, ignoring only the spaces on the new line until a non-space character is reached. I'm okay with losing the spaces at the end of the previous line (my "fixed" code ended up doing this as a side-effect,) but spaces should never be removed between words on the same line, and the removal of spaces should not prevent the creation of a new line.

softmoth: I wasn't trying to match any other word wrapper, and nothing in DFHack depends on its precise behavior, so I'd recommend we make word_wrap() behave as Bumber64 needs, probably. I was thinking of a *possible* case of right-aligned tabulated data which should continue to line up properly after word wrapping; thus I preserved whitespace at the beginning of the new line. But I'm not sure that this is actually needed.

softmoth: By the way, the `compress_spaces = true` result matches the previous `word_wrap()` behavior, and the separate lua function as well. Both just comb through the input looking for chunks of non-whitespace and paste them together with a single space or newline, with no regard for what the input whitespace is.

The problem with trimming the leading whitespace after `word_wrap()` is that it leaves extra space at the end of the line then. If you never have a line wrap more than once, it doesn't cause a problem, but that's an ugly assumption.

Bumber64: > I was thinking of a _possible_ case of right-aligned tabulated data which should continue to line up properly after word wrapping; thus I preserved whitespace at the beginning of the new line. But I'm not sure that this is actually needed.

That's easy enough to keep by not trimming whitespace at all, right?

I just need an option for `word_wrap()` where a bool is set after pushing anything to `out`, which triggers a continue in the `if (isspace(c))` block when true, and is unset in the `else` block.

Bumber64: How does this look?
https://github.com/DFHack/dfhack/compare/develop...Bumber64:patch-1

myk002 (changes requested):

library/modules/Gui.cpp
```diff
@@ -1574,6 +1781,218 @@ void Gui::showAutoAnnouncement(
...
+    if (*gamemode == game_mode::ADVENTURE)
+    {
+        if (r.pos.x != -30000 &&
+            r.type != announcement_type::CREATURE_SOUND &&
+            r.type != announcement_type::REGULAR_CONVERSATION &&
+            r.type != announcement_type::CONFLICT_CONVERSATION &&
+            r.type != announcement_type::MECHANISM_SOUND)
+        {   // If not sound, make sure we can see pos
+            if ((world->units.active.empty() || (r.unit1 != world->units.active[0] && r.unit2 != world->units.active[0])) &&
+                ((Maps::getTileDesignation(r.pos)->whole & 0x10) == 0x0)) // Adventure mode uses this bit to determine current visibility
```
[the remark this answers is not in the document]
lethosor: `tile_designation` 0x10 appears to be _part_ of `tile_designation::dig`, which is 3 bits. From generated headers:

```c++
namespace df {
  union tile_designation {
    uint32_t whole;
    struct {
      uint32_t flow_size : 3; /*!< liquid amount */
      uint32_t pile : 1; /*!< stockpile? */
      df::tile_dig_designation dig : 3;
// ...
    enum Mask : uint32_t {
      mask_flow_size = 0x7U,
      mask_pile = 0x8U,
      mask_dig = 0x70U,
```

Not really sure if there's a better solution here. I doubt df-structures supports generating a union within a bitfield (not even sure if C++ does, honestly).

library/modules/Gui.cpp
```diff
@@ -1624,6 +2043,71 @@ df::coord Gui::getCursorPos()
     return df::coord(cursor->x, cursor->y, cursor->z);
 }
 
+void Gui::recenterViewscreen(int32_t x, int32_t y, int32_t z, df::report_zoom_type zoom)
```
lethosor, line 2046: Is this function meaningfully different from `Gui::revealInDwarfmodeMap(..., center=true)`? I expect there are some slight behavioral differences (e.g. with the separate `Item` vs. `Unit` modes) but does DFHack need two slightly different ways to center the viewport?

docs/Lua API.rst
```diff
@@ -1073,6 +1088,15 @@ Announcements
   Uses the type to look up options from announcements.txt, and calls the above
   operations accordingly. The units are used to call ``addCombatReportAuto``.
 
+* ``dfhack.gui.autoDFAnnouncement(report,text)``
+  ``dfhack.gui.autoDFAnnouncement(type,pos,text,color[,is_bright,unit1,unit2,is_sparring])``
+
+  Takes a ``df.report_init`` (see: `structure definition <https://github.com/DFHack/df-structures/blob/master/df.announcements.xml#L451>`_)
+  and a string and processes them just like DF does. Sometimes this means the announcement won't occur.
```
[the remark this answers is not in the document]
Bumber64: Old API guarantees an announcement is displayed. I'll add more info on how mine ignores hidden units, etc.

library/modules/Gui.cpp
```diff
@@ -1624,6 +2043,71 @@ df::coord Gui::getCursorPos()
...
+
+    if (*gamemode != game_mode::DWARF)
+        return;
+
+    resetDwarfmodeView(pause);
+    if (x != -30000)
+    {
+        recenterViewscreen(x, y, z, report_zoom_type::Item);
+        ui_sidebar_menus->minimap.need_render = true;
+        ui_sidebar_menus->minimap.need_scan = true;
```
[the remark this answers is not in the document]
Bumber64: DF doesn't, but I don't think there's an issue in doing so.

library/modules/Gui.cpp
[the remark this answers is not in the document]
Bumber64: Related: https://github.com/DFHack/df-structures/issues/448

library/modules/Gui.cpp
```diff
@@ -1574,6 +1781,218 @@ void Gui::showAutoAnnouncement(
...
+    }
+
+    // Check for repeat report
+    int32_t repeat_count = check_repeat_report(results); // Does nothing outside dwarf mode
+    if (repeat_count > 0)
+    {
+        if (a_flags.bits.D_DISPLAY)
+        {
+            world->status.display_timer = r.display_timer;
+            Gui::writeToGamelog("x" + to_string(repeat_count + 1));
```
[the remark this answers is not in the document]
Bumber64: Doesn't seem to.

library/modules/Gui.cpp
Bumber64: Was going to handle this in a different PR, where the code looks like this:
```
bool Gui::revealInDwarfmodeMap(df::coord pos, bool center)
{
    using df::global::window_x;
    using df::global::window_y;
    using df::global::window_z;

    if (!window_x || !window_y || !window_z || !world)
        return false;
    if (!Maps::isValidTilePos(pos))
        return false;

    recenterViewscreen(pos, (center ? report_zoom_type::Unit : report_zoom_type::Item));

    return true;
}
```

Should I rename ``recenterViewscreen`` to ``revealInDwarfmodeMap``?

library/modules/Gui.cpp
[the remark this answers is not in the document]
Bumber64: Does this relate at all to ``refreshSidebar()``? Seems like that function might be using a workaround for the kind of stuff we're doing more directly here.

docs/Lua API.rst
```diff
@@ -1004,6 +1004,21 @@ Fortress mode
...
+
+* ``dfhack.gui.recenterViewscreen(pos[,zoom])``
+  ``dfhack.gui.recenterViewscreen(x,y,z[,zoom])``
+  ``dfhack.gui.recenterViewscreen([zoom])``
+
+  Recenter the view on a position using a specific zoom type. If no position is given,
+  recenter on ``df.global.cursor``. Zoom types are ``df.report_zoom_type`` (see: `enum definition <https://github.com/DFHack/df-structures/blob/master/df.announcements.xml#L438>`_),
+  where ``Generic`` skips recentering and enforces valid view bounds (the same as x = -30000,) ``Item`` brings
+  the position onscreen without centering, and ``Unit`` centers the screen on the position. Default zoom type is Item.
+
```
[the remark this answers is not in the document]
Bumber64: Resolving this conversation since lethosor brought it up below.

library/modules/Gui.cpp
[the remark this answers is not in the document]
myk002: That's possible. If we can rewrite `refreshSidebar()` to work without sending actual keystrokes to the UI that would be a win.

myk002:
could you merge latest develop? the diff seems to have gone wonky.

library/include/modules/Gui.h
```diff
@@ -125,12 +127,24 @@ namespace DFHack
...
+
         /*
          * Cursor and window coords
          */
         DFHACK_EXPORT df::coord getViewportPos();
         DFHACK_EXPORT df::coord getCursorPos();
 
+        // Recenter the viewscreen, based on DF code for announcements and scrolling
+        DFHACK_EXPORT void pauseRecenter(int32_t x, int32_t y, int32_t z, bool pause);
+        DFHACK_EXPORT inline void pauseRecenter(df::coord pos, bool pause) { return pauseRecenter(pos.x, pos.y, pos.z, pause); }
```
myk002, line 143: why is there a `return` here? `pauseRecenter(x, y, z, pause)` is a `void` function

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
     fseed.close();
 }
 
+namespace
```
myk002, line 1371: I'm not sure if we have a policy around this, but most other DFHack code uses static functions instead of anonymous namespaces.

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
...
+            word_wrap(out, parsed, line_length, false/*, true*/);
+
+        return true;
+    }
+
+    bool recent_report(df::unit *unit, df::unit_report_type slot)
+    {
+        if (unit && !unit->reports.log[slot].empty() &&
+            *df::global::cur_year == unit->reports.last_year[slot] &&
+            (*df::global::cur_year_tick - unit->reports.last_year_tick[slot]) <= 500)
```
myk002, line 1414: what is the significance of `500`? Can we make it a named constant?

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
...
+        {
+            return true;
+        }
+        return false;
```
myk002, line 1418:
```suggestion
        return unit && !unit->reports.log[slot].empty() &&
            *df::global::cur_year == unit->reports.last_year[slot] &&
            (*df::global::cur_year_tick - unit->reports.last_year_tick[slot]) <= 500;
```

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
...
+            if (recent_report(unit, slot))
+                return true;
+        }
+        return false;
+    }
+
+    void delete_old_reports()
+    {
+        auto &reports = world->status.reports;
+        while (reports.size() > 3000)
```
myk002, line 1434: can we make `3000` into a named constant? maybe `MAX_REPORTS_SIZE`?

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
...
+        {
+            if (reports[0] != NULL)
+            {
+                if (reports[0]->flags.bits.announcement)
+                    erase_from_vector(world->status.announcements, &df::report::id, reports[0]->id);
+                delete reports[0];
+            }
+            reports.erase(reports.begin());
+        }
+    }
```
myk002, line 1444: how often do we expect to run this code? This is O(M*N) where M is the number of erased items and N is the number of items in the vector (which is > 3000). It seems like it could be brought down to O(N) fairly easily by batch removing elements.

library/modules/Gui.cpp
```diff
@@ -1360,6 +1368,100 @@ DFHACK_EXPORT void Gui::writeToGamelog(std::string message)
...
+
```
myk002, line 1445: note to self: I've reviewed up to this line

myk002 (changes requested):

library/modules/Gui.cpp
Bumber64: ``delete_old_reports`` is equivalent to a loop DF runs at the end of each announcement (or report) made. (The part run when ``reports[0] != NULL`` is DF's report destructor.) Technically, we could just not do this, and DF would handle it the next time the game does one of its own announcements. We'd be adding more iterations for DF to do in a single tick, however.

The number of erased items (M) should be equal to the number of report entries (lines of text) we just added, assuming we're keeping on top of things. That's most likely going to be a number < 5, which isn't bad.

So we could do slightly better, but this is what DF is doing all the time.

library/modules/Gui.cpp
Bumber64: It's the number of ticks DF considers a report to be "recent".

lethosor: IMO wouldn't hurt to have a constant for that. If dealing with a `coord` or `coord2d`, we could use its `isValid()` method instead of a raw check.

Bumber64: @lethosor
The ``isValid()`` check itself uses the magic number. It's all over plugins too. I'm not sure of the best place to define it, but it's probably out of the scope of this PR.

myk002:
You might need to merge latest develop to get the tests to pass

myk002:
It looks like your library/xml and scripts are still misaligned. Could you fix those up? You might have committed a change to which commit those submodules point to that you have to revert.

myk002 (commented):
Overall, code looks good. Thanks for getting that submodule thing figured out!

I'd be a lot more comfortable with this if it had some fortress-mode unit tests in https://github.com/DFHack/dfhack/blob/develop/test/gui.lua (or a new sibling file since that one is currently `config.mode = 'none'`)

Bumber64:
> Overall, code looks good. Thanks for getting that submodule thing figured out!
> 
> I'd be a lot more comfortable with this if it had some fortress-mode unit tests in https://github.com/DFHack/dfhack/blob/develop/test/gui.lua (or a new sibling file since that one is currently `config.mode = 'none'`)

Is this something that needs to be done before merging the PR? I don't have much experience with unit testing.

myk002:
It's really not hard. It's like regular coding. It's asking the question "what do I expect to happen when I call this function?" Then you call the function and check that it happened.

Go to your cmake settings and set BUILD_TESTS to install the test harness. It's just a regular script named "test".

In this case, when you run the `test` command, the test harness will load the save in region1 (if a save is not already loaded) and run all tests. Each of your tests should set the viewport to some initial state, run one of the new functions, and then test to make sure the viewport is now set as you expect. Since the functions may change behavior based on window size, the unit tests have to be flexible and test ranges, not exact values.

Here's an end-to-end guide:
1. Build with BUILD_TESTS=ON to install the test harness
2. Create a file named `test/gui_viewport.lua` with the following contents:
```
config.mode = 'fortress'

function test.zoom_to_unit()
  -- set viewport to some initial location

  -- run dfhack.gui. function

  -- test viewport has changed as we expect

end
```
3. ninja install (or however you usually build and install)
4. DFHack# test -t gui_viewport

Look at other tests in the test tree for examples. The `expect` functions used to verify your expectations are defined in https://github.com/DFHack/dfhack/blob/develop/library/lua/test_util/expect.lua

Bumber64:
Do we have a special region1 save for doing tests on?

myk002:
Yeah, you can download it from [here](https://drive.google.com/uc?export=download&id=1XvYngl-DFONiZ9SD9OC4B2Ooecu8rPFz), but really the tests should work on any fortress save. If they don't work on a save you already have available, it's probably a problem with the test.

Bumber64:
Thanks. Looks like a 1x1 embark, which might be inadequate for testing viewport scrolling.

myk002:
> Thanks. Looks like a 1x1 embark, which might be inadequate for testing viewport scrolling.

The "official" tests are run at 80x25, which does allow for scrolling even in a 1x1 (though the tests should still pass even if the dev who runs the tests happens to have such a large screen that the map doesn't actually scroll, but that could be the case no matter how large an embark we use). If you need a larger embark to get any value, though, I can definitely make that happen. I just used a 1x1 embark because it gave the smallest file size so our build runs faster.

myk002 (approved):

myk002:
Sorry for the long delay!

Bumber64:
> Sorry for the long delay!

FYI, I never made any progress on the unit tests. I went on hiatus due to selling house and moving out of state.

myk002:
I understand. I just didn't want the work you did to get lost, especially now that a new release is coming out

PR #2115, opened 2022-04-27
Add "partial-items" tweak.

quietust:
When active, the displayed names of partially-consumed items (e.g. hospital cloth) will display a percentage indicator at the end.

Because otherwise it's impossible to see why certain pieces of cloth or thread are being ignored for other jobs.

### DIFF
--- a/plugins/tweak/tweak.cpp
+++ b/plugins/tweak/tweak.cpp
@@ -49,11 +49,14 @@
 #include "df/item_glovesst.h"
 #include "df/item_shoesst.h"
 #include "df/item_pantsst.h"
+#include "df/item_drinkst.h"
+#include "df/item_globst.h"
 #include "df/item_liquid_miscst.h"
 #include "df/item_powder_miscst.h"
 #include "df/item_barst.h"
 #include "df/item_threadst.h"
 #include "df/item_clothst.h"
+#include "df/item_sheetst.h"
 #include "df/spatter.h"
 #include "df/layer_object.h"
 #include "df/reaction.h"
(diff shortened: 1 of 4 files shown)


plugins/tweak/tweak.cpp
```diff
@@ -329,9 +336,20 @@ DFhackCExport command_result plugin_init (color_ostream &out, std::vector <Plugi
 
     TWEAK_HOOK("nestbox-color", nestbox_color_hook, drawBuilding);
 
+    TWEAK_HOOK("partial-items", partial_items_hook_1, getItemDescription);
```
softmoth, line 339: I would prefer `partial_items_hook_bar`, `_drink`, `_glob`, etc., instead of `_1`, `_2`, `_3`.

plugins/tweak/tweaks/partial-items.h
```diff
@@ -0,0 +1,99 @@
+// When displaying the names of partially-consumed items, show the percentage remaining
+// Potentially useful for revealing why random pieces of cloth or thread aren't suitable for jobs
+
+struct partial_items_hook_1 : df::item_barst {
```
softmoth, line 4: I would prefer eliminating the copy and paste by adding another macro that can be called like

```cpp
// Defines partial_items_hook_bar and calls IMPLEMENT_VMETHOD_INTERPOSE() with it
DEFINE_PARTIAL_ITEM_TWEAK(bar, 150);
// ...
DEFINE_PARTIAL_ITEM_TWEAK(cloth, 1000);
```

softmoth (commented):
It looks useful. I have a few style suggestions which might improve maintainability.

plugins/tweak/tweak.cpp
myk002: +1 on this. I was thinking the same thing as I read the code.

myk002 (approved):

PR #2117, opened 2022-04-27
Update init files

myk002:
Update the default init files with the following changes:

**dfhack.init-example**
- add `gui/quickfort` keybinding (is Ctrl-Shift-Q a good choice?)
- remove keybinding for obsolete `gui/assign-rack`
- remove reference to obsolete `tweak military-training`
- add `tweak do-job-now`
- add `tweak partial-items`
- add `tweak fast-heat 100`
- replaced the `sc-script` example with documentation on how to name files so they load automatically without needing to use `sc-script`

**onLoad.init-example**
- removed (replaced with documentation added to `dfhack.init-example`)

**data/examples/init/onMapLoad_dreamfort.init**
- added more documentation, removed "cheaty" commands and commands now moved to `dfhack.init-example`

Should anything in `onMapLoad_dreamfort.init` be moved to a more "default" file?

### DIFF
--- a/data/examples/init/onMapLoad_dreamfort.init
+++ b/data/examples/init/onMapLoad_dreamfort.init
@@ -24,12 +24,6 @@ repeat -name autoMilkCreature -time 14 -timeUnits days -command [ workorder "{\"
 # Fulfill high-volume orders before slower once-daily orders
 repeat -name orders-sort -time 1 -timeUnits days -command [ orders sort ]
 
-# Reduce performance impact of temperature changes
-tweak fast-heat 100
-
-# Add a hotkey to the job menu to allow boosting the priority of the job
-tweak do-job-now
-
 # Don't let caravans bring barrels of blood and other useless liquids
 fix/blood-del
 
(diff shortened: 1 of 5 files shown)


softmoth (approved):
No comments, looks great.

myk002:
I also wonder about our current guidance to copy the `dfhack.init-example` file and make changes. Should we instead be creating a standard default baseline that is overwritten when you upgrade (e.g. `hack/data/init/dfhack.init`) and people can create their personal files in addition to (as opposed to the current "instead of") the defaults?

softmoth:
That does sound more friendly. Probably worth communicating with PeridexisErrant if the default changes are large?

lethosor (approved):
I can live with this. The argument for the `onLoad.init-example` when it was added was that people who haven't set up init files will still get some "default" behavior that requires a world to be loaded. warn-starving is the only one remaining that we had added to it. I expect the group of people that would be affected is very small.

lethosor:
pre-commit.ci run

myk002:
Should we have a default onLoad.init or onMapLoad.init file, owned by the distribution so we can overwrite it in new versions? What would we put in there to define the "default DFHack experience"? Maybe some of what's in the Dreamfort init example?

myk002:
Ok, I'll merge what we have here and move my questions to an "idea" bug

PR #2118, opened 2022-04-27
add new global function: ensure_key

myk002:
If the Lua table `t` doesn't include the specified `key`, `t[key]` is set to the value of `default_value`, which defaults to `{}` if not set. The new or existing value of `t[key]` is then returned.

### DIFF
--- a/docs/Lua API.rst
+++ b/docs/Lua API.rst
@@ -2643,6 +2643,12 @@ environment by the mandatory init file dfhack.lua:
   Walks a sequence of dereferences, which may be represented by numbers or strings.
   Returns *nil* if any of obj or indices is *nil*, or a numeric index is out of array bounds.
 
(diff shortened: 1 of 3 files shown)


lethosor:
Could this go in utils.lua? I think there are some table-related functions in there already.

myk002:
> Could this go in utils.lua? I think there are some table-related functions in there already.

Well, maybe. I could go either way. I didn't put it in utils because the "table" related functions there seemed geared more towards DF vectors, and I wanted to make it clear that this function is not meant for vectors. However, that restriction is documented, so it's not a very strong point.

lethosor:
We're not very consistent, so I'm ok either way. And now that I've expanded the diff, I see `safe_pairs()` right above this, and `copyall()` farther up, so I'm less hesitant to put it here than I was earlier.

lethosor (approved):

PR #2122, opened 2022-05-01
Adds ccache to build workflow

cppcooper:
I tested on my fork with PR https://github.com/cppcooper/dfhack/pull/3

I added this commit and pushed, opened the PR, added a (temporary) document change commit and pushed. The second commit's build steps were seconds instead of minutes. So this works as far as I tested.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install \
+            ccache \
             libgtk2.0-0 \
             libncursesw5 \
             libsdl-image1.2-dev \
@@ -38,8 +39,7 @@ jobs:
             libxml-libxslt-perl \
             lua5.3 \
             ninja-build \
-            zlib1g-dev \
-            ccache
+            zlib1g-dev
         pip install 'sphinx<4.4.0'
     - name: Install GCC
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -60,6 +61,11 @@ jobs:
       with:
         path: ~/DF
         key: ${{ steps.env_setup.outputs.df_version }}
+    - name: Fetch ccache
+      uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: ${{ matrix.gcc }}
```
lethosor, line 68: Could this also include `${{ matrix.os }}`? That way we would use a clean cache if we move the GCC 7 build to a newer Ubuntu version, for instance.

More generally, I would tend towards more verbose keys to ensure that they are unique across all other caches in this pipeline, but that is probably not important for now. (One example: `ccache-${{ matrix.os }}-gcc-${{ matrix.gcc }}`)

lethosor:
From Discord: looks like 20-30sec build times (with an existing cache) and a cache size of 12-15MB. Impressive results!

.github/workflows/build.yml
```diff
@@ -38,7 +38,8 @@ jobs:
             libxml-libxslt-perl \
             lua5.3 \
             ninja-build \
-            zlib1g-dev
+            zlib1g-dev \
```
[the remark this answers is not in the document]
cppcooper: moved

.github/workflows/build.yml
cppcooper: copy pasta'd

myk002:
The cache will grow over time as old stuff becomes useless but is still cargo culted. Can we limit the size of the cache to working set size times the number of branches that it is building simultaneously? Let's say maybe 15mb * 50 = 750 mb ?

Edit: actually, ccache has a default limit of 1gb, which is close enough. No changes needed, I think

myk002:
Alright, let's ship it and see how it goes!

PR #2134, opened 2022-05-09
fix wrong `Label.frame_body.x2` value

TymurGubayev:
`update_scroll_inset` might change `frame_inset`, i.e. we need to `computeFrame` with the new values.

### DIFF
--- a/test/library/gui/widgets.Label.lua
+++ b/test/library/gui/widgets.Label.lua
@@ -1,8 +1,5 @@
 -- test -dhack/scripts/devel/tests -twidgets.Label
 
---reload('gui')
---reload('gui.widgets')
-
 local gui = require('gui')
 local widgets = require('gui.widgets')
 
@@ -22,7 +19,7 @@ function test.Label_correct_frame_body_with_scroll_icons()
         focus_path = 'test-framed-screen',
     }
 
-    t = {}
+    local t = {}
     for i = 1, 12 do
         t[#t+1] = tostring(i)
         t[#t+1] = NEWLINE
@@ -57,7 +54,7 @@ function test.Label_correct_frame_body_with_few_text_lines()
         focus_path = 'test-framed-screen',
     }
 
-    t = {}
+    local t = {}
     for i = 1, 10 do
         t[#t+1] = tostring(i)
         t[#t+1] = NEWLINE
@@ -92,7 +89,7 @@ function test.Label_correct_frame_body_without_show_scroll_icons()
         focus_path = 'test-framed-screen',
     }
 
-    t = {}
+    local t = {}
     for i = 1, 12 do
         t[#t+1] = tostring(i)
         t[#t+1] = NEWLINE
--- a/test/library/gui/widgets.Label.lua
+++ b/test/library/gui/widgets.Label.lua
@@ -8,17 +8,17 @@ local wait = function(n)
     delay(n or 30) -- enable for debugging the tests
 end
 
-function test.Label_correct_frame_body_with_scroll_icons()
-    local fs = defclass(fs, gui.FramedScreen)
-    fs.ATTRS = {
-        frame_style = gui.GREY_LINE_FRAME,
-        frame_title = 'TestFramedScreen',
-        frame_width = 10,
-        frame_height = 10,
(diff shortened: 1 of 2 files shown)


TymurGubayev:
> How hard would it be to write a unit test for this?

I think it's possible, I'll try to write something in the next few days.

TymurGubayev:
added a few tests.

test/library/gui/widgets.Label.lua
```diff
@@ -0,0 +1,117 @@
+-- test -dhack/scripts/devel/tests -twidgets.Label
+
+--reload('gui')
+--reload('gui.widgets')
```
myk002, line 4: this should not be necessary. the test harness loads all lua libraries cleanly before starting the tests.

test/library/gui/widgets.Label.lua
```diff
@@ -0,0 +1,117 @@
...
+    fs.ATTRS = {
+        frame_style = gui.GREY_LINE_FRAME,
+        frame_title = 'TestFramedScreen',
+        frame_width = 10,
+        frame_height = 10,
+        frame_inset = 0,
+        focus_path = 'test-framed-screen',
+    }
+
+    t = {}
```
myk002, line 25: missing `local`? Also in the other tests.

test/library/gui/widgets.Label.lua
```diff
@@ -0,0 +1,117 @@
...
+function test.Label_correct_frame_body_with_scroll_icons()
+    local fs = defclass(fs, gui.FramedScreen)
+    fs.ATTRS = {
+        frame_style = gui.GREY_LINE_FRAME,
+        frame_title = 'TestFramedScreen',
+        frame_width = 10,
+        frame_height = 10,
+        frame_inset = 0,
+        focus_path = 'test-framed-screen',
+    }
```
myk002, line 23: You can declare this class outside of the function so you can reuse it across tests.

test/library/gui/widgets.Label.lua
```diff
@@ -0,0 +1,117 @@
+-- test -dhack/scripts/devel/tests -twidgets.Label
```
myk002, line 1: There is actually a `test/library/gui/widgets.lua` file that you can add these tests to now. It was created relatively recently, so you might have to merge latest develop to see it.

myk002 (changes requested):
The tests are good to have, but I notice that they don't actually test the change being made here -- the tests pass just fine if the change to `widgets.lua` is reverted.

TymurGubayev:
> The tests are good to have, but I notice that they don't actually test the change being made here -- the tests pass just fine if the change to `widgets.lua` is reverted.

huh, that's strange: the first one (`Label_correct_frame_body_with_scroll_icons`) fails for me without the changes.

test/library/gui/widgets.Label.lua
TymurGubayev: yep, wasn't there last time I checked. But maybe separate widget test files aren't bad? I imagine should every widget get tests, the single file with all the tests will be huge.

myk002:
> > The tests are good to have, but I notice that they don't actually test the change being made here -- the tests pass just fine if the change to `widgets.lua` is reverted.
> 
> huh, that's strange: the first one (`Label_correct_frame_body_with_scroll_icons`) fails for me without the changes.

I must have been in an inconsistent state. I'll check through again on Monday. Sorry!

test/library/gui/widgets.Label.lua
lethosor: I'm ok with separate per-widget tests

lethosor:
I can confirm that one of the three tests fails for me with the implementation taken from `develop` (the other two tests pass with either implementation):

```
Check failed! Label's frame_body.x2 and .width should be one smaller because of show_scroll_icons.: "10" ~= "9"
  at .../test/library/gui/widgets.Label.lua:42

test failed: library/gui/widgets.Label:Label_correct_frame_body_with_scroll_icons
```

I did forget to re-run `ninja install` at first - maybe double-check that you have the right files installed in `hack/lua`?

myk002 (approved):

PR #2153, opened 2022-05-15
Enable fortress tests in CI

### LINKED ISSUE #2132: Get fortress mode tests running in ci
Need to add a prepared fortress save to the df image used by CI. There is a save ready [here](https://drive.google.com/file/d/1XvYngl-DFONiZ9SD9OC4B2Ooecu8rPFz/view?usp=drivesdk) but we need to get it into the image.

IIRC, the plan of record is to add the compressed save file to the dfhack-bin repo, clone it as part of the build action, and decompress it into place.

Then we can remove the `mode` restriction in run_tests.py so fortress tests are run.

Edit: the save can be added to the df cache by modifying https://github.com/DFHack/dfhack/blob/develop/ci/download-df.sh

Do we want to download the binary with wget or clone/pull the dfhack-bin repo?

myk002:
Fixes #2132 

Downloads the test save from Google Drive for now. Eventually I think we should upload the file to somewhere within the DFHack org and download it from there, but this will work for now.

Also enables fortress tests in `ci/run_tests.py` since now we'll have the test save that we can run the tests in.


myk002:
fortress mode tests work! one is failing, though. I suspect this is due to the previous test not restoring state properly. This might get fixed once we merge #2147 since the EditField behavior will be better defined.

myk002:
Note we might not want to merge this until after we have the capability to pull specific PRs in other repos. Otherwise we might have to ignore build failures for DFHack/dfhack PRs that require coordinated changes in DFHack/scripts.

ci/download-df.sh
```diff
@@ -16,40 +17,48 @@ cd "$DF_FOLDER/.."
...
 
 if [ ! -f receipt ]; then
-    rm -f "$tardest"
+    rm -f "$df_tardest" "$save_tardest"
     minor=$(echo "$DF_VERSION" | cut -d. -f2)
     patch=$(echo "$DF_VERSION" | cut -d. -f3)
-    url="http://www.bay12games.com/dwarves/df_${minor}_${patch}_linux.tar.bz2"
-    echo Downloading
+    echo "Downloading DF $DF_VERSION"
     while read url; do
```
cppcooper, line 32: I must be missing something, the url isn't initialized anymore.. or is it?

cppcooper (commented):
So is it the change in run-tests.py that enables fortress tests? I'm very interested to understand how we can run tests under fortress mode.

ci/download-df.sh
myk002: > So is it the change in run-tests.py that enables fortress tests? I'm very interested to understand how we can run tests under fortress mode.

fortress mode tests have been possible for some time now. they're enabled by default when running `test` locally. If a test has `config.mode = 'fortress'` then the test harness will make sure there is a fort loaded before running the test.

The change to `run-tests.py` removes the parameter I put in to skip fortress-mode tests for CI, since the CI environment didn't have a fort save to use. This PR provides that fort save.

myk002: > I must be missing something, the url isn't initialized anymore.. or is it?

the `url` variable on line 30 was never used. It's probably left over from a previous implementation before the while loop was added. the `url` on line 32 is initialized by the while loop.

cppcooper: > `url` on line 32 is initialized by the while loop

ah, maybe I should bother myself to learn basic python =S seemed like it was parsing the url

> If a test has `config.mode = 'fortress'`

cool, thanks

myk002: yeah, it can be difficult to read. note that the `download-df.sh` file is shell script (bash), though, not python.

myk002:
Need to look into that failing test. I can't seem to reproduce it locally. Perhaps it is due to screen size? The message it's looking for shouldn't be affected by screen size (it doesn't disappear when the screen gets small, and is within the top 25 lines), but I'll check.

myk002:
ok, test failure reproduced and fixed. it was indeed caused by the small screen size. I forgot that I made the name field tooltip optional as long as no name conflicts were detected.

I can merge this now to enable the tests and see how it goes. If it causes too much trouble with PRs that are linked across repos, we can walk back the change to `run_tests.py`.

Issue #2155, opened 2022-05-15
Manager: give a warning/confirmation before a Work Order is removed


joeljpa:
Per title. 

Using 0.47.05-r5. 

While I've seen this UI great improvement done to remove uniforms (for example), I was surprised it wasn't there for removing a Work Order at the Manager screen. Kind of new using this awesome mod which makes our lives so much easier so hope I'm not missing something.

myk002:
+1 on this. 'r' is just too easy to hit. I'm not sure how hard the key is to intercept, but the functionality would fit into the `confirm` plugin.

myk002:
There are times when you need to remove a bunch of orders, though. Perhaps there should be a "pause confirmation for 5 minutes" option when the prompt comes up.

lethosor:
There's a pretty straightforward way to disable, but re-enabling at the moment requires pulling up the settings screen or using the console.

Maybe "pause until this screen is closed" could be a reasonable option, since most bulk workflows like you suggested would keep the same screen open.

softmoth:
I remap it to capital R, it solves the problem well for me.

myk002:
That's a really good idea. I might do that too (maybe in addition to the confirmation prompt requested here)

joeljpa:
I remove orders often also and can't imagine the confirmation being that much of a hurdle for the player.  I think it's worth the minor inconvenience versus accidentally removing a Work Order and having no clue what I just removed from my list of ~10 unrelated Work Orders (true story).

PR #2164, opened 2022-05-24
Allow player to pause the confirmation dialog without disabling

myk002:
Hit 'p' to pause confirmation until the current page/sidebar is exited.
As proposed in #2155

adds the "p to pause" hotkey help text to the main dialog.
I would have added it to the bottom bar, but just writing `p: Pause` isn't exactly explaining what will happen when you hit `p`. Adding more help text to the bottom bar makes the dialog unacceptably wide.

Screenshot:

image

Hitting `p` pauses and closes the dialog without confirming the action. The user has to invoke the action again, and it will be let through without the confirm dialog showing up.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Misc Improvements
 - `confirm`: added a confirmation dialog for removing manager orders
+- `confirm`: allow players to pause the confirmation dialog until they exit the current screen
 - `dfhack-examples-guide`: refine food preparation orders and fix conditions for making jugs and pots in the ``basic`` manager orders
 
 ## Documentation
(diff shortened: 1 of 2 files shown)


myk002:
Feedback from discord:
- use standard hotkey formatting for the pause message. Red "p", then colon, then space text
- add paused status and hotkey to unpause to gui/confirm-opts
- use SC_VIEWSCREEN_CHANGED to detect when we should unpause. Track the screen pointer where we paused and unpause when that screen is no longer on the stack

myk002:
@lethosor changes are implemented. I could not go the pure viewscreen transition detection route for figuring out when to unpause since some confirm handlers intercept input on viewscreen_dwarfmodest. On those screens, I fall back to ESC detection. This is not perfect. You can, for example, trigger unpause early by trying to remove your trade depot, pausing confirm, hitting ctrl-N to name your depot, hitting ESC (which is a noop in this mode), then hitting ENTER to exit rename mode. confirm will then be unpaused. I could *potentially* handle this corner case by checking whether we're in rename mode when esc is hit, but it feels like such an edge case that it might not be worth the extra code.

PR #2169, opened 2022-05-28
MiscUtils word_wrap: Add option to trim only leading whitespace after wrapping

Bumber64:
Replace ``bool collapse_whitespace`` option with an enum. Options ``WSMODE_KEEP_ALL``, ``WSMODE_COLLAPSE_ALL``, and ``WSMODE_TRIM_LEADING``.

Related to https://github.com/DFHack/dfhack/pull/2113

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -47,8 +47,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
-- ``word_wrap``: argument ``bool collapse_whitespace`` converted to enum ``word_wrap_wspace_mode mode``,
-  with valid modes ``WSMODE_KEEP_ALL``, ``WSMODE_COLLAPSE_ALL``, and ``WSMODE_TRIM_LEADING``.
+- ``word_wrap``: argument ``bool collapse_whitespace`` converted to enum ``word_wrap_wspace_mode mode``, with valid modes ``WSMODE_KEEP_ALL``, ``WSMODE_COLLAPSE_ALL``, and ``WSMODE_TRIM_LEADING``.
 
 ## Lua
 - ``widgets.HotkeyLabel``: the ``key_sep`` string is now configurable
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -47,7 +47,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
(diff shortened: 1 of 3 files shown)


Bumber64:
``trim_whitespace_after_newline`` is a bit wordy. Any better suggestions?

docs/changelog.txt
```diff
@@ -47,6 +47,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
+- `word_wrap`: add ``trim_whitespace_after_newline`` option, which trims any whitespace at the beginning of a line after wrapping
```
Bumber64, line 50: This causes ``docs/_auto/news.rst:29: WARNING: undefined label: word_wrap``

myk002: yeah, it's not a document label. I'd suggest double backticks.

library/MiscUtils.cpp
```diff
@@ -169,14 +169,15 @@ std::string to_search_normalized(const std::string &str)
 }
 
 
-bool word_wrap(std::vector<std::string> *out, const std::string &str,
-               size_t line_length, bool collapse_whitespace)
+bool word_wrap(std::vector<std::string> *out, const std::string &str, size_t line_length,
+               bool collapse_whitespace, bool trim_whitespace_after_newline)
```
myk002, line 173: since this is mutually exclusive with the collapse_whitespace boolean, maybe it would be better expressed as an enum?

Bumber64: I made an attempt. Let me know what I need to fix.

myk002 (approved):

library/MiscUtils.cpp
myk002: `word_wrap` itself is not in a namespace, so I think it's fine how you have it. The name of the enum and its values are unique enough that I don't think there is much danger of collision.

library/include/MiscUtils.h
```diff
@@ -389,10 +389,16 @@ DFHACK_EXPORT std::string toUpper(const std::string &str);
 DFHACK_EXPORT std::string toLower(const std::string &str);
 DFHACK_EXPORT std::string to_search_normalized(const std::string &str);
 
+enum word_wrap_wspace_mode {
```
lethosor, line 392: I suggest `whitespace` over `wspace`, but LGTM either way

lethosor (approved):

PR #2173, opened 2022-05-30
Maps.cpp: Convert utility fns from anon namespace to static fns

Bumber64:


Bumber64:
Do we need a changelog.txt entry for this? It shouldn't affect anything outwardly, AFAIK.

myk002 (approved):
No changelog needed for stuff that does not change the APIs and is not user visible

lethosor:
Is this something we want to standardize on for some reason? I'm not sure if I missed a discussion on Discord/Bay12. We have a number of anonymous namespaces, and I'm not seeing a clear disadvantage to them, although the only advantage I've seen mentioned is "you can also declare translation-unit-local types in them".

myk002:
I made a comment in another PR that we *seem* to prefer statics, though it's not a very strong preference.

lethosor:
Ah: https://github.com/DFHack/dfhack/pull/2113#discussion_r884473782

For new code, I'd probably personally prefer `static` just because I'm familiar with it. I wouldn't worry too much about changing existing code, since there _may_ have been a reason for choosing anonymous namespaces instead (actually, @ab9rf could confirm in this case since it was added in 59d70d6687). Here, it just affects two functions, so as long as my understanding is correct, there should be no functional difference between either approach.

myk002:
I also have a slight preference for `static`, if only because it reduces the indentation level of the code. My understanding is also that there is no functional difference. This usage of `static` was on track to be deprecated at one point, but apparently they walked that deprecation back in a later revision.

ab9rf:
I don't remember why I used an anonymous namespace specifically. Most likely, it was because these were functions that are not supposed to be used outside the module, and at that time the Received Wisdom was that the proper thing to do in such cases was to enclose them in an anonymous namespace.

It seems that the C++ community is still hotly divided over this. My personal preference is to an anonymous namespace, in part because the ISO CPP Core Guidelines currently state the global use of `static` is discouraged. See also https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities

ab9rf (commented):
The entire purpose of this PR is to convert nonexported functions from enclosure in an anonymous namespace to global static. This contravenes ISO CPP Core Guideline SF.22.

So I must ask: Why?

myk002:
We can revert if this is controversial. I had suggested the change from anonymous namespaces to static functions because we seem to use both and I had a slight preference for static (for the reduced indentation), but my preference is not a strong one.

ab9rf:
I don't have a strong enough preference to argue for reverting this change, but I would caution against further changes of this sort.

This issue was discussed on the isocpp repository about a year ago. Reportedly, the issue was escalated to their editorial committee, which made the call to discourage the use of global static. See isocpp/CppCoreGuidelines#1727.

myk002:
That's perfectly fair. I can promote anonymous namespaces in future PRs.

PR #2175, opened 2022-06-02
[gui/quantum] add integration test

myk002:
Depends on DFHack/scripts#388

uses the quickfort ecosystem test harness to run an end-to-end integration test for `gui/quantum`


test/quickfort/ecosystem.lua
```diff
@@ -345,10 +371,106 @@ function test.end_to_end()
...
 end
+
+local function send_keys(...)
+    local keys = {...}
+    for _,key in ipairs(keys) do
+        gui.simulateInput(dfhack.gui.getCurViewscreen(true), key)
+    end
+end
+
+function test_gui_quantum(pos)
```
lethosor, line 390: In the future, could this be split out into another file? If it's here because there are useful utilities in this file, we could move them to test_util

lethosor: (This is assuming that quantum is really a "separate" script, as opposed to one that's only called by quickfort)

myk002: It's here because it uses quickfort to implement its core logic, and we can verify it the same way we verify quickfort itself - by taking a blueprint snapshot and comparing.

It also needs the ecosystem framework to set up the feeder stockpile it needs to test itself. In short, we *could* put this test in a different file, but all the supporting logic in this file would have to get factored out.

I just realized there may be confusion here. This is not a function called directly by the test harness. It's `test_`, not `test.`. it's called by the `extra_fn` option in the ecosystem test.

lethosor: Oh, yeah, that's why I was confused. Thanks!

PR #2181, opened 2022-06-04
Remove old modules

lethosor:
Notes and Windows both appear to be unused.

I left Graphic alone for now, since it looks like it _might_ do something useful, but I also think it only works on Windows (and is close on macOS).

(please don't squash - would like to maintain the ability to revert as needed)


quietust:
I think we can also get rid of the `Engravings` module (doesn't appear to be used at all) and a small part of the `Constructions` module (specifically the `copyConstruction` function).

The `Kitchen` module could probably also be migrated entirely into the `seedwatch` plugin, since it's really the only one that uses it (aside from one of the Tweaks).

myk002:
Can we also get rid of Core::TileUpdate and hooks-egg.cpp (and the related cmake options)?

I've never been quite clear on how that file makes DF any more strangely egg-shaped..

lethosor:
Yeah, Engravings looks to be in a similar situation as Notes, where it was only really useful when DFHack was out-of-process. It's redundant now that tools can access the engravings list directly. Same goes for Constructions - `t_construction` is no longer needed (and `copyConstruction` with it, like you said).

I think the Kitchen module is useful enough to keep around. It abstracts away some complexity, and if we moved it out of the core library, other tools wouldn't be able to make use of it. (It would probably be more accessible if it were exposed to Lua, but I'm not going to worry about that for now.)

Eggy hooks are on my list too. There are considerably more references that I'd need to clean up, which is the only reason I haven't done it yet.

ab9rf:
I agree with @lethosor regarding the Kitchen module. At the moment, seedwatch is the only thing using it, but there's an open issue on improving `seedwatch` and/or possibly merging it with `autofarm`, which could result in that functionality being used by another plug-in. And someone might well write something new; I've long had an idea about trying to write something that will try to cook foods that dwarfs want to eat and that might involve manipulating the kitchen settings.

Also, I seem to recall that Toady has made noises about changing how food works (although don't ask me why I think that) and thus having the abstraction in place may make updating if and when he makes such changes earlier.

ab9rf (approved):
No objections.

myk002:
Let me split off issues for removing Constructions, Engravings, and eggy so we can merge the work done so far.

At one point, I considered implementing a `getEngravingsMap()` in the Engravings module that would return an O(1) lookup table for engravings, but the lookup table would only be guaranteed consistent until the game unpaused (since we don't have a reliable way of knowing when the engravings vector has been changed). Also, the only place I needed it was in Lua scripts, so I just implemented it [there](https://github.com/DFHack/scripts/blob/master/internal/quickfort/dig.lua#L599-L604).

myk002:
Needs rebasing into current HEAD

myk002 (approved):

lethosor:
Going to take another pass through modules for obviously deprecated functions before merging. That should give more than enough time for CI to pass and for the Stonesense update to make it into develop.

myk002:
As long as you're removing old cruft, what about BUILD_DEVEL?

myk002:
Yeah, it looks like that was part of a workflow for building out-of-tree plugins.

PR #2182, opened 2022-06-05
GCC12 build fixes

cppcooper:
I was having issues building on gcc12, and gcc11 (11.3) as it turned out. These are the modifications I used to obtain a complete build.

### DIFF
--- a/plugins/debug.cpp
+++ b/plugins/debug.cpp
@@ -511,6 +511,8 @@ private:
     DebugManager::categorySignal_t::Connection connection_;
 };
 
+constexpr const char* FilterManager::configPath;
+
 FilterManager::~FilterManager()
 {
 }
--- a/plugins/embark-assistant/matcher.cpp
+++ b/plugins/embark-assistant/matcher.cpp
@@ -476,7 +476,8 @@ namespace embark_assist {
             }
 
             if (!survey_results->at(fetch_x).at(fetch_y).surveyed) {
-                // todo: this has never been executed before /*std::cerr << hello, is anybody out there*/
+                // If the data has been collected, incursion processing should be performed to evaluate whether a match actually is present.
+                // but if it hasn't we need to return with a failed_match
                 *failed_match = true;
                 return;
             }
--- a/plugins/embark-assistant/matcher.cpp
+++ b/plugins/embark-assistant/matcher.cpp
@@ -475,7 +475,8 @@ namespace embark_assist {
                 return;  //  We're at the world edge, so no incursions from the outside.
             }
 
-            if (!&survey_results->at(fetch_x).at(fetch_y).surveyed) {
+            if (!survey_results->at(fetch_x).at(fetch_y).surveyed) {
+                // todo: this has never been executed before /*std::cerr << hello, is anybody out there*/
                 *failed_match = true;
                 return;
             }
--- a/plugins/prospector.cpp
+++ b/plugins/prospector.cpp
@@ -85,10 +85,10 @@ bool operator>(const matdata & q1, const matdata & q2)
     return q1.count > q2.count;
 }
 
-template<typename _Tp = matdata >
-struct shallower : public binary_function<_Tp, _Tp, bool>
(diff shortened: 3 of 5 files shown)


plugins/debug.cpp
```diff
@@ -511,6 +511,8 @@ struct FilterManager : public std::map<size_t, Filter>
     DebugManager::categorySignal_t::Connection connection_;
 };
 
+constexpr const char* FilterManager::configPath;
```
lethosor, line 514: TODO: verify that the initializer in the class above still runs

```c++
    constexpr static const char* configPath{"dfhack-config/runtime-debug.json"};
```

plugins/embark-assistant/matcher.cpp
```diff
@@ -475,7 +475,8 @@ namespace embark_assist {
                 return;  //  We're at the world edge, so no incursions from the outside.
             }
 
-            if (!&survey_results->at(fetch_x).at(fetch_y).surveyed) {
+            if (!survey_results->at(fetch_x).at(fetch_y).surveyed) {
```
myk002, line 478: Was this testing the memory address of the surveyed boolean? What was it supposed to catch?

cppcooper: yea, don't know. I've presumed the `&` was a typo

plugins/debug.cpp
lethosor: Pretty sure it's fine based on https://github.com/DFHack/dfhack/blob/7b2cb8f9a631b8b44bfc2ff1f4e684d34b18b5f6/library/modules/MapCache.cpp#L82 and https://github.com/DFHack/dfhack/blob/7b2cb8f9a631b8b44bfc2ff1f4e684d34b18b5f6/library/include/modules/MapCache.h#L58

plugins/embark-assistant/matcher.cpp
lethosor: @PatrikLundell would you mind confirming that the new logic is correct?

PatrikLundell: It wasn't a typo, but rather poor knowledge of High Level Assembly, a.k.a. C(++). The intention was to check the field, not any address.

Checking the address would never fail, as things would blow up with a segfault before it reached it, as far as I can see.

cppcooper: > The intention was to check the field

Ok. Same outcome as if it had been a typo, off by one extra character.

cppcooper: Oh, one thing I'm still curious about though.. why haven't we seen a problem with whatever error this was intended to prevent?

lethosor: I should clarify - my question is whether the newly-reachable branch (setting `*failed_match = true;`) _should_ be reachable if `.surveyed` is false, or if `.surveyed` is true or something else.

PatrikLundell: Yes. If no survey has been performed the data needed for incursion processing hasn't been collected, and so the code should return with a failed match. If the data has been collected, incursion processing should be performed to evaluate whether a match actually is present.
Performing the incursion processing without (a complete set of) incursion data will result in some false positives and some false negatives, partially depending on the query (it's probably possible to make a query that gets both kinds of false results).

plugins/embark-assistant/matcher.cpp
```diff
@@ -475,7 +475,8 @@ namespace embark_assist {
...
+                // todo: this has never been executed before /*std::cerr << hello, is anybody out there*/
```
myk002, line 479: this comment can be taken out

myk002 (approved):

plugins/embark-assistant/matcher.cpp
cppcooper: I replaced the comment with a better one

myk002 (approved):

PR #2184, opened 2022-06-06
ensure refs are cleaned up when we remove a job

### LINKED ISSUE #2179: Build-now fails on buildings whose items are part of a job
It used to be that jobs.removeJob() would clean up the specific refs in its items. Now that we call the new hook that toady provided, it no longer does so.

I think we need to augment the toady hook with our own cleanup. Looping through job.items and calling jobs.disconnectJobItem() should do it. This works if called from build-now, but should really be moved into the jobs module.

myk002:
Fixes #2179

The new df-provided `cancel_job()` doesn't disassociate items from the job for us whereas the old custom implementation did.
ref: #2028

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,6 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 - ``widgets.CycleHotkeyLabel``: allow initial option values to be specified as an index instead of an option value
+- ``job.removeJob()``: fixes regression in DFHack 0.47.05-r5 where items attached to the job were not disassociated from the job when the job is removed. This allows `build-now` to build buildings whose items are already associated with the job again.
 
 ## Misc Improvements
 - `confirm`: added a confirmation dialog for removing manager orders
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,7 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 - ``widgets.CycleHotkeyLabel``: allow initial option values to be specified as an index instead of an option value
-- ``job.removeJob()``: fixes regression in DFHack 0.47.05-r5 where items attached to the job were not disassociated from the job when the job is removed. This allows `build-now` to build buildings whose items are already associated with the job again.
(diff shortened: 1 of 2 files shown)


lethosor (commented):
Fixes levers for me, and doesn't seem to leave behind any dangling refs.

However, I think we also need to restore at least the functionality that calls `disconnectJobGeneralRef` for certain types of jobs. Right now, jobs with a worker assigned don't appear possible to cancel. This can be reproduced pretty easily with `gui/mass-remove`'s "unremove building" mode - it won't work if a removal job has a unit assigned, and I'm pretty sure it used to work.

Old logic for reference: https://github.com/DFHack/dfhack/blob/cf54b84fd631d0858b4974f1f3d44b7aac3ac0b0/library/modules/Job.cpp#L372-L390

myk002:
> However, I think we also need to restore at least the functionality that calls `disconnectJobGeneralRef`

Done.

ab9rf (approved):
Interesting that this is required, but I'm not terribly surprised. I thought the code I found in disassembly handled this case, but I might easily have misinterpreted something. ghidra doesn't exactly decompile stuff involving `std::vector` into the most readable code.

myk002:
The cleanups appear to be required for both DFHack-initialized jobs and game-created jobs. My guess is that we're simply hooking into a slightly lower layer that expects a certain amount of prepossessing.

Thanks for the review!

ab9rf:
I'm going to have to go back to the disassembly and figure that out, because it looked to me like it was doing everything our code was doing plus a few other things it wasn't. But that will definitely take a while and might never yield useful results, and there's no reason to hold off on a fix while I do that.

ab9rf:
I just looked at the disassembly again and Toady's code does have code to handle both `UNIT_WORKER` and `BUILDING_HOLDER` general references. His code appears to do more than what this code does, but also has conditions that this code doesn't check. For example, the code for `UNIT_WORKER` references clears some pathfinding related variables, and the code for `BUILDING_HOLDER` has stuff to update any open UI screens.

I therefore have nagging concerns that by clearing the references before they get passed to Toady's code, some things might get skipped, which could lead to leaks or other undefined behaviors. This issue deserves further investigation.

(For the interested, this is at `0x140846bc0` in the Win64 version. I haven't located the Win32 or Linux equivalents.)

PR #2186, opened 2022-06-06
namespace the ccache key with the repo and branch

myk002:
as discussed on discord

this builds on https://github.com/DFHack/dfhack/commit/46f80bed3daeb105e6350a6ef12854447c245663 by adding the repo owner and branch name to the key. this allows the ccache that we build on to be the one most relevant to the branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,8 +65,9 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/.ccache
-        key: ccache-${{ matrix.os }}-gcc-${{ matrix.gcc }}-${{ github.sha }}
+        key: ccache-${{ matrix.os }}-gcc-${{ matrix.gcc }}-${{ github.repository_owner }}-${{ github.ref_name }}-${{ github.sha }}
         restore-keys: |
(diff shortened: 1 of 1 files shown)


myk002:
ok, tested by re-running the gcc 7 env. the build is back down to ~40s instead of 6m. yay!

lethosor (approved):
Strictly speaking, we could get by without the `-v2` component, since the cache format hasn't changed and the cache will always be saved under the key specified by `key`, which is more specific now.

I'm also not sure we need the `repository_owner` component, because caches are local to the repository where the workflow runs.

myk002:
Ah, ok, I'll remove the repo owner if the cache is already namespaced that way.

The "v2" we can remove once the existing cache phases out (~7 days). For now, we need to change the name or else the following will occur:
1) new commit fetches cache. No hit on key so it checks restore-keys. Exact match on the base key (i.e. our existing, out of date key) so it uses that
2) every new commit then starts from the out of date cache, requiring a full rebuild every single time and making the cache useless (unless we're rebuilding the same commit twice, where it would match its own full key)

With the v2, there is never an exact match on the base key, and the most recent build that matches that prefix is used instead. This is the behavior we want.

PR #2193, opened 2022-06-08
ensure our docs build cleanly with sphinx-5

myk002:
I installed sphinx-5.0.1 locally to get a preview of what's to come. It displayed a few warnings that looked legit:

- We should explicitly set a language for the docs (it defaults to 'en', but we should make that explicit)
- extlinks should have '%s' in the comment expansion text

Locally-generated docs still look good with the changes in this PR. I'll also check against the build action output.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -201,16 +201,16 @@ extensions = [
 # short alias names to a base URL and a prefix.
 # See http://sphinx-doc.org/ext/extlinks.html
 extlinks = {
-    'wiki': ('https://dwarffortresswiki.org/%s', ''),
+    'wiki': ('https://dwarffortresswiki.org/%s', '%s'),
     'forums': ('http://www.bay12forums.com/smf/index.php?topic=%s',
-               'Bay12 forums thread '),
-    'dffd': ('https://dffd.bay12games.com/file.php?id=%s', 'DFFD file '),
+               'Bay12 forums thread %s'),
+    'dffd': ('https://dffd.bay12games.com/file.php?id=%s', 'DFFD file %s'),
     'bug': ('https://www.bay12games.com/dwarves/mantisbt/view.php?id=%s',
-            'Bug '),
-    'source': ('https://github.com/DFHack/dfhack/tree/develop/%s', ''),
-    'source-scripts': ('https://github.com/DFHack/scripts/tree/master/%s', ''),
-    'issue': ('https://github.com/DFHack/dfhack/issues/%s', 'Issue '),
-    'commit': ('https://github.com/DFHack/dfhack/commit/%s', 'Commit '),
+            'Bug %s'),
+    'source': ('https://github.com/DFHack/dfhack/tree/develop/%s', '%s'),
+    'source-scripts': ('https://github.com/DFHack/scripts/tree/master/%s', '%s'),
+    'issue': ('https://github.com/DFHack/dfhack/issues/%s', 'Issue %s'),
+    'commit': ('https://github.com/DFHack/dfhack/commit/%s', 'Commit %s'),
 }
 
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -201,16 +201,16 @@ def all_keybinds_documented():
 # short alias names to a base URL and a prefix.
 # See http://sphinx-doc.org/ext/extlinks.html
 extlinks = {
-    'wiki': ('https://dwarffortresswiki.org/%s', ''),
+    'wiki': ('https://dwarffortresswiki.org/%s', '%s'),
```
lethosor, line 204: This appears to be a breaking change. Sphinx 1.8-built docs (like the ones [built for this PR](https://dfhack--2193.org.readthedocs.build/en/2193/docs/_auto/fix.html)) don't interpolate the `%s` and look like this:

image

I'm confused exactly when this was introduced. Based on https://github.com/sphinx-doc/sphinx/pull/8898, it sounds like there should have been a warning before the breaking change, but I don't recall seeing one. That issue is also marked with the `4.1.0` milestone, but is linked as a breaking change in 4.0.0 at https://www.sphinx-doc.org/en/master/changes.html (and I've definitely built successfully with 4.x, most recently 4.5.0).

Anyway, we can probably add the `%s` programmatically ourselves, if really needed. Until we decide that we need to take advantage of it not always being at the end of the string...

myk002: Yes, it seems like I only checked extlinks that came with explicit captions (e.g.
```
:dffd:`mycaption <12345>`
```
The extlinks that use the default caption are indeed broken with <sphinx-5. It looks like they added the deprecation warning of the old syntax in the same version that they enabled parsing of the new syntax.

I'll close this PR and just pin my local version of sphinx to 4.X

lethosor: I forgot to follow up. I tested this PR locally and I actually did see correct rendering with Sphinx 4.5, so somehow it supports formats with and without %s. Or it's also possible that my test was flawed.

Pinning to <5 is fine by me. The reason it's pinned to <4.4 in GitHub actions is because of an overly-agressive extlinks warning in 4.4, but 4.5 fixed that.

myk002:
now builds cleanly and correctly locally (with sphinx-5.0.1), on the readTheDocs build action, and in the buildmaster artifacts

lethosor:
Well, it built cleanly before with Sphinx 1.8, just left some `%s` in the docs without warning us. So I think a diff of at least a few pages is in order. I'll see if that's something I can do programmatically somehow.

lethosor:
Running a diff of `html2text` output of a few files turned up at least one other change with how definition lists ("field lists") are rendered. Here's an example in `Lua API.rst`:

Sphinx 4.5:
image

Sphinx 5.0 - note the extra colons:
image

I _think_ this is actually due to docutils 1.17.1 vs. 1.18.1 based on https://docutils.sourceforge.io/RELEASE-NOTES.html#release-0-18-2021-10-26

> Write footnote brackets and field term colons to HTML, so that they are present also without CSS and when copying text.

I don't really know what we can do about it, though, because we're following the syntax at https://docutils.sourceforge.io/docs/user/rst/quickref.html#field-lists. For instance, with `tweak` in Plugins.rst:

```
:clear-missing:  Remove the missing status from the selected unit.
                 This allows engraving slabs for ghostly, but not yet
                 found, creatures.
```

We don't have an "extra" colon in the source beyond what is necessary.

myk002:
This might be solvable with some custom css:
```
dl.field-list > dt:after {
  display: none;
}
```
From:
https://bytemeta.vip/repo/readthedocs/sphinx_rtd_theme/issues/1126

PR #2194, opened 2022-06-08
use extlinks where appropriate in docs

myk002:
I apparently wasn't aware of extlinks when I wrote some documentation. Easy fix.

### DIFF
--- a/docs/guides/examples-guide.rst
+++ b/docs/guides/examples-guide.rst
@@ -91,8 +91,8 @@ order to save resources. It handles:
 - adamantine processing
 - item melting
 
-Orders are missing for plaster powder until DF `bug 11803
-<https://www.bay12games.com/dwarves/mantisbt/view.php?id=11803>`_ is fixed.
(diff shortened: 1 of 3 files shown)


docs/guides/examples-guide.rst
```diff
@@ -91,8 +91,8 @@ order to save resources. It handles:
 - adamantine processing
 - item melting
 
-Orders are missing for plaster powder until DF `bug 11803
-<https://www.bay12games.com/dwarves/mantisbt/view.php?id=11803>`_ is fixed.
+Orders are missing for plaster powder until DF :bug:`bug 11803 <11803>` is
```
lethosor, line 94: In case you were unaware, you can use just 
```
:bug:`11803`
```
and it will render as "Bug 11803" (capitalized, though)

myk002: I should have known that, since I just edited that file. Will fix.

Issue #2189, opened 2022-06-07
EventManager and Eventful UNIT_NEW_ACTIVE is broken


warmist:
The event triggers for all units all the time as the (local static) activeUnits is not updated to have "seen" ids. Also local static does not work here as it needs to be cleared on map unload and etc.

https://github.com/DFHack/dfhack/blob/develop/library/modules/EventManager.cpp#L595

warmist:
Also it could be looking at next unit id change to find if any new unit is created and only then check for updates in the unit list. (Also check with binary search as we know it's sorted?).

myk002:
@cppcooper

myk002:
Also Lua is still seeing this event as NEW_UNIT_ACTIVE. I thought we fixed that, but I guess not.

cppcooper:
yep, he's right. that shouldn't have gotten through review lol
just need an
```cpp
activeUnits.emplace(id);
```
on line 603

cppcooper:
I can implement this, but honestly it'd probably happen faster if yall did it yourselves. Whoever feels like it. I probably won't touch it for a while.. but we'll see maybe I think that now and at 9pm I'll be all `hmm, I need something to do`

cppcooper:
> Also Lua is still seeing this event as NEW_UNIT_ACTIVE

Where do you see that? This is what I see:
![image](https://user-images.githubusercontent.com/7276349/172507628-9fbb90f4-4ee0-415b-991a-5fe2b76cbbca.png)

myk002:
> > Also Lua is still seeing this event as NEW_UNIT_ACTIVE
> 
> Where do you see that? 

I think the culprit is here: https://github.com/DFHack/dfhack/blob/40a6fcdd9928d482e409801078d365b41de4e5c5/plugins/lua/eventful.lua#L157

warmist:
IMHO we can use the rewised (see https://github.com/DFHack/dfhack/issues/2190 and probably this: https://github.com/DFHack/dfhack/pull/2107) which fixes this bug in the process.

cppcooper:
> > > Also Lua is still seeing this event as NEW_UNIT_ACTIVE
> > 
> > 
> > Where do you see that?
> 
> I think the culprit is here:
> 
> https://github.com/DFHack/dfhack/blob/40a6fcdd9928d482e409801078d365b41de4e5c5/plugins/lua/eventful.lua#L157

Ah, I see.



[...]

PR #2197, opened 2022-06-09
PR for issue #2189

cppcooper:
This is a PR to make changes discussed in #2189

### DIFF
--- a/plugins/lua/eventful.lua
+++ b/plugins/lua/eventful.lua
@@ -154,7 +154,7 @@ eventType=invertTable{
     "JOB_INITIATED",
     "JOB_STARTED",
     "JOB_COMPLETED",
-    "NEW_UNIT_ACTIVE",
+    "UNIT_NEW_ACTIVE",
     "UNIT_DEATH",
     "ITEM_CREATED",
     "BUILDING",


myk002:
I updated the description so GitHub knows to close the related bug when this PR is merged.

myk002 (approved):
Code looks good, needs a changelog entry. Have you verified in a running game?

cppcooper:
> Have you verified in a running game?

Not at all.

myk002:
> > Have you verified in a running game?
> 
> Not at all.

Please test the code before submitting a PR.

library/modules/EventManager.cpp
```diff
@@ -199,6 +199,9 @@ static int32_t lastJobId = -1;
 //job completed
 static unordered_map<int32_t, df::job*> prevJobs;
 
+//active units
+static unordered_set<int32_t> activeUnits;
```
cppcooper, line 203: > Also local static does not work here as it needs to be cleared on map unload and etc.

We've raised the scope of `activeUnits` so that it can be seen and cleared when those units are cleared [on map unload] alongside all the other structures for event manager. So its below declaration has been stricken from the manager function.

library/modules/EventManager.cpp
```diff
@@ -256,6 +259,7 @@ void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event
         buildings.clear();
         constructions.clear();
         equipmentLog.clear();
+        activeUnits.clear();
```
cppcooper, line 262: as pointed out, these need to be cleared on unload.

library/modules/EventManager.cpp
```diff
@@ -592,14 +599,14 @@ static void manageNewUnitActiveEvent(color_ostream& out) {
...
 
-    static unordered_set<int32_t> activeUnits;
     multimap<Plugin*,EventHandler> copy(handlers[EventType::UNIT_NEW_ACTIVE].begin(), handlers[EventType::UNIT_NEW_ACTIVE].end());
     // iterate event handler callbacks
     for (auto &key_value : copy) {
         auto &handler = key_value.second;
         for (df::unit* unit : df::global::world->units.active) {
             int32_t id = unit->id;
             if (!activeUnits.count(id)) {
+                activeUnits.emplace(id);
```
cppcooper, line 609: > activeUnits is not updated to have "seen" ids

just blatantly missing

myk002:
> I've just been using this with my new warn-stealers script and I am not seeing the repetition I originally found.

Thanks for the confirmation! Merging.

PR #2199, opened 2022-06-10
add `ItemTraitsDialog` to materials.lua

TymurGubayev:
the dialog allows editing of item traits (where "item" is part of a job or work order or similar).
the list is the same as in vanilla work order conditions `t` change traits (but ordered differently).
example usage: `gui/workorder-details.lua` from https://github.com/DFHack/scripts/pull/358/


myk002 (approved):
looks good, though since you can select multiple traits, I'd prefer `ItemTraitsDialog` instead of `ItemTraitDialog`.

myk002:
It's worth adding this to the "Lua" section in the changelog, too.

TymurGubayev:
pluralized the name and added an entry to the changelog.

PR #2209, opened 2022-06-13
manually handle DestroyBuilding jobs

### LINKED ISSUE #2207: Don't delete refs before calling cancel_job()
I just looked at the disassembly again and Toady's code does have code to handle both `UNIT_WORKER` and `BUILDING_HOLDER` general references. His code appears to do more than what this code does, but also has conditions that this code doesn't check. For example, the code for `UNIT_WORKER` references clears some pathfinding related variables, and the code for `BUILDING_HOLDER` has stuff to update any open UI screens.

I therefore have nagging concerns that by clearing the references before they get passed to Toady's code, some things might get skipped, which could lead to leaks or other undefined behaviors. This issue deserves further investigation.

(For the interested, this is at `0x140846bc0` in the Win64 version. I haven't located the Win32 or Linux equivalents.)

_Originally posted by @ab9rf in https://github.com/DFHack/dfhack/issues/2184#issuecomment-1153238417_

myk002:
Fixes #2207

We still disconnect the refs if we can. @ab9rf is this now the correct behavior? Are there any cases where I shouldn't call `disconnectJobGeneralRef()`

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -38,6 +38,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## New Tweaks
 
 ## Fixes
+- ``job.removeJob()``: ensure jobs are removed from the world list when they are canceled
 
 ## Misc Improvements
 
--- a/library/modules/Job.cpp
+++ b/library/modules/Job.cpp
@@ -362,10 +362,19 @@ bool DFHack::Job::removeJob(df::job* job) {
 
     // cancel_job below does not clean up all refs, so we have to do some work
 
-    // clean up general refs, but do not delete them (cancel_job does further
-    // cleanup)
-    for (auto genRef : job->general_refs) {
-        disconnectJobGeneralRef(job, genRef);
+    // manually handle DESTROY_BUILDING jobs (cancel_job doesn't handle them)
+    if (job->job_type == df::job_type::DestroyBuilding) {
+        for (auto &genRef : job->general_refs) {
+            disconnectJobGeneralRef(job, genRef);
+            if (genRef) delete genRef;
+        }
+        job->general_refs.resize(0);
+
+        // remove the job from the world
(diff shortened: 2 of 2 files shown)


ab9rf:
I _think_ this is safer, but I'm not completely sure. I am going to have to go back and look at the disassembly more. I don't know if Toady's code will be made unhappy by having the backreferences deleted prematurely. Probably not, but my certitude on this is not very firm.

The areas that I've noticed that could be affected, and might be worth checking: 
* the job is in a building that is currently being displayed in the UI (the cancel code forces the UI to update)
* the job is a hunt job and the unit is actively hunting (the cancel code causes the hunter to stop being "steathy")

myk002:
as we discussed on discord/over meet, *not* removing the general refs at all has the strange behavior of making the job not get canceled, which we don't understand. This PR is probably safe, but no more than the current code. Therefore we'll put it on hold until more research can be done.

myk002:
ok, new information here. our current implementation is worse than what it was before. when we remove general refs, the jobs don't get removed from the world jobs list. It seems that it is only DestroyBuilding jobs that should get the special handling. Everything else that I've tested is handled correctly by the global cancel_job() call.

myk002:
test case 1:
1) designate a wall to be built
2) use gui/mass-remove to cancel it
old behavior: building disappears but job remains in world list
new behavior: building disappears and job is removed from world list

test case 2:
1) build a building completely (the starting wagon will do)
2) use gui/mass-remove to remove it (schedules a deconstruction job)
3) use gui/mass-remove to unremove it
old behavior: deconstruct job is removed from building but job remains in world list
new behavior: deconstruct job is removed from building and job is removed from world list

myk002:
Ok, I'll merge this because it appears to fix the problems, but this really needs to get looked into more. The behavior of cancel_job is not what we expect from looking at the disassembly.

PR #2211, opened 2022-06-14
Add default selection handler to `materials.ItemTraitsDialog`

TymurGubayev:
Now the dialog not just shows the traits, but also actually changes them out of the box.

### DIFF
--- a/library/lua/gui/materials.lua
+++ b/library/lua/gui/materials.lua
@@ -601,6 +601,98 @@ function ItemTraitsDialog(args)
         args.on_select = function(idx, obj)
             return cb(obj)
         end
+    else
+        local function toggleFlag(obj, ffield, flag)
+            local job_item = obj.job_item
+            job_item[ffield][flag] = not job_item[ffield][flag]
+        end
+
+        local function toggleToolUse(obj, tool_use)
+            local job_item = obj.job_item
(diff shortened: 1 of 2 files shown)


library/lua/gui/materials.lua
```diff
@@ -600,6 +601,98 @@ function ItemTraitsDialog(args)
...
+                -- tool use
+                toggleToolUse(obj, 'NONE')
+                -- metal ore
+                toggleMetalOre(obj, -1)
+                -- reaction class
+                toggleReactionClass(obj, '')
+                -- producing
+                toggleProductMaterial(obj, '')
+            else
+                print('unknown sel')
```
myk002, line 688: is this indicative of a programming error? if so, this should throw with `error()` to get the stack trace

myk002 (changes requested):
this looks like it could use a changelog entry.

library/lua/gui/materials.lua
TymurGubayev: this would happen if the selected entry in the list is of an unknown type, i.e. a completely new category of traits is added, but a toggle-handler is missing. I've added a call to `error`.

myk002:
The build errors are my fault. Will fix

myk002 (approved):

PR #2212, opened 2022-06-14
[WIP] Improvements to Windows compile docs?

### LINKED ISSUE #2205: Documentation for windows builds
So [take a look](https://docs.dfhack.org/en/stable/docs/Compile.html#windows), that's the beginning of the windows build documentation section. Wouldn't it be nice if we had a table of contents? Perhaps a FAQ section and other improvements?

Perhaps we should separate the documentation of platform dependent builds to different files and keep very short sections in [this file](https://github.com/DFHack/dfhack/blob/develop/docs/Compile.rst) for each platform, which then link to their respective files.
@lethosor I think you were leaning this way

@myk002 @ab9rf thoughts?

cppcooper:
The FAQ could use some massaging [suggestions/change-requests welcome]

resolves #2205 ?

### DIFF
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -689,6 +689,10 @@ an issue or pull request if you have advice to help users get WSL up and running
 
 Build
 -----
+.. contents:: Section Contents
+  :local:
+  :depth: 1
+
 There are several different batch files in the ``win32`` and ``win64``
 subfolders in the ``build`` folder, along with a script that's used for picking
 the DF path. Use the subfolder corresponding to the architecture that you want
@@ -715,10 +719,6 @@ solution file(s):
   Python.
 
 Then you can either open the solution with MSVC or use one of the msbuild scripts.
-
-.. contents:: building options
-  :local:
-  :depth: 1
 
 From the command line:
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/Compile.rst
+++ b/docs/Compile.rst
@@ -463,40 +463,132 @@ Building
 
 Windows
 =======
-On Windows, DFHack replaces the SDL library distributed with DF.
+.. contents:: Section Contents
+  :local:
+  :depth: 3
+
+In order to build on windows you'll need to resolve a few dependencies. The dfhack build system uses perl to generate
+source files using xml files detailing various DF structures. Cmake is used to generate build files, or project files
+for visual studio. Read the sections below for more detail.
+
+Note: On Windows, DFHack replaces the SDL library distributed with DF.
 
 Dependencies
 ------------
-You will need the following:
+These are the general dependencies you will need:
 
 * Microsoft Visual C++ 2022, 2019, 2017, or 2015 (optional)
 * Microsoft Visual C++ 2015 Build Tools
 * Git
 * CMake
-* Perl with XML::LibXML and XML::LibXSLT
+* StrawberryPerl
 
-  * It is recommended to install StrawberryPerl, which includes both.
-
+  * Perl
+  * XML:LibXML
+  * XML:LibXLST
 * Python (for documentation; optional, except for release builds)
 
-Microsoft Visual Studio
+Installing Dependencies
 ~~~~~~~~~~~~~~~~~~~~~~~
-Releases of Dwarf Fortress since roughly 2016 have been compiled for Windows using
-Microsoft's Visual Studio 2015 C++ compiler. In order to guarantee ABI and STL compatibility
-with Dwarf Fortress, DFHack has to be compiled with the same compiler.
+.. contents:: Section Contents
+  :local:
+  :depth: 1
+Chocolatey (`fast install instructions`_) is recommended for installing Git, Cmake, Perl, and Python if they
+are not already installed. Put simply, chocolatey is a package manager for windows so you can install/uninstall
+software from the command line.
 
-Visual Studio 2015 is no longer supported by Microsoft and it can be difficult to obtain
-working installers for this product today. As of 2022, the recommended approach
-is to use Visual Studio 2022 or Visual Studio 2019, installing additional optional
-Visual Studio components which provide the required support for using
-Visual Studio 2015's toolchain. All of the required tools are available from Microsoft as part of
-Visual Studio's Community Edition at no charge.
+.. _fast install instructions: https://chocolatey.org/install
 
-You can also download just the Visual C++ 2015 `build tools`_ if you aren't going to use
-Visual Studio to edit code.
+Choco install commands::
 
-Option 1: Build Tools Only
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+   choco install git
+   choco install cmake
+   choco install strawberryperl
+   choco install visualstudio2015community
+
+Note: strictly speaking, you do not need visual studio 2015, or any other visual studio release, but visual studio
+is recommended by these instructions. More advanced options are available to those who seek them.
+
+Alternative to chocolatey, you're also more than free to download and install manually.
+
+Manually [Fun!]
+^^^^^^^^^^^^^^^
+If you prefer to install manually rather than using Chocolatey, details and
+requirements are as below. If you do install manually, please ensure you
+have all PATHs set up correctly.
+
+Git
+...
+Some examples:
+
+* `Git for Windows <https://git-for-windows.github.io>`_ (command-line and GUI)
+* `tortoisegit <https://tortoisegit.org>`_ (GUI and File Explorer integration)
+
+CMake
+.....
+You can get the win32 installer version from
+`the official site <https://cmake.org/download/>`_.
+It has the usual installer wizard. Make sure you let it add its binary folder
+to your binary search PATH so the tool can be later run from anywhere.
+
+Perl / Strawberry Perl
+......................
+For the code generation stage of the build process, you'll need Perl 5 with
+XML::LibXML and XML::LibXSLT. `Strawberry Perl <http://strawberryperl.com>`_ is
+recommended as it includes all of the required packages in a single, easy
+install.
+
+After install, ensure Perl is in your user's PATH. This can be edited from
+``Control Panel -> System -> Advanced System Settings -> Environment Variables``.
+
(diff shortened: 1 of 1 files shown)


cppcooper:
> These look good to me.

Sorry, at least one more change to take a look at.
Maybe another, not sure. I'n thinking about the fat collected in/around the command line section, maybe finding some short/simple instructions to replace some of it with.. I'm not 100% keen on testing the command line environments for a path with sure footing.. so maybe not.

docs/Compile.rst
```diff
@@ -463,40 +463,102 @@ Building
 
 Windows
 =======
-On Windows, DFHack replaces the SDL library distributed with DF.
+.. contents:: Section Contents
+  :local:
+  :depth: 3
+
+In order to build on windows you'll need to resolve a few dependencies. The dfhack build system uses perl to generate
```
lethosor, line 470:
```suggestion
In order to build on Windows you'll need to install a few dependencies. The DFHack build system uses Perl to generate
```

docs/Compile.rst
```diff
@@ -463,40 +463,102 @@ Building
...
+source files using xml files detailing various DF structures. Cmake is used to generate build files, or project files
+for visual studio. Read the sections below for more detail.
+
```
lethosor, line 473: This bit isn't really Windows-specific.

docs/Compile.rst
```diff
@@ -463,40 +463,102 @@ Building
...
+These are the general dependencies you will need:
 
 * Microsoft Visual C++ 2022, 2019, 2017, or 2015 (optional)
 * Microsoft Visual C++ 2015 Build Tools
 * Git
 * CMake
-* Perl with XML::LibXML and XML::LibXSLT
-
-  * It is recommended to install StrawberryPerl, which includes both.
+* StrawberryPerl
```
lethosor, line 484: I think the previous phrasing was better. Another Perl will work. And in the new version, it's not clear that StrawberryPerl includes LibXML/LibXSLT already, which could lead people on a path to install those manually (unless StrawberryPerl does _not_ include those libraries?)

docs/Compile.rst
```diff
@@ -537,124 +599,69 @@ XP.
...
-guaranteed to have the correct setup (especially PATH).
-
-To install Chocolatey and the required dependencies:
-
-* Go to https://chocolatey.org in a web browser
-* At the top of the page it will give you the install command to copy
+Powershell
+..........
+If PATH in windows is configured correctly, then powershell [or cmd] should just work
+[once the dependencies are installed].
```
lethosor, line 607:
```suggestion
If PATH in Windows is configured correctly, then powershell (or cmd) should just work
(once the dependencies are installed).
```

docs/Compile.rst
```diff
@@ -697,55 +704,89 @@ during a build.  If you get a failure, you may miss important errors or warnings
...
 
+  * then that the path of perl matches what you installed, or if it is perhaps the git-bash packaged perl
+  * cmake output ex: ``-- Found Perl: C:/cmder/vendor/git-for-windows/usr/bin/perl.exe (found version "5.34.0")``
+* If it isn't the version you expect to see, read the section about `git bash`.
+* If it is, ensure the LibXML and LibXLST dependencies are installed for the detected perl installation.
+* If they are, seek help on `discord`_.
 
-Building the documentation
-==========================
+.. _discord: https://dfhack.org/discord
```
lethosor, line 788: I would suggest linking to the support doc with

```
`seek help <support>`
```

or similar. That has links to all of our support channels and is a single place that gets updated.

docs/Compile.rst
```diff
@@ -463,40 +463,102 @@ Building
...
+
+Choco install commands::
+
+   choco install git
+   choco install cmake
+   choco install strawberryperl
+
+There are more dependencies to worry about depending on your intended environment, so read on.
+
+Manually [Fun!]
```
lethosor, line 511: The TOC in this section confuses me. Is it presenting "Manually" as the first option, as an alternative to environments 1 & 2? Or is "Manually" an alternative to "install dependencies with chocolatey", and if so, could that have its own subheader?

image

docs/Compile.rst
```diff
@@ -537,124 +599,69 @@ XP.
...
+Git Bash
+........
+Git bash can be packaged with its own version of perl which will supersede the user installed version
+(StrawberryPerl) already in Windows' PATH variable. So steps should be taken to ensure the necessary
+dependencies are found/used.
 
-  * On Windows 8 and later this can be easily achieved by:
+Option 1: configure PATH
+::::::::::::::::::::::::
+Git bash will modify PATH so that its perl is found under /usr/bin. Using ``.bash_profile`` or ``.bashrc``
```
lethosor, line 619:
```suggestion
Git bash will modify PATH so that its perl is found under ``/usr/bin``. Using ``.bash_profile`` or ``.bashrc``
```

docs/Compile.rst
```diff
@@ -697,55 +704,89 @@ during a build.  If you get a failure, you may miss important errors or warnings
...
-  * Whether you're accessing it locally as with Git's Bash, or remotely through
-    Cygwin's SSH server, this is far superior to using ``cmd.exe``.
+    If you're having trouble building, remember Visual Studio sometimes needs you
+    to rebuild or clean targets. So give that a try if you're seeing build errors.
+    Otherwise check the `FAQ`_ or `discord`_
 
-You don't need to do anything special to compile from Bash. As long as your PATHs
-are set up correctly, you can run the same generate- and build/install/package- bat
-files as detailed above.
+With the documentation
```
lethosor, line 731: This section appears to be copied directly from https://docs.dfhack.org/en/stable/docs/Documentation.html#documentation. Instead, we should link to that so that information is not duplicated. This isn't really platform-specific either.

lethosor (changes requested):
Mostly some minor things, but the "build documentation" section being duplicated is something I would like to see changed.

docs/Compile.rst
cppcooper: You're right, this is unclear. I can do better. I'll think about how to clarify it.

docs/Compile.rst
cppcooper: I was thinking along the lines of: this is simple, why send them to a new page for a simple cmake command modification

docs/Compile.rst
cppcooper: I can respect that, I didn't like the old phrasing because I thought it was giving too many details they didn't need. (Windows User: ok what do I install? A: StrawberryPerl)

Maybe I could add some extra detail to get us back to an equivalent level of detail? While still starting off with what they need to install, and then showing them individual dependencies.

docs/Compile.rst
cppcooper: Too bad you didn't make this as a code suggestion I could accept right now.

docs/Compile.rst
```diff
@@ -697,55 +704,89 @@ during a build.  If you get a failure, you may miss important errors or warnings
...
+folder, and will be installed to ``hack/docs`` when you next install DFHack in a
+DF folder.
+
+FAQ
+---
+
+**What do I do when I'm getting the error** ``fatal error C1083: Cannot open include file: 'CoreProtocol.pb.h': No such file or directory``
+**or** ``..\zlib\lib\zlib.lib : warning LNK4272: library machine type 'X86' conflicts with target machine type 'x64'``?
+
+* Check the architecture CMake has configured: ``cat CMakeCache.txt | grep ARCH`` if it is 32, and you're building 64 [or vice versa] you'll need to regenerate the project from scratch.
```
cppcooper, line 766: This should probably be replaced with a windows-centric command or at least to specify to use git-bash or some such to support the command.

docs/Compile.rst
lethosor: How about having a sub-bullet read:

```
    * Note: if you install a different version of Perl, make sure you also install the ``XML::LibXML`` and ``XML::LibXSLT`` packages
````

also worth noting that it is `::` and not `:` in the package names

File: docs/Compile.rst (hunk 5 above)
Comment (cppcooper) [reply docs/Compile.rst:788]: oh, lame

File: docs/Compile.rst (hunk 3 above)
Comment (cppcooper) [reply docs/Compile.rst:484]: Oops

File: docs/Compile.rst [hunk 10]
```diff
@@ -463,40 +463,102 @@ Building
...
+
+Now that it's installed you'll need to:
+
+1. Open **Visual Studio Installer**.
+2. Select modify, for whichever version you've chosen to utilize.
+3. Check the boxes for the following components:
+
+* "Desktop Development with C++"
+* "C++ Windows XP Support for VS 2017 (v141) tools [Deprecated]"
+* "MSVC v140 - VS 2015 C++ build tools (v14.00)"
```
cppcooper, line 548: should indicate main tab vs Individual Components

docs/Compile.rst
lethosor: I suggest that we just tell people to search for DFHACK_BUILD_ARCH (i.e. the full option name) in the file and let them figure out a way that works for them. The instructions as-is could turn up multiple lines that are more difficult to interpret without context.

quarque2:
Trying the Windows compile docs for the first time right now. I was confused by the following:

```For ABI compatibility with recent releases of Dwarf Fortress, DFHack requires the v140 or v140_xp toolchain to build windows. You’ll need v140_xp if you want to build 32-bit. These instructions will help you obtain v140_xp so you can build both architectures.```

This made me think that I should use v140, because I have a 64 bit machine (as most users do). I thought I would not need to "build both architectures". Only later in the docs I found out that v140_xp is the standard and should be advised here.

myk002:
Also, the multiple `generate` batch files are confusing. It might be better if we remove the `all` one (which requires more dependencies than we have documented as required, like sphinx, and leave just minimal and gui. We should also specifically promote minimal as the script to run unless the developer has specific requirements, then they should run gui

lethosor:
I think just promoting `minimal` is sufficient. Having a script that "builds everything" is useful to some people, and I feel like people who don't want to install extra dependencies would opt for the one named "minimal".

PR #2218, opened 2022-06-19
Update tile-material.lua

### LINKED ISSUE #1956: Lua tile-material module fails on most tiles
Reported by teatime on [Discord](https://discord.com/channels/793331351645323264/807444467140788254/883793343887319090):

> There exists a `tile-material.lua`, and has apparently since 2015. Operation is claimed to just be `GetTileMat(x,y,z)` but this seems to always give me nil, even when there is a valid tile at that position (as tested wtih a more specific function from the same file).
I've narrowed the issue to `tile-material.lua:356`, but I come from a faraway land, the customs are strange to me, and i don't really communicate well in the local language; if someone would care to help, the relevant information seems to be:
> 
> - I have a mineral tile, it exists at a pos, where `local pos = {x  = ???, y = ???, z = ???}`
> - `df.tiletype.attrs[dfhack.maps.getTileType(pos)].material` is used to get a numerical representation of the tile type, in this case, a healthy 5
> - The 5 brings me joy, for it corresponds with `df.tiletype_material.MINERAL`, as expected.
> - There exists a table in `tile-material.lua`, which contains the following:
> ```lua
> BasicMats = {
>     --some stuff
>     [df.tiletype_material.MINERAL] = GetVeinMat,
> ```
> ..which references a function, function `GetVeinMat(x, y, z)`
> - This function works; see how it works:
> ```lua
> local mat = tilemat.GetVeinMat(pos)
> print("lookatme! "..tostring(mat)) -- prints: lookatme! <material 0:196 INORGANIC:COAL_BITUMINOUS>
> ```
> 
> So far so good - but alas, here things go wrong.
> - Once one enters the `GetTileMat(x,y,z)` function, which abstracts over all materials, one is propelled dowards, bringing along the BasicMats, until finally the following code is hit:
>     ```lua
>     local mat_getter = matspec[type_mat]
>     if mat_getter == nil then
>         return nil
>     end
>     return mat_getter(pos)
>     ```
> 
> `matspec` is `BasicMat`, the one that transforms the blessed 5 into `GetVeinMat`, but alas, it always falls through as `nil`.
> BasicMat breaks what appears to be the contract - `BasicMat[df.tiletype_material.MINERAL]` curses me with a `nil`, when it seems it should've given me a function.

@lethosor:

> I think that module is broken
> ```
> [lua]# m=require('tile-material')
> [lua]# ~m.BasicMats
> table: 0x7fffbc147500
> 6                           = function: 0x7fffbc147110
> ```
> because of lua scoping rules, the functions don't actually exist at the time `BasicMats` is initialized, so nearly every entry is nil

quarque2:
Fixes #1956 Moved the definition of the function tables so that the referenced functions are defined first. Also changed the license with explicit permission of the original author, Milo Christiansen.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -46,6 +46,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Lua
+- `tile-material`: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
 
 # 0.47.05-r6
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -46,7 +46,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Lua
-- ``tile-material``: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
+- ``tile-material``: fix the order of declarations. The ``GetTileMat`` function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
 
 # 0.47.05-r6
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -46,7 +46,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Lua
-- `tile-material`: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
+- ``tile-material``: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
 
 # 0.47.05-r6
 
--- a/library/lua/tile-material.lua
+++ b/library/lua/tile-material.lua
@@ -1,7 +1,7 @@
 -- tile-material: Functions to help retrieve the material for a tile.
 
 --[[
-Original code provided by Milo Christiansen in 2015 under the MIT license. Relicensed for ease of maintenance with his permission.
+Original code provided by Milo Christiansen in 2015 under the MIT license. Relicensed under the ZLib license to align with the rest of DFHack, with his permission.
 ]]
 
 local _ENV = mkmodule("tile-material")
--- a/library/lua/tile-material.lua
+++ b/library/lua/tile-material.lua
@@ -1,25 +1,7 @@
 -- tile-material: Functions to help retrieve the material for a tile.
 
 --[[
-Copyright 2015-2016 Milo Christiansen
-
-This software is provided 'as-is', without any express or implied warranty. In
-no event will the authors be held liable for any damages arising from the use of
-this software.
-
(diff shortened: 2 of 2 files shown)


myk002:
Code change looks good. Have you been able to test locally?

Could you add a changelog entry? The file in this repo is docs/changelog.txt

You committed to your develop branch, which will complicate things for you once this is merged, but I'll walk you through how to fix when we get there.

quarque2:
Yeah, I tested it locally for different types of tiles: air, plant, grass, clay and stone. All working as intended now, afaict.

docs/changelog.txt
```diff
@@ -46,6 +46,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Lua
+- `tile-material`: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
```
lethosor, line 49: Needs to be double-backticks because there isn't an anchor of this name

From https://github.com/DFHack/dfhack/runs/6971381872?check_suite_focus=true

```
docs/_auto/news.rst:6: WARNING: undefined label: tile-material
writing output... [ 33%] docs/Authors .. docs/NEWS
docs/_auto/news-dev.rst:6: WARNING: undefined label: tile-material
```

Fix:
```suggestion
- ``tile-material``: fix the order of declarations. The GetTileMat function now returns the material as intended (always returned nil before). Also changed the license info, with permission of the original author.
```

myk002: Also could use double backtics around the function name

myk002 (changes requested):

myk002 (changes requested):

myk002:
Btw. I added the word "fixes" to the description so GitHub knows to close the related issue when this PR is merged

library/lua/tile-material.lua
```diff
@@ -1,25 +1,7 @@
...
-1. The origin of this software must not be misrepresented; you must not claim
-that you wrote the original software. If you use this software in a product, an
-acknowledgment in the product documentation would be appreciated but is not
-required.
-
-2. Altered source versions must be plainly marked as such, and must not be
-misrepresented as being the original software.
-
-3. This notice may not be removed or altered from any source distribution.
+Original code provided by Milo Christiansen in 2015 under the MIT license. Relicensed for ease of maintenance with his permission.
```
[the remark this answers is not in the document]
quarque2: There are still other files with a MIT license: dumper.lua, profiler.lua and maybe more (I didn't do an exhaustive search).
Will add the bit about ZLib.

myk002 (approved):

myk002:
Ok, now let's get your develop branch fixed up. You'll get warnings on your branch that it has diverged from upstream if you don't do this.

If you haven't already, add the upstream remote as per https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-for-a-fork 

Then run:
```
git fetch upstream
git checkout develop
git reset --hard upstream/develop
git push --force
```

In the future, be sure to branch off of develop before committing changes.

PR #2220, opened 2022-06-20
Modding guide

wolfboyft:
As discussed on Discord


wolfboyft:
Oh I forgot to stage the lua api changes

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
...
+What kind of things can we make?
+
+Lots of things, and the list grows as more and more hooks and tools are developed within DFHack. You can modify behaviours by cleverly watching and modifying certain areas of the game with functions that run on every tick or by using hooks. Familiarising yourself with the many structs oof the game will help with ideas immensely, and you can always ask for help in the right places (e.g. DFHack's Discord).
+
+You will need to know Lua for the scripting guides herein.
+
+A script-maker's environment
+----------------------------
+
+A script is run by writing its path and name from a script path folder without the file extension. E.g. gui/gm-editor for hack/scripts/gui/gm-editor.lua. You can make all your scripts in hack/scripts/, but this is not recommended as it makes things much harder to maintain each update. It's recommended to make a folder with a name like "own-scripts" and add it to dfhack-config/script-paths.txt. You should also make a folder for external installed scripts from the internet that are not in hack/scripts/.
```
myk002, line 20: This is generally good advice for in-development scripts, but perhaps you should also explain the structure of an installed mod, with supporting scripts in `raw/scripts` and "autorun" scripts in `raw/init.d`. That might help put the rest of this in context. Otherwise, we're just describing how to write a script, not a "mod". Maybe an "Anatomy of a mod" section before this one.

It seems like it would be a best practice to have `init.d` scripts just call other scripts so it's easy to dynamically change behavior (since `init.d` scripts aren't overridden by similarly-named scripts in your dev directory). I.e. `init.d/dustkicker.lua` would just call `dustkickerctl start`, which is run from `raw/scripts` on an end user machine or from own-scripts in your dev env. Actual modders might have other, better advice, though.

It's a *particularly* bad idea to put in-development scripts in `hack/scripts` when writing a mod, since as soon as you test installing your mod in `raw`, any further changes you think you're making to your copy in `hack/scripts` will never be noticed.

Maybe also mention that you should prepend your "own-scripts" path with a `+` so it takes precedence over the default script directories.

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Why not just use raw modding?
+
+It's much less hacky for many things, easier to maintain, and easier to extend. Of course if you're adding a new creature or whatever do use raw modding, but for things like adding syndromes when certain actions are performed, absolutely try out DFHack. Many things will require a mix of raw modding and DFHack.
+
+What kind of things can we make?
+
+Lots of things, and the list grows as more and more hooks and tools are developed within DFHack. You can modify behaviours by cleverly watching and modifying certain areas of the game with functions that run on every tick or by using hooks. Familiarising yourself with the many structs oof the game will help with ideas immensely, and you can always ask for help in the right places (e.g. DFHack's Discord).
+
+You will need to know Lua for the scripting guides herein.
```
myk002, line 15:
```suggestion
DFHack scripts are written in Lua. If you don't already know Lua, there's a great primer at <link>.
```

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
+DFHack modding guide
+====================
+
+Preface
+-------
+
+Why not just use raw modding?
+
+It's much less hacky for many things, easier to maintain, and easier to extend. Of course if you're adding a new creature or whatever do use raw modding, but for things like adding syndromes when certain actions are performed, absolutely try out DFHack. Many things will require a mix of raw modding and DFHack.
```
myk002, line 9: Could you characterize the "many things" it's less hacky for a little more? adding syndromes is one. What are the other major use cases? You mention "modify behaviours" below, but I think we can get more specific. Getting examples from other modders might help here.

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
...
+
+What kind of things can we make?
+
+Lots of things, and the list grows as more and more hooks and tools are developed within DFHack. You can modify behaviours by cleverly watching and modifying certain areas of the game with functions that run on every tick or by using hooks. Familiarising yourself with the many structs oof the game will help with ideas immensely, and you can always ask for help in the right places (e.g. DFHack's Discord).
+
+You will need to know Lua for the scripting guides herein.
+
+A script-maker's environment
```
myk002, line 17:
```suggestion
A mod-maker's development environment
```

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
...
+====================
+
+Preface
+-------
+
+Why not just use raw modding?
+
+It's much less hacky for many things, easier to maintain, and easier to extend. Of course if you're adding a new creature or whatever do use raw modding, but for things like adding syndromes when certain actions are performed, absolutely try out DFHack. Many things will require a mix of raw modding and DFHack.
+
+What kind of things can we make?
```
myk002, line 11:
```suggestion
What kind of mods can we make?
```

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,74 @@
...
+        return
+    end
+
+Now, the field ``sex`` in a unit is an integer, but each integer corresponds to a string value (it, she, or he). We get this value by indexing the bidirectional map ``df.pronoun_type`` with an integer from the unit. Indexing the other way, with one of the strings, will yield its corresponding number. So:::
+
+    local pronounTypeString = df.pronoun_type[unit.sex]
+    print(pronounTypeString)
+
+Simple.
+
```
myk002, line 50: This is an excellent start. I think we can follow on from here by introducing the concepts that grow the reader's knowledge from "how to write a script" to "how to write a script that can support a mod":

1. "Your first script" (the preceding section)
2. "Detecting triggers" (a script that monitors gameplay, detects a condition, and calls the script from the first section)
3. "Your first mod" (packages the above two scripts (plus an init.d script) into a functioning mod)

"detecting triggers" is essentially a reframing of your "something that uses eventful and/or repeat-util" section below, but I think it would be more impactful to go over it in the context of a functioning mod.

then we can go into more advanced topics (maybe based on ideas from other existing mods so you don't have to come up with the examples from scratch)

docs/changelog.txt
```diff
@@ -69,6 +70,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `quickfort-alias-guide`: new aliases: ``forbidthread``, ``permitthread``, ``forbidadamantinethread``, ``permitadamantinethread``, ``forbidcloth``, ``permitcloth``, ``forbidadamantinecloth``, and ``permitadamantinecloth`` give you more control how adamantine-derived items are stored
 - `quickfort`: `Dreamfort <quickfort-blueprint-guide>` blueprint set improvements: automatically create tavern, library, and temple locations (restricted to residents only by default), automatically associate the rented rooms with the tavern
 - `quickfort`: `Dreamfort <quickfort-blueprint-guide>` blueprint set improvements: new design for the services level, including were-bitten hospital recovery rooms and an appropriately-themed interrogation room next to the jail! Also fits better in a 1x1 embark for minimalist players.
+- `quickfort`: `Dreamfort <quickfort-blueprint-guide>` blueprint set improvements: new design for the services level, including a werebeast-proof hospital recovery rooms and an appropriately-themed interrogation room next to the jail! Also fits better in a 1x1 embark for minimalist players.
```
myk002, line 73: Could you clean this out? Looks like leftover merge conflict text.

docs/changelog.txt
```diff
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Misc Improvements
 
 ## Documentation
+- Added `modding guide <modding-guide>`
```
myk002, line 45: Just
```
- Added `modding-guide`
```
is sufficient. When it's rendered, it will expand to a link with anchor text equal to the title of the document. E.g. it will show as:
`- Added DFHack modding guide`

wolfboyft: Ah, I assumed it would be the label text. Cool and good.

docs/guides/modding-guide.rst
wolfboyft: What do you want pulled out of "modmaker's environment" and into "anatomy of a script"? I feel like they're too similar

myk002: a script..doesn't have much anatomy, does it? unless we're talking about different specific use cases, like how to use `eventful`, how to do timed scans with `repeat-util`, etc.

myk002:
copying from discord so it doesn't get lost:
I think we also need to define some terms in the modding guide
since there are several things that can be called "mods"
there are standalone scripts that users run directly that make a change to the game, either a single change or an ongoing change
there are raw modifications that don't use scripts
and there are raw modifications that are packaged with supporting scripts that are intended to have an automatic, lasting effect on the game that the player doesn't have to explicitly do anything for to start or stop (other than installing the mod) 
it was my understanding that it's the third definition that we were focusing on for the guide, since I thought we were explaining how DFHack could improve raw-modding mods 
With examples for making specific changes that could give modders ideas
but I don't think this document needs to be an overall guide to scripting
I would love to see examples of how to use custom-raw-tokens -- that would work really well here

docs/guides/modding-guide.rst
wolfboyft: Sorry I meant what do you want moved into anatomy of a mod

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,81 @@
+.. _modding-guide:
+
+DFHack modding guide
+====================
+
+Preface
```
myk002, line 6: I think this might need a little reorganization. We should start out with an explicit "What is a mod?" section to define terms and make clear who the intended audience is for this guide.

The section can go over the difference between a script (something you "run" to make a programmatic change to game state) and a mod (something you "install" to get a desired persistent change to game content or behavior) and the relationship between the two 
(e.g. how and why are scripts used by mods)

Then the next section can maybe show an actual comparison between how something complicated to do just in the raws is made simpler by combining raw mods with scripts. Like, actually show the raw-only implementation and then show the raw plus script implementation. Even if people don't know anything about raws or scripts, this example should visually show that the script method is simpler. This will be your "hook" and the reason why people will want to read the rest of the document.

myk002 (changes requested):

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,248 @@
...
+        if entry.mode == df.unit_inventory_item.T_mode.Worn then
+            amount = amount + tonumber((customRawTokens.getToken(entry.item, "EXAMPLE_MOD_MOVEMENT_TIMER_REDUCTION_PER_TICK")) or 0)
+        end
+    end
+    dfhack.units.addMoveTimer(-amount) -- Subtract amount from movement timer if currently moving
+
+Your first whole mod
+--------------------
+
+Now, you may have noticed that you won't be able to run multiple functions on tick/as event callbacks with that ``modId`` idea alone. To solve that we can just define all the functions we want and call them from a single function.
```
myk002, line 233: This isn't exactly true. The mod author could attach as many handlers as they like to an event, each named differently, but with the mod name as a prefix. E.g. mymod-unitchecker, mymod-itemchecker, etc. I'm not sure if we need to recommend one approach over another.

docs/guides/modding-guide.rst
```diff
@@ -0,0 +1,248 @@
...
+
+Create a folder for mod projects within your Dwarf Fortress directory somewhere (e.g. ``hack/my-scripts/mods/``) and use your mod ID (in hyphen-case) as the name for the mod folders within it. The structure of and environment for fully-functioning modular mods are as follows:
```
myk002, line 235: Is this for the dev directory? It doesn't have to be in the df directory. In fact it would probably be more flexible to have it separate so you can nuke your df installation at will and not lose your dev env. This is another thing that I'm not sure we need to recommend a specific approach for. It's up to the dev's preference.

myk002 (commented):
This is shaping up really well! The examples are very nice, and it is easy to follow the progression of topics. I have some editorial and formatting notes, but they can wait until you're done adding content. Again, great job!

docs/guides/modding-guide.rst
wolfboyft: So you can use .. in script paths?

docs/guides/modding-guide.rst
wolfboyft: True true, I will reshape that.

wolfboyft:
"example mod" or "tutorial mod" for prefixes and mod ids?

myk002 (approved):

myk002 (approved):

PR #2223, opened 2022-06-23
Fix use after free's

simotek:
This was detected by gcc and causing the build to fail on my
Linux machine

### DIFF
--- a/plugins/orders.cpp
+++ b/plugins/orders.cpp
@@ -540,9 +540,9 @@ static command_result orders_import(color_ostream &out, Json::Value &orders)
             }
             else
             {
-                delete order;
-
                 out << COLOR_LIGHTRED << "Invalid item subtype for imported manager order: " << enum_item_key(order->item_type) << ":" << it["item_subtype"].asString() << std::endl;
+
+                delete order;
 
                 return CR_FAILURE;
             }
@@ -716,9 +716,9 @@ static command_result orders_import(color_ostream &out, Json::Value &orders)
                     }
                     else
                     {
-                        delete condition;
-
                         out << COLOR_YELLOW << "Invalid item condition item subtype for imported manager order: " << enum_item_key(condition->item_type) << ":" << it2["item_subtype"].asString() << std::endl;
+
+                        delete condition;
 
                         continue;
                     }


myk002 (approved):
I see these would have only been hit for invalid imported items, which explains why this hasn't been reported. Still, very good to fix. Thank you!

myk002:
What compiler version are you using that detected this? Do you have any extra flags set?

myk002:
I noticed that you committed directly to your `develop` branch for this PR. You'll get warnings on your branch that it has diverged from upstream if you don't fix it up now that this PR is merged.

If you haven't already, add the upstream remote as per https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-for-a-fork

Then run:
```
git fetch upstream
git checkout develop
git reset --hard upstream/develop
git push --force
```
In the future, be sure to branch off of develop before committing changes.

simotek:
> What compiler version are you using that detected this? Do you have any extra flags set?

Its gcc 12.1 which is pretty recent with just the default compiler flags from the distro listed below

gcc -v 
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-suse-linux/12/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-suse-linux
Configured with: ../configure --prefix=/usr --infodir=/usr/share/info --mandir=/usr/share/man --libdir=/usr/lib64 --libexecdir=/usr/lib64 --enable-languages=c,c++,objc,fortran,obj-c++,ada,go,d,jit --enable-offload-targets=nvptx-none,amdgcn-amdhsa, --enable-offload-defaulted --without-cuda-driver --enable-host-shared --enable-checking=release --disable-werror --with-gxx-include-dir=/usr/include/c++/12 --enable-ssp --disable-libssp --disable-libvtv --enable-cet=auto --disable-libcc1 --enable-plugin --with-bugurl=https://bugs.opensuse.org/ --with-pkgversion='SUSE Linux' --with-slibdir=/lib64 --with-system-zlib --enable-libstdcxx-allocator=new --disable-libstdcxx-pch --enable-libphobos --enable-version-specific-runtime-libs --with-gcc-major-version-only --enable-linker-build-id --enable-linux-futex --enable-gnu-indirect-function --program-suffix=-12 --without-system-libunwind --enable-multilib --with-arch-32=x86-64 --with-tune=generic --with-build-config=bootstrap-lto-lean --enable-link-mutex --build=x86_64-suse-linux --host=x86_64-suse-linux
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 12.1.0 (SUSE Linux)

simotek:
> I noticed that you committed directly to your `develop` branch for this PR. You'll get warnings on your branch that it has diverged from upstream if you don't fix it up now that this PR is merged.
> 
> If you haven't already, add the upstream remote as per https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-for-a-fork
> 
> Then run:
> 
> ```
> git fetch upstream
> git checkout develop
> git reset --hard upstream/develop
> git push --force
> ```
> 
> In the future, be sure to branch off of develop before committing changes.

Thanks for the heads up, I was being a little lazy, I only created the fork to fix this issue and now i'll delete it and go back to using your upstream repo, if I was planing to keep the repo I would have used a branch :-)

lethosor:
We actually just upgraded to GCC 12 in one of our CI setups (GitHub actions), and have never seen the error you fixed. As far as I know, we're also using the "default" flags - that is, we haven't explicitly changed any.

One possibility: did you override the CMake build type? I think GitHub actions is defaulting to `Release`, and our `RelWithDebInfo` or `Debug` builds might use different flags.

simotek:
Its quite probable that openSUSE / SUSE uses a different set of default gcc flags to the ones used by the default github actions, one of the benefits of the way openSUSE Tumbleweed is built  is its pretty easy to change a default gcc flag then see where things break and fix them where as this is harder to track with things like github actions so its pretty likely that openSUSE ships with a stricter set of flags by default. I couldn't easily see which picked up this issue though.

lewiji:
Just a +1 that I encountered the same issue on EndeavourOS 2021.11.30 (Arch based) gcc 12.1.0, no overrides from default.

lethosor:
The flags both of you posted are actually just compiler configuration flags - they affect how the compiler is built, but they don't affect how the compiler processes code that you are compiling.

Do you have output from a failed ninja build? That should contain the full compiler command that failed, usually starting with `gcc` or `g++` (or a full path to those), and including runtime flags.

PR #2232, opened 2022-07-01
reorganize init scripts into dfhack-config

### LINKED ISSUE #2120: User init files build on defaults instead of overwriting them
I wonder about our current guidance to copy the `dfhack.init-example` file to `dfhack.init` and make changes. Should we instead be creating a standard default baseline that is overwritten on upgrade (e.g. `hack/data/init/dfhack.init`) and instruct people to create their personal files in addition to (as opposed to the current "instead of") the defaults?

This would go for `onLoad.init` and `onMapLoad.init` too.  The defaults would set up the "baseline DFHack experience" and users would augment.

On the other hand, if we set something up in the baseline files and users don't want it, there is no way to turn some things off. For example, once `fix/blood-del` runs, there is no undoing what it did.

myk002:
Fixes #2120 

This change allows player init scripts to build on defaults instead of replace them, and prepares our init system for the usability overhaul by de-emphasizing the init scripts as the method for DFHack configuration (though they will still remain viable and available).

This PR:
- moves the init scripts out of the main df directory and into `dfhack-config/init`
- moves our existing default configuration into files in `hack/init/` which are called via `script` in the default config files in `dfhack-config/init`
- added default config files for `onLoad`, `onUnload`, `onMapLoad` and `onMapUnload`, which all call stubs in the `hack/init/` directory where we can add default configuration in the future
- updates sphinx to look in the new file location for keybindings
- updates init script docs
- updates Core.cpp to look in the new locations
- updates ci to adjust to new locations and mechanisms

What comes next:
- updating `gui/no-dfhack-init.lua` with a new message
- displaying a message to the player if there are unused init files in the DF directory that should be migrated to `dfhack-config/init`


myk002:
pre-commit.ci autofix

myk002:
the buildmaster image will need to be updated to pull in the changes to `run-tests.py`

ci/run-tests.py
```diff
@@ -68,7 +68,11 @@ def change_setting(content, setting, value):
 if args.headless:
     init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
 
-test_init_file = 'dfhackzzz_test.init'  # Core sorts these alphabetically
+try:
+    os.mkdir('dfhack-config/init')
+except OSError as error:
+    pass
+test_init_file = 'dfhack-config/init/dfhackzzz_test.init'  # Core sorts these alphabetically
```
lethosor, line 75: I've been scratching my head on ways to get the old run-tests.py to work with the new DFHack core. I'm not sure there is a good way after all.

However, it might be nice if the new run-tests.py could work with the old DFHack by putting its `.init` file in the right spot. That would enable upgrading the Buildmaster image to allow CI to pass on this PR without failing/hanging on older branches. I think we could do this by installing some sort of "properties" file into `hack` that run-tests.py can read at runtime to figure out where to install a .init file. JSON is pretty easy to parse from Python, and probably easy to generate, although all we would need now is something hardcoded like `{"init_path": "./hack/init"}` (and run-tests could default to `".'` if the file/key isn't present).

conf.py
```diff
@@ -35,7 +35,7 @@ def get_keybinds():
     """Get the implemented keybinds, and return a dict of
     {tool: [(full_command, keybinding, context), ...]}.
     """
-    with open('dfhack.init-example') as f:
```
lethosor, line 38: Might be more technically correct to scan `data/init/*.init` for keybinding commands.

dfhack-config/init/dfhack.init
```diff
@@ -0,0 +1,11 @@
+# This file runs when DFHack is initialized, when Dwarf Fortress is first
```
lethosor, line 1: So I take it none of these files are using the `dfhack-config/default` folder where files are copied in if not present?

conf.py
myk002: That's a fair point. I was planning specifically on defining that file (data/init/dfhack.keybindings.init) as **the** place for default keybindings, but it's true that we could just not have a policy and pick them up from any file in the directory.

dfhack-config/init/dfhack.init
myk002: They are. **Everything** in the dfhack-config source folder gets installed to dfhack-config/default

ci/run-tests.py
myk002: Modified to check which version of DFHack we're building and place the init script accordingly

dfhack-config/init/dfhack.init
lethosor: Oh right... I figured out `data/init` installing to `hack/init` but forgot this path also changes on install

conf.py
myk002: done

myk002:
pre-commit.ci autofix

myk002:
Hrm.. our current guidance in the changelog is to move customized config to dfhack-config/init/dfhack.init, but if they do that before the default dfhack.init is installed, they'll never get the DFHack defaults applied. This seems like frustration waiting to happen. Maybe we need a *different* file in the init dir that loads defaults? Or maybe defaults should always be loaded and the user dfhack.init can unload the ones they don't want.

ci/run-tests.py
```diff
@@ -68,7 +68,16 @@ def change_setting(content, setting, value):
 if args.headless:
     init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
 
-test_init_file = 'dfhackzzz_test.init'  # Core sorts these alphabetically
+init_path = 'dfhack-config/init'
```
lethosor, line 71: Did this change make it into the buildmaster image? I think that should happen before merging this PR. CC @BenLubar

dfhack-config/init/onMapLoad.init
```diff
@@ -0,0 +1,5 @@
+# This file runs when a fortress map is loaded, after any onLoad.init files
```
lethosor, line 1: I think this and onMapUnload apply to adventure mode too, although I could be wrong (depends on whether world.map is populated by DF)

docs/Core.rst
```diff
@@ -385,84 +396,81 @@ save-specific init files in the save folders.
...
+#. The :file:`dfhack-config/init` subdirectory in the main DF directory
 #. :file:`data/save/{world}/raw`, where ``world`` is the current save, and
 #. :file:`data/save/{world}/raw/objects`
 
-When reading commands from dfhack.init or with the `script` command, if the final
-character on a line is a backslash then the next uncommented line is considered a
-continuation of that line, with the backslash deleted.  Commented lines are skipped,
-so it is possible to comment out parts of a command with the ``#`` character.
+For each of those directories, all matching init files will be executed in
+alphebetical order.
```
lethosor, line 404:
```suggestion
alphabetical order.
```

(to be fair, this typo was in a similar sentence earlier)

docs/Core.rst
```diff
@@ -385,84 +396,81 @@ save-specific init files in the save folders.
...
 
+Before running matched init scripts in any of those locations, the
+:file:`dfhack-config/init/default.*` file that matches the event will be run to
+load DFHack defaults. Only the :file:`dfhack-config/init` direcotry is checked
```
lethosor, line 408:
```suggestion
load DFHack defaults. Only the :file:`dfhack-config/init` directory is checked
```

docs/changelog.txt
```diff
@@ -41,10 +41,10 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``job.removeJob()``: ensure jobs are removed from the world list when they are canceled
 
 ## Misc Improvements
-
 - `manipulator`: add a library of useful default professions
 - `manipulator`: move professions configuration from ``professions/`` to ``dfhack-config/professions/`` to keep it together with other dfhack configuration. If you have saved professions that you would like to keep, please manually move them to the new folder.
 - ``materials.ItemTraitsDialog``: added a default ``on_select``-handler which toggles the traits.
+- Init scripts: ``dfhack.init`` and other init scripts have moved to ``dfhack-config/init/``. If you have customized your ``dfhack.init`` file and want to keep your changes, please move the part that you have customized to the new location at ``dfhack-config/init/dfhack.init``. If you do not have changes that you want to keep, do not copy anything, and the new defaults will be used automatically.
```
lethosor, line 47: FYI, this sorts to the top since it doesn't start with ` (which is intended)

lethosor (approved):
Works well on my end!

ci/run-tests.py
BenLubar: It's easiest if it's merged first - the build script for the container pulls the script from the develop branch at build time.

docs/changelog.txt
myk002: Moved to top. It wasn't alphabetized here even if the backtick isn't taken into account! I probably had the line named differently at some point.

dfhack-config/init/onMapLoad.init
myk002: Updated comments to reflect this

ci/run-tests.py
myk002: Merged

BenLubar: Rebuilding the containers...

BenLubar: Should be good to go now.

PR #2243, opened 2022-07-10
Basic doc and sphinx logic reorg

myk002:
#2228

Lots of reorganization to prepare for new individual tool docs.
- get rid of underscore directories so ugly underscores won't appear in our doc URLs (as discussed on discord). files are now generated in `changelogs/` and `tools/`
- `conf.py` now generates wrappers for the documentation in `docs/plugins/*.rst` and `scripts/docs/**/*.rst` these wrappers define the labels for linking to the docs and include the docs themselves. Eventually, the wrappers will also generate a footer for each tool that links to other related tools, but we'll have to assign tags to everything before this becomes useful. The wrappers could also autogenerate the keybinds directives.
- the undocumented keybinds check is commented out until we get the keybinds directives autogenerated (or decide not to)
- fixed links that no longer exist (or were malformed in the first place and were causing errors with the reorg)
- adjusted doc build scripts to match what's in `CMakeLists.txt`
- split plugin docs in `Plugins.rst` into separate files in `docs/plugins/*.rst`


myk002:
pre-commit.ci autofix

myk002:
I note that with the globbing as currently configured, all the tools appear on our main page. We'll get that in editing, though.

PR #2251, opened 2022-07-26
Bump gcc to 12.1 so we can reproduce build errors

### LINKED ISSUE #2250: Compilation fails with GCC 12.1 due to -Wuninitialized raised in jsoncpp-sub
Currently the code is not compiling correctly on my machine (Arch linux) because it raises a "-Werror=uninitialized" when compiling the jsoncpp-sub dependency (full output at the bottom).

I did a bit of research and the issue might come from Arch linux already using GCC 12.1 which [introduced](https://www.gnu.org/software/gcc/gcc-12/changes.html)

> [-Wuninitialized](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/gcc/Warning-Options.html#index-Wuninitialized) warns about using uninitialized variables in member initializer lists ([PR19808](https://gcc.gnu.org/PR19808))

This change in GCC behavior together with the strict CMake flag `-Werror` of the project likely causes the compilation error, although the code itself is the same as before.
If my theory is correct, this will soon affect everyone switching to GCC 12.
Possible fixes would be to either remove the `-Werror` flag for this warning, or to fix the issue upstream.

Full error message below:
```
[ 36%] Building CXX object depends/jsoncpp-sub/src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/json_value.cpp.o
In file included from /usr/include/c++/12.1.0/bits/exception_ptr.h:43,
                 from /usr/include/c++/12.1.0/exception:168,
                 from /usr/include/c++/12.1.0/ios:39,
                 from /usr/include/c++/12.1.0/istream:38,
                 from /usr/include/c++/12.1.0/sstream:38,
                 from /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/include/json/assertions.h:10,
                 from /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:7:
In function ‘std::_Require<std::__not_<std::__is_tuple_like<_Tp> >, std::is_move_constructible<_Tp>, std::is_move_assignable<_Tp> > std::swap(_Tp&, _Tp&) [with _Tp = Json::Value::ValueHolder]’,
    inlined from ‘void Json::Value::swapPayload(Json::Value&)’ at /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:530:12,
    inlined from ‘void Json::Value::swap(Json::Value&)’ at /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:543:14,
    inlined from ‘Json::Value::Value(Json::Value&&)’ at /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:492:7,
    inlined from ‘Json::Value& Json::Value::append(Json::Value&&)’ at /home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:1151:81:
/usr/include/c++/12.1.0/bits/move.h:204:11: error: ‘<unnamed>.Json::Value::value_’ is used uninitialized [-Werror=uninitialized]
  204 |       _Tp __tmp = _GLIBCXX_MOVE(__a);
      |           ^~~~~
/home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp: In member function ‘Json::Value& Json::Value::append(Json::Value&&)’:
/home/benjamin/Repository/pkgs/dfhack/dfhack/depends/jsoncpp-sub/src/lib_json/json_value.cpp:1151:81: note: ‘<anonymous>’ declared here
 1151 |   Value& Value::append(Value&& value) { return (*this)[size()] = std::move(value); }
      |                                                                                 ^
cc1plus: all warnings being treated as errors
make[2]: *** [depends/jsoncpp-sub/src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/build.make:90: depends/jsoncpp-sub/src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/json_value.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1103: depends/jsoncpp-sub/src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

myk002:
Fixes #2250

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         - default
         include:
         - os: ubuntu-22.04
-          gcc: 12
+          gcc: 12.1
           plugins: all
     steps:
     - name: Set up Python 3


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         - default
         include:
         - os: ubuntu-22.04
-          gcc: 12
+          gcc: 12.1
```
lethosor, line 21: This variable gets interpreted as part of a package name, and Ubuntu has no `gcc-12.1` package (only major versions are used). 

https://packages.ubuntu.com/search?keywords=gcc-12
https://packages.ubuntu.com/search?keywords=gcc-12.1

PR #2253, opened 2022-07-26
Support backtick as a keybinding

myk002:
In preparation for binding to `gui/launcher` once it is merged.

### DIFF
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -224,7 +224,8 @@ To set keybindings, use the built-in ``keybinding`` command. Like any other
 command it can be used at any time from the console, but bindings are not
 remembered between runs of the game unless re-created in `dfhack.init`.
 
-Currently, any combinations of Ctrl/Alt/Shift with A-Z, 0-9, or F1-F12 are supported.
+Currently, any combinations of Ctrl/Alt/Shift with A-Z, 0-9, F1-F12 or \`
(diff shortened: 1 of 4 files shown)


lethosor (approved):
Works as intended, although using it in combination with certain modifier keys will conflict with system shortcuts on some systems.

Echoing something I've mentioned earlier: any keybindings that use this should also use something else. It's difficult to access this key on a lot of international keyboards - some require Shift or AltGr to get to it, which would probably break the SDL detection logic (in a similar way as detecting `~` on a US keyboard), and some [simply don't have the key at all](http://kbdlayout.info/KBDIT?arrangement=ISO105).

myk002:
> Echoing something I've mentioned earlier: any keybindings that use this should also use something else.

This is a fair point. I'll see if I can find out what other games (like Skyrim) do on those keyboards.

myk002:
https://www.liveabout.com/cheat-codes-with-international-keyboard-3399944 seems to confirm what I read in the SDL docs -- even if the key below ESC is not the tilde key, it will still send that SDL keycode. That doesn't help with the Italian layout, and the article lists several other keyboard layouts that are similarly problematic, but using the tilde key (backtick) as a hotkey should work as well as it does for AAA games. An alternate hotkey for `gui/launcher` would of course still be a good idea.

PR #2256, opened 2022-08-02
Split autonestbox and autobutcher out from zone into their own plugins

myk002:
the three commands are not closely related. putting them all in one plugin just makes the doc page for zone very cluttered and difficult to read.

the plugin API and command processing portions of autonestbox and autobutcher have been rewritten to conform to the proposed standards. This simplified the code considerably.


myk002:
pre-commit.ci autofix

myk002:
merging to unblock docs work. I'm happy to address any concerns you have about this PR post-merge.

PR #2258, opened 2022-08-07
Docs: custom sphinx directives for tools and commands

lethosor:
See changes under `docs/` for some example usage


docs/plugins/3dveins.rst
```diff
@@ -1,6 +1,9 @@
 3dveins
 =======
-**Tags:** `tag/fort`, `tag/mod`, `tag/map`
+
+.. dfhack-tool::
+    :tags: fort, mod, map
```
lethosor, line 5: Worth noting that with how I've written `directive_arg_str_list()`, we could also leave out commas entirely (but cannot mix-and-match separators within a single directive)

```suggestion
    :tags: fort mod map
```

so what we use depends on whether tags could have spaces in them.

docs/styles/dfhack.css
```diff
@@ -60,3 +60,8 @@ div.body {
 span.pre {
     overflow-wrap: break-word;
 }
+
+.dfhack-tool-summary p {
```
lethosor, line 64: Could potentially style `.dfhack-tool-summary` arbitrarily

docs/plugins/3dveins.rst
```diff
@@ -1,6 +1,9 @@
...
+
 :dfhack-keybind:`3dveins`
```
myk002, line 8:
```suggestion
```

docs/plugins/automelt.rst
```diff
@@ -1,6 +1,8 @@
 automelt
 ========
-**Tags:** `tag/fort`, `tag/auto`, `tag/items`, `tag/stockpiles`
+
+.. dfhack-tool::
+    :tags: fort, auto, items, stockpiles
```
myk002, line 5:
```suggestion
.. dfhack-tool::
    :tags: fort, auto, items, stockpiles
    :no-command:
```

docs/sphinx_extensions/dfhack/tool_docs.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class DFHackToolDirective(DFHackToolDirectiveBase):
+    option_spec = {
+        'tags': dfhack.util.directive_arg_str_list,
+        'no-command': rst_directives.flag,
+    }
+
+    def render_content(self) -> List[nodes.Node]:
+        tag_nodes = [nodes.strong(text='Tags: ')]
```
myk002, line 122:
```suggestion
        tag_nodes = [nodes.strong(text='Tags:'), nodes.inline(text=' ')]
```

myk002: otherwise the text rendering produces `**Tags: **tagname`, instead of `**Tags:** tagname`

docs/sphinx_extensions/dfhack/tool_docs.py
```diff
@@ -0,0 +1,175 @@
...
+                    'reftarget': 'tag/' + tag,
+                    'refexplicit': False,
+                    'refwarn': True,
+                }),
+                nodes.inline(text=' | '),
+            ]
+        tag_nodes.pop()
+
+        return [
+            self.make_labeled_paragraph('Tool', self.get_name_or_docname()),
```
myk002, line 137: As mentioned in discord, this will be appearing just under a header of the same name, so it's not clear that we need to output the tool name here again.

myk002 (approved):

docs/plugins/3dveins.rst
myk002: My hope is that tags can be single words, all lowercase. that way they can be easy to read and easy to remember (not to mention easy to parse)

docs/plugins/automelt.rst
lethosor: `automelt` is a command too, actually

docs/sphinx_extensions/dfhack/tool_docs.py
lethosor: good catch!

docs/plugins/automelt.rst
myk002: I had to go back and check because that didn't fit with the docs. I remember now: the command that prints out the version string. I think I chose not to document that on purpose : )

docs/sphinx_extensions/dfhack/tool_docs.py
myk002: It came up previously when I was testing text rendering with the keybinding directive. I made the same fix there too.

docs/plugins/automelt.rst
lethosor: You... are correct. I only looked for the PluginCommand call, which advertised an ability to actually do something useful.

myk002: #2259

lethosor: Thanks, resolved

docs/plugins/3dveins.rst
lethosor: Removed commas

docs/sphinx_extensions/dfhack/tool_docs.py
lethosor: I'll remove it. I tried adding tool type information automatically based on the document path, but failed because `env.docname` is the including document (/docs/tools/...), not the one being included.

Snippet for posterity:
```python
    def get_type_description(self):
        stem = ''
        docname_parts = self.env.docname.split('/')
        if len(docname_parts) >= 2:
            stem = docname_parts[-2].lower()
        if stem.startswith('plugin'):
            return 'plugin'
        elif stem.startswith('script'):
            return 'script'
        elif stem.startswith('builtin'):
            return 'built-in'
        else:
            return 'unknown type'

```

myk002 (approved):

Issue #2129, opened 2022-05-06
spectate needs better documentation


cppcooper:
It is fairly clear, I think, from the reddit question [here](https://www.reddit.com/r/dwarffortress/comments/uipacv/dfhack_04705r5_released/i7g12w7?utm_medium=android_app&utm_source=share&context=3) that users need better documentation of what to expect from the plugin.

Improving the help command, and the plugin.rst docs for spectate should be a thing that happens.

PR #2270, opened 2022-08-24
Extends spectate plugin

cppcooper:
Using DF AI code as an example I implemented an auto-unpause feature. I tested against a siege event's pause.

resolves https://github.com/DFHack/dfhack/issues/2129


cppcooper:
I've taken well beyond simple already to mean I should try to implement the API as idiomatically as possible. So here is my crack at that, based on the lock idea lethosor gave.

To test I enabled the spectate auto-unpause feature, then unpaused the game, and ran `reveal hell` and could not unpause. If there is another test case I've neglected, I don't know what it is.

docs/changelog.txt
```diff
@@ -57,6 +57,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `orders`: added useful library of manager orders. see them with ``orders list`` and import them with, for example, ``orders import library/basic``
 - `prospect`: add new ``--show`` option to give the player control over which report sections are shown. e.g. ``prospect all --show ores`` will just show information on ores.
 - `seedwatch`: ``seedwatch all`` now adds all plants with seeds to the watchlist, not just the "basic" crops.
+- `spectate`: ``spectate auto-unpause`` is a new feature that will auto-dismiss pause events when enabled. This does not affect the player's ability to pause at a whim.
```
lethosor, line 60: The new API parts should also be mentioned in the `API` section.

library/modules/World.cpp
```diff
@@ -68,6 +75,133 @@ void World::SetPauseState(bool paused)
...
+}
+
+bool PlayerLock::isOnlyLocked() const {
+    return std::all_of(locks.begin(), locks.end(), [&](Lock* lock) {
+        if (lock == this) {
+            return lock->isLocked();
+        }
+        return !lock->isLocked();
+    });
+}
```
lethosor, line 105: I would suggest moving the bodies of these functions into something that can be reused, to minimize duplication. (e.g. helper functions local to this file)

plugins/spectate.cpp
```diff
@@ -31,25 +33,102 @@ using namespace df::enums;
...
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_change_event event) {
+    if (enabled && world) {
+        switch (event) {
+            case SC_WORLD_LOADED:
+            case SC_MAP_LOADED:
+                if (dismiss_pause_events && World::ReadPauseState()) {
+                    *df::global::debug_nopause = false;
```
lethosor, line 77: There should also be a REQUIRE_GLOBAL() line added for this global

plugins/spectate.cpp
```diff
@@ -66,23 +145,77 @@ DFhackCExport command_result plugin_enable(color_ostream &out, bool enable) {
...
 command_result spectate (color_ostream &out, std::vector <std::string> & parameters) {
-    return plugin_enable(out, !enabled);
+    // todo: parse parameters
+    if(!parameters.empty()) {
+        if (parameters[0] == "disable") {
+            return plugin_enable(out, false);
+        } else if (parameters[0] == "enable") {
+            return plugin_enable(out, true);
+        } else if (parameters[0] == "godmode") {
+            out.print("todo?\n"); // todo: adventure as deity?
```
lethosor, line 166: Could consider commenting this entire block out (if you want to keep it around) so that it can't be invoked until it does something

plugins/spectate.cpp
```diff
@@ -66,23 +145,77 @@ DFhackCExport command_result plugin_enable(color_ostream &out, bool enable) {
...
+        if(enabled) {
+            out.print(dismiss_pause_events ? "auto-unpause: on.\n" : "auto-unpause: off.\n");
+        }
+    }
+    return DFHack::CR_OK;
 }
 
+// every tick check whether to decide to follow a dwarf
 void onTick(color_ostream& out, void* ptr) {
+    if (!df::global::ui) return;
```
lethosor, line 210: Not strictly necessary with REQUIRE_GLOBAL() - the plugin simply won't load if `ui` is null

cppcooper: Oh. Neat

plugins/spectate.cpp
cppcooper: Oh that actually can be removed now. I missed it on my spot check

library/modules/World.cpp
cppcooper: I don't understand what you're getting at with this one. They are methods of the locks, making them re-usable?

docs/plugins/spectate.rst
```diff
@@ -11,7 +11,30 @@ Usage
 
 ::
 
-    enable spectate
```
myk002, line 14: if `spectate` shouldn't be directly enabled, then the plugin shouldn't implement the `plugin_enabled` API. You can still have the function, just don't name it `plugin_enabled` and don't export it.

cppcooper: It can be, should I just remove the spec enable ordering? I guess the enable x would obviously be the preferred choice. I'll remove the enable/disable parameters

plugins/spectate.cpp
```diff
@@ -31,25 +33,102 @@ using namespace df::enums;
...
+                                     spectate,
+                                     false,
+                                     ""
+                                     " spectate\n"
+                                     "    displays plugin status\n"
+                                     " spectate enable\n"
+                                     " spectate disable\n"
+                                     " spectate auto-unpause\n"
+                                     "    toggle auto-dismissal of game pause events. e.g. a siege event pause\n"
+                                     "\n"));
```
myk002, line 66: docs go in the .rst files. in-tree plugins should not have any docs here. too easy to let them get out of sync, and this string won't be used anywhere anyway. the rendered help from the .rst file takes precedence.

cppcooper: Oh ok. That's swell

myk002: current documentation standards are [here](https://dfhack--2247.org.readthedocs.build/en/2247/docs/Documentation.html#documentation-standards)

docs/plugins/spectate.rst
```diff
@@ -11,7 +11,30 @@ Usage
...
+    spectate
+    spectate <option>
```
myk002, line 15: Requiring an explicit command to stop spectate mode seems heavy handed. How about this:
`spectate [--nopause]` starts following dwarves, if `--nopause` is specified, game events (e.g. forgotten beast attacks or cavern/magma discoveries) will not cause the game to pause. The player can still pause manually and interact with the game.

Stop spectate mode by moving the game cursor while the game is unpaused.

docs/plugins/spectate.rst
```diff
@@ -11,7 +11,30 @@ Usage
 
 ::
 
-    enable spectate
+    spectate
```
myk002, line 14: if there is now a command, the `dfhack-tool` directive needs to be updated.

plugins/spectate.cpp
```diff
@@ -31,25 +33,102 @@ using namespace df::enums;
...
+df::unit* our_dorf = nullptr;
 void* job_watched = nullptr;
 int32_t timestamp = -1;
 REQUIRE_GLOBAL(world);
 REQUIRE_GLOBAL(ui);
+REQUIRE_GLOBAL(pause_state);
+REQUIRE_GLOBAL(d_init);
+
+// todo: implement as user configurable variables
+#define tick_span 50
```
myk002, line 47: you can steal configuration code from autonestbox

myk002: steal from autonestbox for the commandline parsing, i mean. you don't need to persist the config for `spectate`

docs/plugins/spectate.rst
myk002: I suggested a different API in another comment that would sidestep this issue, I think.

docs/plugins/spectate.rst
myk002: Yeah, try it out and see if it is disruptive. I do strongly feel that the default behavior should match what the DF follow functionality does. Without market research it's hard to say for certain, but spectate is similar enough to `f`ollow that players will likely expect the "cancel" behavior to also be similar.

library/modules/World.cpp
cppcooper: I was suffering code amnesia or something, I figured it out. It will be in 22cb7d20c8a0815fd46bd973f4e2b7d60b232e39

docs/plugins/spectate.rst
cppcooper: I wouldn't claim there is a command. It's more like a self-report the plugin has built-in. I imagine I'll see a related option when I do my documenting documentation reading

docs/plugins/spectate.rst
cppcooper: So, it occurred to me I've already started a methodology for how to implement this when I added `auto-unpause`. I'll just add an `auto-disengage` as well. Logically I think they probably start turned off by default but I can do the other way if it matters to you. I may keep the configuration persistent like I think I saw skimming through autonestbox, on the fence cause then I'd have to actually require setting a state.. but that's probably a good thing to require

cppcooper: alternatively if you can tell me what I'd name the oppositely defaulted setting which does the same thing as auto-disengage I can always name it that and set the default to disengage by default. I cannot think of what that would be called.

docs/plugins/spectate.rst
cppcooper: I think this section is correct, or at least doesn't have this problem anymore

cppcooper:
There is currently a segfault, it happens under the following circumstances:
1. the fort is freshly loaded
2. the game hasn't unpaused at all yet
3. enable spectate
4. unpause the game

Anyone is welcome to debug this for me, but it's gonna be waiting until I pop open the visual studio debugger.. which will probably be waiting for me to make my VM for windows usable (no mouse/keyboard yet)

plugins/spectate.cpp
cppcooper: So I didn't utilize the lua command line parser, but autonestbox was helpful in setting up a persistent configuration

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,48 @@ spectate
...
+which state the feature should be in. Anything else will take any positive
+value.
+
+Examples
+--------
+
+``spectate``
+    The plugin reports its feature status.
+
+
```
cppcooper, line 32:
```suggestion
```

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,47 @@ spectate
...
+Options
+-------
+
+:no option:      Show plugin status.
+:tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
+                 Acts as a maximum wait time when used with focus-jobs.
+:focus-jobs:     Toggle whether the plugin should always be following a job.
+:auto-unpause:   Toggle auto-dismissal of game pause events.
+:auto-disengage: Toggle auto-disengagement of plugin through player intervention.
```
cppcooper, line 48: I believe this resolves #2129

plugins/spectate/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+
+project(spectate)
+
+SET(SOURCES
+        spectate.cpp
+        pause.cpp)
+
```
myk002, line 7: You should probably mark your header file here too so changes to the header file will trigger recompliation. Here's how labormanager does it: 
```
set(PROJECT_HDRS
    labormanager.h
    joblabormapper.h
)
set_source_files_properties(${PROJECT_HDRS} PROPERTIES HEADER_FILE_ONLY TRUE)

# mash them together (headers are marked as headers and nothing will try to compile them)
list(APPEND PROJECT_SRCS ${PROJECT_HDRS})
```

docs/changelog.txt
```diff
@@ -44,8 +44,14 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - UX: List widgets now have mouse-interactive scrollbars
 - UX: You can now hold down the mouse button on a scrollbar to make it scroll multiple times.
 - UX: You can now drag the scrollbar to scroll to a specific spot
+- `spectate`: ``spectate auto-unpause`` is a new feature that will auto-dismiss pause events when enabled. This does not affect the player's ability to pause at a whim.
+- `spectate`: ``spectate auto-disengage`` is a new feature that will auto-disengage the plugin when the player interacts with the game while unpaused
+- `spectate`: ``spectate focus-jobs`` is a new feature that allows the plugin to not always follow a job, so it can trail a dwarf after they finish a job
+- `spectate`: ``spectate tick-threshold`` is a new feature that allows the player to change how long to follow a dwarf
```
myk002, line 50: these don't need the "is a new feature that will" prefix. they're already in the "Improvements" section.

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
     :tags: fort interface
-    :no-command:
 
 Usage
 -----
 
 ::
 
     enable spectate
+    disable spectate
```
myk002, line 14: for better or worse, I've only documented the enable form for all the other commands. you can take this line out to bring it in line with the rest of the documentation.

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
+    spectate <option> <value>
 
-The plugin will automatically switch which dwarf is being followed periodically,
-preferring dwarves on z-levels with the highest job activity.
+
+When enabled, the plugin will automatically switch which dwarf is being
+followed periodically, preferring dwarves on z-levels with the highest
+job activity.
+
+To set features that toggle between two states, use {0,1} to specify
```
myk002, line 22: could we use words instead of numbers to enable and disable? Maybe something similar to https://docs.dfhack.org/en/stable/docs/tools/dwarfmonitor.html#dwarfmonitor and/or https://docs.dfhack.org/en/stable/docs/tools/buildingplan.html#buildingplan?

e.g.

```
Usage
-------

::

    enable spectate
    spectate
    spectate enable|disable <feature>
    spectate set <setting> <value>
```

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
+which state the feature should be in. Anything else will take any positive
+value.
+
+Changes to plugin settings will be saved per world. Whether the plugin itself
+is enabled or not is not saved.
```
myk002, line 27: saving features/settings is a good idea, but should the plugin remember whether it was enabled? Would a player really want a fortress to start back up in spectate mode?

docs/changelog.txt
cppcooper: I'll take this as meaning this is too verbose.. So I'll just change them to the same wording as the others in the same section

docs/plugins/spectate.rst
cppcooper: So, that would look like this. (it's more code than what I have currently)
Making this change will require testing, and reformatting the docs to distinguish "features" and "settings" I don't think it's worth the effort
```cpp
        auto get_feature = [](std::string &arg){
            if(arg == "auto-unpause"){
                return UNPAUSE;
            } else if (arg == "auto-disengage") {
                return DISENGAGE;
            } else if (arg == "focus-jobs") {
                return JOB_FOCUS;
            } else if (arg == "tick-threshold") {
                return TICK_THRESHOLD;
            }
            return INVALID;
        };
        if (parameters.size() >= 2 && parameters.size() <= 3) {
            bool state;
            bool set = false;
            if (parameters[0] == "enable") {
                state = true;
            } else if (parameters[0] == "disable") {
                state = false;
            } else if (parameters[0] == "set") {
                set = true;
            } else {
                return DFHack::CR_WRONG_USAGE;
            }
            switch (get_feature(parameters[1])) {
                case UNPAUSE:
                    if (set) {
                        return CR_WRONG_USAGE;
                    }
                    enable_auto_unpause(out, state);
                    break;
                case DISENGAGE:
                    if (set) {
                        return CR_WRONG_USAGE;
                    }
                    disengage_enabled = state;
                    break;
                case JOB_FOCUS:
                    if (set) {
                        return CR_WRONG_USAGE;
                    }
                    focus_jobs_enabled = state;
                    break;
                case TICK_THRESHOLD:
                    if (!set) {
                        return CR_WRONG_USAGE;
                    }
                    try {
                        tick_threshold = std::stol(parameters[2]);
                    } catch (const std::exception &e) {
                        out.printerr("%s\n", e.what());
                    }
                    break;
                case INVALID:
                    return DFHack::CR_WRONG_USAGE;
            }
        }
```

plugins/spectate/CMakeLists.txt
lethosor: The only thing this really does is cause the headers to be listed in the same "project", like in a Visual Studio sidebar.

CMake detects header dependencies automatically, e.g. with `gcc -M`. If you touch `joblabormapper.h` and rebuild, you'll see the same things being rebuilt with and without `joblabormapper.h` commented out in the labormanager CMakeLists, for instance.

plugins/spectate/pause.cpp
```diff
@@ -0,0 +1,160 @@
...
+            INTERPOSE_NEXT(feed)(input);
+        }
+    };
+
+    IMPLEMENT_VMETHOD_INTERPOSE(player_pause_hook, feed);
+}
+using namespace pausing;
+
+std::unordered_set<Lock*> PlayerLock::locks;
+std::unordered_set<Lock*> AnnouncementLock::locks;
```
lethosor, line 43: From [Discord](https://discord.com/channels/793331351645323264/857489615299411968/1030355598556012555): these sets appear to be initialized after the `announcementLock` and `playerLock` above are initialized, which causes undefined behavior, and is likely the cause of the tests hanging.

Moving these definitions above `namespace pausing` seems to help, but I'm not sure if that is truly sufficient.

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+    if ((focus_jobs_enabled && !job_watched) || (tick - timestamp) > (int32_t) tick_threshold) {
+        following_dwarf = true;
+        // todo: allow the user to configure b, and also revise the math
+        const double b = base;
+        double p = b * ((double) zcount / job_tracker.size());
+        std::bernoulli_distribution follow_job(p);
+        if (!job->flags.bits.special && follow_job(RNG)) {
+            job_watched = job;
+            df::unit* unit = Job::getWorker(job);
+            if (df::global::ui && unit) {
```
myk002, line 349: no need to check for ui

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+    // don't forget about it
+    int zcount = ++freq[job->pos.z];
+    job_tracker.emplace(job->id);
+    // if we're not doing anything~ then let's pick something
+    if ((focus_jobs_enabled && !job_watched) || (tick - timestamp) > (int32_t) tick_threshold) {
+        following_dwarf = true;
+        // todo: allow the user to configure b, and also revise the math
+        const double b = base;
+        double p = b * ((double) zcount / job_tracker.size());
+        std::bernoulli_distribution follow_job(p);
```
myk002, line 345: I like how you've put the random number generator engines to good use in this plugin : )

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+            timestamp = tick;
+            std::vector<df::unit*> nonworkers;
+            for (auto unit: df::global::world->units.active) {
+                if (!Units::isCitizen(unit) || unit->job.current_job) {
+                    continue;
+                }
+                nonworkers.push_back(unit);
+            }
+            std::uniform_int_distribution<> follow_drunk(0, nonworkers.size() - 1);
+            if (df::global::ui) {
```
myk002, line 363: no need to check for ui

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+        }
+    }
+}
+
+// every job completed can be forgotten about
+void onJobCompletion(color_ostream &out, void* job_ptr) {
+    auto job = (df::job*) job_ptr;
+    // forget about it
+    freq[job->pos.z]--;
+    freq[job->pos.z] = freq[job->pos.z] < 0 ? 0 : freq[job->pos.z];
```
myk002, line 375: freq is a std::map<uint16_t,uint16_t> so this bounds checking won't quite work the way it seems to be intended to. It might be better to write the above two lines as:
```
if (freq[job->pos.z])
    --freq[job->pos.z];
```

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+            }
+        } else {
+            if (World::RestoreAnnouncementSettings()) {
+                lock_collision = false;
+            }
+        }
+    }
+    while (unpause_enabled && !world->status.popups.empty()) {
+        // dismiss announcement popup(s)
+        Gui::getCurViewscreen(true)->feed_key(interface_key::CLOSE_MEGA_ANNOUNCEMENT);
```
myk002, line 191: are you certain this is the right interface key for *all* popups? Does this need a failsafe that bails out of the loop if `world->status.popups.size()` does not decrement?

myk002 (changes requested):
Overall, looks great! The pause API looks pretty cohesive and I like what you've done with it. I'm still not comfortable putting it in core just yet, but I'm glad it's here when we need it.

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+bool following_dwarf = false;
+df::unit* our_dorf = nullptr;
+df::job* job_watched = nullptr;
+int32_t timestamp = -1;
+
+std::set<int32_t> job_tracker;
+std::map<uint16_t,uint16_t> freq;
+std::default_random_engine RNG;
+
+void enable_auto_unpause(color_ostream &out, bool state);
```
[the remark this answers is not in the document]
cppcooper: There is no need. Static globals are more of a header thing, you don't gain anything by using them inside a cpp afaik.

To share a global between two or more TU, they all need to declare it with extern with one exception: whichever TU you consider the var not to be external to.  https://stackoverflow.com/a/1106167/1327399

plugins/spectate/spectate.cpp
cppcooper: I have gone to remove these checks in the past and ended up with segfaults. I can try removing them again.. but I don't feel safe doing it

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+void enable_auto_unpause(color_ostream &out, bool state) {
+    // we don't need to do any of this yet if the plugin isn't enabled
+    if (enabled) {
+        // todo: R.E. UNDEAD_ATTACK event [still pausing regardless of announcement settings]
+        if(pause_lock) {
+            // lock_collision == true means: enable_auto_unpause() was already invoked and didn't complete
+            // The onupdate function above ensure the procedure properly completes, thus we only care about
+            // state reversal here ergo `enabled != state`
+            if (lock_collision && unpause_enabled != state) {
+                out.print("handling collision\n");
```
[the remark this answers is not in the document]
cppcooper: Belongs deleted, I was careless and didn't check diligently enough for these. I thought I got them all

plugins/spectate/spectate.cpp
cppcooper: No. This is code pilfered from df-ai. I will double check, but I think that's all it does.

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
+
+Options
+-------
+
+:no option:      Show plugin status.
+:tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
+                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 50)
+:focus-jobs:     Toggle whether the plugin should always be following a job. (default: 0)
+:auto-unpause:   Toggle auto-dismissal of game pause events. (default: 0)
+:auto-disengage: Toggle auto-disengagement of plugin through player intervention while unpaused. (default: 0)
```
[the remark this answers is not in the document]
cppcooper: I had been wondering about this actually. Wasn't sure if all plugins needed this, or if I actually needed the linkage at some point, or other.

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
+--------
+
+``spectate``
+    The plugin reports its feature status.
+
+``spectate auto-unpause 1``
+    Enable the spectate plugin to automatically dismiss pause events caused
+    by the game. Siege events are one example of such a game event.
+
+``spectate tick-interval 50``
```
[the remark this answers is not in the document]
cppcooper: oops, being negligent as ever

plugins/spectate/spectate.cpp
cppcooper: Thanks. I did write the cxxrandom plugin

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,382 @@
...
+//    typedef df::viewscreen_dwarfmodest interpose_base;
+//    DEFINE_VMETHOD_INTERPOSE(void, feed, (std::set<df::interface_key>* input)) {
+//        if ((ui->main.mode == ui_sidebar_mode::Default) && !allow_player_pause) {
+//            input->erase(interface_key::D_PAUSE);
+//        }
+//        INTERPOSE_NEXT(feed)(input);
+//    }
+//};
+//
+//IMPLEMENT_VMETHOD_INTERPOSE(player_pause_hook, feed);
```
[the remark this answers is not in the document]
cppcooper: I'll double check. Probably. I know I copied it into spectate.cpp when I was starting and I was considering needs and possibly linking against both plugins

plugins/spectate/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+dfhack_plugin(${PROJECT_NAME} ${SOURCES} LINK_LIBRARIES lua)
```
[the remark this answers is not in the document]
cppcooper: ¯\_(ツ)_/¯
Your guess is as good as mine. Copy pasted, or was using lua at some point.

plugins/spectate/spectate.cpp
cppcooper: Yes it is. So the presumption here, is @BenLubar has functional code in df-ai
https://github.com/BenLubar/df-ai/blob/345855e39ecdb0e4614620cad12fcee01ac57f1c/pause.cpp#L25-L36

plugins/spectate/spectate.cpp
cppcooper: replaced it with a signed version

plugins/spectate/spectate.cpp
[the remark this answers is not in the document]
myk002: I always thought variables/functions were externally visible unless declared static (at least on *nix -- _WIN32 has always required `__declspec(dllexport)`), but I guess my knowledge is wrong or out of date.

docs/plugins/spectate.rst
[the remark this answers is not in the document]
myk002: You mean the "Options" section header name? It's only needed if the command has options. `spectate` has no options, just subcommands.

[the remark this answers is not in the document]
cppcooper: That was a comment I wrote at 4am one morning, I didn't think I submitted it. It was about the lua linkage

plugins/spectate/pause.cpp
cppcooper: Should be, but i'm no expert in the written law of this land. It works, but afaik something like this could be UB I know static initialization can get a bit complainy. I'd expect a warning [error] if it were. Would enabling sanitiziers on debug builds be feasible?

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,50 @@ spectate
...
+    Set the tick interval the followed dwarf can be changed at back to its
+    default value.
+
+Options
```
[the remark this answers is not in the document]
cppcooper: ok ok, I did it. Since I forgot the `default: 0` stuff

plugins/spectate/spectate.cpp
myk002: Ben hasn't replied, and even if this is the right interface key to send today, it might not be true in the future. Avoiding an infinite loop would be very simple. I think we should add the ~2 lines of code that would prevent it.

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,47 @@ spectate
...
+--------
+
+``spectate``
+    The plugin reports its configured status.
+
+``spectate enable auto-unpause``
+    Enable the spectate plugin to automatically dismiss pause events caused
+    by the game. Siege events are one example of such a game event.
+
+``spectate set tick-interval 50``
```
myk002, line 35: tick-interval -> tick-threshold

docs/plugins/spectate.rst
```diff
@@ -2,16 +2,47 @@ spectate
...
+Features
+--------
+:focus-jobs:     Toggle whether the plugin should always be following a job. (default: disabled)
+:auto-unpause:   Toggle auto-dismissal of game pause events. (default: disabled)
+:auto-disengage: Toggle auto-disengagement of plugin through player intervention while unpaused. (default: disabled)
+
+Settings
+--------
+:tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
+                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 50)
```
myk002, line 48: From some testing, a changing dwarves every 50 ticks seems very jarring. I found somewhere between 1000 and 10,000 to be more comfortable (at 100 FPS).

Of course, this is affected greatly by FPS -- maybe it would be better to measure seconds instead of ticks?

plugins/spectate/spectate.cpp
myk002: that works.

myk002 (changes requested):

docs/plugins/spectate.rst
cppcooper: and how many seconds was that? I think we should at least stick with the ticks for now, and add an issue to switch to seconds in the future if the discussion bears it out. What do you think?

docs/plugins/spectate.rst
cppcooper: sure, that makes more sense

cppcooper:
```suggestion
``spectate set tick-threshold 50``
```

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,363 @@
...
+                lock_collision = false;
+            }
+        } else {
+            if (World::RestoreAnnouncementSettings()) {
+                lock_collision = false;
+            }
+        }
+    }
+    while (unpause_enabled && !world->status.popups.empty()) {
+        // dismiss announcement popup(s)
```
cppcooper, line 178:
```suggestion
    }
    int failsafe = 0;
    while (unpause_enabled && !world->status.popups.empty() && ++failsafe <= 10) {
        // dismiss announcement popup(s)
```

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,363 @@
...
+        Gui::getCurViewscreen(true)->feed_key(interface_key::CLOSE_MEGA_ANNOUNCEMENT);
+        if (World::ReadPauseState()) {
+            // WARNING: This has a possibility of conflicting with `reveal hell` - if Hermes himself runs `reveal hell` on precisely the right moment that is
+            World::SetPauseState(false);
+        }
+    }
+    if (disengage_enabled && !World::ReadPauseState()) {
```
cppcooper, line 185:
```suggestion
    }
    if (failsafe) {
        out.printerr("spectate encountered a problem dismissing a popup!");
    }
    if (disengage_enabled && !World::ReadPauseState()) {
```

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+using namespace DFHack;
+using namespace Pausing;
+using namespace df::enums;
+
+void onTick(color_ostream& out, void* tick);
+void onJobStart(color_ostream &out, void* job);
+void onJobCompletion(color_ostream &out, void* job);
+
+uint64_t tick_threshold = 50;
```
cppcooper, line 43:
```suggestion
uint64_t tick_threshold = 1000;
```

plugins/spectate/spectate.cpp
cppcooper: @myk002
I thought I put this on the thread about the popup dismissal, but I guess not.

plugins/spectate/spectate.cpp
cppcooper: They have been added, see the thread I pinged you in

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,367 @@
...
+    while (unpause_enabled && !world->status.popups.empty() && ++failsafe <= 10) {
+        // dismiss announcement popup(s)
+        Gui::getCurViewscreen(true)->feed_key(interface_key::CLOSE_MEGA_ANNOUNCEMENT);
+        if (World::ReadPauseState()) {
+            // WARNING: This has a possibility of conflicting with `reveal hell` - if Hermes himself runs `reveal hell` on precisely the right moment that is
+            World::SetPauseState(false);
+        }
+    }
+    if (failsafe >= 10) {
+        out.printerr("spectate encountered a problem dismissing a popup!");
```
cppcooper, line 187:
```suggestion
        out.printerr("spectate encountered a problem dismissing a popup!\n");
```
Someone forgot a newline. hint: it was me

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,367 @@
...
+        std::uniform_int_distribution<uint64_t> follow_any(0, dwarves.size() - 1);
+            // if you're looking at a warning about a local address escaping, it means the unit* from dwarves (which aren't local)
+            our_dorf = dwarves[follow_any(RNG)];
+            df::global::ui->follow_unit = our_dorf->id;
+            job_watched = our_dorf->job.current_job;
+            following_dwarf = true;
+            if (!job_watched) {
+                timestamp = tick;
+            }
+
```
cppcooper, line 315:
```suggestion
        std::uniform_int_distribution<uint64_t> follow_any(0, dwarves.size() - 1);
        // if you're looking at a warning about a local address escaping, it means the unit* from dwarves (which aren't local)
        our_dorf = dwarves[follow_any(RNG)];
        df::global::ui->follow_unit = our_dorf->id;
        job_watched = our_dorf->job.current_job;
        following_dwarf = true;
        if (!job_watched) {
            timestamp = tick;
        }
```

There's some extra whitespace

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
```
cppcooper, line 316:
```suggestion
        }
        std::uniform_int_distribution<uint64_t> follow_any(0, dwarves.size() - 1);
        // if you're looking at a warning about a local address escaping, it means the unit* from dwarves (which aren't local)
        our_dorf = dwarves[follow_any(RNG)];
        df::global::ui->follow_unit = our_dorf->id;
        job_watched = our_dorf->job.current_job;
        following_dwarf = true;
        if (!job_watched) {
            timestamp = tick;
        }
    }
```
There's some extra whitespace

plugins/spectate/spectate.cpp
```diff
@@ -0,0 +1,367 @@
...
+            World::SetPauseState(false);
+        }
+    }
+    if (failsafe >= 10) {
+        out.printerr("spectate encountered a problem dismissing a popup!");
+    }
+    if (disengage_enabled && !World::ReadPauseState()) {
+        if (our_dorf && our_dorf->id != df::global::ui->follow_unit) {
+            plugin_enable(out, false);
+        }
```
cppcooper, line 192: I wonder if auto pausing would be desirable to the folks who'd use this feature.
```suggestion
        if (our_dorf && our_dorf->id != df::global::ui->follow_unit) {
            plugin_enable(out, false);
            World::SetPauseState(true);
        }
```

docs/plugins/spectate.rst
myk002: sounds good

plugins/spectate/spectate.cpp
myk002: I don't think so -- follow doesn't do it, and I think we should mimic that behavior

myk002 (approved):

plugins/spectate/spectate.cpp
cppcooper: alright, sounds good

PR #2273, opened 2022-08-30
[dreamfort] add traffic designations for better dining behavior

myk002:
direct dwarves to the grand hall for food selection and eating. otherwise they tend to steal raw ingredients and eat in the manager's office : p

### DIFF
--- a/data/blueprints/library/dreamfort.csv
+++ b/data/blueprints/library/dreamfort.csv
@@ -33,7 +33,7 @@ Put that file in your dfhack-config/init/ directory -- the same directory that h
 ""
 "Also check out https://docs.dfhack.org/en/stable/docs/Plugins.html#professions for more information on the default labor professions that are distributed with DFHack, including suggestions on how many dwarves of each profession you are likely to need at each stage of fort maturity."
 ""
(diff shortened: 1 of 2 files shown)


quietust:
Do traffic designations even help in this case? I thought they only helped dwarves decide *how* to get to their destination, not how to choose *where* their destination is going to be.

myk002:
> Do traffic designations even help in this case? I thought they only helped dwarves decide *how* to get to their destination, not how to choose *where* their destination is going to be.

This is something I recently learned from Reddit, and my preliminary testing has borne out. I know traffic designations won't deter a dwarf that has chosen to go to a particular tile anyway, but it appears to affect *which* available food a hungry dwarf will pick up and *which* available table they choose to sit at to eat it.

I'm still running a larger scale test, and if this technique turns out ineffective then I'll revert this commit, but it seems promising so far!

myk002:
Ok, results are in, but my !!Science!! was insufficient to be conclusive. Traffic designations do not seem to affect where the dwarves choose to eat. It *appears* to affect what food they pick up to eat, though. In my fort of 100, I did not see anyone eating uncooked food (which is a significant change), but they did still choose to eat in the jail and other traffic-restricted areas.

PR #2287, opened 2022-09-16
[overlay] add clickable overlay for launching gui/launcher

myk002:
image
image


plugins/overlay.cpp
```diff
@@ -0,0 +1,80 @@
...
+struct title_hook : df::viewscreen_titlest {
+    typedef df::viewscreen_titlest interpose_base;
+
+    DEFINE_VMETHOD_INTERPOSE(void, feed, (set<df::interface_key> *input)) {
+        if (!handle_click())
+            INTERPOSE_NEXT(feed)(input);
+    }
+    DEFINE_VMETHOD_INTERPOSE(void, render, ()) {
+        INTERPOSE_NEXT(render)();
+        draw_widgets();
```
lethosor, line 47: I would swap the order of these, in case screens are drawing something important in this region

myk002: I tested that, and nothing is shown at all if the order is reversed. At least on the title screen.

lethosor: Bah, it must be drawing blank characters on the whole screen. Any screen using drawBorder() would run into the same thing, probably.

Not sure what a good fix is, then. Maybe scanning for text characters before writing? I don't want to overwrite things like the weather indicator.

myk002: I planned to move the weather indicator somewhere non-conflicting, like a few tiles to the right. Are there others that I need to watch out for?

myk002: done, but you have to `rm dfhack-config/dwarfmonitor.json` to get the update to the weather indicator position.

lethosor: I guess moving that is fine. I noticed the controls on `gui/load-screen` when I was loading a game. That one won't be covered if we are skipping DFHack screens, though.

I don't know of any DF screens that would conflict off the top of my head - most draw a border, which we can draw on top of easily enough. Exceptions that I know of: title, options, save, load, clean

myk002:
merging so TWBT can make progress. more work is needed on our side, but basic mouse grid mapping is tested to work

PR #2293, opened 2022-09-21
[docs] create genindex entries for all tools and commands

myk002:
The resulting index looks the same as the one we prototyped, which is both good and bad. Everything is there, but the links are anchored to the summaries, not the tool names.

Depends on #2294 and #2295


myk002:
interesting, it builds locally, but not for the readthedocs build action. This is probably due to a difference in sphinx version used to build. I'll see if I can find a way to be compatible.

myk002:
ok, the issue is not that the sphinx I'm building with is too old, it's that the readthedocs build uses *sphinx -1.8.6*, which is too old. We need a readthedocs config file to set our versions, since right now we're just using the defaults.

PR #2294, opened 2022-09-21
Update min sphinx version to 3.4.3 to support index generation

myk002:
depends on #2295 so the readthedocs build uses the newer sphinx version


myk002:
ha, I should have guessed that I couldn't make these changes together. I'll split the readthedocs config out into a separate PR

PR #2297, opened 2022-09-22
Generate tag indices

myk002:


docs/sphinx_extensions/dfhack/tool_docs.py
```diff
@@ -212,6 +275,8 @@ def register(app):
 def setup(app):
     app.connect('builder-inited', register)
 
+    init_tag_indices(app)
```
lethosor, line 278: I think this should go in a hook like register(), mainly because extensions I've seen don't do "work" in setup(). Unsure if it's problematic, though.

myk002: much of my initial frustration came from trying to register the domains in a callback hook. registering them in `builder-inited` is too late. the domains never get initialized and the indices never get generated. The example code I saw at https://github.com/ofosos/sphinxrecipes/blob/a6bd74e72779fc7cf5050cd3bb2cc6e6822d194a/sphinxrecipes/sphinxrecipes.py#L204 registers the domains directly in `setup()`, and that's what finally worked for me too.

lethosor: Oh, sorry, I didn't realize this was effectively just calling out to register_index()/register_domain. That sounds fine.

PR #2300, opened 2022-09-23
ensure readthedocs checks out submodules

myk002:
since we no longer appear to be building docs for scripts : /

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,13 +9,13 @@ build:
   tools:
     python: "3"
 
+submodules:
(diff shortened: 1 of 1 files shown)


lethosor:
Oh good catch. Wondering if we might have lost some other defaults...

myk002:
possibly search config? https://docs.readthedocs.io/en/stable/config-file/v2.html#search

I'm going to merge this PR ASAP, though, since reddit users have already started to wonder why the 'latest' links I've been posting are 404ing

PR #2304, opened 2022-09-28
address feedback on the docs

### LINKED ISSUE #2288: Scripts: valid options for the assign-* scripts
Many of the assign-* scripts accept various tokens that can be found on the df wiki. Their help docs provide article/anchor urls to allow users to read more about what they're presumably trying to assign. This does however require the player to stop playing and open their browser so they can browse to the page(s) and read.

With a bit of digging I was able to find these lists (on the wiki pages) in the `df` table with `devel/query`. We should put commands in the docs for users to use to list the valid options so they have a more immediate way of accessing basic option listings.

beliefs: `devel/query --table df.value_type`
facets: `devel/query --table df.personality_facet_type`
goals: `devel/query --table df.goal_type`
skills: `devel/query --table df.job_skill`
attributes: `devel/query --table df --maxdepth 1 --search [ physical_attribute_type mental_attribute_type ]`

As was discussed [but separate from this issue] having these commands in the docs would pair nicely with a click-and-run mechanism for `gui/launcher`. More ideal, perhaps, would be to integrate the commands (above) into the scripts under a `--list` option for each command that has such lists to consider.

myk002:
Fixes #2288 (along with DFHack/scripts#444)

### DIFF
--- a/docs/plugins/autolabor.rst
+++ b/docs/plugins/autolabor.rst
@@ -45,10 +45,11 @@ and manager.
 Hunting is never assigned without a butchery, and fishing is never assigned
 without a fishery.
 
-For each labor, a preference order is calculated based on skill, biased against
-masters of other trades and excluding those who can't do the job. The labor is
-then added to the best <minimum> dwarves for that labor, then to additional
-dwarfs that meet any of these conditions:
+For each labor, a preference order is calculated based on skill, excluding those
+who can't do the job. Dwarves who are masters of particular skills are
+deprioritized in the preference list for other skills. The labor is then added
+to the best <minimum> dwarves for that labor, then to additional dwarfs that
+meet any of these conditions:
 
 * The dwarf is idle and there are no idle dwarves assigned to this labor
 * The dwarf has non-zero skill associated with the labor
--- a/docs/plugins/3dveins.rst
+++ b/docs/plugins/3dveins.rst
@@ -25,10 +25,10 @@ Example
 
     3dveins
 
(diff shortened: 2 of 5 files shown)


docs/plugins/autolabor.rst
```diff
@@ -45,10 +45,11 @@ and manager.
...
 
-For each labor, a preference order is calculated based on skill, biased against
-masters of other trades and excluding those who can't do the job. The labor is
-then added to the best <minimum> dwarves for that labor, then to additional
-dwarfs that meet any of these conditions:
+For each labor, a preference order is calculated based on skill, excluding those
+who can't do the job. Dwarves who are masters of particular skills are
+deprioritized in the preference list for other skills. The labor is then added
+to the best <minimum> dwarves for that labor, then to additional dwarfs that
+meet any of these conditions:
```
cppcooper, line 52: I am not sure this is better. And I've got to presume you know how this script is implemented because I had assumed the exclusion was related to the bias against masters part due to the comma (if so, then this change would be incorrect [I think]).

But more to the point "deprioritized in the preference list for other skills" is just as terse.

So long as I comprehend the intent, I'd go with: "Dwarves who are masters of a skill are deprioritized for other skills." Or something like that.

myk002: fixed. Thanks!

cppcooper: It might even be inferred by saying the reverse: "Masters of a skill are prioritized for that skill's labor"

PR #2305, opened 2022-09-28
fix ALL the typos

myk002:
done by copying everything into Google docs and running spellcheck

also see DFHack/scripts#445


Moth-Tolias:
i wonder if pre-commit has something to catch typos? might be worth looking into.

myk002:
Lethosor commented on this on discord. We could, but there are so many non-words in our vocabulary that the custom dictionary would be large and toilsome to maintain.

PR #2348, opened 2022-10-20
[gui/create-item] Fix hotkeys getting swallowed by the list filter

### LINKED ISSUE #2327: Fix hotkeys being swallowed by filter in gui/create-item

myk002:
Fixes #2327

fixed by ignoring just the hotkeys. we could have added the plumbing to ignore all capital letters, but that didn't seem necessary.

Also fix an off-by-one alignment error for hotkey hints in the scrollable list.

### DIFF
--- a/library/lua/gui/materials.lua
+++ b/library/lua/gui/materials.lua
@@ -56,6 +56,8 @@ function MaterialDialog:init(info)
             frame = { l = 0, r = 0, t = 4, b = 2 },
             icon_width = 2,
             on_submit = self:callback('onSubmitItem'),
+            edit_ignore_keys={'CUSTOM_SHIFT_I', 'CUSTOM_SHIFT_C',
+                              'CUSTOM_SHIFT_P'},
         },
         widgets.Label{
             text = { {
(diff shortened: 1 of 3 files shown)


library/lua/gui/materials.lua
```diff
@@ -56,6 +56,8 @@ function MaterialDialog:init(info)
             frame = { l = 0, r = 0, t = 4, b = 2 },
             icon_width = 2,
             on_submit = self:callback('onSubmitItem'),
+            edit_ignore_keys={'CUSTOM_SHIFT_I', 'CUSTOM_SHIFT_C',
```
lethosor, line 59: Wondering if we could make this a responsibility of `FilteredList`. I think we only want keys to be ignored in the edit field if they are present in a `key` field in `choices`, and `I`, `C`, `P` aren't always present (e.g. they can be disabled with the `use_*` options, and aren't visible in sub-lists).

myk002: I don't know.. It's certainly possible. We could:
- scan for keys in setChoices()
- dynamically set the EditField's ignore list

but there's no guarantee that the keys are set distinct from what the filter needs to intercept to do its job. I think it's the responsibility of the instantiating view to make sure things make sense.

maybe the better solution here is to allow the onchar callback for the EditField to be passed through the FilteredList constructor. In this case, disallowing all capital letters is the "right" thing to do, and doesn't look so hacky as what I did here.

PR #2354, opened 2022-10-21
make scrollbars more responsive when dragging

### LINKED ISSUE #2349: Scrollbars don't scroll smoothly when dragging
They can lag behind the cursor and sometimes when the cursor stops it takes a while for the text to stop adjusting to the new position.

I think we need to have a more specific scroll delta API. We currently support "large" and "small", but I think we need to be able to return specific deltas (or row numbers) to scroll to.

myk002:
Fixes #2349 

This PR sets the target position of the scrollbar directly when dragging instead of nudging it closer to the desired position every rendered frame. The nudge strategy was noticeably laggy when dragging large lists.


myk002:
no changelog entry since scrollbar dragging is already listed as new in this version

PR #2355, opened 2022-10-21
make the materials dialog filter lcase only

myk002:
this allows the capital letter hotkeys to be processed by the list widget without the special-casing of the ignore_keys attribute.

#2327 
This improves the solution merged in #2348


lethosor (approved):
Works for me. It blocks more keys than we really need, but there is an argument to be made for disabling all uppercase letters for consistency if we disable some.

PR #2361, opened 2022-10-24
[mousequery] fix z-level jumps when clicking quickly

myk002:
this happens when TWBT is enabled and multilevel is > 0. `Gui::getDepthAt()` returns inconsistent values for the same position, leading to a mismatch when we "unshift" the coordinates to account for render depth

### DIFF
--- a/plugins/mousequery.cpp
+++ b/plugins/mousequery.cpp
@@ -295,10 +295,7 @@ struct mousequery_hook : public df::viewscreen_dwarfmodest
 
     bool handleLeft(df::coord &mpos, int32_t mx, int32_t my, int32_t depth)
     {
-        if (Core::getInstance().getModstate() & DFH_MOD_SHIFT) {
-            DEBUG(log).print("shift is down; not adjusting tile depth\n");
-        }
-        else
+        if (!(Core::getInstance().getModstate() & DFH_MOD_SHIFT))
             mpos.z += depth;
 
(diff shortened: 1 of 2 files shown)


plugins/mousequery.cpp
```diff
@@ -291,10 +293,13 @@ struct mousequery_hook : public df::viewscreen_dwarfmodest
         return false;
     }
 
-    bool handleLeft(df::coord &mpos, int32_t mx, int32_t my)
+    bool handleLeft(df::coord &mpos, int32_t mx, int32_t my, int32_t depth)
     {
-        if (!(Core::getInstance().getModstate() & DFH_MOD_SHIFT))
-            mpos.z += Gui::getDepthAt(mx, my);
+        if (Core::getInstance().getModstate() & DFH_MOD_SHIFT) {
+            DEBUG(log).print("shift is down; not adjusting tile depth\n");
```
lethosor, line 299: looks like a new feature - is it left in intentionally? could be worth documenting if so.

lethosor (commented):
Makes sense to me, and makes sense to isolate the change to the plugin

plugins/mousequery.cpp
myk002: The feature is unchanged (though undocumented). If you hold shift while left clicking (and you're using twbt and have run multilevel >1) then you can select the tile that you're seeing below you instead of the one on the current zlevel. I could certainly document this.

myk002: after some testing, I think it might be better to leave this feature undocumented. It works..wonkily. not always getting you to where you might expect to go.

PR #2362, opened 2022-10-24
[automaterial] prevent z-level from changing after box select

### LINKED ISSUE #2274: MouseQuery interacts with automaterial in a weird way: up one zlevel occasionally on mouse use.
Ok so popping in here to report that using the mouse while constructing with boxes enable in fort mode will cause the screen to shift up one z level. You can recreate this if you set it to planning mode if you don't have enough resources to rapidly place down a bunch of walls. It might also be from the mouse also panning around the map interacting with auto material. Filing this issue because someone else who heavily uses mouse brushed up on this problem and I figured I better report this.

notice this in r4 of dfhack but the person who found it out first was using PyLNP so it might be more recent.

GIF: <details>  </details>

myk002:
Fixes #2274

If the player clicks the mouse within the SELECTION_IGNORE_TICKS window after finishing a selection, then the z-level changes. Reducing this window to the minimum prevents the issue from happening.

This does not address the root cause, but it does make the issue very unlikely to occur (at least with a human on the mouse).

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## New Plugins
 
 ## Fixes
+- `automaterial`: fix the cursor jumping up a z level when clicking quickly after box select
 - `gui/create-item`: prevent materials list filter from intercepting sublist hotkeys
 - `tiletypes`: no longer resets dig priority to the default when updating other properties of a tile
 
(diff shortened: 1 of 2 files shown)


lethosor (approved):
Can't reproduce the issue anymore.

I am curious if the original issue that inspired this timeout is still around (`//FIXME: Sometimes there's an extra ENTER key left over after box selection`).

myk002:
> I am curious if the original issue that inspired this timeout is still around

I couldn't get an extra ENTER key by holding down the key on the keyboard for a long time or by clicking the mouse. I didn't know the context of the original issue, though, so I was wary of removing the guard entirely.

Issue #2339, opened 2022-10-16
Overlay v2


myk002:
### Motivation

The current implementation of the overlay has some issues:

1. There are complaints that the overlay button is too intrusive, especially in adventure mode where it covers important text.
2. It currently has to interpose every single screen in DF, which leads to a messy implementation and maintenance toil when new DF screens are introduced. This also prevents the overlay from functioning on DFHack-owned screens.
3. It currently can only have one widget and one action associated with left clicking on it, which makes integration with the upcoming control panel or widgets owned by other plugins/scripts (including 3rd party scripts) difficult.

This proposal seeks to solve these shortcomings. In addition, by supplying an overlay framework that other plugins/scripts can use to display information and respond to hotkeys, we can simplify many other DFHack tools and make it easier for new tools to supply viewscreen-bound functionality. Among other things, this solves the issue where scripts that want to overlay information over an existing screen have to add a new viewscreen to the stack, potentially breaking keybindings attached to that screen.

### UX

The "[ DFHack Launcher ]" overlay button can be reduced to a 2x2 hotspot (in the upper left corner by default, but the position will be user-configurable). The hotspot could have a rendered image on some screens (title screen, main map screens, etc) and be invisible on others, but will always be reactive as a hotspot. Once we know more about how the steam renderer works, this can be changed to always rendering a semi-transparent DFHack icon, but we'll take what we can get for this version.

When the mouse rolls over that tile (or Ctrl-Shift-C is hit), a widget grows to show the tools whose hotkeys are associated with the current screen. This list will always include the launcher and (when it's ready) the control panel since their hotkeys are global. The hotkey for the menu itself would be filtered out.

The list will be scrollable with the mouse and keyboard and a short description of the selected tool (selected either by up/down arrow keys or by hovering the mouse cursor over the tool name) will be shown in an adjacent text box.

When the player clicks on a tool, hits enter, or hits a hotkey listed next to one of the tools, that tool is invoked and the widget disappears.

If the player moves the mouse from inside the widget area to outside the widget area, hits the menu hotkey, or presses Esc, then the menu closes without running a tool. If the mouse is still within the hotspot area, it will be ignored until it moves.

On the DF title screen we can display some text indicating where the hotspot is and how to use it. Something like:

> <--- The DFHack menu hotspot is here. Move the mouse cursor
> to this spot or use the Ctrl-Shift-C keybinding to show it.

### Overlay implementation

[...]

PR #2367, opened 2022-11-02
[overlay] implement overlay v2 framework

myk002:
#2339

Implement the design in the linked issue. This PR is probably best reviewed commit by commit to make the changes more digestible.

This includes the following core changes:
- refactor interface key shuttling logic from Screen.cpp into a generic `Lua::PushInterfaceKeys()`
- add a foreach preprocessor macro framework to cut down on repetitive boilerplate in overlay.cpp
- added generic `Lua::CallLuaModuleFunction()` to cut down on boilerplate across plugins

Remaining tasks for this PR:

- [x] assurance that it can handle the hotkeys menu widget (not yet written)
- [x] CLI input sanitization and overall more error handling
- [x] tool doc updates
- [x] dev guide
- [x] changelog entries

unit tests don't gel well with lua code that keeps state, it seems. requiring the plugin lua env will wipe it, which removes all registered widgets, and the overlay framework is left in that state when the tests are over. Moreover, the overlay framework lua env still has `IN_TEST` defined, with the alternate behavior that might entail.

I'll file a bug for this, but I think we need to rethink how our unit tests behave.


library/LuaApi.cpp
```diff
@@ -158,6 +158,50 @@ void Lua::GetVector(lua_State *state, std::vector<std::string> &pvec)
     }
 }
 
+void Lua::PushInterfaceKeys(lua_State *L,
```
lethosor, line 161: I think this should go in `LuaTools.cpp`. Not really sure why `PushPosXY[Z]` wound up in this file, but most other `Push*` functions are in `LuaTools.cpp` or inline in `LuaTools.h`. I think `LuaApi.cpp` is more for Lua-facing functions.

myk002: done. I moved the other Lua:: methods as well.

Issue #2244, opened 2022-07-10
Factor out fastdwarf movement timers into API


wolfboyft:
The fastdwarf movement timer modification code could really benefit modders if it were in its own API.

Fastdwarf breaks for every action type it works on so I can assume a unit won't have multiple move actions at a time, making the API's getter(s) simpler.

I think the API should have: getMoveStatus, which'd return movement timer, timer start and action type (or getMoveTimer, getMoveTimerStart, and getMoveActionType if multiple return can't be supported), and setMoveTimer, addMoveTimer (taking negatives (also maybe addMoveTimerCapped (capped to move timer start))), and multiplyMoveTimer (with flooring for integers. Maybe also multiplyMoveTimerCapped).

wolfboyft:
```cpp
int Units::addActionTimer(df::unit* unit, int amount, int affectedActionType, bool notCappedToTimerInit)
{
    CHECK_NULL_POINTER(unit);
    // Loop over all actions, if an action is of the right type, add to the timer accordingly
    // Don't forget to ignore attack timers of 0, like in fastdwarf
}

int Units::addActionTimerCategory(df::unit* unit, int amount, int affectedActionTypes, bool notCappedToTimerInit)
{
    CHECK_NULL_POINTER(unit);
    // Loop over all actions, if an action's type is in the category, add to the timer accordingly
    // Don't forget to ignore attack timers of 0, like in fastdwarf
}

int Units::multiplyActionTimer(df::unit* unit, float amount, int affectedActionType, bool notCappedToTimerInit)
{
    CHECK_NULL_POINTER(unit);
    // Loop over all actions, if an action is of the right type, multiply the timer accordingly
    // Don't forget to ignore attack timers of 0, like in fastdwarf
}

int Units::multiplyActionTimerCategory(df::unit* unit, float amount, int affectedActionTypes, bool notCappedToTimerInit)
{
    CHECK_NULL_POINTER(unit);
    // Loop over all actions, if an action's type is in the category, mutliply the timer accordingly
    // Don't forget to ignore attack timers of 0, like in fastdwarf
}
```
Current iteration of the API (want me to split those long lines?)
I just had a thought, what if you wanted to only work on action timers that are at init? Should we have a bool for controlling that? It's a believable use case, subtracting a bonus x from the time or splitting it in half but letting it run when it's not at start.

Also I'll make a PR to change the modding API's "use" of this API when this API is done.

wolfboyft:
[...]

PR #2374, opened 2022-11-07
Action timer API

wolfboyft:
Implements #2244 

Haven't made a changelog or documented anything yet because I don't know if this is going to be the final form of the API, but I did remember changelog this time haha. Unless you would have wanted me to changelog it from the get-go?
Anyway the code stuff is not yet finished either, I'm just putting it up as a PR so we can discuss better.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -1553,12 +1553,14 @@ Action Timer API
 
 This is an API to allow manipulation of unit action timers, to speed them up or slow them down.
 All functions in this API have overflow/underflow protection when modifying action timers (the value will cap out).
+Actions with a timer of 0 (or less) will not be modified as they are completed (or invalid in the case of negatives).
+Timers will be capped to go no lower than 1.
 ``affectedActionType`` parameters are integers from the DF enum ``unit_action_type``. E.g. ``df.unit_action_type.Move``.
 ``affectedActionTypeGroup`` parameters are integers from the (custom) DF enum ``unit_action_type_group``. They are as follows:
 
   * ``All`` (does not include unknown unit action types)
   * ``Movement``
-  * ``MovementFeet`` (check if the unit is not on ground before using this one?)
+  * ``MovementFeet`` (intended to only be used when the unit is standing, such as with pegasus boots from the modding guide)
   * ``Combat`` (includes bloodsucking)
   * ``Work``
 
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -1548,6 +1548,49 @@ Units module
 
   Returns a table of the cutoffs used by the above stress level functions.
 
+Action Timer API
+~~~~~~~~~~~~~~~~
+
+This is an API to allow manipulation of unit action timers, to speed them up or slow them down.
+All functions in this API have overflow/underflow protection when modifying action timers (the value will cap out).
+``affectedActionType`` parameters are integers from the DF enum ``unit_action_type``. E.g. ``df.unit_action_type.Move``.
+``affectedActionTypeGroup`` parameters are integers from the (custom) DF enum ``unit_action_type_group``. They are as follows:
+
+  * ``All`` (does not include unknown unit action types)
+  * ``Movement``
+  * ``MovementFeet`` (check if the unit is not on ground before using this one?)
+  * ``Combat`` (includes bloodsucking)
+  * ``Work``
+
+API functions:
+
+* ``subtractActionTimers(unit, amount, affectedActionType)``
+
+  Subtract ``amount`` (32-bit integer) from the timers of any actions the unit is performing of ``affectedActionType``
+  (usually one or zero actions in normal gameplay).
+
+* ``subtractGroupActionTimers(unit, amount, affectedActionTypeGroup)``
+
+  Subtract ``amount`` (32-bit integer) from the timers of any actions the unit is performing that match the ``affectedActionTypeGroup`` category.
+
+* ``multiplyActionTimers(unit, amount, affectedActionType)``
+
+  Multiply the timers of any actions of ``affectedActionType`` the unit is performing by ``amount`` (float)
+  (usually one or zero actions in normal gameplay).
+
+* ``multiplyGroupActionTimers(unit, amount, affectedActionTypeGroup)``
+
+  Multiply the timers of any actions that match the ``affectedActionTypeGroup`` category the unit is performing by ``amount`` (float).
+
+* ``setActionTimers(unit, amount, affectedActionType)``
+
+  Set the timers of any action the unit is performing of ``affectedActionType`` to ``amount`` (32-bit integer)
+  (usually one or zero actions in normal gameplay).
+
+* ``setGroupActionTimers(unit, amount, affectedActionTypeGroup)``
+
+  Set the timers of any action the unit is performing that match the ``affectedActionTypeGroup`` category to ``amount`` (32-bit integer).
+
 Items module
 ------------
 
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -1560,7 +1560,7 @@ Timers will be capped to go no lower than 1.
 
   * ``All`` (does not include unknown unit action types)
   * ``Movement``
-  * ``MovementFeet`` (intended to only be used when the unit is standing, such as with pegasus boots from the modding guide)
+  * ``MovementFeet`` (for walking speed, such as with pegasus boots from the `modding-guide`)
   * ``Combat`` (includes bloodsucking)
   * ``Work``
 
--- a/library/include/modules/Units.h
+++ b/library/include/modules/Units.h
@@ -224,5 +224,13 @@ DFHACK_EXPORT extern const std::vector<int32_t> stress_cutoffs;
 DFHACK_EXPORT int getStressCategory(df::unit *unit);
 DFHACK_EXPORT int getStressCategoryRaw(int32_t stress_level);
 
+enum ActionTypeGroup {All, Movement, MovementFeet, Offensive, Work};
+DFHACK_EXPORT void subtractActionTimer(df::unit *unit, int amount, int affectedActionType);
+DFHACK_EXPORT void subtractActionTimerCategory(df::unit *unit, int amount, int affectedActionTypes);
+DFHACK_EXPORT void multiplyActionTimer(df::unit *unit, float amount, int affectedActionType);
+DFHACK_EXPORT void multiplyActionTimerCategory(df::unit *unit, float amount, int affectedActionTypes);
+DFHACK_EXPORT void setActionTimer(df::unit *unit, int amount, int affectedActionType);
+DFHACK_EXPORT void setActionTimerCategory(df::unit *unit, int amount, int affectedActionTypes);
+
 }
 }
--- a/library/include/modules/Units.h
+++ b/library/include/modules/Units.h
@@ -227,11 +227,11 @@ DFHACK_EXPORT int getStressCategory(df::unit *unit);
 DFHACK_EXPORT int getStressCategoryRaw(int32_t stress_level);
 
 DFHACK_EXPORT void subtractActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type affectedActionType);
-DFHACK_EXPORT void subtractCategoryActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type_group affectedActionCategory);
+DFHACK_EXPORT void subtractCategoryActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type_group affectedActionTypeCategory);
 DFHACK_EXPORT void multiplyActionTimers(color_ostream &out, df::unit *unit, float amount, df::unit_action_type affectedActionType);
-DFHACK_EXPORT void multiplyCategoryActionTimers(color_ostream &out, df::unit *unit, float amount, df::unit_action_type_group affectedActionCategory);
+DFHACK_EXPORT void multiplyCategoryActionTimers(color_ostream &out, df::unit *unit, float amount, df::unit_action_type_group affectedActionTypeCategory);
 DFHACK_EXPORT void setActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type affectedActionType);
-DFHACK_EXPORT void setCategoryActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type_group affectedActionCategory);
+DFHACK_EXPORT void setCategoryActionTimers(color_ostream &out, df::unit *unit, int32_t amount, df::unit_action_type_group affectedActionTypeCategory);
 
 }
 }
--- a/library/include/modules/Units.h
+++ b/library/include/modules/Units.h
@@ -38,6 +38,8 @@ distribution.
 #include "df/misc_trait_type.h"
 #include "df/physical_attribute_type.h"
 #include "df/unit.h"
+#include "df/unit_action.h"
+#include "df/action_type_group.h"
 
 namespace df
 {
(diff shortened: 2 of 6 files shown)


library/include/modules/Units.h
```diff
@@ -205,5 +205,13 @@ DFHACK_EXPORT extern const std::vector<int32_t> stress_cutoffs;
 DFHACK_EXPORT int getStressCategory(df::unit *unit);
 DFHACK_EXPORT int getStressCategoryRaw(int32_t stress_level);
 
+enum ActionTypeGroup {All, Movement, MovementFeet, Offensive, Work};
```
myk002, line 208: a good practice for any enum is to make the first element invalid so unset enums are easily identified.

library/include/modules/Units.h
```diff
@@ -205,5 +205,13 @@ DFHACK_EXPORT extern const std::vector<int32_t> stress_cutoffs;
...
+DFHACK_EXPORT void subtractActionTimer(df::unit *unit, int amount, int affectedActionType);
+DFHACK_EXPORT void subtractActionTimerCategory(df::unit *unit, int amount, int affectedActionTypes);
+DFHACK_EXPORT void multiplyActionTimer(df::unit *unit, float amount, int affectedActionType);
+DFHACK_EXPORT void multiplyActionTimerCategory(df::unit *unit, float amount, int affectedActionTypes);
+DFHACK_EXPORT void setActionTimer(df::unit *unit, int amount, int affectedActionType);
+DFHACK_EXPORT void setActionTimerCategory(df::unit *unit, int amount, int affectedActionTypes);
```
myk002, line 214: The *Category calls should take the third parameter as type `ActionTypeGroup` to indicate that it's not just any int.
Also, the non-*Category signatures should probably take a `df::unit_action_type` neh?

plugins/fastdwarf.cpp
```diff
@@ -82,74 +82,7 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
...
-                case unit_action_type::ReleaseTerrain:
-                case unit_action_type::Parry:
-                case unit_action_type::Block:
-                case unit_action_type::HoldItem:
-                case unit_action_type::ReleaseItem:
-                default:
-                    break;
-                }
-            }
+            Units::setActionTimerCategory(unit, 1, Units::ActionTypeGroup::All);
```
myk002, line 85: love it. this is elegance.

library/include/modules/Units.h
```diff
@@ -205,5 +205,13 @@ DFHACK_EXPORT extern const std::vector<int32_t> stress_cutoffs;
 DFHACK_EXPORT int getStressCategory(df::unit *unit);
 DFHACK_EXPORT int getStressCategoryRaw(int32_t stress_level);
 
+enum ActionTypeGroup {All, Movement, MovementFeet, Offensive, Work};
+DFHACK_EXPORT void subtractActionTimer(df::unit *unit, int amount, int affectedActionType);
+DFHACK_EXPORT void subtractActionTimerCategory(df::unit *unit, int amount, int affectedActionTypes);
```
myk002, line 210: how about "subtractCategoryActionTimers"?

wolfboyft:
Must also remember to update the modding guide's use of this

library/include/modules/Units.h
wolfboyft: How about just "None"?

myk002: It's better to make it explicitly "wrong", like "Unset" or "Invalid". Then you can print a WARNING() and plugin authors have a chance of discovering their error instead of the call failing silently. Naming it "None" implies that it's a valid value and shouldn't generate a warning.

wolfboyft:
I think we would crash if you had an action type out of bounds, I should deal with that

wolfboyft:
Failed run expected (DFHack/df-structures#478) of course

wolfboyft:
OK, now to continue with what you were asking

myk002:
I merged in the df-structures changes, but this PR isn't building. Does it build locally for you?

wolfboyft:
I have some local changes.

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
     }
     return level;
 }
+
+int *getActionTimerPointer(df::unit_action *action) {
```
ab9rf, line 1948: this should be `int32_t *` not `int *`

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+}
+
+void Units::subtractActionTimerCategory(df::unit *unit, int amount, df::unit_action_timer_type_group affectedActionTypes)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        auto list = ENUM_ATTR(unit_action_type, group, action->type);
+        for (size_t i = 0; i < list.size; i++) {
+            if (list.items[i] == affectedActionTypes) {
+                int *timer = getActionTimerPointer(action);
```
ab9rf, line 2020: again, should be `int32_t *` not `int *`

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+        }
+    }
+}
+
+void Units::multiplyActionTimer(df::unit *unit, float amount, df::unit_action_type affectedActionType)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        if (affectedActionType != action->type) continue;
+        int *timer = getActionTimerPointer(action);
```
ab9rf, line 2035: should be `int32_t *` not `int *`

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+}
+
+void Units::multiplyActionTimerCategory(df::unit *unit, float amount, df::unit_action_timer_type_group affectedActionTypes)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        auto list = ENUM_ATTR(unit_action_type, group, action->type);
+        for (size_t i = 0; i < list.size; i++) {
+            if (list.items[i] == affectedActionTypes) {
+                int *timer = getActionTimerPointer(action);
```
ab9rf, line 2049: should be `int32_t *` not `int *`

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+        }
+    }
+}
+
+void Units::setActionTimer(df::unit *unit, int amount, df::unit_action_type affectedActionType)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        if (affectedActionType != action->type) continue;
+        int *timer = getActionTimerPointer(action);
```
ab9rf, line 2064: should be `int32_t *` not `int *`

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,143 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+}
+
+void Units::setActionTimerCategory(df::unit *unit, int amount, df::unit_action_timer_type_group affectedActionTypes)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        auto list = ENUM_ATTR(unit_action_type, group, action->type);
+        for (size_t i = 0; i < list.size; i++) {
+            if (list.items[i] == affectedActionTypes) {
+                int *timer = getActionTimerPointer(action);
```
ab9rf, line 2078: should be `int32_t *` not `int *`

ab9rf (changes requested):
can't depend on the compiler to pick the 'right' data size

library/include/modules/Units.h
ab9rf: you can also overrtide the default 0 for the first element by using `INVALID = -1` as the first entry, which will make the second entry 0 

also, in _most_ of our code base enum constants are defined in ALL CAPS, although there are exceptions to this and i don't think we have a formal style statement on this anywhere

ab9rf (approved):

ab9rf:
since github and i hate each other, let me be clear that i was just approving the fix of the `int32_t` issue and i'm still uncommitted on the change as a whole

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+    case unit_action_type::SuckBlood:
+        return &action->data.suckblood.timer;
+    case unit_action_type::Jump:
+    case unit_action_type::ReleaseTerrain:
+    case unit_action_type::Parry:
+    case unit_action_type::Block:
+    case unit_action_type::HoldItem:
+    case unit_action_type::ReleaseItem:
+    default:
+        break;
```
myk002, line 1996: FYI: if you want the compiler to warn you if you missed an enum, leave out the default case

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+        return &action->data.dodge.timer;
+    case unit_action_type::Recover:
+        return &action->data.recover.timer;
+    case unit_action_type::StandUp:
+        return &action->data.standup.timer;
+    case unit_action_type::LieDown:
+        return &action->data.liedown.timer;
+    case unit_action_type::Job2:
+        return &action->data.job2.timer;
+        // could also patch the unit->job.current_job->completion_timer
```
myk002, line 1984: could or should?

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+    case unit_action_type::Block:
+    case unit_action_type::HoldItem:
+    case unit_action_type::ReleaseItem:
+    default:
+        break;
+    }
+    return nullptr;
+}
+
+void Units::subtractActionTimer(df::unit *unit, int amount, df::unit_action_type affectedActionType)
```
myk002, line 2001: since you're working with int32_t, it would be wise to make the function signatures explicitly take int32_t

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        if (affectedActionType != action->type) continue;
+        int32_t *timer = getActionTimerPointer(action);
+        if (timer != nullptr && *timer != 0) {
+            *timer = max(*timer - amount, 1);
```
myk002, line 2008: be sure to protect against overflow too if amount is a large negative number, also in subtractCategoryActionTimers

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+            if (list.items[i] == affectedActionTypes) {
+                int32_t *timer = getActionTimerPointer(action);
+                if (timer != nullptr && *timer != 0) {
+                    double value = *timer;
+                    value = max(value * amount, 1.0);
+                    if (value > INT32_MAX) {
+                        value = INT32_MAX;
+                    }
+                    *timer = value;
+                }
```
myk002, line 2062: This is a copy of the code in multiplyActionTimer. You should factor this out so you only have to fix bugs in it once. Similarly for subtractCategoryActionTimers and setCategoryActionTimers.

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,153 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+    }
+}
+
+void Units::setActionTimer(df::unit *unit, int amount, df::unit_action_type affectedActionType)
+{
+    CHECK_NULL_POINTER(unit);
+    for (auto action : unit->actions) {
+        if (affectedActionType != action->type) continue;
+        int32_t *timer = getActionTimerPointer(action);
+        if (timer != nullptr && *timer != 0) {
```
myk002, line 2075: in what cases will this fail? do we need to communicate that failure to the API caller?

myk002 (commented):
Make docs/api/index.rst look like https://raw.githubusercontent.com/DFHack/dfhack/00cd52dca723c3c0c30b2fae6807f90b8a847f02/docs/guides/index.rst to get rid of the doc build error.

Also docs/Lua API.rst and docs/api/Units.rst needs some contents ; )

library/modules/Units.cpp
wolfboyft: Feels like could, as was the case with fastdwarf where this was lifted from. This is the action timer API specifically

library/modules/Units.cpp
wolfboyft: OK, that requires me to include the unknowns but I guess that makes sense

library/modules/Units.cpp
wolfboyft: I can't think of any failure cases... Maybe if *timer is negative?

library/modules/Units.cpp
wolfboyft: OK, done, will push

wolfboyft:
Yeah I'm gonna get on the documentation side of things soon

library/modules/Units.cpp
wolfboyft: OK, will do

wolfboyft: 1 - INT32_MAX doesn't underflow according to my googling, so no underflow protection required?

wolfboyft:
Next I need to get that amount input erroring setup.

library/modules/Units.cpp
```diff
@@ -1942,3 +1944,161 @@ int Units::getStressCategoryRaw(int32_t stress_level)
...
+    }
+    return nullptr;
+}
+
+void subtractActionTimerCore(df::unit_action *action, int32_t amount)
+{
+    int32_t *timer = getActionTimerPointer(action);
+    if (timer != nullptr && *timer > 0) {
+        double value = *timer;
+        value = max(value - amount, 1.0);
```
ab9rf, line 2009: This code is repetitive, can we make it less so?

ab9rf: `mutateActionTimerCor(df::unit_action *action, std::function<double(double)> mutator)` maybe

wolfboyft: I uh, this is a bit beyond me.
It looks like you are going to pass a function to a function like in Lua and that function is going to take a double and return a double?

myk002: The syntax can be a little hairy, but once you get it, it's a lot like Lua, as you said.

See https://github.com/DFHack/dfhack/commit/94c6bc8063b8d69c9ebd9014aaade9a8b7dc788a for an example of a function that takes and uses a lambda.

See https://github.com/DFHack/dfhack/blob/develop/plugins/overlay.cpp#L131-L135 to see an example of calling a function with a lambda expression.

wolfboyft:
You know, we call it [xyz]ActionTimer but in theory it would work on multiple action timers of the same type. I have some ideas that would actually have multiple same-type actions ongoing at the same time, so, should we change the name?

myk002 (changes requested):
Looks good! Is it just the docs that remain, or are there any other code changes that are needed?

wolfboyft:
I think just the docs. You might have to walk me through rebasing again to get the new docs file.

docs/dev/Lua API.rst
```diff
@@ -1548,6 +1548,49 @@ Units module
 
   Returns a table of the cutoffs used by the above stress level functions.
 
+Action Timer API
```
myk002, line 1551: You should mention somewhere (probably where you talk about overflow protection) that if an action is in progress, the minimum value that it will be set to is 1 so that the action is completed on the next tick. Similarly, action types that are currently at 0 can't be set to any other value since there is no associated action.

docs/dev/Lua API.rst
```diff
@@ -1548,6 +1548,49 @@ Units module
...
+~~~~~~~~~~~~~~~~
+
+This is an API to allow manipulation of unit action timers, to speed them up or slow them down.
+All functions in this API have overflow/underflow protection when modifying action timers (the value will cap out).
+``affectedActionType`` parameters are integers from the DF enum ``unit_action_type``. E.g. ``df.unit_action_type.Move``.
+``affectedActionTypeGroup`` parameters are integers from the (custom) DF enum ``unit_action_type_group``. They are as follows:
+
+  * ``All`` (does not include unknown unit action types)
+  * ``Movement``
+  * ``MovementFeet`` (check if the unit is not on ground before using this one?)
```
myk002, line 1561: Why is this a question? What happens when callers set this group (or All) to some value and the unit is not on the ground?

wolfboyft: I'm not sure how to word it honestly.

docs/dev/Lua API.rst
```diff
@@ -1548,6 +1548,51 @@ Units module
...
+This is an API to allow manipulation of unit action timers, to speed them up or slow them down.
+All functions in this API have overflow/underflow protection when modifying action timers (the value will cap out).
+Actions with a timer of 0 (or less) will not be modified as they are completed (or invalid in the case of negatives).
+Timers will be capped to go no lower than 1.
+``affectedActionType`` parameters are integers from the DF enum ``unit_action_type``. E.g. ``df.unit_action_type.Move``.
+``affectedActionTypeGroup`` parameters are integers from the (custom) DF enum ``unit_action_type_group``. They are as follows:
+
+  * ``All`` (does not include unknown unit action types)
+  * ``Movement``
+  * ``MovementFeet`` (intended to only be used when the unit is standing, such as with pegasus boots from the modding guide)
```
myk002, line 1563: I guess my question for this is: why does it matter if the dwarf is on the ground? This is just a timer, isn't it? Why is it bad to modify it in since conditions? Fastdwarf doesn't bother to check if the dwarf is standing before setting all timers, which includes the MovementFeet timers. Is that bad?

wolfboyft: No the idea is that if you want something to only work when you are standing, like the pegasus boots, because they work with your feet, you would check if you are standing and if you are then use the MovementFeet category. But fastdwarf has nothing to do with that case, as it wants all timers to be faster, so there's nothing wrong there.
Sorry if it's still not clear!

myk002, line 1563:
```suggestion
  * ``MovementFeet`` (for walking speed, such as with pegasus boots from the `modding-guide`)
```

myk002: is this ok?

wolfboyft:
Committed your change

myk002 (approved):

PR #2375, opened 2022-11-07
Move generally useful string functions from uicommon to MiscUtils

myk002:
and get rid of some "using" statements in headers that were leaking into widespread cpp files.

This will also remove the need for twbt to cargo cult the `int_to_string` function.

### DIFF
--- a/library/include/MiscUtils.h
+++ b/library/include/MiscUtils.h
@@ -24,6 +24,7 @@ distribution.
 
 #pragma once
 #include "Export.h"
+#include <algorithm>
 #include <iostream>
 #include <iomanip>
 #include <climits>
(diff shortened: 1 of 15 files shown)


library/include/MiscUtils.h
```diff
@@ -389,6 +386,30 @@ DFHACK_EXPORT std::string toUpper(const std::string &str);
...
+static inline std::string int_to_string(const int n)
+{
+    std::ostringstream ss;
+    ss << n;
+    return ss.str();
+}
+
+// trim from start
+static inline std::string &ltrim(std::string &s) {
+    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](char x){ return !std::isspace(x); }));
```
lethosor, line 398: On Windows (Buildmaster):
```
z:\home\buildmaster\work\_e0\src\library\include\MiscUtils.h(398): error C2039: 'isspace': is not a member of 'std'
```

not sure where it's sneaking in on Linux, but you probably want to `#include <cctype>`

myk002: Done. Thanks!

PR #2379, opened 2022-11-08
[overlay] reset implementation and simplify with lovely preprocessor macros

myk002:
- removes all logic from overlay and leaves only a plugin skeleton that interposes all the viewscreens
- adds framework for preprocessor macro-based list transformations:
  - ``MAP`` applies a function macro to a list of arguments
  - ``MAP_JOIN`` applies a function macro to a list of arguments and inserts a given separator between the results
  - ``MAP_LIST`` a specialization of ``MAP_JOIN`` with a comma as the separator

does the absorbed code need a LICENSE entry?


myk002:
It appears that windows does not like the preprocessor macros. I'll have to research that a bit to see if there's something I can do for compatibility.

OTOH, we're about to update the windows compiler for the steam version, right? That might fix the problem too.

myk002:
This appears to be a [known issue](https://github.com/swansontec/map-macro/issues/4). Fixed in VS 2019 with `/Zc:preprocessor` (and a little earlier as `/experimental:preprocessor`). Is updating the VS compiler now an option with DF 0.47.05, or should I abandon the macros until the steam version?

myk002:
it appears that our MSVC toolchain is too old even for the experimental flag. I'll revert the macros for now and revisit once we've updated our toolchain for the steam release.

PR #2382, opened 2022-11-08
Updates spectate plugin

cppcooper:
Adds a status message on map load if the plugin is enabled.

edit:
I haven't added a change log entry since the persistent info is new to this release.

### DIFF
--- a/docs/plugins/spectate.rst
+++ b/docs/plugins/spectate.rst
@@ -38,9 +38,11 @@ Examples
 
 Features
 --------
-:focus-jobs:     Toggle whether the plugin should always be following a job. (default: disabled)
 :auto-unpause:   Toggle auto-dismissal of game pause events. (default: disabled)
 :auto-disengage: Toggle auto-disengagement of plugin through player intervention while unpaused. (default: disabled)
+:animals:        Toggle whether to sometimes follow animals. (default: disabled)
+:hostiles:       Toggle whether to sometimes follow hostiles (eg. undead, titan, invader, etc.) (default: disabled)
+:visiting:       Toggle whether to sometimes follow visiting units (eg. diplomat)
 
 Settings
 --------
--- a/docs/plugins/spectate.rst
+++ b/docs/plugins/spectate.rst
@@ -46,5 +46,4 @@ Features
 
 Settings
 --------
-:tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
-                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 1000)
+:tick-threshold: Set the plugin's tick interval for changing the followed dwarf. (default: 1000)
--- a/docs/plugins/spectate.rst
+++ b/docs/plugins/spectate.rst
@@ -45,4 +45,4 @@ Features
 Settings
 --------
 :tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
-                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 50)
+                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 1000)
--- a/plugins/spectate/spectate.cpp
+++ b/plugins/spectate/spectate.cpp
@@ -4,11 +4,13 @@
 
 #include "pause.h"
 
-#include "Core.h"
-#include <modules/Gui.h>
+#include <Debug.h>
+#include <Core.h>
 #include <Console.h>
 #include <Export.h>
 #include <PluginManager.h>
+
+#include <modules/Gui.h>
 #include <modules/World.h>
 #include <modules/EventManager.h>
 #include <modules/Job.h>
@@ -25,6 +27,11 @@
 #include <random>
 #include <cinttypes>
 
+// Debugging
+namespace DFHack {
+    DBG_DECLARE(spectate, plugin, DebugCategory::LINFO);
+}
+
 DFHACK_PLUGIN("spectate");
 DFHACK_PLUGIN_IS_ENABLED(enabled);
 REQUIRE_GLOBAL(world);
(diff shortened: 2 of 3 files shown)


plugins/spectate/spectate.cpp
```diff
@@ -110,6 +110,10 @@ DFhackCExport command_result plugin_load_data (color_ostream &out) {
         pause_lock->unlock();
         enable_auto_unpause(out, unpause_enabled);
     }
+    if (enabled) {
+        std::vector<std::string> params;
+        spectate(out, params);
```
myk002, line 115: this will also cause the config to be saved. is that desirable, or should you factor out the status report into a separate function and call that instead?

cppcooper: The save is logically pointless, but it shouldn't hurt. Do you wish me to factor it out?

cppcooper: I'll just do it, there are other changes I would like to throw in

cppcooper:
This relies on #2387

docs/plugins/spectate.rst
```diff
@@ -38,11 +38,13 @@ Examples
...
 :auto-disengage: Toggle auto-disengagement of plugin through player intervention while unpaused. (default: disabled)
+:animals:        Toggle whether to sometimes follow animals. (default: disabled)
+:hostiles:       Toggle whether to sometimes follow hostiles (eg. undead, titan, invader, etc.) (default: disabled)
+:visiting:       Toggle whether to sometimes follow visiting units (eg. diplomat)
 
 Settings
 --------
 :tick-threshold: Set the plugin's tick interval for changing the followed dwarf.
-                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 50)
+                 Acts as a maximum follow time when used with focus-jobs enabled. (default: 1000)
```
myk002, line 50: if focus-jobs is no longer a thing, this probably needs to get reworded

plugins/spectate/spectate.cpp
```diff
@@ -36,51 +46,300 @@ using namespace DFHack;
...
-        config.ival(DISENGAGE) = disengage_enabled;
-        config.ival(JOB_FOCUS) = focus_jobs_enabled;
-        config.ival(TICK_THRESHOLD) = tick_threshold;
-    }
-}
+static PersistentDataItem pconfig;
 
+DFhackCExport command_result plugin_enable(color_ostream &out, bool enable);
 command_result spectate (color_ostream &out, std::vector <std::string> & parameters);
+#define COORDARGS(id) id.x, (id).y, id.z
```
myk002, line 81: any reason there are params around the y and not the others?

myk002 (changes requested):

plugins/spectate/spectate.cpp
cppcooper: typo

docs/plugins/spectate.rst
cppcooper: I have a couple after-changes to bump this up to v1.0.1a though so I'll get it when I make that commit

plugins/spectate/spectate.cpp
```diff
@@ -36,51 +47,323 @@ using namespace DFHack;
...
+                    range_exists[idx] = ranges[second] >= 0;
+                }
+                // lastly we set the start of the range
+                ranges[first] = ranges[previous] + (range_exists[idx] ? 1 : 0);
+            };
+
+            /// RANGE 0 (in view + working)
+            // grab valid working units
+            add_if([](df::unit* unit) {
+                return valid(unit) && Units::isCitizen(unit, true) && unit->job.current_job;
```
cppcooper, line 248:
```suggestion
                return valid(unit) && unit->job.current_job;
```
Perhaps for in view "working" we can dis-regard whether they are a citizen or not. They probably are.. But if not that might be interesting.

plugins/spectate/spectate.cpp
```diff
@@ -78,14 +77,41 @@ static PersistentDataItem pconfig;
...
 
 namespace SP {
     bool following_dwarf = false;
     df::unit* our_dorf = nullptr;
     int32_t timestamp = -1;
     std::default_random_engine RNG;
 
+    void DebugUnitVector(std::vector<df::unit*> units) {
+        for (auto unit : units) {
+            INFO(plugin).print("[id: %d]\n animal: %d\n hostile: %d\n visiting: %d\n",
```
myk002, line 107: you can change the verbosity level to `DEBUG` and wrap the whole for loop in an
```
if (DBG_NAME(plugin).isEnabled(DebugCategory::LDEBUG)) {
...
}
```
then you can just leave the debug statements in instead of commenting out the call to `DebugUnitVector`

plugins/spectate/spectate.cpp
```diff
@@ -78,14 +77,41 @@ static PersistentDataItem pconfig;
...
+            if (u->pos.x >= x1 && u->pos.x <= x2) {
+                if (u->pos.y >= y1 && u->pos.y <= y2) {
+                    if (u->pos.z >= z1 && u->pos.z <= z2) {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+    }
```
myk002, line 96: Could you move this to the *real* Units module and refactor `Units::getUnitsInBox()` to use it?

myk002 (changes requested):

plugins/spectate/spectate.cpp
cppcooper: Either way a code change and rebuild is required to run that code. I didn't provide debug commands for this cause it's super simple (compared to channel-safely).

cppcooper:
Shouldn't need to, it ought to resolve with a merge

plugins/spectate/spectate.cpp
myk002: why would a code change be required? and you shouldn't need to provide debug commands at all -- `debugfilter` provides that interface.

plugins/spectate/spectate.cpp
```diff
@@ -36,51 +45,336 @@ using namespace DFHack;
 using namespace Pausing;
 using namespace df::enums;
 
-void onTick(color_ostream& out, void* tick);
-void onJobStart(color_ostream &out, void* job);
-void onJobCompletion(color_ostream &out, void* job);
+struct Configuration {
+    bool debug = false;
+
```
myk002, line 50: unused?

myk002 (approved):

plugins/spectate/spectate.cpp
cppcooper: link, or elaborate? `debugfilter`?

plugins/spectate/spectate.cpp
cppcooper: I believe so

myk002 (approved):

PR #2383, opened 2022-11-08
Fixes JOB_STARTED event

cppcooper:


library/modules/EventManager.cpp
```diff
@@ -444,12 +444,14 @@ static void manageJobStartedEvent(color_ostream& out) {
...
-        for (df::job_list_link* link = df::global::world->jobs.list.next; link != nullptr; link = link->next) {
-            df::job* job = link->item;
+    for (df::job_list_link* link = df::global::world->jobs.list.next; link != nullptr; link = link->next) {
+        df::job* job = link->item;
+        bool send = false;
+        for (auto &key_value : copy) {
+            auto &handler = key_value.second;
             // the jobs must have a worker to start
-            if (job && Job::getWorker(job) && !startedJobs.count(job->id)) {
+            if (send || (job && Job::getWorker(job) && !startedJobs.count(job->id))) {
```
myk002, line 453: it might be simpler to put this if statement outside of the "copy" for loop. then you don't need to track the emplace logic so carefully since it will only be done once.

myk002 (changes requested):

myk002 (approved):

myk002:
will merge once CI passes. Just 46 build jobs to go before it's caught up!

PR #2392, opened 2022-11-12
Unit Tests: integrates googletest

cppcooper:
I couldn't figure out how to get ctest working, but the unit tests are integrated into CI at least. Just need more than one unit test now.


.github/workflows/build.yml
```diff
@@ -100,12 +100,15 @@ jobs:
       id: run_tests
       run: |
         export TERM=dumb
-        status=0
-        script -qe -c "python ci/run-tests.py --headless --keep-status \"$DF_FOLDER\"" || status=$((status + 1))
-        python ci/check-rpc.py "$DF_FOLDER/dfhack-rpc.txt" || status=$((status + 2))
-        mkdir -p artifacts
-        cp "$DF_FOLDER"/test*.json "$DF_FOLDER"/*.log artifacts || status=$((status + 4))
-        exit $status
+        if build-ci/library/tests/test-library; then
```
lethosor, line 103: Could `build-ci/library/tests/test-library` be a separate step instead? I think it's conceptually separate from this step, and the `if` condition makes things somewhat messy. A separate step would also avoid the need for `exit 1` at the end.

.gitmodules
```diff
@@ -28,3 +28,6 @@
 [submodule "depends/luacov"]
 	path = depends/luacov
 	url = ../../DFHack/luacov.git
+[submodule "depends/googletest"]
+	path = depends/googletest
+	url = https://github.com/google/googletest.git
```
lethosor, line 33: Please use relative URLs to allow for cloning over different protocols:
```suggestion
	url = ../../google/googletest.git
```

CMakeLists.txt
```diff
@@ -410,13 +405,64 @@ else()
...
+    message("BUILD TESTS: Core, Scripts")
+    set(BUILD_SCRIPT_TESTS ON FORCE)
+    set(BUILD_CORE_TESTS ON FORCE)
+endif()
+
+find_package(ZLIB REQUIRED)
+
+if(BUILD_CORE_TESTS)
+    include_directories(depends/googletest/googletest/include)
+endif()
```
lethosor, line 418: The tinyxml block is another dependency. It may not be obvious, but the entire tinyxml block was formerly `include_directories(depends/tinyxml)`. So I think the BUILD_TESTING stuff should go outside, to avoid interrupting the list of dependencies here.

cppcooper: You're gonna have to be specific, cause I put the stuff there to get variables setup before the depends subdir

cppcooper: I've taken a best guess at what you meant

CMakeLists.txt
```diff
@@ -410,13 +411,58 @@ else()
     set(DFHACK_TINYXML "dfhack-tinyxml")
 endif()
 
+find_package(ZLIB REQUIRED)
+
+if(BUILD_CORE_TESTS)
+    include_directories(depends/googletest/googletest/include)
```
myk002, line 417: this should be set as a property for test files (similar to how the COMPILE_FLAGS are set on line 438). it shouldn't be included in the header search directory for library code, right?

myk002 (commented):
Can we avoid defining BUILD_CORE_TESTS and instead have a `ninja test` target? The test target should not execute (or build gtest or any test files) when you just run `ninja`.

CMakeLists.txt
cppcooper: having it here allows anybody to start writing unit tests plugins and library alike

lethosor: Yeah, adding include paths based on cmake options is likely to cause build errors if e.g. someone introduces code that depends on the added include path. That effectively makes the cmake option required to build, at which point it isn't an option at all.

Enabling _targets_ based on cmake options is fine, and something we do already. So I think tests should be gated by an option, and the include path should be enabled on just those targets.

cppcooper: > so the tests are intended to be written in plugin and core code? not just test files? I don't agree with that plan. I think test code should be restricted to test files, like the example test in this PR.

no?? It means they can write unit tests for plugins and library code alike.. I don't think it's an easy feat to simply write unit tests in the code being tested with the whole: it needs to have a `main()` to function as a unit test.

The point is, that library tests will be with library code and plugin tests would be with the plugin code. Two different sub-directories that would otherwise need to be provided the include path separately.

cppcooper: > Enabling _targets_ based on cmake options is fine, and something we do already. So I think tests should be gated by an option, and the include path should be enabled on just those targets.

I can look into it

cppcooper:
> Can we avoid defining BUILD_CORE_TESTS and instead have a `ninja test` target? The test target should not execute (or build gtest or any test files) when you just run `ninja`.

Can you elaborate? I don't see how `BUILD_CORE_TESTS` and `ninja test` are related.

Is this just about excluding the test target(s) from `all`?

PR #2405, opened 2022-11-18
up MSVC warning level to /W3 /WX

ab9rf:
This makes MSVC warn at a level comparable to what we use on gcc for Linux builds

This basically the same thing for MSVC that we did in #1958 for Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 if(MSVC)
+    # increase warning level
+    add_definitions("/WX")
+    add_definitions("/W3")
+
     # disable C4819 code-page warning
     add_definitions("/wd4819")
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -74,6 +74,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 if(MSVC)
+    # increase warning level
```
myk002, line 77: "and treat warning as errors"

docs/changelog.txt
```diff
@@ -111,6 +111,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``gui.Screen.show()`` now returns ``self`` as a convenience
 - ``gui.View.getMousePos()`` now takes an optional ``ViewRect`` parameter in case the caller wants to get the mouse pos relative to a rect that is not the frame_body (such as the frame_rect)
 
+## Build configuration
```
myk002, line 114: this is failing because the list of valid headers is hard-coded elsewhere. I'd put this in `Internal`.

myk002 (changes requested):

docs/changelog.txt
ab9rf: oops sorry

ab9rf: Meh, `Internals` not `Internal`. Third time's the charm we can hope

myk002 (approved):

PR #2409, opened 2022-11-20
Autolabor refactor, phase 1

ab9rf:
This is a first step in refactoring autolabor, autohauler, and labormanager.

* All three plugins are moved into a single source directory
* The common "job state table" that all three plugins used is moved to a shared header


ab9rf:
Is a changelog appropriate for this since this has no user-facing implications?

myk002:
> Is a changelog appropriate for this since this has no user-facing implications?

yeah, no need for a changelog

myk002 (approved):

PR #2410, opened 2022-11-20
Add robots.txt to our HTML docs

### LINKED ISSUE #2368: DFHack docs SEO
Google searches for DFHack tools rarely take people to useful pages. Look into robots.txt and other techniques to help web crawlers grok our docs

https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html

myk002:
Fixes #2368

Google search does not grok our docs well. It often returns results for outdated versions. ReadTheDocs has an autogenerated robots.txt file we can use, but if we want to hide old DFHack docs from the crawler, we also hide it from the UI (in the flyout menu where doc readers can select the version of the docs they're reading). Therefore, it seems like the best idea to just create a custom robots.txt file to deploy to our docs site that only allows indexing of the `stable` docs. The other versions will still be there, they just won't show up in google search results.

Our generated sitemap.xml *does* appear to correctly give priority to `stable`, but it's easy to see that old docs still seem to outrank current docs in real search results. We may need a custom sitemap.xml as well, but let's see if just replacing robots.txt is enough first.

The robots.txt file in this PR won't take effect on docs.dfhack.org until it's deployed in the "stable" branch.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -256,6 +256,10 @@ html_favicon = 'docs/styles/dfhack-icon.ico'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['docs/styles']
 
+# A list of paths that contain extra files not directly related to the
+# documentation.
+html_extra_path = ['robots.txt']
+
 # Custom sidebar templates, maps document names to template names.
 html_sidebars = {
     '**': [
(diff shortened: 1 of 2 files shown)


lethosor:
The way I usually have handled doc changes that we need on `master` has been to branch off master, merge the needed change into master, then merge master into develop. That way the history between the two branches remains shared (as opposed to cherry-picking from one branch onto the other).

myk002:
> branch off master, merge the needed change into master, 

done

PR #2411, opened 2022-11-20
WIP autolabor refactor phase 2

ab9rf:
refactor autohauler and autolabor to eliminate/align duplicated/parallel code toward an eventual merger


plugins/autolabor/autohauler.cpp
```diff
@@ -86,37 +77,29 @@ REQUIRE_GLOBAL(world);
...
 
-// Yep...don't know what it does
 DFHACK_PLUGIN_IS_ENABLED(enable_autohauler);
 
-// This is the configuration saved into the world save file
-static PersistentDataItem config;
-
-// There is a possibility I will add extensive, line-by-line debug capability
-// later
 static bool print_debug = false;
```
myk002, line 82: it doesn't need to be done in this PR, but nowadays debug printing is handled by the central DFHack debug logging. See https://github.com/DFHack/dfhack/blob/develop/plugins/examples/persistent_per_save_example.cpp#L66

ab9rf: thanks for reminding me

ab9rf:
closing this because the branch is messed up, will reopen shortly

PR #2434, opened 2022-11-28
Implements plugin: channel-safely v1.1b

cppcooper:
Designations in marker mode don't flag their block(s) as having designations. The scan code attempts to avoid scanning every block in the map.

This combination of factors creates problems when, for example, loading a map where mega projects are in progress with many designations in marker mode as those designations won't get added to groups and thus cannot participate in checks.

To remedy this, I've removed the sometimes randomly checking a block anyways which still leads to strange behaviour and instead have added a `bool full_scan` parameter to function arguments to disable this optimization.. then made sure to build groups with that optimization disabled when first loading the map.

### DIFF
--- a/plugins/channel-safely/channel-groups.cpp
+++ b/plugins/channel-safely/channel-groups.cpp
@@ -119,7 +119,13 @@ void ChannelGroups::scan_one(const df::coord &map_pos) {
 }
 
 // builds groupings of adjacent channel designations
-void ChannelGroups::scan() {
+void ChannelGroups::scan(bool full_scan) {
+    static std::default_random_engine RNG(0);
+    static std::bernoulli_distribution sometimes_scanFULLY(0.15);
+    if (!full_scan) {
+        full_scan = sometimes_scanFULLY(RNG);
+    }
+
     // save current jobs, then clear and load the current jobs
     std::set<df::coord> last_jobs;
     for (auto &pos : jobs) {
@@ -138,9 +144,6 @@ void ChannelGroups::scan() {
         remove(pos);
     }
 
-    static std::default_random_engine RNG(0);
-    static std::bernoulli_distribution optimizing(0.75); // fixing OpenSpace as designated
-
     DEBUG(groups).print("  scan()\n");
     // foreach block
     for (int32_t z = mapz - 1; z >= 0; --z) {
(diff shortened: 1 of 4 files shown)


myk002 (approved):

myk002 (approved):

cppcooper:
I'm going to try and add a check for cave-in conditions over the next few days. I'm getting quite a few cave-ins with 4 diggers on this one project on the side of a cliff. The weird thing is it doesn't **look** to me like there should be cave-ins happening but they definitely are.. so I'm gonna try to spot them before they happen. If it works, I'll either add a feature to avoid them.. or make it the default behaviour or both.

myk002:
cave-in detection might be a bit ambitious to get in safely before -r8. Would it be ok to merge this PR as-is and take up further improvements for 0.50?

cppcooper:
I dunno, I think it's doable. I was working on an algorithm earlier, but I got hungry and decided to post on the PR instead of continuing to work it out.

basically.. The idea is to check the current job positions, scan their neighbours for empty space and detect cave-in conditions this way. There might even be a pre-processing step wherein candidates are determined according to distance to other jobs. Ie. If a job is one tile away from another job the tiles between them become candidates

myk002:
regardless of whether cave-in tracking goes in this release, a new feature should go in a new PR. every time this PR is rebased, GitHub loses track of what I've reviewed so far and I have to look through the whole thing again. smaller, focused PRs are much easier to manage. is there anything left for *this* PR?

cppcooper:
If cave-in is successful it'll be v1.2 so it wouldn't be a rebase, but if you want it as a separate PR instead of renaming this one, it can be done

myk002:
yes, please. in general, new set of changes -> new PR

ab9rf:
i agree with the plan to commit this now (if it's ready, it appears to be) and defer the next improvement until after the next release

ab9rf (approved):

cppcooper:
Yea, this passed my testing. If this is being merged, I may just wait until complaints come rolling in about cave-ins. I've been thinking about the windows build documentation, might be time to return to that.

myk002:
Ok, I'll merge it in. It might be best to wait just a little bit before refreshing the windows docs. The Windows build setup will be changing in just a few days.

PR #2445, opened 2022-11-30
correct QUERN construct labor

### LINKED ISSUE #2444: labormanager uses the wrong build labor for querns
querns actually use `HAUL_FURNITURE` but the job mapper is configured to use `MILLER`

ab9rf:
fixes #2444


myk002 (approved):
changelog entry plz

PR #2447, opened 2022-11-30
Fix and organize docs

### LINKED ISSUE #2439: Doc fixups
- fix numbered list in overlay dev guide
- move overlay dev guide to the dev section
- rename DFHack API to DFHack C++ API
- reorganize dev page listing? It seems jumbled

myk002:
Fixes #2439

I renamed "DFHack API" to "DFHack API concepts" to differentiate it from the Lua API reference. Is that accurate enough?


lethosor:
> I renamed "DFHack API" to "DFHack API concepts" to differentiate it from the Lua API reference. Is that accurate enough?

Sure, although I had intended to have an API reference in there as well, so I will probably rename it back once that's done.

PR #2451, opened 2022-12-01
autolabor/autohauler refactor

ab9rf:
This realigns both codebases to be more parallel to one another toward an eventual merge.

In addition both have been modified to use DFHack's internal debug messaging facility. Documentation has been updated to reflect this, since it may have end user impact.

### DIFF
--- a/plugins/autolabor/autohauler.cpp
+++ b/plugins/autolabor/autohauler.cpp
@@ -471,7 +471,7 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
 
         state_count[dwarf_info[dwarf].state]++;
 
-        INFO(cycle, out).print("Dwarf %i \"%s\": state %s\n",
+        TRACE(cycle, out).print("Dwarf %i \"%s\": state %s\n",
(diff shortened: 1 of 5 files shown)


plugins/autolabor/autohauler.cpp
```diff
@@ -583,33 +457,24 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
...
 
-        // Debug: Output dwarf job and state data
-        if(print_debug)
-            out.print("Dwarf %i %s State: %i\n", dwarf, dwarfs[dwarf]->name.first_name.c_str(),
-                      dwarf_info[dwarf].state);
-
-        // Increment corresponding labor in default_labor_infos struct
         state_count[dwarf_info[dwarf].state]++;
 
+        INFO(cycle, out).print("Dwarf %i \"%s\": state %s\n",
```
myk002, line 474: This will be output by default. Is that what you want? It seems like it might get a bit chatty.

ab9rf: oh, yeah, that one probably should be trace, thanks, will fix

myk002 (approved):

PR #2472, opened 2022-12-05
Updates Channel safely

cppcooper:
I've been working on cave-in prevention for several days now, and preliminary testing indicates this achieves it as well as fixing some issues with insta-digging inaccessible designations.


plugins/channel-safely/channel-manager.cpp
```diff
@@ -39,13 +54,87 @@ void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool
...
+            // we can skip designations already queued for insta-digging
+            if (CSP::dignow_queue.count(pos)) continue;
+            if (DFHack::isOpenTerrain(ttype) || DFHack::isFloorTerrain(ttype)) {
+                cavein_possible = config.prevent_caveins;
+                auto access = count_accessibility(miner->pos, pos);
+                if (access == 0) {
+                    if (config.insta_dig){
+                        CSP::dignow_queue.emplace(pos);
+                    } else {
+                        // todo: engage dig management, swap channel designations for dig
```
myk002, line 99: is this todo in this PR or later?

plugins/channel-safely/channel-manager.cpp
```diff
@@ -31,6 +45,7 @@ void ChannelManager::manage_group(const df::coord &map_pos, bool set_marker_mode
 }
 
 void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool marker_mode) {
+    const df::coord stubborn(97,71,135);
```
myk002, line 48: ?

plugins/channel-safely/channel-manager.cpp
```diff
@@ -39,13 +54,87 @@ void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool
...
+                        }
+                    }
+                }
+                manage_one(pos, true, false);
+            } else {
+                manage_one(pos, true, true);
+            }
+        } else {
+            manage_one(pos, true, marker_mode);
+        }
```
myk002, line 132: try inverting the condition and continuing to reduce the block nesting. e.g.
```
if (!cavein_possible) {
    manage_one(pos, true, marker_mode);
    continue;
}
...

**Review by myk002 — COMMENTED** [review]
version numbers as commit comments aren't very informative. could you write *what* was changed instead?

File: plugins/channel-safely/channel-manager.cpp (hunk 1 above)
Comment (cppcooper) [reply plugins/channel-safely/channel-manager.cpp:92]: likely later, depends if a solution is found for the disjointed play problem I brought up on the issue

**Comment by cppcooper** [comment]
I dropped the insta-dig feature from channel-safely in its PR. The dig-now command doesn't seem to work reliably, and trying my own variant didn't pan out. I left the code in (but always have the feature disabled; and not listed in documentation/status) on the chance I work up the willpower to debug it some more. At this point it isn't clear if there is a problem in channel-safely with regards to when it attempts to dig-now, or if there is only a problem with dig-now and the designations listed as *jobs only*.

File: docs/plugins/channel-safely.rst [hunk 4]
```diff
@@ -52,18 +52,17 @@ Commands
...
+:rebuild:           Rebuild the designation group data. You should also read Troubleshooting.
 
 Features
 --------
 
 :require-vision:    Toggle whether the dwarves need vision of a tile before channeling to it can be deemed unsafe. (default: enabled)
-:monitor:           Toggle whether to monitor the conditions of active digs. (default: disabled)
+:risk-averse:       Toggles whether to use cave-in prevention. Designations are activated in stages
+                    and their priorities along edges are modified. (default: enabled)
+:monitoring:        Toggle whether to monitor the conditions of active digs. (default: disabled)
```
myk002, line 63: The description of "monitoring" needs a bit more explanation, I think. How does monitoring change the behavior of the plugin from the player's perspective? Why would a player choose to enable it? If all it does is enable the `resurrect` option to work, maybe we can just get rid of it as an independent option and enable monitoring when `resurrect` is enabled?

docs/plugins/channel-safely.rst
```diff
@@ -74,3 +73,15 @@ Settings
...
+---------------
+
+If designations aren't switching correctly, try putting the designations into marker mode.
+Then press . (next) or resume. If you're debugging code you'll want these:
+
+[todo: make this a block of code]
+
+:filter1: ``debugfilter set Info channel manager``
+:filter2: ``debugfilter set Debug channel plugin``
+:filter3: ``debugfilter set Trace channel group``
```
myk002, line 87:
```suggestion
Then press . (next) or resume. If you're debugging code you'll want these::

    debugfilter set Info channel manager
    debugfilter set Debug channel plugin
    debugfilter set Trace channel group
```

myk002 (commented):
changelog entry?

docs/plugins/channel-safely.rst
cppcooper: It cancels active digs (ie. has a worker), based on safety. Safety being the possibility of a dwarf falling, or directly causing a cave-in. Although not all cave-ins will be prevented, some are... basicalllly.. considered intentional. If an active dig, along with a second active dig, will cause a tile to fall.. then one of the jobs is postponed. (The one that started second) This canceled job is a feature of monitoring.

It may very well be **[*effectively*]** non-functional at this point. It is unclear whether this is the case.

plugins/channel-safely/channel-groups.cpp
```diff
@@ -168,7 +220,15 @@ void ChannelGroups::scan(bool full_scan) {
                                     jobs.erase(map_pos);
                                 }
                                 block->designation[lx][ly].bits.dig = df::tile_dig_designation::No;
-                            } else if (is_dig_designation(block->designation[lx][ly]) || block->occupancy[lx][ly].bits.dig_marked) {
+                            } else if (is_dig_designation(block->designation[lx][ly]) || block->occupancy[lx][ly].bits.dig_marked ) {
```
cppcooper, line 223:
```suggestion
                            } else if (is_dig_designation(block->designation[lx][ly]) || block->occupancy[lx][ly].bits.dig_marked) {
```

plugins/channel-safely/channel-groups.cpp
```diff
@@ -169,6 +221,14 @@ void ChannelGroups::scan(bool full_scan) {
                                 }
                                 block->designation[lx][ly].bits.dig = df::tile_dig_designation::No;
                             } else if (is_dig_designation(block->designation[lx][ly]) || block->occupancy[lx][ly].bits.dig_marked) {
+                                if (!is_channel_designation(block->designation[lx][ly])) {
+                                    if (df::map_block* block_above = Maps::getBlock(bx, by, z+1)) {
```
cppcooper, line 225: move this pointer declaration to the outer loops

Issue #2338, opened 2022-10-14
Enabled state tracking and restoration across all tools


myk002:
As part of the usability overhaul (#2225), we want to be able to make the enabled state of DFHack tools persistent so players can enable a tool once and it will stay that way across reloads until they turn it off. This will reduce the need to edit init scripts to manually maintain the desired state for DFHack tools. The DFHack control panel (#2213) depends on this kind of persistent behavior across DFHack tools.

### Goals

- The player should be able to configure DFHack tools in-game and to have that configuration persisted across program reloads (for globally scoped tools) or savegame reloads (for save-scoped tools). No init file editing should be required (though the player can choose to replace/override the system proposed in this document with init script configuration).
- The player should not have to care whether a tool is a plugin or a Lua script. The player interface should be the same.

### Challenges

1. Whereas plugins have the plugin_init() hook that they could use to restore state, scripts have no such hook. We need to design a mechanism that will allow scripts to bootstrap themselves.
2. In order to know when a tool should be re-enabled, we need to know whether the tool is global or per-save. Right now there is no way to programmatically determine that.
3. Plugins export whether they are enabled but scripts do not. We need to design a way for scripts to report their enabled status.

myk002:
most class (2) tools (UI enhancements) can be ported to the overlay framework and have their state managed by overlay.
we still have to have a solution for class (3) (per-save, persistent config) scripts.

myk002:
I'll write up more about alternatives I've considered, but let me jot down the solution I'm considering:

1. scripts can declare a global function named `script_init()` that is called on DFHack init, after plugins are initialized
2. all scripts are `reqscript()`ed and those functions are discovered and called, similar to how the overlay framework scans scripts for `OVERLAY_WIDGETS`
3. scripts can use the `script_init()` function to load global state or set up state transition hooks as required for loading data from a save

the code that scans scripts and calls their init methods can be a plugin to separate the functionality from core. `script_manager`.

myk002:
in addition, scripts can declare a global function named `isEnabled` which will return the `enabled` status for the script. when the `enable` command is run, scripts are scanned for this global function and it is called to get the current status. Other scripts/plugins can also query the enabled status of the script via this function.

PR #2473, opened 2022-12-06
Allow scripts to self-enable and automatically restore state

myk002:
Basic infrastructure for #2338

This simple plugin scans all scripts and performs the following actions:
- hooks global `onStateChange()` functions up to `dfhack.onStateChange`
- catalogs script `isEnabled()` functions so they can be surfaced via the `enable` command

Sample change to the prioritize script to implement `isEnabled()`:
```lua
function isEnabled()
    for k in pairs(get_watched_job_matchers()) do
        return true
    end
    return false
end
```


lethosor:
Does this need a plugin? For core functionality like this, I think it would be ok to install an onStateChange hook in dfhack.lua (I thought we did already, actually). I'm not sure if there's anything else you're using the plugin for here.

myk002:
The plugin is also acting as the liaison for the enable built-in for listing the state of enable-able scripts. The plugin is certainly light on functionally, though, and if there is a better place for the logic, we can move it

myk002:
A potential modification to this algorithm: instead of scripts declaring global onStateChange functions that are attached by script-manager, they could modify dfhack.onScriptChange themselves in the main script body. For example, [modtools/fire-rate](https://github.com/DFHack/scripts/blob/113d0a85f734f84d4ee4f584819705b26a5ec9f3/modtools/fire-rate.lua#L414) does this.

The advantage is that modified scripts get their hooks updated automatically, which would be useful during development or if scripts can be added by steam workshop while the game is running

The disadvantage is that the names are chosen by the scripts themselves, which can lead to naming collisions if the script authors are not consistent (or if a script with a similar relative path is visible in a different script_path). For example, modtools/fire-rate chose the key "fire-rate" whereas script-manager would have keyed it by 'modtools/fire-rate'.

myk002:
I should note that what I liked about it being a plugin was that the order of events was well defined. I've already discovered how bad an idea it is to load script code before plugins were initialized. The plugin structure guarantees that core and the plugins will be in an appropriate state.

lethosor:
> The plugin is also acting as the liaison for the enable built-in for listing the state of enable-able scripts. The plugin is certainly light on functionally, though, and if there is a better place for the logic, we can move it

How does it do that? It only has `plugin_init()` and `plugin_onstatechange()`, and I don't see any changes to PluginManager.

I think `enable` could integrate with PluginManager (or a new ScriptManager?) directly, rather than going through a plugin.

I think one of my general concerns here is that it shouldn't be possible to break core behavior with `unload script-manager`, and that seems to me like it would break some stuff.

> A potential modification to this algorithm: instead of scripts declaring global onStateChange functions that are attached by script-manager, they could modify dfhack.onScriptChange themselves in the main script body. For example, [modtools/fire-rate](https://github.com/DFHack/scripts/blob/113d0a85f734f84d4ee4f584819705b26a5ec9f3/modtools/fire-rate.lua#L414) does this.

Likewise, I think this is something that dfhack.lua could easily handle as well. It already loads scripts to support `enable`.

> 
I should note that what I liked about it being a plugin was that the order of events was well defined. I've already discovered how bad an idea it is to load script code before plugins are initialized. The plugin structure guarantees that core and the plugins will be in an appropriate state.

Plugin load order isn't defined. Depending on the system, `script-manager` could load before other plugins, or first, or last (depending on directory order). If we define it to be alphabetical, then any plugins after `sc` would still be loaded later. On the other hand, something in core could reliably order itself before or after all plugins load.

myk002:
> > The plugin is also acting as the liaison for the enable built-in for listing the state of enable-able scripts.
> 
> How does it do that? It only has `plugin_init()` and `plugin_onstatechange()`, and I don't see any changes to PluginManager.

`plugin_onstatechange()` calls `Lua::CallLuaModuleFunction(out, L, "plugins.script-manager", "reload");` on `SC_CORE_INITIALIZED`. That's where the plugin lua modules, scripts are scanned, and the `isEnabled()` cache is built. The `enable` command code in `Core.cpp` calls the script manager's Lua `list()` function to list enableable scripts.

> I think `enable` could integrate with PluginManager (or a new ScriptManager?) directly, rather than going through a plugin.
>
> I think one of my general concerns here is that it shouldn't be possible to break core behavior with `unload script-manager`, and that seems to me like it would break some stuff.

Yes, we could shift this into core by making script-manager.lua a library module and calling it's `reload` function from someplace in Core just before the `SC_CORE_INITIALIZED` event is sent. That should be fine.

> > A potential modification to this algorithm
> 
> Likewise, I think this is something that dfhack.lua could easily handle as well. It already loads scripts to support `enable`.

I do not think this functionality can go into dfhack.lua directly. dfhack.lua is loaded well before all plugins have finished initializing, leading to the same problem I had with `overlay` when I tried to do this same sort of scanning during `plugin_init`.

> > I should note that what I liked about it being a plugin was that the order of events was well defined.
> 
> Plugin load order isn't defined. Depending on the system, `script-manager` could load before other plugins, or first, or last (depending on directory order). If we define it to be alphabetical, then any plugins after `sc` would still be loaded later. On the other hand, something in core could reliably order itself before or after all plugins load.

Plugin load order wouldn't matter here -- no lua code is scanned during `plugin_init`. It's only during `SC_CORE_INITIALIZED` that lua is scanned.

I'll move the code out of a plugin and into core and we can see what that looks like.

lethosor:
Ok, so the purpose of the plugin was to handle the SC_CORE_INITIALIZED event, then? If dfhack.lua runs before that event fires, it could install an onStateChange handler to handle that event (there already is one: `dfhack.onStateChange.DFHACK_PER_SAVE`).

Putting the call to Lua in the necessary spot in Core.cpp works just as well, of course.

myk002:
> it could install an onStateChange handler
> 
> Putting the call to Lua in the necessary spot in Core.cpp works just as well, of course.

ha, that's true. Which would you prefer?

lethosor:
Either one. Putting it in Core makes it a bit easier to guarantee execution order, I suppose.

Another question: how do you plan to deal with scripts that are added to a script folder after the initialization step runs?

myk002:
Is there a way to detect that a script has been added without polling?

We might need to just document that if you've just hot-loaded a mod, you'll need to call script-manager.reload() (or some easy to type substitute)

We could mitigate this somewhat by not caching the isEnabled() functions; just do a full scan every time `enable` is run. This would at least pick up new enableable scripts. It's not like `enabled` is run so often that the cache would make a difference

lethosor:
Sorry I am late. ffd646462ab4897399cde9a5da5252add24ebe18 looks like a reasonable approach to me. I'm assuming invoking a script would also attempt to locate the script before bailing, even if it hadn't previously existed?

I'm happy with this being in core.

myk002:
> I'm assuming invoking a script would also attempt to locate the script before bailing, even if it hadn't previously existed?

That's correct. That's what it did before and the logic for searching for scripts upon invocation hasn't changed.

PR #2485, opened 2022-12-09
update google protobuf to 3.21.10

### LINKED ISSUE #2481: update protobuf-lite
our version of protobuf-lite is very old and should probably be updated

i was able to get it to compile with msvc 2022 (for v50.01) by whacking at it, but going forward we should plan on upgrading

@myk002 has 3.21.10 it working with 47.05 (see #2478) but this does not compile on windows (VS2022 using the v140 toolset from VS2015) and i have not been able to find a simple way to make it compile, bunch of stuff related to dllimports that i have yet to make sense of. my guess is that this newer version has not maintained compatibility with the downlevel toolset

myk002:
Fixes #2481


ab9rf (commented):
please hold this until i have a chance to check if this works with msvc v140, i don't want to break our abilty to build a 47.05 release should an -r9 become necessary

myk002:
aborting this attempt at upgrading protobuf. the linkage errors are too much to overcome right now

PR #2490, opened 2022-12-11
make tree compatible with new msvc compiler in cross compilation env

myk002:
This PR will likely prevent us from building with old msvc, but we need these changes for the updated compiler used in our buildmaster build env (which is also required for DF 50 compatibility). if we need to do another 0.47.05 release, it will have to be done on top of the v0.47.05 branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ endif()
 
 if(WIN32)
     if((NOT MSVC) OR (MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1934))
-        message(SEND_ERROR "MSVC 2015 or 2017 is required")
+        message(SEND_ERROR "MSVC 2022 is required")
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ if(UNIX)
 endif()
 
 if(WIN32)
-    if((NOT MSVC) OR (MSVC_VERSION LESS 1900) OR (MSVC_VERSION GREATER 1919))
+    if((NOT MSVC) OR (MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1934))
         message(SEND_ERROR "MSVC 2015 or 2017 is required")
     endif()
 endif()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ if(UNIX)
 endif()
 
 if(WIN32)
-    if((NOT MSVC) OR (MSVC_VERSION LESS 1900) OR (MSVC_VERSION GREATER 1919))
+    if((NOT MSVC) OR (MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1934))
         message(SEND_ERROR "MSVC 2015 or 2017 is required")
```
ab9rf, line 65: the message should say "MSVC 2022 is required"

ab9rf (changes requested):

CMakeLists.txt
ab9rf: i recommend this diff:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 37e657912..a9428a2d2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,8 +61,8 @@ if(UNIX)
 endif()

 if(WIN32)
-    if((NOT MSVC) OR (MSVC_VERSION LESS 1900) OR (MSVC_VERSION GREATER 1919))
-        message(SEND_ERROR "MSVC 2015 or 2017 is required")
+    if((NOT MSVC) OR (MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1934))
+        message(SEND_ERROR "MSVC 2022 is required")
     endif()
 endif()

@@ -102,10 +102,10 @@ endif()

 # Automatically detect architecture based on Visual Studio generator
 if(MSVC AND NOT DEFINED DFHACK_BUILD_ARCH)
-    if(${CMAKE_GENERATOR} MATCHES "Win64")
-        set(DFHACK_BUILD_ARCH "64")
-    else()
+    if ( (${CMAKE_GENERATOR} MATCHES "Win32") OR (${CMAKE_GENERATOR} MATCHES "x86"))
         set(DFHACK_BUILD_ARCH "32")
+    else()
+        set(DFHACK_BUILD_ARCH "64")
     endif()
 else()
     set(DFHACK_BUILD_ARCH "64" CACHE STRING "Architecture to build ('32' or '64')")
```

myk002: done

ab9rf (approved):

PR #2493, opened 2022-12-11
migrate usage of set-output (required by github)

myk002:
as per https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


lethosor:
@myk002 I think this sort of thing could probably go in the "CI" project

myk002:
fair enough

PR #2495, opened 2022-12-11
Add math.floor to module example in Lua API

wolfboyft:
Since non-integers are rejected.


myk002 (approved):

myk002:
wait, doesn't that have to check to make sure it's not rounding down to zero?  (unless the counter was zero to begin with?)

wolfboyft:
Ah, true! Possibly max(x, 1) on its own would be good enough, but ye I'll get that in

myk002 (approved):

PR #2505, opened 2022-12-17
Imtui

20k:
Hello! This is a branch to potentially use the imtui fork of ImGui for DFHack scripts, so that they can have a nice modern UI behind them. Myk002 asked me to file a PR for this to keep a record

So far its all gone surprisingly smoothly:

[button.webm](https://user-images.githubusercontent.com/755197/208227805-3e376a88-09d0-41a3-9da8-a65a0f2784c5.webm)

[multiwin.webm](https://user-images.githubusercontent.com/755197/208228123-86eed276-7387-4ed6-be86-a2794058f75d.webm)

[colourswork.webm](https://user-images.githubusercontent.com/755197/208255309-68afee5d-dc51-48d0-a581-255f828b01cb.webm)

On a technical note, each viewscreen has its own imgui context, which means that every view is fully independent of every other view. The other alternative would be to have one global imgui context (which would allow different viewscreens to have interacting windows, for eg drag + drop), but this adds more complexity

Todo:

- [X] Basic Text Rendering
- [X] Colour internally
- [X] Reference mechanism for APIs that take pointers

Basic widgets functionality:

- [X] Button
- [X] Checkbox
- [X] PushStyleColor/PopStyleColor (necessary for eg coloured buttons)
- [x] Selectable (necessary for full width buttons)
- [x] Input boxes (for search)
- [x] Rework current colour system to reuse existing lua constants

Tedious but easy widget functionality:
- [x] IsItem* + IsWindow*

More difficult
- [x] Keyboard Controls. ImGui has this built in, so I'm hopeful that we can piggyback off this

Things that will inherently never work:
- [ ] Anything involving colour, like ColorPicker3

Technical Questions:
- [x] Is one context per viewscreen the right approach? 
- [x] Overlay interop
- [x] Non lua plugins


library/LuaApi.cpp
```diff
@@ -1617,6 +1617,45 @@ static const luaL_Reg dfhack_gui_funcs[] = {
...
+{
+    ImGui::End();
+}
+
+static void imgui_text(std::string str)
+{
+    ImGui::Text("%s", str.c_str());
+}
+
+static bool imgui_button(std::string name)
```
myk002, line 1644: Note that simple intermediate functions like this are not necessary. You could just WRAPM the original module API

CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 if(MSVC)
     # increase warning level and treat warnings as errors
-    add_compile_options("/WX")
+    #add_compile_options("/WX")
```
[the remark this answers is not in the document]
20k: Aha! I was unaware that you could do this for a specific project, I had a vague hunt around for this previously, thanks

library/LuaApi.cpp
20k: Aha ok! I'll keep this in mind for simple stuff - things like buttons do have default arguments that I'll add down the line

20k:
[dfoverlay.webm](https://user-images.githubusercontent.com/755197/208272435-be921706-f546-42f8-9450-c21da0d5aebf.webm)

Lots more progress, putting together a basic overlay config example

20k:
[overlayintegrations.webm](https://user-images.githubusercontent.com/755197/209250512-799f2695-f014-4ad1-bb4d-854c64047d79.webm)

Current state: the high level integration is mostly (entirely?) done, barring any bugs or unknown unknowns. ImGui windows respect viewscreen sorting, and the technical/architectural work is mostly done. So the current state is that there are a couple of minor local bugfixes, and lots of widgets to add. And the code needs a lot of cleanup

I've been more thoroughly keeping track of what I'm working on over here in case anyone wants more detail of what's to be done generally

https://trello.com/b/zW9Oc48o/df

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "depends/googletest"]
 	path = depends/googletest
 	url = ../../google/googletest.git
+[submodule "depends/imgui"]
+	path = depends/imgui
+	url = https://github.com/20k/dfhack-imgui.git
```
myk002, line 36: be sure to use a relative path so the submodule uses the same transport mechanism as the parent fork.

also, before we merge this, we'll want to fork imgui into the DFHack org like most of the dependencies in this file

CMakeLists.txt
```diff
@@ -419,6 +419,7 @@ include_directories(depends/tthread)
 include_directories(${ZLIB_INCLUDE_DIRS})
 include_directories(depends/clsocket/src)
 include_directories(depends/xlsxio/include)
+include_directories(depends/imgui)
```
myk002, line 422: will plugin code be using this API directly, or will it just be through the DFHack module? If the latter, we should probably restrict the include to target that directly depend on it. Something like:
```
target_include_directories(imgui INTERFACE depends/imgui)
```
(this would have to appear below the `add_subdirectory(imgui)` line below so the target exists

library/CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ set(MODULE_HEADERS
     include/modules/Graphic.h
     include/modules/Gui.h
     include/modules/GuiHooks.h
+    include/modules/ImTuiImpl.h
```
myk002, line 132: As a module name, I'd suggest something like Widgets.h or UI.h

CMakeLists.txt
20k: I spent a while fixing up the C++ support, so the current design should directly allow for plugins to use ImGui directly without many restrictions. I do need to add some testing, but yes ImGui itself would be directly available to any plugins as long as they also decorate their render/feed code correctly from the module itself

myk002:
saving this work in DFHack/dfhack/imtui

PR #2517, opened 2022-12-22
[WIP] Build docs

cppcooper:
This is a stab at organizing the information we throw at new devs into new files.

edit: the commits are not staying

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 
 # any build folders
 build*/
+!docs/dev/building/
 nix
 buntu
 build/VC2010
--- a/docs/dev/building/Compile.rst
+++ b/docs/dev/building/Compile.rst
@@ -227,7 +227,7 @@ host when you want to reattach::
 
     docker start -ai dfhack-win
 
-If you edit code and need to rebuild, run ``dfhack-make`` and then ``ninja install``.
+If you edit code and need to rebuild, run ``dfhack-make install``.
 That will handle all the wineserver management for you.
 
 .. _compile-windows:
--- a/docs/dev/building/Compile.rst
+++ b/docs/dev/building/Compile.rst
@@ -86,44 +86,23 @@ assistance.
   are also able to help with any submodule-related (or Git-related) issues
   you may encounter.
 
-Dependencies
-------------
+All Platforms
+=============
+Before you can compile the code you'll need to configure your build with cmake. Some IDEs can do this,
+but from command line is the usual way to do this; thought the windows section below points out some
+windows batch files that can be used to avoid opening a terminal/command-prompt.
 
-If you haven't already checked, this would be a good point to ensure you have
-all the `dependencies <build-dependencies>`.
-
-
-Contributing to DFHack
-----------------------
-
-For details on contributing to DFHack, including pull requests, code
-format, and more, please see `contributing-code`.
-
-
-Build settings
-==============
-
-This section describes build configuration options that apply to all platforms.
-If you don't have a working build environment set up yet, follow the instructions
-in the platform-specific sections below first, then come back here.
-
-Be sure to check out common `build options <build-options>`.
-
-Instructions
-============
-
-.. contents::
-  :local:
-  :depth: 1
+You should seek cmake's documentation online or via ``cmake --help`` to see how the command works. See
+the `build-options` page for help finding the DFHack build options relevant to you.
 
 .. _compile-linux:
 
 Linux
------
+=====
 On Linux, DFHack acts as a library that shadows parts of the SDL API using LD_PRELOAD.
 
 Build
-~~~~~
+-----
 Building is fairly straightforward. Enter the ``build`` folder (or create an
 empty folder in the DFHack directory to use instead) and start the build like this::
 
@@ -145,97 +124,10 @@ This will show a curses-based interface that lets you set all of the
 extra options. You can also use a cmake-friendly IDE like KDevelop 4
 or the cmake-gui program.
 
-Windows cross compiling from Linux
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. highlight:: bash
-
-If you are on Linux but need to produce a Windows build (for example, because the
-DF version you're working on isn't out for Linux yet), here is how you can build
-and run a Windows binary on Linux.
-
-.. contents::
-  :local:
-  :depth: 1
-
-Step 1: prepare a docker image
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-On your Linux host, install and run the docker daemon and then run these commands::
-
-    xhost +local:root
-    git clone https://github.com/BenLubar/build-env.git
-    cd build-env/msvc
-    docker build .
-    docker image ls
-    IMAGE_ID=<your image id>
-    docker run -it --env="DISPLAY" --env="QT_X11_NO_MITSHM=1" --volume=/tmp/.X11-unix:/tmp/.X11-unix --user buildmaster --name dfhack-win $IMAGE_ID
-
-The ``xhost`` command and ``--env`` parameters are there so you can eventually
-run Dwarf Fortress from the container and have it display on your host.
-
-Step 2: build DFHack
-^^^^^^^^^^^^^^^^^^^^
-
-The ``docker run`` command above will give you a shell prompt (as the `buildmaster` user) in the
-container. Inside the container, run the following commands::
-
-    git clone https://github.com/DFHack/dfhack.git
-    cd dfhack
-    git submodule update --init
-    cd build
-    dfhack-configure windows 64 Release
-    dfhack-make
-
-Inside the ``dfhack-*`` scripts there are several commands that set up the wine
-server. Each invocation of a windows tool will cause wine to run in the container.
-Preloading the wineserver and telling it not to exit will speed configuration and
-compilation up considerably (approx. 10x). You can configure and build DFHack
-with regular ``cmake`` and ``ninja`` commands, but your build will go much slower.
-
-Step 3: copy Dwarf Fortress to the container
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-First, create a directory in the container to house the Dwarf Fortress binary and
-assets::
-
-    mkdir ~/df
-
-If you can just download Dwarf Fortress directly into the container, then that's fine.
-Otherwise, you can do something like this in your host Linux environment to copy an
-installed version to the container::
-
-    cd ~/.steam/steam/steamapps/common/Dwarf\ Fortress/
-    docker cp . dfhack-win:df/
-
-Step 4: install DFHack and run DF
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Back in the container, run the following commands::
-
-    cd dfhack/build
-    cmake .. -DCMAKE_INSTALL_PREFIX=/home/buildmaster/df
-    ninja install
-    cd ~/df
-    wine64 "Dwarf Fortress.exe"
-
-Other notes
-^^^^^^^^^^^
-
-Closing your shell will kick you out of the container. Run this command on your Linux
-host when you want to reattach::
-
-    docker start -ai dfhack-win
-
-If you edit code and need to rebuild, run ``dfhack-make install``.
-That will handle all the wineserver management for you.
-
 .. _compile-windows:
 
 Windows
--------
-Build
-~~~~~
+=======
 There are several different batch files in the ``win32`` and ``win64``
 subfolders in the ``build`` folder, along with a script that's used for picking
 the DF path. Use the subfolder corresponding to the architecture that you want
@@ -263,8 +155,8 @@ solution file(s):
 
 Then you can either open the solution with MSVC or use one of the msbuild scripts.
 
-Building/installing from the Visual Studio IDE
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Visual Studio IDE
+-----------------
 After running the CMake generate script you will have a new folder called VC2015
 or VC2015_32, depending on the architecture you specified. Open the file
 ``dfhack.sln`` inside that folder. If you have multiple versions of Visual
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@
 
 # any build folders
 build*/
+!docs/dev/building/
```
myk002, line 10: what is this for?

cppcooper: It was to track that directory.

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+
+    cd build
+    ccmake .. -G Ninja
+    ninja install
+
+This will show a curses-based interface that lets you set all of the
+extra options. You can also use a cmake-friendly IDE like KDevelop 4
+or the cmake-gui program.
+
+Windows cross compiling from Linux
```
myk002, line 148: I think this should be on the same header level as the main operating systems, since it is a significantly different use case

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+    IMAGE_ID=<your image id>
+    docker run -it --env="DISPLAY" --env="QT_X11_NO_MITSHM=1" --volume=/tmp/.X11-unix:/tmp/.X11-unix --user buildmaster --name dfhack-win $IMAGE_ID
+
+The ``xhost`` command and ``--env`` parameters are there so you can eventually
+run Dwarf Fortress from the container and have it display on your host.
+
+Step 2: build DFHack
+^^^^^^^^^^^^^^^^^^^^
+
+The ``docker run`` command above will give you a shell prompt (as root) in the
```
myk002, line 180:
```suggestion
The ``docker run`` command above will give you a shell prompt (as the `buildmaster` user) in the
```

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+    dfhack-configure windows 64 Release
+    dfhack-make
+
+Inside the ``dfhack-*`` scripts there are several commands that set up the wine
+server. Each invocation of a windows tool will cause wine to run in the container.
+Preloading the wineserver and telling it not to exit will speed configuration and
+compilation up considerably (approx. 10x). You can configure and build DFHack
+with regular ``cmake`` and ``ninja`` commands, but your build will go much slower.
+
+Step 3: copy Dwarf Fortress to the container
```
myk002, line 196:
```suggestion
Step 3: copy or mount Dwarf Fortress to the container
```

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+First, create a directory in the container to house the Dwarf Fortress binary and
+assets::
+
+    mkdir ~/df
+
+If you can just downlaod Dwarf Fortress directly into the container, then that's fine.
```
myk002, line 204:
```suggestion
If you can just download Dwarf Fortress directly into the container, then that's fine.
```

docs/dev/building/Compile.rst
cppcooper: How about under the heading: Docker?
Cause otherwise I have the differing opinion that this is the best hierarchy.

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+
+Other notes
+^^^^^^^^^^^
+
+Closing your shell will kick you out of the container. Run this command on your Linux
+host when you want to reattach::
+
+    docker start -ai dfhack-win
+
+If you edit code and need to rebuild, run ``dfhack-make`` and then ``ninja install``.
```
myk002, line 230:
```suggestion
If you edit code and need to rebuild, run ``dfhack-make install``.
```

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,456 @@
...
+
+Be sure to check out common `build options <build-options>`.
+
+Instructions
+============
+
+.. contents::
+  :local:
+  :depth: 1
+
```
myk002, line 118: since this is the main focus of the page now, I think we can get rid of this header and move everything below it up a header level.

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+
+There are a variety of other settings which you can find in CMakeCache.txt in
+your build folder or by running ``ccmake`` (or another CMake GUI). Most
+DFHack-specific settings begin with ``BUILD_`` and control which parts of DFHack
+are built.
+
+Typical usage may look like::
+
+    # Plugin development with updated documentation
+    cmake ./ -G Ninja -B builds/debug-info/ -DCMAKE_INSTALL_PREFIX=<path to DF> -DCMAKE_BUILD_TYPE:string=RelWithDebInfo -DBUILD_DOCS:bool=ON -DBUILD_PLUGINS=1
```
myk002, line 19:
```suggestion
    cmake .. -G Ninja -DCMAKE_INSTALL_PREFIX=<path to DF> -DCMAKE_BUILD_TYPE:string=RelWithDebInfo -DBUILD_DOCS:bool=ON -DBUILD_PLUGINS=1
```

myk002, line 19: I wouldn't call this "typical usage" with the custom build dir specified like that, but this is probably still legal usage.

docs/dev/building/Compile.rst
cppcooper: I disagree, this makes for a good landing spot for anyone looking for just the instructions. The information would of course be visible from the top level, but it does add clutter. Have you taken a look at the html? The index links to such sections.

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+Build type
+----------
+Release/Debug, this is the type of build you want. This controls what information
+about symbols and line numbers the debugger will have available to it.
+
+Variable: ``CMAKE_BUILD_TYPE``
+
+Usage::
+
+    cmake .. -DCMAKE_BUILD_TYPE:string=RelWithDebugInfo
```
myk002, line 76:
```suggestion
    cmake .. -DCMAKE_BUILD_TYPE:string=RelWithDebInfo
```

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Options:
+
+* Release
+* RelWithDebugInfo
```
myk002, line 81:
```suggestion
* RelWithDebInfo
```

.gitignore
myk002: ah, that makes sense : ) since "building" is so overloaded in DF and DFHack, though, I'd suggest moving the files from `docs/dev/building/` to `docs/dev/compile/`

docs/dev/building/Compile.rst
myk002: I think I'd be happy as long as it shows up in the TOC on this page.

docs/dev/building/Compile.rst
myk002: I did most of the reviewing from the generated HTML. The current TOC on the Building page looks like this:
```
Compilation
How to get the code
Build settings
Instructions
```

I'm saying this would save readers a click if instead of "Instructions" we let them directly access the instructions for the platform they're interested in.

cppcooper: You've convinced me actually. I had considered this originally, but decided against it because I wanted to reinforce that there are sections relevant to everyone. I thought all the submodules being a gotcha was a good reason to avoid adding platforms for users to instantly spot and click and end up not populating submodules. So that was my rational, if you agree with it I'll stick with it otherwise move them up a level

myk002: yeah, I see what you're saying, and that's a good point. How about this:
- Keep `How to get the code` at the top
- Remove the `Dependencies` and `Build Settings` sections (which are both just stubs pointing to other pages)
- Move the platforms up a level (get rid of the `Instructions` header)
- At the start of each platform section, have a short sentence like:
```
Be sure you have `downloaded the code <How to get the code>`_, installed the `build-dependencies`, and configured the `build-options`.
```
Then even if they click straight to their platform, they won't miss common information.

cppcooper:
probably need to update the quick links section on the main page

docs/dev/building/Compile.rst
lethosor: My vote is for "Windows cross-compiling from Linux". No opinion on the hierarchy (nested or sibling with "Windows" are fine with me).

cppcooper: I need a tie breaker

.gitignore
cppcooper: I hadn't considered someone reading "build" and thinking -> buildings (with dfhack)

Though, even now, I don't think it is much to worry about since it is nested under development. Nowadays [I think this is fair to say] virtually everyone in pc gaming knows of "builds" of the game too through version numbering.

docs/dev/building/Compile.rst
cppcooper: That would take us back to that idea of "including the same text in multiple locations"

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+    configuration utility ``ccmake``.
+
+Generator
+---------
+For the uninitiated, the generator is what allows cmake to, of course, generate
+visual studio solution & project files, a makefile, or anything else.
+Your selection of generator comes down to preference and availability.
+
+Visual Studio
+=============
```
lethosor, line 38: The underlines here don't match our usual convention of `===` being a parent of `---`. I would suggest switching them so it's more clear to readers of the source code.

i.e. in this particular case:
```suggestion
-------------
```

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+=====
+The generally preferred build system where available.
+
+example::
+
+    cmake .. -G Ninja
+
+Install Location
+----------------
+This of course uses the default cmake variable.
```
lethosor, line 56:
```suggestion
This is the location where DFHack will be installed.
```

(not sure the current text provides much information)

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+Usage::
+
+    cmake .. -DCMAKE_INSTALL_PREFIX=<path to df>
+
+The path to df will of course depend on your system. If the directory exists it is
+recommended to use ``~/.dwarffortress`` to avoid permission troubles.
+
+Build type
+----------
+Release/Debug, this is the type of build you want. This controls what information
```
lethosor, line 69:
```suggestion
This is the type of build you want. This controls what information
```

(valid types are listed below, and I don't want to confuse people by suggesting `Debug` will work)

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+Variable: ``BUILD_LIBRARY``
+
+Usage::
+
+    cmake .. -DBUILD_LIBRARY:bool=OFF
+    cmake .. -DBUILD_LIBRARY=0
+
+Testing
+-------
+Regression testing will be arriving in the future, but for now there are only tests written in lua.
```
lethosor, line 111: Didn't C++ unit tests get merged?

docs/dev/building/Options.rst
```diff
@@ -0,0 +1,152 @@
...
+    cmake .. -DCMAKE_BUILD_TYPE:string=RelWithDebInfo
+
+Options:
+
+* Release
+* RelWithDebInfo
+
+Target architecture (32/64-bit)
+---------------------------------------
+If need 32-bit binaries or are looking to be explicit about building 64-bit.
```
lethosor, line 85: wording (I know this is a draft still)

.gitignore
lethosor: +1 for `compile` or maybe `build`.

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,435 @@
...
+  export PATH=/usr/local/bin:$PATH
+
+Docker
+======
+
+.. highlight:: bash
+
+You can use docker to build DFHack for windows. These instructions were developed
+on a linux host system.
```
lethosor, line 322: nit: caps

```suggestion
You can use docker to build DFHack for Windows. These instructions were developed
on a Linux host system.
```

docs/dev/building/Compile.rst
lethosor: Ok, my vote is "if it ain't broke, don't fix it". So `Docker` => `Windows cross compiling from Linux`. The instructions only apply to building for Windows currently, so `Docker` is ambiguous.

docs/dev/building/Options.rst
cppcooper: I haven't fully checked, but it seemed like the cmake wasn't setup to build the tests. I probably overlooked something.

docs/dev/building/Compile.rst
```diff
@@ -0,0 +1,435 @@
...
+    IMAGE_ID=<your image id>
+    docker run -it --env="DISPLAY" --env="QT_X11_NO_MITSHM=1" --volume=/tmp/.X11-unix:/tmp/.X11-unix --user buildmaster --name dfhack-win $IMAGE_ID
+
+The ``xhost`` command and ``--env`` parameters are there so you can eventually
+run Dwarf Fortress from the container and have it display on your host.
+
+Step 2: build DFHack
+--------------------
+
+The ``docker run`` command above will give you a shell prompt (as the `buildmaster` user) in the
```
lethosor, line 347: needs double backticks for code

```
/home/runner/work/dfhack/dfhack/docs/dev/building/Compile.rst:347: WARNING: undefined label: 'buildmaster'
```

docs/dev/building/Options.rst
cppcooper: Actually, I think Debug may actually work. Though perhaps it doesn't do as expected.

docs/dev/building/Compile.rst
myk002: yeah, but this "repeated text" is just hyperlinks to other sections, not actual original content. I don't think that crosses the threshold for "duplication"

docs/dev/building/Options.rst
myk002: C++ tests get built by default now, yes. They get run with `ninja && ninja test`

docs/dev/building/Options.rst
lethosor: It doesn't work at all on Windows, from my understanding (different ABI?)

PR #2521, opened 2022-12-24
Prepare for the hook API

myk002:
as per the files in the new DFHack/hooks repo. Note that this API does not break our current injection strategy. Once we have a version that uses this new API, we can turn down the SDL-related injection code.

This PR builds a stub shared object for Linux and OSX and installs it in the main DF dir (where the dfhooks API knows to look for it). That stub calls into hack/dfhack.(so|dylib).

For Windows, where all the dlls are in the main DF directory anyway, there isn't much reason to have an additional stub, so the hooks are baked directly into the dfhack library, and we can just change from building the library as SDL.dll to dfhooks.dll.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1481,11 +1481,12 @@ bool Core::Init()
     // make it obvious what's going on if someone checks the *.txt files.
     #ifndef LINUX_BUILD
         // Don't do this on Linux because it will break PRINT_MODE:TEXT
+        // this is handled as appropriate in Console-posix.cpp
         fprintf(stdout, "dfhack: redirecting stdout to stdout.log (again)\n");
-        fprintf(stderr, "dfhack: redirecting stderr to stderr.log (again)\n");
         freopen("stdout.log", "w", stdout);
-        freopen("stderr.log", "w", stderr);
     #endif
+    fprintf(stderr, "dfhack: redirecting stderr to stderr.log\n");
+    freopen("stderr.log", "w", stderr);
 
     Filesystem::init();
 
@@ -1760,6 +1761,14 @@ bool Core::Init()
     }
 
     cerr << "DFHack is running.\n";
+
+    {
+        auto L = Lua::Core::State;
+        Lua::StackUnwinder top(L);
+        Lua::CallLuaModuleFunction(con, L, "script-manager", "reload");
+        onStateChange(con, SC_CORE_INITIALIZED);
+    }
+
     return true;
 }
 /// sets the current hotkey command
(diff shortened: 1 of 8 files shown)


library/Core.cpp
```diff
@@ -1481,11 +1481,12 @@ bool Core::Init()
     // make it obvious what's going on if someone checks the *.txt files.
     #ifndef LINUX_BUILD
         // Don't do this on Linux because it will break PRINT_MODE:TEXT
+        // this is handled as appropriate in Console-posix.cpp
         fprintf(stdout, "dfhack: redirecting stdout to stdout.log (again)\n");
-        fprintf(stderr, "dfhack: redirecting stderr to stderr.log (again)\n");
         freopen("stdout.log", "w", stdout);
-        freopen("stderr.log", "w", stderr);
     #endif
+    fprintf(stderr, "dfhack: redirecting stderr to stderr.log (again)\n");
```
lethosor, line 1488: The `(again)` was only for Windows.

Also, if you're moving this logic here, you should also remove it from `Hooks-darwin`, like you removed it from `Hooks-linux`

myk002: done

library/include/Core.h
```diff
@@ -197,13 +203,7 @@ namespace DFHack
         struct Private;
         std::unique_ptr<Private> d;
 
-        bool Init();
-        int Update (void);
-        int Shutdown (void);
-        int DFH_SDL_Event(SDL::Event* event);
-        bool ncurses_wgetch(int in, int & out);
-
-        void doUpdate(color_ostream &out, bool first_update);
```
lethosor, line 206: Access to these functions was previously controlled via the `friend` declarations at the top of this class. It could be worth updating those declarations instead of making these public. Or if we really need to make them public, we can remove the `friend`s.

myk002: These functions are now being called from another library (libdfhooks.so), so I don't know if `friend` is the right mechanism anymore. We could probably find a way to keep them from being public if that makes the API cleaner though..

lethosor: Isn't `library/Hooks.cpp` part of our own library? That's the only place that directly calls the Core methods that I can see. Really, `friend` keeps things like plugins from calling `Core::Init()`.

myk002: Only in the windows build. On Linux and osx it's compiled into a separate "dfhooks" library.

We could have yet another bounce point that could befriend Core and keep its functions private, though

lethosor: Oh I see. Well, `friend` is only enforced by the compiler, not the linker, similar to private/public, so `friend` should still be capable of allowing functions to access private members even if those functions are in another translation unit, shared object, etc.

myk002: rewritten to be much cleaner. Thanks!

myk002: notably, we don't actually need to pull in the DFHack/hooks code as a submodule. The header file there is for the DF side of the API, not ours.

lethosor:
Is this able to be merged in as-is, or do we need to gate the new stuff behind some cmake flag?

myk002:
It can be merged as is. It doesn't change anything about our current SDL hook. Once the API is in place on the DF side, we just need to change the library name from SDL.dll to dfhooks.dll (and clean out all the SDL stuff, including what's in the startup scripts). We can merge after we get the ok from Toady.

myk002:
Approved by Tarn.

PR #2531, opened 2022-12-29
Baby GUI steps

myk002:
- implement Gui::getMousePos()
- implement Gui::getDepthAt()
- don't cache USE_GRAPHICS anymore (it was static before. now it's not)
- allow Label widgets to write the tiles that are specified in the pens (was using :char() instead of :tile())
- expose getMousePixels to Lua
- docs

### DIFF
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -2152,10 +2152,17 @@ df::coord Gui::getMousePos()
     df::coord pos;
     if (gps && gps->precise_mouse_x > -1) {
         pos = getViewportPos();
-        int32_t map_tile_pixels = gps->viewport_zoom_factor >> 2;
-        pos.x += gps->precise_mouse_x / map_tile_pixels;
-        pos.y += gps->precise_mouse_y / map_tile_pixels;
+        if (Screen::inGraphicsMode()) {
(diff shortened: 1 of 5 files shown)


library/modules/Gui.cpp
```diff
@@ -2151,16 +2152,31 @@ df::coord Gui::getMousePos()
     df::coord pos;
     if (gps && gps->precise_mouse_x > -1) {
         pos = getViewportPos();
-/* TODO: understand how this changes for v50
-        pos.x += gps->mouse_x_pixel / tile_width;
-        pos.y += gps->mouse_y_pixel / tile_height;
-*/
+        if (Screen::inGraphicsMode()) {
+            int32_t map_tile_pixels = gps->viewport_zoom_factor >> 2;
```
lethosor, line 2156: Maybe `/ 4` would be more intuitive? Hopefully not slower, assuming these are all integers.

lethosor (approved):
static LGTM

library/modules/Gui.cpp
myk002: fair enough. I guess I got into the habit of bit shifting when I could back when it mattered

PR #2533, opened 2022-12-30
adjust Lua widgets to new input scheme

myk002:
and implement context-aware mouse wheel scrolling

@warmist Is it worthwhile to keep `guidm.WorkshopOverlay`? I don't see any usage in `building-hacks`.


warmist:
Sorry i misremembered. Idea was to have scripters use buildingshacks + this overlay for easy "fake building sidebar" usage. E.g. https://github.com/warmist/df-mini-mods/blob/master/displayCase/display_case.lua#L7

PR #2534, opened 2022-12-30
skin our scrollbars to look and hover like DF's

myk002:
matches DF's style and behavior in both graphics and non-graphics modes.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -4347,13 +4347,11 @@ Scrollbar class
 
 This Widget subclass implements mouse-interactive scrollbars whose bar sizes
 represent the amount of content currently visible in an associated display
-widget (like a `Label class`_ or a `List class`_). By default they are styled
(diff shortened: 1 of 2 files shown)


lethosor (commented):
Looks good in-game. Do we know for sure that the tile IDs are stable?

myk002:
> Looks good in-game. Do we know for sure that the tile IDs are stable?

I can't be "sure" sure, but it's very likely that these ids in particular are stable. DF loads the first thousand textures or so before it loads world or mod textures. You can see the effects of this in the tile browser. Run the browser on the title screen, and there are only a few textures loaded, notably including these. After a world is loaded, there are more than 100,000 textures, but these ids stay the same.

myk002:
I should add that for any textures I reference in DFHack tools, I only select from the "basic" ones loaded pre-world.

lethosor:
Yeah, that's good. Maybe we should build an enum of at least the ones we care about eventually, in case they do shift around some day.

PR #2535, opened 2022-12-31
only do a full refresh when needed

myk002:
this significantly reduces CPU utilization when DFHack-owned screens are visible.

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -487,6 +487,10 @@ Window.ATTRS {
     draggable = true,
 }
 
+function Window:postUpdateLayout()
+    gui.Screen.request_full_screen_refresh = true
+end
+
 -------------------
 -- ResizingPanel --
 -------------------
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -442,7 +442,7 @@ end
 -- adding gaps between widgets according to self.autoarrange_gap.
 function Panel:postUpdateLayout()
     -- don't leave artifacts behind on the parent screen when we move
-    gui.Screen.request_full_screen_refresh = true
+    self.request_full_screen_refresh = true
 
     if not self.autoarrange_subviews then return end
 
@@ -464,6 +464,10 @@ end
 
 function Panel:onRenderFrame(dc, rect)
     Panel.super.onRenderFrame(self, dc, rect)
+    if self.request_full_screen_refresh then
+        df.global.gps.force_full_display_count = 1
+        self.request_full_screen_refresh = false
+    end
     if not self.frame_style then return end
     gui.paint_frame(dc, rect, self.frame_style, self.frame_title)
     if self.kbd_get_pos then
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -441,6 +441,9 @@ end
 -- if self.autoarrange_subviews is true, lay out visible subviews vertically,
 -- adding gaps between widgets according to self.autoarrange_gap.
 function Panel:postUpdateLayout()
+    -- don't leave artifacts behind on the parent screen when we move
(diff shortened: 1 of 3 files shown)


library/lua/gui/widgets.lua
```diff
@@ -487,6 +487,10 @@ Window.ATTRS {
     draggable = true,
 }
 
+function Window:postUpdateLayout()
+    gui.Screen.request_full_screen_refresh = true
```
lethosor, line 491: Couldn't this just set `gps.force_full_display_count` directly?

myk002: no, if `gps.force_full_display_count` isn't set *while rendering*, it doesn't have the intended effect (that is, it has no effect)

myk002: looking through g_src, there's lots of code that sets this variable to different values. I suspect that if we set it in `postUpdateLayout` (which is called from feed in this case), it just gets overwritten.

lethosor: What if we set it to 2? :stuck_out_tongue:

It's difficult to trace exactly what could be clobbering `force_full_display_count`. We don't appear to have anything that clears `force_full_display_count` ourselves, and the only mutation that could set it to 0 in g_src is a decrement in enabler.cpp, at the end of `renderer::display()`.

I still think this logic should be scoped to Window, though - as I understand it, it's the only situation (changing the footprint of a screen after it's created) where a full redraw is needed.

myk002: There can be artifact issues whenever a tile was written to by DFHack and then is not written to. Window is one case, but the original inspect-screen would have this issue too when it jumps around to avoid the cursor. I actually originally put this code into Window first before I realized that it's a Screen-level issue.

myk002: Note that the *best* fix would be to understand why the screen artifacts are happening in the first place and fix it there. I've experimented with various flags and buffers, and this is the only solution I've found that works.

myk002: Hold on a minute. To quote Shmee in Hook: "Lightning.. has just struck my brain"

I think the issue might be with the double buffer. Let me run some tests.

myk002: No, memsetting screen to 0, which would have cleared the double buffer, did not work. I don't see a better solution than setting `gps.force_full_display_count` at this point : (

myk002:
> this significantly reduces CPU utilization when DFHack-owned screens are visible.

Some data: this change reduces CPU utilization when a DFHack screen is up from 104% (across multiple CPUs) to 56% on my system. Baseline CPU utilization with no DFHack screen up is 46%.

library/lua/gui/widgets.lua
lethosor: It's not setting `force_full_display_count` that I'm objecting to, it's the split-up logic that I would like to avoid. Maybe it's not possible to avoid.

lethosor:
Yeah, full redraws are pretty expensive. Good to reduce them!

lethosor (approved):

library/lua/gui/widgets.lua
myk002: I understand, and I hear you. The reason I split it up across classes is that there could be many sources of the problem, but only one place to apply the solution (that I have found). For this particular case, we *could* set `gps.force_full_display_count` in `Panel:onRenderFrame()`, but I know there are other places this could happen, and I didn't want to end up copying `gps.force_full_display_count = 1` every place it comes up.

but perhaps I'm being pedantic here. the inspect-screen case is already solved (by inheriting from Window), and the other case I had in mind is overlay widgets changing size, which would have to be handled by `overlay` anyway since it wouldn't be running through the gui.Screen code.

Ok, let me consolidate the logic into `gui.Panel`, and if we find we need this more generic implementation, we can revert this next commit.

PR #2543, opened 2023-01-01
prep for Quietust's structure updates

myk002:
more fields have disapperated

@RosaryMala If you have work for this already prepped, we can merge yours and I can revert this one.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -2266,6 +2266,7 @@ bool Core::ncurses_wgetch(int in, int & out)
     }
     if(in >= KEY_F(1) && in <= KEY_F(8))
     {
+/* TODO: understand how this changes for v50
         int idx = in - KEY_F(1);
         // FIXME: copypasta, push into a method!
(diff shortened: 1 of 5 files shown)


library/Core.cpp
```diff
@@ -2266,6 +2266,7 @@ bool Core::ncurses_wgetch(int in, int & out)
     }
     if(in >= KEY_F(1) && in <= KEY_F(8))
     {
+/* TODO: understand how this changes for v50
```
lethosor, line 2269: What part of this is problematic? As far as I can tell, hotkeys are about the same.

myk002: the reference to `ui`. It's just a name change, but the new names aren't actually in the tree yet.

myk002: and `ui_sidebar_mode`

RosaryMala:
"Work" is a strong word for what I did. Just commented out things until it compiled with no added notes or anything.
I'll be reverting that and pulling this.

PR #2548, opened 2023-01-03
Display DFHack logo for the hotkeys overlay widget

myk002:
add Textures and DFSDL modules, load DFHack logo, and show it for the hotkeys overlay widget hotspot. 

undocumented for now since it's internal. in the future, perhaps this could morph into a dynamic texture loading facility for tools

image


library/modules/Textures.cpp
```diff
@@ -0,0 +1,208 @@
...
+static const std::vector<std::string> SDL_LIBS {
+    "SDLreal.dll",
+    "SDL.framework/Versions/A/SDL",
+    "SDL.framework/SDL",
+    "libSDL-1.2.so.0"
+};
+static const std::vector<std::string> SDL_IMAGE_LIBS {
+    "SDL_image.dll",
+    "SDL.framework/Versions/A/SDL_image",
+    "SDL.framework/SDL_image",
```
lethosor, line 32:
```suggestion
    "SDL_image.framework/Versions/A/SDL_image",
    "SDL_image.framework/SDL_image",
```
confirmed by downloading `df_47_05_osx.tar.bz2`

library/modules/DFSDL.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+DFSDL_Surface * DFSDL::DFSDL_ConvertSurface(DFSDL_Surface *src, const DFSDL_PixelFormat *fmt, uint32_t flags) {
+    return g_SDL_ConvertSurface(src, fmt, flags);
+}
+
+void DFSDL::DFSDL_FreeSurface(DFSDL_Surface *surface) {
+    g_SDL_FreeSurface(surface);
+}
+
+int DFSDL::DFSDL_SemWait(DFSDL_sem *sem) {
```
lethosor, line 114: What I had in mind (and meant on Discord) was just `SDL_SemWait` with C linkage... but I realize now that if we only need to _call_ SDL functions, we can update our own calls appropriately. So I think this should be fine. It's not compatible with any binaries out there that already call `SDL_*` C functions, but we shouldn't need to care about that.

myk002: I understood -- but I can't add SDL_SemWait here in this PR since Hooks-windows is still in the build. This is the framework, and I'll add the actual calls in the hooks PR

lethosor: Yep. And now I'm taking back my Discord suggestion because I realized you probably don't need to define `SDL_SemWait` after all - we can just adjust `df/custom/enabler.methods.inc` to call `DFSDL::DFSDL_SemWait()` instead. (Need to get the right header included somehow, but we're already doing that to allow it to call an SDL function)

plugins/lua/hotkeys.lua
```diff
@@ -41,6 +40,34 @@ function HotspotMenuWidget:overlay_trigger()
         mouseover=self.mouseover}:show()
 end
 
+local dscreen = dfhack.screen
+
+function HotspotMenuWidget:onRenderBody(dc)
+    local tpos = dfhack.textures.getDfhackLogoTexposStart()
+    local x, y = dc.x, dc.y
+
+    if tpos == -1 then
```
lethosor, line 49: Should this also be checking for graphics mode? e.g. if the texture was previously loaded but DF is now in classic mode

lethosor (approved):
static review LGTM

plugins/lua/hotkeys.lua
lethosor: Yes, but my point is I only see one being painted here at a time, and it looks to me like it's switched based on whether the texture was ever loaded, not whether the texture can be displayed right now. But I could be misunderstanding.

myk002: The first branch paints just the characters if the texture is unavailable since otherwise the `tpos+N` lines would refer to incorrect textures. The second branch paints both, suitable for both graphics and non-graphics modes. Notice that both `ch` *and* `tile` are populated in the second branch.

lethosor: Ohh, I did notice that a few days ago but forgot to ask about it when actually reviewing. Thanks!

PR #2551, opened 2023-01-03
refactor EditField text insertion

myk002:


library/lua/gui/widgets.lua
```diff
@@ -657,6 +657,12 @@ function EditField:onRenderBody(dc)
     dc:string((' '):rep(dc.clip_x2 - dc.x))
 end
 
+function EditField:insert(text)
```
lethosor, line 660: idea: might be nice to have an optional `cursor` param someday

lethosor (approved):

library/lua/gui/widgets.lua
myk002: I was thinking that as I wrote the docs, but then I looked at what it would take to bounds check the parameter and I thought..maybe I'll leave that refactoring for another day : p

Issue #2326, opened 2022-10-10
Move dfhack-owned files out of dfhack-config/


myk002:
like the default files.

what about the library files for professions, orders, and (soon) the control panel?

lethosor:
I thought the library files lived under hack/?

myk002:
I believe after discussion we decided to put them in `dfhack-config/foo/library` so users could remove them if they wanted to. I'm not sure that's a strong enough reason to keep them there, though.

myk002:
in fact, let's remove dfhack-config/ from the release archive entirely and create it on demand so users can upgrade DFHack without fear of losing their stored data. E.g. on OSX using Archive Manager, extracting the DFHack installation will *replace* the dfhack-config/ directory, not just overwrite the contents of it.

lethosor:
Sounds good to me

PR #2553, opened 2023-01-04
migrate from SDL interposing to the hooks API

myk002:
- change build library from SDL.dll to dfhooks.dll on windows
- remove SDL interposing code
- update code that looks for SDLreal.dll to instead look for SDL.dll
- fixes return value of SDL_Event hook to what is expected in g_src


docs/Installing.rst
```diff
@@ -28,14 +28,9 @@ both DF and DFHack.
...
 
-* DFHack only supports the SDL version of Dwarf Fortress. The "legacy" version
-  will *not* work with DFHack (the "small" SDL version is acceptable, however).
-* Windows XP and older are *not* supported, due in part to a
-  `Visual C++ 2015 bug <https://stackoverflow.com/questions/32452777/visual-c-2015-express-stat-not-working-on-windows-xp>`_
-
-The Windows build of DFHack should work under Wine on other operating systems,
-although this is not tested very often. It is recommended to use the native
-build for your operating system instead.
+The Windows build of DFHack should work under Wine on other operating systems.
```
lethosor, line 31: I think we should keep the XP note. Feels weird to have the only text under "Windows" be about other systems.

docs/Installing.rst
```diff
@@ -101,56 +96,29 @@ DFHack binary and system data is stored, a ``dfhack-config`` folder where user
...
-=================================
-
-There are :wiki:`several packs available <Utility:Lazy_Newb_Pack>` that include
-DF, DFHack, and other utilities. If you are new to Dwarf Fortress and DFHack,
-these may be easier to set up. Note that these packs are not maintained by the
-DFHack team and vary in their release schedules and contents. Some may make
-significant configuration changes, and some may not include DFHack at all.
+The recommended approach to upgrade DFHack is to remove the ``hack`` folder,
+then install the new version. This will ensure that any files that are only part
+of the older DFHack installation do not affect the new DFHack installation.
```
lethosor, line 121: We should keep the note about preserving dfhack-config - that is still relevant

lethosor: Oh, I just saw your comment at https://github.com/DFHack/dfhack/issues/2326#issuecomment-1371273593. That's fair, but I think removing the note should be done after that change is made (or alongside it).

myk002: Fair enough, even after we remove the directory from the archive and this specific issue goes away, we should still note that that's where all the user-written data is (apart from blueprints and exported stockpile settings, but those might move in someday too)

docs/Installing.rst
myk002: done

docs/Installing.rst
myk002: note that this PR won't be merged until we have a new DF binary in our hands -- otherwise it would break current dev. The dfhack-config PR will be merged before this one.

myk002: > We should keep the note about preserving dfhack-config - that is still relevant

done, but moved to the `Uninstalling` section where it is now more relevant.

myk002:
Tested working with 50.08 pre-release

PR #2554, opened 2023-01-04
Remove dfhack-config from release tarball

### LINKED ISSUE #2326: Move dfhack-owned files out of dfhack-config/

myk002:
Fixes #2326

don't include dfhack-config/ in release tarball -- create it at runtime and move files around so no dfhack-owned files are in there

This PR also fixes a long-standing (but invisible to the user) bug where the command ordering bias file for `gui/launcher` was being installed to the wrong location, resulting in a sort order in `gui/launcher` that was not taking the bias into account.


data/examples/README.md
```diff
@@ -1,9 +0,0 @@
-# The DFHack Example Configuration File Library
```
lethosor, line 1: Forgot to ask: did you intend to remove this? It does break the link at https://docs.dfhack.org/en/stable/docs/guides/examples-guide.html  - not sure off the top of my head where the files that page refers ended up.

myk002: good catch. I did intend to remove the README since it's no longer the root of a sprawling hierarchy but rather just a single file, but it would not be so out of place just to move it to `dfhack-config/init/examples`. I'll take care of that (and the linked doc)

PR #2555, opened 2023-01-05
Remove unused 32-bit Windows support

hawkowl:
Also prevents building on Windows 32-bit entirely for now, since all v50 are on Windows x64 (I put that in a separate commit so it is revertable if/when 32-bit support changes).

Changes are broken into individual commits for revertability.


ab9rf:
we can probably go forward with this at this point, i'm releasing my hold

myk002 (approved):

myk002:
I'll fix up the conflicts and merge. Thanks, @hawkowl !

PR #2556, opened 2023-01-05
framework for passthru, z-order swapping windows

myk002:
Provides a gui.ZScreen class that lua gui tools can inherit from instead of gui.Screen. It allows the parent viewscreen to receive events if the ZScreen subclass doesn't handle them. It also allows other ZScreen subclass screens to raise to the top if windows on those screens are clicked with the mouse. In addition to providing "real window system" behavior for DFHack windows, it also lets the player interact with the map while DFHack tool windows are being shown and lets the game progress while DFHack tools are up if the game is unpaused. Note that DFHack tools can still choose to inherit from gui.Screen if they want the previous "modal" behavior.

ZScreen subclasses should *not* override the `onInput()` function. Instead, they should delegate any input processing to contained widgets (including the main Window panel). This allows top-level input processing to be consistent across all ZScreens. namely:

- otherwise unhandled r-clicks or ESC keypresses close the ZScreen
- keyboard input is passed to the top ZScreen, and if it is unhandled, is sent directly to the first underlying non-ZScreen viewscreen (this is usually the DF viewscreen)
- left mouse clicks outside the ZScreen's window area are passed to the next viewscreen. if a lower ZScreen detects that the mouse click is within its window area, that ZScreen will be raised to be the top viewscreen and receives focus for further mouse and keyboard input

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -2456,6 +2456,17 @@ static int screen_findGraphicsTile(lua_State *L)
     }
 }
 
+static int screen_raise(lua_State *L) {
+    df::viewscreen *screen = dfhack_lua_viewscreen::get_pointer(L, 1, false);
+
+    // remove screen from the stack so it doesn't get returned as an output
+    lua_remove(L, 1);
+
+    Screen::raise(screen);
+
+    return 0;
+}
+
 static int screen_hideGuard(lua_State *L) {
     df::viewscreen *screen = dfhack_lua_viewscreen::get_pointer(L, 1, false);
     luaL_checktype(L, 2, LUA_TFUNCTION);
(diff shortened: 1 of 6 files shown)


myk002:
The `_MOUSE_X_DOWN` inhibition code prevents a mouse click from being signaled for more than one call to `feed()`. however, the inhibition is a two-step process so a lower ZScreen viewscreen receives the *same* `_MOUSE_X_DOWN` values as the viewscreen above it for the case where the upper viewscreen did not handle the click. The previous code started inhibiting immediately, making the mouse click "disappear" for the lower viewscreens.

myk002:
docs version of the initial comment is coming soon. I just wanted to get the code up here first.

library/LuaApi.cpp
```diff
@@ -2456,6 +2456,17 @@ static int screen_findGraphicsTile(lua_State *L)
     }
 }
 
+static int screen_raise(lua_State *L) {
```
lethosor, line 2459: Do we need the custom wrapper here? `raise()` takes a `df::viewscreen*` only and returns `void`, so it ought to be natively wrappable with `WRAPM` and should behave the same way, I think

myk002: The call to `dfhack_lua_viewscreen::get_pointer()` to convert from the Lua Screen table to the screen._native pointer does not seem to be automatic. At least when I tried a standard `WRAPM`, it built, but it didn't function.

lethosor: Oh, I missed the `lua` part of the class name in `dfhack_lua_viewscreen::get_pointer()`. Disregard.

PR #2558, opened 2023-01-06
refine and document ZScreen

myk002:


myk002:
ok, again merging to get some usability feedback. I think it's in a fairly good state now.

PR #2561, opened 2023-01-06
set-df-path: accept path on command line

ab9rf:
This enables specifying the pathname on the command line for `set_df_path.vbs`, which is a longstanding pet peeve of mine


lethosor (approved):
I'm good with this assuming both paths have been tested

ab9rf:
> I'm good with this assuming both paths have been tested

yes, i did test this both ways before submitting it. it's been a decade since i wrote vbscript, i had to look up how to do pretty much everything there

PR #2563, opened 2023-01-06
enable reveal for v50

ab9rf:
`reveal demon` is disabled due to double popup bug 
`reveal` also has guidance text that appears if the player is in graphics mode explaining the limitations of the new renderer


plugins/reveal.cpp
```diff
@@ -256,7 +259,14 @@ command_result reveal(color_ostream &out, vector<string> & params)
             revealed = DEMON_REVEALED;
     }
     is_active = nopause_state || (revealed == REVEALED);
+    bool graphics_mode = Screen::inGraphicsMode();
     con.print("Map revealed.\n");
```
myk002, line 263:
```suggestion
    con.print("Map revealed.\n\n");
```

plugins/reveal.cpp
```diff
@@ -256,7 +259,14 @@ command_result reveal(color_ostream &out, vector<string> & params)
...
+    if (graphics_mode) {
+        con.print("Note that in graphics mode, tiles that are not adjacent to open\n"
+                  "space will not render but can still be examined by hovering over\n"
+                  "them with the mouse. Switching to text mode (in the game settings)\n"
+                  "will allow the display the revealed tiles.\n");
```
myk002, line 268:
```suggestion
                  "will allow the display the revealed tiles.\n\n");
```

plugins/reveal.cpp
```diff
@@ -256,7 +259,14 @@ command_result reveal(color_ostream &out, vector<string> & params)
...
+    }
     if(!no_hell)
         con.print("Unpausing can unleash the forces of hell, so it has been temporarily disabled.\n");
```
myk002, line 271:
```suggestion
        con.print("Unpausing can unleash the forces of hell, so it has been temporarily disabled.\n\n");
```

myk002 (approved):

PR #2564, opened 2023-01-06
add Gui::getDFViewscreen and Lua binding

myk002:
useful for `devel/inspect-screen` and similar tools that want to know what the current DF screen is. it's not so obvious now that there can be multiple Z-layers


lethosor (approved):
Seems useful. Note that there is also a getViewscreenByType() if you know what specific screen you're looking for.

PR #2566, opened 2023-01-07
reenable autofarm and add persistence support

ab9rf:
note that persistence does not appear to work properly right now so persistence hasn't been fully tested; otherwise the plugin works

also updated to use the new debugging logger


myk002 (commented):
needs a few `.c_str()`'s to fix build issues on Linux, but otherwise looks good

ab9rf:
on hold until i figure out what's going on with persistence

PR #2567, opened 2023-01-07
Remote Fortress Reader updates

RosaryMala:
This is more or less working now? I can load up a fortress and view it. There's some bugs but it doesn't crash at least.


plugins/remotefortressreader/building_reader.cpp
```diff
@@ -316,7 +316,8 @@ void CopyBuilding(int buildingIndex, RemoteFortressReader::BuildingInstance * re
     material->set_mat_index(local_build->mat_index);
 
     remote_build->set_building_flags(local_build->flags.whole);
-    remote_build->set_is_room(local_build->is_room);
+    //FIXME: Figure out what to replace this with.
+    //remote_build->set_is_room(local_build->is_room);
```
myk002, line 320: There's a new `building->relations` list that indicates which civzones ("rooms") this building is associated with

RosaryMala: The only thing RFR was using this for was for things that were set as rooms but didn't have the room extents set. There's nothing that does that anymore, so I'll be removing the check entirely.

myk002 (approved):

PR #2571, opened 2023-01-08
Enabled probe, and fixed the one thing preventing it from compiling.

RosaryMala:
Before this can be considered updated, it would need to be able to find the currently selected building or creature from the game, but it does work as-is when the keyboard cursor is enabled in the game.


myk002:
I'd like to hold off merging this until getSelectedUnit() and friends are functional, just to make sure we aren't introducing crash bugs in when we exercise those code paths

myk002:
Tested with #2577. Everything checks out!

PR #2576, opened 2023-01-09
use an actual pin texture for ZScreen pins

myk002:
Pin textures contributed by RafaelKB on Discord (Thanks!)

## Before
Unpinned:
image
image

Pinned:
image
image

## After
Unpinned:
image
image

Pinned:
image
image

If the pin texture is unavailable/can't be loaded, the presentation will fall back to the previous tile. The click detection area will still be 2x2, but that's not necessarily a bad thing


myk002:
It might be better if the green pin were more vertical so they are visually distinct even for those with red/green color blindness. It might also help reinforce that the green means "pinned"

lethosor:
I actually prefer the size/shape from before - in particular, I'd rather have the icon extending into the window than out of the window, if it's that large.

myk002:
the screenshots may make it look large, but in-game it's actually quite small. So much so that I found the one-tile click area to be unusably small. Moreover, the single tile "switch" was easy to miss for such important functionality.

Here are a couple options:
- move back to single tile. I think this is too small, but I'm open to wider feedback.
- move the 2x2 icon inside the window boundary. easy change to make, but I'll have to go through and redo layout across gui screens to avoid the conflict.
- change to a 2x1 icon that sits flush with the frame. Is this a good compromise? I'll have to ask RafaelKB on Discord for ideas on a graphic in those dimensions.

myk002:
ok, merging for now, will get more feedback from pilot testers

lethosor:
I'm somewhat leaning towards 2x1 in the frame, at least in graphics mode, but open to feedback from testers

myk002:
how about after the alpha release, I put out a poll on reddit with a few options and see what people say?

lethosor:
Sure, sounds reasonable

PR #2577, opened 2023-01-09
get rudimentary getAny(Unit|Item|Building) working

myk002:
just implemented for the main view sheet, but it's enough to get started with


myk002:
Note to self: add getSelected passthrough for ZScreens to this PR

PR #2583, opened 2023-01-10
Remove os.execute and io.popen

arekatir:
I don't think anyone uses these functions currently and they're often removed from any program that has Lua scripting available.


myk002:
i'll try patching this in to the 0.47.05 branch and see if the tests all still work (they should)

arekatir:
pre-commit.ci autofix

myk002 (approved):

PR #2588, opened 2023-01-11
fix a crash in Screen when read or paint tile indices are negative

### LINKED ISSUE #2587: fix crash when painting (or getting) out of bounds tiles
specifically, when x == -1, but all bounds should be properly checked.

myk002:
Fixes #2587

### DIFF
--- a/library/modules/Screen.cpp
+++ b/library/modules/Screen.cpp
@@ -208,7 +208,7 @@ static bool doSetTile(const Pen &pen, int x, int y, bool map)
 
 bool Screen::paintTile(const Pen &pen, int x, int y, bool map)
 {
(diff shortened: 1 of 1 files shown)


ab9rf (commented):
I take it that the `if (screen > gps->screen_limit)` provides mostly adequate protection against writing off the other end, but I'll note that if either `x` or `y` is _wildly_ out of range, the calculation in `doSetTile_default` might possibly overflow, which will cause that test to fail to detect the out of bound condition.

myk002:
> might possibly overflow, which will cause that test to fail to detect the out of bound condition.

done

ab9rf (approved):
verified that these changes mitigate the reported crash

PR #2590, opened 2023-01-11
add untested tag to plugins

myk002:
and handle new tag in docs generation

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -695,7 +695,14 @@ local valid_script_flags = {
     scripts = {required = false},
 }
 
+local warned_scripts = {}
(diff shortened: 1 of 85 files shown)


library/lua/dfhack.lua
```diff
@@ -695,7 +695,14 @@ local valid_script_flags = {
     scripts = {required = false},
 }
 
+local warned_scripts = {}
 function dfhack.run_script(name,...)
+    if not warned_scripts[name] and require('helpdb').get_entry_tags(name).untested then
```
lethosor, line 700: I would put this in `run_script_with_env` so it fires if a script is `enable`d as well.

myk002: done

PR #2599, opened 2023-01-13
update behavior of hotspot menu

myk002:
- disappears on click outside its borders
- disappears on r-click
- mouse over the help panel counts as "over the menu" (so the menu doesn't close if the player moves the mouse to the help text)
- menu panels appear next to the logo hotspot instead of over it, allowing players to avoid clicking on the wrong item if they intend to click on the logo

### DIFF
--- a/library/lua/gui.lua
+++ b/library/lua/gui.lua
@@ -725,6 +725,7 @@ function ZScreen:onInput(keys)
     if ZScreen.super.onInput(self, keys) then
         -- ensure underlying DF screens don't also react to handled clicks
         if keys._MOUSE_L_DOWN then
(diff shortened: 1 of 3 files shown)


myk002:
note this does *not* fix the issue where the player is in mining mode, clicks on a hotspot menu list item, and frustratingly starts a mining selection. this is because DF reacts to `enabler->mouse_lbut` instead of `enabler->mouse_lbut_down` for *starting* a selection. there may be a good reason for this, or it may be a bug. @Putnam3145 do you have an opinion?

DFHack cannot clear `enabler->mouse_lbut` without breaking general dragging functionality.

PR #2612, opened 2023-01-15
Add cross-compile script

kelvie:
That will simply generate the win64 artifacts for those of us that run DF under Proton on Steam for Linux.

There are a few points of discussion I'd like to open up:

1. Where should the build script go / be called
2. What should the output directories be called
3. Whether or not the script should use `sudo docker` or just `docker`

To point (3), `docker` installs on most linux distros (and from the docker website) by default force you to be root to run Docker.

In practice, I would *guess* a good number of people set it up so that non-root users can use docker, because it's a lot more convenient.

The change to add BUILDER_UID to the build-env dockerfile has already been merged: https://github.com/BenLubar/build-env/pull/22


kelvie:
Sorry for the many force pushes, it's been a while since I've used RST/sphinx.

kelvie:
OK, the docs look okay finally, so this PR is ready for review/discussion!

kelvie:
Added a patch with a lot of feedback from others (removed sudo, auto-detect UID during run without having to rebuild the 10GB build image every time), and rebased to the latest develop.

this now depends on https://github.com/BenLubar/build-env/pull/23

build/build-win64-from-linux.sh
```diff
@@ -0,0 +1,56 @@
...
+# Assumes you built a container image called dfhack-build-msvc from
+# https://github.com/BenLubar/build-env/tree/master/msvc, see
+# docs/dev/compile/Compile.rst.
+#
+# NOTE: win64-cross is mounted in /src/build due to the hardcoded `cmake ..` in
+# the Dockerfile
+if ! docker run --rm -it -v "$srcdir":/src -v "$srcdir/build/win64-cross/":/src/build  \
+    -e BUILDER_UID=$builder_uid \
+    --name dfhack-win \
+    dfhack-build-msvc bash -c "cd /src/build && dfhack-configure windows 64 Release -DCMAKE_INSTALL_PREFIX=/src/build/output && dfhack-make -j$jobs install" \
```
lethosor, line 43: Curious: how does this behave with incremental builds? Is there an easy way to speed those up?

kelvie: Incremental builds work as they do locally -- I don't think anything gets rebuilt due to the configure running every time. If nothing's changed, it just re-links everything again, which IIRC is how it always worked with cmake and C++ (but my knowledge might be at least a decade old since I've used these two)

myk002:
Could you add a line to docs/changelog.txt? Under Misc Improvements, I think

build/build-win64-from-linux.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env sh
+
+# Number of jobs == core count
+jobs=$(grep -c ^processor /proc/cpuinfo)
```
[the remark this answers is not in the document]
kelvie: I wasn't 100% sure it was installed everywhere.

kelvie:
Done. Let me know if you want me to squash this into a singlke commit also.

myk002 (approved):

PR #2635, opened 2023-01-18
Fix contributing broken hyperlink

Halavus:
Link looped "too many redirects"


lethosor (changes requested):

lethosor:
Also I just fixed the old link as well (we did indeed have a redirect loop on the `readthedocs` side - good catch!), but updating the link here is a good call.

PR #2638, opened 2023-01-18
Updates to Autoclothing

RosaryMala:
There's a few new features that I want to add, but it works as-is right now. Mainly has issues with over-production.


.gitignore
```diff
@@ -72,3 +72,4 @@ tags
 
 # custom plugins
 /plugins/CMakeLists.custom.txt
+out/
```
myk002, line 75: what's this for?

library/modules/Buildings.cpp
```diff
@@ -141,6 +141,8 @@ void buildings_onUpdate(color_ostream &out)
 
     df::job_list_link *link = world->jobs.list.next;
     for (; link; link = link->next) {
+        if (!link->item)
```
myk002, line 144: did you just solve a crash here?

myk002 (commented):
code looks ok, but there is a distinct lack of changes to `autoclothing` is this the correct branch?

RosaryMala:
You're right, this is the wrong branch, closing this.

PR #2641, opened 2023-01-18
minimal implementation of buildingplan

myk002:
just enough to let `quickfort` place buildings without materials.

no material selection, UI, or configuration options yet, just persisted state and business logic.


myk002:
note we keep the old implementation in the `buildingplan/` directory for reference until the rest of the functionality is restored.

PR #2657, opened 2023-01-18
update autodump, only commands, add hotkeys

myk002:
and adapt to a "bad" cursor not always being equal to -30000 anymore


myk002:
Do you know if the change in "bad" cursor detection will have any false negatives?

data/init/dfhack.keybindings.init
```diff
@@ -40,14 +40,15 @@ keybinding add Ctrl-Shift-K gui/cp437-table
 # clean the selected tile of blood etc
 #keybinding add Ctrl-C spotclean
 
-# destroy the selected item
-#keybinding add Ctrl-K@dwarfmode autodump-destroy-item
+# teleport all unforbidden items marked for dumping to the tile under the keyboard cursor
+keybinding add Ctrl-Alt-D autodump
```
lethosor, line 44: What's the conflict with Ctrl-K?

library/include/df/custom/coord2d.methods.inc
```diff
@@ -1,6 +1,6 @@
 coord2d(uint16_t _x, uint16_t _y) : x(_x), y(_y) {}
 
-bool isValid() const { return x != -30000; }
+bool isValid() const { return x >= 0; }
```
lethosor, line 3: This can definitely have some false _positives_, since there is no guarantee that an uninitialized integer is negative. A large positive number could also easily be out of map bounds.

library/modules/Gui.cpp
```diff
@@ -650,7 +649,7 @@ bool Gui::item_details_hotkey(df::viewscreen *top)
 
 static bool has_cursor()
 {
-    return df::global::cursor && df::global::cursor->x != -30000;
+    return Gui::getCursorPos().isValid();
```
lethosor, line 652: I think a better option (for now, anyway) would be to check for UI modes where the cursor is known to be valid, since we can't determine validity solely from looking at the cursor.

data/init/dfhack.keybindings.init
myk002: Nothing, at the moment. I'm just trying to set a pattern that will result in no surprises. This is what I'm tentatively sticking to:  Modal tool windows can use bare keys, Ctrl keys, and alt keys, regular tool windows (ZScreens) can use Ctrl keys and alt keys, invisible keybindings can use ctrl-shift keys and ctrl-alt keys.

library/include/df/custom/coord2d.methods.inc
myk002: That's true, but for the purposes here, lack of a cursor should still be in the neighborhood of -30000, we just don't know the exact value anymore

library/modules/Gui.cpp
myk002: I'm not sure if that's the right strategy. We're already talking about having the cursor set by clicking or cursor follows mouse. That's explicitly so we can have valid cursors outside of the known good modes.

library/include/df/custom/coord2d.methods.inc
lethosor: Ok, I found your post on Discord. I don't have a link handy, but I'd summarize it as: the cursor does seem to be initialized to -30000, but then incremented/decremented by at most some multiple of the map size, so it should always remain negative (probably within 1,000 of -30,000, even).

So I'm ok with this, although it's not ideal.

library/modules/Gui.cpp
lethosor: That is fair, so it would need to check for that somehow. This comment was really conditional on https://github.com/DFHack/dfhack/pull/2657#discussion_r1080776011, though, so you can disregard this one.

data/init/dfhack.keybindings.init
lethosor: There is a bit more risk that using more modifiers will conflict with system shortcuts (e.g. ctrl-alt-T opens a terminal under either Ubuntu or GNOME), as opposed to our own "tried-and-true" shortcuts.

Actually, `ctrl-alt-D` on Ubuntu/GNOME _hides all windows_, equivalent to Windows+D on Windows. I'm pretty sure that's a default, at least.

ab9rf: it's generally risky to use ctrl-alt keys for any application purpose because those are widely considered reserved to system software

also, the msft UI guidelines say this:
> Avoid ALT+letter combinations because they may conflict with access keys. In addition, the system uses many specific key combinations for specialized input; for example, ALT+~ invokes an input editor for the Japanese language.
> Avoid CTRL+ALT combinations because the system interprets this combination in some language versions as an ALTGR key, which generates alphanumeric characters.

myk002: reverting autodump hotkey changes, though I might reinstate existing hotkeys for `quicksave` and `quickfort`

PR #2658, opened 2023-01-18
autofarm: fix repetition in status output

### LINKED ISSUE #2656: autofarm list items duplicated
The autofarm status function generates duplicate line items when run multiple times between closing and exiting DF.

Save can be found at https://dffd.bay12games.com/file.php?id=16366

Attached below are two images showing the console printout, having run the "autofarm status" command, and then some autochop stuff, and then checking "autofarm status" again to see if it was continuing the behavior. The duplicated items start 2/3 of the way down, with BUCKWHEAT (I think, I might have missed an earlier duplicate, but that is the first item in both lists, and shows up twice in the second list). No changes were made to autofarm behavior between status commands.

Autofarm Run 1
Autofarm Run 2

ab9rf:
because C++ std::map is not the same as a ruby table

should close #2656

### DIFF
--- a/plugins/autofarm.cpp
+++ b/plugins/autofarm.cpp
@@ -332,6 +332,7 @@ public:
     void status(color_ostream& out)
     {
         out << "Autofarm is " << (enabled ? "Active." : "Stopped.") << '\n';
+
         for (auto& lc : lastCounts)
         {
             auto plant = world->raws.plants.all[lc.first];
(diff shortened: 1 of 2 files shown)


plugins/autofarm.cpp
```diff
@@ -340,7 +341,7 @@ class AutoFarm {
 
         for (auto& th : thresholds)
         {
-            if (lastCounts[th.first] > 0)
+            if (lastCounts.find(th.first) != lastCounts.end())
```
myk002, line 344: Another way to put this, which I have found more succinct, is `if (lastCounts.count(th.first))`

myk002 (approved):

plugins/autofarm.cpp
ab9rf: that is potentially _slightly_ less efficient at the machine code level but not enough to care, especially in this context

myk002 (approved):

PR #2662, opened 2023-01-19
Zone <-> building interactions

20k:
Brief outline of the way zones + buildings interact:

1. Zones have a contained_buildings vector which is a list of buildings that meet a particular criteria

2. Buildings have a relations vector, which appears to be a list of civzones that contain them

3. The criteria for a building being put into a zone appears to be building->canBeRoom()

When a building is placed in a zone (including abstractly), zone->contained_buildings adds the building, and the civzone is added to the building's relation. When the building is uninstalled, its relations vector is used by the game to remove the building from the civzone's contained_buildings list (quite sure on this point). When a civzone is deleted, its contained_buildings list is used to update the contained building's relations list when it is deleted and is removed from their relations list

Overlapping civzones do not touch this system. canBeRoom encompasses beds, doors, archery targets, chairs, tables, traction benches etc, but does not include eg workshops

What this PR adds:

1. When constructing a building via buildings.constructWithItems or buildings.constructWithFilters, it checks if building->canBeRoom() is true, and if yes it finds all overlapping civzones at its centre. It then inserts itself into the civzones list of `contained_buildings`, and inserts the civzone into its `relations` list

2. When destroying a building, if it is a civzone, it finds (the slow method, via a global lookup) all buildings that have that civzone in their relations lists, and then removes them

3. When destroying a building, if its not a civzone but *is* an abstract building that can be in a civzone (ie canBeRoom), it finds all civzones at the building's centre and updates the civzones contained list. This happens just before it is deleted

5. When making a civzone in constructAbstract, the overlapping buildings are found and its contained_buildings and the overlapping buildings' relations are updated

6. When loading a save, a fixup is applied in case we end up with buildings that should be in zones that are not

What this PR can't do easily:

1. When shrinking, expanding, or generally modifying a zone, the contained_buildings and relations vectors of objects needs to be modified. This means that there are some out of sync issues here - *probably* of the non crashing kind. Civzones as modified by a user need some sort of poke hook when you modify them to correct their internal state. Perhaps buildings.notifyCivzoneModified(zone)

What this PR could add now that I think about it after writing this wall of text

1. Fixing up buildings that shouldn't be in a zone on load. I'll add this, it'll completely rebuild the civzone <-> building relations system when you load a save which should ensure that any non serious mistakes above will be corrected on load

Terribly alarming things about this PR:

1. In the event that a zone's contained_buildings vector contains a building, but the building's relation vector doesn't contain the zone, the game will crash a lot when the building is uninstalled. This whole PR is a bit of a recipe for crashing if mistakes are made

2. This PR is very conservative in terms of assumptions it makes, because there are still some significant unknowns here. There are several things that need more concrete answers

Things that could do with answers:

1. Is a civzone's relations vector always empty

2. Does a buildings relations vector only ever contain civzones

3. Is canBeRoom() the correct indicator function for putting buildings into zones

4. Do the pointers in the relations and contained_buildings vectors need to be sorted by ID, or perhaps even by pointer value? If the latter, this PR is broken. Edit: It does actually look like they do need to be sorted by ID

5. Can buildings not strictly within a civzone legally end up in its contained_buildings list

I use this system and this PR for drawing zones in the ui replacement and today I'm going to be extensively bugtesting this, but some basic experimentation with previous crash cases seems to show that it works on some level. Given that there are some unanswered questions I thought this might be useful to open so people can see what's necessary here, and also to write this all down somewhere so I don't forget

### DIFF
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -102,7 +102,6 @@ struct CoordHash {
 };
 
 static unordered_map<df::coord, int32_t, CoordHash> locationToBuilding;
-static unordered_map<df::coord, vector<int32_t>, CoordHash> locationToCivzones;
 
 static df::building_extents_type *getExtentTile(df::building_extents &extent, df::coord2d tile)
 {
@@ -115,18 +114,36 @@ static df::building_extents_type *getExtentTile(df::building_extents &extent, df
     return &extent.extents[dx + dy*extent.width];
 }
 
+void add_building_to_all_zones(df::building* bld);
+
+static void buildings_fixzones()
+{
+    auto& vec = world->buildings.other[buildings_other_id::IN_PLAY];
+
+    bool changed = false;
+
+    for (size_t i = 0; i < vec.size(); i++)
+    {
+        df::building* bld = vec[i];
+
+        add_building_to_all_zones(bld);
+    }
+}
+
 /*
  * A monitor to work around this bug, in its application to buildings:
  *
  * http://www.bay12games.com/dwarves/mantisbt/view.php?id=1416
  */
 bool buildings_do_onupdate = false;
+static bool buildings_do_fixzones = false;
 
 void buildings_onStateChange(color_ostream &out, state_change_event event)
 {
     switch (event) {
     case SC_MAP_LOADED:
         buildings_do_onupdate = true;
+        buildings_do_fixzones = true;
         break;
     case SC_MAP_UNLOADED:
         buildings_do_onupdate = false;
@@ -138,6 +155,12 @@ void buildings_onStateChange(color_ostream &out, state_change_event event)
 
 void buildings_onUpdate(color_ostream &out)
 {
+    if (buildings_do_fixzones)
+    {
+        buildings_fixzones();
+        buildings_do_fixzones = false;
+    }
+
     buildings_do_onupdate = false;
 
     df::job_list_link *link = world->jobs.list.next;
@@ -164,6 +187,163 @@ void buildings_onUpdate(color_ostream &out)
             iref->role = df::job_item_ref::Hauled;
             iref->job_item_idx = -1;
         }
+    }
+}
+
+static void building_into_zone_unidir(df::building* bld, df::building_civzonest* zone)
+{
+    for (size_t bid = 0; bid < zone->contained_buildings.size(); bid++)
+    {
+        if (zone->contained_buildings[bid] == bld)
+            return;
+    }
+
+    zone->contained_buildings.push_back(bld);
+
+    std::sort(zone->contained_buildings.begin(), zone->contained_buildings.end(), [](df::building* b1, df::building* b2)
+    {
+        return b1->id < b2->id;
+    });
+}
+
+static void zone_into_building_unidir(df::building* bld, df::building_civzonest* zone)
+{
+    for (size_t bid = 0; bid < bld->relations.size(); bid++)
+    {
+        if (bld->relations[bid] == zone)
+            return;
+    }
+
+    bld->relations.push_back(zone);
+
+    std::sort(bld->relations.begin(), bld->relations.end(), [](df::building* b1, df::building* b2)
+    {
+        return b1->id < b2->id;
+    });
+}
+
+static void add_building_to_zone(df::building* bld, df::building_civzonest* zone)
+{
+    if (!bld->canBeRoom())
+        return;
+
+    building_into_zone_unidir(bld, zone);
+    zone_into_building_unidir(bld, zone);
+}
+
+static bool is_suitable_building_for_zoning(df::building* bld)
+{
+    return bld->canBeRoom();
+}
+
+static void add_building_to_all_zones(df::building* bld)
+{
+    if (!is_suitable_building_for_zoning(bld))
+        return;
+
+    df::coord coord(bld->centerx, bld->centery, bld->z);
+
+    std::vector<df::building_civzonest*> cv;
+    Buildings::findCivzonesAt(&cv, coord);
+
+    for (size_t i=0; i < cv.size(); i++)
+    {
+        add_building_to_zone(bld, cv[i]);
+    }
+}
+
+static void add_zone_to_all_buildings(df::building* zone_as_building)
+{
+    if (zone_as_building->getType() != building_type::Civzone)
+        return;
+
+    auto zone = strict_virtual_cast<df::building_civzonest>(zone_as_building);
+
+    if (zone == nullptr)
+        return;
+
+    auto& vec = world->buildings.other[buildings_other_id::IN_PLAY];
+
+    for (size_t i = 0; i < vec.size(); i++)
+    {
+        auto against = vec[i];
+
+        if (zone->z != against->z)
+            continue;
+
+        if (!is_suitable_building_for_zoning(against))
+            continue;
+
+        int32_t cx = against->centerx;
+        int32_t cy = against->centery;
+
+        df::coord2d coord(cx, cy);
+
+        //can a zone without extents exist?
+        if (zone->room.extents && zone->isExtentShaped())
+        {
+            auto etile = getExtentTile(zone->room, coord);
+            if (!etile || !*etile)
+                continue;
+
+            add_building_to_zone(against, zone);
+        }
+    }
+}
+
+static void remove_building_from_zone(df::building* bld, df::building_civzonest* zone)
+{
+    for (int bid = 0; bid < (int)zone->contained_buildings.size(); bid++)
+    {
+        if (zone->contained_buildings[bid] == bld)
+        {
+            zone->contained_buildings.erase(zone->contained_buildings.begin() + bid);
+            bid--;
+        }
+    }
+
+    for (int bid = 0; bid < (int)bld->relations.size(); bid++)
+    {
+        if (bld->relations[bid] == zone)
+        {
+            bld->relations.erase(bld->relations.begin() + bid);
+            bid--;
+        }
+    }
+}
+
+static void remove_building_from_all_zones(df::building* bld)
+{
+    df::coord coord(bld->centerx, bld->centery, bld->z);
+
+    std::vector<df::building_civzonest*> cv;
+    Buildings::findCivzonesAt(&cv, coord);
+
+    for (size_t i=0; i < cv.size(); i++)
+    {
+        remove_building_from_zone(bld, cv[i]);
+    }
+}
+
+static void remove_zone_from_all_buildings(df::building* zone_as_building)
+{
+    if (zone_as_building->getType() != building_type::Civzone)
+        return;
+
+    auto zone = strict_virtual_cast<df::building_civzonest>(zone_as_building);
+
+    if (zone == nullptr)
+        return;
+
+    auto& vec = world->buildings.other[buildings_other_id::IN_PLAY];
+
+    //this is a paranoid check and slower than it could be. Zones contain a list of children
+    //good for fixing potentially bad game states when deleting a zone
+    for (size_t i = 0; i < vec.size(); i++)
+    {
+        df::building* bld = vec[i];
+
+        remove_building_from_zone(bld, zone);
     }
 }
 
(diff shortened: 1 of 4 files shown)


myk002 (commented):
my poor civzone cache! I do have hope for making it usable for v50 if I can ask @Putnam3145 for the proper hook. O(N) lookups for civzone by tile can absolutely kill the performance of `blueprint` and `quickfort`. good thing zone mode is disabled for them right now!

questions: your concern about reshaping zones and updating membership is related to if a DFHack tool does this, right? If players reshape a zone, won't DF handle updating related buildings?

myk002:
btw I believe the answer to (2) Does a buildings relations vector only ever contain civzones
is "yes" according to ab9rf's RE

For (3) Is canBeRoom() the correct indicator function for putting buildings into zones
this can be tested fairly easily, I think. make a zone, put every building inside of it, see what the relations vectors say

For (5) Can buildings not strictly within a civzone legally end up in its contained_buildings list
are there any buildings larger than 1x1 that can be referenced by a civzone?

20k:
> my poor civzone cache! I do have hope for making it usable for v50 if I can ask @Putnam3145 for the proper hook (e.g. `onExtentsChanged(id)` or something). O(N) lookups for civzone by tile can absolutely kill the performance of `blueprint` and `quickfort`. good thing zone mode is disabled for them right now!
> 

If quickfort is doing a lot of lookups in one 'tick', it might be possible to build up a temporary cache and then throw it away next tick

> questions: your concern about reshaping zones and updating membership is related to if a DFHack tool does this, right? If players reshape a zone, won't DF handle updating related buildings?

Yes, df correctly handles all of this, its just if dfhack pokes at the extents directly. It'd be worth investigating if DF has a method for notifying civzones like this. I'm not sure if any dfhack consumer actually directly modifies zone extents after construction (other than me in the ui replacement, I've actually implemented notifyCivzoneModified now and am getting to testing it)

20k:
> should a zone _need_ fixing on every load? is this that common of a problem? is this something that could go in a `fix/` script that runs periodically?

So there's a few ways that this can happen, but anyone who's been using dfhack tools to work with zones or buildings will have a buggy save, and if any tools modify extents without notifying the civzones, it'll also cause saves to be (likely) harmlessly incorrect which is potentially an ongoing problem. On top of that, if there's any non-fatal mistakes in this PR because of overlooked edge cases, they'll get cleaned up on a save reload

DF itself shouldn't generate this issue in a way that zone fixing would apply to, but I have no knowledge about if it does. Errors here are generally either mostly harmless or immediate crashers

I avoided an approach that would run periodically, as the computational complexity for the fixup is O(buildings * civzones), which is pretty bad and may cause stuttering, but there's no reason fundamental it couldn't be made a script

I'm not particularly fussed about this end of things because its just different sets of tradeoffs, but I will say that as a script author its going to be easy to screw this up as zones have a lot of moving parts, and there are also unknown unknowns here because the reverse engineering is extremely fresh

20k:
> btw I believe the answer to (2) Does a buildings relations vector only ever contain civzones is "yes" according to ab9rf's RE

Good to know! 

> For (3) Is canBeRoom() the correct indicator function for putting buildings into zones this can be tested fairly easily, I think. make a zone, put every building inside of it, see what the relations vectors say

Ahh heh true I guess I'll add this one to the todo list at some point for a rainy day

> For (5) Can buildings not strictly within a civzone legally end up in its contained_buildings list are there any buildings larger than 1x1 that can be referenced by a civzone?

Its also that its theoretically possible that eg the contained_buildings list is more an associated list of buildings with the civzone, so maybe things not contained strictly within it could participate (locations? I also noticed that there's a move building function that I've never investigated). I think its probably unlikely, but I haven't played 50.xx extensively enough with tools on hand to have any suspicions here. On top of that, if any bugs in DF ever generate buildings outside the civzone that are associated with it - that's as good as this being true, because it rules out a class of (faster) implementations

myk002:
My concern with "fixing" zones on load is that if our understanding is wrong or incomplete or if the implementation changes, we are then *enforcing* our wrong/incomplete understanding upon all players. I think it would make me more comfortable if we didn't do any onload fixup, but instead released the fix as a script the player could manually invoke if they suspect they have a problem.

> If quickfort is doing a lot of lookups in one 'tick', it might be possible to build up a temporary cache and then throw it away next tick

That's a fair point, and is exactly what quickfort does for engravings, since that's also a linear lookup.

> Ahh heh true I guess I'll add this one to the todo list at some point for a rainy day

Quickfort actually has a testing blueprint that places all buildings, but that of course bypasses the critical manual step that we'd be testing here..

20k:
That's a very good point with respect to zone fixups, I'll take it out of this commit series and port it to a script

20k:
https://github.com/20k/scripts/blob/dfui/fix/zones.lua for future reference will fix zones based on my current understanding - though it needs changes in this PR to function (findCivzonesAt)

20k:
Also, with some hindsight there are two current crashers in dfhack as-is without this PR, both in buildings.deconstruct. If a zone is deconstructed that has live references to it the game'll be corrupted, and additionally if a zone/barracks has ever had a squad assigned to it that still exists, the game will also crash/corrupt if the zone is deleted via buildings.deconstruct (which is a very new one I discovered). I have a fix for the latter locally, which is how I discovered incidentally that squad is misaligned

So basically: I would heavily recommend not using deconstruct at all on game created zones for any functionality, because it'll almost certainly break the game

myk002:
> Also, with some hindsight there are two current crashers in dfhack as-is without this PR, both in buildings.deconstruct. If a zone is deconstructed that has live references to it the game'll be corrupted, and additionally if a zone/barracks has ever had a squad assigned to it that still exists, the game will also crash/corrupt if the zone is deleted via buildings.deconstruct (which is a very new one I discovered).

To my knowledge, the only tool that would deconstruct a zone is `quickfort undo` with a `#zone` blueprint, which is not yet supported in v50, so we should be ok.

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
     }
 }
 
+static void building_into_zone_unidir(df::building* bld, df::building_civzonest* zone)
+{
+    for (size_t bid = 0; bid < zone->contained_buildings.size(); bid++)
+    {
+        if (zone->contained_buildings[bid] == bld)
+            return;
+    }
```
myk002, line 175:
```suggestion
    for (auto contained_building : zone->contained_buildings)
    {
        if (contained_building == bld)
            return;
    }
```

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+
+    zone->contained_buildings.push_back(bld);
+
+    std::sort(zone->contained_buildings.begin(), zone->contained_buildings.end(), [](df::building* b1, df::building* b2)
+    {
+        return b1->id < b2->id;
+    });
```
myk002, line 182: instead of push_back and re-sorting, checkout `insert_into_vector` from MiscUtils.h

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+        if (bld->relations[bid] == zone)
+            return;
+    }
+
+    bld->relations.push_back(zone);
+
+    std::sort(bld->relations.begin(), bld->relations.end(), [](df::building* b1, df::building* b2)
+    {
+        return b1->id < b2->id;
+    });
```
myk002, line 198: same here, `insert_into_vector` can avoid the sort

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+}
+
+static void zone_into_building_unidir(df::building* bld, df::building_civzonest* zone)
+{
+    for (size_t bid = 0; bid < bld->relations.size(); bid++)
+    {
+        if (bld->relations[bid] == zone)
+            return;
+    }
```
myk002, line 191:
```suggestion
    for (auto relation : bld->relations)
    {
        if (relation == zone)
            return;
    }
```

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+
+    df::coord coord(bld->centerx, bld->centery, bld->z);
+
+    std::vector<df::building_civzonest*> cv;
+    Buildings::findCivzonesAt(&cv, coord);
+
+    for (size_t i=0; i < cv.size(); i++)
+    {
+        add_building_to_zone(bld, cv[i]);
+    }
```
myk002, line 228:
```suggestion
    for (auto zone : cv)
        add_building_to_zone(bld, zone);
```

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+    auto zone = strict_virtual_cast<df::building_civzonest>(zone_as_building);
+
+    if (zone == nullptr)
+        return;
+
+    auto& vec = world->buildings.other[buildings_other_id::IN_PLAY];
+
+    for (size_t i = 0; i < vec.size(); i++)
+    {
+        auto against = vec[i];
```
myk002, line 245:
```suggestion
    for (auto against : vec)
    {
```

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+
+        if (!is_suitable_building_for_zoning(against))
+            continue;
+
+        int32_t cx = against->centerx;
+        int32_t cy = against->centery;
+
+        df::coord2d coord(cx, cy);
+
+        //can a zone without extents exist?
```
myk002, line 258: I don't know, but there are buildings whose extents aren't filled out iff they are perfectly rectangular.

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+}
+
+static void remove_building_from_zone(df::building* bld, df::building_civzonest* zone)
+{
+    for (int bid = 0; bid < (int)zone->contained_buildings.size(); bid++)
+    {
+        if (zone->contained_buildings[bid] == bld)
+        {
+            zone->contained_buildings.erase(zone->contained_buildings.begin() + bid);
+            bid--;
```
myk002, line 277: are you saying here that a building can appear more than once in a contained_buildings list?

myk002: and similarly below for a relations list?

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+{
+    df::coord coord(bld->centerx, bld->centery, bld->z);
+
+    std::vector<df::building_civzonest*> cv;
+    Buildings::findCivzonesAt(&cv, coord);
+
+    for (size_t i=0; i < cv.size(); i++)
+    {
+        remove_building_from_zone(bld, cv[i]);
+    }
```
myk002, line 301:
```suggestion
    for (auto zone : cv)
        remove_building_from_zone(bld, zone);
```

library/modules/Buildings.cpp
```diff
@@ -167,6 +166,163 @@ void buildings_onUpdate(color_ostream &out)
...
+    auto& vec = world->buildings.other[buildings_other_id::IN_PLAY];
+
+    //this is a paranoid check and slower than it could be. Zones contain a list of children
+    //good for fixing potentially bad game states when deleting a zone
+    for (size_t i = 0; i < vec.size(); i++)
+    {
+        df::building* bld = vec[i];
+
+        remove_building_from_zone(bld, zone);
+    }
```
myk002, line 323:
```suggestion
    for (auto bld : vec)
        remove_building_from_zone(bld, zone);
```

myk002:
Everything looks functionally correct. The comments I had were all stylistic. I'll merge this now for the release. When you address the comments, I can re-merge.

library/modules/Buildings.cpp
20k: This was one of those things that I did conservatively just in case, I would heavily suspect it is not possible but I don't *know* for a fact it can't happen. Apparently toady has dedicated classes for sorted vectors, if those explicitly never contain duplicates this is probably fine to remove

PR #2664, opened 2023-01-19
Update Units.cpp

cjhammel:
fixes renamed unit flag

### DIFF
--- a/docs/about/Removed.rst
+++ b/docs/about/Removed.rst
@@ -20,7 +20,7 @@ Replaced by `gui/launcher --minimal <gui/launcher>`.
 
 create-items
 ============
-Replaced by `gui/create-item --multi`.
+Replaced by ``gui/create-item --multi``.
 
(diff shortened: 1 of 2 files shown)


myk002 (approved):

docs/about/Removed.rst
```diff
@@ -20,7 +20,7 @@ Replaced by `gui/launcher --minimal <gui/launcher>`.
 
 create-items
 ============
-Replaced by `gui/create-item --multi`.
+Replaced by ``gui/create-item --multi``.
```
lethosor, line 23: Would be nice to keep the link working here, so:

```suggestion
Replaced by `gui/create-item --multi <gui/create-item>`.
```

ab9rf: why is this changelog change even in this change? it doesn't seem related

myk002: I'm also a little confused here.

Issue #52, opened 2010-11-06
"press a key to continue" prevents scripting


rofl0r:
i'd really like to remove that, so i can i.e. run a script that calls dfcleanmap in a loop...
i dont think thats needed at all, it's like those annoying "do you really want to quit [Yes/no]" nag screens...
what do you think ?

PR #2670, opened 2023-01-20
Initial automelt implementation

eamondo2:
Could use some perf. testing/tuning. Seems to work well enough, feedback welcome.


eamondo2:
pre-commit.ci autofix

plugins/automelt.cpp
```diff
@@ -1,294 +1,705 @@
...
 REQUIRE_GLOBAL(world);
 REQUIRE_GLOBAL(cursor);
 REQUIRE_GLOBAL(plotinfo);
 
-static const string PERSISTENCE_KEY = "automelt/stockpiles";
+namespace DFHack
+{
+    DBG_DECLARE(automelt, status, DebugCategory::LINFO);
+    DBG_DECLARE(automelt, cycle, DebugCategory::LINFO);
+    DBG_DECLARE(automelt, perf, DebugCategory::LTRACE);
```
myk002, line 50: this sets the *default* log level to TRACE, which is almost certainly not what you want. This should probably be set to LINFO here. If you want trace output, you'd run `debugfilter set trace automelt perf`

plugins/automelt.cpp
```diff
@@ -1,294 +1,705 @@
...
-        Items::getContainedItems(item, &contents);
-        for (auto child = contents.begin(); child != contents.end(); child++)
-        {
-            marked_count += mark_item(*child, bad_flags, stockpile_id);
+};
+
+enum StockpileConfigValues
+{
+    STOCKPILE_CONFIG_ID = 0,
+    STOCKPILE_CONFIG_MONITORED = 1,
```
myk002, line 70: I believe thresholds for melting objects are better handled in the manager orders, not here. I'm not sure what an extra `monitored` flag would do for this plugin.

plugins/automelt.cpp
```diff
@@ -1,294 +1,705 @@
...
+static vector<PersistentDataItem> watched_stockpiles;
+static unordered_map<int, size_t> watched_stockpiles_indices;
+
 
-static int mark_item(df::item *item, df::item_flags bad_flags, int32_t stockpile_id)
+enum ConfigValues
 {
-    if (item->flags.whole & bad_flags.whole)
-        return 0;
+    CONFIG_IS_ENABLED = 0,
```
myk002, line 63: automelt doesn't have quite the same semantics for enabling as autochop. For autochop, when it's enabled, it starts having an effect on the game. In this case, there is a required manual step for players to take before this plugin does anything.

I recommend a model closer to buildingplan: https://github.com/DFHack/dfhack/pull/2641
where it does not store its enabled state, but rather is enabled unconditionally in the default dfhack.init (but can still be disabled per-session for debugging or preference)

The business logic here in general is all still applicable and correct (and looks good!), but the top level logic should be swapped out for something based on buildingplan.

plugins/automelt.cpp
myk002: reading the gui code.. I see that this can give you a tally of the number of items in monitored stockpiles. Could you explain the cases where this is handy for the player?

plugins/automelt.cpp
eamondo2: Ah, good to know. I need to add a few more debug calls to get timing info and to check perf. impact.

plugins/automelt.cpp
eamondo2: So the plugin will only produce item counts if the pile is `monitored`, and will only melt if both `monitored` and `melt` are enabled.
The actual idea was a bit fuzzier, but would allow you to set a similar flag/value to autochop, in that it will only bother marking items in a `monitored` and `melt` enabled pile if the number of items to be melted is above some threshold, configurable via the UI. This would mean that if I'd had a work order set up to say `melt metal item` with the condition of `if 10 or more meltable items available` I wouldn't get order cancellation spam if there ended up being about 11 or so items marked. E.g, the config threshold value is "only mark for melting if there are 10 or more items to be marked" and the matching work order expects 10 items, the plugin would mark items in batches of 10. 

Further extension would mean you could set a config to leave a minimum number of items in enabled/marked stockpiles. Since the stockpile filters are a bit fiddly, you could end up with a situation where you want to keep at least x of some item around, but melt the excess. Instead of having to carefully set the stockpile size to respect that desired item count (and use pile links etc.) you can just set that in the pile's automelt config. It'd monitor the pile, and mark for melting above a certain count, then stop.

I have no idea if any of the above would actually be useful to anyone, even to me TBH. Can disable the extra flag for now, basically make the `STOCKPILE_CONFIG_MONITORED` flag pull double-duty as the melt-enable flag. Monitored piles get their items counted, all others are skipped in the initial cycle phase, so it cuts down on some exec time.

I also don't know if having the live count/overview of marked/monitored items in enabled piles is useful. I get the count numbers from the cycle pass anyway, and it produces that as output if you run `automelt designate` so I passed it to the GUI because it was easy.

plugins/automelt.cpp
eamondo2: So if I understand correctly, your suggestion is to assume that it's enabled, provided that there are any stockpiles with the monitor/melt condition enabled?
(And in this case, "assume it's enabled" means adding that entry to the `dfhack.tools.init` config)

That makes sense, you're right that additional steps are required for it to do anything. Have to pick piles to enable etc. before it will have an effect on the game and designate items. I can look at making that change as well.

plugins/automelt.cpp
myk002: ok, I understand. I support having only one flag, if only because the use case is not well defined. Sending the count of melt-marked to automelt is absolutely useful. players can use that count to decide whether to build more smelters. I'd actually send the count of *all* melt-marked items, not just the ones in the stockpiles. I'm not sure what decision the stockpile-only count would be useful to make..

plugins/automelt.cpp
myk002: That's correct. This plugin should always *load* and *persist* its state, and if it's disabled, the cycle simply won't run.

eamondo2: Gotcha. That's what the latest ver. does.

plugins/automelt.cpp
eamondo2: Well, the way it's rigged now I only loop over items in the monitored stockpiles, so I don't have to go over every item present in the world. I could add a toggle-able option to scan all items everywhere, and return a count of melt-marked globally as well. Unsure if there's an already filtered list of items marked for melting, I vaguely remember seeing something like that though.

Latest ver. has only one flag to manage, `monitored`. Any stockpile with that enabled gets its items automatically marked and processed.

myk002: It does look like there is an items vector `ANY_MELT_DESIGNATED`. that would make it cheap enough not to have to make it a special flag

myk002:
Please address the build and pre-commit.ci errors in the build actions as well, please.

plugins/automelt.cpp
```diff
@@ -1,294 +1,735 @@
...
+    F(melt);
+#undef F
+
+    if (item->flags.whole & bad_flags.whole)
+        return false;
 
-    void remove(df::building_stockpilest *sp)
+    df::item_type t = item->getType();
+
+    if (t == df::enums::item_type::BOX || t == df::enums::item_type::BAR)
```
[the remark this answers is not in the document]
eamondo2: That's a good question, I don't know whether that is the case or not. I pulled most of the melt flag checks and melt marking code from the original automelt plugin.

eamondo2:
I _think_ I got all your suggestions/requests encapsulated. Still might need some tweaking, but overall seems functional enough.

plugins/automelt.cpp
```diff
@@ -1,294 +1,735 @@
...
+        return false;
+
+    if (!is_metal_item(item))
+        return false;
+
+    for (auto g = item->general_refs.begin(); g != item->general_refs.end(); g++)
```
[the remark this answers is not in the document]
eamondo2: Good suggestion. Implemented.

plugins/automelt.cpp
```diff
@@ -1,294 +1,735 @@
...
 
-        mark_all_in_stockpiles(monitored_stockpiles);
+static inline bool is_set_to_melt(df::item *item)
+{
+    return item->flags.bits.melt;
+}
+
+static inline bool is_in_job(df::item *item) {
+    return item->flags.bits.in_job;
+}
```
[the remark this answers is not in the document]
eamondo2: Removed

plugins/automelt.cpp
```diff
@@ -1,294 +1,735 @@
...
+    F(in_job);
+    F(hostile);
+    F(on_fire);
+    F(rotten);
+    F(trader);
+    F(in_building);
+    F(construction);
+    F(artifact);
+    F(melt);
+#undef F
```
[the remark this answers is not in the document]
ab9rf: I continue to advocate for the approach used in tailor, since that approach generates a constexpr that compiler can reduce at compiletime.

myk002:
I'm about to do the (probably final) review. Would you like to take the next steps on this? These kinds of stockpile functions are the #1 requested DFHack feature right now (https://strawpoll.com/polls/poy9R95NwnJ/results).

Next steps would be:

1) overlay widget (which I know you're already working on)
2) merge autotrade support into this plugin (can remove existing autotrade code and docs)
3) rename this plugin to 'stockpiles' and the associated gui to `gui/stockpiles`; merge in existing `stockpiles` functionality (import/export); add access to import/export to the overlay widget

eamondo2:
I can definitely take a look at those next steps. 
For the "merge `autotrade` support" etc. would it make sense to have that be a separate `.cpp` file, and move both this and the new `autotrade` functionality to a folder? Something like what was done for `autolabor`, `buildingplan` etc.
Might make it tidier than having everything in one file, `automelt.cpp` is getting a bit long as it is.

I've started looking into the overlay widget some more, thanks to your suggestions and pointing me to the `workorders` overlay code.

myk002:
I would like to combine them into one *plugin*, but they can certainly stay in separate .cpp files. Feel free to create a subdirectory (actually, just repurpose the existing `stockpiles` subdir) and move the files in there with their existing names.

The reason I'd like them to be one plugin is that we're going to have one gui, and there is a pretty consistent theme that the gui for plugin `X` is `gui/X`

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+Please see `gui/automelt` for the interactive configuration dialog.
 
 Usage
 -----
 
 ::
 
     enable automelt
+    automelt [status]
+    automelt (designate)
```
myk002, line 20: no need for parens around single literal strings. just `automelt designate` is fine

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+    automelt (monitor|nomonitor) <stockpile>
+
+Examples
+--------
+
+Automatically designate all meltable items in the stockpile ("melt") for melting. ::
```
myk002, line 26:
```suggestion
Automatically designate all meltable items in the stockpile ("melt") for melting::
```

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+
+    enable automelt
+    automelt monitor melt
+
+Enable monitoring for the stockpile ("melt"), and mmediately designate all meltable items in monitored stockpiles for melting. ::
+
+    automelt monitor melt
```
myk002, line 33: I think that adding a stockpile to the monitor list should immediately designate its contents for some immediate feedback, no second command necessary.

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+Commands
+--------
+
+``status``
+    Shows current configuration and relevant statistics
+
+``designate``
+    Designates items in monitored stockpiles for melting right now. This works even if ``automelt`` is not currently enabled.
+
+``(no)monitor <stockpile>``
```
myk002, line 45: no need to document it, but I'd suggest also handling "unmonitor" as an alias for "nomonitor"

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+    Enable/disable monitoring of a given stockpile. Requires the stockpile to have a name set.
```
myk002, line 46: why must it have a name set? why can't we also handle IDs here?

plugins/automelt.cpp
```diff
@@ -265,27 +268,40 @@ static bool isStockpile(df::building * building) {
     return building->getType() == df::building_type::Stockpile;
 }
 
+struct BadFlagsCanMelt {
+    uint32_t whole;
+
+    BadFlagsCanMelt() {
```
myk002, line 274: should this have artifact, spider_web, and owned too?

and for both of them, maybe take away rotten? is there anything meltable that can rot? can it still be melted even if it's rotten?

plugins/lua/automelt.lua
```diff
@@ -47,7 +47,7 @@ function parse_commandline(...)
     return true
 end
 
--- used by gui/autochop
+-- used by gui/auomelt
```
myk002, line 50:
```suggestion
-- used by gui/automelt
```

plugins/automelt.cpp
eamondo2: That is a good question. I don't honestly know if there are rotten meltable items, perhaps some metal-based forgotten beast bits could count? I don't know if they have a `craft_material_class` of Metal or not.

docs/plugins/automelt.rst
eamondo2: Fixed in latest changes. It would already handle enabling by ID, but the GUI didn't show the ID as an alternative to the name if the name was not set.

docs/plugins/automelt.rst
eamondo2: Added in latest

docs/plugins/automelt.rst
eamondo2: Added in latest

docs/plugins/automelt.rst
eamondo2: Fixed in latest

docs/plugins/automelt.rst
```diff
@@ -6,13 +6,41 @@ automelt
...
+
+Examples
+--------
+
+Automatically designate all meltable items in the stockpile ("melt") for melting. ::
+
+    enable automelt
+    automelt monitor melt
+
+Enable monitoring for the stockpile ("melt"), and mmediately designate all meltable items in monitored stockpiles for melting. ::
```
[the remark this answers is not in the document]
eamondo2: Fixed in latest

docs/plugins/automelt.rst
eamondo2: Fixed in latest

plugins/lua/automelt.lua
eamondo2: Fixed in latest

eamondo2:
Fixed/added all your suggested changes. 
Re-enabled toggling monitor status for multiple stockpiles, and handled some cases where the stockpile doesn't have custom name set. 
You can now run `automelt monitor "Stockpile #52"` or `automelt monitor 52`. 
Added note of this in the docs, added some examples as well. 
Works even for mixed names, e.g `automelt monitor melt,"Stockpile #52",36`.
Next bit is to rename everything, and add the remaining general stockpile operations etc. as you'd mentioned @myk002.

eamondo2:
Further update, things look to be a bit broken. Need to figure out how the stockpile name changes are borking stuff.

eamondo2:
Alrighty, CTD issue fixed. Turns out I was tracking the stockpile IDs in a real dumb way.

plugins/automelt.cpp
```diff
@@ -625,48 +626,67 @@ static void automelt_printStatus(color_ostream &out) {
...
     DEBUG(status,out).print("exiting automelt_printStatus\n");
 
 }
 
 static void automelt_setStockpileConfig(color_ostream &out, int id, bool monitored) {
-    DEBUG(status,out).print("entering automelt_setStockpileConfig\n");
+    DEBUG(status,out).print("entering automelt_setStockpileConfig for id=%d and monitored=%d\n", id, monitored);
     validate_stockpile_configs(out);
-    bool isInvalidStockpile = !df::building::find(id);
+    bool isInvalidStockpile = !df::building::find(id) || !isStockpile(df::building::find(id));
```
myk002, line 674: find is O(log n), which is not free. I'd cache the return value instead of running the search twice.

myk002 (commented):
looks good! just one nit and this is ready for merge. Thanks!

plugins/automelt.cpp
```diff
@@ -671,7 +670,8 @@ static void automelt_printStatus(color_ostream &out) {
 static void automelt_setStockpileConfig(color_ostream &out, int id, bool monitored) {
     DEBUG(status,out).print("entering automelt_setStockpileConfig for id=%d and monitored=%d\n", id, monitored);
     validate_stockpile_configs(out);
-    bool isInvalidStockpile = !df::building::find(id) || !isStockpile(df::building::find(id));
+    auto bldg = df::building::find(id);
+    bool isInvalidStockpile = bldg || !isStockpile(bldg);
```
myk002, line 674:
```suggestion
    bool isInvalidStockpile = !bldg || !isStockpile(bldg);
```

myk002:
docs failure is expected, of course, until `gui/automelt` is in the tree

plugins/automelt.cpp
```diff
@@ -1,294 +1,827 @@
...
+        if (watched_stockpiles.count(id)) {
+            DEBUG(cycle,*out).print("indexed_id=%d\n", get_config_val(watched_stockpiles[id], STOCKPILE_CONFIG_ID));
+            push_stockpile_config(L, watched_stockpiles[id]);
+        } else {
+            push_stockpile_config(L, id, false);
+        }
+    }
+    DEBUG(perf, *out).print("exit automelt_getItemCountsAndStockpileConfigs\n");
+
+    return 4+bldg_count;
```
lethosor, line 814: This is coming from a discussion on Discord (not sure if you are on there):

This is unsafe and could overflow the stack. Per https://www.lua.org/manual/5.3/manual.html#4.2:

> Whenever Lua calls C, it ensures that the stack has space for at least LUA_MINSTACK extra slots. LUA_MINSTACK is defined as 20, so that usually you do not have to worry about stack space unless your code has loops pushing elements onto the stack.

This could easily push more than 20 elements, and most stack overflow checks in Lua just call `luai_apicheck()`, which is defined as `assert()` by default, i.e. a no-op in production builds.

It would be safer to return a known number of elements, such as by adding every table here to a single containing table. Alternatively, if you really need to return an arbitrary number of values (which I do not recommend) you would need to use `luaL_checkstack()`.

eamondo2: Oh snap, good catch. I wasn't familiar with how the c++ to Lua push worked, lifted most of that from `autochop`.
Can see about reworking it so it'll shove all the buildings in a single containing table, then push that.

PR #2681, opened 2023-01-21
More civzone <-> building work, add removing zones with associated squads support

20k:
This addresses the style changes, as well as implementing support for removing zones which have ever had a squad assigned to them. The removal of zones from squads is conservative, and handles duplicates if they arise - though I haven't seen it

It *is* worth noting that I discovered surprisingly that there isn't a 1:1 mapping between the zone's room-squad vector, and the squad's room-zone vector. A zone remembers any squads ever associated with it, whereas squads proactively clean up their zone associations

Additionally, given some of the suspicion about vanilla bugs in this kind of bidirectional mapping, I'm not sure how generally conservative the civzone changes want to be. At the moment I've taken the approach generally of df-is-not-to-be-trusted, which is somewhat borne out by some suspicious bug reports

Updating miscutils was necessary to get it to work with derived types in insert_with_vector (eg `insert_with_vector(bld->relations, &df::building_civzonest::id, zone)`), as df::building_civzonest::id is actually technically of type (int32_t df::building*). There don't seem to be that many users of it, but I thought it might be useful and it cleans up this case

I've been bashing on this and the previous PR a lot locally as I use them for drawing zones in the UI rework, and haven't noticed any issues so far which is good

### DIFF
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -169,34 +169,24 @@ void buildings_onUpdate(color_ostream &out)
 
 static void building_into_zone_unidir(df::building* bld, df::building_civzonest* zone)
 {
-    for (size_t bid = 0; bid < zone->contained_buildings.size(); bid++)
+    for (auto contained_building : zone->contained_buildings)
     {
-        if (zone->contained_buildings[bid] == bld)
+        if (contained_building == bld)
             return;
     }
 
-    zone->contained_buildings.push_back(bld);
-
-    std::sort(zone->contained_buildings.begin(), zone->contained_buildings.end(), [](df::building* b1, df::building* b2)
-    {
-        return b1->id < b2->id;
-    });
+    insert_into_vector(zone->contained_buildings, &df::building::id, bld);
 }
 
 static void zone_into_building_unidir(df::building* bld, df::building_civzonest* zone)
 {
-    for (size_t bid = 0; bid < bld->relations.size(); bid++)
+    for (auto relation : bld->relations)
     {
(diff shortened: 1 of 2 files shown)


library/modules/Buildings.cpp
```diff
@@ -1348,6 +1325,43 @@ bool Buildings::constructWithFilters(df::building *bld, std::vector<df::job_item
...
+        if (squad)
+        {
+            for (int i=0; i < (int)squad->rooms.size(); i++)
+            {
+                if (squad->rooms[i]->building_id == bld->id)
+                {
+                    auto room = squad->rooms[i];
+                    squad->rooms.erase(squad->rooms.begin() + i);
+                    delete room;
+                    i--;
```
myk002, line 1354: if you traverse the vector backwards, you won't need to adjust this

20k:
I added in cleaning up assignments on a zone destroy as well, which I've tested a fair amount locally. I suspect that assigned_units and assigned_items are both also going to need some cleanup on zone destroy, but I haven't hit that portion of my reverse engineering adventures yet. That's to SeerSkye because I had absolutely 0 idea that owned_buildings was a thing that interacted with civzones

The owned_buildings bug unfortunately justifies a lot of the other paranoia with this PR series, as it can (and regularly does) contain duplicates and invalid pointers in vanilla

20k:
Well, some more work has shown that there isn't a 1:1 mapping between assigned unit ids, and dwarves that point to that civzone, eg spouses, so I'm going to implement a more bruteforcey approach here to pick up all the cases

myk002 (approved):

PR #2682, opened 2023-01-21
makeSquad, updateRoomAssignments

20k:
Both of these functions are absolutely hideously complicated to implement correctly and required a lot of reverse engineering, and I thought they might well be useful to have generally. I've tested makeSquad a fair amount in 50.xx and it seems to work, updateRoomAssignments is a lot more straightforward to test and I've bashed on it a lot. Both of these are unimplementable from lua, because of memory allocation

These functions take IDs currently but I've noticed that a lot of the rest of the API takes raw values so that probably wants to change? And the sorts below could probably be changed to a sorted insert, as the sorting is necessary

Other than that: the main question is is this something generally that you'd want in dfhack core? I imagine it'll need reworking a fair bit, but I'm happy to make it happen because I'm hoping to shrink the size of my fork for the ui replacement to just imgui


myk002:
I already merged #2684, but could you tack on a changelog entry for that PR here? in the API section for c++ and Lua for Lua

myk002 (commented):
I'm not sure if this should go into the Units module. How about starting a Military module instead? and maybe take `getSquadName(unit *)` with you and reframe it as `getSquadName(squad_id)`. The only user is manipulator. there are no Lua users.

docs/dev/Lua API.rst
```diff
@@ -1563,6 +1563,22 @@ Units module
 
   Returns a table of the cutoffs used by the above stress level functions.
 
+Military Module API
```
myk002, line 1566:
```suggestion
Military module
```

myk002: for consistency with the preceding section titles

docs/dev/Lua API.rst
```diff
@@ -1563,6 +1563,22 @@ Units module
...
+  Creates a new squad associated with the assignment. Fails if one already exists.
+  Note: This function does not name the squad, but they are otherwise complete.
+
+* ``dfhack.military.updateRoomAssignments(squad_id, assignment_id, squad_use_flags)``
+
+  Sets the sleep, train, indiv_eq, and squad_eq flags when training at a barracks.
+
+* ``dfhack.military.getSquadName(squad_id)``
+
+  Returns the name of a squad/
```
myk002, line 1580:
```suggestion
  Returns the name of a squad.
```

library/include/modules/Military.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "Export.h"
+#include "DataDefs.h"
+
+#include "df/squad.h"
+#include "df/unit.h"
```
myk002, line 7: is this header necessary? I don't see any unit pointers here

library/modules/Units.cpp
```diff
@@ -1932,19 +1939,6 @@ bool Units::isGoalAchieved(df::unit *unit, size_t goalIndex)
         && unit->status.current_soul->personality.dreams[goalIndex]->flags.whole != 0;
 }
 
-std::string Units::getSquadName(df::unit *unit)
```
myk002, line 1935: how come removing this code resulted in an *increase* of included header files in Units.cpp?

library/modules/Military.cpp
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    if (corresponding_position == nullptr)
+        return nullptr;
+
+    df::squad* result = new df::squad();
+    result->id = *df::global::squad_next_id;
+    result->cur_routine_idx = 0;
+    result->uniform_priority = result->id + 1; //no idea why, but seems to hold
```
myk002, line 88: yes, this is curious. what information is `uniform_priority` supposed to convey?

library/modules/Military.cpp
```diff
@@ -0,0 +1,304 @@
...
+    result->activity = -1; //??
+    result->carry_food = 2;
+    result->carry_water = 1;
+    result->entity_id = df::global::plotinfo->group_id;
+    result->leader_position = corresponding_position->id;
+    result->leader_assignment = found_assignment->id;
+    result->unk_1 = -1;
```
myk002, line 95: can this get initialized in the structure defs so we don't have to refer to the unknown field by name?

library/modules/Military.cpp
```diff
@@ -0,0 +1,304 @@
...
+    result->name = name;
+    result->ammo.unk_v50_1 = 0;
```
myk002, line 97: this one too

library/modules/Military.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    int16_t squad_size = corresponding_position->squad_size;
+
+    for (int i=0; i < squad_size; i++)
+    {
+        //construct for squad_position seems to set all the attributes correctly
+        //except I've observed unk_2 is -1 generally
+        df::squad_position* pos = new df::squad_position();
+        pos->unk_2 = -1;
```
myk002, line 106: this one three

myk002:
ping on this

20k:
I should have some free time to finish this up fairly soon, sorry for the long delay!

docs/changelog.txt
```diff
@@ -148,9 +148,12 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
...
 - ``Units::getCitizens()``: gets a list of citizens, which otherwise you'd have to iterate over all units the world to discover
-- ``Screen::Pen``: now accepts ``top_of_text`` and ``bottom_of_text`` properties to support offset text in graphics mode
-- `overlay`: overlay widgets can now specify a default enabled state if they are not already set in the player's overlay config file
-- ``Lua::Push``: now supports ``std::unordered_map``
+- Military: New module for military functionality
+- Military: new ``makeSquad`` to create a squad
+- Military: changed ``getSquadName`` to take a squad identifier
+- Military: new ``updateRoomAssignments`` for assigning a squad to a barracks and archery range
+- ``Maps::GetBiomeType`` renamed to ``Maps::getBiomeType`` for consistency
+- ``Maps::GetBiomeTypeRef`` renamed to ``Maps::getBiomeTypeRef`` for consistency
```
myk002, line 156: changelog additions should be shifted up to the current "Future" section

docs/dev/Lua API.rst
```diff
@@ -1580,6 +1580,22 @@ Units module
 
   Returns a table of the cutoffs used by the above stress level functions.
 
+Military module
+~~~~~~~~~~~~~~~~~~~
+
+* ``dfhack.military.makeSquad(assignment_id)``
+
+  Creates a new squad associated with the assignment. Fails if one already exists.
+  Note: This function does not name the squad, but they are otherwise complete.
```
myk002, line 1589: perhaps document how the caller is supposed to name the squad?

myk002: also, what's an assignment and how to you get one?

20k: On names: It uses the general df language_name system, so I'm not 100% sure what the docs would look like here - just say eg "which can be modified via name"? Its worth noting that this uses df's language_name system, and that I'm not sure the reverse engineering has been done for how squad names are generated by the game

20k: On assignments: These are `entity_position_assignment`s, I can update the docs to reflect that but if it might be more helpful to actually document where/how to get one specifically I can do that too

ab9rf: > I'm not sure the reverse engineering has been done for how squad names are generated by the game

ran into that code just the other day, what do you want to know about it?

20k: Oh! I'd love to know how they're generated. So each name has a `words` array, and I *think* for squads only [3] and [6] are actually used - how they're generated would be incredibly useful to know. I'd happily cram that into this PR so that we get game-equivalent randomly generated naming for squads here

I think also that the name for a squad is also generated with a `language` of 0, but I don't know that for a provable fact concretely

Overall I'm pretty unfamiliar with what the language word table is actually doing

ab9rf: ok, here's the code that creates a new squad:
```
                local_88 = (squad *)operator_new(0x1e8);
                psVar24 = squad::_construct(local_88,'\0');
                local_80 = psVar24;
                historical_entity::name_randomizer
                          (plotinfo.main.fortress_entity,&psVar24->name,Squad);
```
`historical_entity::name_randomizer` picks a random name using the word tables for the specified historical entity (which in this case is the fortress entity). The `Squad` name type tells it that the name is being requested for a "squad". However, there is no specific rule for choosing names for squads so the selections are made from the entity's generic word list instead of one of the purpose-specific lists.

The process by which a random name is chosen is _quite_ complex, and while I've disassembled that code I can't claim I fully understand it and certainly can't give you a precise recap of the algorithm, at least not right now.

docs/changelog.txt
```diff
@@ -59,11 +59,18 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
   - ``isGeldable()``
   - ``isMarkedForGelding()``
   - ``isPet()``
+- `Military`: New module for military functionality
+- `Military`: new ``makeSquad`` to create a squad
+- `Military`: changed ``getSquadName`` to take a squad identifier
+- `Military`: new ``updateRoomAssignments`` for assigning a squad to a barracks and archery range- ``Maps::GetBiomeType`` renamed to ``Maps::getBiomeType`` for consistency
```
myk002, line 65: looks like two lines got concatenated here

library/modules/Military.cpp
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    //all we've done so far is leak memory if anything goes wrong
+    //modify state
+    (*df::global::squad_next_id)++;
+    fort->squads.push_back(result->id);
+    df::global::world->squads.all.push_back(result);
+    found_assignment->squad_id = result->id;
+
+    //todo: find and modify old squad
```
myk002, line 219: why? what needs to be done?

library/modules/Military.cpp
```diff
@@ -0,0 +1,296 @@
...
+            {
+                asched[i].sleep_mode = 0;
+                asched[i].uniform_mode = 0;
+            }
+        }
+
+        result->schedule.push_back(reinterpret_cast<df::squad::T_schedule*>(asched));
+    }
+
+    //all we've done so far is leak memory if anything goes wrong
```
myk002, line 212: where are we leaking? I don't see any `return` statements that come after `new` statements above

myk002 (commented):
library/xml is causing conflicts. could you rebase to fix?

library/modules/Military.cpp
20k: This comment is out of date, when I wrote this I intended to have the ability to make a new squad on an existing assignment - deleting the old one in the process, but the current plan is to add an equivalent deleteSquad instead of bundling up too much functionality in

library/modules/Military.cpp
20k: I'm probably too used to code that can throw randomly, you're right in that there isn't currently a recoverable failure case where this can leak memory so I've removed this

docs/dev/Lua API.rst
20k: I've expanded the docs here to make it more clear what the assignment is, and expanding on naming a bit. I could also add a bit on how to find the entity assignment you're looking for, but its not tremendously straightforward so it'd likely add a reasonable chunk to be complete. Its the kind of thing that might be much better in a general overview of the noble position/assignment system in general, but then again maybe something here is better than nothing

20k:
Cleaned up a bunch more of the PR. The only major code-wise unanswered question as far as I can tell is what uniform_priority is doing there

docs/changelog.txt
```diff
@@ -79,11 +79,18 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 - ``Gui::any_civzone_hotkey``, ``Gui::getAnyCivZone``, ``Gui::getSelectedCivZone``:  new functions to operate on the new zone system
 - Units module: added new predicates for ``isGeldable()``, ``isMarkedForGelding()``, and ``isPet()``
+- `Military`: New module for military functionality
+- `Military`: new ``makeSquad`` to create a squad
+- `Military`: changed ``getSquadName`` to take a squad identifier
+- `Military`: new ``updateRoomAssignments`` for assigning a squad to a barracks and archery range
```
myk002, line 85:
```suggestion
- ``Military``: New module for military functionality
- ``Military``: new ``makeSquad`` to create a squad
- ``Military``: changed ``getSquadName`` to take a squad identifier
- ``Military``: new ``updateRoomAssignments`` for assigning a squad to a barracks and archery range
```

myk002 (approved):

PR #2689, opened 2023-01-22
Update orders related to bags for v50.x

dawnmist:
Bags now have their own item type, they're no longer mixed in with boxes


myk002 (approved):

lethosor:
This should have a changelog entry. Moving to in progress as a reminder.

Issue #2688, opened 2023-01-22
fix lye and minecart library orders


myk002:
lye is missing the lye-containing trait and so never stops
minecarts match in-use minecarts and so never starts

cdombroski:
Minecarts aren't fixed yet. An empty minecart registers as empty even if assigned to a route. Not sure what the fix should be yet

PR #2690, opened 2023-01-22
Attempt to fix lye order limits

dawnmist:
This does require the lye to be in buckets rather than barrels for the limits to apply to a reasonable amount of lye. Make sure that you bar barrels from the lye stockpile.


myk002 (approved):

myk002:
#2688

PR #2714, opened 2023-01-25
tailor: increased check frequency

ab9rf:
i think the previous window of 600 ticks was allowing it to miss bookkeeper runs, causing tailor to run less often than intended

reducing to 50 to reduce the chances of this happening


myk002 (approved):
and we can likely remove this entirely once we have a proper EventManager event for this.

PR #2715, opened 2023-01-25
Add Github Actions build of win64 artifacts, and update compile instructions.

kelvie:
Thanks to @BenLubar for setting up the official build images on ghcr.io, we can now have automated win64 builds without requiring a win64 builder.

This updates the build to now generate Windows artifacts (for Steam users) for every single PR and commit. See here (scroll down):

https://github.com/kelvie/dfhack/actions/runs/4002114641

One side effect this will have is that the build now takes 20 minutes, as the new bottleneck is this new cross-win64 build, and this may be an issue with e.g. PRs that should be quick.

I haven't tried getting `cccache` working, though I don't think this will help the worst case PRs anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,6 +134,14 @@ jobs:
         with:
           submodules: true
           fetch-depth: 0
+      - name: Fetch ccache
+        uses: actions/cache@v3
+        with:
+          path: build/win64-cross/ccache
+          key: ccache-win64-cross-msvc-${{ github.ref_name }}-${{ github.sha }}
+          restore-keys: |
+            ccache-win64-cross-msvc-${{ github.ref_name }}
+            ccache-win64-cross-msvc
       - name: Cross-compile win64 artifacts
         run: |
           cd build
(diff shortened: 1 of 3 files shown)


kelvie:
Looks like the artifact name generation isn't working for PR builds. I'll try to fix it, but I now see the value of ccache.

myk002:
This isn't based on the image I put together? ccache was working there.

myk002:
note that this should use the ccache built for the previous build image since stock ccache has a processing error when used with cl.exe in this linux-based environment.

myk002 (commented):
ok, so ccache should already be installed in the image, right? you just need to add an appropriate ccache action

kelvie:
Yup, ccache is already in there and would probably work, I just need to choose a cache key, do the cache fetch and save and test it.

kelvie:
Oh, and I guess figure out how to mount the cache directory from within that script. Hmm, I may need to give it a way to pass in arguments (or just run the docker command directly in github actions)

kelvie:
OK, well setting up the cache step was a lot easier than I thought. The rebased build took 4 mins (vs 20 from scratch). I think this is good to go now.

myk002:
where can the built artifact be downloaded from?

kelvie:
Yeah the artifacts aren't in the per-build page, but rather the "Summary" page

PR #2719, opened 2023-01-25
Expose winapi heap debugging tools to lua

20k:
_msize returns the real allocation size of a pointer by using windows heap black magic. This function is significantly more useful than the equivalent on linux (which seems to have no equivalent), which only returns the *block* size - and so cannot really be used for what we need it for. Eg running it on a unit pointer returns 5136 , whereas on linux it'd probably return something much more block-sized 

This would have detected more than one misalignment as well (aliased pointers) and is how I found that vision_cone was misaligned, and would very likely be useful to integrate into gm-edit

I don't have a wine install, but its absolutely worth checking the behaviour there, as this is one of the things that will likely diverge

Edit:

This PR now contains a reasonably complete set of tools for debugging the heap. I'm going to file a PR to update sizecheck for both dangling pointers (which can now detect the zone bug!), and memory misalignments


docs/dev/Lua API.rst
```diff
@@ -614,6 +614,10 @@ Miscellaneous
   both from the curry call and the closure call itself. I.e.
   ``curry(func,a,b)(c,d)`` equals ``func(a,b,c,d)``.
 
+* ``dfhack.msize(ptr)``
```
myk002, line 617: if we expose this, I think it should be in the `dfhack.internal` namespace since we do not expect end user scripts to be calling it

20k: Will do. I'm going to rework this PR because as-is its insufficient for a lot of things we'd want to do with it, so I'll give you a ping

myk002:
Is this PR ready for review? It's not quite my forte, though, and I'd want opinions from the others on the utility of these APIs

myk002 (commented):
is there an accompanying script that uses these calls?

20k:
Sorry for the late reply! I've been taking a break for a while, but I should be more around now

https://github.com/20k/scripts/blob/dfuirebase/devel/sc.lua

This is the modified sizecheck I've been using to find misaligns. In hindsight after thinking about this for a while, the api is more inefficient than it could be (returning a struct/map with a general getPointerInfo function), but the intent was to allow potential implementation on non windows platforms, as some information can't be exposed on linux (what's the size of this pointer?), but some could be (is this pointer dangling?)

There are also issues around using this for heap debugging when the heap is corrupt, given that both lua and the snapshot function heap allocate. These are both fixable, but this is probably a separate project, and is of questionable utility outside of the occasional post hoc heap corruption debugging situation. Though, that's not incredibly uncommon either, so if someone wanted it I could add it to the todo list

I do need to change this to return nil when a pointer isn't found instead of returning 0/-1 though, so I'll update this pr

`msizeAddress`, `getHeapState`, and `getRootAddressOfHeapObject` are not used in that script. msizeAddress wraps `_msize` on windows and is much faster if you know a pointer is valid, and getHeapState wraps `_heapchk` which is largely there for completeness (and its occasionally useful in a post-hoc crash debugging situation). `getRootAddressOfHeapObject` is one of the building blocks for detecting if any arbitrary pointer is strictly valid and dereferenceable, which would very likely be useful for a variety of things - eg gm-edit, or making the lua api safer

myk002:
that sounds good - will you be submitting a PR for the updated `sc.lua` too?

20k:
I was planning to file that separately, but I can file that at the same time as this after I've updated everything if it would be helpful

myk002:
It'll have to be a separate PR since it's a different repo, but it would be nice to have the PRs concurrently so we could see how the new functions are used.

myk002 (approved):

myk002:
when you get a chance, the script that uses these functions would still be welcome

lethosor:
I am late, but I think generally this sort of API should be made available to C++ as well (i.e. LuaApi is for Lua wrappers around C++ APIs, aside from cases where there really is no possible C++ equivalent).

myk002:
We can definitely refactor these if there is a need for it. As it is, the calls are not expected to be used by any C++ tools (though they could be written). I think the proof will be with this mysterious `sc.lua` file that makes use of these APIs. If it meets our requirements, then a Lua API is all we really need.

20k:
I'm likely to have some time to do more dfhack development in the relatively near future, I'll get on it sc.lua because its helpful. Its floating around in a fork at the moment, so I'll make a PR once I've got my dev environment up again

Creating a C++ API for it would be straightforward to do, but the API is a pretty thin wrapper around the winapi heap debugging system with some room left for a hypothetical linux implementation. Its essentially a simple cache to mitigate the performance problems of repeatedly iterating over the heap. That said, the use cases for this in C++-land are small, because there's few situations in which you *want* to call these APIs in general other than to do possible pre-crash debugging, or structure misalignment checking

I could imagine it being useful for doing debug checking of lua pointers in the C++ land, to cut down on crashes, so some consideration could be made for that, but you'd likely want a very different implementation *anyway*

That said, I'm totally up for whatever if there's call for it

lethosor:
`sc` is not just in a fork: https://github.com/DFHack/scripts/blob/master/devel/sc.lua

> That said, the use cases for this in C++-land are small, because there's few situations in which you want to call these APIs in general other than to do possible pre-crash debugging, or structure misalignment checking

We have a `check-structures-sanity` tool written in C++, partly for performance and type-safety reasons. (I don't remember the entire reasoning). There are also tools like `memview` that can and do make use of size information when it's available through previous custom tooling.

PR #2721, opened 2023-01-26
Enable Confirm

robob27:
- Rename `Gui::getFocusString` -> `Gui::getFocusStrings`. Now returns a table of strings representing the current focuses. This now uses data from `df.global.game.main_interface` instead of separate viewscreens since almost everything is contained in a single viewscreen called dwarfmode now. Currently the taxonomy is very different from what it was previously.
- Re-enable confirm, mostly, so far
- Add the ability to detect left/right clicks in confirmations. Intercepted left clicks are passed to the screen in the position they were intercepted in if the action is confirmed. Intercepted right clicks are sent as `interface_key::LEAVESCREEN` to avoid doing more click simulation jank.
- Confirmations now keep track of their paused status individually
- Added a `focusString` param to the `DEFINE_CONFIRMATION` template. When a confirmation is paused, if ESC/right clicks are detected, the confirmation will check for the presence of `focusString` in `Gui::getFocusStrings` on the next render. If not found, that individual confirmation will be unpaused.
- Add new removal confirms: zone, stockpile and burrow


library/Core.cpp
```diff
@@ -948,7 +948,7 @@ command_result Core::runCommand(color_ostream &con, const std::string &first_, v
                 << "Context may be used to limit the scope of the binding, by" << endl
                 << "requiring the current context to have a certain prefix." << endl
                 << "Current UI context is: "
-                << Gui::getFocusString(Core::getTopViewscreen()) << endl;
+                << Gui::getFocusString(Core::getTopViewscreen())[0] << endl;
```
lethosor, line 951: Would it make sense to print the entire vector here?

Also, is it possible for this vector to be empty? A brief look at `Gui.cpp` suggests that it could be.

library/Core.cpp
```diff
@@ -2419,7 +2419,7 @@ bool Core::SelectHotkey(int sym, int modifiers)
                                         binding.modifiers, modifiers);
                 continue;
             }
-            string focusString = Gui::getFocusString(screen);
+            string focusString = Gui::getFocusString(screen)[0];
```
lethosor, line 2422: Do you not want to check against every string?

library/modules/Gui.cpp
```diff
@@ -114,14 +114,14 @@ static df::layer_object_listst *getLayerList(df::viewscreen_layer *layer, int id
 }
 */
 
-static std::string getNameChunk(virtual_identity *id, int start, int end)
+static std::string getNameChunk(virtual_identity* id, int start, int end)
```
lethosor, line 117: We don't really have a standard for `type *var` vs `type* var`, and there are merits for both, but I'd rather avoid the noisy diffs from changing it from one to the other

plugins/confirm.cpp
```diff
@@ -16,27 +16,28 @@
...
-#include "df/viewscreen_jobmanagementst.h"
-#include "df/viewscreen_justicest.h"
-#include "df/viewscreen_layer_militaryst.h"
-#include "df/viewscreen_locationsst.h"
-#include "df/viewscreen_tradegoodsst.h"
+//#include "df/viewscreen_jobmanagementst.h"
+//#include "df/viewscreen_justicest.h"
+//#include "df/viewscreen_layer_militaryst.h"
+//#include "df/viewscreen_locationsst.h"
+//#include "df/viewscreen_tradegoodsst.h"
```
lethosor, line 28: These can be deleted entirely. The classes are removed from the game and are unlikely to come back soon.

plugins/confirm.cpp
```diff
@@ -16,27 +16,28 @@
...
 
 using namespace DFHack;
 using namespace df::enums;
 using std::map;
 using std::queue;
 using std::string;
 using std::vector;
+using df::global::game;
```
lethosor, line 36: You should use `REQUIRE_GLOBAL(game)` instead. That ensures that the global address is available when loading the plugin.

plugins/uicommon.h
```diff
@@ -198,13 +198,14 @@ static inline char get_string_input(const std::set<df::interface_key> *input)
 
 static inline df::building_stockpilest *get_selected_stockpile()
 {
-    if (!Gui::dwarfmode_hotkey(Core::getTopViewscreen()) ||
+    return nullptr;
```
lethosor, line 201: Was this not working in v50? As far as I can tell, it should have been compiling. If there's an ability to select a building in DF (which I'm pretty sure there is), we should be returning the selected building as before.

plugins/hotkeys.cpp
```diff
@@ -90,7 +90,7 @@ static void find_active_keybindings(df::viewscreen *screen, bool filtermenu) {
 
     valid_keys.push_back("`");
 
-    current_focus = Gui::getFocusString(screen);
+    current_focus = Gui::getFocusString(screen)[0];
```
lethosor, line 93: I think we should probably have a way to check against multiple focus strings.

plugins/confirm.cpp
```diff
@@ -565,7 +580,8 @@ DFhackCExport command_result plugin_enable (color_ostream &out, bool enable)
     }
     if (is_enabled)
     {
-        conf_lua::simple_call("check");
+        //conf_lua::simple_call("check"); this doesn't work anymore, why?
```
lethosor, line 583: What do you mean by "doesn't work"? Was it giving an error? It will throw an error deliberately if the list of confirmations between Lua and C++ doesn't line up.

lethosor (changes requested):
(I realize this is a draft... the only hard requested change here is the REQUIRE_GLOBAL one)

library/modules/Gui.cpp
robob27: Yeah this was an accidental "format selection" by the looks of it. My visual studio settings are very opinionated about the positioning of the `*` lol. Will revert for sure, along with any other style only changes that aren't driven by CI yelling at me.

plugins/confirm.cpp
robob27: Ah good to know. I was getting an error "attempted to call a nil value. stack trace:" (no stack trace) when I enabled the plugin, after I had made a bunch of changes. I eventually traced it to this line and noticed that the rest of the script seemed to work when I commented it out. So I wrote a note to myself to find out why.

library/Core.cpp
robob27: Yeah this is janky atm. I have a fallback at the end of the focus string handler for dwarfmode that ensures at least "dwarfmode" is in the vector if it didn't add anything else. Assuming the dwarfmode screen is always present now, it should at least have that, but yeah I definitely need to update any code that relies on `getFocusString` to not just take the first element.

myk002: the dwarfmode screen will not be in every context. only specifically when the fort map is visible, and sometimes not even then if a separate viewscreen (not just a panel) is partially covering it. For a simple example, when you first start the game and you see the dwarf face being constructed line by line, that's viewscreen_setupdwarfgamest, and the title screen (with the splash image in the background) is viewscreen_titlest

robob27: Hah, yeah I actually found this out by hitting the launcher hotkey earlier while in the title screen and immediately thought of this comment 🙈

plugins/uicommon.h
robob27: Would it make sense to migrate this function to `Gui:getAnyStockpile()`/`Gui:getSelectedStockpile()`?

robob27:
```
df::building_stockpilest* Gui::getAnyStockpile(df::viewscreen* top) {
    if (Gui::getFocusString(Core::getTopViewscreen())[0] == "dwarfmode/Some/Stockpile") {
        return game->main_interface.stockpile.cur_bld;
    }

    return NULL;
}

df::building_stockpilest* Gui::getSelectedStockpile(color_ostream& out, bool quiet) {
    df::building_stockpilest* stockpile = getAnyStockpile(Core::getTopViewscreen());

    if (!stockpile && !quiet)
        out.printerr("No stockpile is selected in the UI.\n");

    return stockpile;
}
```

?

myk002: I'm actually surprised that it's not already returned by getSelectedBuilding() -- stockpiles are just buildings. If callers care if it's a stockpile, they can check themselves.

myk002: if it's covered by getSelectedBuilding(), I'd say we don't even need it. Adding a getSelectedFoo() call comes with quite a bit of overhead since it then needs to be passed through a lot of other tools that implement the interface

robob27: Makes sense! I think the [current implementation](https://github.com/DFHack/dfhack/blob/develop/library/modules/Gui.cpp#L1279-L1281) of `getSelectedBuilding` won't cover it because it doesn't use view sheets, but that should be easy enough to fix.

library/Core.cpp
```diff
@@ -947,8 +947,8 @@ command_result Core::runCommand(color_ostream &con, const std::string &first_, v
                 << "Supported keys: [Ctrl-][Alt-][Shift-](A-Z, 0-9, F1-F12, `, or Enter)." << endl
                 << "Context may be used to limit the scope of the binding, by" << endl
                 << "requiring the current context to have a certain prefix." << endl
-                << "Current UI context is: "
-                << Gui::getFocusString(Core::getTopViewscreen()) << endl;
+                << "Current UI context is: " << endl
+                << join_strings("\n", Gui::getFocusString(Gui::getDFViewscreen())) << endl;
```
myk002, line 951: I think this should be `join_strings("\n", Gui::getFocusString(Gui::getTopViewscreen(true)))`. That is:
1. ignore dismissed viewscreens at the top of the stack
2. unfocused ZScreens are ignored when determining the focus string (this might take changes to the Lua ZScreen implementation)
3. the viewscreen used to determine the focus string is the focused ZScreen (if any) or the first non-ZScreen on the stack. if that screen is a DF viewscreen, then further context can be determined

This will make the focus string correspond to the viewscreen that actually has keyboard focus

this might take some explaining, but I wanted to get the requirements written down here first

library/Core.cpp
```diff
@@ -2419,11 +2419,22 @@ bool Core::SelectHotkey(int sym, int modifiers)
...
-            if (!binding.focus.empty() && !prefix_matches(binding.focus, focusString)) {
-                DEBUG(keybinding).print("skipping keybinding due to focus string mismatch: '%s' !~ '%s'\n",
-                                        focusString.c_str(), binding.focus.c_str());
-                continue;
+            if (!binding.focus.empty()) {
+                // TODO: understand more about this to figure out if this solution works
+                bool found = false;
+                std::vector<std::string> focusStrings = Gui::getFocusString(Core::getTopViewscreen());
+                // is there convention for when to use size_t vs int?
+                for (size_t ii = 0; ii < focusStrings.size(); ii++) {
```
myk002, line 2427:
```suggestion
                for (std::string focusString : focusStrings) {
```

myk002: the standard is to use foreach syntax when possible, but if not then the var should be size_t and cast where necessary.

library/LuaApi.cpp
```diff
@@ -1621,6 +1634,7 @@ static const luaL_Reg dfhack_gui_funcs[] = {
     { "pauseRecenter", gui_pauseRecenter },
     { "revealInDwarfmodeMap", gui_revealInDwarfmodeMap },
     { "getMousePos", gui_getMousePos },
+    { "getFocusString", gui_getFocusString },
```
myk002, line 1637: you might as well rename this to `getFocusStrings` since we're changing the return value anyway.

library/LuaApi.cpp
```diff
@@ -1493,9 +1493,22 @@ static const LuaWrapper::FunctionReg dfhack_gui_module[] = {
...
 };
 
+static int gui_getFocusString(lua_State *state) {
+    df::viewscreen *r = Lua::GetDFObject<df::viewscreen>(state, 1);
+    std::vector<std::string> focusStrings = Gui::getFocusString(r);
+
+    lua_newtable(state);
+    for (size_t i = 0; i < focusStrings.size(); i++) {
+        Lua::TableInsert(state, i, focusStrings[i]);
+    }
```
myk002, line 1507:
```suggestion
    Lua::PushVector(focusStrings);
```

library/include/modules/Gui.h
```diff
@@ -107,6 +110,10 @@ namespace DFHack
         DFHACK_EXPORT df::building *getAnyBuilding(df::viewscreen *top);
         DFHACK_EXPORT df::building *getSelectedBuilding(color_ostream &out, bool quiet = false);
 
+        DFHACK_EXPORT bool any_stockpile_hotkey(df::viewscreen* top);
+        DFHACK_EXPORT df::building_stockpilest *getAnyStockpile(df::viewscreen* top);
+        DFHACK_EXPORT df::building_stockpilest *getSelectedStockpile(color_ostream& out, bool quiet = false);
```
myk002, line 115: `any_stockpile_hotkey` is good, but I do recommend against `getAnyStockpile` and `getSelectedStockpile` unless there is a proven need for them. This will affect all tools that implement the `getSelected*` API (e.g. `gui/launcher`)

actually, let me walk that back. if we can treat a non-focused ZScreens as invisible, then there is no need for these tools to implement these functions just to pass them through to the lower layer. So go ahead and add these. We can clean up the tools later and simplify them instead of make them more complicated.

library/include/modules/Gui.h
```diff
@@ -191,7 +198,7 @@ namespace DFHack
             return strict_virtual_cast<T>(getViewscreenByIdentity(T::_identity, n));
         }
 
-        inline std::string getCurFocus(bool skip_dismissed = false) {
+        inline std::vector<std::string> getCurFocus(bool skip_dismissed = false) {
```
myk002, line 201: @lethosor do you happen to know why `skip_dismissed` defaults to false? what is the use case for wanting a reference to a dismissed viewscreen?

library/modules/Gui.cpp
```diff
@@ -130,529 +130,307 @@ static std::string getNameChunk(virtual_identity *id, int start, int end)
...
+    if (game->main_interface.squad_equipment.open) {
+        newFocusString = baseFocus;
+        newFocusString += "/SquadEquipment";
+        focusStrings.push_back(newFocusString);
     }
-*/
-}
-
-/* TODO: understand how this changes for v50
-DEFINE_GET_FOCUS_STRING_HANDLER(dungeonmode)
```
myk002, line 366: you should probably leave this in but commented out for the time when dungeonmode (adventure mode) rises again.

library/modules/Gui.cpp
```diff
@@ -130,529 +130,307 @@ static std::string getNameChunk(virtual_identity *id, int start, int end)
...
-    default:
-        break;
+    // squads should be last because it's the only one not exclusive with the others? or something?
+    if (game->main_interface.squads.open) {
+        newFocusString = baseFocus;
+        newFocusString += "/Squads";
+        focusStrings.push_back(newFocusString);
+    }
+
+    if (focusStrings.size() == 0) {
```
myk002, line 342: since you don't clear focusStrings (and I'm not sure if you should), this can't truly be relied on. `if (!newFocusString.size()) {` would work though, I think

library/modules/Gui.cpp
```diff
@@ -130,529 +130,307 @@ static std::string getNameChunk(virtual_identity *id, int start, int end)
...
-        focus += "/Orders";
-        break;
-    case T_tab::Restrictions:
-        focus += "/Restrictions";
-        break;
-    }
-}
+bool Gui::matchFocusString(std::string focusString) {
+    focusString = toLower(focusString);
+    std::vector<std::string> currentFocus = getFocusString(Core::getTopViewscreen());
```
myk002, line 349:
```suggestion
    std::vector<std::string> currentFocus = getFocusString(Core::getTopViewscreen(true));
```

library/modules/Gui.cpp
```diff
@@ -130,529 +130,307 @@ static std::string getNameChunk(virtual_identity *id, int start, int end)
...
-
-DEFINE_GET_FOCUS_STRING_HANDLER(layer_overall_health)
-{
-    auto list1 = getLayerList(screen, 0);
-    if (!list1) return;
-
-    focus += "/Units";
+    return std::find_if(currentFocus.begin(), currentFocus.end(), [&focusString](std::string item) {
+        //Transform the each vector item to lower temporally 
+        return focusString == toLower(item);
```
myk002, line 353: I think this should be a prefix match like keybinding expects. that will be the general use case: "does the focus string spec I'm passing in match the current context?"

library/modules/Gui.cpp
```diff
@@ -130,529 +130,307 @@ static std::string getNameChunk(virtual_identity *id, int start, int end)
...
-}
-*/
-
-std::string Gui::getFocusString(df::viewscreen *top)
-{
     if (!top)
-        return "";
+        return focusStrings;
 
     if (dfhack_viewscreen::is_instance(top))
```
myk002, line 364:
```suggestion
    if (dfhack_viewscreen::is_instance(top))
        top = static_cast<dfhack_viewscreen*>(top)->getFocusedScreen();

    if (dfhack_viewscreen::is_instance(top))
```

myk002: We'll need to check here whether we fall through ZScreens. getFocusedScreen() will need to be implemented for DFHack's viewscreen implementation in the Screen module

library/modules/Gui.cpp
```diff
@@ -1601,27 +1384,27 @@ bool Gui::addCombatReport(df::unit *unit, df::unit_report_type slot, int report_
...
         rvec.push_back(vec[i]->id);
 
     return true;
 }
 
-bool Gui::addCombatReportAuto(df::unit *unit, df::announcement_flags mode, int report_index)
+bool Gui::addCombatReportAuto(df::unit* unit, df::announcement_flags mode, int report_index)
 {
-    auto &vec = world->status.reports;
+    auto& vec = world->status.reports;
```
myk002, line 1407: I'll review this again after the formatting changes are cleaned up. this is breaking my brain : p

plugins/uicommon.h
myk002: My eventual plan is to move useful stuff out of uicommon and into core, then delete that file. Option 2 above is good.

library/modules/Gui.cpp
robob27: Sorry ><. I didn't think about editor formatting settings going into this at all and it's biting me in the ass. I'll get it sorted and comment when I do.

robob27: Formatting stuff fixed up. Still lots of work to do though.

plugins/uicommon.h
lethosor: I don't think we need `getAnyStockpile`, just `getSelectedStockpile`. `getAnyBuilding`/`getSelectedBuilding` already return the selected building, and the only purpose of `get_selected_stockpile()` is to filter out any selected buildings that aren't stockpiles. It's just a thin wrapper around `getSelectedBuilding()`, even though it isn't actually implemented as such.

It is probably also true that `get_selected_stockpile()` could be replaced with `virtual_cast<df::building_stockpilest>(getSelectedBuilding())`. I think that's true in all cases...

robob27:
@myk002 I'm a bit stuck on how to properly detect left/right clicks for confirmations.. are you able to help shed any light on that?

robob27:
The last thing I'm not really sure how to sort is how to detect if hovering over a button when hovering over the button doesn't seem to change something. When you hover over "Disband Squad", `game.main_interface.current_hover` is set to 341 which makes this easily detectable. I'm unable to find something similar for the Trade button (and likely many others). Is there actually any reliable solution to this currently?

myk002:
you're exploring new territory here. I don't have any answers, unfortunately. Are there any hints for which screen coordinates the button is being displayed at?

Putnam may have some insights on the value of current_hover, and what it means for the trade button.

myk002:
just as a note, it makes it difficult to track what is changing when you squash and force push. if you're worried about many commits, we can squash them at the end. in the meantime, it would be useful to keep the commits separate so I can check on your progress and offer guidance.

robob27:
> Are there any hints for which screen coordinates the button is being displayed at?

I haven't found anything yet. I will keep searching on the side as I work on implementing the rest of the feedback and re-enabling the confirmations that don't require exploring the unknown.

> Putnam may have some insights on the value of current_hover, and what it means for the trade button.

From what I can tell `current_hover` gets set when you're hovering over any UI element with some kind of tooltip (whether it replaces the minimap or displays beside the cursor). Unfortunately some stuff doesn't do this, such as the trade button, uniform delete button etc.

I can't really see a way forward on some confirmations without being able to find something that tells us the coordinates of the buttons, or when hovering over the button mutates some data that we can access. But maybe we'll find something eventually.

myk002:
> I can't really see a way forward on some confirmations without being able to find something that tells us the coordinates of the buttons, or when hovering over the button mutates some data that we can access.

if the buttons are a constant distance from any two edges, we have a way to detect hover by adding an invisible widgets.Panel with a frame that holds that distance from the two edges. If we can find such relationships with the screen edges we also need to verify that the numbers are constant at different zoom levels. But if all that checks out, then we may have a semi-convenient way of doing it.

robob27:
Still trying to figure this out but mentioning here in case there is a known easy fix.

I'm having an issue with the confirmation dialogs lingering when certain UIs are open:

- if i open the hauling window and delete a route, and dismiss the confirmation dialog by right clicking or pressing escape, the dialog will linger.
- if i open the squads window and disband a squad, when i dismiss the "Disband Squad" confirmation dialog by right clicking or pressing escape, it does not linger
- if i open up any window on the left side of the screen that doesn't completely obscure where the confirmation window appears alongside the Squads window (hauling window in this video), the "Disband Squad" dialog now also lingers after being dismissed.

https://user-images.githubusercontent.com/2104555/216230887-b0b089e5-1341-418a-9e9b-44bf9d369001.mp4

So far the only thing I've found to try was `Screen::invalidate()` which doesn't seem to fix it.

myk002:
This graphical "shadow" or "ghosting" or "dirty blitting" sounds similar to an issue I solved with ZScreens. Does it help if you set `df.global.gps.force_full_display_count = 1` when you dismiss the confirm dialog?

library/lua/gui/dwarfmode.lua
```diff
@@ -48,7 +48,7 @@ function enterSidebarMode(sidebar_mode, max_esc)
     local focus_string = ''
     while remaining_esc > 0 do
         local screen = dfhack.gui.getCurViewscreen(true)
-        focus_string = dfhack.gui.getFocusString(screen)
+        focus_string = dfhack.gui.getFocusStrings(screen)
```
myk002, line 51: this clearly will not work, but this entire function is not necessary anymore and can be deleted, so it doesn't matter.

library/modules/Screen.cpp
```diff
@@ -1081,7 +1081,7 @@ using df::identity_traits;
 #define CUR_STRUCT dfhack_viewscreen
 static const struct_field_info dfhack_viewscreen_fields[] = {
     { METHOD(OBJ_METHOD, is_lua_screen), 0, 0 },
-    { METHOD(OBJ_METHOD, getFocusString), 0, 0 },
+    { METHOD(OBJ_METHOD, getFocusStrings), 0, 0 },
```
myk002, line 1084: I believe this is the (singular) focus string provided by the DFHack viewscreen. it shouldn't be changed to the plural.

plugins/buildingplan/buildingplan.cpp
```diff
@@ -56,7 +56,7 @@ class ViewscreenChooseMaterial : public dfhack_viewscreen
 
     void render();
 
-    std::string getFocusString() { return "buildingplan_choosemat"; }
+    std::string getFocusStrings() { return "buildingplan_choosemat"; }
```
myk002, line 59: same here. this is a singular focus string. no need to change.

plugins/dwarfmonitor.cpp
```diff
@@ -410,7 +410,7 @@ class ViewscreenDwarfStats : public dfhack_viewscreen
         OutputHotkeyString(x, y, "Zoom Unit", CUSTOM_SHIFT_Z);
     }
 
-    std::string getFocusString() { return "dwarfmonitor_dwarfstats"; }
+    std::vector<std::string> getFocusStrings() { return std::vector<std::string>{"dwarfmonitor_dwarfstats"}; }
```
myk002, line 413: same here. this is singular

myk002: changes to this file can be reverted.

plugins/embark-assistant/finder_ui.cpp
```diff
@@ -1613,7 +1613,7 @@ namespace embark_assist {
 
             void render();
 
-            std::string getFocusString() { return "Finder UI"; }
+            std::vector<std::string> getFocusStrings() { return std::vector<std::string>{"Finder UI"}; }
```
myk002, line 1616: revert

plugins/embark-assistant/help_ui.cpp
```diff
@@ -31,7 +31,7 @@ namespace embark_assist{
 
             void render();
 
-            std::string getFocusString() { return "Help UI"; }
+            std::vector<std::string> getFocusStrings() { return std::vector<std::string>{"Help UI"}; }
```
myk002, line 34: revert

robob27:
> This graphical "shadow" or "ghosting" or "dirty blitting" sounds similar to an issue I solved with ZScreens. Does it help if you set `df.global.gps.force_full_display_count = 1` when you dismiss the confirm dialog?

Tried adding it [here](https://github.com/DFHack/dfhack/pull/2721/commits/ac2568c6ba00049b233d9e6544264a0ba8f15be9) and it doesn't seem to 🤔

plugins/embark-tools.cpp
```diff
@@ -572,7 +572,7 @@ class embark_tools_settings : public dfhack_viewscreen
     embark_tools_settings () { };
     ~embark_tools_settings () { };
     void help () { };
-    std::string getFocusString () { return "embark-tools/options"; };
+    std::vector<std::string> getFocusStrings () { return std::vector<std::string>{"embark-tools/options"}; };
```
myk002, line 575: revert

plugins/hotkeys.cpp
```diff
@@ -90,7 +90,7 @@ static void find_active_keybindings(df::viewscreen *screen, bool filtermenu) {
 
     valid_keys.push_back("`");
 
-    current_focus = Gui::getFocusString(screen);
+    current_focus = Gui::getFocusStrings(screen)[0];
```
myk002, line 93: this is the same algorithm as the keybinding match, essentially. it will need all the focus strings, and will need to match any of them down on line 115

plugins/manipulator.cpp
```diff
@@ -864,7 +864,7 @@ class viewscreen_unitbatchopst : public dfhack_viewscreen {
             }
         }
     }
-    std::string getFocusString() { return "unitlabors/batch"; }
+    std::vector<std::string> getFocusStrings() { return std::vector<std::string>{"unitlabors/batch"}; }
```
myk002, line 867: can revert changes to this file

plugins/stocks.cpp
```diff
@@ -620,7 +620,7 @@ class search_help : public dfhack_viewscreen
         for (auto line = lines.begin(); line != lines.end(); ++line)
             OutputString(COLOR_WHITE, x, y, line->c_str(), true, left_margin);
     }
-    std::string getFocusString() { return "stocks_view/search_help"; }
+    std::vector<std::string> getFocusStrings() { return std::vector<std::string>{"stocks_view/search_help"}; }
```
myk002, line 623: can revert changes to this file

test/library/gui/dwarfmode.lua
```diff
@@ -26,7 +26,7 @@ function test.enterSidebarMode()
     -- Simulate not being able to get to default from a screen via mocks. This
     -- failure can actually happen in-game in some situations, such as when
     -- naming a building with ctrl-N (no way to cancel changes).
-    mock.patch({{dfhack.gui, 'getFocusString', mock.func()},
+    mock.patch({{dfhack.gui, 'getFocusStrings', mock.func()},
```
myk002, line 29: this test can be removed

myk002:
> > This graphical "shadow" or "ghosting" or "dirty blitting" sounds similar to an issue I solved with ZScreens. Does it help if you set `df.global.gps.force_full_display_count = 1` when you dismiss the confirm dialog?
> 
> Tried adding it [here](https://github.com/DFHack/dfhack/pull/2721/commits/ac2568c6ba00049b233d9e6544264a0ba8f15be9) and it doesn't seem to 🤔

try setting it while rendering. See the implementation in gui.Screen here: https://github.com/DFHack/dfhack/blob/develop/library/lua/gui.lua#L637

plugins/confirm.cpp
```diff
@@ -16,27 +16,24 @@
...
-#include "df/viewscreen_locationsst.h"
-#include "df/viewscreen_tradegoodsst.h"
 
 using namespace DFHack;
 using namespace df::enums;
 using std::map;
 using std::queue;
 using std::string;
 using std::vector;
+using df::global::game;
```
myk002, line 31: this line isn't necessary. it's provided by `REQUIRE_GLOBAL(game)`

test/library/gui/dwarfmode.lua
robob27: Tonight I was thinking "man, I really hope myk002 doesn't think this has been more trouble than it's worth..". That was 20 comments ago lmao. I super appreciate your help and hope my lack of knowledge here isn't _too_ annoying 😅

myk002: not at all. you're doing great!

It would be nice to merge in the widespread focus string changes in a separate PR and leave this PR for just confirm.  That way I can work on focus string handling in ZScreens while you are finishing up confirm

robob27:
> > > This graphical "shadow" or "ghosting" or "dirty blitting" sounds similar to an issue I solved with ZScreens. Does it help if you set `df.global.gps.force_full_display_count = 1` when you dismiss the confirm dialog?
> > 
> > 
> > Tried adding it [here](https://github.com/DFHack/dfhack/pull/2721/commits/ac2568c6ba00049b233d9e6544264a0ba8f15be9) and it doesn't seem to 🤔
> 
> try setting it while rendering. See the implementation in gui.Screen here: https://github.com/DFHack/dfhack/blob/develop/library/lua/gui.lua#L637

Works like a charm! Thank you!

robob27:
I need to fix up the pausing logic. Previously if you paused confirmations and then did `interface_key::LEAVESCREEN` it was pretty much certain the the paused focus would no longer be on screen. Now with multiple focuses possible this isn't always the case. Buyer beware for now.

test/library/gui/dwarfmode.lua
robob27: Thanks 😄.

>It would be nice to merge in the widespread focus string changes in a separate PR

Okay cool I can work on that. I'll focus my next efforts on getting the rest of the focus strings I was missing in and getting strictly focus string related stuff into a separate PR.

Then I'll try the suggestion about the detecting the trickier buttons. I wish every button just had a tooltip 😢

myk002:
> I need to fix up the pausing logic. Previously if you paused confirmations and then did `interface_key::LEAVESCREEN` it was pretty much certain the the paused focus would no longer be on screen (for dwarfmode confirms which required escape detection - which is now all confirms). Now with multiple focuses possible this isn't always the case. Buyer beware for now.

instead of a single isPaused boolean, does it now make sense to have a flag per confirm dialog, which is set to false when the context for that dialog no longer matches?

library/modules/Gui.cpp
robob27: `Gui::getCurViewscreen(true)`? `Core::getTopViewscreen` doesn't seem to take a boolean.

plugins/uicommon.h
robob27: So far I've implemented option 2 above. I really feel like this approach makes the most sense since we retrieve the data about which stockpile is active from a different place in `game` than where we retrieve the building data.

If I make the suggested change, `getSelectedBuilding` would need to be updated to check another spot in `game` to return the currently selected stockpile, before or after checking for a building (`main_interface.view_sheets.active_id` vs `game->main_interface.stockpile.cur_bld`), right? Is that a better tradeoff than implementing the methods as is? Apologies in advance if I'm misunderstanding 😅.

Either way, happy to do whichever approach is considered best!

library/Core.cpp
robob27: Do you want me to look into the ZScreen stuff you mentioned here or is that the stuff you wanna work on after the focus string stuff is merged?

robob27:
> if the buttons are a constant distance from any two edges, we have a way to detect hover by adding an invisible widgets.Panel with a frame that holds that distance from the two edges. If we can find such relationships with the screen edges we also need to verify that the numbers are constant at different zoom levels. But if all that checks out, then we may have a semi-convenient way of doing it.

What are your thoughts on shipping this without the confirms which rely on that detection method for now? The PR is already getting pretty large to the point where we have discussed splitting it. I feel like we're getting very close to a point where this will be the only thing not yet implemented. Maybe instead of splitting it, we merge what's here so far after all feedback/review is complete? I'm down to work on implementing it whichever way we go.

myk002:
I'm perfectly happy with merging the confirm protections that work and filling in the rest later. There is a lot that is very valuable in this PR already.

robob27:
Explaining a bit what happened with pause in some of those commits:

At one point I moved unpause detection out of feed and into render to try to make it independent of input. I realized after some testing that right click and esc couldn't be used as a reliable way to detect if the window was closed. I personally use `Q` to show/hide my squads window for example (right click doesn't even work, and escape opens up the save/exit menu). And it's also possible to go from one screen to another without ever pressing esc or right click by just pressing another hotkey. I'm also unable to check that the focus no longer matches in feed since the input hasn't been processed yet.

Then I realized that it's probably okay if the confirmation doesn't unpause _as soon as_ the focus is dismissed. So I moved it back to feed, but without trying to detect any specific input. So now it works like:

- Press Q to open squads
- Attempt to delete, pause confirmation
- Press Q to close squads
- The squads window is closed and the confirmation is still paused
- Press Q to open squads - the window is not yet open, so the confirmation gets unpaused before the window is opened

So I think this works out this way..

robob27:
I think the only outstanding stuff is:

- some ZScreen stuff you mentioned (not sure if you want me to take a look at that as part of this)
- the fate of `getSelectedStockpile` etc (I made another comment on that thread)
- any potential feedback from all of the changes yesterday

Getting so close 😄

robob27:
Sooo it was really bugging me that I couldn't get the uniform deletion confirm working without the overlay stuff so I decided to take a stab at it. It works a lot better than I expected, though it does seem to have one issue. If I resize the window in a certain way, it stops being positioned at the right place.

https://user-images.githubusercontent.com/2104555/216687165-7ded15e7-5f4f-44e3-9c94-34417fe931ba.mp4

it's also possible to resize the uniform the "AssignUniform" window right off the screen. I guess probably I'd have to find some minimum window size and don't render the overlay if the window is less than that size?

For making sure it works at different zoom levels, do you mean this zoom level? 
image

Or ctrl+scroll zoom level? Or both? Both of those seem to work minus the issue above.

library/Core.cpp
myk002: I can work on that after this PR is merged. That functionality is tied to how hotkeys will work when ZScreens of different properties are onscreen.

library/modules/Gui.cpp
myk002: `Gui::getCurFocus(true)`, I think. Focus strings for the top non-dismissed viewscreen

myk002:
> it stops being positioned at the right place

its frame looks like it is being forced to change based on screen positioning. You could have a check after layout is updated to force the frame back into position. IOW, it looks fixable

robob27:
>its frame looks like it is being forced to change based on screen positioning. You could have a check after layout is updated to force the frame back into position. IOW, it looks fixable

One other challenge here.. when you have 10+ uniforms, a scrollbar renders and pushes the delete buttons over to the left. If I could find a list of the uniforms, or something telling me if the scrollbar was rendered, I could adjust it based on that, but so far I can't find any uniform data (other than uniforms assigned to individuals).

plugins/confirm.cpp
```diff
@@ -46,10 +42,8 @@ static map<string, conf_wrapper*> confirmations;
 string active_id;
 queue<string> cmds;
 
-// true when confirm is paused
-bool paused = false;
-// if set, confirm will unpause when this screen is no longer on the stack
-df::viewscreen *paused_screen = NULL;
+// TODO: move to confirmation instance with paused_focus etc
+ df::viewscreen *paused_screen = NULL;
```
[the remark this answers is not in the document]
robob27: I _think_ this is no longer needed since we track if each confirmation is paused individually and they will each unpause during render if the focus string is no longer found. I commented this during my first pass at refactoring pause to use focus strings before implementing the per-confirmation pause stuff. So I've gone ahead and removed it for now.

robob27:
Okay, I think it's _actually_ done for now, minus any new review comments or me finding something broken while I play with it more 😌

plugins/hotkeys.cpp
```diff
@@ -38,14 +37,13 @@ static bool can_invoke(const string &cmdline, df::viewscreen *screen) {
 }
 
 static int cleanupHotkeys(lua_State *) {
-    DEBUG(log).print("cleaning up old stub keybindings for %s\n", current_focus.c_str());
+    DEBUG(log).print("cleaning up old stub keybindings for:\n %s\n", join_strings(", ", Gui::getCurFocus(true)).c_str());
```
myk002, line 40: can we keep this debug log line internal-newline-free?

plugins/confirm.cpp
```diff
@@ -501,45 +548,59 @@ struct cls##_hooks : cls::screen_type { \
...
     IDs (used in the "confirm enable/disable" command, by Lua, and in the docs)
     are obtained by replacing '_' with '-' in the first argument to DEFINE_CONFIRMATION
+
+    The second argument to DEFINE_CONFIRMATION determines the viewscreen that any
+    intercepted input will be fed to.
+
+    The third argument to DEFINE_CONFIRMATION determines the focus string that will
+    be used to determine if the confirmation should be unpaused. If a confirmation is paused
+    and the focus string is no longer found in the current focus, the confirmation will be
+    unpaused. Focus strings ending in "*" will use prefix matching e.g. "dwarfmode/Info*" would
```
myk002, line 578: Is this true? I thought we were matching prefixes unconditionally. Where is the `*` interpreted?

plugins/confirm.cpp
```diff
@@ -306,45 +293,70 @@ class confirmation : public confirmation_base {
...
-            if (!paused_screen && input->count(df::interface_key::LEAVESCREEN))
-                conf_lua::api::unpause(NULL);
-            return false;
+        bool mouseExit = false;
+        if(df::global::enabler->mouse_rbut) {
+            mouseExit = true;
         }
-        else if (state == INACTIVE)
+        bool mouseSelect = false;
+        if(df::global::enabler->mouse_lbut) {
```
myk002, line 301: should these be `...but` or `...but_down`? If the mouse is clicked, held, and then moved to the button, does it count?

myk002 (commented):
Please also look over the docs in `docs/plugins/confirm.rst` to see if it needs any updates.

This PR will need a sister PR on the scripts branch that addresses:
- rename gui/confirm-opts.lua to gui/confirm.lua (and the docs too)
- update the GUI as appropriate (including making it a ZScreen) (see gui/autochop.lua for an example)
- add confirm to the "System Services" list in gui/control-panel.lua

plugins/confirm.cpp
robob27: I made `matchFocusString` also take a boolean arg which is `true` by default (defined in the header). If false, it will look for an exact (still case insensitive) match. If "*" is at the end of the focus string in the `DEFINE_CONFIRMATION`, it will set `match_prefix` to true on the confirmation instance, otherwise it'll be false. `match_prefix` is then passed to `matchFocusString` [here](https://github.com/DFHack/dfhack/pull/2721/files#diff-e005088f61392cf1a500dac515066f9cfcb18894fdb97549ead1631aa39668b8R308)

plugins/hotkeys.cpp
robob27: Oops, forgot to get this newline when I changed the separator earlier, will fix

robob27:
>Please also look over the docs in docs/plugins/confirm.rst to see if it needs any updates.

:+1:

>This PR will need a sister PR on the scripts branch that addresses:
>
>  rename gui/confirm-opts.lua to gui/confirm.lua (and the docs too)

I assumed it was named confirm-opts.lua to make it easier to see which `confirm.lua` you're opening if you CTRL+P. Still worth changing? I did actually make a few other changes to confirm-opts.lua locally. I actually kinda forgot it wasn't part of this repo. Will be sure to push that up.

>update the GUI as appropriate (including making it a ZScreen) (see gui/autochop.lua for an example)

You just mean the confirmation options interface and not the confirmation dialog yeah? :+1:

>add confirm to the "System Services" list in gui/control-panel.lua

Didn't even know about this, very cool! :+1:

plugins/confirm.cpp
robob27: As soon as the mouse buttons are pushed down it counts, even if still holding the button. But it doesn't spam the events, so for example if you open trade, select some items and hold right click, the confirmation dialog appears and is not dismissed unless you let go of right click and click again.

myk002:
> > update the GUI as appropriate (including making it a ZScreen) (see gui/autochop.lua for an example)
> 
> You just mean the confirmation options interface and not the confirmation dialog yeah? 👍

right, just the gui/confirm script. the confirm dialogs here *could* be ZScreens, but they don't have to be since they're modal anyway. what you have is fine.

> > add confirm to the "System Services" list in gui/control-panel.lua
> 
> Didn't even know about this, very cool! 👍

it was added literally just a few days ago. eventually I'll have a query API that allows gui/control-panel to populate the fort and system services list automatically. right now I can get the list of enableable tools, but that's not specific enough for gui/control-panel.

robob27:
@myk002 Added the other PR, hopefully I did that right. Not super familiar with submodule stuff, should probably fix that

myk002:
> > > update the GUI as appropriate (including making it a ZScreen) (see gui/autochop.lua for an example)
> > 
> > You just mean the confirmation options interface and not the confirmation dialog yeah? 👍
> 
> right, just the gui/confirm script. the confirm dialogs here _could_ be ZScreens, but they don't have to be since they're modal anyway. what you have is fine.

Now that I'm playtesting this, I'm seeing that the dialogs aren't rendered with the standard DFHack themeing. This can be fixed in a new PR (this PR is far too large as it is), but the screens should be drawn in Lua with the standard frames and widgets.

myk002 (approved):

docs/plugins/confirm.rst
```diff
@@ -5,7 +5,7 @@ confirm
     :summary: Adds confirmation dialogs for destructive actions.
     :tags: untested fort interface
 
-Now you can get the chance to avoid seizing goods from traders or deleting a
+Now you can get the chance disbanding a squad or deleting a
```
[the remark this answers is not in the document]
robob27: Far too tired last night lol

robob27:
Oh noooo it got merged in unsquashed. My sloppiness forever memorialized 😭 haha

myk002:
> Oh noooo it got merged in unsquashed. My sloppiness forever memorialized 😭 haha

Oh yeah. Oh well. I wouldn't worry about it : ) I tried to keep our commit history super clean for a while but it just caused pain. This is fine.

lethosor:
This seems to have broken a number of clients of getCurFocus, e.g. exportlegends: #2936 

Did we ever settle on a migration path?

PR #2731, opened 2023-01-27
getplants fixes

### LINKED ISSUE #1805: Make getplants plugin case-insensitive
The "getplants" command requires the plant types to be inputted in all capital letters. It would be more convenient to allow them to be inputted with any capitalization, as with the "exterminate" script.

### LINKED ISSUE #2043: `getplants` will not designate trees
Selecting any (or all) tree types using getplants always returns, "Updated 0 plant designations."  No trees are designated for cutting.

gearsix:
Hey, just some small first-timer fixing issues I found for the *getplants plugin*.

Fixes #1805 
ID parameters are now added to `plantNames` with a `toUppercase` call, so case doesn't matter for input.

Fixes #2043 
The `designate` call was returning *false* for all trees because the `ripe` check was failing.
My guess is that the trees either have a different lifetime or aren't ripe but it doesn't matter because they can still be chopped. I'm wasn't sure what the best way to fix this one was, so I added an `istree` check for the tile and now the check passes if `istree || isripe`.

I did some brief testing and it seems good.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -37,7 +37,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 -@ Fix issues with clicks "passing through" some DFHack window elements, like scrollbars
-- `plugin/getplants`: tree are now designated correctly
+- `getplants`: tree are now designated correctly
 
 ## Misc Improvements
 - A new cross-compile build script was added for building the Windows files from a Linux Docker builder (see the Compile instructions in the docs)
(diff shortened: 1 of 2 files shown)


myk002 (commented):
looks good! Thank you for this! Are there any updates that need to be made to the documentation in `docs/plugins/getplants.rst`?

Also, could you add these fixes/improvements to `docs/changelog.txt`?

gearsix:
> looks good! Thank you for this! Are there any updates that need to be made to the documentation in `docs/plugins/getplants.rst`?

I don't believe so, no big changes have been made to it's behaviour.

> Also, could you add these fixes/improvements to `docs/changelog.txt`?

Updated the changelog & merged develop :)

docs/changelog.txt
```diff
@@ -37,6 +37,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Fixes
 -@ Fix issues with clicks "passing through" some DFHack window elements, like scrollbars
+- `plugin/getplants`: tree are now designated correctly
```
myk002, line 40:
```suggestion
- `getplants`: tree are now designated correctly
```

docs/changelog.txt
```diff
@@ -46,6 +47,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `gui/quickcmd`: now has it's own global keybinding for your convenience: Ctrl-Shift-A
 - Many DFHack windows can now be unfocused by clicking somewhere not over the tool window. This has the same effect as pinning previously did, but without the extra clicking.
 - `overlay`: overlay widgets can now specify a default enabled state if they are not already set in the player's overlay config file
+- `plugin/getplants`: ID values will now be accepted regardless of case
```
myk002, line 50:
```suggestion
- `getplants`: ID values will now be accepted regardless of case
```

myk002 (approved):

PR #2733, opened 2023-01-27
Add show-unit-syndromes to removed scripts

vallode:
As per: https://github.com/DFHack/scripts/pull/562#issuecomment-1405838879


docs/about/Removed.rst
```diff
@@ -151,6 +151,12 @@ ruby
 Support for the Ruby language in DFHack scripts was removed due to the issues
 the Ruby library causes when used as an embedded language.
 
+.. _show_unit_syndromes:
+
+show_unit_syndromes
+====
```
myk002, line 157:
```suggestion
===================
```

PR #2736, opened 2023-01-28
cross-compile-script: Fix permissions in ccache dir for sudo users

kelvie:
I think this affects basically nobody but me since everyone else seems to run docker rootless.


Gaelmare:
I had this problem too, just too lazy to fix. I added ccache explicitly to the chown instead, but this works.

lethosor:
I suspect Docker still runs as root for most people - it's not truly "rootless", just provides an easy way to configure the CLI to not require root.

lethosor (approved):

Issue #2652, opened 2023-01-18
feature request: autoslab


ab9rf:
some sort of tool to automate or at least make easier the process of engraving memorial slabs. the current selection menu is unsorted and is filled with units for which engraving is not required, so the ability to filter or search that list would be deemed useful, as would automation that automatically engraves (and possibly places) memorial slabs

johncosker:
I'll take a stab at this one

PR #2737, opened 2023-01-28
Implement autoslab engraving feature

johncosker:
Fixes https://github.com/DFHack/dfhack/issues/2652

Adds plugin `autoslab` that will automatically queue up engrave slab orders for ghosts.

Will check two conditions before creating an EngraveSlab job:

1. That there are no existing engraved slab items for that unit.
2. There are no existing EngraveSlab orders queued for the dwarf in work orders.

The check for existing work orders uses the historical figure id for the job which is fine, but for the existing engraved slab item I wasn't able to figure out a way to do it other than checking for the dwarf's name in the engraved slab description. It seems to work well enough in my testing but ideally we'd have a more solid way to do it.

Added some notes for possible enhancements that I might look into if this ends up merged.

Plugin code based on the example in the repo.


docs/plugins/autoslab.rst
```diff
@@ -0,0 +1,20 @@
+autoslab
+========
+
+.. dfhack-tool::
+    :summary: Automatically engrave slabs for ghostly citizens!
+    :tags: untested fort auto workorders
```
myk002, line 6: I presume you've tested this, so no need to add the `untested` tag : )

docs/plugins/autoslab.rst
```diff
@@ -0,0 +1,20 @@
+autoslab
+========
+
+.. dfhack-tool::
+    :summary: Automatically engrave slabs for ghostly citizens!
```
myk002, line 5: summaries must actually end in a period to conform with the in-game help text parser

docs/plugins/autoslab.rst
```diff
@@ -0,0 +1,20 @@
...
+    :tags: untested fort auto workorders
+    :no-command:
+
+Automatically queue orders to engrave slabs of existing ghosts. Will only queue
+an order if there is no existing slab with that unit's memorial engraved and
+there is not already an existing work order to engrave a slab for that unit
```
myk002, line 11: Maybe note that the player is still responsible for both crafting the base unengraved slab and also for placing the engraved slabs as a building in-game.

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
+/* Simple plugin to check for ghosts and automatically queue jobs to engrave slabs for them.
+ *
+ * Enhancement idea: Queue up a ConstructSlab job, then link the engrave slab job to it. Avoids need to have slabs in stockpiles
+ *                   Would require argument parsing, specifying materials
+ * Enhancement idea: Automatically place the slab. This seems like a tricky problem but maybe solveable with named zones?
+ *                   Might be made obsolete by people just using buildingplan to pre-place plans for slab?
```
myk002, line 6: I'm not sure about this -- in its current form, buildingplan doesn't differentiate between engraved and unengraved slabs, so it would likely just steal the unengraved slabs before dwarves have a chance to engrave them. We could certainly modify buildingplan to only choose engraved slabs (or make that an option).

docs/plugins/autoslab.rst
myk002: if they're not already there, we might want to add slabs to the default library manager orders in the `orders` library. they might be in `rockstock`, and we might move them to `basic`?

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
...
+ * Enhancement idea: Optionally enable autoengraving for pets.
+ * Enhancement idea: Try to get ahead of ghosts by autoengraving for dead dwarves with no remains.
```
myk002, line 8: or unreachable corpses (e.g. at the bottom of a pool or in a closed-off cavern -- see autochop.cpp for how I detect reachability for logs)

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
...
+DFhackCExport command_result plugin_load_data(color_ostream &out)
+{
+    config = World::GetPersistentData(CONFIG_KEY);
+
+    if (!config.isValid())
+    {
+        DEBUG(status, out).print("no config found in this save; initializing\n");
+        config = World::AddPersistentData(CONFIG_KEY);
+        set_config_bool(CONFIG_IS_ENABLED, is_enabled);
+        set_config_val(CONFIG_CYCLE_TICKS, 1200);
```
myk002, line 126: it would be fine to take this out and just use a constant value (e.g. see autochop.cpp). I've been using constants for the cycle period for most new/updated plugins and should probably take this example out of the sample skeleton code.

johncosker:
Looks like I should set up a Linux build VM

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector<PluginCommand> &commands)
+{
+    DEBUG(status, out).print("initializing %s\n", plugin_name);
+
+    // provide a configuration interface for the plugin
+    commands.push_back(PluginCommand(
+        plugin_name,
+        "Automatically engrave slabs of ghostly citizens!",
+        do_command));
```
myk002, line 84: since you have no configuration you can take this out. you've already marked the tool as `:no-command:` in the docs.

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
...
+    if (!Core::getInstance().isWorldLoaded())
+    {
+        out.printerr("Cannot run %s without a loaded world.\n", plugin_name);
+        return CR_FAILURE;
+    }
+
+    // TODO: decide what, if any configuration should be here
+
+    return CR_OK;
+}
```
myk002, line 172: you can remove this until the time when this plugin has a config interface

plugins/autoslab.cpp
```diff
@@ -0,0 +1,267 @@
...
+                 std::back_inserter(ghosts),
+                 [](const auto &unit)
+                 { return unit->flags3.bits.ghostly; });
+
+    for (auto ghost : ghosts)
+    {
+        // Only create a job is the map has no existing jobs for that historical figure or no existing engraved slabs
+        if (histToJob.count(ghost->hist_figure_id) == 0 && !engravedSlabItemExists(ghost, engravedSlabs))
+        {
+            createSlabJob(ghost);
```
myk002, line 254: if there exist no free unengraved slabs, you might also want to create a job for an unengraved slab to use for the engrave job.

plugins/autoslab.cpp
johncosker: It might be better to update buildingplan to allow specifying engraved slabs, and let this plugin just handle making the engravings. Better to leave the job of placing buildings to the plugin that already does it well. You've got a better sense of where things belong though so I'll defer to whatever you think is best.

plugins/autoslab.cpp
johncosker: Good idea, I'll append the enhancement comment.

johncosker:
I was able to reproduce the linux build failures in the WSL subsystem on Windows so hopefully that also fixes the issues in the CI checks.

For the record I haven't _tested_ the changes on linux, if required I can do that I'll just need to get a setup going.

docs/plugins/autoslab.rst
myk002: > For the record I haven't _tested_ the changes on linux

DF only has a windows build, so testing on Linux (via wine/proton) isn't any better than testing on Windows. In general, though, DF does not have behavior differences on different platforms, and we don't ask contributors to test on anything other than what they normally run.

plugins/autoslab.cpp
myk002: That sounds like a reasonable plan. Could you file a feature request for `buildingplan` for that? Once I get `gui/control-panel` written, `buildingplan` is my next target.

docs/plugins/autoslab.rst
```diff
@@ -0,0 +1,20 @@
...
+========
+
+.. dfhack-tool::
+    :summary: Automatically engrave slabs for ghostly citizens.
+    :tags: fort auto workorders
+    :no-command:
+
+Automatically queue orders to engrave slabs of existing ghosts. Will only queue
+an order if there is no existing slab with that unit's memorial engraved and
+there is not already an existing work order to engrave a slab for that unit
```
myk002, line 11:
```suggestion
there is not already an existing work order to engrave a slab for that unit.
Make sure you have spare slabs on hand for engraving! If you run
`orders import library/rockstock <orders>`, you'll be sure to always have
some slabs in stock. 
```

plugins/autoslab.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // for logging during the periodic scan
+    DBG_DECLARE(autoslab, cycle, DebugCategory::LINFO);
+}
+
+static const auto CONFIG_KEY = std::string(plugin_name) + "/config";
+static PersistentDataItem config;
+enum ConfigValues
+{
+    CONFIG_IS_ENABLED = 0,
+    CONFIG_CYCLE_TICKS = 1,
```
myk002, line 48: can remove

plugins/autoslab.cpp
johncosker: Created https://github.com/DFHack/dfhack/issues/2754

docs/plugins/autoslab.rst
johncosker: Fixed

plugins/autoslab.cpp
johncosker: Fixed

myk002 (approved):

PR #2743, opened 2023-01-28
Fix dig plugin and enable building it

johncosker:
Found that the only compiler errors were cause by an (apparently unused) header include and a single other line that apparently just needed to be updated because some structures change.

I tested all the types of dig* commands and it seems to work fine. I also confirmed that the line of code changed is correctly setting the default priority to 4000.

I'd like to try creating a GUI for dig to allow mouse based circle/ellipses/etc... drawing and also adding a WxH label and maybe even X/Y guidelines to the designation to make creating symmetrical designations easier, but this is a good first step.

### DIFF
--- a/plugins/dig.cpp
+++ b/plugins/dig.cpp
@@ -9,7 +9,6 @@
 #include "Console.h"
 #include "Export.h"
 #include "PluginManager.h"
-#include "uicommon.h"
 
 #include "modules/Gui.h"
 #include "modules/MapCache.h"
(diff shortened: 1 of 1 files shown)


myk002:
A GUI for dig would be much appreciated. I have some long-term thoughts about a general "gui/designer" interface that includes this, and a `dig`-specific UI could be a step towards that goal. Or it might be better to have a separate `gui/dig` indefinitely. It's wide open at this point, and I'd love to see what you come up with.

myk002:
btw you can copy the structure of `gui/quickfort` to kickstart your `gui/dig`

johncosker:
I was literally just about to ask what the most up-to-date example for gui stuff would be, thanks!

myk002 (approved):

myk002:
marked as tested at <https://docs.google.com/spreadsheets/d/1hiDlo8M_bB_1jE-5HRs2RrrA_VZ4cRu9VXaTctX_nwk/edit#gid=1774645373> (no need to update the tag in `docs/plugins/dig.rst` -- I periodically sync the tags in the spreadsheet to the docs)

PR #2745, opened 2023-01-28
Prevent DF screens from getting stuck and corrupting the stack on transitions

### LINKED ISSUE #2739: Multiple Embarks bug
Quoting from Discord:

damn. i have no idea how i did this, idk if it was a DF bug or a dfhack bug, but uh.. i just embarked twice. at the same time. and now i have all 14 dwarfs from both embarks, lol
here's what happened:

1) i embarked like normal
2) i set up my embark with prepare carefully
3) when i hit embark, it came up with a box in the middle of the screen with tons of garbled text, and the "Confirm/Abort" box in the top left like when you are choosing an embark location on the map
4) I hit Confirm in that little box
5) it brought me back to the embark carefully screen, where i embarked again
6) the error box came up again, but this time i just right clicked and tada, now im on my embarked tile with 14 dwarfs and 2x the materials lmao

(unpausing caused all trees to turn to clay, several animals and dwarves to be crushed to death, and a bunch of "newly appointed noble" popups, despite no nobles being appointed.)

---------------------------

User SeerSkye pointed out that a similar bug occurs if you open up the dfhack console (with `) while on the prepare embark screen, and hit "I'm ready!" a bunch of times. This caused all the same problems except for the screen of garbled text. I was unable to replicate that, nor did I get a screenie, sorry.

Edit: Oh, one more thing, if I try to quit back to title after causing this bug, DF crashes.

Edit 2: Sorry, basic info: Win7 64bit, DFHack 50.05-alpha2, DF 50.05 Steam.

### LINKED ISSUE #2741: DF screens that are under ZScreens don't go away when dismissed
E.g. when loading a game, the loading screen can't disappear and the fort screen can't appear until all DFHack tool windows are dismissed.

I believe we can fix this by scanning the viewscreen stack and popping the dismissed screens ourselves instead of waiting for DF to do it

myk002:
Fixes #2741
Fixes #2739

There are two parts to this problem.
1. Dismissed DF viewscreens that are not at the top of the stack don't get cleaned up and can appear to hang after they are dismissed.
2. Some viewscreen transitions destroy the top viewscreen, which is sometimes a DFHack viewscreen and not the intended DF viewscreen.

(1) is fairly easy to fix -- when a DFHack viewscreen is on top, check the underlying DF viewscreen, and if it is dismissed, raise it to the top so it can be cleaned up.

(2) only happens for a few DF viewscreens. Instead of dismissing themselves and pushing a viewscreen on the stack on a screen transition, they instead *destroy* the top viewscreen unconditionally. When a DFHack tool is visible, this means the DFHack viewscreen is destroyed and the DF viewscreen is left hanging.

There doesn't seem to be a good way to detect when the viewscreen is *about* to be destroyed (is there?) so the solution I found was to force-pause those screens (that is, don't call their `logic()` functions) while DFHack screens are up. The "force pause" indicator is displayed on the DFHack window to show what is happening. The DF viewscreen will continue when the DFHack screen is closed. This prevents the transition from occurring when a DFHack screen is up, at the cost of preventing the DF screen from making any progress whatsoever while a DFHack screen is up.


library/lua/gui.lua
```diff
@@ -706,27 +706,41 @@ ZScreen.ATTRS{
...
         df.global.pause_state = df.global.pause_state and self.saved_pause_state
     end
 end
 
+local NO_LOGIC_SCREENS = utils.invert{
+    'viewscreen_loadgamest',
+    'viewscreen_export_regionst',
+    'viewscreen_choose_game_typest',
+    'viewscreen_worldst',
+}
```
lethosor, line 728: I would tend to prefer to verify that these names are correct by looking up the classes, e.g. with
```lua
for k, v in pairs(NO_LOGIC_SCREENS) do
    NO_LOGIC_SCREENS[k] = df[v] or error('invalid class name: ' .. v)
end
```

and then replacing `vs_name = getmetatable(...)` later with `vs_type = getDFViewscreen()._type`

lethosor (commented):
Just for my own sake, is the issue that DF is _deleting_ the topmost viewscreen in some cases or just dismissing it (by setting breakdown_level)? The latter should be straightforward to detect and prevent, but the former is probably not feasible, yeah.

myk002:
> Just for my own sake, is the issue that DF is _deleting_ the topmost viewscreen in some cases or just dismissing it (by setting breakdown_level)?

Deleting. I discovered this when our viewscreen's onDestroy function was called and onDismiss was not.

lethosor:
Interesting. That seems really inconsistent with how DF has ever behaved in the past, so I'm surprised.

Worth noting that when DF dismisses a screen, it does _not_ call our `onDismiss()`, ever, so that's not necessarily an indication that it isn't dismissing the screen. onDismiss() is only a method of dfhack_viewscreen and its subclasses, so DF doesn't know to call it.

myk002:
let me test that a bit more, then. it could be that it's setting our breakdown_level against our will.

myk002:
no, I forgot, I already checked this. our breakdown_level == interface_breakdown_types::NONE at the time our destructor is run

lethosor:
Sounds like we should bug Bay12 :)

(kidding, but I guess what we thought was an invariant is no longer)

PR #2752, opened 2023-01-30
Add API for getting a list of citizens

myk002:
for use in DFHack/scripts#575 and other places where we need to calculate reachability by a citizen


myk002:
merging so `unforbid` can use

PR #2755, opened 2023-01-30
remove autohauler

ab9rf:
`autohauler` just doesn't make sense in v50: for `autohauler` to work at all, it would have to disable work details entirely, but if it did this then there'd be nothing to manage the _other_ labors.

there might be a point to something _like_ `autohauler` in the future, but any such tool should probably be reimplemented _ab initio_ to interact with the work detail system


myk002 (changes requested):
also remember to remove the docs in docs/plugins/autohauler.rst and add it to docs/about/Removed.rst

myk002 (approved):

PR #2757, opened 2023-01-30
reenable autolabor

ab9rf:
minimum required to make work:
* realign the job table
* add code to flip/flop the work detail enable flag


ab9rf:
This test version has multiple caveats:
* there is no particular guarantee that the labor assignments autolabor is making are good
* there is no way to easily see what labors are being assigned to whom, because there is no in-game mechanic for seeing the labor assignment table as Toady hid all of this away from the player. Until `manipulator` is ported to v50, probably the best way to see what autolabor is doing is to use Dwarf Therapist.
* autolabor disables the work detail system when it is enabled using the mechanic provided by Toady, but there is _no_ visual indication that this has happened. The work detail screen will pop up if the player opens it and the player can manipulate the settings there to their heart's content all without having any actual effect on the game.
* disabling autolabor reenables the work detail system by unsetting the flag provided by Toady, but work details are only applied when the player actually opens the work detail screen and does some de minimus thing. So whatever labor settings were in place when autolabor is disabled will remain in place until the player actually opens the work detail screen (and possibly touches something on that screen to force it to recompute). This is a deficiency in the core game design that we cannot readily work around, at least not without doing code detours or calling a nonvirtualized function that would have to be located by reverse engineering

myk002 (commented):
code looks reasonable, but I'm wary about merging this without some stronger UI support that tells players what's going on. Everyone has heard of `autolabor`, but there will be confusion unless we clearly label the work details screen as non-functional while `autolabor` is active.

I think we might need:
1) updates to docs/plugins/autolabor.rst that explains the caveats
2) a simple overlay that appears when the work details screen is shown and autolabor is enabled that says that the user can make changes, but they won't have any effect on the game until `autolabor` is disabled. A pared-down version of https://github.com/DFHack/dfhack/blob/develop/plugins/lua/orders.lua would be fine

myk002:
and even then, perhaps we should keep it as "untested" until some reports come in saying how well it works

myk002:
Because I like you, I implemented (2). WDYT?

docs/plugins/autolabor.rst
```diff
@@ -18,8 +18,31 @@ untouched by autolabor.
...
+    At this time there is no way to easily see what labors are being assigned to whom, as there is no longer
+    any vanilla means for seeing the labor assignment table. Until `manipulator` is once again available,
+    probably the best way to see what autolabor is doing is to use
+    `Dwarf Therapist <https://github.com/Dwarf-Therapist/Dwarf-Therapist>`_. You can also increase autolabor's
+    logging level using the `debugfilter<debug>` command (setting either ``debug`` or ``trace`` level for the
+    ``cycle`` mode) but be warned that this may generate a large amount of console spam, especially in a large fort.
+
+    When it is enabled, autolabor automatically disables the work detail system. You cannot
+    use autolabor and work details at the same time. If you attempt to open the work detail screen while
+    autohauler is active, a warning box should appear advising you that autolabor is managing labors and preventing
```
myk002, line 37:
```suggestion
    autolabor is active, a warning box should appear advising you that autolabor is managing labors and preventing
```

myk002 (approved):

PR #2759, opened 2023-01-30
Always build stonesense

CoreyJ87:
Stonesense was not being built by github actions for steam v50 builds. Thus not matching release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
           -DBUILD_DEV_PLUGINS:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SIZECHECK:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SKELETON:BOOL=${{ matrix.plugins == 'all' }} \
-          -DBUILD_STONESENSE:BOOL=1 \
+          -DBUILD_STONESENSE:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SUPPORTED:BOOL=1 \
           -DCMAKE_C_COMPILER_LAUNCHER=ccache \
           -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
@@ -145,7 +145,7 @@ jobs:
       - name: Cross-compile win64 artifacts
         run: |
           cd build
-          bash -x build-win64-from-linux.sh
+          env EXTRA_CMAKE_ARGS="-DBUILD_STONESENSE:BOOL=1" bash -x build-win64-from-linux.sh
       - name: Format artifact name
         id: artifactname
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
           -DBUILD_DEV_PLUGINS:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SIZECHECK:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SKELETON:BOOL=${{ matrix.plugins == 'all' }} \
-          -DBUILD_STONESENSE:BOOL=${{ matrix.plugins == 'all' }} \
+          -DBUILD_STONESENSE:BOOL=1 \
           -DBUILD_SUPPORTED:BOOL=1 \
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -87,7 +87,7 @@ jobs:
           -DBUILD_DEV_PLUGINS:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SIZECHECK:BOOL=${{ matrix.plugins == 'all' }} \
           -DBUILD_SKELETON:BOOL=${{ matrix.plugins == 'all' }} \
-          -DBUILD_STONESENSE:BOOL=${{ matrix.plugins == 'all' }} \
+          -DBUILD_STONESENSE:BOOL=1 \
```
lethosor, line 90: What I meant was add `-DBUILD_STONESENSE:BOOL=1 \` to the _Windows_ section, which starts around line 128 right now.

We want to keep this line how it was before because we decided we don't need to build Stonesense for every permutation of our Linux builds - just checking that it compiles once on Linux is fine.

lethosor (changes requested):

.github/workflows/build.yml
myk002, line 90: this is the for the linux build actions. You'd need to pass `-DBUILD_STONESENSE:BOOL=1` through `build-win64-from-linux.sh` on line 148

myk002: which may take some editing of `build-win64-from-linux.sh`

CoreyJ87: ohhhhh i see. Thanks

CoreyJ87: That makes sense. thats why i was asking why that line exists.

lethosor: yeah, it looks like the build script isn't flexible enough to allow extra build args at the moment. Sorry I didn't check it before sending you down this path :/

CoreyJ87: No problem at all. Im happy to learn alil anyways.

kelvie: you can either add this directly to the script, right after `-DBUILD_DOCS=1` (most people would want to build stonesense I think?) or add `$EXTRA_CMAKE_ARGS` there, and pass it in line 148 below, e.g.:

```
 env EXTRA_CMAKE_ARGS="-DBUILD_STONESENSE:BOOL=1" bash -x build-win64-from-linux.sh
```

Though will the `EXTRA_CMAKE_ARGS` version we may need to be careful about quoting if we're gonna do multiple `-D`s in the future.

CoreyJ87: Yeah @myk002 recommended the idea of using "$@" so we could just take in as many as we want in the future.

kelvie: The problem is that it needs to be double quoted, as it's being passed into `bash -c`. Or refactored so that it's no longer necessary (we need to run multiple commands inside the docker container).

kelvie: Yeah just testing it, you can just put in $@ (without quotes, since it's already quoted), and it should be fine, but I *think* it can't handle arguments with spaces in them (which should be OK, since I feel like 90% of the shell scripts out there don't get quoting right)

lethosor: I'm not too concerned about the `EXTRA_CMAKE_ARGS` approach. The quoting problem I'm expecting, which would happen both with `$@` and an environment variable like `EXTRA_CMAKE_ARGS`, is preserving spaces within individual arguments, which we should pretty much never need to do (e.g. there is no space in `-DBUILD_STONESENSE:BOOL=1`). `EXTRA_CMAKE_ARGS` would support multiple arguments separated by spaces without a problem.

myk002:
Do you mind if I push this along a little bit? We may need this in the build for the release later today

myk002:
well that didn't work. I'll look into it a bit with a local build

kelvie:
https://github.com/DFHack/dfhack/pull/2788 I had this locally for a while now

lethosor:
For future reference, GitHub Actions has a more canonical way of setting environment variables with `env:`: https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

PR #2763, opened 2023-01-30
Update hotkeys and quickstart guide for gui/control-panel

myk002:
Depends on DFHack/scripts#579

should we give the control panel an alternate hotkey for keyboards without a usable backtick? it's not as important as `gui/launcher`, so I'm not sure it's warranted.


myk002:
merging for feedback

PR #2807, opened 2023-02-05
Update Units::isFortControlled

Bumber64:
Add check for `agitated_wilderness_creature`.


myk002 (commented):
could you add a changelog entry?

myk002 (approved):

PR #2810, opened 2023-02-05
de-hardcoding scrollbar spritesheet definitions

TaxiService:
added a new variable "SBSO", and redefined all scrollbar tile definitions as this variable + an offset. this could make it easier to switch scrollbar graphics from within the game, if necessary.


myk002 (approved):
I'm going to trust your math here : p

PR #2811, opened 2023-02-05
increase min scrollbar height to 2 (from 1)

TaxiService:
rationale:
- vanilla scrollbars dont get shorter than 2.
- 2-tall 'bars are easier to click on than 1-tall 'bars.
- this avoids having to make short custom graphics for 1-tall tiles, which most of the time look terrible...

cons:
- short scrollbars are cool : (

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -808,7 +808,7 @@ local function scrollbar_get_max_pos_and_height(scrollbar)
     local frame_body = scrollbar.frame_body
     local scrollbar_body_height = (frame_body and frame_body.height or 3) - 2
 
-    local height = math.max(1, math.floor(
+    local height = math.max(2, math.floor(
         (math.min(scrollbar.elems_per_page, scrollbar.num_elems) * scrollbar_body_height) /
         scrollbar.num_elems))
 


library/lua/gui/widgets.lua
```diff
@@ -808,7 +808,7 @@ local function scrollbar_get_max_pos_and_height(scrollbar)
     local frame_body = scrollbar.frame_body
     local scrollbar_body_height = (frame_body and frame_body.height or 3) - 2
 
-    local height = math.max(1, math.floor(
+    local height = math.max(2, math.floor(
```
myk002, line 811: I'm not sure if this has any practical effect on the scrollbar. It's still beholden to the parent frame. What difference does this change actually make visually?

TaxiService: it turns this...
Capture1

into this...
Capture3

and if you use a custom texture, the scrollbar will always look good
Capture2

like, if we wanted 1-tall scrollbar graphics to "look good", we could probably define about 6 more sprites for different states... like so:
(from left to right: 2-tall, current 1-tall, "good-looking" 1-tall, possible implementations of different states of it)
(those red-highlit parts are to show that the goodlooking 1tall overflows into the lower tiles)
Captur4e

...but! maybe that's just too much of a useless detail to be concerned about. Like, limiting scrollbars' height to 2 lets us *not* implement that crap, *and* keep the ui looking clean, consistent.


worthy of note though, i just noticed a bit of a weird behaviour. When the 'bar visually reaches the bottom of the list, you can still drag it downwards 1 more tile and scroll some more.
[2023-02-06 12-54-06-1.webm](https://user-images.githubusercontent.com/5853667/216968171-006fabc6-5aac-4601-bf99-dbd44515a5d8.webm)

TaxiService: >I'm not sure if this has any practical effect on the scrollbar. It's still beholden to the parent frame.

come to think of it, i guess that weird behaviour i described is surely related to what you just said.
I... I, like, honestly just tried changing numbers here and there, and _that one_ i changed _seemed_ to do what i wanted... 😂 so i felt like a supa haka and was all like PULLREQUEST THIS SHIT and here we are.

if we want to do all this 2-tall nonsense, we can probably iron out that behaviour as well, if needed.
or even, if you're really interested in the "goodlooking" system, i could make some 'sheets for those i guess 🤔 i guess it would be like having a 2-tall bar anyway, just with a different look and more logic to it. Not sure if it's worth doing though

myk002 (approved):

PR #2813, opened 2023-02-06
allow focus string generation to fall through unfocused ZScreens

myk002:
next steps for the focus string update

this allows the focus strings from the underlying DF viewscreen to come through, even when there is an unfocused DFHack tool window on the stack. Note that *focused* DFHack tool windows will be reflected in the focus string and will (correctly) *not* let the hotkey guard to function.

### DIFF
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -484,6 +484,12 @@ bool Gui::matchFocusString(std::string focusString, bool prefixMatch) {
     }) != currentFocusStrings.end();
 }
 
+static void push_dfhack_focus_string(dfhack_viewscreen *vs, std::vector<std::string> &focusStrings)
+{
+    auto name = vs->getFocusString();
+    focusStrings.push_back(name.empty() ? "dfhack" : "dfhack/" + name);
+}
+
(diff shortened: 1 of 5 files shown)


library/modules/Gui.cpp
```diff
@@ -484,6 +484,12 @@ bool Gui::matchFocusString(std::string focusString, bool prefixMatch) {
     }) != currentFocusStrings.end();
 }
 
+static void push_dfhack_focus_string(dfhack_viewscreen *vs, std::vector<std::string> &focusStrings)
+{
+    auto name = vs->getFocusString();
+    focusStrings.push_back(name.empty() ? "dfhack" : "dfhack/" + name);
+}
+
 std::vector<std::string> Gui::getFocusStrings(df::viewscreen* top)
```
myk002, line 493: note I reworked the logic in this function to allow focus strings to work at all for viewscreens without a specific handler (they were entering the second case and getting lost)

robob27 (approved):
LGTM. Also playtested and appears to work as described.

myk002:
Thanks!

PR #2832, opened 2023-02-06
Add 'key_back' optional hotkey for CycleHotkeyLabel to allow cycling backwards.

johncosker:
While working on https://github.com/DFHack/scripts/pull/593 I found that the CycleHotkeyLabel only going 'forwards' was kind of annoying while working with a longer list of options.

As I mentioned here: https://github.com/DFHack/scripts/pull/593#discussion_r1096543719 I thought a 'backwards' cycle key would be useful.

This is a pretty straightforward implementation of that.

I did see that there are GUI's that implement something similar using regular Labels, so if this change is deemed unnecessary then I'll withdraw the PR, but I think it might be useful.

I tested both my new GUI using the new key and `gui/quickfort` which doesn't use the new key and they both seem to work fine.

Example:
image

Which is created with:

```lua
widgets.CycleHotkeyLabel {
            view_id = "shape_name",
            key = "CUSTOM_Z",
            key_back = "CUSTOM_SHIFT_Z",
            label = "Shape: ",
            label_width = 8,
            active = true,
            enabled = true,
            options = options,
            disabled = false,
            show_tooltip = true,
            on_change = self:callback("change_shape"),
        }
```


docs/dev/Lua API.rst
```diff
@@ -4817,6 +4817,7 @@ cycle through by pressing a specified hotkey or clicking on the text.
 It has the following attributes:
 
 :key: The hotkey keycode to display, e.g. ``'CUSTOM_A'``.
+:key_back: Similar to ``key``, but will cycle backwards (optional)
```
myk002, line 4820: also please document the new optional parameter to `cycle()` on line 4838 below

johncosker: Done

myk002 (approved):

PR #2837, opened 2023-02-07
Add case_sensitive attr to FilteredList

robob27:
Add a `case_sensitive` attr to `FilteredList` to allow for case insensitive matching. Defaults to `true` so that existing `FilteredList`s keep the same behavior.

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -1866,6 +1866,7 @@ end
 FilteredList = defclass(FilteredList, Widget)
 
 FilteredList.ATTRS {
+    case_sensitive = true,
     edit_below = false,
     edit_key = DEFAULT_NIL,
     edit_ignore_keys = DEFAULT_NIL,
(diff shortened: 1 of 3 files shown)


library/lua/gui/widgets.lua
```diff
@@ -2026,11 +2027,17 @@ function FilteredList:setFilter(filter, pos)
...
                         not search_key:match('%f[^%s\x00]'..key) then
-                    ok = false
-                    break
+                            ok = false
+                            break
+                    elseif not string.lower(search_key):match('%f[^%p\x00]'..string.lower(key)) and
+                        not string.lower(search_key):match('%f[^%s\x00]'..string.lower(key)) then
+                            ok = false
+                            break
+                    end
```
myk002, line 2040: could we factor out the lowercasing and leave the rest of the code un-duplicated? every time a regular expression is duplicated in code, a magical forest creature dies.

myk002 (commented):
it would be very nice to have the match be case insensitive by default, but we need to check all users of the widget for negative interactions first. I'm happy merging this with default case sensitivity now, but once we have time to check the (in-tree) users of List for negative interactions, we can switch the default value to false.

also, changelog plz

myk002 (commented):
looks good; just needs a changelog entry

robob27:
> looks good; just needs a changelog entry

Would it go in the future section currently?

myk002:
Yes, all new changelog entries go in "Future"

myk002 (approved):

Issue #2065, opened 2022-03-31
add event for recordkeeper is updating stockpile records


ab9rf:
Several plugins (e.g. `stockflow`, `tailor`) currently have code that detects when the recordkeeper starts updating the fortress's stockpile records and use this to trigger the effects of these plugins.

This code is duplicated and seems to fit within the scope of eventful; thus, it seems reasonable for "recordkeeper begins updating stockpile records" to be an event that other plugins/scripts can capture.

PR #2843, opened 2023-02-08
[tailor] persist state, use best practices

### LINKED ISSUE #2835: `tailor` needs to be manually enabed every time you start DF
Not sure if it's intended, but I enabled `tailor` on the control panel, and it only lasts for the session. If I restart DF, it gets turned off again.

myk002:
Fixes #2835

Includes adding Lua for the eventual gui config.

Because of the questions that have been raised recently about bookkeepers and their schedules, I changed the scanning from being based on the bookkeeper to just once daily (like other plugins). We might change it back when we address #2065 though.

Note I added a few flags to the "bad" mask, stolen from similar constructs in other plugins:

- in_chest
- removed
- encased
- spider_web

Is there any reason these flags should not be added?


moontayle:
Doing Armok's work. Praise Urist!

PR #2844, opened 2023-02-08
Implements plugin: channel-safely v1.2.3

### LINKED ISSUE #2818: Channel-safely cannot set marker mode
A player reported bug on discord regarding channel-safely not working correctly and crashing. Details on the crash are unclear and may very well be unrelated to channel-safely. The main trouble identified and described the player was marker mode not being set quickly enough for designations. Of course upon load or enabling the plugin the marker mode should be set for designations that require it.

I was able to reproduce the issue using the save game provided by the player. Since marker mode is part of the manager class I enabled the trace logs for the manager via `debugfilter set Trace channel manager`. Additionally also adding the debug logs for the groups processing in order to have access to some of the coordinates being processed.

The manager logs are as follows:
```
managing group #3
cave-in evaluated false
cave-in evaluated false
cave-in evaluated false
cave-in evaluated false
manage_group() is done
```

Group 3:
```
  group 3 (size: 4)
   (123,29,124)
   (123,30,124)
   (123,31,124)
   (124,31,124)
```

The manager logs are as deep as the log line seen here:
https://github.com/DFHack/dfhack/blob/a7b5ce418e931fdf19dd6662385828173d186cbf/plugins/channel-safely/channel-manager.cpp#L110-L116

This code of course calls `manage_one` after printing the log line to indicate where the algorithm is. So taking a look at that function we can see it should print another log line so long as the position passed is a valid one.
https://github.com/DFHack/dfhack/blob/a7b5ce418e931fdf19dd6662385828173d186cbf/plugins/channel-safely/channel-manager.cpp#L153-L155

So then, looking at the function used to check whether that is the case or not. We can see it checks if `IsValid()` then simply checks if the coordinates are within the bounds of the map.

https://github.com/DFHack/dfhack/blob/a7b5ce418e931fdf19dd6662385828173d186cbf/library/modules/Maps.cpp#L193-L202

https://github.com/DFHack/dfhack/blob/a7b5ce418e931fdf19dd6662385828173d186cbf/library/modules/Maps.cpp#L104-L107

Looking at both these functions, the data being compared against (other than what the plugin contains) appears to all be within `world->map`. So logging the data inside we see:
```
df.global.world.map: <world.T_map: 00007FF7646B7EA8>
| map_blocks: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ <vector<map_block*>[11664]: 00007FF7646B7EA8>
| block_index: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <map_block***: 000001DC678B5810>
| map_block_columns: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <vector<map_block_column*>[81]: 00007FF7646B7EC8>
| column_index: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ <map_block_column**: 000001DC678B2110>
| x_count_block: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  9
| y_count_block: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  9
| z_count_block: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  144
| x_count: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  144
| y_count: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  144
| z_count: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  144
| region_x: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ 147
| region_y: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ 208
| region_z: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ 1
| distance_lookup: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <int16_t[53][]: 00007FF7646B7F0C>
```

We can see that `block_index` appears to be not `NULL` nor do the coordinates of group 3 fall outside the ranges here. Which of course means that it does not make a lot of sense for this check to fail. As the coordinate data is valid and the map appears valid.

- Nothing in the algorithm modifies the coordinates.
- The coordinates are valid
- The coordinates are output to console showing that they can be programmatically accessed and iterated over

cppcooper:
- Revises a few log lines
- Adds d_assert macro to replace assert usage
  - provides a log line printed to the console to indicate d_assert failed in Release/Debug builds
  
This resolves #2818 (presumably)

### DIFF
--- a/plugins/channel-safely/channel-manager.cpp
+++ b/plugins/channel-safely/channel-manager.cpp
@@ -5,7 +5,17 @@
 #include <modules/EventManager.h> //hash function for df::coord
 #include <df/block_square_event_designation_priorityst.h>
 
+#define NUMARGS(...) std::tuple_size<decltype(std::make_tuple(__VA_ARGS__))>::value
+#define d_assert(condition, ...) \
+            static_assert(NUMARGS(__VA_ARGS__) >= 1, "d_assert(condition, format, ...) requires at least up to format as arguments"); \
+            if (!condition) {                                                   \
+                DFHack::Core::getInstance().getConsole().printerr(__VA_ARGS__); \
+                assert(0);                                                      \
+            }
+
+
 df::unit* find_dwarf(const df::coord &map_pos) {
+
     df::unit* nearest = nullptr;
     uint32_t distance;
     for (auto unit : df::global::world->units.active) {
@@ -57,6 +67,7 @@ void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool
     // cavein prevention
     bool cavein_possible = false;
     uint8_t least_access = 100;
+
     std::unordered_map<df::coord, uint8_t> cavein_candidates;
     if (!marker_mode) {
         /* To prevent cave-ins we're looking at accessibility of tiles with open space below them
@@ -111,7 +122,7 @@ void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool
         // if no cave-in is possible [or we don't check for], we'll just execute normally and move on
         if (!cavein_possible) {
             TRACE(manager).print("cave-in evaluated false\n");
-            assert(manage_one(pos, true, marker_mode));
+            d_assert(manage_one(pos, true, marker_mode), "manage_one() is failing under !cavein");
             continue;
         }
         // cavein is only possible if marker_mode is false
@@ -136,16 +147,16 @@ void ChannelManager::manage_group(const Group &group, bool set_marker_mode, bool
                     evT->priority[Coord(local)] = v;
                 }
             }
-            assert(manage_one(pos, true, false));
+            d_assert(manage_one(pos, true, false), "manage_one() is failing for cavein ");
             continue;
         }
         // cavein possible, but we failed to meet the criteria for activation
         if (cavein_candidates.count(pos)) {
-            DEBUG(manager).print("cave-in evaluated true and no dignow and (%d > %d)\n", cavein_candidates[pos], least_access+OFFSET);
+            DEBUG(manager).print("cave-in evaluated true and the cavein candidate's accessibility check was made as (%d <= %d)\n", cavein_candidates[pos], least_access+OFFSET);
         } else {
-            DEBUG(manager).print("cave-in evaluated true and no dignow and pos is not a candidate\n");
+            DEBUG(manager).print("cave-in evaluated true and the position was not a candidate, nor was it set for dignow\n");
         }
-        assert(manage_one(pos, true, true));
+        d_assert(manage_one(pos, true, true), "manage_one() is failing to set a cave-in causing designation to marker mode");
     }
     INFO(manager).print("manage_group() is done\n");
 }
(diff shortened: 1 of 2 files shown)


plugins/channel-safely/channel-manager.cpp
```diff
@@ -5,7 +5,17 @@
 #include <modules/EventManager.h> //hash function for df::coord
 #include <df/block_square_event_designation_priorityst.h>
 
+#define NUMARGS(...) std::tuple_size<decltype(std::make_tuple(__VA_ARGS__))>::value
+#define d_assert(condition, ...) \
+            static_assert(NUMARGS(__VA_ARGS__) >= 1, "d_assert(condition, format, ...) requires at least up to format as arguments"); \
+            if (!condition) {                                                   \
+                DFHack::Core::getInstance().getConsole().print(__VA_ARGS__);    \
```
myk002, line 12: this should probably be `ERROR(manager).print(__VA_ARGS__)'`

myk002 (commented):
How does this PR resolve the linked issue, though? All I see is instrumentation.

cppcooper:
> How does this PR resolve the linked issue, though? All I see is instrumentation.

the only logical conclusion I was able to make was that the function call was never made as all the evidence pointed to the valid tile pos function being able to return true yet inexplicably returning false instead. So by revising the assertions in the manager with that d_assert the function should now always be invoked.

I'm going to test out the github windows artifact tomorrow, hopefully. I'm pretty sure this will fix the problem though.

plugins/channel-safely/channel-manager.cpp
cppcooper: I was going to go with printerr like lethosor suggested on discord. Should I prefer this?

myk002: looking at the code, no, not yet. use printerr so it goes to stderr. we need to modify the debug logging framework to also print to stderr for WARN and above

cppcooper: Earlier today I had a brief thought about a plugin that would share stdout and stderr with a file log. I vaguely recall having written some code to do that, pretty sure it was some C code

cppcooper:
> I'm pretty sure this will fix the problem though.

verified

plugins/channel-safely/channel-manager.cpp
myk002: will you change this to printerr for this PR? it seems like output splitting would be a larger project

cppcooper: Yea, my local has the commit. I was just waiting to boot back to linux to push

PR #2845, opened 2023-02-08
Remove silver crossbow work-order from library/military

dawnmist:
It is not possible to create a work-order for making silver crossbows, so the silver crossbow recipe belongs in the alternative military library with other artifact/non-craftable weapon recipes.

It's already present in the library/military_include_artifact_materials work-orders, so it simply needed removal from the vanilla library/military orders.

### DIFF
--- a/data/orders/military.json
+++ b/data/orders/military.json
@@ -1142,42 +1142,6 @@
 		"amount_left" : 1,
 		"amount_total" : 1,
 		"frequency" : "Daily",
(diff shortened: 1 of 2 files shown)


myk002 (commented):
could you also remove the conditions from the other metal crossbows that check for silver?

myk002:
also, could you possibly add a changelog entry for this?

myk002 (approved):

PR #2847, opened 2023-02-08
Add setLabel/setOnActivate to HotkeyLabel

robob27:
- Move the `HotkeyLabel.init` logic to a a new method: `HotkeyLabel.initializeLabel`. 
- Adds `HotkeyLabel.setLabel` and `HotkeyLabel.setOnActivate` to HotkeyLabel to allow easily updating a the label or onactivate function without having to call `initializeLabel` yourself.

I'm unsure if there was already a "proper" way to do this already that I couldn't figure out (without having to call `setText` the same way that it was called during `init()`). Directly setting the text mangles the keyboard shortcut part, and directly setting the `on_activate` doesn't seem to update the `on_activate` function right away. This is basically what I want to achieve:

```
self.subviews.context_controls.visible = self:contextControlsEnabled()

-- setup context controls
if view_mode == VIEW_MODE.Animals then
    self.subviews.context_control_1:setLabel('Butcher')
    self.subviews.context_control_1:setOnActivate(function() print('Butcher') end)
    self.subviews.context_control_2:setLabel('Geld')
    self.subviews.context_control_2:setOnActivate(function() print('Geld') end)
elseif view_mode == VIEW_MODE.Citizens then
    self.subviews.context_control_1:setLabel('Citizen Action 1')
    self.subviews.context_control_1:setOnActivate(function() print('Citizen Action 1') end)
    self.subviews.context_control_2:setLabel('Citizen Action 2')
    self.subviews.context_control_2:setOnActivate(function() print('Citizen Action 2') end)
end
```


myk002:
I would actually suggest having two HotkeyLabels at the same frame position and switching which one is visible. That will also control which interprets the hotkey, since onInput propagation checks the `visible` property.

robob27:
> I would actually suggest having two HotkeyLabels at the same frame position and switching which one is visible. That will also control which interprets the hotkey, since onInput propagation checks the `visible` property.
> 
> Would that solve your use case? If not, we certainly could consider something like this PR

I think it would, though it feels like it would be a lot messier if you have more than 2 contexts. Instead of having just 2 `HotkeyLabel`s for my 2 context controls, I'd have 2 * numberOfContextsWithContextControls `HotkeyLabel`s. And then to control the visibility, I'd have to either keep track of the last context to know which ones to hide, or just blanket hide  2 * numberOfContextsWithContextControls `HotkeyLabel`s and then show the right ones. Not impossible but it does seem nicer to have the option to work with just 2 `HotkeyLabel`s for this sort of thing imo. I can make it work if that's preferred though!

myk002:
If you have many, there *is* the `widget.Pages` widget that will do the tracking for you, but I agree that there can be cases where setting the properties directly can be useful. I'm good with this PR.

myk002 (approved):

PR #2853, opened 2023-02-09
[nestboxes] remove in-progress jobs when protecting fert eggs

### LINKED ISSUE #2846: Nestboxes should also cancel jobs attached to eggs
So if the haul/cook job won't continue to grab the forbidden egg anyway

myk002:
Fixes #2846

in DF 50, forbidding an item is not enough to cancel the associated job. we now need to do that manually.


myk002:
verified by Hiran on Discord

Issue #2850, opened 2023-02-08
autoslab creating duplicate engraved slabs


johncosker:
There's a ghost in my fort named Zasit Åkumurdim. The engrave job for him keeps getting queued over again, which means the string comparison to check for existing engraved slab items isn't working correctly in `autoslab`.

It just seems to be that one dwarf which makes me wonder if it has to do with the string encoding or something and it's not handling the 'Å' correctly when calling `std::string::find`

When writing the plugin I did test with dwarves with non-ASCII characters and it seemed to work so not sure yet what the exact issue is.

johncosker:
This is why in the plugin I commented I'm not super happy about having to use string comparison for the slab text compared to something more solid like checking the historical figure id, still don't have an alternative though.

johncosker:
For now a workaround is to place the slabs and get rid of the ghost.

lethosor:
Can you upload the save?

johncosker:
I added a debug log and saw that this is the issue

Checking 'marble memorial to Zasit Åkumurdim' for Zasit åkumurdim: Result false

The last name is generated from a function I stole from another plugin:

```c++
static std::string get_last_name(df::unit *unit)
{
    df::language_name name = unit->name;
    std::string ret = "";
    for (int i = 0; i < 2; i++)
    {
        if (name.words[i] >= 0)
            ret += *world->raws.language.translations[name.language]->words[name.words[i]];
    }
    return Translation::capitalize(ret);
}
```

So it looks like the capitalization isn't working for at least that letter.

johncosker:
`Translation::capitalize` is just calling `std::toupper(input_string[0])` so I think the issue is that in this case the first letter is a unicode character it's failing...

Indeed, if you try to do this it doesn't compile

`std::cout << std::toupper('å');`
```
main.cpp:15:31: warning: multi-character character constant [-Wmultichar]
   15 |     std::cout << std::toupper('å');
      |                               ^~~
```

I think the reason I didn't see this in testing with other names with non-ascii characters is because in those cases the letter was in some other position in the name other than 0.

johncosker:
This seems to work: https://stackoverflow.com/questions/17991431/convert-a-unicode-string-in-c-to-upper-case

```c++
#include <iostream>
#include <locale.h>

int main() {
    std::locale::global(std::locale(""));  // (*)
    std::wcout.imbue(std::locale());
    auto& f = std::use_facet<std::ctype<wchar_t>>(std::locale());

[...]

PR #2854, opened 2023-02-09
autoslab: Improve check for existing engraved slab.

johncosker:
https://github.com/DFHack/dfhack/issues/2850 caused me to revisit `autoslab`. I found that the issue was the `Translation::capitalize()` function wasn't handling ~~unicode~~ CP437 characters correctly if one happened to be at the start of a word.

This caused an issue for `autoslab` because it was using a string comparison to check for the dwarf's name in the slab description.

As I mentioned in the Issue I never felt great about that solution so I took a second look to see if I could find where the slab stores the historical id of the memorialized dwarf and lo and behold I found it was actually the `topic` field in the `df::item_slabst` struct.

So the code is now improved and no longer dependent on string comparison.


plugins/autoslab.cpp
```diff
@@ -228,7 +210,14 @@ static void checkslabs(color_ostream &out)
     for (auto ghost : ghosts)
     {
         // Only create a job is the map has no existing jobs for that historical figure or no existing engraved slabs
-        if (histToJob.count(ghost->hist_figure_id) == 0 && !engravedSlabItemExists(ghost, engravedSlabs))
+        if (histToJob.count(ghost->hist_figure_id) == 0 &&
+            !std::any_of(engravedSlabs.begin(),
+                        engravedSlabs.end(),
+                        [&ghost](df::item *slab){
+                            auto slabst = virtual_cast<df::item_slabst>(slab);
```
myk002, line 217: this caught my eye here since it's usually dangerous to virtual_cast without checking for NULL. it should always work here, but you wouldn't have to convert if `engravedSlabs` were a `std::vector<df::item_slabst>`, and you can cut down on your scanning *and* get the correct type if you scan `world->items.other.SLAB` instead of `world->items.all`

myk002 (approved):

PR #2859, opened 2023-02-10
Add getSelectedCivZone to dfhack.gui

kelvie:
Mostly helpful for my own automation in my custom lua scripts.

Tested it with barracks, meeting areas, and pastures.

image

### DIFF
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -98,14 +98,15 @@ namespace DFHack
 
 using namespace df::enums;
 
+using df::building_civzonest;
+using df::global::game;
 using df::global::gamemode;
 using df::global::gps;
 using df::global::gview;
 using df::global::init;
-using df::global::selection_rect;
 using df::global::plotinfo;
+using df::global::selection_rect;
 using df::global::ui_menu_width;
-using df::global::game;
 using df::global::world;
 
 /* TODO: understand how this changes for v50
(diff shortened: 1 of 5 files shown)


library/modules/Gui.cpp
```diff
@@ -1111,6 +1112,26 @@ df::building_stockpilest* Gui::getSelectedStockpile(color_ostream& out, bool qui
...
 }
 
+bool Gui::any_civzone_hotkey(df::viewscreen* top) {
+    return getAnyCivZone(top) != NULL;
+}
+
+df::building_civzonest *Gui::getAnyCivZone(df::viewscreen* top) {
+    if (matchFocusString("dwarfmode/Zone")) {
+        return game->main_interface.civzone.cur_bld;
+    }
```
myk002, line 1122: Just to check, does this get the right value when multiple zones overlap?

kelvie: I shall check tonight, I more or less copy and pasted this logic from the other `get` functions.

myk002: ack. the reason I'm suspicious is that civzones are the only building type that can overlap, so the rules might be different.

kelvie: Just tested, this works properly when they're completely overlapped (tested a meeting hall and an office stacked on top of each other exactly, when I change the selection using the left and right arrows here:

image

the function returns the correct civzone.

myk002 (commented):
could you add changelog entries under API for the cpp code and under Lua for the Lua function?

kelvie:
> could you add changelog entries under API for the cpp code and under Lua for the Lua function?

Oh, yeah, didn't realize these were documented. I'll do this tomorrow if I have time.

kelvie:
Updated dogs and changelog, waiting for build to check formatting.

myk002 (approved):

PR #2860, opened 2023-02-10
Detect manual save when persisting data

robob27:
Include manual saves when detecting when to write persisted data


myk002 (approved):
tested with regular saves, timeline saves, and manual saves. seems to all work. Thanks!

lethosor:
Thanks! Changelog LGTM :)

PR #2867, opened 2023-02-11
nestboxes: avoid forbidding non-eggs

lethosor:
The rewrite in #2773 introduced a bug (identified by KMFrench on Discord) where nestboxes themselves would be forbidden, which prevents egglaying citizens from using them since the nestboxes would be claimed before use. The previous version of the plugin handled this by skipping the first item in `contained_items`, which is the building material (in this case, the nestbox).

This fix limits the `forbid` flag manipulation to egg items only, which should be a reasonable limitation that still allows the plugin to serve its purpose.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,6 +36,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## New Plugins
 
 ## Fixes
+-@ `nestboxes`: fixed bug causing nestboxes themselves to be forbidden, which prevented citizens from using them to lay eggs. Now only eggs are forbidden.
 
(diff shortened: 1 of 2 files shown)


myk002 (approved):
changelog entry

lethosor:
hah, true

KMFrench reports that this solves their issue

ab9rf:
as discussed on discord, i think we should probably exclude from the item loop any items that are in use_mode 2, even if they are eggs. while i don't know of any way to build a nestbox out of eggs, i don't want to underestimate the ingenuity of df modders to do things we don't expect

i'll address this in a separate PR probably

PR #2875, opened 2023-02-11
Add Tab/TabBar to widgets

robob27:
Attempt at adding Tab/TabBar as widgets, with some slight modifications to allow customization where users may want to do so, and first pass at some docs to go along with it.

If/once this looks alright I can do a followup either here or in another PR to switch control-panel.lua over to use these.

If it looks like I had an aneurysm with spelling anywhere I just wanna throw it out there in advance.. I have covid and fully blame that 😛

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -52,7 +52,6 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## API
 - ``Gui::any_civzone_hotkey``, ``Gui::getAnyCivZone``, ``Gui::getSelectedCivZone``:  new functions to operate on the new zone system
-
 - Units module: added new predicates for:
   - ``isGeldable()``
   - ``isMarkedForGelding()``
@@ -60,8 +59,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - ``dfhack.gui.getSelectedCivZone``: returns the Zone that the user has selected currently
-
 - ``widgets.FilteredList``: Added ``edit_on_change`` optional parameter to allow a custom callback on filter edit change.
+- Added ``widgets.TabBar`` and ``widgets.Tab`` (migrated from control-panel.lua)
 
 ## Removed
 
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -5038,11 +5038,11 @@ the width of the window and will continue rendering on the next line(s) if all t
 :key_back: Specifies a keybinding that can be used to switch to the previous tab.
 :labels: A table of strings; entry representing the label text for a single tab. The order of the entries
          determines the order the tabs will appear in.
-:get_cur_page: Returns the current "page". This function does not have a default implementation; you must provide
-               an implementation that returns the current value of whichever variable your script uses to keep track of the
-               current "page" (this does not need to relate to an actual Pages widget).
-:on_select: Callback executed when a tab is selected. It receives the selected tab index as an argument. Your implementation
-            should likely update the value of whichever variable your script uses to keep track of the current page.
+:on_select: Callback executed when a tab is selected. It receives the selected tab index as an argument. The provided function
+            should update the value of whichever variable your script uses to keep track of the currently selected tab.
+:get_cur_page: The function used by the TabBar to determine which Tab is currently selected. The function you provide should
+               return an integer that corresponds to the non-zero index of the currently selected Tab (i.e. whatever variable
+               you update in your ``on_select`` callback)
 :active_tab_pens: A table of pens used to render active tabs. See the default implementation in widgets.lua for an example
                   of how to construct the table. Leave unspecified to use the default pens.
 :inactive_tab_pens: A table of pens used to render inactive tabs. See the default implementation in widgets.lua for an example
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -5028,6 +5028,41 @@ The widget implements:
 
   Same as with an ordinary list.
 
+TabBar class
+------------
+
+This widget implements a set of one or more tabs to allow navigation between groups of content. Tabs automatically wrap on
+the width of the window and will continue rendering on the next line(s) if all tabs cannot fit on a single line.
+
+:key: Specifies a keybinding that can be used to switch to the next tab.
+:key_back: Specifies a keybinding that can be used to switch to the previous tab.
+:labels: A table of strings; entry representing the label text for a single tab. The order of the entries
+         determines the order the tabs will appear in.
+:get_cur_page: Returns the current "page". This function does not have a default implementation; you must provide
+               an implementation that returns the current value of whichever variable your script uses to keep track of the
+               current "page" (this does not need to relate to an actual Pages widget).
+:on_select: Callback executed when a tab is selected. It receives the selected tab index as an argument. Your implementation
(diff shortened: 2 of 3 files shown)


robob27:
pre-commit.ci autofix

docs/dev/Lua API.rst
```diff
@@ -5012,6 +5012,39 @@ The widget implements:
...
+
+This widget implements a single clickable tab and is the main component of the TabBar widget. Usage of the ``TabBar``
+widget does not require direct usage of ``Tab``.
+
+:id: The id of the tab.
+:label: The text displayed on the tab.
+:on_select: Callback executed when the tab is selected.
+:get_pens: A function that is used during ``Tab:onRenderBody`` to determine the pens that should be used for drawing. See the
+           usage of ``Tab`` in ``TabBar:init()`` for an example. See the default value of ``active_tab_pens`` or ``inactive_tab_pens``
+           in ``TabBar`` for an example of how to construct pens.
```
myk002, line 5047: This is why I didn't push the widgets straight into the library. The API as it stands just feels overly complicated and unencapsulated.

I think what I'd like to see is there not being any externally-visible Tab or TabBar class at all. Instead, widget.Pages can take a `labels` attribute, and if it is set, then it creates and manages the tabs itself. Is there a strong use case for these tabs *outside* of a Pages widget?

robob27: I don't know if it's a strong use case but.. I'll describe how I'm currently using it outside of a Pages widget and see how that goes haha

In that units search/overview script I've been working on:

image

I'm [not using a Pages widget](https://gist.github.com/robob27/1a30758877566def65f80a8ae3a22fc3#file-unit-search-lua-L1128-L1324). I make the top TabBar responsible for controlling a variable called `view_mode`, and the second TabBar responsible for a variable called `active_subviews`. Each time the view mode or subview mode is updated, I refresh the `units` table [with units that are valid for the current mode/submode](https://gist.github.com/robob27/1a30758877566def65f80a8ae3a22fc3#file-unit-search-lua-L567-L634) and then update the single shared `FilteredList` called `unit_list`. All of the hotkeylabels etc are shared by all views and just have their labels and callbacks updated as needed based on the view/subview mode.

It feels like if I were to use a Pages widget for this, I would end up having to duplicate a lot of widgets to achieve the same sort of layout. For my particular use case, it feels like I just want a pretty way to control the value of 2 variables rather than a way to control the visibility of grouped sets of widgets.

myk002: That sounds legit. Essentially a CycleHotkeyLabel with all options visible and directly selectable.

docs/dev/Lua API.rst
```diff
@@ -5012,6 +5012,40 @@ The widget implements:
...
+------------
+
+This widget implements a set of one or more tabs to allow navigation between groups of content. Tabs automatically wrap on
+the width of the window and will continue rendering on the next line(s) if all tabs cannot fit on a single line.
+
+:labels: A table of strings; entry representing the label text for a single tab. The order of the entries
+         determines the order the tabs will appear in.
+:get_cur_page: Returns the current "page". This function does not have a default implementation; you must provide
+               an implementation that returns the current value of whichever variable your script uses to keep track of the
+               current "page" (this does not need to relate to an actual Pages widget).
```
myk002, line 5025: shouldn't this be a little more prescriptive? shouldn't it return the index that corresponds to the active tab? otherwise the default `get_pens` function won't work.

docs/dev/Lua API.rst
```diff
@@ -5012,6 +5012,40 @@ The widget implements:
...
+:on_select: Callback executed when a tab is selected. It receives the selected tab index as an argument. Your implementation
+            should likely update the value of whichever variable your script uses to keep track of the current page.
+:active_tab_pens: A table of pens used to render active tabs. See the default implementation in widgets.lua for an example
+                  of how to construct the table. Leave unspecified to use the default pens.
+:inactive_tab_pens: A table of pens used to render inactive tabs. See the default implementation in widgets.lua for an example
+                    of how to construct the table. Leave unspecified to use the default pens.
+:get_pens: A function used to determine which pens should be used to render a tab. Receives the index of the tab as the first
+           argument and the TabBar widget itself as the second. The default implementation, which will handle most situations,
+           returns ``self.active_tab_pens``, if ``self.get_cur_page() == idx``, otherwise returns ``self.inactive_tab_pens``.
+:switch_tab_key: Specifies a keybinding that can be used to switch to the next tab.
```
myk002, line 5035: CycleHotkeyLabel now has a hotkey for both forwards and backwards. You might want the same flexibility here. Maybe also use the same attribute names: `key` and `key_back`.

this is also more important than the `pens` attributes, and should probably be moved higher up.

robob27: Moved to top, renamed `switch_tab_key` to `key`, added `key_back`

docs/changelog.txt
```diff
@@ -52,16 +52,15 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## API
 - ``Gui::any_civzone_hotkey``, ``Gui::getAnyCivZone``, ``Gui::getSelectedCivZone``:  new functions to operate on the new zone system
-
```
robob27, line 55: I'm not sure why these extra line breaks were in here, and if I should just leave them?

docs/dev/Lua API.rst
robob27: Yeah, that's pretty much what I was trying to get across here. I struggle with doc writing a lot haha. I'll try to reword it.

robob27: Attempted to make it a bit more prescriptive.

docs/changelog.txt
myk002: the GitHub web UI is a beast to use for resolving merge conflicts on mobile. they might be left over from that. Taking them out is fine. Otherwise I'll clean them up in some future edit.

library/lua/gui/widgets.lua
```diff
@@ -2091,4 +2091,156 @@ function FilteredList:onFilterChar(char, text)
...
+        local next_zero_idx = (zero_idx + 1) % #self.labels
+        self.on_select(next_zero_idx + 1)
+        return true
+    end
+    if self.key_back and keys[self.key_back] then
+        local zero_idx = self.get_cur_page() - 1
+        local prev_zero_idx = (zero_idx - 1) % #self.labels
+        self.on_select(prev_zero_idx + 1)
+        return true
+    end
```
myk002, line 2243:
```suggestion
    if self.key_back and keys[self.key_back] then
        local cur_page = self.get_cur_page()
        if cur_page == 1 then
            self.on_select(#self.labels)
        else
            self.on_select(cur_page - 1)
        end
        return true
    end
```

robob27: It seemed like this was being handled elsewhere. When I tested with what I had it just worked, including wrapping around to the end. Can update if it's preferred to explicitly handle it here though.

myk002: ah, I guess in Lua an index of -1 would wrap around. that's true, though the client-provided `on_select` function might not be expecting it. the `% #self.labels` is a noop, though.

myk002 (approved):

myk002:
could you make a followup PR to migrate gui/control-panel and gui/launcher to the library implementation?

robob27:
> could you make a followup PR to migrate gui/control-panel and gui/launcher to the library implementation?

Yeah will do

PR #2882, opened 2023-02-12
buildingplan, first draft

### LINKED ISSUE #2632: quickfort/blueprint doesn't allow for material selection, acting always like "Use closest material"
When using a /construct blueprint, i.e. build walls for a room instead of digging them out, the constructed walls behave as if you had selected "Use closest material" during build.

As far as I can see there is no way to change this or select material.

To make the /construct fully usable some form of material selection should be implemented so that the user can implement the blueprint and have it using the same resources on the whole build. As it stands, it's close to unusable in that mode.

Thanks for all your work, this thing is amazing.

### LINKED ISSUE #2644: Suggestion: Bring back up, down, up-down stairs somehow
The new "intuitive" stairbuilding works for most cases but in some cases I found it to be quite a pain to make it do exactly what I want. Is it possible for dfhack to bring back the old stairs somehow; maybe next to the existing system?

myk002:
Should implement (at least most of) #2883
Fixes #2644
Fixes #2632

Starting a PR just to get some sharable builds

Checklist of ideas so I don't forget (even if not all of them will be going into this PR):

- [ ] prevent buildingplan-suspended buildings from being unsuspended
- [ ] allow construction areas to be placed even if some tiles are invalid (just skip the invalid tiles)
- [ ] allow filters to be updated from the Inspection overlay (will likely require canceling the building and recreating in order to ensure consistency in the task queues)
- [ ] alternate mode for item choosing - select from map: shrink dialog, highlight possible choices on map, track selections and add them to the list
- [ ] CycleHotkeyLabel to choose between Single and Multi placement mode


silverwing235:
Hey there - one of the checks (https://github.com/DFHack/dfhack/pull/2882/checks?check_run_id=11313054872) seems to be stuck on repeat?

PR #2891, opened 2023-02-12
Add edit_on_change to FilteredList

robob27:
Allow providing a custom callback for `edit.on_change` in the same way we do for `edit.on_char`. I had a need to get the final value of the edit widget after a change, and `on_char` wasn't able to reliably give me what I needed (can't concat `c` and `text` without taking the cursor position into account).

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -53,6 +53,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 
+- ``widgets.FilteredList``: Added ``edit_on_change`` optional paramter to allow a custom callback on filter edit change.
+
 ## Removed
 
 # 50.07-alpha1
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -53,7 +53,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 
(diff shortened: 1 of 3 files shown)


docs/changelog.txt
```diff
@@ -53,6 +53,8 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 
+- ``widgets.FilteredList``: Added ``edit_on_change`` optional paramter to allow a custom callback on filter edit change.
```
myk002, line 56:
```suggestion
- ``widgets.FilteredList``: Added ``edit_on_change`` optional parameter to allow a custom callback on filter edit change.
```

myk002 (commented):
That's a very good point about on_char no longer being able to discern how the text is changing now that we support an edit cursor (the edit cursor is a very recent addition, just prior to the steam release) I'll file an issue to look into that. there are definitely scripts that think they have the complete text in on_char and use that info to make decisions.

robob27:
>That's a very good point about on_char no longer being able to discern how the text is changing now that we support an edit cursor (the edit cursor is a very recent addition, just prior to the steam release) I'll file an issue to look into that. there are definitely scripts that think they have the complete text in on_char and use that info to make decisions.

Oh nice, yeah I didn't even consider that other scripts might be using `onChar` in an attempt to get the full text as I was unaware edit cursor was new-ish. Nice random catch here.

myk002 (approved):

PR #2904, opened 2023-02-13
Persist confirm config

### LINKED ISSUE #2815: Persist confirm state
User config should be stored in dfhack-config/confirm.json and not be associated with the current fort since the choices are global

robob27:
Load confirm config from dfhack-config/confirm.json. Sister scripts PR: https://github.com/DFHack/scripts/pull/617

Closes #2815

### DIFF
--- a/plugins/confirm.cpp
+++ b/plugins/confirm.cpp
@@ -612,15 +612,7 @@ DFhackCExport command_result plugin_init (color_ostream &out, vector <PluginComm
 
 DFhackCExport command_result plugin_enable (color_ostream &out, bool enable)
 {
-    if (is_enabled != enable)
-    {
-        for (auto c : confirmations)
-        {
-            if (!c.second->apply(enable))
-                return CR_FAILURE;
-        }
-        is_enabled = enable;
-    }
+    is_enabled = enable;
     if (is_enabled)
     {
         conf_lua::simple_call("check");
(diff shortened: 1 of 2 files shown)


plugins/lua/confirm.lua
```diff
@@ -1,4 +1,26 @@
...
+        local ok, config = pcall(json.decode_file, CONFIRM_CONFIG_FILE)
+        local conf_data = ok and config or get_conf_data()
+        if ok then
+            for _, c in ipairs(conf_data) do
+                set_conf_state(c.id, c.enabled)
+            end
+        end
+        loaded = true
+    end
+end
```
myk002, line 23: splitting the save/restore logic across files seems like asking for trouble down the line. could you move this logic into gui/confirm?

myk002: note that you have to make `gui/confirm` a `--@ module=true`

robob27: I did initially try doing this but it made the confirm options pop up as soon as the game loaded and I wasn't sure the best way to stop that from happening

robob27: As soon as I add `--@ module=true`:

image

myk002: you can't just *say* you're a module. you also have to *act* like a module : )

add:
```lua
if dfhack_flags.module then
    return
end
```

before the script has any side effects (like bringing up a window)

myk002: e.g. https://github.com/DFHack/scripts/blob/b13f351b22de50210b9b71d34806ebc2b84351ca/gui/quickfort.lua#L677

robob27: Haha.. sorry. I'm still covid dumb. That's what I missed. Thanks.

myk002 (approved):

robob27:
@myk002 You probably already know but just a heads up that there was the c++ changes since the approval

plugins/confirm.cpp
```diff
@@ -612,15 +612,7 @@ DFhackCExport command_result plugin_init (color_ostream &out, vector <PluginComm
...
-    if (is_enabled != enable)
-    {
-        for (auto c : confirmations)
-        {
-            if (!c.second->apply(enable))
-                return CR_FAILURE;
-        }
-        is_enabled = enable;
-    }
+    is_enabled = enable;
```
robob27, line 615: Not sure if I should just check if `is_enabled != enabled` before setting it always, or only if we are doing something when the state changes. Maybe just check anyways?

myk002: I think this is fine.

myk002 (approved):

PR #2906, opened 2023-02-14
Update Buildings.cpp

silverflyone:
Add = to block loop check. If the edge of a stockpile is at a block edge, the comparison was checking 15 tiles (<) rather than 16 tiles (<=). This was resulting in incomplete items being returned for Items::getContainedItems.


silverflyone:
I found this while testing the combine.lua stockpile merge; some items were not being returned for 2 two tile wide stockpile. The combine.lua change needs this change in order to merge properly.

Issue #2912, opened 2023-02-16
List:setChoices rendering issues


robob27:
The test script used to demonstrate these issues can be found [here](https://gist.github.com/robob27/ece52d0adc9310b0018f8c05a8ec7b74). I initially found the issues in a much more complex script so I tried to distill it down to something very simple that could also demonstrate the issues.

If you call `setSelected` in the same frame as the list is rendered, `page_top` is not calculated correctly. This results in the scrollbar and available options being rendered incorrectly:
![image](https://user-images.githubusercontent.com/2104555/219223222-8bad930c-a9f3-40b7-ae71-2388cb8d4f7f.png)

This only happens if there is not enough space to display all of the options, unlike in `gui/confirm` where we have exactly as many confirms as we have available space in the list, so this is not yet a problem. Here is same list choices, with setSelected also called, but with enough space to display all options:

![image](https://user-images.githubusercontent.com/2104555/219223448-6aa062bb-1d09-436d-8e58-cc92003e0828.png)

I believe the issue is only when these both happen in the same frame because if I render the list first, and then do some action to populate the list after, even if there is not enough room to display all options, it renders correctly:

Rendering the list first without any choices:

![image](https://user-images.githubusercontent.com/2104555/219223621-6e91f30f-02ea-4912-95c9-b3de3fa90f59.png)

Then using "Switch lists" to populate the list and set the selected index:

![image](https://user-images.githubusercontent.com/2104555/219223667-f89b7c6b-b136-4001-b8ce-a6f16fb95a55.png)

However, if I now switch between the lists, when I switch to the list that _would_ have enough space to render all of its choices, it does not (One, Two and Three should be visible here):

![image](https://user-images.githubusercontent.com/2104555/219223858-0451c487-10e3-4a3d-a6d3-90e0aea3014e.png)

robob27:
I tried to work on a fix for this, but I didn't get very far by myself. Then I had an idea to ask ChatGPT about it. I fed it the full code for the List widget and described the problem. After some back and forth we ended up with this:

https://github.com/DFHack/dfhack/compare/develop...robob27:list-setselected-fix?expand=1

This _does_ appear to fix the issues described here, but... I didn't come up with it myself entirely, and there was a reason why I asked ChatGPT about it 😂. I figured I would post it and someone who is more comfortable with everything else going on the widget could maybe make use of it to make the ideal fix, assuming this _probably_ isn't exactly it.

myk002:
[...]

PR #2914, opened 2023-02-16
List:setChoices rendering fix

robob27:
Attempt at fixing the issues described in #2912 

WIP because I don't fully understand _everything_ going on in the widget yet so I'm unsure if this is 100% the right fix, though it does seem to resolve the issues so I think it's pretty close.


library/lua/gui/widgets.lua
```diff
@@ -1611,6 +1611,15 @@ function List:setChoices(choices, selected)
...
     self:setSelected(selected)
+
+    -- Check if page_top needs to be adjusted
+    if #self.choices - self.page_size < 0 then
+        self.page_top = 1
+    elseif self.selected <= math.floor(self.page_size / 2) then
+        self.page_top = 1
+    elseif self.selected >= #self.choices - math.floor(self.page_size / 2) then
+        self.page_top = #self.choices - self.page_size + 1
+    end
```
myk002, line 1622: could you clean this up a little? e.g. #self.choices <= self.page_size is more natural (and I think the <= rather than just < is warranted here)

library/lua/gui/widgets.lua
```diff
@@ -1651,10 +1660,26 @@ local function update_list_scrollbar(list)
 end
 
 function List:postComputeFrame(body)
-    self.page_size = math.max(1, math.floor(body.height / self.row_height))
-    if #self.choices - self.page_size < 0 then
+    local row_count = math.floor(body.height / self.row_height)
```
myk002, line 1663: prefer the `//` operator instead of flooring

myk002 (approved):

Issue #2916, opened 2023-02-16
'tailor' module continually requests new job for toads


Stealthii:
I'm utilizing the tailor module to keep up with tattered clothing demands in my fort. Recently, I've been seeing the module request new silk trousers sized for toads, despite my stockpile now containing an innumerable number of toad sized trousers.
```
tailor: confiscated rope reed dress from Imush äkilmonom.
tailor: added order #1448 for 1 silk trousers, sized for toads
tailor: civilization cannot make skirts, skipped
tailor: ordered 1 items of clothing
Fixed priority of manager orders.
tailor: confiscated kenaf cloak from Sazir Uzolzulban.
tailor: added order #1449 for 1 silk trousers, sized for toads
tailor: civilization cannot make skirts, skipped
tailor: ordered 1 items of clothing
Fixed priority of manager orders.
tailor: added order #1450 for 1 silk trousers, sized for toads
tailor: civilization cannot make skirts, skipped
tailor: ordered 1 items of clothing
Fixed priority of manager orders.
tailor: added order #1451 for 1 silk trousers, sized for toads
tailor: civilization cannot make skirts, skipped
tailor: ordered 1 items of clothing
Fixed priority of manager orders.
tailor: added order #1452 for 1 silk trousers, sized for toads
tailor: civilization cannot make skirts, skipped
tailor: ordered 1 items of clothing
Fixed priority of manager orders.
```
![image](https://user-images.githubusercontent.com/998920/219454771-3099aaf0-5bf2-4a9b-898a-442c8dc80104.png)

There are no toads or toad men in my fortress, the closest thing I have is a single green tree frog, but no-one owns her as a pet or has tried to put little trousers on her yet, so I'm not sure where this order is being generated from.

ab9rf:
I've added a test to prevent attempting to make clothing for a bodysize that isn't mapped to a race. Toads are race 0 in the default raws, which means an unmapped body size gets mapped to 0 (toads) by default. The code will now check to see if there is a known mapping for a given body size and refuse to order anything if the size is unmapped.

The diagnostic generated when this happens will identify the size that was not mapped correctly, which may help to identify why this is happening in the first place.

[...]

PR #2917, opened 2023-02-16
tailor: try to squash toad clothing bug

ab9rf:
this adds a test for an unmapped clothing size which will at least stop the making of toad-sized clothing -see #2916. a diagnostic is issued when this happens as it is a bug

### DIFF
--- a/plugins/tailor.cpp
+++ b/plugins/tailor.cpp
@@ -56,6 +56,7 @@ enum ConfigValues {
     CONFIG_CLOTH_IDX = 2,
     CONFIG_YARN_IDX = 3,
     CONFIG_LEATHER_IDX = 4,
+    CONFIG_ADAMANTINE_IDX = 5,
 };
 
 static int get_config_val(PersistentDataItem &c, int index) {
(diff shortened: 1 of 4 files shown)


myk002 (commented):
need to add adamantine to:
- `set_material_order()`
- `tailor_setMaterialPreferences()`
- `setMaterials()` (in `plugins/lua/tailor.lua`)
- docs (in `docs/plugins/tailor.rst`)

plugins/tailor.cpp
```diff
@@ -698,6 +712,7 @@ static void tailor_setMaterialPreferences(color_ostream &out, int32_t silkIdx,
     set_config_val(config, CONFIG_CLOTH_IDX, clothIdx - 1);
     set_config_val(config, CONFIG_YARN_IDX, yarnIdx - 1);
     set_config_val(config, CONFIG_LEATHER_IDX, leatherIdx - 1);
+    set_config_val(config, CONFIG_ADAMANTINE_IDX, leatherIdx - 1);
```
myk002, line 715:
```suggestion
    set_config_val(config, CONFIG_ADAMANTINE_IDX, adamantineIdx - 1);
```

myk002: function needs a new `int32_t adamantineIdx` param.

ab9rf: this is what i get for coding while still on my first cup of coffee

myk002 (approved):

PR #2925, opened 2023-02-17
Fix confirm performance

robob27:
When I updated `confirm` I made it so `df::global::gps->force_full_display_count = 1;` was running every frame ><.

This fixes that by adding a `dirty` field to `confirmation_base`. It  defaults to `false`, gets set to `true` during `set_state`, and if `true` we will `df::global::gps->force_full_display_count = 1` at the end of `render()` and set it back to `false`.

I am unable to actually replicate the fps issues on my machine, but this seems like the culprit. Playtested with a breakpoint to confirm it is only executed after a state change. Playtested without the breakpoint and confirmed that the dialogs are still getting cleaned up properly.

While I was at it I moved some other variables into the only branch that uses them, that were getting initialized every frame unnecessarily.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -43,7 +43,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 -@ `autochop`: generate default names for burrows with no assigned names
 - ``Buildings::StockpileIterator``: check for stockpile items on block boundary.
 - `tailor`: block making clothing sized for toads; make replacement clothing orders use the size of the wearer, not the size of the garment; add support for adamantine cloth (off by default); improve logging
-- `confirm`: fix fps drop when enabled
+-@ `confirm`: fix fps drop when enabled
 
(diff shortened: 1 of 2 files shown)


docs/changelog.txt
```diff
@@ -43,6 +43,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 -@ `autochop`: generate default names for burrows with no assigned names
 - ``Buildings::StockpileIterator``: check for stockpile items on block boundary.
 - `tailor`: block making clothing sized for toads; make replacement clothing orders use the size of the wearer, not the size of the garment; add support for adamantine cloth (off by default); improve logging
+- `confirm`: fix fps drop when enabled
```
myk002, line 46:
```suggestion
-@ `confirm`: fix fps drop when enabled
```

myk002 (approved):

Issue #2921, opened 2023-02-17
Channel Safely Is Crashing on launch.


RyushoYosei:
Just as said, I  open the GUI for enabling things such as Auto butcher, auto farm and similar, if I click Channel Safely, Game instantly CTD's.

myk002:
could you upload your savegame and installed mods to a cloud service (or [dffd](https://dffd.bay12games.com/)) and attach a link here so we can investigate?

RyushoYosei:
I would but I think it was a   messed up install of DFHACK on my end, as I had re-installed it and..now it's working fine, So it might of just been an issue on my end.

lethosor:
Are you the same person who reported this on Discord a couple days ago (nickname starting with "N")? Want to make sure we aren't losing track of this issue if it was someone else.

RyushoYosei:
No, actually, I am not in the DFHack discord.

lethosor:
Ok, reopening until we can track this down

myk002:
@cppcooper FYI, though not a lot to go on here

cppcooper:
Ya no kidding. Without some kind of crash logs.
I could try to integrate some signal handling and stacktracing like discussed on discord, but won't help unless it happens to somebody else in the future.

RyushoYosei:
I had it happen a bit ago, but it's bizzare, It's inconsistant it seems, Though I mostly have encountered it running mods for sure, since I have been running mods, ..Every time i have noticed it I have been using like Playable animal people, but I suppose it would be best to see if I can get it to happen while using no mods.

myk002:
reproducible crash with this save: https://drive.google.com/file/d/1_S2Yvy99IT8xC5zQW7DxVQ3cDygz_fJV/view?usp=share_link
mods: https://drive.google.com/file/d/1okCd4khCGCHLuwxb_JNs7LpRfVvWwMwj/view?usp=share_link

steps:
1) load game
2) enable channel-safely (via gui/control-panel, but it shouldn't matter how)

myk002:
Fixed in #2945

PR #2940, opened 2023-02-21
Integrates signal handling to output a stacktrace before crashing

cppcooper:
Debugging user crashes can be difficult, this will help with that.

Needs to be tested on each platform as it becomes possible. I believe this code should all be signal safe and (less sure) async safe as well, but could really use a review by an expert.

From my understanding each thread will have its signals handled in that thread's call stack and needs to have the handler registered. In other words, from some of what I've read it seems as though running the install handlers function once in the init won't be sufficient for additional threads. Each thread will need to register the handler function separately.

To help test I added a plugin with a command to send/create signal conditions. Sigint doesn't really have a good way to be produce so it's just a simple raise(sigint) but the others create signals with intentionally bad code and abort/terminate functions.


plugins/devel/signals.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace DFHack {
+    DBG_DECLARE(signals, log, DebugCategory::LERROR);
+}
+
+static command_result send_signal(color_ostream &out, vector<string> &parameters);
+
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector <PluginCommand> &commands) {
+    DEBUG(log, out).print("initializing %s\n", plugin_name);
+
+    commands.emplace_back(plugin_name, "", send_signal);
```
myk002, line 27: I'll let others review the actual code, but I'll just add that since this doesn't have a help file, this is the only place where you can add command docs, and it should have some sort of docs to explain usage.

cppcooper: You're saying to add the help/usage string?

myk002: yeah, in lieu of a full .rst file. that's what the other devel plugins do.

myk002:
what does this PR print out for the crash in #2921 ?

cppcooper:
> what does this PR print out for the crash in #2921 ?

Nothing, I presume it's a separate thread from the one Core::Init executes on.

cppcooper:
So, I (re-)discovered the library: https://github.com/bombela/backward-cpp

Which claims to do what I was attempting, with a single header. Since I don't have high hopes for this PR's implementation I'll likely give that library a shot down the road.

myk002:
Should we close this PR, then?

cppcooper:
> Should we close this PR, then?

Probably, I put it into draft thinking I'd see about integrating that library relatively soon. But I haven't been active in dfhack for weeks (months?)

myk002:
Ok, given the merge conflicts and change of contents, I'll close this one. We can open a new one if/when you look into backward-cpp more

PR #2941, opened 2023-02-22
Updating c make lists to vs1935

PopnROFL:
Updating CMakeLists to the VS version I installed today.
Also changing the error message so others won't go down "crazy rabbit holes" like I did, trying to figure out what was going wrong!


CMakeLists.txt
```diff
@@ -61,8 +61,10 @@ if(UNIX)
 endif()
 
 if(WIN32)
-    if((NOT MSVC) OR (MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1934))
-        message(SEND_ERROR "MSVC 2022 is required")
+    if(NOT MSVC)
+        message(SEND_ERROR "No MSVC found! MSVC 2022 version 1930 to 1935 is required.")
+    elseif((MSVC_VERSION LESS 1930) OR (MSVC_VERSION GREATER 1935))
+        message(SEND_ERROR "MSVC 2022 version 1930 to 1935 is required, Version Found: ${MSVC_VERSION}")
```
ab9rf, line 67: good improvement

ab9rf (approved):
I have confirmed that MSVC version 17.5 (which dropped on February 21, 2023) does indeed report as MSVC_VER 1935. I've reviewed MSFT's release notes and do not believe that any of the changes will impact us. The changes in `std` support (which is what mandated the version bump as far as I can tell) relate mostly or entirely to `std::atomic`, which neither DF nor DFHack uses, at least as far as I know.

PR #2945, opened 2023-02-24
Fix out-of-bounds error in Channel safely

cppcooper:


plugins/channel-safely/channel-groups.cpp
```diff
@@ -21,24 +21,27 @@ void ChannelJobs::load_channel_jobs() {
...
-    auto ttype = *Maps::getTileType(p);
-    if (!DFHack::isOpenTerrain(ttype)) {
-        // check shared neighbour for cave-in conditions
-        df::coord neighbours[4];
-        get_connected_neighbours(map_pos, neighbours);
-        int connectedness = 4;
-        for (auto n: neighbours) {
-            if (active.count(n) || DFHack::isOpenTerrain(*Maps::getTileType(n))) {
-                connectedness--;
+    if likely(Maps::isValidTilePos(map_pos)) {
```
myk002, line 24: in the future, you can avoid indenting the entire body of a function by inverting the condition and returning early.

myk002 (approved):

plugins/channel-safely/channel-groups.cpp
cppcooper: I chose against that because of the extra returns, but I suppose it's more for the diff?

myk002: it's for readability and simplicity of the code. every additional indent is an indent that the reader has to follow and remember. small blocks with return statements provide a visual guarantee of "below this line, this is true". it's fine to have multiple returns if each return is filtering out an additional class of bad input.

cppcooper: It's strange cause I find the opposite problematic [sometimes], but especially in these sorts of functions. The flow control sort of melds in with the code and I forget it's there, something I suppose I'll need to train myself to look for [even] in simple functions.

PR #2953, opened 2023-02-25
clean up code for C++20 readiness

ab9rf:
two change:

* remove use of `register` in `md5.cpp`
* remove use of `using namespace std` in `Core.cpp` (which causes an ambiguous name resolution error between `byte` and `std::byte`). while there are other ways to resolve this, `using namespace std` is a code smell anyway, so eliminating it is the best option

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1648,7 +1648,7 @@ bool Core::Init()
     }
     std::cerr << "Initializing textures.\n";
     Textures::init(con);
-    // create std::mutex for syncing with interactive tasks
+    // create mutex for syncing with interactive tasks
     std::cerr << "Initializing plugins.\n";
     // create plugin manager
     plug_mgr = new PluginManager(this);
@@ -1883,7 +1883,7 @@ void Core::doUpdate(color_ostream &out)
         if(isMapLoaded())
             onStateChange(out, SC_MAP_LOADED);
     }
-    // otherwise just check for std::map change...
+    // otherwise just check for map change...
     else if (new_mapdata != last_local_map_ptr)
     {
         bool had_map = isMapLoaded();
@@ -2029,7 +2029,7 @@ namespace DFHack {
                     const char *v = va_arg(list, const char *);
                     if (!v || !v[0])
                         break;
-                    val.push_back(std::string(v));
+                    val.emplace_back(v);
                 }
                 result.push_back(Entry(key,val));
             }
@@ -2164,7 +2164,7 @@ void Core::onStateChange(color_ostream &out, state_change_event event)
     if (event == SC_WORLD_LOADED && Version::is_prerelease())
     {
         runCommand(out, "gui/prerelease-warning");
-        std::cerr << "loaded std::map in prerelease build" << std::endl;
+        std::cerr << "loaded map in prerelease build" << std::endl;
     }
 
     if (event == SC_WORLD_LOADED)
@@ -2424,7 +2424,7 @@ bool Core::SelectHotkey(int sym, int modifiers)
             if (!binding.focus.empty()) {
                 if (!Gui::matchFocusString(binding.focus)) {
                     std::vector<std::string> focusStrings = Gui::getCurFocus(true);
-                    DEBUG(keybinding).print("skipping keybinding due to focus std::string mismatch: '%s' !~ '%s'\n",
+                    DEBUG(keybinding).print("skipping keybinding due to focus string mismatch: '%s' !~ '%s'\n",
                         join_strings(", ", focusStrings).c_str(), binding.focus.c_str());
                     continue;
                 }
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1870,15 +1870,15 @@ void Core::doUpdate(color_ostream &out)
     // if the world changes
     if (new_wdata != last_world_data_ptr)
     {
-        // we check for std::map change too
+        // we check for map change too
         bool had_map = isMapLoaded();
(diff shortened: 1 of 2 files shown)


library/Core.cpp
```diff
@@ -2165,7 +2164,7 @@ void Core::onStateChange(color_ostream &out, state_change_event event)
     if (event == SC_WORLD_LOADED && Version::is_prerelease())
     {
         runCommand(out, "gui/prerelease-warning");
-        std::cerr << "loaded map in prerelease build" << std::endl;
+        std::cerr << "loaded std::map in prerelease build" << std::endl;
```
myk002, line 2167:
```suggestion
        std::cerr << "loaded map in prerelease build" << std::endl;
```

library/Core.cpp
```diff
@@ -2425,7 +2424,7 @@ bool Core::SelectHotkey(int sym, int modifiers)
             if (!binding.focus.empty()) {
                 if (!Gui::matchFocusString(binding.focus)) {
                     std::vector<std::string> focusStrings = Gui::getCurFocus(true);
-                    DEBUG(keybinding).print("skipping keybinding due to focus string mismatch: '%s' !~ '%s'\n",
+                    DEBUG(keybinding).print("skipping keybinding due to focus std::string mismatch: '%s' !~ '%s'\n",
```
myk002, line 2427:
```suggestion
                    DEBUG(keybinding).print("skipping keybinding due to focus string mismatch: '%s' !~ '%s'\n",
```

library/Core.cpp
```diff
@@ -2030,7 +2029,7 @@ namespace DFHack {
                     const char *v = va_arg(list, const char *);
                     if (!v || !v[0])
                         break;
-                    val.push_back(string(v));
+                    val.push_back(std::string(v));
```
myk002, line 2032:
```suggestion
                    val.emplace_back(v);
```

library/Core.cpp
```diff
@@ -1871,20 +1870,20 @@ void Core::doUpdate(color_ostream &out)
...
         if(had_map)
             onStateChange(out, SC_MAP_UNLOADED);
-        // and if the world is appearing, we report map change after that
+        // and if the world is appearing, we report std::map change after that
         onStateChange(out, new_wdata ? SC_WORLD_LOADED : SC_WORLD_UNLOADED);
         if(isMapLoaded())
             onStateChange(out, SC_MAP_LOADED);
     }
-    // otherwise just check for map change...
+    // otherwise just check for std::map change...
```
myk002, line 1886:
```suggestion
    // otherwise just check for map change...
```

library/Core.cpp
```diff
@@ -1871,20 +1870,20 @@ void Core::doUpdate(color_ostream &out)
...
         bool had_map = isMapLoaded();
         last_world_data_ptr = new_wdata;
         last_local_map_ptr = new_mapdata;
 
-        // and if the world is going away, we report the map change first
+        // and if the world is going away, we report the std::map change first
         if(had_map)
             onStateChange(out, SC_MAP_UNLOADED);
-        // and if the world is appearing, we report map change after that
+        // and if the world is appearing, we report std::map change after that
```
myk002, line 1881:
```suggestion
        // and if the world is appearing, we report map change after that
```

library/Core.cpp
```diff
@@ -1871,20 +1870,20 @@ void Core::doUpdate(color_ostream &out)
...
     if (new_wdata != last_world_data_ptr)
     {
-        // we check for map change too
+        // we check for std::map change too
         bool had_map = isMapLoaded();
         last_world_data_ptr = new_wdata;
         last_local_map_ptr = new_mapdata;
 
-        // and if the world is going away, we report the map change first
+        // and if the world is going away, we report the std::map change first
```
myk002, line 1878:
```suggestion
        // and if the world is going away, we report the map change first
```

library/Core.cpp
```diff
@@ -1871,20 +1870,20 @@ void Core::doUpdate(color_ostream &out)
     // if the world changes
     if (new_wdata != last_world_data_ptr)
     {
-        // we check for map change too
+        // we check for std::map change too
```
myk002, line 1873:
```suggestion
        // we check for map change too
```

library/Core.cpp
```diff
@@ -1642,27 +1641,27 @@ bool Core::Init()
...
     if (!DFSDL::init(con)) {
         fatal("cannot bind SDL libraries");
         return false;
     }
-    cerr << "Initializing textures.\n";
+    std::cerr << "Initializing textures.\n";
     Textures::init(con);
-    // create mutex for syncing with interactive tasks
-    cerr << "Initializing plugins.\n";
+    // create std::mutex for syncing with interactive tasks
```
myk002, line 1651:
```suggestion
    // create mutex for syncing with interactive tasks
```

library/Core.cpp
```diff
@@ -35,7 +35,6 @@ distribution.
 #include <forward_list>
 #include <type_traits>
 #include <cstdarg>
-using namespace std;
```
myk002, line 38: while I get that `using namespace std;` is bad, I wouldn't have complained if we kept `using std::string` and `using std::vector`. That's pretty commonplace in our code, and I don't see anything wrong with it.

ab9rf: i did consider doing that, but have mixed feelings here

myk002:
still some unaddressed comments

myk002 (approved):

Issue #2977, opened 2023-03-06
[stockpiles] savestock errors when targeting a refuse pile


TSM-EVO:
Steps to reproduce:

Create a stockpile, set it to 'refuse'
run `savestock refuse`

The following error is returned, the file `refuse.dfstock` is created but empty. 

`[DFHack]# savestock refuse
serialization failed: protobuf exception: invalid vector subscript`

TSM-EVO:
The error happens when you select any of the subcategories of refuse as well.

It also happens when you select a different, working category, like cloth, and then select a refuse category alongside it, same error.

TSM-EVO:
I reproduced this in a clean world with no mods, I'll upload it on the discord

Issue #2974, opened 2023-03-05
[quickfort] Stockpile and zone support


myk002:
The most important missing feature of `quickfort` today is stockpile placement and configuration. Placement is done by `place` mode, and should not be too difficult to update on its own. However, stockpile configuration is done in `query` blueprints, and is a greater challenge.

`query` and `config` blueprints are "freeform keystroke repeaters" -- you could have them reproduce any key sequence. This was fine when everything could be done with the keyboard. In v50, it is a great liability.

I propose we deprecate `query` and `config` modes entirely (along with keystroke aliases in general) and move stockpile configuration into `place` mode, introducing a DSL for configuration. This approach has already been proven to work with the existing `zone` blueprint format, where you can set zone properties as you declare the zones.

All existing logic in `query` blueprints would need to be migrated, but given the change in the operating environment, I don't think this is avoidable. We can keep compatibility with existing `place` blueprints, though, with simple "all or nothing" category enablement.

The DSL could be based on the current (successful) keystroke alias library, using a similar vocabulary. e.g.:
```
f{plants}(10x10)
f{forbid plants; forbid tallow; forbid dye}(4x7)
pd{bronzeweapons; permit steelweapons, forbid masterworkweapons; forbid artifactweapons}
```

or more formally:
```
<stockpile-config> ::= "{" <commands> "}"
<commands> ::= <command> | <command> ";" <commands>
<command> ::= <name> | "permit" <name> | "forbid" <name>
```
where `<name>` is the name of a saved stockpile config that the `stockpile` plugin can apply to the given stockpile. `stockpile` will need to be extended to add ("permit") or subtract ("forbid") stockpile settings in addition to just overwriting the config (when the `<name>` is given without a command prefix).

We would also need to *create* all these library stockpile configs and make them accessible.

myk002:
Project plan

- [ ] implement "permit" and "forbid" `stockpile` commands
- [ ] implement reading library stockpile configs from a different directory (same structure as `orders`)
- [ ] create stockpile config library
- [ ] parse new `place` text syntax and use `stockpiles` to apply
- [ ] convert dreamfort blueprints to use new syntax
- [ ] remove dreamfort query and config blueprints (keep unmigrated functionality (like QSP logic) in "ignore" blueprints)
- [ ] update dreamfort walkthrough
- [ ] remove query and config modes from quickfort
- [ ] update quickfort blueprint creation guide with new syntax
[...]

PR #2978, opened 2023-03-06
[quickfort] new alias structure

TSM-EVO:
Files created with `savestock` for each existing [alias](https://docs.dfhack.org/en/stable/docs/guides/quickfort-alias-guide.html). Some aliases missing (search as an example) currently.


TSM-EVO:
Refuse piles are also not currently represented correctly, due to #2977

myk002:
let me see if I can get the refuse exporting fixed before we merge this PR

myk002:
I actually can't reproduce the specific error here. could I get a copy of the world you're using?

myk002:
Refuse pile errors should be fixed now that #3005 is merged. Could you try those again?

myk002:
mentioning #2974 to link this PR to that issue

TSM-EVO:
The refuse files are now nonzero and working,

myk002 (approved):
Thanks you for creating these!

PR #2983, opened 2023-03-06
update all build envs to ubuntu-22.04

myk002:
since ubuntu-18.04 is apparently becoming verboten


myk002:
merging to unbreak build actions. we can adjust more later once we find out what our *real* compatibility range should be

PR #3008, opened 2023-03-11
support scripts in mods

### LINKED ISSUE #2667: Define how mods with scripts should be handled
There are two types of scripts that mods can come with:

1. scripts that should be generally available regardless of whether the mod is "active" for a world
2. scripts that depend on changes to the raws and should only be accessible if the mod is active

I think that we should define a directory format for mod authors to follow so we can handle them appropriately. A mod can come with both types of scripts, so it stands to reason that they could be identified by directory name. For example:

- `general_scripts/` could hold scripts that should always be in the script path, regardless of whether the mod is active
- `active_scripts/` could hold scripts that should only be in the script path if the mod is active

at dfhack init, we can scan for `general_scripts/` directories and add them all to the script paths, maybe *after* `hack/scripts` so mods can't hijack a standard script, accidentally or intentionally

on world load, we can check the active mod list and append relevant `active_scripts/` directories to the script paths

We should decide on a standard soon to reduce the mounting confusion about how this should be handled.

myk002:
scripts in downloaded mods are now automatically added to the DFHack script path. DFHack recognizes two directories in a mod's folder: `scripts_modinstalled/` and `scripts_modactive/`. `scripts_modinstalled/` folders will always be added to the script path, regardless of whether the mod is active in a world. `scripts_modactive/` folders will only be added to the script path when the mod is active in the current loaded world.

Fixes #2667

side note: if a mod adds a script in its `scripts_modactive` folder, and that script attaches itself to `dfhack.onStateChange`, then the script must remove its handler when a world is unloaded or else risk being called in a world where the mod is not active.

docs: 

Script directories in installed mods are automatically added to the script path
according to the following rules:

**If a world is not loaded**, then directories matching the pattern
``data/installed_mods/*/scripts_modinstalled/`` are added to the script path
in alphabetical order of the mod name.

**If a world is loaded**, then the ``scripts_modactive`` directories of active
mods are also added to the script path according to the active mod load order,
and scripts in active mods take precedence over scripts in
``scripts_modinstalled`` in non-active mods. For example, the search paths for
mods might look like this::

    activemod1/scripts_modactive
    activemod1/scripts_modinstalled
    activemod2/scripts_modactive
    activemod2/scripts_modinstalled
    inactivemod1/scripts_modinstalled
    inactivemod2/scripts_modinstalled

Not all mods will have script directories, of course, and those mods will not be
added to the script search path. Mods are re-scanned whenever a world is loaded
or unloaded.

### DIFF
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -248,7 +248,7 @@ according to the following rules:
 
 **If a world is not loaded**, then directories matching the pattern
 ``data/installed_mods/*/scripts_modinstalled/`` are added to the script path
-in alphabetical order of the mod name.
+in alphabetical order.
 
 **If a world is loaded**, then the ``scripts_modactive`` directories of active
 mods are also added to the script path according to the active mod load order,
@@ -256,12 +256,14 @@ and scripts in active mods take precedence over scripts in
 ``scripts_modinstalled`` in non-active mods. For example, the search paths for
 mods might look like this::
 
-    activemod1/scripts_modactive
-    activemod1/scripts_modinstalled
-    activemod2/scripts_modactive
-    activemod2/scripts_modinstalled
+    activemod_last_in_load_order/scripts_modactive
+    activemod_last_in_load_order/scripts_modinstalled
+    activemod_second_to_last_in_load_order/scripts_modactive
+    activemod_second_to_last_in_load_order/scripts_modinstalled
+    ...
     inactivemod1/scripts_modinstalled
     inactivemod2/scripts_modinstalled
+    ...
 
 Not all mods will have script directories, of course, and those mods will not be
 added to the script search path. Mods are re-scanned whenever a world is loaded
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -235,9 +235,40 @@ root DF folder):
 #. :file:`dfhack-config/scripts`
 #. :file:`save/{world}/scripts` (only if a save is loaded)
 #. :file:`hack/scripts`
+#. :file:`data/installed_mods/...` (see below)
 
 For example, if ``teleport`` is run, these folders are searched in order for
 ``teleport.lua``, and the first matching file is run.
+
+Scripts in installed mods
+.........................
+
+Script directories in installed mods are automatically added to the script path
+according to the following rules:
+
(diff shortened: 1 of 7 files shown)


myk002:
I have noticed that the scanning of so many files takes about a second on my system (Linux, via wine). In doing some profiling, I discovered that the bulk of the time is spent in the `dfhack.filesystem.listdir_recursive` call (!). I was expecting the time to be spent in loading all the scripts as modules, but that only takes about 30% of the time.

The delay is just long enough to be noticeable. It may or may not be a problem on systems running native versions. The delay might go away once the Linux and Mac ports are released.

ab9rf:
> In doing some profiling, I discovered that the bulk of the time is spent in the `dfhack.filesystem.listdir_recursive` call (!).

this is probably because this function enumerates that entire tree in one go to generate a map containing _every_ file in those trees, whether or not we care about them. it would probably be better to, instead, implement this scan using an iterator, which would avoid creating such a large data structure almost all of which is not needed, and let the caller traverse the iterator as it goes

myk002:
> this is probably because this function enumerates that entire tree in one go to generate a map containing _every_ file in those trees, whether or not we care about them. it would probably be better to, instead, implement this scan using an iterator, which would avoid creating such a large data structure almost all of which is not needed, and let the caller traverse the iterator as it goes, capturing only the specific information it needs without the need to inter the rest into a large and expensive data structure

I'm not fully convinced.
```
$ find scripts/ plugins/lua library/lua/ -name '*.lua' | wc -c
12033
```
More profiling is warranted, but 12KB just doesn't seem like very much. I would have thought this operation would have completed just about instantly.

ab9rf:
`fstat` on NTFS is a slow operation, and it's possible that the enumeration being done is calling `fstat` unnecessarily. since the decision whether we're even interested in a file can be made based solely on its _name_, we don't need any of the file's other metadata.

caveat: i haven't looked at the implementation

myk002:
ah, that might explain it. it's not the listing of the files, then, but rather the check to see if they are a directory. that *does* call `stat()`. Is there a way we can avoid that?

myk002:
I opened an issue to look into avoiding `stat`/`fstat` #3097

PR #3051, opened 2023-03-20
Add default keybinding an alias for gui/design

johncosker:
Should not be merged until after https://github.com/DFHack/scripts/pull/643

Creates an `alias` for `gui/dig` -> `gui/design` to preserve existing custom user hotkeys.

### DIFF
--- a/docs/about/Removed.rst
+++ b/docs/about/Removed.rst
@@ -146,6 +146,12 @@ This script is no longer useful in current DF versions. The script required a
 binpatch <binpatches/needs-patch>`, which has not been available since DF
 0.34.11.
 
+.. _gui/dig:
+
+gui/dig
+=======
+Renamed to gui/design
+
 .. _gui/hack-wish:
 
 gui/hack-wish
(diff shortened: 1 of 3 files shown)


myk002 (commented):
One more item of business the could be added to this PR: `gui/dig` needs a tombstone entry in docs/about/Removed.rst so the old entries in the changelog have something to resolve to. Otherwise we'll get build errors on the docs.

Ironically, the tombstone entry will cause *this* PR to fail until the scripts PR is merged due to duplicate doc labels, but don't worry about that.

myk002 (approved):

docs/about/Removed.rst
```diff
@@ -192,3 +192,9 @@ Replaced with a GUI version: `gui/unit-syndromes`.
 warn-stuck-trees
 ================
 The corresponding DF :bug:`9252` was fixed in DF 0.44.01.
+
+.. gui/dig:
```
myk002, line 196: Ha, the fact that the PR didn't fail tipped me off that there was an issue here. Missing an underscore

myk002: Also please keep in alphabetical order

myk002:
Ah, I didn't see this before, could you fix this in the scripts repo?
scripts/docs/gui/design.rst:3: WARNING: Title underline too short.

johncosker:
https://github.com/DFHack/scripts/pull/644

myk002 (approved):

PR #3058, opened 2023-03-21
[stockpiles] support selective stockpile modifications

myk002:
and update the stockpile library with new formats

#2974


ab9rf (approved):
I wish there was a way to genericize at least some of the repeated code, but the only ways I can think of to do it will likely result in really hard to understand code

myk002:
yeah, the repeated pattern of "if all then do_all else go_element_by_element" would have been nice to extract, but there was enough variation in required state that it kept me from attempting it

PR #3076, opened 2023-03-23
allow metal chests to be melted

### LINKED ISSUE #2982: Automelt doesn't designate steel chests
Using a stockpile for meltable steel furniture - all items seem to get designated for melting properly via automelt except for steel chests, which clutter up the stockpile over time.

myk002:
Fixes #2982


ab9rf:
There used to be a bug in DF (see https://www.bay12games.com/dwarves/mantisbt/view.php?id=2493) that made chests unmeltable. This was resolved in v0.40.05, which was _after_ the initial version of `automelt`.

PR #3080, opened 2023-03-23
[buildingplan] actually record the deserialized material mask val

### LINKED ISSUE #3044: Buildingplan does not remember planned filters on save/load
[Version: Build 4529](https://github.com/DFHack/dfhack/actions/runs/4452039003)

Steps to reproduce:
Have multiple kinds of blocks available
Plan a bunch of constructions at once for ONE specific material
Save and quit (i closed the game entirely), load back in
Dwarves will build with any blocks

Expected behavior:
Dwarves will build only with the one intended material

### LINKED ISSUE #3065: Bug: Buildingplan forgets material/quality settings on reload
Saved twice with planned walls, floors, cage traps, cabinets and chests (and I assume other things will be affected) each with a single specific material chosen, when loading each time the planned furniture/constructions would be built with whatever was available. One load between the two bugged loads didn't do this, but I'm almost certain that was because the planned squares had all already been assigned (lost their P sign) when I saved.

eg: this was all planned pink:
image

myk002:
member field was passed by value instead of reference, causing the deserialized state to just be lost to the aether instead of remembered in the ItemFilter

Fixes #3044 
Fixes #3065

### DIFF
--- a/plugins/buildingplan/itemfilter.cpp
+++ b/plugins/buildingplan/itemfilter.cpp
@@ -35,7 +35,7 @@ bool ItemFilter::isEmpty() const {
             && materials.empty();
 }
 
-static bool deserializeMaterialMask(string ser, df::dfhack_material_category mat_mask) {
+static bool deserializeMaterialMask(const string& ser, df::dfhack_material_category& mat_mask) {
     if (ser.empty())
         return true;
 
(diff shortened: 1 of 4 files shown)


plugins/buildingplan/itemfilter.cpp
```diff
@@ -35,7 +35,7 @@ bool ItemFilter::isEmpty() const {
             && materials.empty();
 }
 
-static bool deserializeMaterialMask(string ser, df::dfhack_material_category mat_mask) {
+static bool deserializeMaterialMask(const string& ser, df::dfhack_material_category& mat_mask) {
```
myk002, line 38: all other changes in this PR are incidental. This is the functionally different line.

ab9rf (approved):

PR #3091, opened 2023-03-25
reenable getplants

ab9rf:
not clear why this wasn't enabled previously


myk002:
I seem to recall a report of "getplants not getting plants", but I don't see it in our issue history.. Did you test that it worked as advertized?

ab9rf:
i did brief tests on my "garbage" save and it did mark plants for harvest as expected

there were some issues with getplants last year, but i believe they were all resolved

PR #3094, opened 2023-03-25
autoclothing: support disable

ab9rf:
this also includes some minor refactorings:
* some debug messages have been "tagged" so it's obvious they're from autoclothing
* a serialized list of flag checks has been replaced with a maskcheck
* partial lua api added, currently only supports enable/disable


ab9rf:
pre-commit.ci autofix

plugins/autoclothing.cpp
```diff
@@ -523,7 +565,7 @@ static void find_needed_clothing_items()
 
                 if (!item)
                 {
-                    WARN(cycle).print("Invalid inventory item ID: %d\n", ownedItem);
+                    WARN(cycle).print("autoclothing: Invalid inventory item ID: %d\n", ownedItem);
```
myk002, line 568: `debugfilter` has the capability to turn on source plugin labels in debug messages, but it's a configurable header element. What if, instead of tagging here (and below), we make that header element always on for WARN and above?

https://github.com/DFHack/dfhack/blob/develop/library/Debug.cpp#L152

ab9rf: i have mixed feelings about this idea, it'll force edits to a number of plugins that already tag all their messages to avoid doubletagging

myk002: we can keep it as a follow-up issue, then. I think plugins tagging their messages manually is redundant, but it doesn't have to be solved in this PR

PR #3101, opened 2023-03-25
remove lua's ability to load binary chunks

### LINKED ISSUE #3070: disable ability for Lua to run compiled scripts
to inhibit malware distributed as 3rd party scripts

myk002:
to protect against malware attacks

Fixes #3070

### DIFF
--- a/depends/lua/src/ldo.c
+++ b/depends/lua/src/ldo.c
@@ -767,14 +767,14 @@ static void f_parser (lua_State *L, void *ud) {
   LClosure *cl;
   struct SParser *p = cast(struct SParser *, ud);
   int c = zgetc(p->z);  /* read first character */
-  if (c == LUA_SIGNATURE[0]) {
-    checkmode(L, p->mode, "binary");
-    cl = luaU_undump(L, p->z, p->name);
-  }
-  else {
+  // if (c == LUA_SIGNATURE[0]) {
+  //   checkmode(L, p->mode, "binary");
+  //   cl = luaU_undump(L, p->z, p->name);
+  // }
+  // else {
     checkmode(L, p->mode, "text");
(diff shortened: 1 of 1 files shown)


myk002:
we could also completely remove `luaU_undump` from the library, but that might cause conflicts if we ever want to upgrade Lua and merge our changes back in

ab9rf (approved):

depends/lua/src/ldo.c
```diff
@@ -767,14 +767,14 @@ static void f_parser (lua_State *L, void *ud) {
   LClosure *cl;
   struct SParser *p = cast(struct SParser *, ud);
   int c = zgetc(p->z);  /* read first character */
-  if (c == LUA_SIGNATURE[0]) {
-    checkmode(L, p->mode, "binary");
-    cl = luaU_undump(L, p->z, p->name);
-  }
-  else {
+  // if (c == LUA_SIGNATURE[0]) {
```
lethosor, line 770: Generally when modifying upstream dependencies that we may upgrade in the future, I would prefer to modify as few lines as possible (kind of how we changed CMakeLists for the libzip/libexpat family of dependencies, although I wouldn't call that a "few" lines...)

Maybe something like
```suggestion
  if (0) { // disabled by DFHack
```

but I think our changes to Lua, including this one, are small enough that I'm not very concerned about maintanability here.

lethosor:
Good with this change in general - we may want to allow it to linger in develop a bit just in case it happens to break something internally (which I think is unlikely)

myk002:
Let me promote the current dev build a bit on Discord. There were many changes today, and I'd like to see them go through some beta testing as well.

If we can get some usage from Discord friends and no reports of breakages, I'm planning on another Friday release (6 days from now)

PR #3123, opened 2023-03-29
WIP compound buildingplan ui tweaks (dont merge!)

TaxiService:
needed for buildingplan's various ui's


TaxiService:
like, this is marked as WIP, but up to the last commit this could have been merged.

what i've done so far is:
- adding two new borders, BOLD_FRAME and MEDIUM_FRAME_2
-- the first is a chonky border for when you need a panel emphasized,
-- the second is a variant of MEDIUM_FRAME without the `DFHack` text at the bottom. (it's used in planneroverlay)
- rewrote itemselection's ui to look like this:
image
- started reworking planneroverlay
-- applied the new borders to it (although you can't tell it's MEDIUM_FRAME_2 because i forgot about moving the error panel downwards)
-- started redoing the minimize-button, but right now it acts as two separate buttons instead of a single big button (https://i.imgur.com/dXi8V9C.mp4)

docs/dev/Lua API.rst
```diff
@@ -4352,6 +4352,15 @@ There are the following predefined frame style tables:
...
   A frame suitable for overlay widget panels.
 
+* ``MEDIUM_FRAME_2``
+
+  This version of MEDIUM_FRAME does *not* have a visible ``DFHack`` signature on it,
+  so it must not be used as the most external frame for a DFHack-owned UI.
+
+* ``BOLD_FRAME``
+
+  A frame suitable for overlay widget panels.
```
myk002, line 4362: `BOLD_FRAME` may have a good place in the frame lineup, but I've specifically been using `MEDIUM_FRAME` for overlay panels. Overlay panels are often in crowded spaces where thick borders make the layout too crowded. The buildingplan overlay is special since it's on its own without vanilla elements around it, but even there why not `PANEL_FRAME`?

library/lua/gui.lua
```diff
@@ -916,6 +916,9 @@ end
 WINDOW_FRAME = make_frame('Window', true)
 PANEL_FRAME = make_frame('Panel', false)
 MEDIUM_FRAME = make_frame('Medium', false)
+MEDIUM_FRAME_2 = make_frame('Medium', false) --this is a panel style used in buildingplan's planner overlay
```
myk002, line 919:
```suggestion
MEDIUM_FRAME_2 = copyall(MEDIUM_FRAME)
```

docs/dev/Lua API.rst
```diff
@@ -4352,6 +4352,15 @@ There are the following predefined frame style tables:
 
   A frame suitable for overlay widget panels.
 
+* ``MEDIUM_FRAME_2``
```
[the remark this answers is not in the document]
TaxiService: yes great idea! As you can tell by variable names like MINI_BUTT_PEN, my coming-up-with-names skill is not super well-honed 🤣

docs/dev/Lua API.rst
TaxiService: honestly-- i just copied `MEDIUM_FRAME`'s description because i didn't know what to write. :flushed:

what i meant to write was something along the lines of: 
"use this when you need your panel to have a bigger impact on screen", or
"a frame suitable for _bold_ panels" in the sense that it's an important panel/one you should notice.

i didn't know how to word that and i thought it'd be best to discuss it.

as for why not `PANEL_FRAME`... i think that in planneroverlay's case it complements vanilla's UI the best.
Capture1 Capture2

`PANEL_FRAME`'s square angles and its slightly lighter shade of yellow feel a bit out of place, imo.

i feel that, generally, a `PANEL_FRAME` border should encompass the whole contents of a window. :thinking: 
Since part of this window is already using a `MEDIUM_FRAME` external border, having a panel using a `PANEL_FRAME` border tacked under it feels a bit jarring. 
In this case `BOLD_FRAME` looks more, like, "organic", less "tacked-on" somehow. _And_ in this case also helps visually separating its contents from the controls you can directly interact with.

I agree that `BOLD_FRAME` could be considered redundant in other places... But maybe we can find another use for it for windows like `confirm`'s confirmation dialogs. 🤔

library/lua/gui.lua
TaxiService: that's sick! totes gonna add that in as i commit the `INTERIOR_MEDIUM_FRAME` change

docs/dev/Lua API.rst
myk002: you're talking above my pay grade here. I trust you on the visual style decisions.

myk002:
image
most players will already have the previous default_pos saved, so their panel will look like the screenshot above. Moreover, there is text that appears on the line just below the building name (though I forget now what the circumstances are where it appears. it's pretty rare). You could use the existing overlay position and have a blank panel that blacks out the area to the left of the minimize button.

myk002:
image
I'm guessing the item description isn't supposed to be clipped like that?

myk002:
image
Block of what? This dialog is keeping me in suspense!

myk002:
I just noticed that the `[x]` is still visible when there is no filter, it's just very hard to see. perhaps we should hide it entirely when there's no filter set?

TaxiService:
>I'm guessing the item description isn't supposed to be clipped like that?
>[...]
>Block of what? This dialog is keeping me in suspense!
>[...]
>I just noticed that the `[x]` is still visible when there is no filter, it's just very hard to see. perhaps we should hide it entirely when there's no filter set?

yes, it's bad. 🤣 let me explain...

This whole pr is meant as a stepping stone towards towards the "complete"/final ui.
progression

I've tried turning the window as close as i could to the mockup, but some things i just wasn't sure how to implement. :\\
Like, i tried to make the window height change dynamically using ResizingPanels, but failed miserably. (possibly we might need to extend that class' functionality to allow for minimum window height or smth)

so, like, i did all i could without delving too deeply into trying to implement things by botching stuff, since surely someone that knew what they're doing could do the same in a better way and less time. 😳 

The behaviour of text in those item lines should be something like this:
Untitled-2
The shaded areas are different Labels, and their width/left-offsets all depend on the previous Label. 🤔 
from left to right:
- the yellow areas represent the borders, those are the extremes of our total Label cluster.
- the purple area is the amount of needed items. I guess it should stretch for bigger numbers. 🤔 
- the white area is the item name. It should stretch depending on string length, and comes after purple.
- the blue area is for the item's status. (i guess it should align itself with the white area, but we'll have to think about this)
- the cyan area is the filter. It should display what the filter is, and get cut with an ellipsis (...) if too long. 
- the red area is reserved for the red x button, which appears only when a filter is not `[any material]` (it's always 3-wide, and constrains the cyan label's max width!)

like, i haven't found a way to make those labels' starting position depend on the previous label's width. ;_; I'm sure it's somewhat easy to do, but i didn't want to commit a botched-ass solution since probably anyone else can come up with a better one. 😳 

Another missing feature is that the selected item in the list should be highlit (WHITE item name/LCYAN filter text) and the other elements shouldn't be. (LGRAY item name/CYAN filter text)

so, in the end... yes, this pr is bad, but it's not *too bad*, and is closer to what we'll eventually need to do. °-° 
(tl;dr: i cant do this alone ;_; halp)



>most players will already have the previous default_pos saved, so their panel will look like the screenshot above. 

Now, this is a problem. Can't we reset this overlay's position to default the first time a user loads the new planneroverlay? 🤔 
The minimize button really should be up there in that specific spot: Above vanilla's green button, and below vanilla's building-type text.

If, by the way, you feel that the minimize button is too big while minimized, we could possibly have it so that the ` show Planner ` text only shows if you're hovering over it, and be hidden when the mouse is not in that area.

>You could use the existing overlay position and have a blank panel that blacks out the area to the left of the minimize button.

🤔 i guess we could do that, but... then if the user wanted to move this window elsewhere, it'd always have that black bar.
like, the "cleanest" solution (ui-wise) would be to forcibly reset this panel's position to default. I don't know how to go about coding this, and it doesn't sound like a clean solution code-wise, but... ehhhh °-°

library/lua/gui.lua
```diff
@@ -916,8 +916,11 @@ end
 WINDOW_FRAME = make_frame('Window', true)
 PANEL_FRAME = make_frame('Panel', false)
 MEDIUM_FRAME = make_frame('Medium', false)
+BOLD_FRAME = make_frame('Bold', true)
```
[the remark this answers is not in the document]
TaxiService: ; ) it ain't called BOLD for no reason!

myk002:
I had an idea at one point (I'm not sure if I wrote it down anywhere) that overlay panels could have a "version" field. the version would be saved into the `dfhack-config/overlay.json` state file, and if you bumped your version, you'd get a new record in the state file (which would reset your state to the (possibly new) defaults). That would work here, but the black panel approach is simple and cheap, and also I think the correct approach, since there could be vanilla text that appears on the line where you currently have the minimize button appearing.

To implement, you'd add a `widgets.Panel{frame={t=0, l=0, r=0, h=1}, frame_background=gui.CLEAR_PEN, visible=function() return not self.minimized end}`

myk002:
This PR is a bit large to review effectively. Could you split the changes into different PRs? That would help get the easy stuff committed quickly and let us focus on specific issues in different PRs. Essentially, could you:

1. move planneroverlay changes to a new PR
2. move itemselection dialog changes to a new PR

that would leave this PR with just the border-related changes, and we can merge it and get all those files out of the way.

It can get a little tricky when you have multiple related changes in progress at once. The best workflow I've found is to have a branch with everything combined on your local computer so you can build and test everything together. When you are ready to commit something, switch to the appropriate branch(es), commit your changes, push them to update their PRs, then switch back to your local combined branch and merge the changes from the other branches. It's a bit convoluted, but keeping different changes on different branches makes them much much easier to reason about. Single large PRs tend to languish since they're just too hard to review.

myk002:
I should note that if you want other people to test the combined code, you can push your combined branch as well to start a PR that is not intended for review. I've done that frequently when I've been in this situation.

TaxiService:
> there could be vanilla text that appears on the line where you currently have the minimize button appearing.

after some painstaking research, it seems that text down there only appears while building *some* constructions that span multiple z-layers.

Dwarf_Fortress_9rFmrnAMGL

the constructions are:
- wall
- floor
- ramp
- stairs
- fortification
- track

There are other things that can be built spanning multiple z-levels in some cases, like rollers or paved roads, but the text is not shown for those.
so, like, good news! unless you go through a gogol z-layers, that button should never overlay anything!

TaxiService:
alright, i'll divide up these commits into different branches and commit them separately. makes sense!
it'll take a bit, but if we're lucky i'll have them ready by tomorrow!

TaxiService:
christ on a tandem, handling all these branches really makes me feel like a monkey!

myk002:
you get used to it : )
I'm not sure how many branches I've created, but it must be well into the hundreds

PR #3137, opened 2023-03-31
fix masterwork/exceptional/decorated symbols

TaxiService:
this seems to fix the issue in which some quality levels were displayed with incorrect characters in some DFHack windows, namely:

`@` instead of `☼`
`=` instead of `≡`
`<>` instead of `«»`

### DIFF
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -720,13 +720,13 @@ df::coord Items::getPosition(df::item *item)
     return item->pos;
 }
 
-static char quality_table[] = { 0, '-', '+', '*', '=', '@' };
+static int quality_table[] = { 0, 45, 43, 42, 240, 15 };
 
 static void addQuality(std::string &tmp, int quality)
 {
     if (quality > 0 && quality <= 5) {
-        char c = quality_table[quality];
-        tmp = c + tmp + c;
+        int c = quality_table[quality];
+        tmp = static_cast<char>(c) + tmp + static_cast<char>(c);
     }
 }
 
@@ -825,7 +825,7 @@ std::string Items::getDescription(df::item *item, int type, bool decorate)
         addQuality(tmp, item->getQuality());
 
         if (item->isImproved()) {
-            tmp = "<" + tmp + ">";
+            tmp = static_cast<char>(174) + tmp + static_cast<char>(175);
             addQuality(tmp, item->getImprovementQuality());
         }
     }
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -720,12 +720,7 @@ df::coord Items::getPosition(df::item *item)
     return item->pos;
 }
 
-static const char MARKER_EXCEPTIONAL = static_cast<char>(240);
-static const char MARKER_MASTERWORK = static_cast<char>(15);
-static const char MARKER_IMPROVED_LEFT = static_cast<char>(174);
-static const char MARKER_IMPROVED_RIGHT = static_cast<char>(175);
-
-static char quality_table[] = { 0, '-', '+', '*', MARKER_EXCEPTIONAL, MARKER_MASTERWORK };
+static char quality_table[] = { 0, '-', '+', '*', '\xF0', '\x0F' };
 
 static void addQuality(std::string &tmp, int quality)
 {
@@ -830,7 +825,7 @@ std::string Items::getDescription(df::item *item, int type, bool decorate)
         addQuality(tmp, item->getQuality());
 
         if (item->isImproved()) {
-            tmp = MARKER_IMPROVED_LEFT + tmp + MARKER_IMPROVED_RIGHT;
+            tmp = '\xAE' + tmp + '\xAF';
             addQuality(tmp, item->getImprovementQuality());
         }
     }
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -720,7 +720,13 @@ df::coord Items::getPosition(df::item *item)
     return item->pos;
 }
 
-static char quality_table[] = { 0, '-', '+', '*', '\xF0', '\x0F' };
+// These '\xFF' chars refer to quality markers from curses.png, namely: 250 (≡), 15 (☼), 174 («) and 175 (»).
+static const char MARKER_EXCEPTIONAL = '\xF0';
+static const char MARKER_MASTERWORK = '\x0F';
+static const char MARKER_IMPROVED_LEFT = '\xAE';
+static const char MARKER_IMPROVED_RIGHT = '\xAF';
+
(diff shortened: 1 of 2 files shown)


library/modules/Items.cpp
```diff
@@ -720,13 +720,13 @@ df::coord Items::getPosition(df::item *item)
     return item->pos;
 }
 
-static char quality_table[] = { 0, '-', '+', '*', '=', '@' };
+static int quality_table[] = { 0, 45, 43, 42, 240, 15 };
```
lethosor, line 723: What's the reasoning for changing these to `int`s? IMO they decrease readability.

If it must contain integers due to the `240` being out of range for a signed int8, you could use `static_cast<int>('-')`, etc.

library/modules/Items.cpp
```diff
@@ -825,7 +825,7 @@ std::string Items::getDescription(df::item *item, int type, bool decorate)
         addQuality(tmp, item->getQuality());
 
         if (item->isImproved()) {
-            tmp = "<" + tmp + ">";
+            tmp = static_cast<char>(174) + tmp + static_cast<char>(175);
```
myk002, line 828: these as well should be given names so they don't appear to be "magic numbers". e.g. up under `quality_table`, you could add:
```cpp
static const char IMPROVEMENT_MARKER_LEFT = static_cast<char>(174);
static const char IMPROVEMENT_MARKER_RIGHT = static_cast<char>(175);
```

library/modules/Items.cpp
TaxiService: sorry, the reasoning was "i have no idea what i'm doing, but this seems to work!" :flushed:

all i knew was that the following didn't work:
(https://github.com/TaxiService/dfhack/actions/runs/4576523769)
```
[-Werror=multichar]
  723 | static char quality_table[] = { 0, '-', '+', '*', '≡', '☼' };
      |                                                   ^~~
/home/runner/work/dfhack/dfhack/library/modules/Items.cpp:723:58: error: multi-character character constant [-Werror=multichar]
  723 | static char quality_table[] = { 0, '-', '+', '*', '≡', '☼' };
      |                                                        ^~~
/home/runner/work/dfhack/dfhack/library/modules/Items.cpp:723:64: error: narrowing conversion of ‘14846369’ from ‘int’ to ‘char’ [-Wnarrowing]
  723 | static char quality_table[] = { 0, '-', '+', '*', '≡', '☼' };
      |                                                            ^
/home/runner/work/dfhack/dfhack/library/modules/Items.cpp:723:64: error: narrowing conversion of ‘14850236’ from ‘int’ to ‘char’ [-Wnarrowing]
```
and then myk suggested to use this "static_cast<char>(999)" thing; so i cobbled something together and it miraculously worked. 

but yeah he too had some concerns about readability. I'll see what i can do to make this pr more sensible. 💪

library/modules/Items.cpp
TaxiService: roger. I'll go and try making everything work in a more readable way.
thanks for the tips by the way, both of you!

library/modules/Items.cpp
quietust: Is there a reason why we can't just use `\xF0` for '≡' and `\x0F` for '☼'?

library/modules/Items.cpp
quietust: As above, these could probably just be `'\xAE'` and `'\xAF'` without having to deal with all of the casts.

library/modules/Items.cpp
TaxiService: the only reason is the same of my previous post. (idk what im doing + it seems to work!)

how would the readability compare with my latest commit? https://github.com/DFHack/dfhack/pull/3137/commits/3a6893de53ee3994aa68641dd8b49a423ff617ab

library/modules/Items.cpp
TaxiService: needless to say, i only saw your comments after posting my last commit 🙄 i'll do another one later as soon as i can

myk002 (commented):
this looks good. needs a changelog entry to describe the improvement.

TaxiService:
how about...
"Corrected quality markers characters for Exceptional, Masterwork and Decorated items."
or something like that?

also, i've tried to simplify the code even more by just doing what Quietust suggested, which seems like a good idea. Is it readable enough?

myk002:
> how about... "Corrected quality markers characters for Exceptional, Masterwork and Decorated items." or something like that?

They weren't *wrong* before, just approximate. Maybe "use same quality markers for item descriptions as are used in the base game"? You also have to attribute the changelog entry to something. this is technically a change to the Items module, but people might not care so much about that. It might make more sense to put it in Misc Improvements and attribute it to `buildingplan`:

```
Misc Improvements
-----------------
`buildingplan`: use same quality markers for item descriptions as are used in the base game
```

> also, i've tried to simplify the code even more by just doing what Quietust suggested, which seems like a good idea. Is it readable enough? (basically, all the changes amount to this [develop...TaxiService:dfhack:patch-6](https://github.com/DFHack/dfhack/compare/develop...TaxiService:dfhack:patch-6))

I respectfully disagree that this is more readable. using bare constants in code essentially makes them "magic". They need names, or at the very least explanatory comments. Otherwise how is a person who is new to the code to figure out what they mean?

myk002 (commented):
Looks good (still needs changelog tho)

TaxiService:
```
Misc Improvements
--------------------
buildingplan: manual item selection window should now use the same item quality symbols as the base game.
```
i guess? 🤔 i'm not sure, how would you describe the change? I'll be okay with whatever wording you end up choosing! Choosing word is not my strong suit ;_;

library/modules/Items.cpp
lethosor: I like that new commit.

The reason pasting the literals in didn't work (https://github.com/DFHack/dfhack/pull/3137#discussion_r1155103028) is that a `char` in C is only 8 bits, and the Unicode characters you pasted in require more than 8 bits to represent (that's what the `14846369` means in one of your error messages - it's the numerical representation of your character in Unicode, which is much bigger than the allowed maximum of 255).

myk002:
this is a particularly difficult changelog to wordsmith. I'll tack something in. I don't want to hold up this PR on something so nitpicky.

quietust:
For the record, what I was trying to suggest was just doing something like this:
```
static const char quality_table[] = {
    '\0',   // (base)
    '-',    // well-crafted
    '+',    // finely-crafted
    '*',    // superior quality
    '\xF0', // (≡) exceptional
    '\x0F'  // (☼) masterful
};
```

Strictly speaking, **all** of the characters in that list are "magic numbers", but in my opinion, splitting them into separate constants provides no benefit - on the contrary, I think it makes the code more difficult to understand.

myk002:
Agree. This is the best option.

PR #3148, opened 2023-04-02
[hotkeys] adjust text margin in popup menu for wide scrollbar

### LINKED ISSUE #3146: Keybind widget needs resize fix
When you have more than a page full of keybinds a scrollbar covers up the key used in the combo:

image

myk002:
adjustment needed for new double-wide scrollbars

Fixes #3146

### DIFF
--- a/plugins/lua/hotkeys.lua
+++ b/plugins/lua/hotkeys.lua
@@ -107,7 +107,7 @@ local function get_bindings_to_hotkeys(hotkeys, bindings)
 end
 
 -- number of non-text tiles: icon, space, space between cmd and hk, scrollbar
-local LIST_BUFFER = 2 + 1 + 1
+local LIST_BUFFER = 2 + 1 + 1 + 2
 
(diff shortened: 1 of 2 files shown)


plugins/lua/hotkeys.lua
```diff
@@ -143,7 +143,7 @@ local function get_choices(hotkeys, bindings, is_inverted)
     -- adjust width of command fields so the hotkey tokens are right justified
     for _,choice in ipairs(choices) do
         local command_token = choice.text[1]
-        command_token.width = max_width - choice.hk_width - 3
+        command_token.width = max_width - choice.hk_width - 5
```
ab9rf, line 146: that 5 is feeling a bit magical

myk002: very fair, especially given the conversation we just had over TaxiService's PR : P

will fix

myk002: done

ab9rf (approved):

Issue #3159, opened 2023-04-03
Read access violation in StockpileIterator::operator++


silverflyone:
_Originally posted by @flarejoltporeon in https://github.com/DFHack/dfhack/issues/2957#issuecomment-1493960863_

@myk002 and @silverflyone 

_You can find a save at https://dffd.bay12games.com/file.php?id=16545 (Please ignore it says it's for the 9,000 pants issue - `combine all` still crashes it, and I reverted to the unmodified version of combine just to make sure._ 

_I was always just assuming I'd done something wrong in the past with my commands and borked it entirely._

_Mod List (Zip attached to this comment):_ 
_See-Through Smoothing Designations_
_clinodev's dry mines (no aquifers)_
_Dwarven Weapons Fixed 2.0_
_Interface Tweaks 1.3_
_Let Dwarven Women Have Beards 1.4_
_No Attribute/Skill Rust 1.0.0_
_Rounded Hills 1.6_

[Mods.zip](https://github.com/DFHack/dfhack/files/11136664/Mods.zip)

Triage:
Unhandled exception raised in Buildings.cpp.

```
Exception thrown: read access violation.
this->**block** was nullptr.
```

code block:
```
using Buildings::StockpileIterator;
StockpileIterator& StockpileIterator::operator++() {
...
        while (current >= block->items.size()) {  // **line 1644 * exception thrown here.**
            // Out of items in this block; find the next block to search.
            if (block->map_pos.x + 16 <= stockpile->x2) {
                block = Maps::getTileBlock(block->map_pos.x + 16, block->map_pos.y, stockpile->z);
                current = 0;
            } else if (block->map_pos.y + 16 <= stockpile->y2) {
                block = Maps::getTileBlock(stockpile->x1, block->map_pos.y + 16, stockpile->z);
                current = 0;
            } else {
                // All items in all blocks have been checked.
                block = NULL;
                item = NULL;
                return *this;
            }
        }

...
}
```

call stack:
```
	SDL.dll!DFHack::Buildings::StockpileIterator::operator++() Line 1644	C++
 	SDL.dll!DFHack::Buildings::getStockpileContents(df::building_stockpilest * stockpile, std::vector<df::item *,std::allocator<df::item *>> * items) Line 1572	C++
 	SDL.dll!buildings_getStockpileContents(lua_State * state) Line 2384	C++
[...]

PR #3160, opened 2023-04-03
Fix to read access violation in StockpileIterator::operator++

silverflyone:
Fix to #3159.
Valid map coordinates from (0, 0, 0) to (world->map.x_count - 1, world->map.y_count - 1, world->map.z_count - 1).
Stockpile coords (x1, y1, z) to (x2, y2, z) may lie outside of this region. 
Exit with no items if the stockpile exists outside of the valid map plane.
While iterating the block (top-left to right, then top to bottom), ensure block coords stay within the valid map plane.


library/modules/Buildings.cpp
```diff
@@ -1645,19 +1645,25 @@ StockpileIterator& StockpileIterator::operator++() {
...
+            block = Maps::getTileBlock(std::min(std::max(stockpile->x1, 0), world->map.x_count-1), std::min(std::max(stockpile->y1, 0), world->map.y_count-1), stockpile->z);
             current = 0;
         }
 
         while (current >= block->items.size()) {
             // Out of items in this block; find the next block to search.
-            if (block->map_pos.x + 16 <= stockpile->x2) {
-                block = Maps::getTileBlock(block->map_pos.x + 16, block->map_pos.y, stockpile->z);
+            if (std::max(block->map_pos.x + 16, 0) <= std::min(std::max(stockpile->x2, 0), world->map.x_count-1)) {
+                block = Maps::getTileBlock(std::min(std::max(block->map_pos.x + 16, 0), world->map.x_count-1), block->map_pos.y, stockpile->z);
```
myk002, line 1663: this seems a little too conservative for the current invariants. for example, we know at this point that `stockpile->x2 >= 0` and `stockpile->x1 <= world->map.x_count - 1`. similarly for `y` below

myk002 (commented):
this looks like it will fix the problem. do you know what *causes* the bounds of a stockpile to be outside of the map? is it corruption, or just the way stockpiles are stored when near the edges of the map?

edit: and if it's corruption, should we also verify that `x1 <= x2` and `y1 <= y2`?

silverflyone:
It is strange. I'm not sure how you could create a stockpile through the GUI with bounds outside the map; the DF validations should kick in to limit the selectable area.

DF doesn't complain when accessing that stockpile, and lists the items contained. So it handles this case gracefully.

For x1>x2 and y1>y2? Current functionality will return items from a single block, being the top left one. It will then exit due to the >x2 and >y2 next block check.

However, if the corners are interchangeable (meaning x1/y1 can be bottom left or bottom right, or top right), then not all items would be found. I'll create some stockpiles and check the corresponding x1,y1 and x2,y2.

silverflyone:
I've created some stockpiles starting at top-left, bottom-left, bottom-right, too-right, added some areas and removed some areas so they were non rectangular. Used gm-editor to check: x1,y1 and x2,y2 define a rectangle that bounds the area of the stockpile. X1,y1 is top left in all these cases.

So I think the current functionality is good: start at the top left block and go right, then down to check for items contained by the stockpile. If ever x2,y2 is less than x1,y1 then the top left block will be checked and then iterator will finish.

library/modules/Buildings.cpp
silverflyone: Yes, agree. I've removed the extra min/max checks.

docs/changelog.txt
```diff
@@ -39,6 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `hotkeys`: hotkey hints on menu popup will no longer get their last character cut off by the scrollbar
 - ``launchdf``: launch Dwarf Fortress via the Steam client so Steam Workshop is functional
 - `blueprint`: interpret saplings, shrubs, and twigs as floors instead of walls
+- ``Buildings::StockpileIterator:``: fix to use world map boundaries when stockpile coordinates exist outside of the world map. Fixes unhandled exception for `combine`.
```
myk002, line 42:
```suggestion
- `combine`: fix error processing stockpiles with boundaries that extend outside of the map
```

myk002 (approved):

PR #3164, opened 2023-04-03
buildingplan: itemselection ui rework

TaxiService:
changes:
- `R: Sort by: ...` has been moved _above_ the item list.
- the item list now has an `INTERIOR_FRAME` around it.
(since the search box is tied to the list, the border is weaved between it and the list. Not super happy about this, but it works well enough for now.)
- the search box now has a `>` in front of it, to draw attention to the search bar (since it's too close to the frame 🙄 )
- removed brackets from around the item amounts
- rearranged the "controls" below the list, item-interaction is all neatly on the left side, and controls to proceed/go back are on the right
- the window's minimum height has probably increased by a couple of tiles 🤔 

https://github.com/DFHack/dfhack/pull/3123


myk002:
also note that you can set the `frame_inset` of the Window to 0 so the interior border can sit flush with the exterior border (filterselection also does this)

TaxiService:
> You can see in the filterselection dialog how I replaced the FilteredList's EditField with my own, allowing more flexible placement. would that help?

oh blyat, yeah! that would've been cool to know 🤣 
initially the idea was to have the sorting-control _between_ the searchbar and the list... i will look into this, but this shouldn't stop the merge imo.

> also note that you can set the `frame_inset` of the Window to 0 (or just set the inset for the top and bottom edges to 1) so the interior border can sit flush with the exterior border (filterselection also does this)

yeah i think i experimented with that, but opted against it because we don't _need_ the horizontal space. Having that sort of empty padding space is good for the eyes. 
If we ever add a magma-safety column, we might reconsider.

docs/changelog.txt
```diff
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Misc Improvements
 - `buildingplan`: items in the item selection dialog should now use the same item quality symbols as the base game
+- `buildingplan`: rearranged elements of ``itemselection`` interface
```
```suggestion
-@ `buildingplan`: rearranged elements of ``itemselection`` interface
```

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -134,53 +125,71 @@ function ItemSelection:init()
...
-        widgets.HotkeyLabel{
-            frame={l=22, b=1},
-            key='CUSTOM_SHIFT_C',
-            label='Continue',
-            auto_width=true,
-            on_activate=self:callback('submit'),
+        widgets.Label{
+            frame={l=1, t=5},
+            text_pen=COLOR_LIGHTCYAN,
+            text={">"},
```
myk002, line 140: once you provide your own EditField, this can be integrated into that widget

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -134,53 +125,71 @@ function ItemSelection:init()
...
+            frame={l=6, b=2, w=5},
             text_pen=COLOR_LIGHTGREEN,
             text='Right',
         },
         widgets.HotkeyLabel{
-            frame={l=23, b=0},
+            frame={l=1, b=1},
             key='KEYBOARD_CURSOR_LEFT_FAST',
-            key_sep='   : ',
+            key_sep='eft: ',
```
myk002, line 162: is there a functional difference to this change (and the `ight` change above)? These will just get overwritten, correct?

myk002 (commented):
I'd suggest getting the EditField separated out in this PR. It's not that hard and it would clean up the layout a bit.

plugins/lua/buildingplan/itemselection.lua
TaxiService: yeah, the only "function" of these changes is, like, to let us know if for some reason the label that overwrites these is not working. 🤔 

like, i guess i did this to make it clearer why there were so many spaces in front of that colon. °-°

myk002: Yeah, an explanatory comment to the right of the line would be a good idea - and also would be a bit more appropriate than a non-functional code change (which could be confusing to future maintainers -- "*should* this be visible???")

TaxiService:
[Dwarf_Fortress_de5QJEksWq.webm](https://user-images.githubusercontent.com/5853667/229862794-567c6886-ed04-48cb-a0c4-2d342c30fe6f.webm)

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -81,107 +81,147 @@ function ItemSelection:init()
...
-                {label='Recently used', value=sort_by_recency},
-                {label='Name', value=sort_by_name},
-                {label='Amount', value=sort_by_quantity},
+        widgets.Panel{
+            view_id='footer',
+            frame={l=1, r=1, b=0, h=3},
+            subviews={
+                --[[
+                add an explanation for this terribleness
+                --]]
```
myk002, line 169: terribleness meaning the overwritten hyphens? yeah, maybe we should add a "key_label" parameter that overrides the resolved key name? that would get rid of the need to hackily overwrite the text. Not in this PR, though.

can this comment be removed, though?

myk002 (commented):
this looks great! you're really getting the hang of this!

plugins/lua/buildingplan/itemselection.lua
TaxiService: sorry, the usage of the word "terribleness" was probably influenced by having enjoyed this video a little too much https://youtu.be/k238XpMMn38?t=61

i meant to come back and write a more descriptive comment about what is going on in the next section, (didnt do it immediately 'cause it really takes me _a lot_ of time to find a "good way" to explain what's happening, and i still had to fix a lot more things at that point) ;_; 
but then i forgot about it, had things irl distract me and so i committed in a hurry...

by "explain what's happening" i mean the whole "we are using key_sep to move the text out of the way so we can overlay other labels on it, because the original characters suck" thing.

the point is:
how do i put it? like, how do we explain what's going on in the next section, in a concise _and_ thorough manner?
(in the end, that comment was left there so that we could have this discussion, i guess :joy:)
or do i just not put any comments at all? which yes, could be a viable option as well...

myk002 (approved):

PR #3166, opened 2023-04-03
buildingplan: planneroverlay ui rework

TaxiService:
changes:
- the minimize button has been made more noticeable, and moved up 1 line.
- the "controls" have been rearranged, to hopefully increase intuitiveness
- the item's status (`Will link later`/`Available`) has been moved _under_ the item entry, instead than on the right
- the item names have ` of` concatenated at the end (in anticipation of future changes to filter)
- the `[filter]` button has been turned into ascii and changed color to CYAN (in anticipation of future changes)
- the `[x]` button's colors have changed to RED/BLACK (active/inactive)
- the weapons counter now shows yellow text, and parenthesis around the amount (in anticipation of future changes)

what is still needed:
- the weapons counter should have a slider on the right of the number
- the filter button should be rendered immediately after the ` of` of the item name
- the filter button should show `[any material]` if unset, or the filter's text when set. (like `[microcline blocks]`)
- the `[x]` button should be rendered immediately after the filter button
- the filter button's text should be cut with an ellipsis (`...`) if too long. (like `[microcline, orthoclase, pyr...] [x]`)
- bonus: the [x] button could use a custom graphic in premium mode.

https://github.com/DFHack/dfhack/pull/3123


docs/changelog.txt
```diff
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Misc Improvements
 - `buildingplan`: items in the item selection dialog should now use the same item quality symbols as the base game
+- `buildingplan`: rearranged elements of ``planneroverlay`` interface
```
```suggestion
-@ `buildingplan`: rearranged elements of ``planneroverlay`` interface
```

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -207,40 +207,47 @@ ItemLine.ATTRS{
 }
 
 function ItemLine:init()
-    self.frame.h = 1
+    --self.frame.h = 2
```
myk002, line 210: this is subtle, but by commenting this out, then the bounding box for the ItemLine becomes unrestricted (within the bounds of its parent). The effect of this is that clicking anywhere on the panel registers as a click on the first visible ItemLine. try this: plan a well, hit `q`  a few times so a lower item is selected, then click on the second ItemLine text. The *first* ItemLine will register the click and steal the selection.

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -207,40 +207,47 @@ ItemLine.ATTRS{
...
+            text=string.char(26),
+            --auto_width=true,
             visible=self.is_selected_fn,
         },
+        widgets.Label{
+            frame={t=0, l=2},
+            text={
+                {text=self:callback('get_item_line_text')},
+                --{text='[filter]', pen=self:callback('get_f_pen')},
+                --{text='[x]', pen=self:callback('get_x_pen')},
```
myk002, line 224: no need to leave leftover comments in. they'll always be available in the git history

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -307,10 +318,10 @@ end
 
 PlannerOverlay = defclass(PlannerOverlay, overlay.OverlayWidget)
 PlannerOverlay.ATTRS{
-    default_pos={x=5,y=9},
+    default_pos={x=5,y=8},
```
myk002, line 321: I don't really want to change `overlay` significantly to implement versioning right before the stable release. We should avoid changing the `default_pos` value until that's implemented.

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -422,17 +434,28 @@ function PlannerOverlay:init()
         },
         widgets.CycleHotkeyLabel {
             view_id='weapons',
-            frame={t=5, l=4},
+            frame={b=4, l=1, w=22},
             key='CUSTOM_T',
             key_back='CUSTOM_SHIFT_T',
-            label='Num weapons:',
+            label='# of weapons:',
```
myk002, line 440: I think "Number of weapons" is more readable -- we have the space for it anyway.

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -441,7 +464,8 @@ function PlannerOverlay:init()
             end,
         },
         widgets.Label{
-            frame={b=3, l=17},
+            frame={b=2, l=25},
+            text_pen=dfhack.pen.parse{fg=COLOR_DARKGREY},
```
myk002, line 468: I like the color, but the positioning seems..distracting:
image

The right alignment of the `:` I get, but the `z:` sticking out on the left, not aligned with the line above or the line below just below it is strange. Not sure what the best solution is, though.

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -490,19 +514,20 @@ function PlannerOverlay:init()
...
+                    label='Choose items:',
+                    label_below=true,
                     options={
-                        {label='Use filters', value=0},
+                        {label='with Filters', value=0},
                         {
                             label=function()
                                 local automaterial = itemselection.get_automaterial_selection(uibs.building_type)
-                                return ('Last choice (%s)'):format(automaterial or 'Will ask')
+                                return ('Last used (%s)'):format(automaterial or 'n/a')
```
myk002, line 526: I settled on "Will ask" because I wanted to convey that it was *ok* to choose this option even if you haven't made a prior selection. "n/a" makes me worried that people will think that this option isn't valid.

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -490,19 +514,20 @@ function PlannerOverlay:init()
...
-                    frame={b=0, l=0},
-                    key='CUSTOM_I',
-                    label='Item selection:',
+                    frame={b=0, l=23},
+                    key='CUSTOM_Z',
+                    label='Choose items:',
+                    label_below=true,
                     options={
-                        {label='Use filters', value=0},
+                        {label='with Filters', value=0},
```
myk002, line 522: why capital F?

plugins/lua/buildingplan/planneroverlay.lua
myk002: I think the only change necessary would be to add a full-width Panel around the MINI_BUTT Label that clears the background.

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: no we don't! 😭 
bplanoverlay_6_weaptrap3

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: a-ha! so _that's_ what that was!! I was sure i encountered a mis-behaviour like this, at some point!
yeah, most of these commented parts are probably leftovers of my experimentations towards using ResizingPanels for list entries... i'll have to take a look at all of these and fix 'em D:

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: > "n/a" makes me worried that people will think that this option isn't valid.

that is a valid concern, but i'm not sure how i feel about saying "Will ask" 🤔 
like... "use last filter (will ask)" is like... what?! will ask what? my last filter? i don't have a last filter! 

how can i say, it shouldn't read like that at all! 
i mean, the text in the parentheses should be the last filter, not some reassuring text 🤔 the optimal thing would be to not give you the option of choosing the last used item if you didn't choose any item before!

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -529,23 +554,30 @@ function PlannerOverlay:init()
...
     }
 
+    local divider_widget = widgets.Panel{
+        view_id='divider',
+        frame={t=10, l=0, r=0, h=1},
+        on_render=self:callback('draw_divider_h'),
+        visible=function() return not self.minimized end,
+    }
+
     local error_panel = widgets.ResizingPanel{
```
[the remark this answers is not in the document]
TaxiService: you mean the error panel? if so, i like the way it is! 

The fact that it's separated makes it more obvious that, like, the errors are _a result_ of the actions you take in the previous panel. (or with your mouse movements... but anyway, not something you interact with directly)

either this, or i smoked too much plastic or shoe polish or whatever 🤔

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: i... 🤔 hm.
yeah, i'll see how to handle this tomorrow, alongside all of the other things!

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: this makes me sad, but okay. 
i mean, it feels kind of putting a band-aid on it 😕 this will need to be addressed later

TaxiService:
[Dwarf_Fortress_n0vPvEp3Im.webm](https://user-images.githubusercontent.com/5853667/229915006-7c8776b3-ee20-48e2-8c46-0a0fd5aa79f9.webm)

TaxiService:
wait a minute... 🤦 ARGGGHH

myk002:
If you're willing to wrestle with the git commandline a bit, this should get you fixed:
```
git reset --hard 78d5b2690a60ed5c30451144e035325ad72d9a51
git push --force
```
that will reset your branch to before you nuked Items.cpp and re-upload to GitHub. I have a copy of your work in case something goes horribly wrong.

TaxiService:
🙏 thank you. yesterday i literally had to turn my pc off after this. I didn't have the energy to try finding out how to restore a previous build or whatever 😞 

still... somehow Items.cpp has found its way on this pr. 🤦 and it totally shouldn't be here. idk how i managed to commit it...
what i _meant_ to do was to "uncommit" it, or unstage it or whatever the hell it's called... 
today i have new energy and maybe i can find out how to do what i want...

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: image
now it's aligned with the `z:`

plugins/lua/buildingplan/planneroverlay.lua
TaxiService: i tried looking into ways to hide that option if the last-used item was not set, but i couldn't find a way to do it ;_;
if we _can't_ hide it, what should the text say? (really, it should be hidden though)

TaxiService:
[Dwarf_Fortress_gkwsR58IAZ.webm](https://user-images.githubusercontent.com/5853667/230408327-a0bda26f-89d5-4d24-94f7-c90226aecc23.webm)

plugins/lua/buildingplan/planneroverlay.lua
myk002: I used the longer text for now. We can re-shorten once we add the slider

myk002 (approved):

myk002 (approved):
This is a huge improvement! Great job!

PR #3168, opened 2023-04-03
read mods from all mod dirs and only use most recent versions

### LINKED ISSUE #3113: only read the most recent version of a mod in mod_installed
for non-active mods, it currently reads the dirs in alphabetical order. if there are multiple versions of a mod in the mod_installed directory, it will pick up the oldest version of the mod first (or at least the one with the lowest digit in the first position).

only one version of a mod should be in the script path, and it should be the active mod (if any) or the most recent version of the mod (if the mod id is not active)

### LINKED ISSUE #3141: read mods from the steam workshop directory
script-only steam workshop mods are in a bad place, user experience-wise
DF only pulls in updates for mods that are "active" in the current world. script-only mods are *never* active. this means that they never get added to a place where DFHack is looking for script paths.

Steam *does* keep the mod workshop dir up to date for subscribed mods, so we need to start looking there, in addition to the "mods" staging dir and the "data/installed_mods" working dir.

we'll need to parse all the info.txt files to discover which mod directories are for the same mod and figure out which is the most recent version.

when a world is not loaded, the script paths should include the `scripts_modinstalled` directories from most recent versions for the union of mods found in all three folders

when a world is loaded, the script paths should include the `scripts_modactive` and `scripts_modinstalled` directories from the active version of active mods in the `data/installed_mods` folder first, then the `scripts_modinstalled` directories from the most recent versions of the remaining mods (in all three directories), not including any other versions of mods that are active

the workshop content dir is `../../workshop/content/975370/`, but we may be able to query that from the steam API (or read a cached var from DF?) to "discover" it dynamically.

myk002:
and update docs and modding guide. Also fix a behavior discrepancy for `dfhack.filesystem.listdir_recursive` where the list of files couldn't be accessed if you specified a recursive depth less than the full depth of the tree.

Fixes #3141
Fixes #3113

### DIFF
--- a/library/lua/script-manager.lua
+++ b/library/lua/script-manager.lua
@@ -2,6 +2,9 @@ local _ENV = mkmodule('script-manager')
 
 local utils = require('utils')
 
+---------------------
+-- enabled API
+
 -- for each script that can be loaded as a module, calls cb(script_name, env)
 function foreach_module_script(cb)
     for _,script_path in ipairs(dfhack.internal.getScriptPaths()) do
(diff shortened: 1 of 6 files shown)


library/lua/script-manager.lua
```diff
@@ -57,4 +60,107 @@ function list()
     end
 end
 
+---------------------
+-- mod script paths
+
+-- this perhaps could/should be queried from the Steam API
+-- are there any installation configurations where this will be wrong, though?
```
ab9rf, line 67: steam (on windows at least) allows for you to install individual games on a separate drive. i don't know whether, if you do this, the workshop for any given game will still be on the same drive as the game, or if there is a single central workshop off the main steam installation folder in such situations. if the latter is the case, this approach will fail for someone who installs DF on a different drive than their main steam installation

myk002: I'll open this as a followup issue. We'd need a Steam library wrapper like our DFSDL module, I think, which is something we can add later.

myk002: It also *may* not be necessary to read from the steam workshop directory at all. I noticed while testing that mods appeared to be copied from the steam workshop directory into the df mods/ directory on startup, so **maybe** just reading the mods/ directory is good enough. I'll have to test this more when I have more time, though.

PR #3178, opened 2023-04-07
Add "faststart" plugin to make DF start faster

quietust:
In particular, it makes the game's "Loading..." screen animate as quickly as possible, shortening it from around 10 seconds to slightly more than 1 second.

A conditional build setting makes it skip the animation as well, making it slightly faster yet.

If the Tweak plugin ever comes back, this can probably be merged into it (since it's otherwise extremely minimal).


myk002 (approved):
I am appalled that this is all that it takes to speed up that agonizingly slow screen. And here I thought it was actually loading things...

myk002:
pre-commit.ci autofix

quietust:
It actually **is** loading things, but it just loads them very slowly - each "pixel" of the Dwarf face corresponds to it loading one single file, but it waits 4 frames between each pixel.

As the plugin is currently configured, all it does is remove the "idle" frames between each step - the extra `#define` near the top makes it so it performs all of the initialization steps in one single frame, making it load a little bit faster but also making the window unresponsive for a second or so.

PR #3179, opened 2023-04-07
enable bugfix services by default

myk002:
and the new `faststart` plugin


myk002:
pre-commit.ci autofix

PR #3181, opened 2023-04-07
allow launching DF to work under wine

myk002:
if it's native windows, launches by finding Steam in the registry and running Steam.exe
if it's wine, launches by discovering `steam` in the path and running the native client that way
if the chosen path above fails, launches the `Dwarf Fortress.exe` file in the current dir

### DIFF
--- a/package/windows/launchdf.c
+++ b/package/windows/launchdf.c
@@ -1,7 +1,26 @@
+#include <process.h>
 #include <windows.h>
 
-int WINAPI wWinMain(HINSTANCE hi, HINSTANCE hpi, PWSTR cmd, int ns)
-{
+static BOOL is_running_on_wine() {
+    static const char *(CDECL *pwine_get_version)(void);
+    HMODULE hntdll = GetModuleHandle("ntdll.dll");
+    if(!hntdll)
+        return FALSE;
+
+    pwine_get_version = (void *)GetProcAddress(hntdll, "wine_get_version");
+    return !!pwine_get_version;
+}
+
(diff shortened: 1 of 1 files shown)


package/windows/launchdf.c
```diff
@@ -12,28 +31,40 @@ int WINAPI wWinMain(HINSTANCE hi, HINSTANCE hpi, PWSTR cmd, int ns)
...
+// but that's better than not having DF start at all.
+static BOOL launch_direct() {
+    STARTUPINFOA si;
+    PROCESS_INFORMATION pi;
+
+    ZeroMemory(&si, sizeof(si));
+    si.cb = sizeof(si);
+    ZeroMemory(&pi, sizeof(pi));
+
+    return CreateProcessA("Dwarf Fortress.exe",
```
ab9rf, line 59: we used `CreateProcessW` earlier but `CreateProcessA` here. this is technically not really a problem since `"Dwarf Fortress.exe"` works fine as a narrow string in this context, but the nonparallelism irks me a bit

for consistency i'd probably use `CreateProcessW(L"Dwarf Fortress.exe",` here. you'll also have to replace `STARTUPINFOA si` with `STARTUPINFOW si` above.

this might become more important if we ever need to actually keep the launcher running for some reason (workshop interaction, perhaps), in which case we might actually have to use the `si` and `pi` values later on, which means they need to be the same type in either launch scenario, and using the unicode versions elsewhere provides for better i18n support

myk002: done

PR #3184, opened 2023-04-08
Include proto directory from current source directory

cvuchener:
I had issues creating an external plugins with protocol buffers. The dfhack_plugin macro looks for proto files in the current directory but the include directory is added globally when Plugins.cmake is included (so in plugins/). I moved the include_directory inside the macro so that ${CMAKE_CURRENT_SOURCE_DIR} is the plugin directory.

This broke remotefortressreader. It looked like most of the content of its CMakeLists.txt was already done by dfhack_plugin macro. So I cleaned it and moved the proto files inside its own directory. I hope I did not break anything. It builds, but I did not test the plugin.


myk002 (commented):
Changes look reasonable (though I need to look again more carefully later), but just so I know, were you building the external plugins from the `plugins/external/` directory or from somewhere else?

cvuchener:
From `plugins/external/myplugin/`, but it affects any subdirectory (like `remotefortressreader`).

ab9rf (approved):
lgtm

PR #3194, opened 2023-04-09
add raw/cooked z level output to prospector

### LINKED ISSUE #3187: Prospect does not appear to be calculating Z level properly
Testing on DFHack 50.07b2, in Steam.

Prospect returns the Z level on which the material is located. However, it appears to be calculating strangely. For example, in my current fortress, the surface is at Elevation 44 in the GUI. Prospect says that various shrubs are on Z: 144..146, which if I subtract 100, gets me 44. So perhaps it's adding 100 to the Z level somehow? 

My current fort has surface at 44, and if I go down as far as possible, the Elevation in the GUI reads -112. The range of values returned from Prospect is -12 to 146; -12 is the SLADE.

So it looks like I subtract 100 if the reported Z is below 0, and _add_ 100 if it's above.

ab9rf:
toady shifted displayed z levels by 100 for v50

closes #3187

### DIFF
--- a/plugins/prospector.cpp
+++ b/plugins/prospector.cpp
@@ -168,9 +168,13 @@ static void printMatdata(color_ostream &con, const matdata &data, bool only_z = 
         con << std::setw(9) << int(data.count);
 
     if(data.lower_z != data.upper_z)
-        con <<"   Z:" << std::setw(4) << data.lower_z << ".." <<  data.upper_z << std::endl;
+        con <<"   Z:" << std::setw(4) << data.lower_z << " .." << std::setw(4) << data.upper_z
+            <<"   Elev:" << std::setw(4) << (data.lower_z - 100) << " .." << std::setw(4) << (data.upper_z - 100)
+            << std::endl;
     else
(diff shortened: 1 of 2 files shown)


plugins/prospector.cpp
```diff
@@ -168,9 +168,13 @@ static void printMatdata(color_ostream &con, const matdata &data, bool only_z =
         con << std::setw(9) << int(data.count);
 
     if(data.lower_z != data.upper_z)
-        con <<"   Z:" << std::setw(4) << data.lower_z << ".." <<  data.upper_z << std::endl;
+        con <<"   Z:" << std::setw(4) << data.lower_z << " .." << std::setw(4) << data.upper_z
+            <<"   Elev:" << std::setw(4) << (data.lower_z - 100) << " .." << std::setw(4) << (data.upper_z - 100)
```
myk002, line 172: I was going to ask about the regional offset, but I see this was already discussed here: https://discord.com/channels/793331351645323264/1017471248567124049/1088569309149798441

myk002 (approved):

PR #3195, opened 2023-04-09
prospector redux

ab9rf:
this is arguably a better approach for dealing with the change in elevation display in v50


myk002 (approved):
Ref: #3194

PR #3199, opened 2023-04-10
hide terminal console when running on steam deck

### LINKED ISSUE #3182: DFHack terminal should auto-hide on Steam Deck
Wanted to follow up and create this issue as suggested by someone on Reddit. 

When loading into DFHack and Dwarf Fortress on Steam Deck, there files load and I’m given the message DFHack is loaded. Have a nice day! But I cannot move on from here. 

The suggestion given is to go into the text files and add a command to “hide” but the Reddit user also said they might need to put more priority into adding this themselves. 

I’m just creating this issue for visibility per their suggestion. Thank you.

myk002:
- Finds and loads Steam lib
- initializes Steam lib (since we need it before DF initializes it itself)
- discovers the SteamDeck detection function
- calls it just before we process dfhack.init
- if we're on steam deck, hide the console window

I'm not entirely happy with having to initialize the steam lib. it's now being initialized twice, which might leak memory, but it doesn't appear to cause any real problems.

This is something we might be able to avoid once we move to the dfhooks api, since we'll be initialized later in the loading process.

Fixes #3182

### DIFF
--- a/library/modules/DFSteam.cpp
+++ b/library/modules/DFSteam.cpp
@@ -39,6 +39,8 @@ bool DFSteam::init(color_ostream& out) {
         }
 
     bind(g_steam_handle, SteamAPI_Init);
+
+    // TODO: can we remove this initialization of the Steam API once we move to dfhooks?
     if (!g_SteamAPI_Init || !g_SteamAPI_Init())
         return false;
 
--- a/library/modules/DFSteam.cpp
+++ b/library/modules/DFSteam.cpp
@@ -21,7 +21,7 @@ static const std::vector<std::string> STEAM_LIBS {
 
 bool (*g_SteamAPI_Init)() = nullptr;
 void (*g_SteamAPI_Shutdown)() = nullptr;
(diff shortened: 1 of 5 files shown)


library/modules/DFSteam.cpp
```diff
@@ -0,0 +1,66 @@
...
+    return true;
+}
+
+void DFSteam::cleanup() {
+    if (!g_steam_handle)
+        return;
+
+    ClosePlugin(g_steam_handle);
+    g_steam_handle = nullptr;
+}
```
ab9rf, line 60: this function probably should call `SteamAPI_Shutdown` before releasing the DLL

https://partner.steamgames.com/doc/api/steam_api#SteamAPI_Shutdown

need to test first to make sure that this doesn't interfere with DF calling `SteamAPI_Init` itself, but i expect it won't

myk002: done; needs testing on Windows and Steam Deck

ab9rf (approved):
pending validation from an actual steam deck user ofc

PR #3209, opened 2023-04-11
update Installing docs for Steam

### LINKED ISSUE #3208: documentation is out of date
> The release archives contain several folders, including a hack folder where DFHack binary and system data is stored, a dfhack-config folder where user data and configuration is stored, and a blueprints folder where [quickfort](https://docs.dfhack.org/en/latest/docs/tools/quickfort.html#quickfort) blueprints are stored.

the `blueprints` folders have been moved into `hack` but the documentation has not been updated to reflect this

also v50 no longer has a `raw` folder in the top-level so "the root DF folder, which should already include a data folder and a raw folder" is misleading and confusing

myk002:
This removes the sections for other platforms. I'll include a link to this commit in an issue so we can find it and retrieve some of the text when it comes time to support linux and osx again.

Fixes #3208 

Edit: direct link to rendered page from this PR: https://dfhack--3209.org.readthedocs.build/en/3209/docs/Installing.html#installing

### DIFF
--- a/docs/Installing.rst
+++ b/docs/Installing.rst
@@ -75,7 +75,7 @@ The release archives contain a ``hack`` folder where DFHack binary and system
 data is stored, a ``stonesense`` folder that contains data specific to the
 `stonesense` 3d renderer, and various libraries and executable files. To
 install DFHack, copy all of the files from the DFHack archive into the root DF
-folder, which should already include a ``data`` folder and a ``mods`` folder,
+folder, which should already include a ``data`` folder and a ``save`` folder,
 among other things. Some redistributions of Dwarf Fortress may place DF in
 another folder, so ensure that the ``hack`` folder ends up next to the ``data``
 folder, and you'll be fine.
--- a/docs/Installing.rst
+++ b/docs/Installing.rst
@@ -7,78 +7,50 @@ Installing DFHack
 .. contents::
     :local:
 
-
 Requirements
 ============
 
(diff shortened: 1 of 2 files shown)


docs/Installing.rst
```diff
@@ -91,19 +63,22 @@ restrictions apply (e.g. a file named ``Windows64`` is for 64-bit Windows DF).
...
-blueprints are stored. To install DFHack, copy all of the files from the DFHack
-archive into the root DF folder, which should already include a ``data`` folder
-and a ``raw`` folder, among other things. Some packs and other redistributions
-of Dwarf Fortress may place DF in another folder, so ensure that the ``hack``
-folder ends up next to the ``data`` folder.
+The release archives contain a ``hack`` folder where DFHack binary and system
+data is stored, a ``stonesense`` folder that contains data specific to the
+`stonesense` 3d renderer, and various libraries and executable files. To
+install DFHack, copy all of the files from the DFHack archive into the root DF
+folder, which should already include a ``data`` folder and a ``mods`` folder,
```
ab9rf, line 78: the `mods` folder may not be present if the player has never used mods - might be best to state that this folder _may_ be present

ab9rf (approved):

docs/Installing.rst
myk002: I could mention `save` instead. That might be more universal

PR #3239, opened 2023-04-16
remove stubs for deprecated travis scripts

myk002:
These haven't been used in years. Time to clean them out?


lethosor:
I think these were left in for buildmaster, actually, since some paths were hardcoded in one of those images. If that is no longer a concern, I'm ok with this.

myk002:
We haven't discussed it explicitly, but it seems we're moving towards GitHub actions and away from buildmaster. We will need to migrate the OS X build and modify the Linux build to upload artifacts, but overall I think the GitHub path is probably the right one, especially given the unreliability buildmaster was showing before it disappeared.

PR #3260, opened 2023-04-18
reenable 3dveins

ab9rf:
3dveins mostly works for 50x. it has trouble with the new HFS gem-studded columns and will occasionally delete some of the gems in them, but typically not all that many of them, and it's my opinion that the impact of this is at most minimal.

in order for this to work, however, i had to make the inability to place an inclusion not a fatal error, and so that's the one change

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -34,6 +34,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 # Future
 
 ## New Plugins
+- ``3dveins``: reinstated for v50, this plugin replaces vanilla DF's blobby vein generation with veins that flow smoothly and naturally between z-levels
 
 ## Fixes
 - Fix extra keys appearing in DFHack text boxes when shift (or any other modifier) is released before the other key you were pressing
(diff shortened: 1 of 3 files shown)


myk002 (commented):
could use a "new plugin" changelog entry. at this point I think it counts as "new" again.

Also, if you implement a "quiet" parameter, we could add this tool to the Fortress Autostart list in `gui/control-panel` so it can run automatically for new forts.

ab9rf:
those are meritorious additions, hold on this until i update the PR, probably won't be before tomorrow

ab9rf:
returning to this now that we're past sdl2

still need to rewrite for logging, so do not merge yet

ab9rf:
switched (i think) all of the messaging to use debugfilter. still needs a changelog

myk002:
> Also, if you implement a "quiet" parameter, we could add this tool to the Fortress Autostart list in gui/control-panel so it can run automatically for new forts.

don't forget `gui/control-panel` integration on the Autostart tab.

ab9rf:
> Also, if you implement a "quiet" parameter, we could add this tool to the Fortress Autostart list in `gui/control-panel` so it can run automatically for new forts.

i'm not entirely sure this is a good idea, 3dveins currently causes a hang of up to a minute and having this happen automatically and silently could lead to user confusion. i'm open to ideas on how to fix that, but i don't want to delay getting this into prod where people can actually use it while we sort that out. don't let the perfect be the enemy of the good

myk002 (approved):

PR #3263, opened 2023-04-18
updated steam df launcher

### LINKED ISSUE #3232: Make Steam metrics accurate and useful
### Background

It would be useful to have metrics on the numbers of active DFHack players to support any requests we have of Kitfox or Bay12. For example, if we want to coordinate feature roadmaps between DF and DFHack, we should be able to say what percentage of DF players DFHack features will be able to service. Similarly, if there is a change/integration/feature that DFHack is requesting of DF, it will be easier for Kitfox to make a decision about dev resource allocation if the number of DF players that it will benefit is known and clearly substantial.

In the past, we have estimated the number of active DFHack players based on download counts from our own GitHub plus the download counts of the various starter packs that redistribute DFHack. This number was fuzzy at best, and never *really* represented the number of players that intentionally use DFHack since downloading a starter pack does not necessarily indicate the intent to use DFHack.

When DFHack v50 essentially squashed all starter packs (at least for the time being), we just had our GitHub download counts, which was a better metric for measuring intentional DFHack acquisition, but was still difficult to get an accurate number from because:
1. Again, downloading doesn't mean using
2. Players tended to not download every release

(2) was probably more impactful than (1), so the download counts per release are likely to be less than the actual active DFHack player count, though we can't be sure.

Now that most of our player base is getting DFHack via Steam, we have an opportunity to get accurate metrics of actual usage.

The numbers we have right now come from our Steam [Sales and Activation report](https://partner.steampowered.com/app/details/2346660/)
- Lifetime total units (User has added DFHack to their library, but hasn't necessarily downloaded it)
- Daily active users (Based on users who have launched DFHack **explicitly** in their Steam client)
- Total first-time downloads per day (Player must have downloaded and explicitly launched DFHack)

These are interesting, but not precisely what we want. If players launch DF instead of DFHack in their Steam client, they're still using DFHack; Steam just isn't recording any playtime for DFHack and it doesn't count towards our metrics. Therefore, we have to consider other options:

### Option 1: Store DFHack usage as a DF stat

DF would maintain a minutes played stat and DFHack would maintain a similar minutes played stat. These metrics would have the "aggregated" property checked so the values would be summed globally (across all players). The ratio of the globally summed values indicates the percentage of DF playtime that uses DFHack.

Similarly, there could be two stats: vanilla_player and dfhack_player. DF would set vanilla_player to 1 and dfhack_player to 0 at startup. If DFHack is loaded, it would change dfhack_player to 1. The ratio of the globally summed values indicates the percentage of DF players that use DFHack

There is a way to query only the past N days, so these metrics can be used to find recent data and approximate a time series, satisfying our requirement to know *current* statistics, not just lifetime totals.

The potential downside is that they can be queried by any player via the DFHack APIs that we'd need to implement to see this data ourselves. Even if we don't provide those APIs, a 3rd party plugin could be loaded that makes the calls itself. In short, these metrics would be effectively public. Whether that is acceptable or not is up to Kitfox.

We would view the data in-game via a script that we would write that shows the numbers/graphs.

API docs here: https://partner.steamgames.com/doc/features/achievements

### Option 2: Launch a shadow process that tracks DFHack playtime

Summarized from Discord discussion: https://discord.com/channels/793331351645323264/1046628873397358602/1096587959869915197

This would allow Steam to calculate our metrics for us and make use of the analytics built into the web UI.

We cannot write to the DFHack metrics from the DF process, but we can spawn a second process that "represents" DFHack as far as Steam is concerned. It can even communicate with the DFHack inside of DF via RPC if necessary to write things to our namespace on the player's behalf.

The simple implementation would just be a process that hangs around while DF is running, resulting in one of the two flows:

#### Dwarf Fortress is launched from the Steam client

- DF calls dfhooks::init()
- dfhooks::init() (inside dfhack) detects that DF was launched via Steam and spawns launchdf (via steam)
- launchdf detects that Dwarf Fortress is running, so doesn't launch another copy. instead, it waits for DF to exit, then exits itself

#### DFHack is launched from the Steam client

- launchdf starts Dwarf Fortress
- if launchdf was able to start DF via steam, then it finds the started process and sleep loops until DF exits, then exits itself
- when dfhack attempts to start launchdf via Steam, Steam detects that it is already running and doesn't do anything (I think)

ab9rf:
This is an update of the DF steam launcher to support the launcher "hanging around" until DF exits, in order to accrue playtime statistics for DFHack while DF runs.

Please do not merge until we've had a chance to test this across multiple platforms; a beta channel ~~will be created (once a GHA build of this branch is ready)~~ has been created (called `launcher-test`) on steam for testing purposes.

Closes #3232.


package/windows/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
...
 
 if(WIN32)
-    add_executable(launchdf WIN32 launchdf.c)
-    install(TARGETS launchdf
-        DESTINATION ${DFHACK_DATA_DESTINATION})
+    include_directories(${STEAMAPI_SOURCE_DIR})
+    link_libraries(${STEAMAPI_LIBRARY})
+    add_executable(launchdf WIN32 launchdf.cpp)
+    install(TARGETS launchdf DESTINATION ${DFHACK_DATA_DESTINATION})
+    install(FILES ${STEAMAPI_SHARED_LIBRARY} DESTINATION ${DFHACK_DATA_DESTINATION})
```
myk002, line 8: Is this a good idea? this is the same situation we're in now with SDL.dll. Uninstalling DFHack will remove the steam dll and render DF unstartable. Do we have to install it? Won't DF already have a copy we can use?

ab9rf: `${DFHACK_DATA_DESTINATION}` is is the `hack` folder

myk002: ok, good.

PR #3264, opened 2023-04-18
WIP separating Slider logic from buildingplan

TaxiService:
hey there. We have to start somewhere to "refine and extend widgets library". 
I begun trying to move the Slider class out of `filterselection.lua`, and actually i renamed it to `RangeSlider`, since it's a slider to define a range of options...

i will need some help though in generalizing a simpler "slider" class. <_<
and a button class.
and that separator class...

like, right now this is *a* beginning...


myk002 (commented):
no logic changes, right?

- needs a changelog entry (Under "Lua", I think)
- needs API docs in Lua API.rst explaining the widget behavior and attributes

TaxiService:
no logic changes, just a couple of sprite/character changes i was messing with...

and yeah, this needs a docs entry. But maybe before all this we should make a "normal" slider as well?
i mean, it is kind of weird to have this advanced slider but no simple single-value slider. 0_0 Maybe this class should kind of be derived from the first "normal" Slider class, which is currently not even a thing!

I think the changes i've made so far are as follows:
- renamed the Slider class to RangeSlider 
- moved it in widgets.lua
- changed SLIDER_LEFT_END and SLIDER_RIGHT_END to `▐` and `▌` respectively
- added SLIDER_TAB_RANGE_LEFT and SLIDER_TAB_RANGE_RIGHT, and assigned `[` and `]` to them
- modified the slider tabs to use these last two characters on the inside (`<○] [○>`) (although i was just testing around...)

TaxiService:
i... i have been trying to "dumb down" the RangeSlider class to make just a "Slider" class... but like i'm... having a really hard time. >_> I am not sure i can do it, i can't seem to visualize what the hell is going on with the code... Or rather, how to re-define it in a way that it only has 1 handle for the user to interact with. ;_; i am dumb ok, this really isn't an efficient use of anyone's time

TaxiService:
this is what i've got so far... 

[Dwarf_Fortress_rKuRBoNsND.webm](https://user-images.githubusercontent.com/5853667/232901801-b24d65e7-6bdf-4745-a7ee-15524e06ab2c.webm)

The slider kind of reflects the CycleHotkeyLabel's value, but...
- if 1 weapon is selected, there's a weird track selected overflow to the right of the handle.
- you can't drag the handle
- the only way to move the handle is to click to its left, at which point it'll go 1 step to the left... _but without updating the actual weapon count_

i'll just push these changes for the moment (which by the way are HORRIBLY put inside planneroverlay.lua, instead of being separated in widgets.lua because i couldn't find a way to reload widgets.lua in-game🙄)
...so that you can take a look and maybe point to what is wrong (probably EVERYTHING)

myk002:
Let's keep this PR simple -- just move the code from `buildingplan` verbatim and keep any changes (even to the tiles) for a separate PR. Moving the code and changing the code at the same time makes it very difficult to keep track of.

For supporting a single ended version of the slider, I suggest trying to build it from scratch. As you figure out what the next steps are, check out the double-ended version to see how it solved that particular problem and then use that as inspiration for your implementation. Once you have a working widget, then go back and see where the similarities are and where the differences are in the final code. That will help us decide if we need a separate widget or just a new property or two on the existing one.

PR #3271, opened 2023-04-19
move Slider class from filterselection.lua into widgets.lua

TaxiService:
alright, all i did was moving the whole Slider definition out of filterselection.lua and into widgets.lua.

To make everything work, the other changes were:
- in widgets.lua, `Slider = defclass(Slider, widgets.Widget)` is now just `Slider = defclass(Slider, Widget)`
- in filterselection, `Slider{...` is now `widgets.Slider{...`, since Slider is now not locally defined anymore.

i almost forgot but i've even written something to changelog.txt 😆


myk002 (commented):
needs docs in Lua API.rst, but otherwise looks good

TaxiService:
ok, like... i attempted to add some sort of documentation about this to `Lua API.rst`, it's just that i don't fully understand how this Slider works, and i need someone to double-check if whatever i said even makes sense.

Additionally, i've renamed the class to RangeSlider. I just had to ;_; If not now, this would be needed later when an actual Slider class is implemented... and i see no reason why not just... doing it now?

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,17 @@ widget does not require direct usage of ``Tab``.
            usage of ``Tab`` in ``TabBar:init()`` for an example. See the default value of ``active_tab_pens`` or ``inactive_tab_pens``
            in ``TabBar`` for an example of how to construct pens.
 
+RangeSlider class
+-----------------
+
+This widget implements a mouse-interactable range-slider. The user can move its two handles to set minimum and maximum values to define a range.
```
myk002, line 5136: Should also mention that both handles can be moved at once by dragging the bar between the two handles.

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,17 @@ widget does not require direct usage of ``Tab``.
...
+
+:num_stops: Specifies the amount of "notches" in the range slider, the places where handles can stop.
```
myk002, line 5138: "amount" -> "number"

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,17 @@ widget does not require direct usage of ``Tab``.
...
+:get_left_idx_fn: The function used by the RangeSlider to determine what value to display on its left handle.
+:get_right_idx_fn: The function used by the RangeSlider to determine what value to display on its right handle.
```
myk002, line 5140: it's more to get the index of the stop to display the handle at -- we don't actually display the value. the purpose of these callbacks is to allow the parent to keep the slider value state and thus give the parent control over the values (allowing them to be changed by hotkeys, for example). I did it this way to give flexibility to the layout. If this class kept the slider position state internally, we'd also need to provide options for how to display the hotkey hint widgets that allow you to change the values with the keyboard.

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,17 @@ widget does not require direct usage of ``Tab``.
...
+:on_left_change: Callback executed when a handle is moved leftwards.
+:on_right_change: Callback executed when a handle is moved rightwards.
```
myk002, line 5142: `on_left_change` is actually called when the left handle changes (to the left or the right), not when any handle is moved to the left. similarly for `on_right_change`

docs/dev/Lua API.rst
myk002: It might be worth mentioning that the parent controls the index state in the introduction paragraph.

docs/dev/Lua API.rst
TaxiService: wow. i never realized this was a thing, and would've never expected it to be a thing 😆 
i mean, never in my life i had encountered a range slider in which you could _drag the line to move the handles_ 🤯 

Was about to say that the handles really need to provide some feedback on hover, like the text turning white or something... but man. The line would _also_ need to provide such feedback! Nobody is ever going to guess that this functionality exists! Should it even exist?

i mean, having one slider "push" the other when they're both on the same value is intuitive. This is just... i don't know, we maybe need to re-work the graphics to make this more obvious. :\ 
Instead of a green line, it should become an extension of the yellow handles, or something.

TaxiService:
i'll try rewording these texts tomorrow, it will take me a bit. ;_; 
sorry but i suck at wordsmithing in general, and trying to document code that i didn't write exacerbates this :joy::gun: it's really a deadly combo

docs/dev/Lua API.rst
TaxiService: So hold on... does the rangeslider always need a parent CycleHotkeyLabel?
Why can we define the number of stops then? 🤔 shouldn't the slider just mirror the number of available options from the parent?

i ran into this weirdness while attempting to write the single-handle slider class. I could set the slider to have less notches than the parent CycleHotkeyLabel, but then the handle would go _past_ the end anyway. <_<

docs/changelog.txt
```diff
@@ -49,6 +49,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Lua
+- ``widgets``: "Slider" class has been moved from ``filterselection.lua`` into ``widgets.lua`` and renamed to "RangeSlider"
```
myk002, line 52:
```suggestion
- ``widgets.RangeSlider``: new mouse-controlled two-headed slider widget
```

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,20 @@ widget does not require direct usage of ``Tab``.
            usage of ``Tab`` in ``TabBar:init()`` for an example. See the default value of ``active_tab_pens`` or ``inactive_tab_pens``
            in ``TabBar`` for an example of how to construct pens.
 
+RangeSlider class
+-----------------
+
+This widget implements a mouse-interactable range-slider. The user can move its two handles to set minimum and maximum values
```
myk002, line 5136:
```suggestion
This widget implements a mouse-interactable range-slider. The player can move its two handles to set minimum and maximum values
```

docs/dev/Lua API.rst
```diff
@@ -5130,6 +5130,20 @@ widget does not require direct usage of ``Tab``.
...
+to define a range, or they can drag the bar itself to move both handles at once.
+The handles mirror the state of its two parent ``CycleHotkeyLabels``.
```
myk002, line 5138:
```suggestion
The parent widget owns the range values, and can control them independently (e.g. with ``CycleHotkeyLabels``). If the range values change, the ``RangeSlider`` appearance will adjust automatically.
```

myk002 (approved):

TaxiService:
thanks <3 i'm totally useless when it comes to handling text for human consumption

PR #3277, opened 2023-04-21
planner panel is minimized by default

### LINKED ISSUE #3276: Start buildingplan minimized by default and persist minimized state
So DFHack can make stronger assurances that we don't hijack vanilla functionality by default.

Reddit request: https://www.reddit.com/r/dwarffortress/comments/12m8omx/comment/jh49vy4/?utm_source=share&utm_medium=web2x&context=3

Discord discussion: https://discord.com/channels/793331351645323264/807444515194798090/1098987112252047491

myk002:
and minimized state is now persisted across reloads

Fixes #3276

### DIFF
--- a/data/dfhack-config/buildingplan.json
+++ b/data/dfhack-config/buildingplan.json
@@ -1,0 +1,5 @@
+{
+	"planner": {
+		"minimized": true
+	}
+}
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -38,7 +38,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 
 ## Misc Improvements
-- `buildingplan`: minimized planner panel stays minimized until you change it again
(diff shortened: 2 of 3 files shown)


data/dfhack-config/buildingplan.json
TaxiService, line 1: this is pretty cool.

potentially, i guess this thing could be extended to provide more options about persistence?

like, you know how currently the `z: Choose items:` thing is persistent per-building? What if these buildings were also stored in here?

OR... what if we included a setting (maybe in control-panel?) that let the user choose if they want `z: Choose items`'s persistence to be either per-building or global?

TaxiService (approved):

data/dfhack-config/buildingplan.json
TaxiService: by "global" i mean... that if you choose "manually" it would always be "manually" no matter the building you pick, until you switch it to something else.

myk002: absolutely. we can tack on whatever state we want into that json file.

Issue #3251, opened 2023-04-17
buildingplan filter doesn't include gems


Franzyd:
If building something with gems you cannot select them for filtering. I don't know of any other missing items.
![image](https://user-images.githubusercontent.com/56082757/232413030-9d1c2702-a2f2-4990-b805-32937466a860.png)

fricy:
Can't filter for yarn/wool ropes either. 

[img]https://i.imgur.com/W8s4uhD.png[/img]

myk002:
for some reason I thought that was covered by Cloth. Will add. thanks!

myk002:
There ya go:
![image](https://user-images.githubusercontent.com/977482/233713305-866ba9aa-7c8b-4f59-b1dd-8bcf3054e949.png)

PR #3278, opened 2023-04-21
[buildingplan] allow filtering by yarn

myk002:
fixes https://github.com/DFHack/dfhack/issues/3251#issuecomment-1518139164

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -39,7 +39,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Misc Improvements
 - `buildingplan`: minimized planner panel stays minimized until you change it again
-- `buildingplan`: can now filter by gems (for gem windows)
+- `buildingplan`: can now filter by gems (for gem windows) and yarn (for ropes in wells)
 - ``toggle-kbd-cursor``: add hotkey for toggling the keyboard cursor (Alt-K)
 - `gui/control-panel`: add preference option for hiding the terminal console on startup
 - `gui/control-panel`: add preference option for hiding "armok" tools in command lists
--- a/plugins/buildingplan/buildingplan.cpp
+++ b/plugins/buildingplan/buildingplan.cpp
@@ -154,6 +154,7 @@ static const df::dfhack_material_category gem_cat(df::dfhack_material_category::
 static const df::dfhack_material_category clay_cat(df::dfhack_material_category::mask_clay);
 static const df::dfhack_material_category cloth_cat(df::dfhack_material_category::mask_cloth);
 static const df::dfhack_material_category silk_cat(df::dfhack_material_category::mask_silk);
+static const df::dfhack_material_category yarn_cat(df::dfhack_material_category::mask_yarn);
 
 static void cache_matched(int16_t type, int32_t index) {
     MaterialInfo mi;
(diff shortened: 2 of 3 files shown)


TaxiService:
°-° does commenting count as reviewing? i hope so...

like, this is all fine and dandy as a temporary solution, but truthfully the behaviour of this filter selector does not look good. imo.

Selecting an item from the Type list should not automatically select *all* materials of that type. It should _filter_ the material list on the right, in the sense that: 
when a type is selected, only the materials of the selected type would be shown in the material list. 

Each of these lists should also include an "`any <materialtype>`" entry _at the top of the list_, and _that's_ what the user would click to select the whole category.
Doing so, would show just "any cloth" or "any leather" (or whatever the material is) in the `Current filter` panel/label.

despite requiring 1 additional click, this should make selecting a filter easier to read/comprehend. 
Right now the behaviour is unintuitive and the feedback too wordy.

this is an old mockup in which i had attempted to show what the type list should actually do.
In this image the types are different tabs, and clicking on each tab would switch to a new list for only that material.

All these new materials that popped in the Types list should've been put in the Other category.
When i made this though, i didn't think about letting the user choose individual types of leather/cloth/gems.
If we really want to provide such granularity, here are some ideas:
- we could simply have a tab for each of these categories, although it might get a little needlessly crowded.
- we could "merge" some categories together, such as `cloth`+`yarn`+`silk`->`textiles`, merge `glass` into `gem`, merge `clay` into `stone`, and anything else in `other`
- we could make some types "subcategories" of `other`, and have their tabs appear only after the `other` tab is selected...
- ...alternatively we could just have them all into `other` and have the first list entries be `any gem`, `any clay`, etc.

(this is an old mockup, but shows some concepts, such as having the 'any x', 'any y'... list entries at the top of the list, and the effect that this has visually on the `Current filter` panel.)
image

(also notice how selecting `any stone` puts a checkmark next to 'any stone' and a smaller green dot next to stones in the list, which are selected _indirectly_)
(the point is that the user _really_ doesn't need to see the whole list of minerals in the game if all they wanted was "any stone". :\ Hell, i'd rather have the filter text tell me if i have blocks/boulders/logs selected instead!)

TaxiService (approved):
°-° does commenting count as reviewing? i hope so...

like, this is all fine and dandy as a temporary solution, but truthfully the behaviour of this filter selector does not look good. imo.

Selecting an item from the Type list should not automatically select *all* materials of that type. It should _filter_ the material list on the right, in the sense that: 
when a type is selected, only the materials of the selected type would be shown in the material list. 

Each of these lists should also include an "`any <materialtype>`" entry _at the top of the list_, and _that's_ what the user would click to select the whole category.
Doing so, would show just "any cloth" or "any leather" (or whatever the material is) in the `Current filter` panel/label.

despite requiring 1 additional click, this should make selecting a filter easier to read/comprehend. 
Right now the behaviour is unintuitive and the feedback too wordy.

this is an old mockup in which i had attempted to show what the type list should actually do.
In this image the types are different tabs, and clicking on each tab would switch to a new list for only that material.

All these new materials that popped in the Types list should've been put in the Other category.
When i made this though, i didn't think about letting the user choose individual types of leather/cloth/gems.
If we really want to provide such granularity, here are some ideas:
- we could simply have a tab for each of these categories, although it might get a little needlessly crowded.
- we could "merge" some categories together, such as `cloth`+`yarn`+`silk`->`textiles`, merge `glass` into `gem`, merge `clay` into `stone`, and anything else in `other`
- we could make some types "subcategories" of `other`, and have their tabs appear only after the `other` tab is selected...
- ...alternatively we could just have them all into `other` and have the first list entries be `any gem`, `any clay`, etc.

(this is an old mockup, but shows some concepts, such as having the 'any x', 'any y'... list entries at the top of the list, and the effect that this has visually on the `Current filter` panel.)
image

(also notice how selecting `any stone` puts a checkmark next to 'any stone' and a smaller green dot next to stones in the list, which are selected _indirectly_)
(the point is that the user _really_ doesn't need to see the whole list of minerals in the game if all they wanted was "any stone". :\ Hell, i'd rather have the filter text tell me if i have blocks/boulders/logs selected instead!)

myk002:
I fully agree that this is where we should go. I'm calling your proposed layout "buildingplan v2" and I intend to get it prioritized against the other items in the strategic roadmap. This PR is just solving a little nit in the current design so at least the functionality is there.

myk002:
Oh, and to "approve" a review,

1. click "Files changed"
image

2. click "review changes"
image

3. leave comments, and select either the "comment" or "approve" radio button, then submit:
image

TaxiService (approved):
!!!!

PR #3303, opened 2023-04-25
steam build workflow improvements

ab9rf:
improve cmake handling for downloading the sdk

set up GHA to cache the SDK download

### DIFF
--- a/.github/workflows/steam.yml
+++ b/.github/workflows/steam.yml
@@ -36,6 +36,12 @@ jobs:
           restore-keys: |
             ccache-win64-cross-msvc-develop-${{ github.event.inputs.commit_hash }}
             ccache-win64-cross-msvc
+      - name: Restore steam SDK
+        uses: action/cache@v3
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/steam.yml
```diff
@@ -36,6 +36,12 @@ jobs:
           restore-keys: |
             ccache-win64-cross-msvc-develop-${{ github.event.inputs.commit_hash }}
             ccache-win64-cross-msvc
+      - name: Restore steam SDK
+        uses: actions/cache@v3
+        with:
+          path: depends/steam/steamworks_sdk_156.zip
```
myk002, line 42: shouldn't this be a *directory* rather than a file name? i.e. `depends/steam`?

ab9rf: per github docs either is fine, see no reason not to cache just the file

image

myk002: fair enough. let's give this a try

PR #3306, opened 2023-04-25
Add `key_sep` and `val_gap` attributes to `CycleHotkeyLabel` for more stylistic control

TaxiService:
hello. This PR adds two attributes to `CycleHotkeyLabel`

`key_sep=': '` could already be user-defined in `HotkeyLabel`, but was hard-coded in `CycleHotkeyLabel`.
`val_gap=1` was another variable that could not (easily?) be accessed externally. It's the whitespace between the _label_ and the _value_ of a CycleHotkeyLabel. It defaulted to 1, but exposing it like this lets us set it to 0, and have both strings be "flush" with each other/have no space.

i've also added some changelog/docs entries... if i mis-describe something, feel free to change it to better wording. ;_;

### DIFF
--- a/library/lua/gui/widgets.lua
+++ b/library/lua/gui/widgets.lua
@@ -1488,6 +1488,8 @@ CycleHotkeyLabel = defclass(CycleHotkeyLabel, Label)
 CycleHotkeyLabel.ATTRS{
     key=DEFAULT_NIL,
     key_back=DEFAULT_NIL,
+    key_sep=': ',
+    val_gap=1,
     label=DEFAULT_NIL,
     label_width=DEFAULT_NIL,
     label_below=false,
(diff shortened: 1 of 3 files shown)


TaxiService:
goddamnit!!!!!!! 🤦 

image

library/lua/gui/widgets.lua
```diff
@@ -1499,17 +1501,16 @@ CycleHotkeyLabel.ATTRS{
 function CycleHotkeyLabel:init()
     self:setOption(self.initial_option)
 
-    local val_gap = 1
     if self.label_below then
-        val_gap = 0 + (self.key_back and 1 or 0) + (self.key and 3 or 0)
+        self.val_gap = 0 + (self.key_back and 1 or 0) + (self.key and 3 or 0)
```
myk002, line 1505: perhaps mention that the `val_gap` value is ignored if `label_below` is `true`

TaxiService: Good idea.

I wonder if we should somehow make `val_gap` affect the label's position even when `label_below` is true. 🤔 
for now i'll just update the documentation.

PR #3319, opened 2023-04-27
[buildingplan] check items for accessibility for dialogs

### LINKED ISSUE #3318: DFHack tries to use unreachable items when building constructions
DFHack version `50.07-beta1 (release) on x86_64`
Game Version: `Dwarf Fortress v50.07 (Steam)`
OS: `Windows 11 Pro (22621.1555)` 
CPU: `Intel i9-12900K x64`
RAM: `32GB`

## Issue: 
When attempting to build something with an object that is unreachable, the vanilla build menu doesn't list the item - but DFHack's build construction menu will, and assign a construction job that can never be completed.

## Steps To Reproduce:
I have reproduced a minimally viable test fort [here](https://drive.google.com/file/d/1SdoRRcOKK9Bm0mq6ih3-JizZ5TIfljGZ/view?usp=sharing).

There are "Willow Logs" in the brook (fallen there when cut) - and attempting to construct a floor with them in one of the open spaces will have different behaviors when using Vanilla vs DFHack.  

Vanilla: There are no "Willow Logs" available to build with.
DFHack: There are "Willow Logs" listed, but construction jobs will never complete.

## Expected Behavior:
 - Unreachable items are filtered from the available build materials
 
## Actual Behavior:
 - Unreachable items are listed and can be selected

## Workaround:
 - If I "forbid" the willow logs in the water, Vanilla and DFHack agree they can't be used.

myk002:
before we only checked when doing the cycle, so if an inaccessible item were manually selected, we'd never be able to build

Fixes #3318

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -38,6 +38,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Fixes
 - `autoclothing`: eliminate game lag when there are many inventory items in the fort
 - `buildingplan`: fixed size limit calculations for rollers
+- `buildingplan`: fixed items not being checked for accessibility in the filter and item selection dialogs
 - `dig-now`: properly detect and complete smoothing designations that have been converted into active jobs
 
 ## Misc Improvements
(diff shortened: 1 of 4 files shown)


fricy:
As per reddit discussion here's [a save](https://drive.google.com/file/d/1FD3ppHOIUviTZ379xdklCLjRjQfNQi7q/view?usp=sharing) with the issue in dfhack 50.08-r2rc1 (git: 732e042)

Tower cap wall gets assigned the unaccessible towercap block behind the wall to the left.

PR #3324, opened 2023-04-28
C++ gui/design Plugin Code for drawing shapes faster.

johncosker:
The lua code for drawing shapes with borders was slow due to the number of tiles and boundary checks, so I've ported that code to a new C++ plugin. 

Results in a noticeable improvement in responsiveness when drawing shapes, especially large ones.

**THIS PR should be merged before the associated scripts PR:** https://github.com/DFHack/scripts/pull/693


plugins/CMakeLists.txt
```diff
@@ -77,7 +77,8 @@ set_source_files_properties( Brushes.h PROPERTIES HEADER_FILE_ONLY TRUE )
 #dfhack_plugin(add-spatter add-spatter.cpp)
 dfhack_plugin(autobutcher autobutcher.cpp LINK_LIBRARIES lua)
 dfhack_plugin(autochop autochop.cpp LINK_LIBRARIES lua)
-dfhack_plugin(autoclothing autoclothing.cpp LINK_LIBRARIES lua)
+dfhack_plugin(design design.cpp LINK_LIBRARIES lua)
+dfhack_plugin(autoclothing autoclothing.cpp)
```
myk002, line 81: this looks like a merge error (missing `LINK_LIBRARIES lua` for autoclothing)

plugins/lua/design.lua
```diff
@@ -0,0 +1,15 @@
...
+    design_draw_shape(arr)
+end
+
+function draw_points(points_obj)
+    design_draw_points(points_obj)
+end
+
+function clear_shape(arr)
+    design_clear_shape(arr)
+end
```
myk002, line 13: are these here just for the rename? otherwise, the plugin-exported functions are directly usable by other scripts. This file needs to exist, but could just be:
```lua
local _ENV = mkmodule('plugins.design')

return _ENV
```

plugins/design.cpp
myk002, line 1: I think this plugin started from the wrong skeleton. The skeleton you chose is for "automatic agent" plugins that run in the background. This looks more like a plugin that provides a service. You should model it more on plugins like `pathable` or `xlsxreader`.

johncosker: I'm persisting the data in `arr` between calls to avoid having to load the shape multiple times, would changing the plugin in the way you're suggesting keep it loaded between calls?

myk002: "enabled" means that `plugin_onupdate` will be called so the plugin can have an ongoing effect in the running game. It doesn't change anything about persistence.

docs/changelog.txt
```diff
@@ -50,6 +50,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - `gui/control-panel`: add preference option for hiding "armok" tools in command lists
 - ``Dwarf Therapist``: add a warning to the Labors screen when Dwarf Therapist is active so players know that changes they make to that screen will have no effect. If you're starting a new embark and nobody seems to be doing anything, check your Labors tab for this warning to see if Dwarf Therapist thinks it is in control (even if it's not running).
 - `overlay`: add the DFHack version string to the DF title screen
+- `gui/design`: Improved performance for drawing shapes
```
myk002, line 53: There has been a release, and the changelog got cycled -- be sure to put this up in the new "Future" section

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+REQUIRE_GLOBAL(plotinfo);
+using namespace DFHack;
+using namespace df::enums;
+
+enum ConfigValues {
+  CONFIG_IS_ENABLED = 0,
+};
+
+namespace DFHack {
+    DBG_DECLARE(pathable, log, DebugCategory::LINFO);
```
myk002, line 35:
```suggestion
    DBG_DECLARE(design, log, DebugCategory::LINFO);
```

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include "PluginManager.h"
+#include "df/graphic_viewportst.h"
+#include "df/world.h"
+#include "modules/Gui.h"
+#include "modules/Persistence.h"
+#include "modules/Screen.h"
+#include "modules/World.h"
+
+DFHACK_PLUGIN("design");
+DFHACK_PLUGIN_IS_ENABLED(is_enabled);
```
myk002, line 20: since the plugin isn't enableable, you don't need this line. The services offered by this plugin will be accessible as long as it is loaded. it doesn't need an explicit call to `enable` to function.

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+REQUIRE_GLOBAL(window_y);
+REQUIRE_GLOBAL(world);
+REQUIRE_GLOBAL(plotinfo);
+using namespace DFHack;
+using namespace df::enums;
+
+enum ConfigValues {
+  CONFIG_IS_ENABLED = 0,
+};
+
```
myk002, line 33: Also don't need this since there's nothing persisted in the savegame

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+    DBG_DECLARE(pathable, log, DebugCategory::LINFO);
+}
+
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector<PluginCommand> &commands) {
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_shutdown(color_ostream &out) {
+    return CR_OK;
+}
```
myk002, line 44: you don't technically need this either. only `plugin_init()` is required.

plugins/design.cpp
myk002, line 1: our files normally have four-space indents. could you run this through an autoformatter?

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+      PENS.emplace(pen_key,
+                   make_pen(CURSORS_MAP.at({n, w, e, s}), is_drag_point,
+                            mouse_over, is_in_shape, is_extra));
+      if (type == "" && has_point(x,y)) {
+        arr[x][y].penKey = pen_key;
+      }
+    }
+  }
+
+  // DEBUG(status).print("not cached lmao\n");
```
myk002, line 186: feel free to leave these uncommented. they have very little runtime overhead, and they're useful for debugging when you need them. If a player has an issue, you can ask them to turn on debug logging and give you a copy of the output.

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+  bool mouse_over = mouse_pos.x == x && mouse_pos.y == y;
+
+  uint32_t pen_key =
+      gen_pen_key(n, s, e, w, is_drag_point, mouse_over, is_in_shape, is_extra);
+
+  if (CURSORS_MAP.count({n, w, e, s}) > 0 && has_point(x,y)) {
+    arr[x][y].cursor_coords = CURSORS_MAP.at({n, w, e, s});
+  }
+
+  if (PENS.find(pen_key) == PENS.end()) {
```
myk002, line 168: I find `if (!PENS.count(pen_key))` to be more succinct, but this works too

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+      cur_tile.tile = pen.tile;
+      Screen::paintTile(cur_tile, x - *window_x, y - *window_y, true);
+    }
+    lua_pop(L, 1);
+  }
+
+  return 0;
+}
+
+DFHACK_PLUGIN_LUA_COMMANDS{DFHACK_LUA_COMMAND(design_draw_shape),
```
myk002, line 268: Not for this PR, but future, what would be the best way to have gui/blueprint and `gui/quickfort` use this API? Would it be a good idea to move the rendering logic from the `gui/dwarfmode.lua` library to this plugin?

plugins/design.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector<PluginCommand> &commands) {
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_shutdown(color_ostream &out) {
+    return CR_OK;
+}
+
+std::map<uint32_t, Screen::Pen> PENS;
```
myk002, line 46: I believe it is possible for a world to have the cursors loaded as different texpos values (though I have never proved it). I suspect it would be a good idea to clear PENS on world unload:
```cpp
DFhackCExport command_result plugin_onstatechange(color_ostream &out, state_change_event event) {
    if (event == DFHack::SC_WORLD_UNLOADED) {
        DEBUG(log,out).print("clearing PENS\n");
        PENS.clear();
    }
    return CR_OK;
}
```

plugins/design.cpp
johncosker: Good to know, but this one wouldn't actually be helpful so I'm removing it.

plugins/design.cpp
johncosker: Yeah I've kept that in mind, I wanted to get a smaller scope solution working first, then maybe it can be factored into a more general feature.

docs/changelog.txt
```diff
@@ -67,6 +68,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
...
```
myk002, line 71: leftover line?

myk002 (approved):

PR #3332, opened 2023-04-30
make val_sep useful if label_below == true

TaxiService:
i should've come up with this earlier... 

basically this makes the `val_sep` attribute able to affect the option value's position even if `label_below` is set to true.
meaning that if we set `val_sep` to `-3`, we can do this:

image

setting it to `-4` crops the first letter of the option value, but we don't care.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -4918,11 +4918,11 @@ It has the following attributes:
     hotkey.
 :label_width: The number of spaces to allocate to the ``label`` (for use in
     aligning a column of ``CycleHotkeyLabel`` labels).
-:label_below: If ``true``, then the option value will apear below the label
+:label_below: If ``true``, then the option value will appear below the label
     instead of to the right of it. Defaults to ``false``.
 :val_gap: The size of the gap between the label text and the option value.
     Default is ``1``. If set to ``0``, there'll be no gap between the strings.
-    Note that ``val_gap`` is ignored if ``label_below`` is set to ``true``.
+    If ``label_below`` == ``true``, negative values will shift the value leftwards.
 :options: A list of strings or tables of
     ``{label=string or fn, value=val[, pen=pen]}``. String options use the same
     string for the label and value and use the default pen. The optional ``pen``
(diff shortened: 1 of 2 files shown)


TaxiService:
additionally... someday we might want to be able to do this as well
image

like, have this `val_sep` thing be definable individually for each option. 🤔 idk if we can do this already with some of that _self._ magic or what

myk002 (approved):

myk002 (commented):
also needs a changelog entry (in the Lua section, I think)

TaxiService:
wait- i added wip because i meant to include the attribute name change to "option_gap" in this pr. 🤔 
should i make another one instead? This change is already so small...

...also when you said "needs docs", you mean the changelog?
i didn't include a changelog entry because, like... isn't it weird to say:
"we exposed a new parameter!" and then immediately after
"we changed how the new parameter we just introduced works, slightly!"

...like, why not just merging both changelog entries into the first? 🤔 or is it just too late for that?

silverwing235:
@TaxiService; its because the WIP check is broken somehow, in such a way that it does _not_ like WIPs in titles ATM, and will refuse to complete properly for that reason, as far as some of us can ascertain.

TaxiService:
@silverwing235 oh!!! i see. 😳 well, it has been removed now, but still-- consider this kind of a WIP.
Idk if committing the name change or just the changelog... i'll probably wait to hear from myk again before touching anything

myk002:
I think just a small note in the changelog saying that the option name has changed is fine. We don't need to worry about backwards compatibility for something that has existed for such a short time (and really nobody is likely to have used yet)

I think the problem with the WIP in the title is that GitHub itself decided that never-ending actions needed notification emails. The behavior of how the WIP action hasn't changed -- it always just never ended if WIP was in the title -- it's just that GitHub now thinks (as of a few months ago) that that's an error.

The reason that the WIP action doesn't end is because it was coded to loop and detect the eventual absence of "WIP" and then turn green. Nobody is wrong here, it's just a mismatch of expectations.

Issue #3183, opened 2023-04-08
Rename unforbid to claim


myk002:
and add an alias for unforbid -> claim so the old name continues to work

ref: https://discord.com/channels/793331351645323264/1017471248567124049/1091810230972723230

Penumbra69:
This looks like some good low hanging fruit for me to begin with.

Looks like this is:
- Renaming "unforbid.lua" -> "claim.lua" (along with all of the internal references).
- Renaming "unforbid.rst" -> "claim.rst" (along with all of the internal spellings)
- Adding an alias to "dfhack.tools.init" for the "unforbid -> claim" backwards compatible usage

If this is correct, I can push the PR.

Q: Do we want to change "unforbid*" in other places? 
I see it being used client facing in "forbid.lua" as well as "makeown.lua" and "kitchen-info.lua".

Penumbra69:
I think strpcaged.rst is fine, upon re-reading it.. It makes sense in context there.

PR #3339, opened 2023-05-02
Renaming "unforbid" to "claim".

Penumbra69:
- Added an alias "unforbid -> claim" in the init
 - related to [dfhack #3183](https://github.com/DFHack/dfhack/issues/3183)


lethosor:
Per my ramblings on Discord: I'm wondering if this is the direction we want to go, or if just aliasing `claim` to `unforbid` would better serve more users. As it stands, without changes to helpdb, one of the commands (whichever is an alias) will be excluded from certain lists and functionality, and `unforbid` has been around for longer.

myk002:
I put my thoughts on this subject here: #3342

myk002:
@Penumbra69 do you think you could possibly implement the first two items in #3342 ? That is, link up aliases in `helpdb` so they have help entries? That will take care of them appearing in the autocomplete list and showing appropriate help text when they're in the active commandline.

Penumbra69:
> @Penumbra69 do you think you could possibly implement the first two items in #3342 ? That is, link up aliases in `helpdb` so they have help entries? That will take care of them appearing in the autocomplete list and showing appropriate help text when they're in the active commandline.

I'll take a look at this tonight, sure!

ab9rf:
This PR appears abandoned, and in any case I think I would prefer to implement a "first class" alias strategy that would allow leaving `unforbid` in place, making `claim` an equally-visible alias, than the approach taken here.

PR #3374, opened 2023-05-16
support disabling DFHack with --disable-dfhack

### LINKED ISSUE #3198: Option to start DF without DFHack
Can you add logic to disable DFHack if a specific parameter is passed?launching DFHack would get you DF with DFHack and launching DF directly would get you vanilla DF. This would allow people to simply add the parameter say to their steam launch options and switch back and forth as desired without messing with the files.

myk002:
As requested by at least two different streamers.

This PR recognizes `--disable-dfhack` on the DF commandline and stubs out all our hooks if the flag is detected.

df::globals initialization moves to the `dfhooks_init` call in the main thread to support reading the commandline.

if the DF version is not matched in symbols.xml, the error response is exactly the same as it was before.

in order to keep the logic from changing too much, I also moved stderr/stdout redirection to the new main thread init function. this means DFHack will still have a *little* effect on the game when it is disabled, but I think this is acceptable (though always open for discussion).

Fixes #3198

### DIFF
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -336,6 +336,11 @@ on UNIX-like systems:
 
   DFHACK_SOME_VAR=1 ./dfhack
 
+- ``DFHACK_DISABLE``: if set, DFHack will not initialize, not even to redirect
+  :file:`stdout.txt` or :file:`stderr.txt`. This is provided as an alternative
+  to the ``--disable-dfhack`` commandline parameter above for when environment
+  variables are more convenient.
+
 - ``DFHACK_PORT``: the port to use for the RPC server (used by ``dfhack-run``
   and `remotefortressreader` among others) instead of the default ``5000``. As
   with the default, if this port cannot be used, the server is not started.
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -337,7 +337,7 @@ on UNIX-like systems:
   DFHACK_SOME_VAR=1 ./dfhack
 
 - ``DFHACK_DISABLE``: if set, DFHack will not initialize, not even to redirect
-  :file:`stdout.txt` or :file:`stderr.txt`. This is provided as an alternative
+  standard output or standard error. This is provided as an alternative
   to the ``--disable-dfhack`` commandline parameter above for when environment
   variables are more convenient.
 
--- a/docs/Core.rst
+++ b/docs/Core.rst
@@ -53,7 +53,7 @@ double quotes.  To include a double quote character, use ``\"``.
 If the first non-whitespace character is ``:``, the command is parsed in
 an alternative mode.  The non-whitespace characters following the ``:`` are
 the command name, and the remaining part of the line is used verbatim as
-the first argument.  This is very useful for the `lua` and `rb` commands.
+the first argument.  This is very useful for the `lua` command.
 As an example, the following two command lines are exactly equivalent::
 
   :foo a b "c d" e f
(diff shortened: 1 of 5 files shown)


docs/changelog.txt
```diff
@@ -48,6 +48,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Misc Improvements
 - Terminal console no longer appears in front of the game window on startup
 - `gui/design`: Improved performance for drawing shapes
+- ``Core``: For debugging purposes, you can now pass ``--disable-dfhack`` on the Dwarf Fortress commandline to disable DFHack for the session.
```
lethosor, line 51: Stylistic nitpick: this isn't a literal name, so I would drop the backticks:
```suggestion
- Core: For debugging purposes, you can now pass ``--disable-dfhack`` on the Dwarf Fortress commandline to disable DFHack for the session.
```
and as a bonus (or not, depending on your preferences), this would sort it closer to the top of the list

lethosor (commented):
As mentioned on Discord: I was thinking an environment variable would be easier and more consistent with our other core/startup-altering flags. Looking at the implementation, it looks pretty straightforward either way. I would still suggest an environment variable for consistency, maybe as a synonym instead of a replacement.

docs/changelog.txt
myk002: ha ; ) fair enough.

lethosor:
I'm always a little concerned when threading refactors come up, but I think the stuff that's in the main thread with this change is fine. It's more than just I/O redirection, and we could theoretically still crash DF on startup, but it looks like a reasonable split in terms of functionality.

I haven't tracked it down, but I _thought_ all of Init() was formerly called from SDL_Init() directly, which would have been running on the main thread anyway. So I don't expect any significant regressions from moving some of the logic back there.

myk002:
> As mentioned on Discord: I was thinking an environment variable would be easier and more consistent with our other core/startup-altering flags

I did some initial analysis on the related issue:
```
environment variables are not easy to set in Windows, and even if you do set them on the commandline (e.g. DFHACK_DISABLE=1 steam steam://rungameid/975370), the trip through Steam to launch DF loses the environment.
```

but adding it as a synonym would be very simple.

lethosor:
See, I would have thought setting command-line flags would be just as difficult. Does the `steam` CLI pass those through?

myk002:
> I'm always a little concerned when threading refactors come up

Me too. From our [little foray](https://discord.com/channels/793331351645323264/1057104743337570414/1095972703879573535) trying to do all of Core::Init in the main thread, I know that our code expects any thread-related bits to be done from the simulation thread. I specifically didn't move any Lua initialization or locking calls for that reason.

One bit that I was on the fence about moving was the console initialization. It probably *could* go in the main thread, but it was a larger chunk of code than I felt comfortable moving for this PR.

myk002:
> See, I would have thought setting command-line flags would be just as difficult. Does the `steam` CLI pass those through?

yes, but with a big fat warning to the user that something shady is happening.

running:
```
steam steam://rungameid/975370//--disable-dfhack
```
results in this screen:
image

myk002:
Added env var as *near* synonym. We don't have to do any initialization in that case, so I didn't. We could easily make the behavior exactly the same between the env var and the commandline param if we want to, though.

docs/Core.rst
```diff
@@ -319,6 +336,11 @@ on UNIX-like systems:
 
   DFHACK_SOME_VAR=1 ./dfhack
 
+- ``DFHACK_DISABLE``: if set, DFHack will not initialize, not even to redirect
+  :file:`stdout.txt` or :file:`stderr.txt`. This is provided as an alternative
```
lethosor, line 340: Nitpick:
```suggestion
  standard output or standard error. This is provided as an alternative
```
DF (or is it SDL?) only writes to the .txt files on Windows. (Admittedly my information is based on 0.47, so maybe it will change once those ports come out for v50)

lethosor (commented):
No real objections, but hesitant to approve without doing more testing than I have the ability to right now. Would be interested in seeing this tested on a beta branch first.

myk002:
> Would be interested in seeing this tested on a beta branch first.

Maybe we can spin a beta release later this week. People will be eager to get their hands on `exportlegends`, I think.

myk002:
I was originally planning on doing a beta once I got stockpiles, logistics, and gui/logistics to a good point, but the more I do there, the more there is to do, and I don't know when it will be shippable.

myk002:
merging for testing in the beta branch

PR #3380, opened 2023-05-16
[overlay] add links to quickstart-guide and gui/control-panel to the title screen

### LINKED ISSUE #3315: Button to launch DFHack quickstart guide and control panel on title screen
as suggested by u/old-tobie on [reddit](https://www.reddit.com/r/dwarffortress/comments/12zob2m/comment/jhu8njs/?utm_source=share&utm_medium=web2x&context=3):

more user friendly, trying DFhack on steam I personally have no idea what are and aren't DFhack features. A more obvious control window on launch or at the title screen to turn features on or off and or explain them would be cool imo.

myk002:
image

The hotkey for starting gui/control-panel is listed as `/` instead of `Ctrl-Shift-E` since the latter is not an interface key (at least not until https://github.com/Putnam3145/Dwarf-Fortress--libgraphics--/pull/4 is merged)

Fixes #3315


TaxiService (commented):
Since these will be the very first "buttons" the user will see, we should try to make it as immediately clear as possible that they _can be clicked_.

something like this might work better to convey the clickable-ness of those HotkeyLabels.

this is just a quick and dirty mockup; i am not sure if this is the best placement for the buttons, and if the buttons should have an on-hover state or not.

238747274-e5416666-ad62-492e-8385-75e6e1036977

further button-related discussion in here
https://github.com/DFHack/dfhack/issues/3382

myk002:
I agree with this change. How about if we make it a follow-up task attached to #3382 since the button classes aren't ready yet?

Issue #3383, opened 2023-05-16
rename window frame constants


myk002:
so they all start with `FRAME_`

now that we have so many, it makes sense to pay more attention to the naming scheme. starting them all with `FRAME_` will identify them as a group and help with autocomplete.

docs should be updated, and old names left in the code as mappings to the new names.

PR #3393, opened 2023-05-18
rename window frame constants

TaxiService:
Fixes https://github.com/DFHack/dfhack/issues/3383

changed frame constant naming scheme in `gui.lua` from "XYZ_FRAME" to "FRAME_XYZ".
Also added aliases for old frame names, pointing to the new ones.

Originally i had also renamed every instance of `getXyzBordersTexposStart` to `getBorderXyzTexposStart`, but then `gui/launcher` and `gui/cp437-table` would need to be modified as well, and i didn't want to venture into trying to make cross-module PRs ;_;

ideally both of those scripts wouldn't have to reference these commands at all, but we'll see about that in the future.
Our current `Frame`s don't actually list all possible sprites, so after we rework those, we might try to remove those "dependencies" or whatever they'd be called in this case.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -62,6 +62,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - ``overlay.reload()``: has been renamed to ``overlay.rescan()`` so as not to conflict with the global ``reload()`` function. If you are developing an overlay, please take note of the new function name for reloading your overlay during development.
+- `gui.lua`: changed frame naming scheme to `FRAME_X` rather than `X_FRAME`, and added aliases for backwards compatibility. (for example `BOLD_FRAME` is now called `FRAME_BOLD`)
 
 ## Removed
 - `orders`: ``library/military_include_artifact_materials`` library file removed since recent research indicates that platinum blunt weapons and silver crossbows are not more effective than standard steel. the alternate military orders file was also causing unneeded confusion.
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -62,7 +62,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - ``overlay.reload()``: has been renamed to ``overlay.rescan()`` so as not to conflict with the global ``reload()`` function. If you are developing an overlay, please take note of the new function name for reloading your overlay during development.
-- `gui`: changed frame naming scheme to `FRAME_X` rather than `X_FRAME`, and added aliases for backwards compatibility. (for example `BOLD_FRAME` is now called `FRAME_BOLD`)
+- gui.lua: changed frame naming scheme to `FRAME_X` rather than `X_FRAME`, and added aliases for backwards compatibility. (for example `BOLD_FRAME` is now called `FRAME_BOLD`)
 
 ## Removed
 - `orders`: ``library/military_include_artifact_materials`` library file removed since recent research indicates that platinum blunt weapons and silver crossbows are not more effective than standard steel. the alternate military orders file was also causing unneeded confusion.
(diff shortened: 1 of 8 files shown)


TaxiService:
ah crap-- one undocumented change is in this line...

image

i've made "GREY_LINE_FRAME" alias to "FRAME_PANEL" instead of "FRAME_WINDOW" because... i guess that a window that uses GREY_LINE_FRAME is probably not going to be resizable... right? 🤔

myk002:
> solves this #3383

Could you change this to say "Fixes #3383"? GitHub recognizes that keyword and sets things up so that merging the PR closes the issue.

> Originally i had also renamed every instance of `getXyzBordersTexposStart` to `getBorderXyzTexposStart`, but then `gui/launcher` and `gui/cp437-table` would need to be modified as well, and i didn't want to venture into trying to make cross-module PRs ;_;

Probably best to make that a standard three-part migration across 3 PRs:
1. rename the functions, leave aliases so old code still works
2. update old code to use the new function names
3. remove the aliases

If we think there are 3rd party scripts that use the old names, we can just not do step 3

docs/changelog.txt
```diff
@@ -62,6 +62,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 
 ## Lua
 - ``overlay.reload()``: has been renamed to ``overlay.rescan()`` so as not to conflict with the global ``reload()`` function. If you are developing an overlay, please take note of the new function name for reloading your overlay during development.
+- ``gui.lua``: changed frame naming scheme to ``FRAME_X`` rather than ``X_FRAME``, and added aliases for backwards compatibility. (for example ``BOLD_FRAME`` is now called ``FRAME_BOLD``)
```
myk002, line 65:
```suggestion
- ``gui``: changed frame naming scheme to ``FRAME_X`` rather than ``X_FRAME``, and added aliases for backwards compatibility. (for example ``BOLD_FRAME`` is now called ``FRAME_BOLD``)
```

library/lua/gui.lua
```diff
@@ -923,17 +923,26 @@ local function make_frame(name, double_line)
...
+FRAME_MEDIUM = make_frame('Medium', false)
+FRAME_BOLD = make_frame('Bold', true)
+FRAME_INTERIOR = make_frame('Thin', false)
+FRAME_INTERIOR.signature_pen = false
+FRAME_INTERIOR_MEDIUM = copyall(FRAME_MEDIUM)
+FRAME_INTERIOR_MEDIUM.signature_pen = false
 
 -- for compatibility with pre-steam code
-GREY_LINE_FRAME = WINDOW_FRAME
+GREY_LINE_FRAME = FRAME_PANEL
```
myk002, line 936: I'd keep this as pointing to FRAME_WINDOW since that was its original purpose -- as the external frame for an entire tool window.

TaxiService: FRAME_PANEL is also for external window, but it's the one without the thick title bar and little draggable corner. It's the one meant for panels that can't be resized/moved.

Like, FRAME_WINDOW is for draggable/resizable panels.

myk002: the "original" windows weren't draggable or resizable, but they still need the thick title bar, since they're for top-level windows, right?

TaxiService: not necessarily. :thinking: Originally (in .47) they all just used the thick character for all sides... right?

Like, the thick bar is thick because it's supposed to be grabbed. If the window can't be moved, it shouldn't be there!
It's like the hotkeys menu frame, that one can't be moved and lacks a thick/title bar

myk002: ok, that's fair. it will still have the "title" area in either case, which is fine. ok. approved. thanks!

PR #3403, opened 2023-05-19
launch DFHack through steam if DF is run from steam

### LINKED ISSUE #3333: launch DFHack via Steam if not already running and DF is launched via Steam
so we get proper playtime metrics

hopefully, Steam will detect if launchdf is already running and break the cycle so we don't get into an infinite launch loop

myk002:
Except on wine, where `launchdf` can't find the DF process.

Fixes #3333

If the player launches DFHack from the Steam client (windows):
1. launchdf.exe runs
2. launchdf launches Dwarf Fortress (via steam if possible) finds the launched process, and waits for it to exit
3. Dwarf Fortress runs and loads DFHack
4. DFHack recognizes launchdf is already running and doesn't launch it again
5. Player plays DF and eventually exits
6. launchdf exits

If the player launches DF from the Steam client (windows):
1. Dwarf Fortress runs and loads DFHack
2. DFHack recognizes that it was loaded through Steam, and runs launchdf
3. launchdf reloads itself via steam since it wasn't started from the steam client
4. launchdf finds the running Dwarf Fortress process and waits for it to exit
5. Player plays DF and eventually exits
6. launchdf exits

If the player launches DFHack from the Steam client (wine):
1. launchdf.exe runs
2. launchdf launches Dwarf Fortress (via steam if possible) and exits
3. Dwarf Fortress runs and loads DFHack
4. DFHack recognizes that it is running on wine and does not do anything

If the player launches DF from the Steam client (wine):
1. Dwarf Fortress runs and loads DFHack
2. DFHack recognizes that it is running on wine and does not do anything

If the player loads DF or DFHack outside of the Steam client (or it is not a Steam version of DF), then it is essentially the same as the wine cases above. No relaunching happens (unless the player runs launchdf directly for some reason, then it will launch Dwarf Fortress).

In the future, the native Linux build is expected to behave the same as the windows cases, but some of that code hasn't been written yet since we have no Linux build to test with.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1682,8 +1682,10 @@ bool Core::InitSimulationThread()
         fatal("cannot bind SDL libraries");
         return false;
     }
-    if (DFSteam::init(con))
+    if (DFSteam::init(con)) {
         std::cerr << "Found Steam.\n";
+        DFSteam::launchSteamDFHackIfNecessary(con);
+    }
     std::cerr << "Initializing textures.\n";
     Textures::init(con);
     // create mutex for syncing with interactive tasks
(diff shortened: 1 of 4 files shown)


myk002:
confirmed working on Windows with helpful beta testers

PR #3404, opened 2023-05-20
[RemoteFortressReader] encode transmitted names in utf-8

### LINKED ISSUE #3304: [RemoteFortressReader] Some materials names are encoded in CP437 instead of UTF-8
When retrieving the material list with RemoteFortressReader::GetMaterialList, (some) materials names  are not valid UTF-8, which makes the protobuf deserialization fail. Maybe some protobuf library will be able to continue, but unfortunately at least the rust one just fails to deserialize the entire message.

Depending on the worldgen, sometimes it will be fine, but here is one where the issue can be seen: [save](https://drive.google.com/file/d/1klKpt3K98TIixYCL1HGzC_1C6a1aceM9/view?usp=share_link). In that save, one of the material is named `Behemoth of Zikth bone` which is the CP437 encoding of `Behemoth of Zikâth bone`.

Client side, it can be worked around by commenting or replacing the `string` declaration in the proto with a `bytes` declaration, and decode it from here.

myk002:
the `DF2UTF` encoding function was being applied inconsistently. I added it to the places it seemed to be missing.

Fixes #3304

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -47,6 +47,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - DFHack screen backgrounds now use appropriate tiles in DF Classic
 - RemoteServer: fix crash on malformed json in ``dfhack-config/remote-server.json``
 - `autolabor`: work detail override warning now only appears on the work details screen
+- `RemoteFortressReader`: ensured names are transmitted in UTF-8 instead of CP437
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


myk002:
@plule could you possibly test with the MSVC build from this PR to see if I fixed the issue for you? I don't have a remote client handy.

plule:
@myk002 I confirm it fixes the issue, thank you!

PR #3410, opened 2023-05-22
get things ported to SDL2

myk002:
Changes:
- The SDL binary release (same one as is used by DF) is now downloaded and used for the DFHack build. SDL1 headers and types that were copied into our tree have been removed.
- we no longer install SDL.dll (or SDL2.dll) into DF
- the sym field of the KEYDOWN event is now unicode, and respects the keyboard layout, so we can use it directly (it seems. would need a non-US keyboard layout to test properly)
- commented out `enabler->zoom_display` and its various access methods until we can figure out what to do about the semaphores
- updated the Textures module to match updated logic in g_src


myk002:
includes some changes that were split out into #3418; will de-dup once that PR is merged

myk002:
either is fine. we can build steam deployments off this PR just as easily as a DFHack org branch

PR #3414, opened 2023-05-23
scroll mouse wheel to focus window under cursor

myk002:
because otherwise it's a pain in the butt to focus another `gui/gm-editor` window without accidentally clicking on (and therefore editing/changing) one of the displayed fields.

### DIFF
--- a/library/lua/gui.lua
+++ b/library/lua/gui.lua
@@ -789,7 +789,8 @@ function ZScreen:onInput(keys)
     if not self:hasFocus() then
         if has_mouse and
                 (keys._MOUSE_L_DOWN or keys._MOUSE_R_DOWN or
-                 keys.CONTEXT_SCROLL_UP or keys.CONTEXT_SCROLL_DOWN) then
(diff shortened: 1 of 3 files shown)


library/lua/gui.lua
```diff
@@ -818,10 +820,14 @@ function ZScreen:onInput(keys)
         end
         local passit = self.pass_pause and keys.D_PAUSE
         if not passit and self.pass_mouse_clicks then
-            for key in pairs(MOUSE_KEYS) do
-                if keys[key] then
-                    passit = true
-                    break
+            if keys.CONTEXT_SCROLL_UP or keys.CONTEXT_SCROLL_DOWN then
```
lethosor, line 823: Does this also trigger if you press the keys bound to these bindings?

myk002: if there are keyboard keys bound to these interface keys, then yes. these interface keys are normally bound to the mouse scroll wheel, though.

myk002: image

PR #3418, opened 2023-05-24
clean up some cmake

myk002:
as I was looking into offline builds, I found it to be a good opportunity to clean up some cmake logic.

- put the version vars closer to the top since those are the most often edited
- update minimum cmake version to 3.18
- update some incorrect comments
- include newer submodules in the submodule check
- use variables for destination paths so we don't repeat the string "hack" everywhere
- put all zlib logic in one place
- download zlib to the build dir instead of the source dir to keep the source dir cleaner

I didn't touch anything related to osx since we have no way to test that right now


CMakeLists.txt
```diff
@@ -276,30 +263,37 @@ elseif(WIN32)
...
-    else()
-        download_file("https://github.com/DFHack/dfhack-bin/releases/download/0.44.09/win32-zlib.lib"
-            ${ZLIB_DOWNLOAD_DIR}/zlib.lib
-            "f4ebaa21d9de28566e88b1edfcdff901")
+    set(ZLIB_FILE zlib.lib)
+    set(ZLIB_PATH ${CMAKE_BINARY_DIR}/depends/zlib/)
+    set(ZLIB_MD5 a3b2fc6b68efafa89b0882e354fc8418)
+    file(COPY ${dfhack_SOURCE_DIR}/depends/zlib/ DESTINATION ${ZLIB_PATH})
+    download_file("https://github.com/DFHack/dfhack-bin/releases/download/0.44.09/win64-${ZLIB_FILE}"
+        ${ZLIB_PATH}lib/${ZLIB_FILE}
```
lethosor, line 279: I disagree on the zlib change. The file pretty much never changes, so putting it in the source tree allows for an arbitrary number of build folder creations/deletions without needing to redownload zlib. The md5 check allows us to force a redownload if the file ever changes.

For consistency with other downloads, though, we could consider storing it in `package/windows/*/` instead of `depends/`.

ab9rf (changes requested):

CMakeLists.txt
```diff
@@ -182,62 +180,51 @@ endif()
...
 endif()
 
 # external tools will be installed here:
 set(DFHACK_BINARY_DESTINATION .)
-# dfhack data goes here:
-set(DFHACK_DATA_DESTINATION hack)
 # plugin libs go here:
-set(DFHACK_PLUGIN_DESTINATION hack/plugins)
-# dfhack header files go here:
-set(DFHACK_INCLUDES_DESTINATION hack/include)
```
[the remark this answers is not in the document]
myk002: this removal was unintentional. fixed.

CMakeLists.txt
myk002: that's a good point. I was trying to decide what would be best for new downloads (like the steam sdk and sdl2) and was seeing what it would be like to make our existing downloads follow suit. It appears that "nothing in the source tree" is never going to fly because of generated headers that we *want* in the tree (not to mention generated docs that have to be there because sphinx). It's not even like moving zlib to `package` will make it more consistent. Only SDL was downloaded there and soon we won't even be doing that anymore. Steam sdk is expected to be in `depends/steam`. Right now in #3410, SDL2 is downloaded and extracted into `build/depends/sdl2`. I was going to move steam there too, but maybe that's the wrong direction and I should be moving SDL2 into `depends` instead of `build/depends`.

lethosor: There _was_ more stuff than just SDL going into `package`. Looks like currently the only other thing is `libstdc++` on macOS, and that's subject to change if the compiler changes. libruby used to go into `package/` as well, but doesn't anymore.

Actually, I think the distinction there is that *runtime* dependencies were downloaded into `package/`, and build-time dependencies went in `depends/`. I'm not opposed to keeping it that way. Either way, these downloads have a lot in common with git LFS files - we would keep them in the source tree, if that wasn't problematic with them being binaries.

lethosor: Oh, and technically the generated headers could go in the build folder (a few other generated headers do already, currently all of which are from dependencies), but that opens up a new possible set of problems with IDEs not finding them properly

PR #3421, opened 2023-05-24
don't throw if json is unreadable

myk002:
just act like the file didn't exist (unless strict is set)

Fixes bug brought up on the Steam forums: https://steamcommunity.com/app/2346660/discussions/0/3827540383542758371/


lethosor (approved):
Ok with this given that it seems to fit with the intent of our existing `strict` option

PR #3422, opened 2023-05-24
[sdl2] prevent DF interface events for handled hotkeys

myk002:
this is a change in behavior from earlier versions. now that we have the capability to inhibit SDL events in DF's event loop, we can prevent our handled hotkeys from also triggering DF input events.

Among other things, this prevents a backtick from appearing in an active text entry field when players bring up the launcher.


lethosor:
I'm not sure this is always what we want... if someone has bound a key globally, which only conflicts on specific screens, this change would prevent it from working on those screens. Our solution could be to say "don't do that", but I worry that it would be significantly harder to troubleshoot with this change in place.

(Unrelated, but I'm pretty sure we were able to inhibit SDL events even before dfhooks, and chose not to - I had a patch in place to inhibit all events if DF is unfocused, although I'm not sure if that made it into mainline DFHack.)

myk002:
Reading through g_src, I think a good compromise might be to inhibit the TEXTINPUT event, not the KEYUP/KEYDOWN events. That would prevent handled hotkeys from adding to text fields, which is the real problem I'm trying to solve

myk002:
I'm a bit torn on this. You make a valid point that an inattentively configured global keybinding could cause confusion. I added help text and debug logging to mitigate, but if this case is going to happen, it would still be frustrating for the player.

The question is: is the current frustration caused by having keybindings "fall through" to the UI worth the risk of causing that aforementioned potential confusion? It's hard to judge. I don't even know if the current "fall through" behavior is causing general frustration, or if it's just me. I'm open to more opinions on this.

lethosor:
> The question is: is the current frustration caused by having keybindings "fall through" to the UI worth the risk of causing that aforementioned potential confusion? It's hard to judge. I don't even know if the current "fall through" behavior is causing general frustration, or if it's just me. I'm open to more opinions on this.

I've never really had an issue with it myself. I think that's because most (of my?) DFHack keybindings use modifiers like Ctrl, and most DF keybindings don't, so the risk of overlap is low. In special cases (like Ctrl-R/P/S/L for macros in DF), it's not too hard to work around. Likewise, DFHack keybindings that do have a high likelihood of conflicting can be scoped to just the screens where they're needed (e.g. single-letter keybindings - not sure how many we still have, but I think we used to have some that applied to very specific sidebar modes).

myk002:
The only problem that I run into is with backtick. it's a text input key so if an edit field is focused, either vanilla or df, then the field will get a backtick in it in addition to bringing up `gui/launcher`. it's a side effect I didn't anticipate when choosing that keybinding.

Issue #3342, opened 2023-05-02
Make aliases useful for deprecating commands


myk002:
The current alias system is intended to allow a player to make a shorter version of frequently used commands and/or parameter lists. However, aliases have recently begun to be used as a way to allow an old name for a renamed/deprecated tool to function. The idea is that players who are used to typing the old name will still be able to do so, and renamed tools that can be `enable`d will still be `enable`able via their old name.

Aliases, however, are not well integrated into the rest of the system and do not really provide players and script writers with a good overall experience when they are used to make deprecated names "still available". Here are some ideas for improving the player experience for aliases:

- [ ] show appropriate help in `gui/launcher` when the alias is in the commandline. It is debatable whether they should be in the autocomplete list, but at least the help text for the command that the alias expands to should be shown in the help area. It would also be useful to prepend the help text with the alias expansion so the player can see exactly what they will be running.
- [ ] possibly already covered by the `gui/launcher` item, `helpdb` should display useful help text if an alias is given as an entry. modifying aliases should poke `helpdb` to reindex
- [ ] allow `reqscript` to find the correct script if an alias is named (e.g. `reqscript('unforbid')` would redirect to `reqscript('claim')`)
- [ ] allow `require` to find the correct module if an alias is named (e.g. `require('plugins.oldname')` would redirect to `require('plugins.newname')`)

Penumbra69:
Taking a look at this.

myk002:
do you need any help on this one? I could do this part to unblock the claim rename PRs

jppessoap:
Hello. Could aliases be made for the commands "show" and "hide" so they include "console" or "terminal" on their name, which would better reflect what exactly is shown or hidden, making finding those commands easier?

Penumbra69:
> 

Apologies - I had some personal matters to deal with -- I will dig back into this. 
I believe I know what needs to be done, will try to get a PR pushed asap.

Penumbra69:
Almost ready to push this PR -- one Q:
' document that players should poke helpdb to reindex after modifying aliases' 

What kind of documentation are you suggesting here?

myk002:
in the `alias` command docs, just a line that says something to the effect of "help text is refreshed whenever a game is loaded, but if you modify alias definitions during a running game and want to enable autocomplete and help text for the alias in `gui/launcher`, be sure to run this command: `:lua require('helpdb').refresh()`

Penumbra69:
[...]

PR #3423, opened 2023-05-26
Expose Alias API to Lua

Penumbra69:
Issue (#3342)
 - Make aliases available to lua
 - Ensure aliases typed in `gui/launcher` show proper help entry
 - Enhance help entry with aliased command information
 - added a simple set of tests

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -3577,6 +3577,68 @@ static int internal_md5file(lua_State *L)
     }
 }
 
+/**************************
+ * Wrappers for Alias API *
+ **************************/
+
+static int internal_getAliasCommand(lua_State *L)
+{
+    const char *name = luaL_checkstring(L, 1);
+    std::string ret = Core::getInstance().GetAliasCommand(name);
+    if (!ret.empty())
+        lua_pushstring(L, ret.c_str());
+    else
+        lua_pushnil(L);
+    return 1;
+}
+
+static int internal_listAliases(lua_State *L)
+{
+    auto aliases = Core::getInstance().ListAliases();
+
+    lua_newtable(L);
+    for (const auto& p : aliases)
+    {
+        Lua::Push(L, p.first.c_str()); // alias
+        Lua::PushVector(L, p.second);
+        lua_settable(L, -3);
+    }
+    return 1;
+}
+
+static int internal_isAlias(lua_State *L)
+{
+    const char *name = luaL_checkstring(L, 1);
+    lua_pushboolean(L, Core::getInstance().IsAlias(name));
+    return 1;
+}
+
+static int internal_addAlias(lua_State *L)
+{
+    const char *new_alias = luaL_checkstring(L, 1);
+    std::vector<std::string> alias;
+    split_string(&alias, new_alias, " ");
+
+    if (!alias.empty())
+    {
+        std::string name = alias[0];
+        alias.erase(alias.begin());
+        lua_pushboolean(L, Core::getInstance().AddAlias(name, alias));
+    } else {
+        lua_pushboolean(L, false);
+    }
+
+    return 1;
+}
+
+static int internal_removeAlias(lua_State *L)
+{
+    const char *rem_alias = luaL_checkstring(L, 1);
+
+    lua_pushboolean(L, Core::getInstance().RemoveAlias(rem_alias));
+    return 1;
+}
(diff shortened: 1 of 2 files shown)


library/LuaApi.cpp
```diff
@@ -3577,6 +3577,68 @@ static int internal_md5file(lua_State *L)
...
+    const char *name = luaL_checkstring(L, 1);
+    lua_pushboolean(L, Core::getInstance().IsAlias(name));
+    return 1;
+}
+
+static int internal_addAlias(lua_State *L)
+{
+    const char *new_alias = luaL_checkstring(L, 1);
+    std::vector<std::string> alias;
+    split_string(&alias, new_alias, " ");
```
lethosor, line 3620: I would prefer to have this function take a list of arguments instead; otherwise there is no way to set up certain aliases, e.g. with spaces in arguments.

It looks like there is a `Lua::GetVector()` that could work here. It's a more specific function than its name would apply - it reads strings starting at the stack position you give (`1`) and pushes them all into a `vector<string>`, but that looks like it should work for our purposes here.

Penumbra69: I'll adjust.

Penumbra69: This change is complete, tests have been updated.

lethosor: I was mistaken - on second look, it looks like GetVector takes a table and converts it to a vector, but your code is expecting that.

Have you tested passing a non-table to make sure the code doesn't crash? An error message is fine (and expected).

library/LuaApi.cpp
```diff
@@ -3577,6 +3577,67 @@ static int internal_md5file(lua_State *L)
...
+}
+
+static int internal_addAlias(lua_State *L)
+{
+    std::vector<std::string> alias;
+    Lua::GetVector(L, alias, 1);
+
+    if (!alias.empty())
+    {
+        std::string name = alias[0];
```
lethosor, line 3623: I think we should match the AddAlias() signature instead by taking two arguments: `name` (a string) and `arguments` (a list of arguments). It looks like you're retrieving the alias name from the first element of the list here, which doesn't match the C++ API.

library/LuaApi.cpp
Penumbra69: I have tested that, and it does crash. I'll take a look and find a way to ensure stability.

library/LuaApi.cpp
Penumbra69: I'll adjust.

library/LuaApi.cpp
lethosor: `luaL_checktype(L, 1, LUA_TTABLE)` before the GetVector call would probably work.

lethosor:
Which PR addresses this point?
> Enhance help entry with aliased command information

I don't see changes to `helpdb` in either.

library/LuaApi.cpp
```diff
@@ -3577,6 +3577,67 @@ static int internal_md5file(lua_State *L)
...
+    if (!ret.empty())
+        lua_pushstring(L, ret.c_str());
+    else
+        lua_pushnil(L);
+    return 1;
+}
+
+static int internal_listAliases(lua_State *L)
+{
+    auto aliases = Core::getInstance().ListAliases();
```
myk002, line 3597: If we add a `Lua::Push` override for `std::vector`, this function could be reduced to `Lua::Push(L, Core::getInstance().ListAliases()); return 1;`

test/library/alias.lua
```diff
@@ -0,0 +1,11 @@
+-- Simple tests to ensure the alias functions work as expected
+function test.aliases()
+    expect.eq(false, dfhack.internal.isAlias("foo"))
```
myk002, line 3: this can actually be flaky since it depends on the current active environment. these function tests allow you to specify a setup/teardown wrapper where you can save the current aliases, clear the alias list for the test, and restore the original aliases afterwards. See the orders plugin tests for an example.

library/LuaApi.cpp
Penumbra69: Nice - will look at this.

myk002:
I think that would be required, yeah, otherwise it wouldn't show up as an entry in `ls` and other places.

ab9rf:
This appears to be abandoned, with no action in over three years

PR #3431, opened 2023-05-29
[SDL2] temporarily disable faststart

myk002:
Just until DF is race-condition-free. prevents crashes for experimental branch users (since `faststart` is on by default)

### DIFF
--- a/plugins/faststart.cpp
+++ b/plugins/faststart.cpp
@@ -48,8 +48,10 @@ DFhackCExport command_result plugin_enable(color_ostream &out, bool enable)
 {
     if (enable != is_enabled)
     {
+#if 0
         if (!INTERPOSE_HOOK(prep_hook, logic).apply(enable))
             return CR_FAILURE;
+#endif
 
         is_enabled = enable;
     }


ab9rf:
faststart has _not_ been upstreamed. the `logic` method for `viewscreen_initial_prepst` still counts up to 4 for each step. while there have been changes to the code to the `logic` method, this is not part of them, and `faststart` is not (itself) responsible for the crashy behavior we've been seeing

i am categorically opposed to this PR being merged at this time

ab9rf:
@quietust and i have dissected the code in question and determined that there is fundamentally an internal race condition within DF that `faststart` makes _much_ more likely to be won by the "wrong" thread, leading to misoperation. ultimately this is a serious bug in Dwarf Fortress, (and i've reported it as such to Bay12 via the kitfox discord) and once Bay12 fixes it `faststart` will be safe to use, but also probably not as necessarily as it once was. at this time i'm going to recommend setting `faststart` to disabled by default, but not removing the tool as it will may conceivably still serve a purpose

quietust:
To clarify, the `initial_prep` viewscreen looks like it's been modified to spawn a separate thread which performs all of the relevant initialization as quickly as possible and updates the "percentage complete" variable in the viewscreen in order to drive the animation, but if the very first step took longer than 4 frames then it would spawn _another_ thread and things would basically crash horribly.

myk002:
So even if `faststart` has not been truly upstreamed, should we keep it or remove it? Does it add benefit above the now-async driving of the animation?

ab9rf:
it still adds a (small) benefit of updating the animation more frequently, and in theory trimming up to 4 frames off the total startup time

i would do nothing until we see the next update from bay12 and we see how the bug `faststart` exposed is fixed

PR #3435, opened 2023-05-30
UI: "divider" rendering rework

TaxiService:
by 'divider' i mean one of these horizontal or vertical lines that visually "divide" a frame.
Dwarf_Fortress_tyNzHqdCjv

Previously, these were rendered using a custom function within each script that used a divider.

This PR modifies the way a Panel is rendered, so that to achieve the same effect, you can just make a 1-width or 1-height Panel to get a vertical or horizontal divider respectively.

### gui.lua
- `make_frame` - now assigns a name to every tile in a border spritesheet, so we can access T pieces etc.
- `paint_frame` - a Frame with height/width equal to 1 will now render appropriate T-pieces at its extremities, and an horizontal/vertical bar between them. 
It will also draw a cross-piece if both height and width are equal to 1.
(Additionally, logic has been added that disables the `signature_pen` whenever a Panel's height and/or width is equal to 1.)

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -75,6 +75,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Lua
 - ``overlay.reload()``: has been renamed to ``overlay.rescan()`` so as not to conflict with the global ``reload()`` function. If you are developing an overlay, please take note of the new function name for reloading your overlay during development.
 - ``gui``: changed frame naming scheme to ``FRAME_X`` rather than ``X_FRAME``, and added aliases for backwards compatibility. (for example ``BOLD_FRAME`` is now called ``FRAME_BOLD``)
+- ``gui.make_frame()``: now maps every tile sprite from a border spritesheet.
+- ``gui.paint_frame()``: Panels with width or height equal to 1 will now draw a ``divider``.
+- ``buildingplan``: has been updated to make use of the new gui elements internally.
 - ``ensure_keys``: walks a series of keys, creating new tables for any missing values
 
 ## Removed
--- a/library/lua/gui.lua
+++ b/library/lua/gui.lua
@@ -899,13 +899,13 @@ GREY_FRAME = {
 -- The boundary used by the pre-steam DF screens.
 -- deprecated
 BOUNDARY_FRAME = {
-    frame_pen = to_pen{ ch = 0xDB, fg = COLOR_GREY, bg = COLOR_BLACK },
+    frame_pen = to_pen{ ch = 0xDB, fg = COLOR_GREY, bg = COLOR_BLACK }, -- ch=0xDB is "full block" (█)
     title_pen = to_pen{ fg = COLOR_BLACK, bg = COLOR_GREY },
     signature_pen = to_pen{ fg = COLOR_BLACK, bg = COLOR_GREY },
 }
 
 local BASE_FRAME = {
-    frame_pen = to_pen{ ch=206, fg=COLOR_GREY, bg=COLOR_BLACK },
+    frame_pen = to_pen{ ch=206, fg=COLOR_GREY, bg=COLOR_BLACK }, -- ch=206 is "box drawings double vertical and horizontal" (╬)
     title_pen = to_pen{ fg=COLOR_BLACK, bg=COLOR_GREY },
     inactive_title_pen = to_pen{ fg=COLOR_GREY, bg=COLOR_BLACK },
     signature_pen = to_pen{ fg=COLOR_GREY, bg=COLOR_BLACK },
@@ -920,14 +920,42 @@ local function make_frame(name, double_line)
     end
 
     local frame = copyall(BASE_FRAME)
+    -- external horizontal/vertical bars
     frame.t_frame_pen = to_pen{ tile=tp(1), ch=double_line and 205 or 196, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.l_frame_pen = to_pen{ tile=tp(7), ch=double_line and 186 or 179, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.b_frame_pen = to_pen{ tile=tp(15), ch=double_line and 205 or 196, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.r_frame_pen = to_pen{ tile=tp(9), ch=double_line and 186 or 179, fg=COLOR_GREY, bg=COLOR_BLACK }
+    -- external corners
     frame.lt_frame_pen = to_pen{ tile=tp(0), ch=double_line and 201 or 218, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.lb_frame_pen = to_pen{ tile=tp(14), ch=double_line and 200 or 192, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.rt_frame_pen = to_pen{ tile=tp(2), ch=double_line and 187 or 191, fg=COLOR_GREY, bg=COLOR_BLACK }
     frame.rb_frame_pen = to_pen{ tile=tp(16), ch=double_line and 188 or 217, fg=COLOR_GREY, bg=COLOR_BLACK }
+    -- external T-junctions
+    frame.tte_frame_pen = to_pen{ tile=tp(10), ch=double_line and 203 or 194, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.bte_frame_pen = to_pen{ tile=tp(11), ch=double_line and 202 or 193, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.lte_frame_pen = to_pen{ tile=tp(12), ch=double_line and 204 or 195, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.rte_frame_pen = to_pen{ tile=tp(13), ch=double_line and 185 or 180, fg=COLOR_GREY, bg=COLOR_BLACK }
+    -- internal horizontal/vertical bars (and cross junction)
+    frame.v_frame_pen = to_pen{ tile=tp(4), ch=double_line and 186 or 179, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.h_frame_pen = to_pen{ tile=tp(5), ch=double_line and 205 or 196, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.x_frame_pen = to_pen{ tile=tp(3), ch=double_line and 206 or 197, fg=COLOR_GREY, bg=COLOR_BLACK }
+    -- internal T-junctions
+    frame.tti_frame_pen = to_pen{ tile=tp(20), ch=double_line and 203 or 194, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.bti_frame_pen = to_pen{ tile=tp(19), ch=double_line and 202 or 193, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.lti_frame_pen = to_pen{ tile=tp(18), ch=double_line and 204 or 195, fg=COLOR_GREY, bg=COLOR_BLACK }
+    frame.rti_frame_pen = to_pen{ tile=tp(17), ch=double_line and 185 or 180, fg=COLOR_GREY, bg=COLOR_BLACK }
+    --[[ naming scheme for t-junctions:
+        the first letter means:
+            "t" for "top":    the T is not rotated, it looks like the character "T"
+            "b" for "bottom": the T is upside-down
+            "l" for "left":   the T is rotated 90 degrees counter-clockwise
+            "r" for "right":  the T is rotated 90 degrees clockwise
+        the second letter means:
+            "t" for "t-junction": this is a t-junction piece...
+        the third letter means:
+            "i" for "internal": this sprite is meant to be drawn inside the frame.
+            "e" for "external": this sprite is meant to be drawn on the frame itself.
+    --]]
     return frame
 end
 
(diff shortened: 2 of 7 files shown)


library/lua/gui.lua
```diff
@@ -920,14 +920,42 @@ local function make_frame(name, double_line)
...
+            "t" for "top":    the T is not rotated, it looks like the character "T"
+            "b" for "bottom": the T is upside-down
+            "l" for "left":   the T is rotated 90 degrees counter-clockwise
+            "r" for "right":  the T is rotated 90 degrees clockwise
+        the second letter means:
+            "t" for "t-junction": this is a t-junction piece...
+        the third letter means:
+            "i" for "internal": this sprite is meant to be drawn inside the frame.
+            "e" for "external": this sprite is meant to be drawn on the frame itself.
+    --]]
```
TaxiService, line 958: idk if this is necessary but ive included it for now

library/lua/gui.lua
```diff
@@ -955,19 +983,35 @@ INTERIOR_MEDIUM_FRAME = FRAME_INTERIOR_MEDIUM
...
+        signed = style.signature_pen
     end
-    dscreen.paintTile(rb_frame_pen or pen, x2, y2)
-    dscreen.fillRect(style.t_frame_pen or style.h_frame_pen or pen,x1+1,y1,x2-1,y1)
-    dscreen.fillRect(style.b_frame_pen or style.h_frame_pen or pen,x1+1,y2,x2-1,y2)
-    dscreen.fillRect(style.l_frame_pen or style.v_frame_pen or pen,x1,y1+1,x1,y2-1)
-    dscreen.fillRect(style.r_frame_pen or style.v_frame_pen or pen,x2,y1+1,x2,y2-1)
-    if style.signature_pen ~= false then
+
+    if signed ~= false then
```
TaxiService, line 1014: working on this `signed` thing made me realise that we don't need a whole new FRAME_INTERIOR_MEDIUM for planneroverlay... 🤦 i could've just turned off the signature_pen for _that_ particular panel... so expect PRs for that as well.

but anyway- this code basically only renders the `DFHack` thingy if the panel has a width/height > 1.

TaxiService (commented):
i didn't add a changelog yet because my idea initially was to also commit some changes to some buildingplan scripts... and i've committed some of those changes locally, but i haven't pushed them yet because on second thought it would be idiotic

and now i am having trouble locally with git, and i cant try to figure out how to fix them at the moment... so, for now here's just the code changes.

I tested them in-game and this all seemed to work, at least. ;_; (so far)

plugins/lua/buildingplan/pens.lua
```diff
@@ -1,38 +1,15 @@
...
-    local mb_texpos = dfhack.textures.getMediumBordersTexposStart()
-    HORI_LEFT_PEN = to_pen{tile=tp(mb_texpos, 12), ch=195, fg=COLOR_GREY, bg=COLOR_BLACK}
-    HORI_MID_PEN = to_pen{tile=tp(mb_texpos, 5), ch=196, fg=COLOR_GREY, bg=COLOR_BLACK}
-    HORI_RIGHT_PEN = to_pen{tile=tp(mb_texpos, 13), ch=180, fg=COLOR_GREY, bg=COLOR_BLACK}
-
-    local cp_texpos = dfhack.textures.getControlPanelTexposStart()
-    BUTTON_START_PEN = to_pen{tile=tp(cp_texpos, 13), ch='[', fg=COLOR_YELLOW}
-    BUTTON_END_PEN = to_pen{tile=tp(cp_texpos, 15), ch=']', fg=COLOR_YELLOW}
-    SELECTED_ITEM_PEN = to_pen{tile=tp(cp_texpos, 9), ch=string.char(251), fg=COLOR_YELLOW}
-
```
TaxiService, line 35: this thing displays a yellow checkmark in ascii mode, and a white masterwork-symbol in premium mode. 🤔 (it's actually one of the sprites from the control-panel spritesheet, but i only realized this after checking out the code)

like, i've removed the control-panel stuff completely, and right now i've set it to use the yellow checkmark in both ascii and premium modes.

If you think we should use a white masterwork symbol instead, we can do it, but whatever we do, it should be the same for both modes.

TaxiService (commented):
alright this is it.
The changes in gui.lua are the most significant.
The most significant changes in the other scripts are in `planneroverlay` and `filterselection`.
In both of them, the lines that would callback the draw_divider function, now simply sets the frame's style. The stuff modified in gui.lua (paint_frame mostly) now achieves the same as the now removed draw_divider functions.

other small tweaks in the other scripts were changing the old XYZ_FRAME names to the new format, and "fixing" an inconsistency i found while cleaning pens.lua.

plugins/lua/buildingplan/pens.lua
myk002: I'm fine with either as long as it is clear and consistent

docs/changelog.txt
```diff
@@ -75,6 +75,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Lua
 - ``overlay.reload()``: has been renamed to ``overlay.rescan()`` so as not to conflict with the global ``reload()`` function. If you are developing an overlay, please take note of the new function name for reloading your overlay during development.
 - ``gui``: changed frame naming scheme to ``FRAME_X`` rather than ``X_FRAME``, and added aliases for backwards compatibility. (for example ``BOLD_FRAME`` is now called ``FRAME_BOLD``)
+- ``gui.make_frame()``: now maps every tile sprite from a border spritesheet.
+- ``gui.paint_frame()``: Panels with width or height equal to 1 will now draw a ``divider``.
+- ``buildingplan``: has been updated to make use of the new gui elements internally.
```
myk002, line 80: the `gui` functions are undocumented, and the `buildingplan` change is internal, so I think we can get by without a changelog entry for this PR

library/lua/gui.lua
myk002: I certainly needed the explanation. I'd say keep it.

myk002: In fact, maybe add a little explanation about the difference between an external and internal t-junction (which I only now understand because I looked at the sprite sheet).

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -345,11 +345,13 @@ function PlannerOverlay:init()
     local main_panel = widgets.Panel{
         view_id='main',
         frame={t=1, l=0, r=0, h=14},
-        frame_style=gui.INTERIOR_MEDIUM_FRAME,
+        frame_style=gui.FRAME_MEDIUM,
         frame_background=gui.CLEAR_PEN,
         visible=self:callback('is_not_minimized'),
     }
 
+    main_panel.frame_style.signature_pen = false
```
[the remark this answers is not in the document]
myk002: as written, this is changing the `signature_pen` for every tool that uses `gui.FRAME_MEDIUM`, which is probably not what you want. Option 1 would be to use `copyall(gui.FRAME_MEDIUM)` and alter that copy. Option 2 would be to use `gui.FRAME_INTERIOR_MEDIUM` that already has the change. I'd go with option 2 and keep the `FRAME_INTERIOR_MEDIUM` in `gui`.

library/lua/gui.lua
```diff
@@ -955,19 +983,35 @@ INTERIOR_MEDIUM_FRAME = FRAME_INTERIOR_MEDIUM
...
-    dscreen.paintTile(style.lb_frame_pen or pen, x1, y2)
-    local rb_frame_pen = style.rb_frame_pen
-    if rb_frame_pen == FRAME_WINDOW.rb_frame_pen and not resizable then
-        rb_frame_pen = FRAME_PANEL.rb_frame_pen
+    local signed = false
+
+    if x1 == x2 and y1 == y2 then
+        dscreen.paintTile(style.x_frame_pen or pen, x1, y2)
+    elseif x1 == x2 then
+        dscreen.paintTile(style.tte_frame_pen or style.tti_frame_pen or pen, x1, y1)
```
myk002, line 991: it does not seem convenient to get an "interior"-style border with this logic. a frame would have to remove all the `t*e` fields so this logic would fall through to using the `t*i` fields.

what if:
- the border just had one set of t-junction variables (and maybe capitalize the `T` to distinguish it from "top"? e.g. `tT_frame_pen`, `bT_frame_pen`, etc.)
- we choose which t-junction tiles to use in `make_frame` based on an `interior` parameter
 would that work?

myk002:
good job with this PR, by the way. Your Lua coding has improved rapidly!

docs/changelog.txt
TaxiService: yaaaaaay :D

plugins/lua/buildingplan/planneroverlay.lua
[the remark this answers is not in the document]
TaxiService: > as written, this is changing the `signature_pen` for every tool that uses `gui.FRAME_MEDIUM`, 

oh god no °-° no, i didn't mean that at all, of course!
crap, i didn't even notice this. 🤦 i didn't think about checking other MEDIUM frames tbh.

dammit! I always forget all these intricacies of Lua. I mean-- i presumed this would only affect the "copy" that planneroverlay is using, but i see how that might not really be what is happening D:

yeah no ok, we should probably go back to using INTERIOR_MEDIUM. I'll edit that back in the next update to the PR

library/lua/gui.lua
TaxiService: oh you mean the `style.tte_frame_pen or style.tti_frame_pen`?

yeah, like, i put those `or style.tti_frame_pen` there mostly to have this conversation. 🤣 
written like that, it would require what you described. Which is of course inconvenient af.

I too thought about adding some sort of "interior" flag or something, but i wanted to ask you if we could even do that. 🤔 what would adding such a flag entail, for the rest of dfhack? 
(like-- what would we break, what would we have to update?)

> what if:
> 
>     * the border just had one set of t-junction variables (and maybe capitalize the `T` to distinguish it from "top"? e.g. `tT_frame_pen`, `bT_frame_pen`, etc.)
> 
>     * we choose which t-junction tiles to use in `make_frame` based on an `interior` parameter
>       would that work?

This is the point i wanted to get to: 
i think `make_frame` should output a FRAME with both internal and external Ts, and *then* the "interior" flag should be settable from each Panel individually. 🤔 

Meaning we won't end up with FRAME_MEDIUM_INTERNAL and FRAME_MEDIUM_EXTERNAL, but just a FRAME_MEDIUM that could be set to use both "internal" or "external" sprites based on an additional field in the Panel/Widget or whatever. 🤔 on a case-by-case basis.

TaxiService: additionally, great idea about using capital Ts to differentiate from t(op). Will definitely do that 👍

library/lua/gui.lua
TaxiService: right. yeah, will do. 
i've taken their obviousness for granted because i've been analyzing/staring at/editing those spritesheets for a long time, but it's such an unintuitive concept 🤣 
In ascii mode there's no difference between internal/external either!

library/lua/gui.lua
myk002: > Meaning we won't end up with FRAME_MEDIUM_INTERNAL and FRAME_MEDIUM_EXTERNAL, but just a FRAME_MEDIUM that could be set to use both "internal" or "external" sprites based on an additional field in the Panel/Widget or whatever.  on a case-by-case basis.

what are the pros and cons of each option, as you see it?

TaxiService: let's see... 

### Approach 1: a single FRAME_X per border texture
#### Pros
- a FRAME would contain a reference to \*all\* sprites in the spritesheet.
(imo this is very important, because it makes a FRAME a "complete set". Meaning that if for whatever reason you need sprite X, you know it's in there and that you can access it.)

#### Cons
- a FRAME would have 20 sprite references instead of 16, making it a bit "bigger" in memory than approach 2, i guess?
- we will have to implement some logic for choosing which sprites to use, which does not exist yet. 🤔

-------

### Approach 2: two FRAME_X_INTERNAL / FRAME_X_EXTERNAL per border texture
#### Pros
- probably easier to implement? 
- a FRAME would have up to 16 sprite references in it, making the whole structure smaller i guess?

#### Cons
- no FRAME would contain all sprites of a border spritesheet. 
- you need to have two `make_frame()` functions, or otherwise make it do "different things"
- you have two FRAMEs where you could've had just one. 🤔 also 2\*16  > 1\*20 sprites, negating one of the pros 🤦 
- if for any reason part of your frame needs to switch from internal to external, you must come up with weird solutions.

-------
#### more considerations
in gui/launcher there's an example of a divider with internal and external ends
Untitled-1
(well, at the moment the "external" end is not even there, but you get the idea)

like, idk why, but i feel that having some way of saying:
"make a 1-tall frame from here to there, and have its rightmost end be `internal`."
would sound better than
"make a 1-tall external frame from here to halfway-there, then 
 make another 1-tall internal frame from halfway-there to there"
or... something like that. °-°


though, i must stress this-- 
these are kinda uneducated opinions of a person that does not know much at all about programming, its best practices, how to be efficient at it, and stuff like that! 😫 i'm just like... "having all the parts in one container is neat!!"

myk002: table sizes between 16 and 20 are negligible. these frames will have 1 copy that all windows share. size won't count for pro or con here.

switching between internal and external (if we were to want to do so, and I don't really understand the case where we'd want to do this) would be easy in either option, I'd think. for option 1 we'd flip a flag and for option 2 we'd assign a different `frame-style`.

from a programming perspective, option 2 is easier. consolidating the logic in `make_frame` is easier than making dynamic rendering decisions at draw time.

it's the flexibility angle that I want to understand more. in order to handle the `gui/launcher case`, we'd need to use T junctions from multiple sprite sheets since they're different widths. We'd need to construct a frankenframe for this specific use case. possible, but not elegant. if this will be a common use case, we'd want a better solution for it.

From a design perspective, in what cases would we want a single frame to have some edges be internal and others be external?

PR #3437, opened 2023-06-01
reinstated Buildings.setOwner

myk002:
now restricted to civzones since buildings can no longer have owners

### DIFF
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -350,46 +350,45 @@ df::specific_ref *Buildings::getSpecificRef(df::building *building, df::specific
     return findRef(building->specific_refs, type);
 }
 
-bool Buildings::setOwner(df::building *bld, df::unit *unit)
+bool Buildings::setOwner(df::building_civzonest *bld, df::unit *unit)
 {
     CHECK_NULL_POINTER(bld);
-/* TODO: understand how this changes for v50
-    if (!bld->is_room)
-        return false;
-    if (bld->owner == unit)
+
+    if (bld->assigned_unit == unit)
         return true;
 
-    if (bld->owner)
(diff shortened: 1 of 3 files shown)


library/modules/Buildings.cpp
```diff
@@ -350,46 +350,45 @@ df::specific_ref *Buildings::getSpecificRef(df::building *building, df::specific
...
-/* TODO: understand how this changes for v50
-    if (!bld->is_room)
-        return false;
-    if (bld->owner == unit)
+
+    if (bld->assigned_unit == unit)
         return true;
 
-    if (bld->owner)
+    df::building * pbld = dynamic_cast<df::building*>(bld);
```
lethosor, line 360: This needs to be virtual_cast, not dynamic_cast, since this is a `df` class. I suspect `pbld` is null when used below.

myk002: it works when tested, but will change to virtual_cast for consistency

lethosor: Well, it's probably always returning null, which means `linear_index(blist, pbld)` would return -1... that would get converted to UINT_MAX in the vector_erase_at() calls, which means the erase would typically silently do nothing.

Or are you saying that dynamic_cast is returning something non-null? As far as my understanding of our type system goes, that can't happen, because our df::building has completely different RTTI from DF's.

myk002: Looking back on this, I believe the answer is that I didn't test it properly the first time. This code path is only relevant when another unit is already assigned to the zone, where my test case was starting from a fresh zone.

I have since verified that both previous unit and spouse are correctly removed from the zone with the virtual_cast, so I think your review has saved us from a bug.

PR #3439, opened 2023-06-01
Textures.cpp: add tile resolution arguments to `load_textures()`

TaxiService:
this pr makes pathable look like this:
[Dwarf_Fortress_mAuqleDA4E.webm](https://github.com/DFHack/dfhack/assets/5853667/1c64d8f6-fe79-4a33-a2fa-bf2ee1773cfe)

more specifically, the `load_textures()` function now needs to have the resolution of a spritesheet's tile passed into it.
this lets us choose if we want an 8x12-tile 'sheet or a 32x32 'sheet. (or even whatever other resolution, i guess!)

and to test this all, i've added a new texture and set gui/pathable to use it.


TaxiService:
NOTE: despite the branch being called `sdl2-test`, there is nothing related to sdl2 in this PR. 🤔 😩 
originally i meant to try cloning locally the "sdl2" branch from the main repo or whatever, but i couldn't do it. 🤷
another series of messes/misguidedness led me to have this PR like this, but incredibly it seems to work even without sdl2!!

library/modules/Textures.cpp
```diff
@@ -64,7 +65,7 @@ const uint32_t TILE_WIDTH_PX = 8;
 const uint32_t TILE_HEIGHT_PX = 12;
 
 static size_t load_textures(color_ostream & out, const char * fname,
-                            long *texpos_start) {
+                            long *texpos_start, int tile_w, int tile_h) {
```
TaxiService, line 68: i am really not fluent in c++... at all. °-°
are these argument/parameter/variable names okay? 

also, is it ok to use `int` as types, here? or should i have used `uint32_t`?

originally i meant to do something like...
```cpp
int tile_w = 8, int tile_h = 12)
```
but i really dont know anything about c++ and didn't know (or wanted to find out) if that actually works. 🤔

lethosor:
> NOTE: despite the branch being called `sdl2-test`, there is nothing related to sdl2 in this PR.

But doesn't the issue you're solving only occur on the SDL2 builds? If so, this should be targeting our `sdl2` branch.

TaxiService:
ah crap. that's right 🤦 i committed this here because: 
a) i couldnt get a working copy of the sdl2 branch locally with git ;_;
b) i thought "well, if i fix this here, when sdl2 will be merged it would work there too!"

but now i guess... i dont really know if the sdl2 branch will be "merged" into develop. 🤔 is that going to happen?

lethosor:
This code as-is is only useful on the sdl2 branch, so it should be merged there. You can change the target branch with the "Edit" button at the top of the PR, although it might also need to be rebased. Not quite sure what you mean by "couldn't get a working copy of the sdl2 branch" - if you need git help (either checking out the sdl2 branch locally or rebasing), Discord is probably the better place to figure it out.

TaxiService:
ok, sorry but i'm just gonna close this PR and do a new one without all this mess °-°

Issue #3098, opened 2023-03-25
stockpile presets (?) for butcherable races only


ab9rf:
Now that corpse stockpiles are filterable, it would be useful to have some way to quickly create a stockpile that only accepts _butcherable_ corpses (or at least corpses for butcherable races).

Note that this can't exactly be a static preset because which races are butcherable depends on the fortress entity's ethics.

myk002:
what is the code for determining if the race is butcherable? And are there other examples of properties that we might want to filter on?

Cubittus:
I'd find these useful as presets:
- Caged creatures that are trainable
- Caged creatures that are butcherable
- Any other caged creatures (eg for pitting)
- Corpses/bodyparts that can be butchered (ie not intelligent and not too small)
- Corpses/bodyparts that cannot be butchered
- Seeds that are plantable in Temperate biomes
- Seeds that are plantable in Tropical biomes
- Seeds that are plantable in Temperate and Tropical biomes
- Seeds that are plantable Underground
- Seeds that can only be eaten/cooked (acorns etc)
- Seed that are useless (inedible tree seeds)
- Plant products that can be brewed
- Plant products that can be processed (perhaps split by method/dye/flour etc)
- Plant products that can only be cooked

PR #3461, opened 2023-06-07
[stockpiles] add moar presets

myk002:
barrels and organic

two presets that came up as needed when porting Dreamfort


TaxiService:
in my lazy newb's starter pack installation (.47.05), i remember having some pre-set stockpile settings! Were those included in dfhack or were they added by LNP?

in any case, while testing with your latest stocks build, i have exported some stockpile settings as well.
In this screenshot, on the left i have my .47 stockpiles, on the right the ones i generated in .50

image

some stockpiles i think would be great for anyone are the:
- underground seeds
- above-ground seeds
ideally, this could further be separated into: 
edible/inedible, 
cookable/non-cookable, 
plantable/non-plantable
- probably a general "animal-cooking-ingredient" and "plant-cooking-ingredient" could be useful as well

now, if we could "perform operations" on stockpiles, like "subtract this stockpile's stuff from this other stockpile", then i'd say we'd need more presets. 🤔 
for example, an "artifact-quality" stockpile, in which the "artifact" quality is selected in *every* possible place; so that you could then "subtract" its settings from another existing stockpile to prevent artifacts being stocked in it. 🤔

myk002:
a general artifacts preset (and masterworks) is an interesting idea. currently, you can subtract artifacts from an existing pile, but the syntax is long and unwieldy. a general "no artifacts" preset is an excellent idea, and one I can use in dreamfort to tidy things up too.

for example:
```
pd{name="Other meltables" take_from="Metalworker quantum,Bar/military feeder"}:-cat_weapons/other/,core/masterful,core/artifact-cat_armor/other/,core/masterful,core/artifact-bronzeweapons-bronzearmor-steelweapons-steelarmor(7x3)
```
would become
```
pd{name="Other meltables" take_from="Metalworker quantum,Bar/military feeder"}:-cat_weapons/other/-cat_armor/other/-bronzeweapons-bronzearmor-steelweapons-steelarmor-artifacts-masterworks(7x3)
```
which is certainly cleaner (if still long)

programmatic presets (e.g. seeds that are edible or plantable in the current biomes) are on the roadmap. I've already implemented "tameable" as a property filter for animals. Those ideas should probably go on #3098

TaxiService:
> programmatic presets (e.g. seeds that are edible or plantable in the current biomes) are on the roadmap. I've already implemented "tameable" as a property filter for animals. Those ideas should probably go on https://github.com/DFHack/dfhack/issues/3098

oh wow. I wasn't even thinking about that! 
i meant a stockpile setting with all the "Farmable: Yes" seeds from [this list](https://dwarffortresswiki.org/index.php/Seed#Seed_List), one with all the "Farmable: No", and so on.

programmatic seed stockpile presets would be pretty cool! But \*personally\* i would be fine with just having a general "biome-agnostic" 'pile. :p I mean, most of it would be full of seeds that i found _in the biome_ i am in anyway... right?

i'd much rather be able to do stuff like:
- set stockpile to "inedible_seeds"
- subtract "underground_seeds"
- add "non-farmable_seeds"

to get specific settings-combinations very easily!

PR #3465, opened 2023-06-12
allow wide-character directories

mikerenfro:
As discussed in Discord cpp channel. Fix should support accented characters in directory names.


myk002 (commented):
I think more needs to be done here. The function signatures for `_mkdir()` and `_wmkdir()` differ:
```cpp
int _rmdir(
   const char *dirname
);
int _wrmdir(
   const wchar_t *dirname
);
```
which means you'd have to translate between utf-8 (or is it cp-437 here?) and utf-16
in fact, if path *is* cp-437, then the easier solution might be to translate into utf-8.

myk002:
So perhaps the real solution here is to translate from CP437 to windows-1252

ab9rf:
> So perhaps the real solution here is to translate from CP437 to windows-1252 and keep the existing call to _mkdir.

this may cause problems for people whose primary language setting is a language that does not use cp-1252, such as japanese or arabic

PR #3474, opened 2023-06-15
[sdl2] ensure we get all string input

### LINKED ISSUE #3433: Figure out why keybindings and text entry are weird in SDL2
most symbols don't work, like underscore or shift-number, until their interface keys are re-bound to the keypress (which changes, for example, `RParen` to `Shift + 0`)

### LINKED ISSUE #3471: SDL2 Experimental version keyboard input bug in GUI terminal of DFhack
I found a input bug in the sdl2 experimental version of DFhack. The bug appears when using the DFhack GUI terminal. 

When you use a swedish keyboard the / among all shift+number key keys and § (tilde) keys dont work. 

However, ctrl+alt+D key opens the DFhack gui terminal without much issue.

When I swap to ENG keyboard language the ` (tilde) key works but the shift+number keys dont. 

It does not seem to be a dwarf fortress native bug since the keypresses work in Dwarf fortress stockpile search and such 

Currently I am using the "show" command to open the console application window. This window seems to work fine with all kinds of text input as long as I use Windows 11's windows-terminal. 

See this page for a guide on setting windows-terminal as default on W11:
https://devblogs.microsoft.com/commandline/windows-terminal-as-your-default-command-line-experience/

I'm unsure how to do it on W10 but someone somewhere probably have a guide for it

myk002:
This only gets the first character in order to keep the logic simple, but I have yet to find a case where there is more than one character in the buffer (maybe with a system IME?).

If we have to loop, then we have to make multiple calls to `feed`/`onInput`, *without* duplicating mouse clicks or other forms of input. That would get hairy. Let's hope the single character approach is sufficient.

There are no other plugins that currently interpose feed, but if we add any more/add any more back, we need to make sure to add in the `Screen:add_text_keys()` wrapper.

Fixes #3433
Fixes #3471

### DIFF
--- a/library/include/modules/Screen.h
+++ b/library/include/modules/Screen.h
@@ -229,6 +229,9 @@ namespace DFHack
         DFHACK_EXPORT bool hasActiveScreens(Plugin *p);
         DFHACK_EXPORT void raise(df::viewscreen *screen);
 
+        // returns a new set with text interface keys from the text buffer added in (if any)
+        DFHACK_EXPORT std::set<df::interface_key> add_text_keys(const std::set<df::interface_key>& keys);
+
         /// Retrieve the string representation of the bound key.
         DFHACK_EXPORT std::string getKeyDisplay(df::interface_key key);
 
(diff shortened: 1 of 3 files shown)


myk002:
merging so we can put out a beta build and test with a wider audience

PR #3475, opened 2023-06-15
[sdl2] prefer symbols to descriptions for hotkey names

myk002:
prints "?" instead of "Question"
prints "U" instead of "Shift + u"

in short; it makes the hotkey hints on our UIs readable again

Before:

image
image
image

After:

image
image
image

### DIFF
--- a/library/modules/Screen.cpp
+++ b/library/modules/Screen.cpp
@@ -588,10 +588,12 @@ void Hide::merge() {
 
 string Screen::getKeyDisplay(df::interface_key key)
 {
-    if (enabler)
-        return enabler->GetKeyDisplay(key);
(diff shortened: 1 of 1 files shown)


lethosor (commented):
My understanding is that Putnam changed the display of shift+letter hotkeys in SDL2 DF recently - not sure if those changes are in 50.09-sdl2-2 or not. It may be worth checking how this was handled in libgraphics to see if we still need this patch.

myk002:
Yeah, I'll have to double check. She said she was going to update it, but we need to make sure that all the existing cases are covered. She may just do upper case letters, whereas this patch handled symbols too

PR #3481, opened 2023-06-17
get compiling with c++-20

myk002:


depends/CMakeLists.txt
```diff
@@ -1,14 +1,26 @@
 # list depends here.
 add_subdirectory(lodepng)
-add_subdirectory(lua)
 add_subdirectory(md5)
+
+add_subdirectory(lua)
```
lethosor, line 5: (these were sorted alphabetically)

library/CMakeLists.txt
```diff
@@ -377,6 +377,9 @@ if(WIN32)
     set_target_properties(dfhack-client PROPERTIES COMPILE_FLAGS "/FI\"Export.h\"" )
 else()
     set_target_properties(dfhack PROPERTIES COMPILE_FLAGS "-include Export.h" )
+    # required because of protobuf headers
+    target_compile_options(dfhack
+        PUBLIC -Wno-deprecated-declarations -Wno-restrict)
```
lethosor, line 382: Does setting these flags as PUBLIC on the protobuf targets not work? I thought something along those lines was supposed to work.

If this is required for DFHack, is it also required for plugins?

myk002: I believe it's required here because while plugins include protobuf headers, they do not directly link against the protobuf libs, so they don't pick up the flags.

A better solution might be to not include protobuf headers from DFHack headers

myk002: > If this is required for DFHack, is it also required for plugins?

It's actually not required for DFHack, since it picks up the flags from protobuf. I could have set the DFHack flags as INTERFACE instead of PUBLIC

depends/CMakeLists.txt
myk002: done. I was separating them out to mirror the new logic below, but that's not really necessary to do

library/CMakeLists.txt
myk002: changed to INTERFACE

lethosor: > It's actually not required for DFHack

I'm confused, then, why set the flags on the `dfhack` target at all? Is the issue that plugins won't pick up on INTERFACE flags set on the `protobuf` target, but will pick up on flags set on `dfhack`?

I'm also not a huge fan of disabling warnings globally due to an issue in a third-party dependency. Since we have full control over our copies of the protobuf headers, would we be able to add `#pragma GCC diagnostic ignored` and `#pragma GCC diagnostic pop` around the relevant bits? Or maybe this would work, which we use in DF headers to silence most warnings - unsure if this covers the warnings in question or not:
```c
#ifdef __GNUC__
#pragma GCC system_header
#endif
```

myk002:
That's a good idea. I don't like the way this PR looks right now either. Feels dirty.

The reason that DFHack needs the INTERFACE bit right now is because the protobuf headers get included indirectly by all plugins, but most don't depend on protobuf directly so they don't get the flags from there.

This means that we should also look into *why* one of our DFHack headers is including protobuf instead of just our cpp files.

lethosor:
RemoteClient.h and RemoteTools.h both include a generated .pb.h file, which includes protobuf headers. Those dependencies don't appear easy to rip out to me, unfortunately. A couple things might be possible to forward-declare, but not all of them.

myk002:
I was able to remove the inclusion of the Remote* header files from Core.h, which allowed us to avoid including protobuf directly in everything. Also your suggestion of `#pragma GCC system_header` worked like a charm.

myk002:
I moved the command_result enum from `RemoteClient.h` into the less externally-polluted but equally-widely-included `Export.h`. The choice was relatively arbitrary, though, and we can move it somewhere else if there's an obviously appropriate place.

library/include/Export.h
```diff
@@ -69,3 +69,17 @@ distribution.
...
+    {
+        CR_LINK_FAILURE = -3,    // RPC call failed due to I/O or protocol error
+        CR_NEEDS_CONSOLE = -2,   // Attempt to call interactive command without console
+        CR_NOT_IMPLEMENTED = -1, // Command not implemented, or plugin not loaded
+        CR_OK = 0,               // Success
+        CR_FAILURE = 1,          // Failure
+        CR_WRONG_USAGE = 2,      // Wrong arguments or ui state
+        CR_NOT_FOUND = 3         // Target object not found (for RPC mainly)
+    };
+}
```
lethosor, line 85: Can this go in Core.h instead? Export.h is really just for macros that affect symbol visibility (well, and Wformat, which should probably move too).

Not familiar with what dependencies exist here. I know Export.h is convenient because it's always included.

library/modules/Gui.cpp
```diff
@@ -182,23 +182,23 @@ DEFINE_GET_FOCUS_STRING_HANDLER(dwarfmode)
     if (game->main_interface.info.open) {
         newFocusString = baseFocus;
         newFocusString += "/Info";
-        newFocusString += "/" + enum_item_key(game->main_interface.info.current_mode);
+        newFocusString += '/' + enum_item_key(game->main_interface.info.current_mode);
```
lethosor, line 185: Are you sure these changes are correct? I would interpret `char` + anything as integer math, not string concatenation.

library/include/Export.h
lethosor: Hm, of the headers that use `command_result`, `RemoteClient.h` and `RemoteTools.h` are the only two that don't include `Core.h` already. But `Core.h` includes one of them, `RemoteClient.h`, so we can't introduce a circular include there.

I wouldn't be opposed to a separate `Command.h`. I guess there is also `Types.h`, which looks mostly outdated to me, and `Internal.h`, which at first glance seems _similar_ to `Export.h` but may cause some behavioral changes that we wouldn't want to force on _everything_ that includes DFHack headers (such as disabling warnings and forcing large file support on Linux. Worth noting that `Types.h` is currently only included by `.cpp` files within the `library` folder.)

library/modules/Gui.cpp
myk002: Tested and it works out. std::string has an overload for this, just like +=

lethosor: Ok, looks like there has been an `operator+(char, std::string)` available since C++11: https://en.cppreference.com/w/cpp/string/basic_string/operator%2B

Wondering why the changes, though, since they did catch me off guard. They were bundled in a protobuf commit but looked unrelated to protobuf to me.

library/include/Export.h
myk002: Note that Core.h included RemoteClient.h *because* of command_result. It might be ok to include the other way now

lethosor: Ah that would make sense. I didn't really dive into _why_ those headers were being included.

library/modules/Gui.cpp
myk002: Otherwise we fail on -Wrestrict. It's "related" to the protobuf change since we lost the absorbed -Wnorestrict

lethosor: ...how the heck does `"a" + std::string("b")` trigger `-Wrestrict`? I realize that's calling `operator+()` under the hood, but that doesn't have `restrict` qualifiers on either of its arguments.

myk002: It is truly one of the more esoteric error messages I've seen. I'm not at the computer right now, but the issue was that it constructs a temporary string with a corrupt (? Maybe just ill -defined?) buffer size

lethosor: No luck reproducing with
```cpp
#include <string>

template<typename T>
inline std::string enum_item_key(T val) {
    return "foo";
}

int main() {
    std::string s = "a";
    s += "b" + enum_item_key(3);
}
```

I'll see if I can reproduce with the full codebase.

lethosor: It only seems to happen for `"/"` literals inside the `switch` block. Outside, either `"/"` or `'/'` is fine. Weird.

```
In static member function ‘static constexpr std::char_traits<char>::char_type* std::char_traits<char>::copy(char_type*, const char_type*, std::size_t)’,
    inlined from ‘static constexpr void std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::_S_copy(_CharT*, const _CharT*, size_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’ at /usr/include/c++/12/bits/basic_string.h:423:21,
    inlined from ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Allocator>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::_M_replace(size_type, size_type, const _CharT*, size_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’ at /usr/include/c++/12/bits/basic_string.tcc:532:22,
    inlined from ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::replace(size_type, size_type, const _CharT*, size_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’ at /usr/include/c++/12/bits/basic_string.h:2171:19,
    inlined from ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::insert(size_type, const _CharT*) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’ at /usr/include/c++/12/bits/basic_string.h:1928:22,
    inlined from ‘constexpr std::__cxx11::basic_string<_CharT, _Traits, _Allocator> std::operator+(const _CharT*, __cxx11::basic_string<_CharT, _Traits, _Allocator>&&) [with _CharT = char; _Traits = char_traits<char>; _Alloc = allocator<char>]’ at /usr/include/c++/12/bits/basic_string.h:3541:36,
    inlined from ‘bool DFHack::Gui::autoDFAnnouncement(df::report_init, std::string)’ at dfhack/library/modules/Gui.cpp:1692:37:
/usr/include/c++/12/bits/char_traits.h:431:56: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ accessing 9223372036854775810 or more bytes at offsets [2, 9223372036854775807] and 1 may overlap up to 9223372036854775813 bytes at offset -3 [-Werror=restrict]
  431 |         return static_cast<char_type*>(__builtin_memcpy(__s1, __s2, __n));
      |                                        ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
```

lethosor: Preliminary searching around suggests that this is a case of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104336 or https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329.

Someone else ran into the same issue in https://github.com/DFHack/dfhack/issues/2250#issuecomment-1196549459.

I guess this only happens for me (and you?) in C++20 with GCC 12? I wasn't able to get it to fail earlier, and curiously, CI didn't fail before your change either. (It is running GCC 12.1, vs 12.2 for me locally)

myk002: The lack of CI failure may just be because I didn't push until it built successfully locally. This being a bug makes sense since the code we had looked just fine.

library/include/Export.h
myk002: done

library/CMakeLists.txt
```diff
@@ -408,8 +408,8 @@ endif()
 target_link_libraries(dfhack protobuf-lite clsocket lua jsoncpp_static dfhack-version ${PROJECT_LIBS})
 set_target_properties(dfhack PROPERTIES INTERFACE_LINK_LIBRARIES "")
 
-target_link_libraries(dfhack-client protobuf-lite clsocket jsoncpp_static)
-target_link_libraries(dfhack-run dfhack-client)
+target_link_libraries(dfhack-client protobuf-lite clsocket jsoncpp_static dfhack)
+target_link_libraries(dfhack-run dfhack-client dfhack)
```
lethosor, line 412: Coming from https://github.com/DFHack/dfhack/pull/3524#issuecomment-1620621678... why's the link-time dependency necessary now? As far as I can tell, the only change to RemoteClient is that it pulls in the `command_result` enum from Core.h now, which shouldn't introduce a link-time dependency. Attempting to call any functions from Core.h would introduce such a dependency, but we can avoid that.

myk002: yeah, I wasn't clear on why this dependency showed up either. let me look into this a little more.

myk002: the link error only occurs on Windows:
```gcc
FAILED: library/dfhack-client.dll library/dfhack-client.lib 
: && /opt/cmake/bin/cmake -E vs_link_dll --intdir=library/CMakeFiles/dfhack-client.dir --rc=/opt/msvc/bin/x64/rc --mt=/opt/msvc/bin/x64/mt --manifests  -- /opt/msvc/bin/x64/link /nologo library/CMakeFiles/dfhack-client.dir/RemoteClient.cpp.obj library/CMakeFiles/dfhack-client.dir/ColorText.cpp.obj library/CMakeFiles/dfhack-client.dir/MiscUtils.cpp.obj library/CMakeFiles/dfhack-client.dir/Error.cpp.obj library/CMakeFiles/dfhack-client.dir/proto/Basic.pb.cc.obj library/CMakeFiles/dfhack-client.dir/proto/BasicApi.pb.cc.obj library/CMakeFiles/dfhack-client.dir/proto/CoreProtocol.pb.cc.obj library/CMakeFiles/dfhack-client.dir/Console-windows.cpp.obj  /out:library/dfhack-client.dll /implib:library/dfhack-client.lib /pdb:library/dfhack-client.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO  depends/protobuf/protobuf-lite.lib  depends/clsocket/clsocket.lib  depends/jsoncpp-sub/src/lib_json/jsoncpp.lib  depends/zlib/lib/zlib.lib  Ws2_32.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib  && :
LINK: command "/opt/msvc/bin/x64/link /nologo library/CMakeFiles/dfhack-client.dir/RemoteClient.cpp.obj library/CMakeFiles/dfhack-client.dir/ColorText.cpp.obj library/CMakeFiles/dfhack-client.dir/MiscUtils.cpp.obj library/CMakeFiles/dfhack-client.dir/Error.cpp.obj library/CMakeFiles/dfhack-client.dir/proto/Basic.pb.cc.obj library/CMakeFiles/dfhack-client.dir/proto/BasicApi.pb.cc.obj library/CMakeFiles/dfhack-client.dir/proto/CoreProtocol.pb.cc.obj library/CMakeFiles/dfhack-client.dir/Console-windows.cpp.obj /out:library/dfhack-client.dll /implib:library/dfhack-client.lib /pdb:library/dfhack-client.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO depends/protobuf/protobuf-lite.lib depends/clsocket/clsocket.lib depends/jsoncpp-sub/src/lib_json/jsoncpp.lib depends/zlib/lib/zlib.lib Ws2_32.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:library/dfhack-client.dll.manifest" failed (exit code 96) with the following output:
   Creating library library/dfhack-client.lib and object library/dfhack-client.exp
RemoteClient.cpp.obj : error LNK2019: unresolved external symbol "private: __cdecl DFHack::Core::Core(void) __ptr64" (??0Core@DFHack@@AEAA@XZ) referenced in function "public: static class DFHack::Core & __ptr64 __cdecl DFHack::Core::getInstance(void)" (?getInstance@Core@DFHack@@SAAEAV12@XZ)
RemoteClient.cpp.obj : error LNK2019: unresolved external symbol "private: __cdecl DFHack::Core::~Core(void) __ptr64" (??1Core@DFHack@@AEAA@XZ) referenced in function "??__Finstance@?1??getInstance@Core@DFHack@@SAAEAV12@XZ@YAXXZ" (??__Finstance@?1??getInstance@Core@DFHack@@SAAEAV12@XZ@YAXXZ)
library\dfhack-client.dll : fatal error LNK1120: 2 unresolved externals
```

because this is in the header file:
```cpp
        static Core& getInstance()
        {
            static Core instance;
            return instance;
        }
```

I don't see why this is in the header file and not the cpp file. I'll try to move it.

library/include/modules/Buildings.h
```diff
@@ -237,14 +237,20 @@ DFHACK_EXPORT std::string getRoomDescription(df::building *building, df::unit *u
  * starting at the top left and moving right, row by row,
  * the block's items are checked for anything on the ground within that stockpile.
  */
-class DFHACK_EXPORT StockpileIterator : public std::iterator<std::input_iterator_tag, df::item>
```
lethosor, line 240: Why did this need to change? New version seems harder to maintain to me.

myk002: iterator subclasses are deprecated. I had to migrate to the new way of doing things.
https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

lethosor: Sounds good. I think I was confusing this with `std::vector::iterator`, which I don't think we can use here.

CMakeLists.txt
```diff
@@ -226,9 +227,7 @@ if(UNIX)
     ## flags for GCC
     # default to hidden symbols
     # ensure compatibility with older CPUs
-    # enable C++11 features
     add_definitions(-DLINUX_BUILD)
-    add_definitions(-D_GLIBCXX_USE_C99)
```
lethosor, line 231: Heads-up that this came from 7bbfd456d3, so we may be breaking compatibility with older systems here.

lethosor (approved):
LGTM from static review

PR #3483, opened 2023-06-18
add unsuspend map textures

### LINKED ISSUE #3438: Map annotation tiles for SDL2

myk002:
Closes #3438


myk002:
I traced out some letters in gimp, but maybe you could do better? The red 'X', in particular, should be larger than the yellow one.

PR #3498, opened 2023-06-25
Fix so failed accept doesn't terminate the server

orrche:
On my steam install the accept function failes with SocketEwouldblock, this doesn't fix that but the fail would terminate the RemoteServer completely instead of waiting for the next attempt to establish an connection.

### DIFF
--- a/library/RemoteServer.cpp
+++ b/library/RemoteServer.cpp
@@ -476,11 +476,14 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
     CActiveSocket *client = nullptr;
 
     try {
-        while ((client = server.socket.Accept()) != NULL)
+        while (true)
         {
-            BlockGuard lock;
-            ServerConnection::Accepted(client);
-            client = nullptr;
+            if ((client = server.socket.Accept()) != NULL)
+            {
+                BlockGuard lock;
+                ServerConnection::Accepted(client);
+                client = nullptr;
+            }
         }
     } catch(BlockedException &) {
         if (client)


library/RemoteServer.cpp
```diff
@@ -476,11 +476,14 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
     CActiveSocket *client = nullptr;
 
     try {
-        while ((client = server.socket.Accept()) != NULL)
+        while (true)
         {
-            BlockGuard lock;
-            ServerConnection::Accepted(client);
-            client = nullptr;
+            if ((client = server.socket.Accept()) != NULL)
```
ab9rf, line 481: my main concern here that you are not guaranteed that `socket` will be in a valid state after `Accept` fails. if the socket is an invalid state, all subsequent calls to `Accept` will fail, which will lead to livelock

you need to verify that the socket is still in a state able to accept new connections after a failure before making another attempt; if you can't do that you have to close that listener and create a new one

ab9rf (changes requested):
in addition to the livelock issue mentioned, this needs a changelog

library/RemoteServer.cpp
orrche: Ouch, committed an corrected version that checks if the socket is valid before attempting to accept, not entirely sure that is enough tho. 

Added an error output as well so I can diagnose my actual problem better.

library/RemoteServer.cpp
```diff
@@ -476,11 +476,16 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
...
+            {
+                BlockGuard lock;
+                ServerConnection::Accepted(client);
+                client = nullptr;
+            } else {
+                std::cerr << "Accepting connection error: " << server.socket.DescribeError() << std::endl;
+            }
         }
     } catch(BlockedException &) {
```
myk002, line 490: This should be done with the debug logging system so the output verbosity is filterable.

```cpp
WARN(socket).print("Accepting connection error: %s\n", server.socket.DescribeError());
```

Might need a `.c_str()` in there if the error comes back as a std::string.

You'll need to include Debug.h and set up the category at the top of the file:

```cpp
namespace DFHack {
    DBG_DECLARE(core, socket, DebugCategory::LINFO);
}
```

orrche: Updated the merge request, left all the other outputs in there. Imagine they would be better in another merge request.

library/RemoteServer.cpp
ab9rf: This might not help, since all `isSocketValid` does is check if the socket descriptor is not -1; it does not check to see if the underlying socket is still able to accept connections as that would require asking the OS if the socket is still in a valid state, which `clsocket` is not programmed to do. I checked the POSIX documentation (from Open Group), and if there a reliable way to interrogate a listener as to whether it is still able to accept connections, I can't find it, nor do I remember there being one from the days I actually did low-level networking programming in UNIX.

So what I'd probably do here is have a failure counter, which gets reset to zero after each successful accept, but gets incremented after each failed accept. If the counter gets too high (3 or 5 or something), close the listening socket and attempt to allocate a new one, exiting if that fails (again, possibly trying more than once).

The sorts of circumstances in which `accept` is likely fail persistently are fairly dire (out of memory, out of kernel resource, the interface on which the socket was allocated has been shut down) but good code should still deal with such circumstances as gracefully as possible.

orrche: Will make an attempt at it, the occasions accept would fail shouldn't be to often so that the listening socket is disconnected shouldn't be to big of a problem.

Thanks for the feedback!

library/RemoteServer.cpp
```diff
@@ -476,17 +478,33 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
...
         if (client)
             client->Close();
         delete client;
     }
+
+    if (server.socket.IsSocketValid())
+    {
+        WARN(socket).print("To many failed accept, shutting down RemoteServer \n");
```
ab9rf, line 502: s/To/Too/

ab9rf (approved):

library/RemoteServer.cpp
lethosor: I would also suggest `s/accept/accepts/`

```suggestion
        WARN(socket).print("Too many failed accepts, shutting down RemoteServer\n");
```

library/RemoteServer.cpp
```diff
@@ -476,17 +478,33 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
...
-            client = nullptr;
+            if ((client = server.socket.Accept()) != NULL)
+            {
+                BlockGuard lock;
+                ServerConnection::Accepted(client);
+                client = nullptr;
+            }
+            else
+            {
+                WARN(socket).print("Accepting connection error: %s %d of %d\n", server.socket.DescribeError(), acceptFail + 1, 5);
```
myk002, line 491:
```suggestion
                WARN(socket).print("Connection failure: %s (%d of %d)\n", server.socket.DescribeError(), acceptFail + 1, 5);
```

PR #3502, opened 2023-06-27
rework wide mkdir/rmdir with correct EOL

mikerenfro:
Unintentionally closed https://github.com/DFHack/dfhack/pull/3465 when fixing my wrong EOL characters.

### DIFF
--- a/library/modules/Filesystem.cpp
+++ b/library/modules/Filesystem.cpp
@@ -99,7 +99,13 @@ bool Filesystem::mkdir (std::string path)
 {
     int fail;
 #ifdef _WIN32
-    fail = ::_mkdir(path.c_str());
+    ::std::wstring wide_path;
+    int convertResult = MultiByteToWideChar(437, 0, path.c_str(), (int)strlen(path.c_str()), NULL, 0);
+    if (convertResult > 0) {
+        wide_path.resize(convertResult+10);
+        convertResult = MultiByteToWideChar(437, 0, path.c_str(), (int)strlen(path.c_str()), &wide_path[0], (int)wide_path.size());
+    }
+    fail = ::_wmkdir(wide_path.c_str());
 #else
     fail = ::mkdir(path.c_str(), S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP |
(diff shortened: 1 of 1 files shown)


library/modules/Filesystem.cpp
```diff
@@ -99,7 +99,13 @@ bool Filesystem::mkdir (std::string path)
 {
     int fail;
 #ifdef _WIN32
-    fail = ::_mkdir(path.c_str());
+    ::std::wstring wide_path;
+    int convertResult = MultiByteToWideChar(437, 0, path.c_str(), (int)strlen(path.c_str()), NULL, 0);
+    if (convertResult > 0) {
+        wide_path.resize(convertResult+10);
```
lethosor, line 105: Why not `+1`?

library/modules/Filesystem.cpp
```diff
@@ -99,7 +99,13 @@ bool Filesystem::mkdir (std::string path)
 {
     int fail;
 #ifdef _WIN32
-    fail = ::_mkdir(path.c_str());
+    ::std::wstring wide_path;
+    int convertResult = MultiByteToWideChar(437, 0, path.c_str(), (int)strlen(path.c_str()), NULL, 0);
```
lethosor, line 103: Does this assume the input is CP437? If so, I don't think that's what we want, and I believe that came up in #3465 too. Requiring CP437 input would severely restrict the characters that callers of these functions can use in paths.

library/modules/Filesystem.cpp
mikerenfro: Mostly due to the source I was copy/pasting from (and lack of sleep).

library/modules/Filesystem.cpp
mikerenfro: I may have misread. Thought I was converting *from* CP437 to CP1252, If I mixed those up, then my mistake. Is 437 not the default in DF?

lethosor: CP437 is the default for text displayed on the DF screen. It's very much not the default for text in our code. We have a few CP437 <> UTF8 conversion functions, so I think UTF8 would be a reasonable assumption.

PR #3524, opened 2023-07-03
support copy/paste from system clipboard

### LINKED ISSUE #3523: Implement paste from system clipboard
I think this is possible now that we're on SDL2.

also copy to clipboard? that might mean that Ctrl-C can no longer mean "clear" for some DFHack textboxes.

myk002:
Fixes #3523

### DIFF
--- a/library/modules/DFSDL.cpp
+++ b/library/modules/DFSDL.cpp
@@ -135,15 +135,29 @@ int DFSDL::DFSDL_PushEvent(SDL_Event *event) {
     return g_SDL_PushEvent(event);
 }
 
-std::string DFSDL::DFSDL_GetClipboardText() {
-    if (g_SDL_HasClipboardText() != SDL_TRUE)
-        return "";
-    char *text = g_SDL_GetClipboardText();
-    std::string ret = text;
-    g_SDL_free(text);
-    return ret;
+void DFSDL::DFSDL_free(void *ptr) {
+    g_SDL_free(ptr);
 }
 
-bool DFSDL::DFSDL_SetClipboardText(const char *text) {
-    return g_SDL_SetClipboardText(text) == 0;
+char * DFSDL::DFSDL_GetClipboardText() {
+    return g_SDL_GetClipboardText();
+}
+
+int DFSDL::DFSDL_SetClipboardText(const char *text) {
+    return g_SDL_SetClipboardText(text);
+}
+
+DFHACK_EXPORT std::string DFHack::getClipboardTextCp437() {
+    if (!g_sdl_handle || g_SDL_HasClipboardText() != SDL_TRUE)
+        return "";
+    char *text = g_SDL_GetClipboardText();
+    std::string textcp437 = UTF2DF(text);
+    DFHack::DFSDL::DFSDL_free(text);
+    return textcp437;
+}
+
+DFHACK_EXPORT bool DFHack::setClipboardTextCp437(std::string text) {
+    if (!g_sdl_handle)
+        return false;
+    return 0 == DFHack::DFSDL::DFSDL_SetClipboardText(DF2UTF(text).c_str());
 }
--- a/library/modules/DFSDL.cpp
+++ b/library/modules/DFSDL.cpp
@@ -3,7 +3,10 @@
 #include "modules/DFSDL.h"
 
 #include "Debug.h"
+#include "MiscUtils.h"
 #include "PluginManager.h"
+
+#include <SDL.h>
 
 namespace DFHack {
     DBG_DECLARE(core, dfsdl, DebugCategory::LINFO);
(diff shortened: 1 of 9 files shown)


library/modules/DFSDL.cpp
```diff
@@ -124,3 +135,33 @@ void DFSDL::DFSDL_FreeSurface(SDL_Surface *surface) {
...
+    char *text = g_SDL_GetClipboardText();
+    std::string ret = text;
+    g_SDL_free(text);
+    return ret;
+}
+
+std::string DFSDL::DFSDL_GetClipboardTextCp437() {
+    std::string utf8text = DFSDL_GetClipboardTextUtf8();
+    return UTF2DF(utf8text);
+}
```
lethosor, line 151: I think we should keep this module to stubs around the SDL functions. Any additional logic can go in a separate module (MiscUtils, maybe?)

myk002: that sounds fair. will do.

myk002: though I think keeping the encapsulation of `g_SDL_free` would probably still be a good idea

lethosor: Oh, because you're changing the return type to `std::string`. I think even that conversion could go into another module. Ideally we could just generate the stubs for the SDL functions we need, and any logic changes (including type conversions) would make that hard.

myk002: My concern is that we expose ourselves to memory leaks by leaving the call to `SDL_free` up to the client.

lethosor: We can leak memory by calling all sorts of low-level C functions. I would argue that the client should be well aware of the need to free memory. For convenience, _most_ clients should use a higher-level wrapper that handles memory for them. Same reason most client code should use std::string's `operator+` instead of `strcat()`.

lethosor:
Is that because MiscUtils is used in both? Maybe MiscUtils is not the right place, then. We could potentially put the higher-level wrappers in DFSDL, just with a different namespace that makes it clearer that they aren't SDL functions.

myk002:
Is the link dependency worth avoiding? It actually shows up in the cxx-20 PR as well since RemoteClient.h now includes Core.h

lethosor:
That header dependency by itself shouldn't introduce a link dependency - isn't RemoteClient.h just pulling in the command_result enum from Core.h?

As for whether the link-time dependency is worth avoiding, I think it is, in theory. I don't know if there are third-party applications out there that use `libdfhack-client.so`, but if there are any (now or in the future), they shouldn't have to start depending on libdfhack and SDL just because we added some helper functions to MiscUtils that they don't use. In fact, I think remote clients _can't_ use these SDL functions if they aren't using SDL, so the link-time dependency is useless to them - if they do need to do something like copy text from the clipboard, they can pull in the necessary deps themselves, but that's not a common-enough use-case to force the new dep on everyone.

So I think my suggestion to put the clipboard functions in MiscUtils was a mistake. My new suggestion is:
* DFSDL functions should be simple pass-through wrappers over SDL functions
* Our own higher-level wrappers can go in a separate namespace in DFSDL.cpp, or in another module that is only used by libdfhack (and plugins)

myk002:
done. I moved the wrapper functions from MiscUtils to the DFSDL module (but in the `DFHack` namespace, not the `DFHack::DFSDL` namespace)

library/modules/DFSDL.cpp
```diff
@@ -124,3 +134,30 @@ void DFSDL::DFSDL_FreeSurface(SDL_Surface *surface) {
...
+
+char * DFSDL::DFSDL_GetClipboardText() {
+    return g_SDL_GetClipboardText();
+}
+
+int DFSDL::DFSDL_SetClipboardText(const char *text) {
+    return g_SDL_SetClipboardText(text);
+}
+
+DFHACK_EXPORT std::string DFHack::getClipboardTextCp437() {
```
lethosor, line 150: Would the `Utf8` wrappers from before still be useful (mainly to handle memory and the std::string conversion)? I suppose if there's no immediate use-case for them, they aren't too hard to add later, but I wouldn't object to adding them now either if you think they're useful.

myk002: I debated it -- and you can see both versions were in my original commit -- but after reflection, I couldn't think of a use case for the UTF-8 versions and I didn't want to clutter the API

library/modules/DFSDL.cpp
```diff
@@ -5,6 +5,8 @@
 #include "Debug.h"
 #include "PluginManager.h"
 
+#include <SDL.h>
```
lethosor, line 8: If this is just for SDL_bool, looks like that's defined in `SDL_stdinc.h` if you want to narrow this

lethosor (approved):
Looks good from static review

PR #3530, opened 2023-07-04
fix item prices and algorithm

myk002:
adjust prices according to race WAR affinity

depends on #3481 for c++-17-isms

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -50,8 +50,9 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - Price calculations fixed for many item types
 
 ## Lua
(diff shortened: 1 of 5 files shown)


docs/changelog.txt
```diff
@@ -44,9 +44,12 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## API
 
 ## Internals
+- Price calculations fixed for many item types
 
 ## Lua
-- ``dfhack.items.markForTrade``: new API for marking items for trade
+- ``dfhack.items.markForTrade``: mark items for trade
+- ``dfhack.items.isRequestedTradeGood``: discover whether an item is named in a trade agreement with an active caravan
```
lethosor, line 51: Should go under `API` since it's also new to C++

PR #3543, opened 2023-07-09
suppress SDL events if a DFHack keybinding is matched

myk002:
This PR tracks whether a keybinding has been matched and suppresses SDL events related to that key so DF doesn't react to them (or pass them through as text events)

This prevents such inconveniences as:
- Hitting Ctrl-Shift-A to bring up `gui/quickcmd` and having the "Add new command" hotkey immediately trigger because DF generates a `CUSTOM_SHIFT_A` event
- Having hotkeys get added as text in a DF or DFHack text field if one happens to be focused when a DFHack keybinding is pressed

However, it introduces a risk of preventing vanilla keyboard shortcuts from working if a DFHack keybinding is active in the same context. We could just document "don't do this", but solving some players' frustrations may introduce frustrations for others

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -2359,8 +2359,10 @@ bool Core::DFH_ncurses_key(int key)
 // returns true if the event is handled
 bool Core::DFH_SDL_Event(SDL_Event* ev)
 {
+    static std::map<int, bool> hotkey_states;
+
     // do NOT process events before we are ready.
-    if(!started || !ev)
+    if (!started || !ev)
         return false;
 
     if (ev->type == SDL_WINDOWEVENT && ev->window.event == SDL_WINDOWEVENT_FOCUS_GAINED) {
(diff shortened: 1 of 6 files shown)


lethosor:
+1 for the mentioned risks - thought I had brought those up before on GitHub, but I'm not seeing a relevant [GitHub issue](https://github.com/DFHack/dfhack/issues?q=is%3Aissue+keybinding), so maybe that happened on Discord.

I think a global suppression like this should be configurable, if we do go with it, and should be possible to disable from the command line in the worst case where a user can't pull up the settings UI to disable it.

library/Core.cpp
```diff
@@ -2356,6 +2356,18 @@ bool Core::DFH_ncurses_key(int key)
     return ncurses_wgetch(key, dummy);
 }
 
+static bool getSuppressDuplicateKeyboardEvents() {
+    auto L = Lua::Core::State;
+    color_ostream_proxy out(Core::getInstance().getConsole());
+    Lua::StackUnwinder top(L);
+    bool suppress = false;
+    Lua::CallLuaModuleFunction(out, L, "dfhack", "getSuppressDuplicateKeyboardEvents", 0, 1,
```
lethosor, line 2364: We could probably just access the table value directly to simplify things, unless you're thinking you might want more logic in the function in the future.

myk002: No, the function exists purely to give it a "standard" access method from cpp. What would direct value access look like?

lethosor: I think just `DFHack::Lua::PushModulePublic(out, L, "dfhack", "SUPPRESS_DUPLICATE_KEYBOARD_EVENTS")` should be enough to get it on the stack, followed by some call to `lua_toboolean()` to retrieve it.

myk002: done. A few other calls to Lua inside of Core.cpp could be simplified as well, either with this method (for fields) or by using `Lua::CallLuaModuleFunction` (for functions), but that can be done some other time.

PR #3546, opened 2023-07-09
[misery] realign the docs to the actual implementation

### LINKED ISSUE #3544: Incorrect thoughts/memories
All fort dwarves, be it the initial 7 or new migrants, experience soapy baths even when there's no soap available, and are miserable due to those baths, when a soapy bath should be a positive thought.
image
Tested without any other mods and after re-installing DFHack - on complete vanilla, this issue doesn't happen.
Used release 50.09 Steam of Dwarf Fortress and DFHack 50.05-r1

myk002:
Related to #3544

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -43,6 +43,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 - ``widgets.EditField``: DFHack edit fields now support cut/copy/paste with the system clipboard with Ctrl-X/Ctrl-C/Ctrl-V
 
 ## Documentation
+- `misery`: rewrite the documentation to clarify the actual effects of the plugin
 
 ## API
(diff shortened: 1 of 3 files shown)


lethosor:
(Removed the "fixes" tag since I don't believe the docs were the underlying cause of the issue)

Could you also update the help string in `misery.cpp:plugin_init()` while you're at it? Maybe `Make citizens more miserable`? I might also suggest this for the summary in the docs, but more of a weak suggestion.

myk002:
that sounds good. done.

myk002:
> (Removed the "fixes" tag since I don't believe the docs were the underlying cause of the issue)

fair enough -- my theory at this point is that `misery` is enabled in the `gui/control-panel` Autostart tab, in which case the "fix" here is advertising more clearly that `misery` will add that soapy bath thought.

PR #3597, opened 2023-07-23
deploy linux build to steam

myk002:
This PR adds a linux build and deploy step to our steam pipeline

it also adds a Linux package build action for PRs to complement the existing MSVC package build action.

question: do we care which gcc version we use to build the distribution packages? the ubuntu image has gcc-11 by default, but we can install and use a different version if we want to. our test actions use gcc-10 and gcc-12.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,9 +3,9 @@ name: Build
 on: [push, pull_request]
 
 jobs:
-  build:
+  build-test:
     runs-on: ${{ matrix.os }}
-    name: build (Linux, GCC ${{ matrix.gcc }}, ${{ matrix.plugins }} plugins)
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -29,25 +29,18 @@ jobs:
...
-            libncursesw5 \
-            libsdl-image1.2-dev \
-            libsdl-ttf2.0-dev \
-            libsdl1.2-dev \
-            libxml-libxml-perl \
             libxml-libxslt-perl \
-            lua5.3 \
-            ninja-build \
-            zlib1g-dev
+            libgl-dev \
```
lethosor, line 33: not (quite) sorted

myk002: I didn't even notice that there was sorting going on : p will fix

myk002: done

PR #3612, opened 2023-08-01
Fix mangling of `plugin_globals` with GCC's C++11 ABI

lethosor:
Without being declared with `extern "C"`, `plugin_globals` is mangled, with a `cxx11` suffix.

We can't add `extern "C"` to the `DFhackDataExport` macro because GCC does not allow initializing any `extern` variables inline, including `extern "C"`.

### DIFF
--- a/library/include/PluginManager.h
+++ b/library/include/PluginManager.h
@@ -298,14 +298,16 @@ namespace DFHack
 };
 
 #define DFHACK_PLUGIN_AUX(m_plugin_name, is_dev) \
+extern "C" { \
     DFhackDataExport const char * plugin_name = m_plugin_name;\
     DFhackDataExport const char * plugin_version = DFHack::Version::dfhack_version();\
     DFhackDataExport const char * plugin_git_description = DFHack::Version::git_description();\
     DFhackDataExport int plugin_abi_version = DFHack::Version::dfhack_abi_version();\
     DFhackDataExport DFHack::Plugin *plugin_self = NULL;\
-    std::vector<std::string> _plugin_globals;\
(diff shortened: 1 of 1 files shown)


library/include/PluginManager.h
```diff
@@ -298,14 +298,16 @@ namespace DFHack
...
 
 #define DFHACK_PLUGIN_AUX(m_plugin_name, is_dev) \
+extern "C" { \
     DFhackDataExport const char * plugin_name = m_plugin_name;\
     DFhackDataExport const char * plugin_version = DFHack::Version::dfhack_version();\
     DFhackDataExport const char * plugin_git_description = DFHack::Version::git_description();\
     DFhackDataExport int plugin_abi_version = DFHack::Version::dfhack_abi_version();\
     DFhackDataExport DFHack::Plugin *plugin_self = NULL;\
-    std::vector<std::string> _plugin_globals;\
+    DFhackDataExport std::vector<std::string> _plugin_globals;\
```
myk002, line 307: do we need to export the underscore non-pointer version? is `_plugin_globals` referenced outside the plugin? I would think we'd only need the `extern "C"` change and not this export line.

lethosor: This line was left over from testing... you are probably right that this one doesn't need to be exported.

myk002: `unresolved external symbol _plugin_globals` -- might be fixed by reverting that line

lethosor: Yeah, pushing up a new version. I'm also going to switch away from the `_` prefix, since that's reserved, and `C` globals actually get a `_` prefix added so I'm not sure if that was potentially problematic.

lethosor: Built successfully this time!

Issue #3427, opened 2023-05-28
codegen: new types needed for SDL2 builds


lethosor:
- [x] `std::mutex`
- [x] `std::condition_variable`
- [ ] `std::atomic_int` or `std::atomic<T>`
- [ ] `std::variant` (this one is probably challenging - TBD if we can reuse our tagged union code for Lua)
- [x] `size_t` (and `ssize_t` while we're at it?)

ab9rf:
improving how we do `std::variant` in codegen is still an open issue
and we could do better with `std::atomic` as well

PR #3617, opened 2023-08-02
DataIdentity support for new types (std::mutex, std::condition_variable)

lethosor:
Requires https://github.com/DFHack/df-structures/pull/622
Needed to fix enabler on Linux (#3427)

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -62,6 +62,7 @@ set(MAIN_SOURCES
     ColorText.cpp
     CompilerWorkAround.cpp
     DataDefs.cpp
+    DataIdentity.cpp
     Debug.cpp
     Error.cpp
     VTableInterpose.cpp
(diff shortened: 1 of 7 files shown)


lethosor:
I think this should work now. Needs testing on Windows to be sure.

lethosor:
enabler still appears to be aligned on Windows. Merging.

myk002 (approved):
I'm not an expert in everything that was changed here, but it all looks reasonable. and proof that windows still works gives me some confidence

lethosor:
I'm not too worried either. Just wanted to make sure I didn't accidentally introduce a misalignment. ab9rf also confirmed that gps is the correct size still. 

On the DFHack side (this PR), most the changes are either strict additions, or changes to Lua wrappers for types that shouldn't generally be accessed from Lua anyway (std::fstream).

PR #3620, opened 2023-08-03
remove buildmaster integration files

myk002:
unless we want to keep these around?

if we remove these files, we could/should also remove the DFHACK_BUILDMASTER_WEBHOOK_URL and DFHACK_BUILDMASTER_WEBHOOK_SECRET secrets

and also the corresponding build action and secrets in the df-structures repo

related question: should we remove the travis-ci integrations from our repo settings?


lethosor (approved):
Fine by me, easy to re-add these if we need them

PR #3622, opened 2023-08-03
first attempt to run tests on windows

myk002:


ci/run-tests.py
```diff
@@ -33,7 +33,7 @@
 
 MAX_TRIES = 5
 
-dfhack = 'Dwarf Fortress.exe' if sys.platform == 'win32' else './dfhack'
+dfhack = 'Dwarf Fortress.exe' if sys.platform == 'win32' else './dwarfort'
```
lethosor, line 36: We should keep this as `./dfhack` on Linux for now so that we can see output.

myk002: done

.github/workflows/test.yml
```diff
@@ -33,51 +45,78 @@ jobs:
...
+          #   compiler: gcc-10
+          #   gcc: 10
+          #   plugins: "default"
+          # - os: ubuntu
+          #   compiler: gcc-10
+          #   gcc: 12
+          #   plugins: "all"
     steps:
-    - name: Download artifact
+    - name: Set env (windows)
```
lethosor, line 71: We can probably unify these two steps across platforms by using "shell: bash"

PR #3625, opened 2023-08-04
re-enable unit tests

myk002:
and fix up the issues we find

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -379,7 +379,7 @@ local function load_tests(file, tests)
         dfhack.printerr('Skipping tests for unspecified target in ' .. file)
         return true  -- TODO: change to false once existing tests have targets specified
     end
-    local targets = type(env.config.target) == table and env.config.target or {env.config.target}
+    local targets = type(env.config.target) == 'table' and env.config.target or {env.config.target}
     for _,target in ipairs(targets) do
         if target == 'core' then goto continue end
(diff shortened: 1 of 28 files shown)


test/gui.lua
```diff
@@ -1,3 +1,5 @@
+config.targets = 'core'
```
lethosor, line 1: Are we planning to support multiple targets per test? Would suggest `target` if not.

myk002: I had the quickfort suite of integration tests in mind when I made this plural. I could also make a singular variant of the property. The overwhelming majority of tests test one target (being "unit" tests and all). Maybe calling it "target" but still letting it take a list for integration cases is the way to go.

ci/test.lua
```diff
@@ -379,7 +379,7 @@ local function load_tests(file, tests)
         dfhack.printerr('Skipping tests for unspecified target in ' .. file)
         return true  -- TODO: change to false once existing tests have targets specified
     end
-    local targets = type(env.config.targets) == table and env.config.targets or {env.config.targets}
+    local targets = type(env.config.target) == table and env.config.target or {env.config.target}
```
lethosor, line 382: `type()` returns strings:
```suggestion
    local targets = type(env.config.target) == 'table' and env.config.target or {env.config.target}
```

myk002: ha whoops. thanks. fixed.

PR #3642, opened 2023-08-06
Add option to force-fetch world blocks in RemoteFortressReader

randombk:
The current behavior of GetBlockList in the RemoteFortressReader
Protobuf RPC API is to only return blocks that have changed since
the last fetch. This causes problems when the RPC client (i.e. a
world renderer) wants to restart, as it can no longer fetch the
full world state.

This patch adds a `force_reload` option to BlockRequest, defaulting
to `false` (the current behavior). When passed, it returns all
requested blocks regardless of whether they have changed or not.

Tested against the `develop` and `v0.47.05` branches.
Using the `dfhack-remote` JS RPC client:
Screenshot from 2023-08-06 13-33-02

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
+- `RemoteFortressReader`: add a `force_reload` option to the GetBlockList RPC API to return blocks regardless of whether they have changed since the last request
 
 ## Internals
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -42,7 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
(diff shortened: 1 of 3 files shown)


myk002:
@RosaryMala FYI

myk002 (commented):
code change looks good. could you add a line to `docs/changelog.txt` in the "Future" version under the "API" section?

randombk:
Done!

myk002 (approved):

docs/changelog.txt
```diff
@@ -42,6 +42,7 @@ changelog.txt uses a syntax similar to RST, with a few special sequences:
 ## Documentation
 
 ## API
+- `RemoteFortressReader`: add a `force_reload` option to the GetBlockList RPC API to return blocks regardless of whether they have changed since the last request
```
```suggestion
- `RemoteFortressReader`: add a ``force_reload`` option to the GetBlockList RPC API to return blocks regardless of whether they have changed since the last request
```

PR #3646, opened 2023-08-07
add changelog template for new versions

myk002:
and mark unused headers as deprecated

### DIFF
--- a/docs/sphinx_extensions/dfhack/changelog.py
+++ b/docs/sphinx_extensions/dfhack/changelog.py
@@ -16,11 +16,11 @@ CHANGELOG_PATHS = (
 CHANGELOG_PATHS = (os.path.join(DFHACK_ROOT, p) for p in CHANGELOG_PATHS)
 
 CHANGELOG_SECTIONS = [
-    'New Plugins',
-    'New Scripts',
-    'New Tweaks',
+    'New Plugins', # deprecated
+    'New Scripts', # deprecated
+    'New Tweaks', # deprecated
     'New Features',
-    'New Internal Commands',
+    'New Internal Commands', # deprecated
     'Fixes',
     'Misc Improvements',
     'Removed',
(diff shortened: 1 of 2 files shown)


docs/sphinx_extensions/dfhack/changelog.py
```diff
@@ -16,11 +16,11 @@
 CHANGELOG_PATHS = (os.path.join(DFHACK_ROOT, p) for p in CHANGELOG_PATHS)
 
 CHANGELOG_SECTIONS = [
-    'New Plugins',
-    'New Scripts',
-    'New Tweaks',
+    'New Plugins', # deprecated
```
lethosor, line 19: I think separating "New Tools" from "New Features" would still be a useful distinction, but I agree we probably don't need to differentiate between plugins/scripts/tweaks.

myk002: I went back and forth on that. I originally *did* add a "New Tools" section (in fact I even added that section manually in the 50.09-r2 release notes), then backed out the change and went with the existing "New Features" heading.

I think the distinction between a new tool and a new feature is getting fuzzier. Is the new trade interface a new tool? When an end user reads the release notes, what do they care most about? If the answer is just "what's new?" then I'd say a single section is probably sufficient, and we should only break it out into separate sections if a single "New Features" section gets too large.

lethosor: I would argue that we shouldn't be putting logically separate features in the same tool (e.g. the caravan script should probably not have gained a trade UI). But more generally, I think there is still an important distinction between "new tools" and "new features of existing tools". A completely new UI could fall into the former category, regardless of how it's technically implemented. We don't have any enforceable checks that a tool referenced under "new tools" actually didn't exist beforehand.

myk002: Fair enough. I'll add "New Tools" as a valid header and add it to the templates.

We may have to have a more focused "logically separate features in the same tool" discussion at some point. The overlay-driven UIs I'm adding tend to have more thematic links to the tool they're associated with than actual code dependencies.

lethosor: In some cases, splitting those out into "gui/" and non-"gui/" scripts might work (e.g. teleport), but I'm sure there are more complicated cases. And I understand naming is hard, especially when we started with one tool that claimed the name for something that multiple tools would like to manage now.

PR #3653, opened 2023-08-08
add functions to Panel in addition to attributes

myk002:
so they are easy to work with in subclasses


lethosor (approved):
LGTM from static review

PR #3662, opened 2023-08-11
Fix crash in df.global:_field() when global address is unknown

lethosor:
Also added a test. Probably minor enough to skip the changelog.

### DIFF
--- a/test/structures/globals.lua
+++ b/test/structures/globals.lua
@@ -1,0 +1,33 @@
+config.mode = 'title'
+config.target = 'core'
+
+local function with_temp_global_address(name, addr, callback, ...)
+    dfhack.call_with_finalizer(2, true,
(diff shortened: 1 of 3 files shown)


test/structures/globals.lua
```diff
@@ -0,0 +1,33 @@
+config.mode = 'title'
```
myk002, line 1: why is this restricted to the title screen? It looks pretty generic

lethosor: I copied it from `find.lua`, which is similarly generic.

lethosor: Thinking about it a little more, `find.lua` is not safe to run when a world is loaded because it permanently modifies world state - that's probably why it was configured that way. `globals.lua` does clean up after itself properly.

myk002 (approved):

PR #3663, opened 2023-08-11
feat: dynamic texture loading

shevernitskiy:
This PR introduce dynamic texture loading.

The main issue with textures is that game clear texture cache in some points of the game process, and texpos, that refering to the texture, become invalid. In this moment game reregister textures in `enabler.textures.raws` vector and get new texposes. So we need to do the same thing, if we want keep use our texture.

There are four points of texture resetting:
- game continue
- new game in existing world
- new world generation
- new arena

Here we track loading process step and reset our texpos cache as most closely as game does.

Typical usage:
1. Register texture and get `TexposHandle` with `Textures::loadTexture()`. Keep it wherever you want.
2. Get valid texpos with `Textures::getTexposByHandle()`.

Possible improvements:
- add possibility to load images as texture (like static dfhack textures) and expose this method to lua
- if needed, existing code could be refactored to use dynamic mechanism without adjusting `init_texture_size`


library/modules/Textures.cpp
```diff
@@ -72,13 +83,148 @@ SDL_Surface * canonicalize_format(SDL_Surface *src) {
...
+    g_handle_to_texpos.emplace(handle, texpos);
+    return handle;
+}
+
+std::optional<long> Textures::getTexposByHandle(TexposHandle handle) {
+    if (!handle)
+        return std::nullopt;
+
+    if (auto it = g_handle_to_texpos.find(handle); it != g_handle_to_texpos.end())
+        return it->second;
```
myk002, line 111:
```suggestion
    if (g_handle_to_texpos.count(handle))
        return g_handle_to_texpos[handle];
```

myk002: what you had wasn't wrong, but this is (subjectively) easier to read

library/modules/Textures.cpp
```diff
@@ -72,13 +83,148 @@ SDL_Surface * canonicalize_format(SDL_Surface *src) {
...
+        INTERPOSE_NEXT(logic)();
+    }
+
+  private:
+    inline static int m_cur_step = -2; // not valid state at the start
+};
+IMPLEMENT_VMETHOD_INTERPOSE(tracking_stage_new_arena, logic);
+
+static void
+  install_reset_point() {
```
myk002, line 207:
```suggestion
static void install_reset_point() {
```

library/modules/Textures.cpp
```diff
@@ -120,7 +266,12 @@ static size_t load_textures(color_ostream & out, const char * fname,
 // texture count so our textures will no longer be freed when worlds are
 // unloaded.
 //
-void Textures::init(color_ostream &out) {
+void Textures::init(color_ostream& out) {
```
myk002, line 269: with the interposed handlers in place, we no longer need to call `Textures::init(out)` from Core on every screen transition. You can modify Core.cpp to only call `Textures::init` once.

library/include/modules/Textures.h
```diff
@@ -12,6 +18,19 @@ namespace DFHack {
  */
 namespace Textures {
 
+/**
+ * Load texture and get handle.
+ * Keep it to obtain valid texpos.
+ */
+DFHACK_EXPORT TexposHandle loadTexture(SDL_Surface* surface);
```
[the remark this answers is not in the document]
shevernitskiy: I can make additional method like `loadTileset()`, that could do `image -> surface`, merging on tiles and calling this one. Surface operation could be useful for iternal cpp plugins like mine translation thing (i got surfaces from sdl_ttf).

library/include/modules/Textures.h
```diff
@@ -12,6 +18,19 @@ namespace DFHack {
...
+
+/**
+ * Get texpos by handle.
+ * Always use this function, if you need to get valid texpos for your texure.
+ * Texpos can change on game textures reset, but handle will be the same.
+ */
+DFHACK_EXPORT std::optional<long> getTexposByHandle(TexposHandle handle);
```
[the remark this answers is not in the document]
shevernitskiy: Sure, i can return -1 on invalid value

library/modules/Textures.cpp
```diff
@@ -72,13 +83,148 @@ SDL_Surface * canonicalize_format(SDL_Surface *src) {
...
+struct tracking_stage_new_region : df::viewscreen_new_regionst {
+    typedef df::viewscreen_new_regionst interpose_base;
+
+    DEFINE_VMETHOD_INTERPOSE(void, logic, ()) {
+        if (this->m_raw_load_stage != this->raw_load_stage) {
+            this->m_raw_load_stage = this->raw_load_stage;
+            if (this->m_raw_load_stage == 1)
+                reset_texpos();
+        }
+        INTERPOSE_NEXT(logic)();
```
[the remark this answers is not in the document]
shevernitskiy: I've tried different combination of `INTERPOSE_NEXT ` and actual trigger stage/step number. If I place it before step check, then it would be +1 step (for the closeset reset point) comapring to texture cleaning in ghidra listing. Should I do that?

library/modules/Textures.cpp
shevernitskiy: Then it would be something like new `Texture::dynamicInit()`, because we still need to handle static asstest case, right?

library/modules/Textures.cpp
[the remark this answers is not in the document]
shevernitskiy: Сhecked all the options again. If `INTERPOSE_NEXT` placed after step check, then itpossible to use the same step number which game use. And there is no significant interval with invalid texposes during loading screen. There are only single "blink" in couple of loading screen.

shevernitskiy:
ready to test, needs https://github.com/DFHack/scripts/pull/798

shevernitskiy:
If the mechanism will be approved, then it will be possible to move whole assets part to lua.
Register on init, receive hanldes and declare callbacks to get valid texpos by handles.
Cpp part will provide api only.
But it will need to rework lua part all over codebase.

plugins/lua/buildingplan/pens.lua
```diff
@@ -9,29 +9,30 @@ MINI_TEXT_PEN, MINI_TEXT_HPEN, MINI_BUTT_PEN, MINI_BUTT_HPEN = nil, nil, nil, ni
 
 local to_pen = dfhack.pen.parse
 
-local tp = function(base, offset)
-    if base == -1 then return nil end
-    return base + offset
+local tp = function(asset, offset)
+    local texpos = dfhack.textures.getAsset(asset, offset)
+    if texpos == -1 then return nil end
+    return texpos
```
myk002:
```suggestion
        return texpos >= 0 and texpos or nil
```

library/Core.cpp
```diff
@@ -2218,7 +2218,6 @@ void Core::onStateChange(color_ostream &out, state_change_event event)
         }
         break;
     case SC_VIEWSCREEN_CHANGED:
-        Textures::init(out);
```
myk002, line 2221: you can take out the entire SC_VIEWSCREEN_CHANGED case, I think

library/include/modules/Textures.h
```diff
@@ -1,7 +1,14 @@
 #pragma once
 
-#include "Export.h"
+#include <string>
+#include <vector>
+
 #include "ColorText.h"
+#include "Export.h"
+
+#include <SDL_surface.h>
```
myk002, line 9: I'm still uncomfortable including a SDL header at this scope. Since we just interact with pointers to SDL_Surface objects, you can replace the `#include <SDL_surface.h>` with a forward declaration: `struct SDL_Surface;`

library/lua/gui.lua
```diff
@@ -913,10 +913,10 @@ local BASE_FRAME = {
 }
 
 local function make_frame(name, double_line)
-    local texpos = dfhack.textures['get'..name..'BordersTexposStart']()
     local tp = function(offset)
+        local texpos = dfhack.textures.getAsset('hack/data/art/border-'..name:lower()..'.png', offset)
```
myk002, line 917: I suggest making the strings that this function is called with lower case instead of paying the cost of lower casing each time this function is called

library/lua/gui.lua
```diff
@@ -951,8 +951,28 @@ BOLD_FRAME = FRAME_BOLD
...
+        local frame = make_frame('Thin', true)
+        frame.signature_pen = false
+        return frame
+    end
+    if style == FRAME_INTERIOR_MEDIUM or style == INTERIOR_MEDIUM_FRAME then
+        local frame = make_frame('Medium', true)
+        frame.signature_pen = false
+        return frame
+    end
+end
```
myk002, line 971: This will only work for non-custom frames. What about if you replace the `tile` value with a function that returns a texpos? Then you can call that function in LuaApi.cpp in `decode_pen()` (replace the line `get_int_field(L, &pen.tile, idx, "tile", 0);`)

that is, extend Pen abstraction to allow the `tile` field to be either an integer or a function that returns an integer. If it's the former, then the current handling in `decode_pen()` stays the same. If it's a function, then call it to get the integer.

library/modules/Textures.cpp
```diff
@@ -41,205 +48,236 @@ static long g_window_borders_texpos_start = -1;
...
 
-static size_t load_textures(color_ostream & out, const char * fname,
-                            long *texpos_start,
-                            int tile_w = TILE_WIDTH_PX,
-                            int tile_h = TILE_HEIGHT_PX) {
-    SDL_Surface *s = DFIMG_Load(fname);
-    if (!s) {
-        out.printerr("unable to load textures from '%s'\n", fname);
-        return 0;
+    auto handle = reinterpret_cast<TexposHandle>(surface); // not the best way? but cheap
```
myk002, line 100: it might be better to just typedef TexposHandle as a SDL_Surface * in the header file and avoid the need to cast at all.

library/modules/Textures.cpp
```diff
@@ -41,205 +48,236 @@ static long g_window_borders_texpos_start = -1;
...
+    g_handle_to_surface.emplace(handle, surface);
+    surface->refcount++; // prevent destruct on next FreeSurface by game
+    auto texpos = add_texture(surface);
+    g_handle_to_texpos.emplace(handle, texpos);
+    return handle;
+}
+
+std::vector<TexposHandle> Textures::loadTileset(const std::string& file,
+                                                int tile_px_w = TILE_WIDTH_PX,
+                                                int tile_px_h = TILE_HEIGHT_PX) {
```
myk002, line 110: these defaults need to be in the header file to have any effect

library/modules/Textures.cpp
```diff
@@ -41,205 +48,236 @@ static long g_window_borders_texpos_start = -1;
...
+    install_reset_point();
+    DEBUG(textures, out).print("dynamic texture loading ready");
+
+    for (auto& [key, value] : g_static_assets) {
+        auto tile_w = TILE_WIDTH_PX;
+        auto tile_h = TILE_HEIGHT_PX;
+        if (key == "hack/data/art/pathable.png" || key == "hack/data/art/unsuspend.png") {
+            tile_w = 32;
+            tile_h = 32;
+        }
```
myk002, line 272: this should be in the g_static_assets data structure so it can be maintained in one place

library/lua/gui.lua
shevernitskiy: This is just a crutch to bring current codebase to work with introduced pipeline. Should be reworked fro sure!
I'm not sure i understand you clearly:) Imho, cpp should know nothing about special cases.
Maybe it will be better to rework lua part to work with callbacks but not with static values? (more work, i guess)

library/lua/gui.lua
shevernitskiy: Yeah, we can avoid addition cost here.
Can it be in the next iteraction of refactor? (If it will be assets init in lua)

shevernitskiy:
done:)

myk002 (commented):
I want to go through this in detail a bit more, but I want to say that it looks surprisingly clean. You've done a really good job with this!

I also want to do some local testing to see if it has any noticeable impact on performance, since this is such a significant change to the rendering step.

myk002:
for the build error, use `%zd` in the format string instead of `%i`

shevernitskiy:
> 

Thanks for kind words:)
Yes, perf testing is needed as well simple bug testing in affected plugins/scripts!

myk002:
testing has shown this anomaly while starting a new game in an existing world:

https://github.com/DFHack/dfhack/assets/977482/d966f6c1-4376-46f9-8e82-f7b68af17527

shevernitskiy:
> testing has shown this anomaly while starting a new game in an existing world:
> 
>  simplescreenrecorder-2023-08-27_11.55.47.mp4

That because there is a time between game texture raws reset and our cache reset.
I'll look if it possible intorpose reset trigger more precisely.

library/lua/gui.lua
```diff
@@ -885,6 +885,56 @@ ZScreenModal.ATTRS{
...
+function tp_border_bold(offset)
+    return dfhack.textures.getTexposByHandle(texpos_handles.border_bold[offset])
+end
+function tp_border_panel(offset)
+    return dfhack.textures.getTexposByHandle(texpos_handles.border_panel[offset])
+end
+function tp_border_window(offset)
+    return dfhack.textures.getTexposByHandle(texpos_handles.border_window[offset])
+end
+
```
myk002, line 937: Looking at the usage in `plugins/lua/buildingplan/pens.lua`, These feel a little out of place in this file. could you move them to a new `gui/textures.lua` file? `gui.lua` can add `local textures = require('gui.textures')`

myk002 (approved):

myk002:
In a separate PR, could you possibly add your name to docs/about/Authors.rst? You've definitely earned it!

shevernitskiy:
> In a separate PR, could you possibly add your name to docs/about/Authors.rst? You've definitely earned it!

I'll add in the next PR with texture generation:)

Issue #3668, opened 2023-08-12
Parse global addresses from global table at runtime


lethosor:
This would allow us to avoid adding all globals to symbols.xml (we would just need `global_table`), and would potentially be more resilient to ASLR in the future if needed.

PR #3671, opened 2023-08-12
Expose struct_field_info to Lua

lethosor:
For #3668 

To do:
- [x] fix initialization order for the `type` field (`A._fields.b.type` won't be set to `B` if `A._fields` is initialized first)
- [x] add `struct_field_info_extra` fields

### DIFF
--- a/library/LuaTypes.cpp
+++ b/library/LuaTypes.cpp
@@ -1365,6 +1365,13 @@ static void IndexFields(lua_State *state, int base, struct_identity *pstruct, in
     }
 }
 
+static void PushTypeIdentity(lua_State *state, const type_identity *id)
+{
+    lua_rawgetp(state, LUA_REGISTRYINDEX, &DFHACK_TYPEID_TABLE_TOKEN);
+    lua_rawgetp(state, -1, id);
+    lua_remove(state, -2);  // TYPEID_TABLE
+}
+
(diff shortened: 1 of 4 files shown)


library/LuaTypes.cpp
```diff
@@ -1337,21 +1354,179 @@ static void IndexFields(lua_State *state, int base, struct_identity *pstruct, bo
...
+    while (lua_next(state, ix_fieldinfo)) {
+        auto field = static_cast<const struct_field_info*>(lua_touserdata(state, -1));
+        lua_pushvalue(state, -2);  // field name
+        PushFieldInfoSubTable(state, field);
+        lua_settable(state, ix_fields);
+        lua_pop(state, 1); // struct_field_info
+    }
+
+    // lua_pushvalue(state, ix_fields);
+    // freeze_table(state); // TODO: figure out why this creates an __index cycle for nonexistent fields
```
myk002, line 1521: does the index cycle happen for other frozen tables, such as the `attrs` tables?

lethosor: Nope, but those define `__index` to be a metamethod, not a table.

I'm effectively freezing the table by setting `__newindex`, anyway, so I'm not really worried about this. It's a nice-to-have that could save maybe 5-10 lines of code, but now that I'm generating the field-specific tables on the fly, I only really need to freeze the outer table, so there's not much duplication.

myk002 (approved):

PR #3689, opened 2023-08-21
fix tests sometimes timing out in ci

myk002:
initial_prep takes longer than our timeout for initial load

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -162,16 +162,19 @@ end
 local function ensure_title_screen()
     for i = 1, 100 do
         local scr = dfhack.gui.getCurViewscreen()
-        if is_title_screen(scr) then
+        if df.viewscreen_initial_prepst:is_instance(scr) then
+            delay(10000)
+        elseif is_title_screen(scr) then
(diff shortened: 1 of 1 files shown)


ci/test.lua
```diff
@@ -162,16 +162,19 @@ end
 local function ensure_title_screen()
     for i = 1, 100 do
         local scr = dfhack.gui.getCurViewscreen()
-        if is_title_screen(scr) then
+        if df.viewscreen_initial_prepst:is_instance(scr) then
+            delay(10000)
```
lethosor, line 166: This is 100 seconds with the default frame rate - do we really need the delay to be this long?

myk002: I'm pretty sure loading does not respect frame rate limits. Testing with less does still result in timeouts.

lethosor: Then we should probably make the check for `viewscreen_initial_prepst` not time out, or maybe add a wall-time limit for it instead. I would rather not risk a 100-second delay here if DF behavior changes.

PR #3710, opened 2023-08-31
[tailor] fix status output character encoding

myk002:
so dwarf and item names and item descriptions are printed correctly

probably only an issue on Linux since DF2CONSOLE is a noop on Windows

### DIFF
--- a/plugins/tailor.cpp
+++ b/plugins/tailor.cpp
@@ -222,7 +222,7 @@ public:
                 // only count dyed
                 std::string d;
                 i->getItemDescription(&d, 0);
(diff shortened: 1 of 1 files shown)


ab9rf:
as an aside, `DF2CONSOLE` probably shouldn't always be a no-op on windows, insofar as not everyone has their console in CP-1252. my Windows conhost is set to UTF-8, which means i would probably benefit from using that translation as well

quietust:
I assume you meant CP437 (the "OEM" codepage), not CP-1252 (the "ANSI" codepage), but yes, non-English versions of Windows can easily use something different.

PR #3712, opened 2023-08-31
remove unneeded modification of CMAKE_CXX_FLAGS_RELWITHDEBINFO

myk002:
-g is already in the flags, we just end up adding it a second time


lethosor (approved):
Looks like this is left over from when we added `-g -Wall -Wno-unused-variable` to RELWITHDEBINFO builds only, but those were moved to all build types in 8d551c7a7cbf05a65df498c467efcfb4efcbd4fe

PR #3713, opened 2023-08-31
Hack to force GCC to invoke cancel_job() through DF's vtable

lethosor:
GCC appears to be optimizing the call to `cancel_job()` to use the stub in *DFHack's* job_handler vtable, which is a no-op. Lua was unaffected because it invokes vmethods through method pointers (without knowing the target instance at compile time), so use a similar approach here for now.

As mentioned by @ab9rf on Discord, we should pursue an alternative like asking Bay12 to expose the relevant code through a global `std::function` instead of a vmethod.

### DIFF
--- a/library/modules/Job.cpp
+++ b/library/modules/Job.cpp
@@ -387,7 +387,14 @@ bool DFHack::Job::removeJob(df::job* job) {
     // call the job cancel vmethod graciously provided by The Toady One.
     // job_handler::cancel_job calls job::~job, and then deletes job (this has
     // been confirmed by disassembly).
-    world->jobs.cancel_job(job);
+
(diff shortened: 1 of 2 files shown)


myk002 (approved):

lethosor:
hold, adding a test

ab9rf:
lemme verify that this works on windows. it should, but still...

lethosor:
Good to check. From CI, [test-output-msvc-default_plugins-default_config](https://github.com/DFHack/dfhack/suites/15705909338/artifacts/896695384) -> `test_status.json` contains `"modules/job.lua:removeJob": "passed",` which is hopefully a good sign.

myk002 (approved):

ab9rf (approved):
this change does not impair compatibility with MSVC - lgtm

Issue #3709, opened 2023-08-31
textures become corrupted on game reload


myk002:
to reproduce:
1. load a fort
2. quit to main menu without saving
3. load the fort again

The DFHack logo will become corrupted during the loading sequence.

![image](https://github.com/DFHack/dfhack/assets/977482/d86819c8-a388-4c04-abea-ae2e99ad059b)

@shevernitskiy

myk002:
disregard -- this is likely due to the tests I was doing with `cancel_job`, which apparently is corrupting things.

myk002:
it doesn't appear to be a refcount issue. increasing the refcount by 10 instead of 1 did not solve the issue.

shevernitskiy:
> it doesn't appear to be a refcount issue. increasing the refcount by 10 instead of 1 did not solve the issue.

if it something with refcount -> surface not exist -> nullref -> crash
cached texpos (index of texture.raws vector) refering to not target texture

PR #3716, opened 2023-09-01
reserved texpos range

shevernitskiy:
This PR add `reserved range` feature to dynamic texture module.

Reserved range is a buffer in the `textures.raws` vector right after static game textures. Game will never wipes items of this range from the vector.

Now `loadTilset()`, `loadTexture()`, `createTile()`, `createTileset()` can receive additional optional parameter `bool reserved` which define target range (false by default, dynamic range).

close https://github.com/DFHack/dfhack/issues/3709


docs/dev/Lua API.rst
```diff
@@ -2595,10 +2595,12 @@ The ``textures`` module solves this problem by providing a stable handle instead
 raw ``texpos``. When we need to draw a particular tile, we can look up the current
 ``texpos`` value via the handle.
 
-* ``loadTileset(file, tile_px_w, tile_px_h)``
+* ``loadTileset(file, tile_px_w, tile_px_h, reserved?)``
```
myk002, line 2598:
```suggestion
* ``loadTileset(file, tile_px_w, tile_px_h[, reserved])``
```

docs/dev/Lua API.rst
```diff
@@ -2595,10 +2595,12 @@ The ``textures`` module solves this problem by providing a stable handle instead
...
 
   Loads a tileset from the image ``file`` with give tile dimensions in pixels. The
   image will be sliced in row major order. Returns an array of ``TexposHandle``.
+  ``reserved`` is optional boolean argument, which indicates texpos range.
+  ``true`` - reserved, ``false`` - dynamic (default).
```
myk002, line 2603: maybe add a few sentences to the introduction section (line 2567) explaining what the reserved range is. Also, we should document what happens if the reserved range is already full.

docs/dev/Lua API.rst
```diff
@@ -2611,18 +2613,22 @@ raw ``texpos``. When we need to draw a particular tile, we can look up the curre
   get the ``texpos`` for your texture. ``texpos`` can change when game textures are
   reset, but the handle will be the same.
 
-* ``createTile(pixels, tile_px_w, tile_px_h)``
+* ``createTile(pixels, tile_px_w, tile_px_h, reserved?)``
```
myk002, line 2616:
```suggestion
* ``createTile(pixels, tile_px_w, tile_px_h[, reserved])``
```

docs/dev/Lua API.rst
```diff
@@ -2611,18 +2613,22 @@ raw ``texpos``. When we need to draw a particular tile, we can look up the curre
...
 
   Create and register a new texture with the given tile dimensions and an array of
   ``pixels`` in row major order. Each pixel is an integer representing color in packed
   RBGA format (for example, #0022FF11). Returns a ``TexposHandle``.
+  ``reserved`` is optional boolean argument, which indicates texpos range.
+  ``true`` - reserved, ``false`` - dynamic (default).
 
-* ``createTileset(pixels, texture_px_w, texture_px_h, tile_px_w, tile_px_h)``
+* ``createTileset(pixels, texture_px_w, texture_px_h, tile_px_w, tile_px_h, reserved?)``
```
myk002, line 2624:
```suggestion
* ``createTileset(pixels, texture_px_w, texture_px_h, tile_px_w, tile_px_h[, reserved])``
```

library/modules/Textures.cpp
```diff
@@ -71,6 +96,12 @@ static long add_texture(SDL_Surface* surface) {
     return texpos;
 }
 
+// register surface in texture raws to specific texpos, returns a texpos
```
myk002, line 99:
```suggestion
// register surface in texture raws to specific texpos
```

library/modules/Textures.cpp
```diff
@@ -118,42 +150,66 @@ std::vector<TexposHandle> slice_tileset(SDL_Surface* surface, int tile_px_w, int
     return handles;
 }
 
-TexposHandle Textures::loadTexture(SDL_Surface* surface) {
+TexposHandle Textures::loadTexture(SDL_Surface* surface, bool reserved) {
     if (!surface)
         return 0; // should be some error, i guess
+    if (loading_state) {
+        ERR(textures).printerr("unable to load texture during game loading\n");
```
myk002, line 157: I thought we were going to use the reserved buffer for this? e.g. instead of printing an error and returning 0, set `reserved = true` and continue on

library/modules/Textures.cpp
```diff
@@ -118,42 +150,66 @@ std::vector<TexposHandle> slice_tileset(SDL_Surface* surface, int tile_px_w, int
...
     auto handle = reinterpret_cast<uintptr_t>(surface);
     g_handle_to_surface.emplace(handle, surface);
     surface->refcount++; // prevent destruct on next FreeSurface by game
-    auto texpos = add_texture(surface);
-    g_handle_to_texpos.emplace(handle, texpos);
+    if (reserved) {
+        auto texpos = reserved_range.get_new_texpos();
+        if (texpos == -1) {
+            ERR(textures).printerr("reserved range limit has been reached, use dynamic range\n");
+            return 0;
```
myk002, line 168: how about if this only becomes an error if `loading_state == true`? otherwise we can automatically fall back to making it a dynamic texture.

library/modules/Textures.cpp
```diff
@@ -109,7 +141,7 @@ std::vector<TexposHandle> slice_tileset(SDL_Surface* surface, int tile_px_w, int
                 surface->format->Bmask, surface->format->Amask);
             SDL_Rect vp{tile_px_w * x, tile_px_h * y, tile_px_w, tile_px_h};
             DFSDL_UpperBlit(surface, &vp, tile, NULL);
-            auto handle = Textures::loadTexture(tile);
+            auto handle = Textures::loadTexture(tile, reserved);
```
myk002, line 144: if `reserved == true` we should do a pre-check before starting the loop to ensure that all tiles will fit in the remaining reserved buffer, and fall back to making them all dynamic if they won't all fit.

library/modules/Textures.cpp
```diff
@@ -292,13 +368,28 @@ static void uninstall_reset_point() {
     INTERPOSE_HOOK(tracking_stage_new_arena, logic).remove();
 }
 
+static void reserve_static_range() {
+    reserved_range.init(enabler->textures.init_texture_size);
```
myk002, line 372: we should check to make sure that the number of current textures == enabler->textures.init_texture_size. if not, we're in an invalid state and the reserved range cannot be initialized. in this case, we should print a warning and continue, and all reserved textures should fall back to becoming dynamic

library/modules/Textures.cpp
shevernitskiy: It will be hidden behavior.
Also  you said we can't touch raws vector, when game does. So this is short interval when textures can't be loaded.
We can make kinda delayed queue for such case. But i think it is too much:) who need load dynamic texposes in this particular short interval?

library/modules/Textures.cpp
shevernitskiy: it can be, but dev can count on reserved range during development (it is a special range  for special purposes)
what if user will install to many plugins and will getting error because some texposes will be dynamic?

library/modules/Textures.cpp
myk002: I said we couldn't *extend* the raws vector. the buffer range will be safe to modify at all times

library/modules/Textures.cpp
shevernitskiy: and again it is hiden behaviour... does this fallback should be automatic?

library/modules/Textures.cpp
myk002: I think it's more important that this code always returns a valid handle. whether a texture is reserved or dynamic should not materially matter to the *code* that uses it. the experience will just gracefully degrade as pressure is increased, which is what we want. If a texture cannot be loaded because there is no room in the reserved buffer and loading_state is true, then we should return a valid handle that points to texpos (0) and we should try to add the real texture to the vector the next time the handle is resolved.

shevernitskiy: in case of texpos (0) and  `loading_state`user will receive incorrect texpos for ever call..
next handle resolve can be in loading_state as well

shevernitskiy: if you insist on loading during loading state, so it can be delayed queue, that will be loaded after loading state become false

library/modules/Textures.cpp
shevernitskiy: stl vector not thread safe, are you sure?

myk002: there isn't any concurrent access here, is there? we're running in DF's simulation thread, and you have locks protecting the Textures API from concurrent access from DFHack threads

shevernitskiy:
I almost lost in logic branches:)

library/modules/Textures.cpp
```diff
@@ -28,9 +30,36 @@ namespace DFHack {
...
+    void init(int32_t start) {
+        this->start = start;
+        this->end = start + ReservedRange::size;
+        this->current = start;
+        this->is_installed = true;
+    }
+    long get_new_texpos() {
+        if (this->current == this->end)
+            return -1;
+        current = this->current;
```
myk002, line 43: I don't see another `current` being declared anywhere. Isn't this line just assigning to itself?

myk002: moreover, these three lines can be replaced by a single `return current++`

library/modules/Textures.cpp
```diff
@@ -292,13 +412,33 @@ static void uninstall_reset_point() {
...
+    if (static_cast<size_t>(enabler->textures.init_texture_size) != enabler->textures.raws.size()) {
+        WARN(textures).print(
+            "reserved range can't be installed! all textures will be loaded to dynamic range!");
+        return;
+    }
+    reserved_range.init(enabler->textures.init_texture_size);
+    auto dummy_surface =
+        DFSDL_CreateRGBSurfaceWithFormat(0, 0, 0, 32, SDL_PixelFormatEnum::SDL_PIXELFORMAT_RGBA32);
+    for (int32_t i = 0; i < ReservedRange::size; i++) {
+        add_texture(dummy_surface);
```
myk002, line 425: at program close, DF will attempt to free all these textures. Is the refcount incremented somewhere so that it doesn't end up doing a double free?

library/modules/Textures.cpp
```diff
@@ -118,30 +160,57 @@ std::vector<TexposHandle> slice_tileset(SDL_Surface* surface, int tile_px_w, int
...
     auto handle = reinterpret_cast<uintptr_t>(surface);
     g_handle_to_surface.emplace(handle, surface);
     surface->refcount++; // prevent destruct on next FreeSurface by game
-    auto texpos = add_texture(surface);
-    g_handle_to_texpos.emplace(handle, texpos);
+
+    if (reserved && reserved_range.is_installed) {
+        auto texpos = reserved_range.get_new_texpos();
+        if (texpos != -1) {
+            insert_texture(surface, texpos);
```
myk002, line 176: this should also decrement the refcount of the dummy texture it's replacing

shevernitskiy: there is just one dummy texture placed all over reserved range... it has refcount = 1...
if we decrement, it will be destroyed

library/modules/Textures.cpp
shevernitskiy: game will perform `FreeSurface` for entire `textures.raws` (i guess)... so every texture with `refcount = 1` will be destroyed (here we destroy dummy)
our cached surfaces has `refcount = 2` and it will be decrmented to 1..
then `Texture::cleanup()` will be called in `Core` shutdown sequence, there we will one more time call `FreeSurface` for our cached surfaces and finaly destroy it

library/modules/Textures.cpp
myk002: I'm not entirely sure of the code path here, but this is what I'm concerned about.
On close, DF would iterate through the textures vector and call FreeSurface on each texture. Since the dummy texture appears more than once, it will get passed to FreeSurface many times. The first time it passes it to FreeSurface, the refcount would be decremented to 0 and the surface would be deleted. What happens when DF tries to pass the dummy surface to FreeSurface for the second time?

shevernitskiy: according to SDL docs passing nullptr to `FreeSurface` is safe...

shevernitskiy: Ok, found impl https://github.com/libsdl-org/SDL/blob/SDL2/src/video/SDL_surface.c#L1572

myk002: The pointer will *not* be null -- it will be the dummy surface pointer that we put in the vector. The second call to FreeSurface would be operating on freed memory. Maybe the safest and simplest approach would be to set the `SDL_DONTFREE` flag on the dummy surface?

myk002:
sorry for the conflicting changes in Textures.cpp -- `df::global::enabler` is `NULL` if we don't have the symbol map populated for a version, and the current usage was causing crashes and preventing us from investigating the upcoming 50.10 release

library/modules/Textures.cpp
shevernitskiy: Ok, i'll put a flag

library/modules/Textures.cpp
myk002: I note that we currently double free on exit with current code in 50.09-r3, so this is a real issue. I'll test to make sure this PR fixes that issue before merging, but otherwise this is looking pretty good

myk002:
There is still a double free going on here and DF is aborting when it frees the textures in its array. Could you double check the refcounting logic?

shevernitskiy:
On a win there is no errors on close, there visible only while debugging. That's why i missed that.
For me commenting out `reset_surface()` solves nothing, debbuger still catches exception.
Increasing refcount for dummy to equal size of reserved range eliminate exception.

myk002:
could you possibly update this branch with latest changes from develop?

shevernitskiy:
rebased onto develop

myk002 (approved):

PR #3717, opened 2023-09-01
zero ccache stats between runs

### LINKED ISSUE #3514: Clarify dreamfort help
Especially around how to keep miners on the right level. See discord discussion here: https://discord.com/channels/793331351645323264/807444467140788254/1124522719409606676

myk002:
and allow the cache to expand past the persisted max size during the build

Fixes #3514 just because I needed an innocuous change to test the rebuild speed

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -142,9 +142,11 @@ jobs:
       - name: Finalize cache
         if: inputs.platform-files
         run: |
+          ccache --show-stats --verbose
           ccache --max-size 40M
           ccache --cleanup
-          ccache --show-stats --verbose
+          ccache --max-size 500M
+          ccache --zero-stats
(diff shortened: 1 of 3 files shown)


lethosor:
Why is `--zero-stats` desirable?

myk002:
so we get statistics specific to the last build and not averages from the beginning of time

lethosor:
I guess a better question is: what are those stats used for? If they are used to determine which files to evict after multiple builds, I would think we _do_ want them to persist between runs. If they are informational only, then it probably doesn't matter.

myk002:
They are for measuring the effectiveness of the cache, for informational purposes only. clearing the stats does not affect the LRU measurements in the cache itself.

see: https://interrupt.memfault.com/blog/ccache-debugging#1-measure-total-cache-misses

PR #3728, opened 2023-09-03
[buildingplan] fix and reinstate heat safety filter

### LINKED ISSUE #3692: Remove reference to magma safe filter in buildingplan
The feature is currently disabled. We should remove it from the docs until it's working again

myk002:
It turns out the problem was *not* with item heat safety not being transferable to building heat safety. That was never true. The problem was that our job item flag matcher matches the logic of the vanilla matcher, and many item types, including screw pump components, are not checked for heat safety. therefore, no enormous corkscrews would ever match the filter if the heat safety filter was on.

I didn't want to change the function that mirrors vanilla logic, so I added heat safety detection in buildingplan logic itself.

Fixes #3692

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,7 @@ Template for new versions:
 ## New Features
 - `sort`: search and sort for squad assignment screen
 - `zone`: advanced unit assignment screens for cages, restraints, and pits/ponds
+- `buildingplan`: one-click magma/fire safety filter for planned buildings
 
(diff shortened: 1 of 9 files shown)


myk002:
merging so reddit user who reminded me about this can test the fix
https://www.reddit.com/r/dwarffortress/comments/15wwklt/comment/jx7r8fd/?utm_source=share&utm_medium=web2x&context=3

PR #3737, opened 2023-09-04
Update strangemood.rst example section

Tjudge1:
Updated example to accurately reflect arguments. Compliant with getopt-style parameter syntax.

### DIFF
--- a/docs/plugins/strangemood.rst
+++ b/docs/plugins/strangemood.rst
@@ -12,10 +12,10 @@ Usage
 
     strangemood [<options>]
 
-Examples
+Example
 --------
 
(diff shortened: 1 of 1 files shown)


docs/plugins/strangemood.rst
```diff
@@ -12,10 +12,10 @@ Usage
 
     strangemood [<options>]
 
-Examples
+Example
```
myk002, line 15: The underline has to match the length of the text

docs/plugins/strangemood.rst
```diff
@@ -12,10 +12,10 @@ Usage
...
 --------
```
myk002, line 16:
```suggestion
-------
```

PR #3742, opened 2023-09-06
Fix unchecked lua_tostring calls

lethosor:


myk002 (commented):
It still asserts when building RelWithDebInfo, but that is expected. Throws an exception in Release (as expected):

```c
[DFHack]# :lua @dfhack.units.getUnitsByNobleRole('manager')
table: 0x7fff7e074890
1                        = <unit: 0x7fff85d8a950>
[DFHack]# :lua @dfhack.units.getUnitsByNobleRole('manofwar')
table: 0x7fffdc0997e0
[DFHack]# :lua @dfhack.units.getUnitsByNobleRole()
(lua command):1: bad argument #-1 to 'getUnitsByNobleRole' (string expected, got function)
stack traceback:
        [C]: in field 'getUnitsByNobleRole'
        (lua command):1: in main chunk
        [C]: in function 'dfhack.safecall'
        /home/myk/src/dfhack-scripts/lua.lua:39: in local 'script_code'
        ./hack/lua/dfhack.lua:830: in function 'dfhack.run_script_with_env'
        (...tail calls...)
```

what's not completely expected is that it identifies the var as a function instead of nil. What function is it finding? I'm not sure if that implies that we could get a false negative here and absorb some random string.

myk002:
some discussion here: https://discord.com/channels/793331351645323264/1017471248567124049/1149114196127125604

PR #3743, opened 2023-09-06
Add `boolean` function for argparse

AndrielChaoti:
Implementing a boolean parser function for `argparse`

Accepts a "truthy" value like "yes" , "no",  "true", etc. and converts it to a lua boolean value. Throws an appropriate error should the value not be accepted.

Was discussing this on discord and thought it might be simple but useful addition to `argparse.`
(still needs docs update and likely further improvements)

### DIFF
--- a/library/lua/argparse.lua
+++ b/library/lua/argparse.lua
@@ -196,4 +196,17 @@ function coords(arg, arg_name, skip_validation)
     return pos
 end
 
+function boolean(arg, arg_name)
+    local toBool={["true"]=true,["yes"]=true,["y"]=true,["on"]=true,["1"]=true,
+                  ["false"]=false,["no"]=false,["n"]=false,["off"]=false,["0"]=false}
+
+    arg = string.lower(arg)
+    if toBool[arg] == nil then
+        arg_error(arg_name,
+            'unknown value: "%s"; expected "true", "yes", "false", or "no"')
+    end
+
+    return toBool[arg]
+end
+
 return _ENV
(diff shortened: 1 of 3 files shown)


library/lua/argparse.lua
```diff
@@ -196,4 +196,17 @@ function coords(arg, arg_name, skip_validation)
     return pos
 end
 
+function boolean(arg, arg_name)
+    local toBool={["true"]=true,["yes"]=true,["y"]=true,["on"]=true,["1"]=true,
+                  ["false"]=false,["no"]=false,["n"]=false,["off"]=false,["0"]=false}
```
myk002, line 201: might as well declare this above the function so it doesn't get needlessly reevaluated on every call

library/lua/argparse.lua
```diff
@@ -196,4 +196,17 @@ function coords(arg, arg_name, skip_validation)
...
+
+    arg = string.lower(arg)
+    if toBool[arg] == nil then
+        arg_error(arg_name,
+            'unknown value: "%s"; expected "true", "yes", "false", or "no"')
```
myk002, line 206:
```suggestion
            'unknown value: "%s"; expected "true", "yes", "false", or "no"', arg)
```

myk002: you probably should keep the original casing of `arg` saved somewhere so it can be displayed in this error message

myk002 (commented):
code looks good; just needs docs and a changelog entry

myk002 (approved):

PR #3753, opened 2023-09-09
[seedwatch] fix crash on bad seed ids in configs

myk002:
reported on Bay 12 forums: http://www.bay12forums.com/smf/index.php?topic=164123.msg8499609#msg8499609

seedwatch crashes when reading config data with invalid seed ids. the very first v50 seedwatch plugin did not record the seed id properly, and unlucky players who happened to use that version to save their game run into this crash.

### DIFF
--- a/plugins/seedwatch.cpp
+++ b/plugins/seedwatch.cpp
@@ -103,6 +103,10 @@ static bool validate_seed_config(color_ostream& out, PersistentDataItem c)
 {
     int seed_id = get_config_val(c, SEED_CONFIG_ID);
     auto plant = binsearch_in_vector(world->raws.plants.all, &df::plant_raw::index, seed_id);
+    if (!plant) {
+        WARN(config, out).print("discarded invalid seed id: %d\n", seed_id);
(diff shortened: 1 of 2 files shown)


plugins/seedwatch.cpp
```diff
@@ -103,6 +103,10 @@ static bool validate_seed_config(color_ostream& out, PersistentDataItem c)
 {
     int seed_id = get_config_val(c, SEED_CONFIG_ID);
     auto plant = binsearch_in_vector(world->raws.plants.all, &df::plant_raw::index, seed_id);
```
lethosor, line 105: I realize you didn't change this, but I'm pretty sure it could be reduced to `df::plant_raw::find(seed_id)`

lethosor (approved):

plugins/seedwatch.cpp
ab9rf: that was my doing, i forgot about `find`

ab9rf (approved):

PR #3755, opened 2023-09-10
update changelog for autofish changes.

AndrielChaoti:


myk002:
this *can* go here, but it would be more appropriate in the scripts repo

PR #3770, opened 2023-09-14
check-structures-sanity: add support for GCC std::map

lethosor:
still could use cleanup - opening so I don't lose track of it


myk002:
the code looks useful -- could we possibly merge this as-is?

Issue #3768, opened 2023-09-12
[beta] [linux] Steam Overlay and game streaming do not work when started as 'DFHack' in Steam


Jajcus:
When I was running vanilla Dwarf Fortress I was able to use Shift-Tab to open Steam Overlay, which was useful for making notes.
After I installed DFHack and start the game via DFHack the steam overlay is not available any more.

Possibly related: game streaming from other PC does not work. I have only tried that after installing DFHack, so I am not sure if this is also a problem in unmodded Dwarf Fortress, but it might be caused by the missing steam overlay which seems to affect DFHack only.

By the way: thanks for the native Linux build! I really do appreciate that.

Jajcus:
With the new DF beta release v50.10 I can start Dwarf Fortress using 'Dwarf Fortress' and not 'DFHack' entry in Steam again.  And when started this way both streaming and steam overlay work as expected.

The problem remains when started as 'DFHack'.

ab9rf:
I cannot replicate this on Windows, but I'm going to guess that this is with the native linux build, which suggests that the issue is that the dfhack launcher for linux is not properly propagating whatever environmental information the Steam client uses to communicate to the game that it's running under the client when it launches DF.

Jajcus:
yes. I think it might be the ``LD_PRELOAD`` environment variable - the ``dfhack`` script sets its without preserving previous value. I will check this…

Jajcus:
Yes, that was that. I'll try to make a pull request.

myk002:
There is a separate issue at hand here, and that is that we need to relaunch DF from steam when DFHack is launched from Steam. This is the pattern we use on Windows to ensure both DFHack and DF are properly attributed play time. The relaunch mechanism hasn't been written yet for Linux

ab9rf:
i figured as much but i don't know enough about the linux environment to implement the equivalent functionality to our windows relauncher

PR #3776, opened 2023-09-15
linux: preserve existing LD_PRELOAD content

Jajcus:
that is needed for Steam overlay and game streaming to work

fixes https://github.com/DFHack/dfhack/issues/3768

### DIFF
--- a/docs/about/Authors.rst
+++ b/docs/about/Authors.rst
@@ -83,6 +83,7 @@ Herwig Hochleitner      bendlas
 Hevlikn                 Hevlikn
 Ian S                   kremlin-
 IndigoFenix
+Jacek Konieczny         Jajcus
 James                   20k
 James Gilles            kazimuth
 James Logsdon           jlogsdon
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- Linux launcher: preserve LD_PRELOAD set by Steam, so Steam Overlay and game streaming works
 
 ## Misc Improvements
 
--- a/package/linux/dfhack
+++ b/package/linux/dfhack
@@ -60,7 +60,7 @@ fi
 case "$1" in
   -g | --gdb)
     shift
-    echo "set exec-wrapper env LD_LIBRARY_PATH='$LD_LIBRARY_PATH' LD_PRELOAD='$PRELOAD_LIB' MALLOC_PERTURB_=45" > gdbcmd.tmp
+    echo "set exec-wrapper env LD_LIBRARY_PATH='$LD_LIBRARY_PATH' LD_PRELOAD='${LD_PRELOAD:+$LD_PRELOAD:}$PRELOAD_LIB' MALLOC_PERTURB_=45" > gdbcmd.tmp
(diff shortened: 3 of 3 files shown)


myk002 (commented):
Thanks for this! Could you add a line to `docs/changelog.txt` under the "Fixes" header for the "Future" version?

Also, please feel free to add your name to the `docs/about/Authors.rst` file

lethosor (changes requested):
We already have PRELOAD_LIB for this purpose. Is that not usable in your case? 

The reason for the separate variable is because if LD_PRELOAD is set, *every process in the launcher script* will preload the specified libraries. That includes every `echo` command, `gdb`, `strace`, `stty`, etc. In the case of GDB and others, that can cause unexpected behavior.

If we need to pull in LD_PRELOAD because of limitations on the Steam side, we should also clear LD_PRELOAD after adding its contents to PRELOAD_LIB.

Jajcus:
> Thanks for this! Could you add a line to docs/changelog.txt under the "Fixes" header for the "Future" version?>>
>
> Also, please feel free to add your name to the docs/about/Authors.rst file

Will do that tomorrow, with other changes.

> We already have PRELOAD_LIB for this purpose. Is that not usable in your case?

I do put LD_PRELOAD into PRELOAD_LIB, as that is used to set LD_PRELOAD (the actual mechanism to preload libraries in Linux).
Now I think this is not the optimal solution, though I do not fully understand what is the meaning of PRELOAD_LIB, so I just did the simplest thing.

I will try to find something a bit better.

> The reason for the separate variable is because if LD_PRELOAD is set, every process in the launcher script will preload the specified libraries.

This should not be a problem – that is what steam expects with its use of LD_PRELOAD and preloading libs in Linux is AFAIK very cheap (these are called 'shared libraries' for a reason).

> In the case of GDB and others, that can cause unexpected behavior.

Running GDB without the LDD environment the application requires can also give unexpected results.

Jajcus:
I have updated the commit not to modify ``$PRELOAD_LIB``, just make sure  current ``$LD_PRELOAD`` contents is preserved everywhere where ``$LD_PRELOAD`` is set.

This would probably only matter in case the script is run recursively or if anything else uses ``$PRELOAD_LIB`` set there, which does not seem to be the case. Seems reasonable thing to do anyway.

lethosor:
> Now I see how it is currently handled – my current change would also be cover that – LD_PRELOAD is only set when it was already being set to $PRELOAD_LIB.

Ok, this is fine. What I had in mind was actually unsetting `LD_PRELOAD` after adding its contents to `PRELOAD_LIB`, just to avoid modifying 10 places in the same way, but your way works too, and we already have some duplication between branches.

> This should not be a problem – that is what steam expects with its use of LD_PRELOAD and preloading libs in Linux is AFAIK very cheap (these are called 'shared libraries' for a reason).

The problem I am pointing out is that we _shouldn't_ be preloading DFHack with LD_PRELOAD for processes other than DF. It's not a problem of not being "cheap" - it's a problem with introducing unexpected behavior. I don't have specific examples, but I know we have had trouble with preloading libdfhack into a process that isn't DF.

lethosor (approved):

Jajcus:
>  It's not a problem of not being "cheap" - it's a problem with introducing unexpected behavior.

Fair enough.

Anyway – may change does not change where the dfhack libraries preload is enabled.

PR #3780, opened 2023-09-16
reenable `spectate`

ab9rf:
by request from the community

do _not_ include in 50.10-r1, this should go through a beta cycle before being added to production


docs/changelog.txt
```diff
@@ -54,6 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `spectate`: automatically follow productive dwarves (returned to availability)
```
myk002, line 57: this would be a "New Tool", ne?

myk002 (commented):
also the docs probably need to have the "unavailable" tag removed

docs/changelog.txt
ab9rf: lol yes bad me

ab9rf:
closing because the merge history is ugly, will reopen

Issue #3555, opened 2023-07-10
Feature Req: New Sortable/Filterable Work Order Display Window


wbschultz:
I'm a relatively new player on the steam version, but even with just a meager fort, I find it painful to check if I'm regularly producing certain goods, if my output is high enough, etc. I would _**really**_ like a way to view all existing work orders in an easier-to-use window, with some sort of sorting and filters available. 

User story: As a user, I want to have access to view, sort, and filter existing work orders so that I can easily monitor which jobs are recurring, output levels, etc. 

The ability to tag and filter by tag, or organize orders into folders within this special display would be incredible so that I could see orders by various industries.

myk002:
thoughts from u/wheatleygone on [Reddit](https://www.reddit.com/r/dwarffortress/comments/14tne4z/comment/jrfnxjo/?utm_source=share&utm_medium=web2x&context=3):

It might be simpler if rather than allowing custom user-created folders, work orders were just categorized by the workshop they're performed at. The relative position of a "Cut ten gems" order to a "Make three wooden buckets" order does not matter, because they're performed at different workshops. It doesn't meaningfully change anything to separate the two out into different lists visually. Being able to collapse all of my mason orders or just view my glass forge orders would still be pretty great for keeping things organized, compared to the way it is now. It would also be quite intuitive since when you add work orders, the list is categorized by workshop.

Off the top of my head, the only thing that would complicate this is if there are multiple vastly different workshops that partially overlap in tasks, but I don't believe this is the case? There's overlap between the forges and their magma variants, and mill/quern, but those are cases where all of the tasks are shared so it's perfectly convenient for them to share a category.

I'm sure there are probably issues I'm not anticipating, but I think this could allow for a much better work order screen that sidesteps the issues with the list ordering.

Halifay:
Here is the code to sort by material or job type. It was written for orders.cpp file.

static bool compare_type(df::manager_order *a, df::manager_order *b)
{
    // Goal: Sort orders to easily find them in the list and to see dupclicated orders.
    // Sorting by job types

    // Divide orders by frequency first
    if (a->frequency != b->frequency)
    {
        return compare_freq(a, b);
    }

    // Determine if only one order has reaction_name
    bool a_has_reaction_name = !a->reaction_name.empty();
    bool b_has_reaction_name = !b->reaction_name.empty();

    if (a_has_reaction_name != b_has_reaction_name)
    {
        return a_has_reaction_name;
    }
[...]

PR #3781, opened 2023-09-16
Work orders manager enhancement

Halifay:
Added functions to reset and sort manager orders. Reset is needed to recheck conditions for all orders as they sometimes get stuck and manager doesn't recheck them. New sortings by job type and material might be useful to see duplicating orders or/and orders for similar items/with similar materials.

Also changed melee and ranged skill effectiveness thresholds to be more descriptive about strong soldiers. I think this change is too small for it's own branch.


Halifay:
pre-commit.ci autofix

plugins/lua/orders.lua
```diff
@@ -107,6 +119,27 @@ function OrdersOverlay:init()
...
+                frame={t=1, l=30},
+                label='sort by type',
+                key='CUSTOM_CTRL_T',
+                auto_width=true,
+                on_activate=do_sort_type,
+            },
+            widgets.HotkeyLabel{
+                frame={t=0, l=52},
+                label='sort by mat',
+                key='CUSTOM_CTRL_T',
```
azrazalea, line 139: (not a maintainer, just was curious and looking)

Looks like this hot tea label accidentally has the same key to activate as the one for type!

myk002 (commented):
could you possibly split the `sort.lua` changes out into a separate PR?

docs/plugins/orders.rst
```diff
@@ -17,10 +17,22 @@ Usage
     manager orders. It will not clear the orders that already exist.
 ``orders clear``
     Deletes all manager orders in the current embark.
+``orders reset``
```
myk002, line 20: I'd call this "recheck", not "reset". "reset" is too similar to "clear" in semantics.

myk002: and it might also be useful to differentiate between UI states -- `orders recheck` when viewing the details for a particular order could just reset that order, whereas it could reset all orders if called at any other time. alternately, the syntax could be `orders recheck [this]` where if `this` is specified, it will just reset the currently selected order, and all orders otherwise.

plugins/lua/orders.lua
Halifay: Thanks! It should be M.

Halifay:
> could you possibly split the `sort.lua` changes out into a separate PR?

Done https://github.com/DFHack/dfhack/pull/3783

docs/plugins/orders.rst
Halifay: Changed 'reset' to 'recheck'. 
We already have keybing for a single order on order conditions screen. Do you mean to rewrite it?

myk002: yes, I think it makes sense to merge that into the `orders` plugin.

docs/plugins/orders.rst
```diff
@@ -17,10 +17,22 @@ Usage
...
+``orders sort_type``
+    Sorts current manager orders by job type, making it easier to locate orders
+    that produce similar items. The sorting is done by reaction name, job type
+    and item subtype. If orders are equal by these fields the sorting falls back
+    to sort by frequency.
+``orders sort_material``
+    Sorts current manager orders by material, making it easier to locate orders
+    that produce items of the same material. The sorting is done by material type
+    and material category. If orders are equal by these fields the sorting falls back
+    to sort by job type.
```
myk002, line 35: I am very wary of using these to sort the base orders list. The reason for the existing sort is to work around deficiencies in the vanilla order processing code. any other sort order risks making some orders non-completable.

What I was thinking about instead was an alternate UI where you could sort, filter, and group orders by any criteria. the organization would be saved separately from the vanilla orders list, which would stay in the existing sorted order for safety.

Halifay: Well, they are not sorting automatically. This is not a problem at all if orders-sort enabled. 
About UI, I think that this is a great idea. I just don't think I can write something big right now.

docs/plugins/orders.rst
Halifay: Added `this` keyword to recheck only selected order. Moved single order invalidation to the orders plugin.

docs/about/Removed.rst
```diff
@@ -10,6 +10,13 @@ work (e.g. links from the `changelog`).
   :local:
   :depth: 1
 
+.. _workorder-recheck:
+
+workorder-recheck
+=================
+Tool to set 'Checking' status of the selected work order forcing manager to reevaluate its
+conditions. Merged into `orders`.
```
myk002, line 18:
```suggestion
Tool to set 'Checking' status of the selected work order, allowing conditions to be
reevaluated. Merged into `orders`.
```

docs/plugins/orders.rst
```diff
@@ -17,10 +17,27 @@ Usage
     manager orders. It will not clear the orders that already exist.
 ``orders clear``
     Deletes all manager orders in the current embark.
+``orders recheck [this]``
+    Sets the status to ``Checking`` (from ``Active``) of all work orders or one
+    currently viewed if 'this' option is passed. Work order conditions screen 
+    should be open in this case. This makes the manager reevaluate its conditions.
```
myk002, line 23:
```suggestion
    Sets the status to ``Checking`` (from ``Active``) for all work orders. if the
    "this" option is passed, only sets the status for the workorder whose condition
    details page is open. This makes the manager reevaluate its conditions.
```

docs/plugins/orders.rst
myk002: I think remove them until we can add a proper UI layer. Could you add the code as a comment at https://github.com/DFHack/dfhack/issues/3555 so your work isn't lost?

Halifay: Removed these functions and posted them in a comment.

plugins/lua/orders.lua
```diff
@@ -62,12 +62,24 @@ local function do_export()
...
+end
+
+local function do_sort_type()
+    dfhack.run_command('orders', 'sort_type')
+end
+
+local function do_sort_mat()
+    dfhack.run_command('orders', 'sort_material')
+end
+
```
myk002, line 76:
```suggestion
```

myk002 (approved):

PR #3783, opened 2023-09-18
Combat ratings polishing

Halifay:
Removed redundant uppercase in mental stability formula. Reworked thresholds for combat skill effectiveness formulas to have a higher 100 cap (more descriptive about very strong warriors).

### DIFF
--- a/plugins/lua/sort.lua
+++ b/plugins/lua/sort.lua
@@ -212,7 +212,7 @@ local function melee_skill_effectiveness(unit)
 end
 
 local function get_melee_skill_effectiveness_rating(unit)
-    return get_rating(melee_skill_effectiveness(unit), 350000, 2350000, 78, 64, 49, 35)
+    return get_rating(melee_skill_effectiveness(unit), 350000, 2750000, 64, 52, 40, 28)
 end
 
 local function make_sort_by_melee_skill_effectiveness_desc()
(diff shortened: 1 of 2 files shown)


plugins/lua/sort.lua
```diff
@@ -212,7 +212,7 @@ local function melee_skill_effectiveness(unit)
 end
 
 local function get_melee_skill_effectiveness_rating(unit)
-    return get_rating(melee_skill_effectiveness(unit), 350000, 2350000, 78, 64, 49, 35)
+    return get_rating(melee_skill_effectiveness(unit), 350000, 2750000, 64, 52, 40, 28)
```
myk002, line 215: I thought the point of the lower "maximum" thresholds was to allow more dwarves to have higher ratings. Do players get value from being able to distinguish between 2,350,000-rated dwarves and 2,750,000-rated dwarves? I fear that by raising the maximum here, we will lose the more important distinctions between dwarves with more common ratings.

Halifay: I adjusted colors to be around the same thresholds (in melee case). So for the casual players it is the same. The colors might be described as the following: red - can't handle aggressive wildlife, orange - in danger fighting a goblin, yellow - weak against any megabeast, green - weak against strong demons and megabeasts, light green - can handle an archangel. 
There is a big issue with marksdwarves, because data I built it on lacked legendary marksdwarves. They get to 100 very easily and therefore it is hard to distinguish between uber-super-mega commander who was training for 10 years and a rookie who was shooting goblins for 2 years. It is better for melee fighters, but at L+some they basically get to 100.
There is another argument for a larger visible scale that generic monsters are wastly varried in form/material/effects. In some occasions enemy might be considerably stronger than a generic archangel.
At last, there are some people (including me) who likes distinguishing the best of the best. For 10+ years forts it is not uncommon to have several 100. Both data analysis and my personal sparring observations tell that there is a difference after 100 cap.
To summarize, new thresholds are as good or better in describing relative strength using colors. On top of that they are more descriptive for legendary warriors.

myk002 (commented):
since this change is user-visible, could you add a note to the changelog (under "Misc Improvements")?

myk002 (approved):

Issue #3795, opened 2023-09-21
remotefortressreader.cpp - ambiguous function call to `lerp` when built on Visual Studio using C++20


master-spike:
In `plugins/remotefortressreader/remotefortressreader.cpp`

- line 133: `using namespace std;`
```

- lines 1636 to 1639:
float lerp(float a, float b, float f)
{
    return a + f * (b - a);
}
```

and while `cmath` is not explicitly included, it is included indirectly. Hence at lines 1799-1801:
```
send_unit->set_subpos_x(lerp(0, unit->path.path.x[0] - unit->pos.x, (float)(action->data.move.timer_init - action->data.move.timer) / action->data.move.timer_init));
send_unit->set_subpos_y(lerp(0, unit->path.path.y[0] - unit->pos.y, (float)(action->data.move.timer_init - action->data.move.timer) / action->data.move.timer_init));
send_unit->set_subpos_z(lerp(0, unit->path.path.z[0] - unit->pos.z, (float)(action->data.move.timer_init - action->data.move.timer) / action->data.move.timer_init));
```

The compiler cannot deduce whether to call `lerp` as defined in the source file or `std::lerp` as defined in the `cmath` header. (`std::lerp` was added to the standard in C++20)

On my machine I get the following error:
```
D:\DFHACK\DFHACK_DEVEL\dfhack\plugins\remotefortressreader\remotefortressreader.cpp(1799,187): error C2668: 'lerp': amb
iguous call to overloaded function [D:\DFHACK\DFHACK_DEVEL\dfhack\build\win64\VC2022\plugins\remotefortressreader\Remot
eFortressReader.vcxproj]
```

I am assuming that the programmer intended to use their own provided `lerp` function. To that end, the `using namespace std;` could be replaced by only using the parts of the standard library that were intended to be used in this file?

ab9rf:
`using namespace std` violates our coding guidelines anyway, so the real solution here is to remove that code smell

PR #3797, opened 2023-09-21
Fix lerp() name conflict on Windows

master-spike:
Fixed a potential build issue on some platforms on C++20 where `<cmath>` was getting shadow-included and causing a naming conflict with `lerp` due to the existence of `std::lerp` (#3795). Solution: `using namespace std;` was removed, and all instances of using `std` elements (in this case, `vector`, `map` and `string`) were changed to have the `std::` prefix (`std::vector`, `std::map`, `std::string`). As we are no longer using `std` namespace, the compiler can unambiguously know which `lerp` function we are referring to.

### DIFF
--- a/plugins/remotefortressreader/remotefortressreader.cpp
+++ b/plugins/remotefortressreader/remotefortressreader.cpp
@@ -130,7 +130,6 @@
 using namespace DFHack;
 using namespace df::enums;
 using namespace RemoteFortressReader;
-using namespace std;
 
 DFHACK_PLUGIN("RemoteFortressReader");
 
@@ -193,7 +192,7 @@ const char* growth_locations[] = {
 #include "df/art_image.h"
 #include "df/art_image_chunk.h"
 #include "df/art_image_ref.h"
-command_result loadArtImageChunk(color_ostream &out, vector <string> & parameters)
+command_result loadArtImageChunk(color_ostream &out, std::vector<std::string> & parameters)
 {
     if (parameters.size() != 1)
         return CR_WRONG_USAGE;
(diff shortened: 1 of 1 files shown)


plugins/remotefortressreader/remotefortressreader.cpp
```diff
@@ -193,7 +192,7 @@ const char* growth_locations[] = {
 #include "df/art_image.h"
 #include "df/art_image_chunk.h"
 #include "df/art_image_ref.h"
-command_result loadArtImageChunk(color_ostream &out, vector <string> & parameters)
+command_result loadArtImageChunk(color_ostream &out, std::vector<std::string> & parameters)
```
ab9rf, line 195: it's fairly common, and acceptable by our standards, to import just `vector` and `string` with 'using std::string; using std::vector;`

master-spike: > it's fairly common, and acceptable by our standards, to import just `vector` and `string` with 'using std::string; using std::vector;`

I did consider resolving it that way, there are already multiple instances of `std::vector` being written before I made any changes so I opted for that way instead, personally I only do `using std::xxx` if some types are getting unweildy to write out which wasn't really the case here (there are only 18 usages of `std::xxx` in the entire file).

ab9rf (approved):

lethosor (commented):
Would it have been possible to just specify `std::lerp` or `::lerp` instead of changing the other stuff? That can sometimes work, but not sure what the specifics were in this situation.

myk002 (approved):
seems fair, though I would have added more specific `using` clauses instead

PR #3798, opened 2023-09-21
Use SDL_ShowSimpleMessageBox in Core::fatal()

lethosor:
This allows the message to be displayed consistently in a dialog on all platforms.

image

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1459,12 +1459,9 @@ void Core::fatal (std::string output)
         con.print("\n");
     }
     fprintf(stderr, "%s\n", out.str().c_str());
-#ifndef LINUX_BUILD
-    out << "Check file stderr.log for details\n";
-    MessageBox(0,out.str().c_str(),"DFHack error!", MB_OK | MB_ICONERROR);
-#else
+    out << "Check file stderr.log for details.\n";
     std::cout << "DFHack fatal error: " << out.str() << std::endl;
-#endif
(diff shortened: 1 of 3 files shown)


library/Core.cpp
```diff
@@ -1506,6 +1503,12 @@ bool Core::InitMainThread() {
     std::cerr << "DFHack build: " << Version::git_description() << "\n"
          << "Starting with working directory: " << Filesystem::getcwd() << std::endl;
 
+    std::cerr << "Binding to SDL.\n";
+    if (!DFSDL::init(con)) {
+        fatal("cannot bind SDL libraries");
+        return false;
+    }
+
```
lethosor, line 1511: I'm hoping that moving this block to InitMainThread() is safe. This is needed so that the SDL function is bound before we call fatal() if the DF version is unknown.

I'm realizing now that there is also a fatal() call inside this block, so I should protect it somehow...

ab9rf: the only risk i can see is if somehow `InitSimulationThread` executed before, or in parallel with, `InitMainThread`

we call `initMainThread` from `dfhooks_init` and DF calls this before it starts the simulation thread, so it'll be completed before DF's simulation thread can ever call `dfhooks_update` and that appears to be the only code path that can reach `initSimulationThread`

so i think this should be safe, from a thread safety standpoint

ab9rf (approved):

PR #3800, opened 2023-09-22
Adjusted behaviour of "digtype" to only designate non-hidden squares, and designate them as auto-dig

master-spike:
The current behaviour of `digtype` designates all tiles on the map of the same type, regardless of visibility. The player might not want to have the locations of all the veins revealed to them, so designating visible veins of that type as "auto-dig" is probably more desirable and useful behaviour for players. I also took the liberty of adding a new option `+z` to this command which is analogous to the existing `-z` option but to only dig layers above. (Hence, the player can combine the two to only dig at the same z-level as the cursor).

Note that for designations other than the default we can't set them to autodig, it simply does not work. So the dig_auto flag is only set if the designation is default.

### DIFF
--- a/docs/plugins/dig.rst
+++ b/docs/plugins/dig.rst
@@ -50,7 +50,7 @@ Usage
     Designate circles. The diameter is the number of tiles across the center of
     the circle that you want to dig. See the `digcircle`_ section below for
     options.
-``digtype [<designation>] [-p<number>] [-z]``
+``digtype [<designation>] [-p<number>] [--zup|-u] [--zdown|-zu] [--cur-zlevel|-z] [--hidden|-h] [--no-auto|-a]``
     Designate all vein tiles of the same type as the selected tile. See the
     `digtype`_ section below for options.
 ``digexp [<pattern>] [<filter>] [-p<number>]``
@@ -119,9 +119,11 @@ the last selected parameters.
 digtype
 -------
 
-For every tile on the map of the same vein type as the selected tile, this
-command designates it to have the same designation as the selected tile. If the
-selected tile has no designation, they will be dig designated.
+For every tile on the map of the same vein type as the selected tile, this command
+designates it to have the same designation as the selected tile. If the selected
+tile has no designation, they will be dig designated. By default, only designates
+visible tiles, and in the case of dig designation, applies automatic mining to them
+(designates uncovered neighbouring tiles of the same type to be dug).
 
 If an argument is given, the designation of the selected tile is ignored, and
 all appropriate tiles are set to the specified designation.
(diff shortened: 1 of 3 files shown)


master-spike:
As discussed on discord - it might be good to implement the full vein designation as an option `--hidden`, with this behaviour as the default, as well as a possible `--no-auto` option in case players want to avoid the auto-designation digging into obstacles

docs/plugins/dig.rst
```diff
@@ -143,9 +143,10 @@ Designation options:
 ``clear``
     Clear any designations.
 
-You can also pass a ``-z`` option, which restricts designations to the current
-z-level and down. This is useful when you don't want to designate tiles on the
-same z-levels as your carefully dug fort above.
+You can also pass a ``-z`` and/or a ``+z``` option, which restricts designations to the
```
myk002, line 146: clever, but not in line with how DFHack options generally work. 

How about if we align `-z` with the definition used by `dig-now` https://docs.dfhack.org/en/stable/docs/tools/dig-now.html?highlight=dig-now
and add --zdown and --zup to mean "this z level and down" and "this z level and up"

master-spike: I've implemented these changes, and added single-character equivalent options.

`--zdown` == `-d`, `--zup` == `-u`, `--cur-zlevel` == `-z`.

master-spike:
As of right now, the original behaviour (digging all hidden tiles in non-auto mode) can be achieved using `--hidden` and `--no-auto` options (`-h` and `-a`) for short. Example behaviour

`digtype`
Screenshot 2023-09-24 124238

`digtype --hidden`
Screenshot 2023-09-24 124416

`digtype --no-auto`
image

`digtype --no-auto --hidden` - equivalent to old `digtype` by itself
image

docs/plugins/dig.rst
```diff
@@ -143,9 +145,18 @@ Designation options:
...
+Other options:
+
+``--zdown``, ``-d``
+    Only designates tiles on the cursor's z-level and below.
+``--zup``, ``-u``
+    Only designates tiles on the cursor's z-level and above.
+``--cur-zlevel``, ``-z``
+    Only designates tiles on the same z-level as the cursor.
+``--hidden``, ``-h``
+    Allows designation of hidden tiles, and using a hidden tile as the "palette".
```
myk002, line 157: What do you mean by `and using a hidden tile as the "palette"`? I don't understand that part.

myk002 (commented):
needs a line in `docs/changelog.txt` to describe the change

myk002:
`--hidden` could potentially imply `--no-auto` since there's no reason to mark a vein as "auto" if it's already completely selected, but it's fine the way it is too.

docs/plugins/dig.rst
master-spike: The selected tile on the cursor which is used to determine what type to dig. I think the word choice is poor on my part but couldn't think of a better word to describe it.

myk002: oh I see, so it will allow a hidden tile to be selected, and it will mark other tiles based on that tile's material? Is that...desired? If a player has a hidden tile with an unknown (to them) material selected when they run `digtype`, isn't that more likely to be a mistake?

master-spike: It's how the functionality originally was - I do agree that it's probably a mistake but perhaps there's someone out there who really like to live dangerously in that way so I left it in for the hidden mode.

myk002 (approved):

PR #3807, opened 2023-09-23
re-enable tubefill.

donhth:
Compiled and tested that it does indeed refill the tube with adamantine.


lethosor (commented):
Can you mention this in docs/changelog.txt? There should be some examples of re-enabled plugins from previous releases.

Also, update docs/plugins/tubefill.rst to remove the `unavailable` tag

donhth:
Note: the changelog doesn't tell the whole truth. It replenishes mined out adamantine and replaces all the mined out special stuff (clusters consisting of: water, magma, obsidian, gemstones) in the tube with adamantine.  

I think this is ok because:

1) Users interested in this tool are most likely only after the adamantine.
2) I believe the tool is clear enough about its intention to fill the tube with adamantine.

If there's community interest in replenishing the tube to its original state (as much as possible) then an option can always be added in for that later.  Better to have something than nothing, for players that want more adamantine now.

docs/changelog.txt
```diff
@@ -67,6 +67,11 @@ Template for new versions:
 
 ## Removed
 
+# 50.10-r2
+
+## New Plugins
+- `tubefill`: (reinstated) replenishes mined-out adamantine
+
```
myk002, line 74: We'll create the 50.10-r2 header when we tag the release. For now, the `tubefill` entry should go under "New Tools" under the "Future" heading.

myk002:
> Note: the changelog doesn't tell the whole truth. It replenishes mined out adamantine and replaces all the mined out special stuff (clusters consisting of: water, magma, obsidian, gemstones) in the tube with adamantine.

Feel free to add more info to the help file!

I do agree that replacing mined-out gems with adamantine is perfectly fine

donhth:
Alright, maybe I did it correctly this time. I think I get it now. 

I can add more info to the changelog regarding the differences if that's what you were referring to by 'help file.'  Otherwise, I'd prefer to do that in a separate PR, if that's ok.

myk002:
By "help file" I was referring to `tubefill.rst` -- that's what players will see online at https://docs.dfhack.org/en/latest/docs/tools/tubefill.html and in-game in `gui/launcher` when they type `tubefill`.

myk002 (commented):
code change looks good, though a note about how mined-out gem-studded walls will be replaced with adamantine in `tubefill.rst` would be welcomed

PR #3811, opened 2023-09-24
enable fortress mode tests in CI

myk002:
significant changes:

- tests on linux are enabled
- fortress mode tests are enabled
- test harness code is changed to wait for a number of ms instead of a number of frames
- tests don't start until DF initial prep is complete
- crashing tests are marked as failed and skipped instead of retried

### DIFF
--- a/ci/run-tests.py
+++ b/ci/run-tests.py
@@ -31,7 +31,7 @@ if args.test_dir is not None:
     if not os.path.isdir(args.test_dir):
         print('ERROR: invalid test folder: %r' % args.test_dir)
 
-MAX_TRIES = 5
+MAX_TRIES = 1
 
 dfhack = 'Dwarf Fortress.exe' if sys.platform == 'win32' else './dfhack'
 test_status_file = 'test_status.json'
--- a/ci/run-tests.py
+++ b/ci/run-tests.py
@@ -65,14 +65,12 @@ if not os.path.exists(init_txt_path):
 shutil.copyfile(init_txt_path, init_txt_path + '.orig')
 with open(init_txt_path) as f:
     init_contents = f.read()
-init_contents = change_setting(init_contents, 'INTRO', 'NO')
 init_contents = change_setting(init_contents, 'SOUND', 'NO')
 init_contents = change_setting(init_contents, 'WINDOWED', 'YES')
-init_contents = change_setting(init_contents, 'WINDOWEDX', '80')
-init_contents = change_setting(init_contents, 'WINDOWEDY', '25')
-init_contents = change_setting(init_contents, 'FPS', 'YES')
-if args.headless:
-    init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
+init_contents = change_setting(init_contents, 'WINDOWEDX', '1200')
+init_contents = change_setting(init_contents, 'WINDOWEDY', '800')
+#if args.headless:
+#    init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
 
 init_path = 'dfhack-config/init'
 if not os.path.isdir('hack/init'):
--- a/ci/run-tests.py
+++ b/ci/run-tests.py
@@ -31,7 +31,7 @@ if args.test_dir is not None:
     if not os.path.isdir(args.test_dir):
         print('ERROR: invalid test folder: %r' % args.test_dir)
 
-MAX_TRIES = 1
+MAX_TRIES = 5
 
 dfhack = 'Dwarf Fortress.exe' if sys.platform == 'win32' else './dfhack'
 test_status_file = 'test_status.json'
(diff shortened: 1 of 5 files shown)


ci/test.lua
```diff
@@ -189,39 +189,67 @@ local function ensure_title_screen()
                          dfhack.gui.getCurFocus(true)[1]))
 end
 
-local function is_fortress(focus_string)
-    focus_string = focus_string or dfhack.gui.getCurFocus(true)
-    return focus_string == 'dwarfmode/Default'
+local function is_fortress()
+    return dfhack.gui.matchFocusString('dwarfmode/Default')
```
lethosor, line 193: I think a strict match is better than a prefix match here. We don't want to be stuck in a submenu unexpectedly.

ci/test.lua
```diff
@@ -189,39 +189,67 @@ local function ensure_title_screen()
...
+    end
+    df.global.gps.precise_mouse_y = df.global.gps.mouse_y * df.global.gps.tile_pixel_y
+    df.global.enabler.tracking_on = 1
+    df.global.enabler.mouse_lbut = 1
+    df.global.enabler.mouse_lbut_down = 1
+    dfhack.screen._doSimulateInput(scr, {})
+end
+
+local function load_first_save(scr)
+    if #scr.savegame_header == 0 then
```
lethosor, line 213: To avoid unexpectedly altering a user fort, I think it would be good to check that there is exactly one save, with the expected name, and bail out if not.

ci/test.lua
```diff
@@ -189,39 +189,67 @@ local function ensure_title_screen()
...
             print('Loaded fortress map')
             -- pause the game (if it's not already paused)
             dfhack.gui.resetDwarfmodeView(true)
             return
         end
-        local scr = dfhack.gui.getCurViewscreen(true)
-        if focus_string == 'title' or
-                focus_string == 'dfhack/lua/load_screen' then
+        local scr = dfhack.gui.getDFViewscreen(true)
+        if dfhack.gui.matchFocusString('title') then
```
lethosor, line 235: Similar comment regarding prefix matching - we don't want to be in a submenu here.

ci/test.lua
```diff
@@ -189,39 +189,67 @@ local function ensure_title_screen()
...
+            -- TODO: reinstate loading of a specified save dir; for now
+            -- just load the first possible save, which will at least let us
+            -- run fortress tests in CI
             -- qerror()'s on falure
-            dfhack.run_script('load-save', config.save_dir)
-        elseif focus_string ~= 'loadgame' then
+            -- dfhack.run_script('load-save', config.save_dir)
+            load_first_save(scr)
+        elseif dfhack.gui.matchFocusString('loadgame') then
```
lethosor, line 242: Should be negated like before, I think.

Prefix match is probably ok for this one because it has no submenus, and even if it did, we're just waiting for the screen to disappear.

ci/test.lua
myk002: `dwarfmode/Default` *is* a strict match here. the "`/Default`" part indicates that no other panel is covering the map

ci/test.lua
myk002: good idea; added

ci/test.lua
myk002: done (by restricting to "title/Default")

ci/test.lua
lethosor: Right. Good point.

ci/run-tests.py
```diff
@@ -31,7 +31,7 @@
     if not os.path.isdir(args.test_dir):
         print('ERROR: invalid test folder: %r' % args.test_dir)
 
-MAX_TRIES = 5
+MAX_TRIES = 1
```
myk002, line 34: This change was made so we keep the original stdout.log and stderr.log. further attempts were always failing in the same way, so we just lost information by having multiple retries. My stance is that if a test is making DF crash, we should fix the test instead of hiding it with retries. To my knowledge, we don't have flaky tests where retries are warranted.

myk002: crashing tests are now marked as failed and skipped instead of retried

ci/run-tests.py
```diff
@@ -65,14 +65,12 @@ def change_setting(content, setting, value):
...
 init_contents = change_setting(init_contents, 'WINDOWED', 'YES')
-init_contents = change_setting(init_contents, 'WINDOWEDX', '80')
-init_contents = change_setting(init_contents, 'WINDOWEDY', '25')
-init_contents = change_setting(init_contents, 'FPS', 'YES')
-if args.headless:
-    init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
+init_contents = change_setting(init_contents, 'WINDOWEDX', '1200')
+init_contents = change_setting(init_contents, 'WINDOWEDY', '800')
+#if args.headless:
+#    init_contents = change_setting(init_contents, 'PRINT_MODE', 'TEXT')
```
myk002, line 73: commented out since TEXT mode is no longer supported -- but may be supported again in the future.
in the meantime, we launch with Xfrb to provide an X screen on which to run the graphical window

PR #3813, opened 2023-09-25
autolabor: fix #3812

### LINKED ISSUE #3812: autolabor leaves external_flag set when world or plugin is unloaded
so the labor system might be disabled incorrectly when a new fort that does not have autolabor enabled is loaded

ab9rf:
make sure autolabor resets the work detail bypass flag whenever autolabor is unloaded for _any_ reason

i tested `disable autolabor`, `unload autolabor`, and unloading a fort with autolabor enabled; in all cases the work detail bypass flag was cleared as desired

closes #3812

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- `autolabor`: now unconditionally clears ``game.external_flag`` when unloading a fort or the plugin
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


docs/changelog.txt
```diff
@@ -56,6 +56,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- `autolabor`: now unconditionally clears ``game.external_flag`` when unloading a fort or the plugin
```
myk002, line 59:
```suggestion
- `autolabor`: now unconditionally re-enables vanilla work details when the fort or the plugin is unloaded
```

myk002 (approved):

PR #3818, opened 2023-09-27
Eventmanager cleanup

master-spike:
Pulled out the event handler calls outside of the iterations over `df::global` structures so that if any of the calls to `handleEvent` change those structures those iterators aren't invalidated. Also made a few changes to tidy up some things here and there.

Changed the way `manageConstructionEvent()` works so that - rather than doing a binary search for every one of the constructions in our memorised set to check if its been destroyed, instead does a linear pass over the game's construction vector (which had to be done regardless to check for new constructions) and moves those constructions which intersect to a new set, and whatever remains in the old set is our destroyed constructions.

Range based for loops over map is now written as a structured binding for clarity. (no more ugly `.second`).

In `manageUnitAttackEvent` replaced the weird `alreadyDone` (it was a map of maps to ints which represented bools?) with a simple hashset of pairs (hash function provided - `hash_pair`) to check if that interaction was already done.

### DIFF
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -43,6 +43,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <array>
+#include <utility>
 
 namespace DFHack {
     DBG_DECLARE(eventmanager, log, DebugCategory::LINFO);
@@ -246,6 +247,15 @@ static int32_t reportToRelevantUnitsTime = -1;
 //interaction
 static int32_t lastReportInteraction;
 
+struct hash_pair {
+    template<typename A, typename B>
+    size_t operator()(const std::pair<A,B>& p) const {
+        auto h1 = std::hash<A>{}(p.first);
+        auto h2 = std::hash<B>{}(p.second);
+        return h1 ^ (h2 << 1);
+    }
+};
+
 void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event event) {
     static bool doOnce = false;
 //    const string eventNames[] = {"world loaded", "world unloaded", "map loaded", "map unloaded", "viewscreen changed", "core initialized", "begin unload", "paused", "unpaused"};
@@ -276,9 +286,9 @@ void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event 
         gameLoaded = false;
 
         multimap<Plugin*,EventHandler> copy(handlers[EventType::UNLOAD].begin(), handlers[EventType::UNLOAD].end());
-        for (auto &key_value : copy) {
+        for (auto &[_,handle] : copy) {
             DEBUG(log,out).print("calling handler for map unloaded state change event\n");
-            key_value.second.eventHandler(out, nullptr);
+            handle.eventHandler(out, nullptr);
         }
     } else if ( event == DFHack::SC_MAP_LOADED ) {
         /*
@@ -375,8 +385,7 @@ void DFHack::EventManager::manageEvents(color_ostream& out) {
             continue;
         int32_t eventFrequency = -100;
         if ( a != EventType::TICK )
-            for (auto &key_value : handlers[a]) {
-                EventHandler &handle = key_value.second;
+            for (auto &[_,handle] : handlers[a]) {
                 if (handle.freq < eventFrequency || eventFrequency == -100 )
                     eventFrequency = handle.freq;
             }
@@ -439,8 +448,7 @@ static void manageJobInitiatedEvent(color_ostream& out) {
             continue;
         if ( link->item->id <= lastJobId )
             continue;
-        for (auto &key_value : copy) {
-            EventHandler &handle = key_value.second;
+        for (auto &[_,handle] : copy) {
             DEBUG(log,out).print("calling handler for job initiated event\n");
             handle.eventHandler(out, (void*)link->item);
         }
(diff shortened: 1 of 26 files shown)


library/modules/EventManager.cpp
```diff
@@ -902,6 +935,26 @@ static void manageEquipmentEvent(color_ostream& out) {
             equipment.push_back(item);
         }
     }
+
+    // now handle events
+    std::for_each(equipment_pickups.begin(), equipment_drops.end(), [&](InventoryChangeData& data) {
```
donhth, line 940: equipment_pickups != equipment_drops. 

Is this intentional?

master-spike: no, it was not

library/modules/EventManager.cpp
```diff
@@ -455,18 +463,21 @@ static void manageJobStartedEvent(color_ostream& out) {
...
-                handler.eventHandler(out, job);
-            }
+            new_started_jobs.emplace_back(job);
+        }
+    }
+    for (df::job* job : new_started_jobs) {
+        for (auto &[_,handle] : copy) {
+            // the jobs must have a worker to start
+            DEBUG(log,out).print("calling handler for job started event\n");
+            handle.eventHandler(out, job);
```
myk002, line 480: this still suffers from the problem that if a handler cancels the job (which deletes it and removes it from the list), it will crash (probably).

to improve safety, it might be better to move the handler calls back inline and check after each handler call that the previous link still points to this job

library/modules/EventManager.cpp
```diff
@@ -586,10 +595,19 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
-        for (auto &key_value : copy) {
-            EventHandler &handle = key_value.second;
+        new_jobs_completed.emplace_back(&job0);
+    }
+
+    for (df::job* job : new_jobs_completed_repeats) {
+        for (auto &[_,handle] : copy) {
+            DEBUG(log,out).print("calling handler for repeated job completed event\n");
+            handle.eventHandler(out, (void*) job);
+        }
```
myk002, line 605: what does moving this logic gain us? the `prevjobs` structure cannot be modified by the handlers (though the jobs could potentially be deleted, which would cause crashes both in the original code and in the proposed change)

library/modules/EventManager.cpp
```diff
@@ -711,29 +742,32 @@ static void manageBuildingEvent(color_ostream& out) {
...
     nextBuilding = *df::global::building_next_id;
 
+    std::for_each(new_buildings.begin(), new_buildings.end(), [&](int32_t building){
+        for (auto &[_,handle] : copy) {
+            DEBUG(log,out).print("calling handler for created building event\n");
+            handle.eventHandler(out, (void*)intptr_t(building));
+        }
+    });
+
     //now alert people about destroyed buildings
```
myk002, line 757: could you take this opportunity to move the destroyed buildings events before the new building events? this would solve a longstanding bug in our building location cache where a new building is added in the spot a removed building just vacated.

myk002:
could you also make sure to *test* all these event code changes? you can use `devel/eventful-client` to register debug listeners and sanity check behavior

master-spike:
Interestingly in testing I've found a bug in the pre-existing code that fires unit death events when units leave the map - probably because the only check it does is whether the unit is active or not

library/modules/EventManager.cpp
master-spike: Yeah, there might be no way around that. Somehow I'm able to hold a job open in the gm-editor with auto-update long after its completion though, when it's no longer findable in the linked list? I might do some experiments later to determine the lifetime of a job.

master-spike: Actually, I think it's fine - the jobs in `prevJobs` are actually cloned job structures (by `Job::cloneJobStruct()`) and so won't be deleted by the game.

library/modules/EventManager.cpp
master-spike: done!

master-spike:
As of the latest changes I have tested the events for the following events (using 2 forts, one with 300 pops and another newly embarked)
`UNIT_NEW_ACTIVE` - works as expected
`UNIT_DEATH` - works as expected (note: changed behaviour - now doesn't register units leaving the map as deaths)
`JOB_COMPLETED` - works as expected. For cancelled and repeated jobs works as expected too.
`ITEM_CREATED` - works as expected.
`JOB_INITIATED` - works as expected.
`JOB_STARTED` - works as expected.
`CONSTRUCTION` - works as expected.
`BUILDING` - works as expected.
`SYNDROME` - works as expected. (i.e. I see when dwarves in the tavern are getting drunk and stuff)
`INVENTORY_CHANGE` - works as expected (tested equipment use changes by drafting and undrafting squads so they'd strap their gear then put it in their hand and back again)
`UNIT_ATTACK` - works as expected

All the other event types I didn't test as I didn't change them.

library/modules/EventManager.cpp
myk002: oh that's good - I thought I remembered something like that, but I didn't see it when I was scanning the code for the review. So, this part can be refactored or not. You can refactor it for consistency with the other handlers or revert this section for fewer local variables. I'm ok with either way.

library/modules/EventManager.cpp
myk002: thanks! I keep on forgetting to do that. it's been bugging me for more than a year. It wasn't causing errors, just unnecessary O(N) lookups.

myk002:
> As of the latest changes I have tested the events for the following events

Awesome! Thank you!

library/modules/EventManager.cpp
```diff
@@ -635,22 +650,26 @@ static void manageUnitDeathEvent(color_ostream& out) {
...
         return;
     multimap<Plugin*,EventHandler> copy(handlers[EventType::UNIT_DEATH].begin(), handlers[EventType::UNIT_DEATH].end());
+    vector<int32_t> dead_unit_ids;
     for (auto unit : df::global::world->units.all) {
         //if ( unit->counters.death_id == -1 ) {
         if ( Units::isActive(unit) ) {
             livingUnits.insert(unit->id);
             continue;
         }
+        if (!Units::isDead(unit)) continue; // for units that have left the map but aren't dead
```
myk002, line 660: I had to check the difference between `isDead()` and `! isAlive()` -- you chose the correct one.

myk002 (commented):
please add changelog entries for:
- fixing the unit dead event (Fixes)
- changing the order of the events in the building handler (Misc Improvements)
- protecting against list/vector modification while sending events (Misc Improvements instead of Fixes since this wasn't actively causing reported crashes)

master-spike:
> > As of the latest changes I have tested the events for the following events
> 
> Awesome! Thank you!
> 
> Edit: just to check, did you test with _more than one registered handler_? Whenever we've found bugs in this code the past, it was with that use case.

Ah, I did not for everything, I'll go ahead and double check that.

myk002:
fantastic! Thank you!

myk002 (approved):

PR #3829, opened 2023-09-29
Plugin preservetombs - fix for assigned tombs to living units becoming unassigned when the unit dies

master-spike:
I noticed a bug in the game a while ago where a unit assigned a tomb in life, would lose that assignment when they die, which is clearly unintended behaviour. I haven't seen anyone file a report for this bug so I don't have a bug tracker entry to link, but I've written up a plugin which fixes this bug. It basically tracks tomb assignments and listens for a `UNIT_DEATH` event to see if a unit we are tracking ended up dying, so that their tomb is reassigned to them.


plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+// ASSIGN UNIT TO TOMB
+//
+//
+static bool assign_to_tomb(int32_t unit_id, int32_t building_id) {
+
+    const int unit_idx = Units::findIndexById(unit_id);
+    if (unit_idx == -1) return false;
+
+    df::unit* unit = world->units.all[unit_idx];
```
myk002, line 258:
```suggestion
    df::unit* unit = df::unit::find(unit_id);
```

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+        auto* tomb = virtual_cast<df::building_civzonest>(bld);
+        if (!tomb || !tomb->flags.bits.exists) continue;
+        if (!tomb->assigned_unit) continue;
+        if (Units::isDead(tomb->assigned_unit)) continue; // we only care about living units
+
+        auto it = tomb_assignments.find(tomb->assigned_unit_id);
+
+        if (it == tomb_assignments.end()) {
+            tomb_assignments.emplace(tomb->assigned_unit_id, tomb->id);
+            out.print("%s new tomb assignment, unit %d to tomb %d\n", plugin_name, tomb->assigned_unit_id, tomb->id);
```
myk002, line 213: these print statements should be `DEBUG`s, perhaps with a new `cycle` category

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // check if unit was assigned a tomb in life
+    auto it = tomb_assignments.find(unit_id);
+    if (it == tomb_assignments.end()) return;
+
+    // assign that unit to their previously assigned tomb in life
+    int32_t building_id = it->second;
+    if (!assign_to_tomb(unit_id, building_id)) return;
+
+    // success, print status update and remove assignment from our memo-list
+    out.print("Unit %d died - assigning them to tomb %d\n", unit_id, building_id);
```
myk002, line 190: this should be an INFO statement instead of a raw print (so players have control over whether it is output)

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+            if (it->second != tomb->id) {
+                out.print("%s tomb assignment to %d changed, (old: %d, new: %d)\n", plugin_name, tomb->assigned_unit_id, it->second, tomb->id);
+            }
+            it->second = tomb->id;
+        }
+
+    }
+
+    // now check our civzones for unassignment / deleted zone
+    std::erase_if(tomb_assignments,[&](const auto& p){
```
myk002, line 226: don't you want to do this block *before* the `check tomb civzones for assigned units` block? what if I've swapped assignments between two tombs?

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+DFhackCExport command_result plugin_load_data (color_ostream &out) {
+    cycle_timestamp = 0;
+    config = World::GetPersistentData(CONFIG_KEY);
+
+    if (!config.isValid()) {
+        DEBUG(config,out).print("no config found in this save; initializing\n");
+        config = World::AddPersistentData(CONFIG_KEY);
+        set_config_bool(config, CONFIG_IS_ENABLED, is_enabled);
+        set_config_val(config, CONFIG_CYCLES, 25);
```
myk002, line 140: this is even more frequent than `nestboxes`, which uses a super low cycle time of 50 ticks (half a second at 100 fps). I would think 100 is more appropriate here, which is once per second at 100 fps and once every 5 seconds at 20 fps. would that be ok?

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    is_enabled = get_config_bool(config, CONFIG_IS_ENABLED);
+    cycle_freq = get_config_val(config, CONFIG_CYCLES);
```
myk002, line 144: I highly recommend making this a constant (like in `nestboxes`) and not making it configurable. making the tick count configurable in `autobutcher` causes confusion, is not useful to the end user, and was a mistake. I plan to remove that code.

docs/plugins/preserve-tombs.rst
```diff
@@ -0,0 +1,22 @@
...
+If you find that the tombs you assign to units get unassigned from them when
+they die (e.g. your nobles), this tool can help fix that.
+
+Usage
+-----
+
+::
+
+    enable preserve-tombs
+    preserve-tombs status
```
myk002, line 17: I suggest printing status when no options are passed as well, that is, changing the syntax to `preserve-tombs [status]`

docs/plugins/preserve-tombs.rst
```diff
@@ -0,0 +1,22 @@
+preserve-tombs
+==============
+
+.. dfhack-tool::
+    :summary: Fix tombs being unassigned to units on death
+    :tags: fort bugfix
```
myk002, line 6: Be sure to add this tool to `gui/control-panel` in the FORT_SERVICES list

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+        out.print("%s wrong usage", plugin_name);
+        return CR_WRONG_USAGE;
+    }
+    else {
+        out.print("%s is currently %s\n", plugin_name, is_enabled ? "enabled" : "disabled");
+        if (is_enabled) {
+            out.print("tracked tomb assignments:\n");
+            std::for_each(tomb_assignments.begin(), tomb_assignments.end(), [&out](const auto& p){
+                auto& [unit_id, building_id] = p;
+                out.print("unit %d -> building %d\n", unit_id, building_id);
```
myk002, line 92: it would be useful to print the unit name here as well. see `Translation::TranslateName(&unit->name, false)`

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+    if (enable != is_enabled) {
+        is_enabled = enable;
+        DEBUG(config,out).print("%s from the API; persisting\n",
+                                is_enabled ? "enabled" : "disabled");
+        set_config_bool(config, CONFIG_IS_ENABLED, is_enabled);
+        EventManager::registerListener(EventManager::EventType::UNIT_DEATH, assign_tomb_handler, plugin_self);
+        if (enable)
+            update_tomb_assignments(out);
+    } else {
+        EventManager::unregisterAll(plugin_self);
```
myk002, line 118: Be sure to also unregister the handler when the plugin or the world is unloaded.

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+    tomb_assignments.erase(it);
+
+}
+
+
+// Update tomb assignments
+//
+//
+static void update_tomb_assignments(color_ostream &out) {
+
```
myk002, line 200: you need to update `cycle_timestamp` here, otherwise you'll run on every frame

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+        }
+        return CR_OK;
+    }
+}
+
+// event listener
+EventManager::EventHandler assign_tomb_handler(onUnitDeath, 0);
+
+DFhackCExport command_result plugin_enable(color_ostream &out, bool enable) {
+    tomb_assignments.clear();
```
myk002, line 103: this will clear tomb assignments if the plugin is loaded and running and then the player runs `enable preserve-tombs`, which is probably not what you want

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,269 @@
...
+        set_config_bool(config, CONFIG_IS_ENABLED, is_enabled);
+        set_config_val(config, CONFIG_CYCLES, 25);
+    }
+
+    is_enabled = get_config_bool(config, CONFIG_IS_ENABLED);
+    cycle_freq = get_config_val(config, CONFIG_CYCLES);
+    DEBUG(config,out).print("loading persisted enabled state: %s\n",
+                            is_enabled ? "true" : "false");
+
+    if (is_enabled) update_tomb_assignments(out);
```
myk002, line 148: this is still while things are loading and might be too early to be running. I'd wait for `plugin_onupdate` before running the business logic

plugins/preserve-tombs.cpp
master-spike: The previous block that iterates over the `TOMB_ZONE` vector would already have assigned them correctly so they wouldn't be erased here.

plugins/preserve-tombs.cpp
master-spike: Ah forgot - fixed it up now

plugins/preserve-tombs.cpp
master-spike: That probably would be fine yeah - I've adjusted the defaults

plugins/preserve-tombs.cpp
master-spike: I've adjusted `plugin_enable` to do `tomb_assignments.clear()` when going from enabled to disabled, and `update_tomb_assignments(out)` vice versa.

plugins/preserve-tombs.cpp
master-spike: Perhaps you may be right, I have made this change. I've also added option of forcing an immediate update by the user `preserve-tombs update`.

myk002: sorry, should have suggested that. to be consistent, other tools phrase it as `preserve-tombs now`

master-spike:
Alright, everything _should_ be ready now, I've tested it and it works as expected. Let me know @myk002 if I missed anything

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,285 @@
...
+        DEBUG(config,out).print("%s from the API, but already %s; no action\n",
+                                is_enabled ? "enabled" : "disabled",
+                                is_enabled ? "enabled" : "disabled");
+    }
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_shutdown (color_ostream &out) {
+    DEBUG(config,out).print("shutting down %s\n", plugin_name);
+
```
myk002, line 146:
```suggestion

// PluginManager handles unregistering our handler from EventManager,
// so we don't have to do that here
```

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,285 @@
...
+    int32_t unit_id = reinterpret_cast<std::intptr_t>(ptr);
+
+    // check if unit was assigned a tomb in life
+    auto it = tomb_assignments.find(unit_id);
+    if (it == tomb_assignments.end()) return;
+
+    // assign that unit to their previously assigned tomb in life
+    int32_t building_id = it->second;
+    if (!assign_to_tomb(unit_id, building_id)) {
+        DEBUG(event, out).print("Unit %d died - but failed to assign them to tomb %d\n", unit_id, building_id);
```
myk002, line 204:
```suggestion
        WARN(event, out).print("Unit %d died - but failed to assign them back to their tomb %d\n", unit_id, building_id);
```

plugins/preserve-tombs.cpp
```diff
@@ -0,0 +1,285 @@
...
+        return;
+    }
+    // success, print status update and remove assignment from our memo-list
+    INFO(event, out).print("Unit %d died - assigning them to tomb %d\n", unit_id, building_id);
```
myk002, line 208:
```suggestion
    INFO(event, out).print("Unit %d died - assigning them back to their tomb\n", unit_id);
```

myk002 (approved):

PR #3833, opened 2023-09-30
use df's translate_name function if available

ab9rf:
requires dfhack/df-structures#666

tested on windows, still needs testing on linux


myk002:
can't automatically update -- could you rebase this off latest develop?

ab9rf:
> can't automatically update -- could you rebase this off latest develop?

there was a conflict over the changelog that had to be merged out, done

docs/changelog.txt
```diff
@@ -59,6 +59,7 @@ Template for new versions:
 ## Fixes
 
 ## Misc Improvements
+- Translate: will use DF's ``translate_name`` function, if available, instead of the DFHack emulation
```
myk002, line 62: since this should be transparent to the end user, I don't think we need a changelog for this, at least not in this section. If it goes anywhere, it should be in an "Internals" section

library/modules/Translation.cpp
```diff
@@ -167,6 +174,23 @@ string Translation::TranslateName(const df::language_name * name, bool inEnglish
     CHECK_NULL_POINTER(name);
 
     string out;
+
+    auto fp = df::global::translate_name;
+
+    if (fp) {
+        DEBUG(translate).print("using df provided translate_name function\n");
+        typedef std::function<void(const df::language_name&, std::string&, bool, bool)> fun_type;
+        auto f = reinterpret_cast<fun_type*>(fp);
```
myk002, line 183: this would be terrible to require all callers to do. can we make codegen produce a more usable type signature?

ab9rf: not easily

myk002: That would be sufficient, yeah, and it would give us a place to wrap the functionality if we need to, like with cancel_job.

docs/changelog.txt
ab9rf: going to move it under API

PR #3841, opened 2023-10-01
centralize management of mouse state

myk002:
Thank you @lethosor for [bringing this up](https://github.com/DFHack/dfhack/commit/cd12e41fd7aa7d7a52c9e9cab97d89273bef0764#r128858036). so much hackery could be removed.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -3952,6 +3952,14 @@ Misc
   of keycodes to *true* or *false*. For instance, it is possible to use the
   table passed as argument to ``onInput``.
 
+  You can send mouse clicks as will by setting the ``_MOUSE_L`` key or other
+  mouse-related pseudo-keys documented with the ``screen:onInput(keys)``
+  function above. Note that if you are simulating a click at a specific spot on
+  the screen, you must set ``df.global.gps.mouse_x`` and
+  ``df.global.gps.mouse_y`` if you are clicking on the interface layer or
+  ``df.global.gps.precise_mouse_x`` and ``df.global.gps.precise_mouse_y`` if
+  you are clickin on the map.
+
 * ``mkdims_xy(x1,y1,x2,y2)``
 
   Returns a table containing the arguments as fields, and also ``width`` and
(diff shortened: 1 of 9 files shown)


docs/dev/Lua API.rst
```diff
@@ -3952,6 +3952,14 @@ Misc
   of keycodes to *true* or *false*. For instance, it is possible to use the
   table passed as argument to ``onInput``.
 
+  You can send mouse clicks as will by setting the ``_MOUSE_L`` key or other
+  mouse-related pseudo-keys documented with the ``screen:onInput(keys)``
+  function above. Note that if you are simulating a click at a specific spot on
+  the screen, you must set ``df.global.gps.mouse_x`` and
+  ``df.global.gps.mouse_y`` if you are clicking on the interface layer or
+  ``df.global.gps.precise_mouse_x`` and ``df.global.gps.precise_mouse_y`` if
+  you are clickin on the map.
```
lethosor, line 3961:
```suggestion
  you are clicking on the map.
```
or
```suggestion
  you are clickin' on the map.
```
:)

library/modules/Screen.cpp
```diff
@@ -956,6 +956,18 @@ int dfhack_lua_viewscreen::do_notify(lua_State *L)
...
+void dfhack_lua_viewscreen::markInputAsHandled() {
+    if (!enabler)
+        return;
+
+    // clear text buffer
+    enabler->last_text_input[0] = '\0';
+
+    // mark clicked mouse buttons as handled
+    enabler->mouse_lbut = 0;
+    enabler->mouse_rbut = 0;
```
lethosor, line 968: Should this also handle mbut? or the _down fields?

lethosor (commented):
Looks like a good simplification. Haven't tested it myself yet. Supportive of waiting for 50.11-r2.

library/modules/Screen.cpp
myk002: We don't touch mbut in general so that map scrolling continues to work. The _down fields should not be modified since it can break dragging. Widgets capture the mouse on button click and then react to _down if they handled the click. if we mess with _down, then we can break whatever thinks the mouse belongs to them (including our own window dragging mechanism).

Issue #3844, opened 2023-10-04
Cannot set logistics automelt to melt masterworks


wintrparkgrl:
Melt stockpiles getting filled with masterworks over time. Should have an option for what quality to melt for. If I set a stockpile to accept masterworks and put auto-melt on it, it should melt the masterworks.

myk002:
Fixed in https://github.com/DFHack/dfhack/pull/3848

It should be out in a beta release next week or so, or you can try it now by switching to the DFHack Steam `testing` branch

PR #3848, opened 2023-10-07
[logistics] support optionally melting masterworks

myk002:
- add persisted config elements per stockpile\
- add CLI support
- add GUI elements for setting option from `stockpiles` overlay
- add button to stockpiles overlay for showing help

image


myk002:
Fixes #3844

PR #3907, opened 2023-10-20
Add to Lua library df time class

dhthwy:
Initial work.


dhthwy:
- Pretty sure this needs an API doc and probably changelog entry.
- Need a consensus on the name. 'Datetime' may be more befitting.
- File placement is currently 'library/lua.' Is this the best place?

dhthwy:
No clue why the ci build is failing. Builds for me.

`Make Error at /home/runner/work/dfhack/dfhack/library/git-describe.cmake:33 (message):
  git-describe failed: 128`

Maybe I am missing some magic git command to make it work?

myk002:
I vote for `datetime.lua`. The location in `library/lua` is good.

There seems to be an unintentional change to the `googletest` dependency submodule. That should get reverted

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+ 'Hematite',
+ 'Malachite',
+ 'Galena',
+ 'Limestone',
+ 'Sandstone',
+ 'Timber',
+ 'Moonstone',
+ 'Opal',
+ 'Obsidian',
+}
```
myk002, line 24: This feels like it *should* be a data structure enum (in the df-structures repo and accessed here as `df.months` or something), but I know the code that this came from did it this way too, so I'm not going to force that.

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+
+Time = defclass(Time)
+function Time:init(args)
+ self.year = args.year or 0
+ self.ticks = args.ticks or 0
+end
```
myk002, line 30:
```suggestion
Time.ATTRS{
    year=0,
    ticks=0,
}
```

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+ return self.year * 336 + (self.ticks / TU_PER_DAY)
+end
+function Time:getDayInYear() -- >>int<< Current day of the year
+ return math.floor (self.ticks / TU_PER_DAY)
+end
+function Time:getDayInMonth()
+ return math.floor ( (self.ticks % TU_PER_MONTH) / TU_PER_DAY ) + 1
+end
+function Time:getMonths() -- >>int<< Months as age (not including years)
+ return math.floor (self.ticks / TU_PER_MONTH)
```
myk002, line 41: note that Lua has a `//` operator for integer division which would be cleaner to use here instead of calling `math.floor()` everywhere

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+
+Time = defclass(Time)
+function Time:init(args)
+ self.year = args.year or 0
+ self.ticks = args.ticks or 0
+end
+function Time:getDays() -- >>float<< Days as age (including years)
+ return self.year * 336 + (self.ticks / TU_PER_DAY)
+end
+function Time:getDayInYear() -- >>int<< Current day of the year
```
myk002, line 34: is the intention for this to be one based or zero based?

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+  return Time{ year = (self.year - other.year - 1) , ticks = (TU_PER_YEAR + self.ticks - other.ticks) }
+ else
+  return Time{ year = (self.year - other.year) , ticks = (self.ticks - other.ticks) }
+ end
+end
+function Time:_debugOps(other)
+ print(self.year,self.ticks)
+ print(other.year,other.ticks)
+end
+
```
myk002, line 86: a `function now()` that returns a Time/DateTime object initialized to the current in-game date and time would be a useful addition here

library/lua/time.lua
myk002: though it would still be useful to keep the `init` function so you could normalize the values if ticks is < 0 or > TU_PER_YEAR

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+function Time:getDayInYear() -- >>int<< Current day of the year
+ return math.floor (self.ticks / TU_PER_DAY)
+end
+function Time:getDayInMonth()
+ return math.floor ( (self.ticks % TU_PER_MONTH) / TU_PER_DAY ) + 1
+end
+function Time:getMonths() -- >>int<< Months as age (not including years)
+ return math.floor (self.ticks / TU_PER_MONTH)
+end
+function Time:getYears() -- >>int<<
```
myk002, line 43: why does getDays return a float and partial days but getYears does not? I'm not exactly sure which style is better, but I think they should at least be consistent.

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+ elseif d % 10 == 1 then
+  d = tostring(d)..'st'
+ elseif d % 10 == 2 then
+  d = tostring(d)..'nd'
+ elseif d % 10 == 3 then
+  d = tostring(d)..'rd'
+ else
+  d = tostring(d)..'th'
+ end
+ return d
```
myk002, line 62: type changing confuses the linter. It would be clearer if you return the value in each of those branches instead of setting `d` and returning at the end

myk002:
API docs would go in `docs/dev/Lua API.rst` under the `Lua Modules` section
https://github.com/DFHack/dfhack/blob/develop/docs/dev/Lua%20API.rst

library/lua/time.lua
```diff
@@ -0,0 +1,87 @@
...
+ 'Timber',
+ 'Moonstone',
+ 'Opal',
+ 'Obsidian',
+}
+
+Time = defclass(Time)
+function Time:init(args)
+ self.year = args.year or 0
+ self.ticks = args.ticks or 0
```
myk002, line 29: calling it `year_tick` instead of `ticks` might be clearer

library/lua/time.lua
dhthwy: Thanks for the tip. Will try that way.

library/lua/time.lua
dhthwy: Good catch and one to remember to add to the docs. One based probably makes more sense.

library/lua/time.lua
dhthwy: I can't think of a good reason. I lean toward making everything a float for this stuff unless it absolutely makes no sense to do so. The user can always convert to int.

library/lua/time.lua
myk002: @ab9rf what do you think? should this list go in df-structures as an enum?

dhthwy: I misread at first, I assumed they were already there. Assumed wrong, as usual. Will wait for you guys to decide.

ab9rf: > @ab9rf what do you think? should this list go in df-structures as an enum?

there is no corresponding enum or `#define` in the bay12 info we have received, so if it does it should be scoped to DFHack. toady appears to use an `int32_t` whenever he need to represent a month and uses a magic 12 instead of a `#define` or enum for things like squad schedules

library/lua/time.lua
dhthwy: I'm thinking use case for this class will want an integer most of the time. But as we know, ints lose data. So to keep it consistent we either have methods for both floats and ints? Or add a parameter to the methods to indicate which type to return. Or use floats to preserve data and let the user convert.

So assuming we keep things simple and use floats, when the user needs ints, the code will look like:

string.format("%03d-%02d-%02d",agr_time:getYears()//1, agr_time:getMonths()//1, agr_time:getDayInMonth()//1)  Or math.floor instead of //1

%.0f isn't a viable substitute due to rounding up and formatting.

These points are probably all obvious. Just thought I'd mention it regardless. Also my prior experience with Lua is close to 0.

myk002: another option is to return the result in two integer variables, whole and part. e.g.

```lua
function Datetime:getYears()
    return self.years, self.year_tick
end

Comment (ab9rf) [reply library/lua/time.lua:43]: DF consistently manages times as a pair of integers `(years, year_tick)`. i strongly recommend that we assiduously use this format for representing times and time intervals and eschew entirely the use of floating point for this purpose

Comment (dhthwy) [reply library/lua/time.lua:43]: sgtm

**Comment by dhthwy** [comment]
This iteration may be more or less palatable, although untested, I'm sure it has bugs :)

As soon as we get the API finalized, I'll write the docs. I don't have any problem changing things upl then.

**Comment by dhthwy** [comment]
DateTime as it currently stands is no good for measuring time spans between two times. I'm debating on adding a TimeSpan class for this purpose. 

e.g. if we subtract times and call getMonth(). getMonth() returns the one-based month of the year. So 1 for the first month. But for time spans, we need it to be zero based.  

We can keep it to a single class, but then we have one class trying to do two separate things.

**Comment by myk002** [comment]
Yeah, I think separate `DateTime` and `Duration` classes make sense.

**Comment by dhthwy** [comment]
There was some common code amongst the classes so I added a DwarfCalendar parent  where I placed the various common utility/conversion/convenience functions -- not just for the child classes, but also for potential users. It was either that or move these functions outside the classes as local module functions. 

Don't have a problem changing things up again if this approach isn't going to fly.

File: library/lua/datetime.lua [hunk 9]
```diff
@@ -0,0 +1,237 @@
...
+end
+
+function DwarfCalendar:normalize()
+    if (self.year_tick > TICKS_PER_YEAR) then
+        self.year = self.year + (self.year_tick // TICKS_PER_YEAR)
+        self.year_tick = self.year_tick % TICKS_PER_YEAR
+    elseif (self.year_tick < 0) then
+        -- going backwards in time, subtract year by at least one.
+        self.year = self.year - math.max(1, math.abs(self.year_tick) // TICKS_PER_YEAR)
+        self.year_tick = self.year_tick % TICKS_PER_YEAR
```
myk002, line 99: won't self.year_tick still be negative here?

ab9rf: also do remember that lua's mechanics for `%` with respect to negative values may be unintuitive, due to the fact that lua rounds toward negative infinity instead of toward zero

dhthwy: I should have commented that part. self.year_tick will be positive. This behavior is documented in Lua's manual 3.4.1.  

Lua's % operator should be equivalent to:
function m(a, b)
  return a - (b * (a // b))
end

If I understand this correctly, the difference between modulo operator in Lua and C-lang+friends is the direction of quotient rounding via the use of the integer floor op where it effectively rounds toward minus infinity (i.e. -1.123 becomes -2). 

C-lang+friends round toward zero (-1.123 becomes -1). Consequently, in this direction, it ensures the remainder's sign agrees with the dividend, as it should. I believe Lua's math.fmod() rounds toward zero -- so not the same as %.

Incidentally, Python, Ruby, and Perl also use this variant of modulo.  I think it's safe to depend on this behavior going forward.

dhthwy: For example:
max year tick = 403200

if year = 2 and year_tick is -1, the result is:
year: 1
year_tick: 403199

library/lua/datetime.lua
```diff
@@ -1,72 +1,161 @@
...
+        self.ticks_per_day = (ticks_per_day > 0) and ticks_per_day or DWARF_TICKS_PER_DAY
+    end
+    self.ticks_per_month = self.ticks_per_day * DAYS_PER_MONTH
+    self.ticks_per_year  = self.ticks_per_month * MONTHS_PER_YEAR
+end
+
 function DwarfCalendar:addTicks(ticks)
     self.year_tick = self.year_tick + ticks
     self:normalize()
+    return self
```
myk002, line 100: so fluid. i like

dhthwy:
I started writing tests for this, and then I'll do the docs. Please let me know if you think this might need any major overhaul.

If we keep the Duration class vs duration methods in DateTime, it can technically represent a date and time, and still bound by the rules and system of the dwarven calendar, this is why I made it a child of DwarfCalendar.

In library/lua: dfcalendar.lua instead of datetime.lua might avoid conflating real time with dwarf time.  

DwarfCalendar can also be augmented to provide info on seasons as well as the expected caravans for the seasons, if we want that stuff.

For performance, if this might be in hot zones, I suppose we can precompute or cache some of these date calculations whenever year_tick changes. We'll want to ensure the caller can't directly mutate self.year_tick.

dhthwy:
Might be better implementing this in C++ so that it can benefit all. Given the simplicity of this class in its current form, it should be trivial. Tho, I'm mostly daft these days and it's been a few years since I've touched C++. 

In any case, Lua is a great lang for prototyping.

myk002:
All the current consumers of the logic are in Lua right now, so there's no reason to port it quite yet.

library/lua/datetime.lua
```diff
@@ -0,0 +1,486 @@
...
+function DateTime:monthsToTicks(months)
+    return months * self.ticks_per_month
+end
+
+function DateTime:addTicks(ticks)
+    if not ticks then qerror('ticks must be a number') end
+    self.year_tick = self.year_tick + ticks
+    self:normalize()
+    return self
+end
```
dhthwy, line 294: I'm not sure how robust we want our argument checking to be. We could do:

`ticks = tonumber(ticks)
if not ticks...
`

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+========
+
+A module to display, compare, and manipulate dates and times for the dwarven calendar.
+The *DateTime* and *Duration* objects in this module support *addition*, *subtraction*,
+and *<, <=, >, >=, ==* equality operators
+
+DateTime class
+==============
+
+* ``DateTime{ year = ..., year_tick = ..., ticks_per_day = ...}``
```
myk002, line 3941: the other examples in this file put the module name in the examples, which would make this `datetime.DateTime`

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+                  note: if ticks_per_day is given a value <= 1,
+                  it will default to fortress mode.
+  Examples::
+
+    -- create a new instance with year and year_tick set to
+    -- the current game time.
+    local dateTime = DateTime.now()
+    -- construct and set year to 1, year_tick to 1
+    dateTime = DateTime{ year = 1, year_tick = 1 }
+    -- set ticks_per_day to adventure mode
```
myk002, line 3958: this comment should also indicate that the time itself will be set to the beginning of year 0

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+    dateTime = DateTime{ ticks_per_day = df.game_mode.ADVENTURE }
+
+    local dateTimeOne = DateTime{ year = 1, year_tick = 1}
+    local dateTimeTwo = DateTime{ year = 1, year_tick = 2}
+    local deltaDateTime = dateTimeTwo - dateTimeOne
+    deltaDateTime = dateTimeOne + dateTimeTwo
```
myk002, line 3964: this example sets the same variable twice in a row. is that intentional?

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+    if (dateTimeOne < deltaDateTime) then print('is less than') end
+
+* ``dateTime:year()``
+
+  Returns an integer pair: *year*, *year tick count*.
+
+* ``dateTime:setYear(year)``
+
+  Sets the year. Returns *DateTime*.
```
myk002, line 3973: this should document what happens to the ticks when the year is set.

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+  assigning to year_tick directly due to normalization.
+  returns *DateTime*.
+
+* ``dateTime:dayOfYear()``
+
+  Returns an integer pair: *the current day of the year starting from one*, *day tick count*.
+
+* ``dateTime:setDayOfYear(day)``
+
+  Sets the day of year starting from one.
```
myk002, line 3988: what effect does this have on the existing year and tick?

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+
+* ``dateTime:monthTick()``
+
+  Returns *ticks since the beginning of a month*.
+
+* ``dateTime:daysToTicks(days)``
+
+  Returns *days converted to ticks*.
+
+* ``dateTime:monthsToTicks(months)``
```
myk002, line 4045: these look like class functions instead of instance functions. Should they be documented as `DateTime.` instead of `dateTime:`?

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+* ``DateTime{ year = ..., year_tick = ..., ticks_per_day = ...}``
+
+  A DateTime object is constructed with the following properties:
+
+  :year: A year. Defaults to zero.
+  :year_tick: A year tick. Defaults to zero.
+  :ticks_per_day: The rate at which time passes in a day.
+                  Defaults to DWARF/fortress mode.
+                  note: if ticks_per_day is given a value <= 1,
+                  it will default to fortress mode.
```
myk002, line 3950: this section should also explain what normalization is in this context and when it is applied.

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+
+* ``dateTime:toDuration()``
+
+  Returns a *Duration object*.
+
+* ``dateTime:normalize()``
+
+  Ensure's year_tick is within bounds of a year, otherwise
+  adjusts year and year_tick accordingly. Note: should only
+  call this function when directly modifying year_tick.
```
myk002, line 4085: this *should* not need to be exposed. what reason would callers have to modify `year_tick` directly?

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+* ``DateTime.now([game_mode])``
+
+  Returns *DateTime* initialized from the current game time.
+  If game_mode is not given. It detects the current game mode.
+  When in *adventure mode*, DateTime is initialized by adventure mode time.
+  Otherwise *defaults to fortress mode* time. This function sets tick rate by game mode.
+  If you need a custom tick rate, call dateTime:setTickRate(custom_ticks_per_day).
+
+Duration Class
+==============
```
myk002, line 4103:
```suggestion
Duration class
--------------
```

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+* ``duration:months()``
+
+  Returns an integer pair: *months since year zero*, *month tick count*
+
+* ``duration:yearsMonthsDays()``
+
+  Returns four parts of an elapsed time: *years*, *months*, *days*, *day tick count*
+
+DwarfCalendar
+=============
```
myk002, line 4125:
```suggestion
DwarfCalendar class
-------------------
```

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+  If game_mode is not given. It detects the current game mode.
+  When in *adventure mode*, DateTime is initialized by adventure mode time.
+  Otherwise *defaults to fortress mode* time. This function sets tick rate by game mode.
+  If you need a custom tick rate, call dateTime:setTickRate(custom_ticks_per_day).
+
+Duration Class
+==============
+
+Inherits DateTime's constructor and methods. Used for measuring the difference
+between DateTime and Duration objects.
```
myk002, line 4106: A lot of the DateTime instance functions don't really make sense for a Duration class. Would this be better the other way? Duration as the superclass and DateTime as the subclass?

docs/dev/Lua API.rst
dhthwy: Doesn't need to be exposed. Thinking was: 'you shouldn't be assigning to year_tick yourself, but if you insist then here's this function'

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+DateTime class
+==============
+
+* ``DateTime{ year = ..., year_tick = ..., ticks_per_day = ...}``
+
+  A DateTime object is constructed with the following properties:
+
+  :year: A year. Defaults to zero.
+  :year_tick: A year tick. Defaults to zero.
+  :ticks_per_day: The rate at which time passes in a day.
```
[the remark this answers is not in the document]
dhthwy: I'm not thrilled with the ticks_per_day name. Originally ticks_per_day was going to be a mode (I believe (df.game_mode is an enum with DWARF=0, ADVENTURE=1, AND LEGENDS=3). I think you mentioned letting the tick rate be customizable (if can change rate of time in game, then tick rate in a DateTime object may need to coincide).

We can substitute ticks_per_day for mode in the docs, but keep ticks_per_day (unexposed) as an object property so that setTickRate() can change it?

docs/dev/Lua API.rst
```diff
@@ -3928,6 +3928,211 @@ Examples:
...
+                  Defaults to DWARF/fortress mode.
+                  note: if ticks_per_day is given a value <= 1,
+                  it will default to fortress mode.
+  Examples::
+
+    -- create a new instance with year and year_tick set to
+    -- the current game time.
+    local dateTime = DateTime.now()
+    -- construct and set year to 1, year_tick to 1
+    dateTime = DateTime{ year = 1, year_tick = 1 }
```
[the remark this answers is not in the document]
dhthwy: admittedly the examples are kinda crappy. I was just being lazy.

docs/dev/Lua API.rst
dhthwy: These are calendrical functions that use the current tick rate for a particular instance (either by game mode or custom tick rate).  They could be class functions if we want all instances to share the same tick rate.  There used to be a mutable TICKS_PER_DAY module scoped variable for this, but taken out when custom tick rate logic was added.

docs/dev/Lua API.rst
dhthwy: The other way works. Some calendrical functions will need hoisting to Duration. And the addDays(), addMonths(), etc ones would probably be useful for Duration as well. In fact, it's not uncommon from implementations that I've looked at, to expose duration-style functions in their date/time objects.

It's either that, or DwarfCalendar can be the super class for common calendrical logic. I don't know if the calendar is actually moddable.  

Additionally, Duration wouldn't need to know about the calendar at all if it didn't include magnitudes of months and years. Just some thoughts.

Thanks for looking at this.

docs/dev/Lua API.rst
myk002: that's a good point. I forgot that the tick rate could differ.

How about making them class functions, but taking the mode as an optional parameter, defaulting to the current mode like the constructor does?

dhthwy: Fine with me; however, if we treat mode as a boolean, we won't be able to support a custom tick rate. Tick rate will either be fort or adv.

dhthwy: Or we could pass in an optional object and get the mode and stuff from there.

myk002: Either a boolean or a documented constant are fine, I think. The tick rate of the various modes is *not* moddable afaik. even `timestream` doesn't change the ticks per month -- it just advances the ticks faster than they would normally advance.

dhthwy:
Okay, in case it helps someone push this feature forward, I implemented the changes - I think. It didn't take long to do.  Untested and tests weren't updated to reflect yet.

ab9rf:
Closing as abandoned - this has had no effort in over two years

PR #3923, opened 2023-10-27
[burrows] rename burrows plugin to burrow to match the command name

myk002:
before I start making updates


plugins/lua/burrow.lua
lethosor, line 1: the `mkmodule` call will need to be updated

myk002: I made minimal code changes so this could be a clean rename. The plugin hasn't been added to CMakeLists.txt so it doesn't quite matter if it works yet.

PR #3929, opened 2023-10-28
adjust price of vermin swarms

myk002:
divide by 10000

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -58,6 +58,7 @@ Template for new versions:
 
 ## Fixes
 - `stockpiles`: hide configure and help buttons when the overlay panel is minimized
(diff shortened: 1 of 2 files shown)


quietust:
I don't believe this is correct - DF's item value code doesn't have a special case for vermin colonies, just `PETVALUE_DIVISOR` which happens to be 1000 for honey bees.

quietust (changes requested):
Change is incorrect - the actual game only checks `petvalue_divisor` and does not check the VERMIN_SOIL_COLONY flag at all.

myk002:
So the issue here is then that bees have a matgloss of -1, which cannot be used as a caste index. What should I do exactly in that case?

quietust:
In that case, the game itself uses caste 0 instead, which it looks like we're now doing.

PR #3935, opened 2023-10-30
[burrow] update and integrate with UI

myk002:
Adds UI integration for box and flood fill

reinstates existing CLI functionality

Fixes https://github.com/DFHack/dfhack/issues/3838


myk002:
CLI isn't finished, but merging so the UI can be tested and feedback can be gathered

PR #3955, opened 2023-11-03
[buildingplan] allow choosing of mechanisms when linking levers

myk002:
adds an overlay that extends the link lever panel. it offers a selection dialog for manual choices and heat safety filters for autoselection


silverflyone:
@myk002 it might be just me, but my build is complaining about a missing mechanisms.rst.

PR #3974, opened 2023-11-07
Add trap type to focus string

robob27:
When building type is trap, add the trap type to the focus string:

image

image

image

### DIFF
--- a/library/modules/Gui.cpp
+++ b/library/modules/Gui.cpp
@@ -380,8 +380,14 @@ DEFINE_GET_FOCUS_STRING_HANDLER(dwarfmode)
         case df::view_sheet_type::BUILDING:
             if (game->main_interface.view_sheets.linking_lever)
                 newFocusString = baseFocus + "/LinkingLever";
-            else if (auto bld = df::building::find(game->main_interface.view_sheets.viewing_bldid))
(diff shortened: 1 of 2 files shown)


robob27:
Adding changelog

library/modules/Gui.cpp
```diff
@@ -380,8 +380,14 @@ DEFINE_GET_FOCUS_STRING_HANDLER(dwarfmode)
         case df::view_sheet_type::BUILDING:
             if (game->main_interface.view_sheets.linking_lever)
                 newFocusString = baseFocus + "/LinkingLever";
-            else if (auto bld = df::building::find(game->main_interface.view_sheets.viewing_bldid))
-                newFocusString += '/' + enum_item_key(bld->getType());
+            else if (auto bld = df::building::find(game->main_interface.view_sheets.viewing_bldid)) {
+                std::string buildingType = enum_item_key(bld->getType());
+                newFocusString += '/' + buildingType;
+                if (buildingType == "Trap") {
```
myk002, line 386: It's more efficient to match the integer type rather than do a string compare. This function is called many times per frame and efficiency is important here. Otherwise, looks great!

robob27: Fixed!

myk002 (approved):

PR #3975, opened 2023-11-07
Remove old trackstop

robob27:
Remove the old trackstop plugin, docs and other references


robob27:
For historical purposes, if anyone comes across this while looking for trackstop, it has been replaced by https://github.com/DFHack/scripts/pull/884

docs/about/History.rst
myk002, line 1: changes to this file can be reverted -- it's ok if the links being to point to the new `trackstop` tool

robob27: Reverted

PR #3977, opened 2023-11-07
Add search to Zones page

ToxicBananaParty:
Let me know if there's a better way to get_default_zone_name -- this is so that you can still search "Garbage dump" or "Bedroom" if you haven't set an actual name for it


plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+    return table.concat(result, ' ')
+end
+
+-- ----------------------
+-- PlacesOverlay
+--
+
+PlacesOverlay = defclass(PlacesOverlay, sortoverlay.SortOverlay)
+PlacesOverlay.ATTRS{
+    default_pos={x=71, y=8},
```
ToxicBananaParty, line 71: I feel like y=8 is too high but y=9 is too low.
I went with y=8 because it at least lined up with the edges of some of the other UI elements.

ToxicBananaParty:
pre-commit.ci autofix

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+
+    return '' -- If zone_type is anything else it will not have a default name populated in the Zones page when left nameless
+end
+
+local function get_zone_search_key(zone)
+    local site = df.global.world.world_data.active_site[0]
+    local result = {}
+
+    -- allow zones to be searchable by their name
+    if zone.name == nil or zone.name == '' then
```
myk002, line 37: it's more idiomatic to treat values as "truthy" rather than comparing to `nil`. Also comparing the string length to 0 is preferable to comparing against the empty string. this would normally be written as:
```lua
if not zone.name or #zone.name == 0 then
```

however, in this case, `zone.name` is a value field, not a pointer, so you don't have to check that for truthiness. just compare the length to 0

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+        table.insert(result, get_default_zone_name(zone.type))
+    else
+        table.insert(result, zone.name)
+    end
+
+    -- allow zones w/ assignments to be searchable by their assigned unit
+    if zone.assigned_unit ~= nil then
+        table.insert(result, dfhack.TranslateName(zone.assigned_unit.name))
+        table.insert(result, dfhack.units.getReadableName(zone.assigned_unit))
```
myk002, line 46: You only need the last name (in English) from TranslateName since everything else is provided by `getReadableName`

However, it would be even better to call `sortoverlay.get_unit_search_key(zone.assigned_unit)` instead (unless you *don't* want the unit profession in the search key).

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    -- allow zones to be searchable by type
+    if zone.location_id == -1 then -- zone is NOT a special location and we don't need to do anything special for type searching
+        table.insert(result, df.civzone_type[zone.type]);
+    else -- zone is a special location and we need to get its type from world data
+        table.insert(result, dfhack.TranslateName(site.buildings[zone.location_id].name, true))
```
myk002, line 53: use `utils.binsearch` to get the element out of the array. `zone.location_id` is the *id*, not necessarily the *index* (though it often is).

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+        table.insert(result, df.civzone_type[site.buildings[zone.location_id].name.type])
```
myk002, line 54: site.buildings is a vector of `df.abstract_building` subclasses, not all of which have a `name` field

moreover, `name.type` is a `df.language_name_type`, not a `df.civzone_type`, so it's not going to map correctly

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+        self:do_search(self.subviews.search.text)
+    end
+end
+
+function PlacesOverlay:onInput(keys)
+    if keys._MOUSE_L then
+        self.refresh_search = true
+    end
+    return PlacesOverlay.super.onInput(self, keys)
+end
```
myk002, line 119: is this necessary on these tabs? The WORK_DETAILS screen needs this because clicking on any of the buttons refreshes the underlying vectors, which then need to be re-filtered. I don't think that situation applies to the Places subtabs (does it?)

plugins/lua/sort/places.lua
myk002: finding a spot can be tricky. could you post a screenshot of what it looks like now?

Also, be sure the position works (or is adjusted) when the screen is resized narrower and the top row of tabs becomes two lines

plugins/lua/sort/places.lua
ToxicBananaParty: I copied this from UnitSelectorOverlay -- but no, the vector's don't need to be reloaded when any buttons are clicked. Will remove and make sure everything still works.

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,121 @@
...
+    if zone_type == df.civzone_type.PlantGathering then return 'Gather fruit' end
+    if zone_type == df.civzone_type.Dungeon then return 'Dungeon' end
+    if zone_type == df.civzone_type.MeetingHall then return 'Meeting area' end
+    if zone_type == df.civzone_type.Barracks then return 'Barracks' end
+    if zone_type == df.civzone_type.ArcheryRange then return 'Archery range' end
+    if zone_type == df.civzone_type.Office then return 'Office' end
+    if zone_type == df.civzone_type.DiningHall then return 'Dining hall' end
+    if zone_type == df.civzone_type.Dormitory then return 'Dormitory' end
+    if zone_type == df.civzone_type.Bedroom then return 'Bedroom' end
+    if zone_type == df.civzone_type.Tomb then return 'Tomb' end
```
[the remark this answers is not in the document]
ToxicBananaParty: Can do. What does enclosing the enums in square brackets on the left side of the = accomplish?

plugins/lua/sort/places.lua
ToxicBananaParty: Here is y=8:
image

And yeah it looks fine when compressed:
image

Here is y=9 for reference:
image

Now that I'm looking at them side by side maybe y=9 is actually better since it's less likely to have misclicks onto the top bar? As much as it's bothersome that it doesn't line up.

ToxicBananaParty:
pre-commit.ci autofix

plugins/lua/sort/places.lua
myk002: agreed. I think I chose a y=9 style for other searches as well

plugins/lua/sort/places.lua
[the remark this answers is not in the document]
myk002: that's just the Lua syntax for how to declare a key inline that isn't a string literal. docs here: https://www.lua.org/manual/5.3/manual.html#3.4.9

plugins/lua/sort/places.lua
```diff
@@ -0,0 +1,113 @@
...
+        table.insert(result, sortoverlay.get_unit_search_key(zone.assigned_unit))
+    end
+
+    -- allow zones to be searchable by type
+    if zone.location_id == -1 then -- zone is NOT a special location and we don't need to do anything special for type searching
+        table.insert(result, df.civzone_type[zone.type]);
+    else -- zone is a special location and we need to get its type from world data
+        local building, success, _ = utils.binsearch(site.buildings, zone.location_id, 'id')
+        if success and building.name then
+            table.insert(result, df.language_name_type[building.name.type])
```
myk002, line 57: I think you're going to need a different way to map the location type to a searchable string. Nobody is going to search for `SymbolFood` when looking for a Tavern, for example.

ToxicBananaParty: My first thought would be to create table matching arbitrary `df.language_name_type`s to `df.civzone_type` similar to how I created the `zone_type` table earlier.

Is there a more idiomatic or API-using solution you would recommend?

plugins/lua/sort/places.lua
```diff
@@ -28,10 +28,22 @@ local zone_names = {
...
+    [df.language_name_type.SymbolDomestic] = 'MarketStall'
+}
+
 local function get_default_zone_name(zone_type)
     return zone_names[zone_type] or ''
 end
 
+local function get_location_type_from_language_name(language_name_type)
+    -- Looking at df.language.xml I think SymbolFood and SymbolDomestic are the only two we need to override?
+    -- Every other df.language_name_type for a location matches the displayed type name (e.g. Temples are language_name_type.Temple, etc)
```
myk002, line 43: The ones that we care about are `Guildhall`, `Tavern`, `Temple`, and `Hospital`. The language name type seems to match up for everything except taverns, so you can create a static mapping for everything or just special case the taverns. Where does SymbolDomestic come up? I don't think I've seen that one in use.

ToxicBananaParty: SymbolDomestic is for MarketStall -- I added it because both it and SymbolFood are written in df.language.xml as
`<enum-item name='SymbolDomestic' comment='MarketStall'>`

Looking through the wiki I don't think that's actually used anywhere (maybe Adventure Mode?) so I'll update it.
Hopefully final commit incoming ASAP! :)

plugins/lua/sort/places.lua
```diff
@@ -29,21 +29,17 @@ local zone_names = {
 }
 
 -- I used strings rather than df.civzone_type because nobody is going to search "MeadHall" they're going to search "Tavern"
-local language_name_type_strings = {
+local language_name_types = {
     [df.language_name_type.SymbolFood] = 'Inn/Tavern',
```
myk002, line 33: This is going in the search key so it should be space separated

myk002:
Thanks for this! The remaining Places tabs should be much easier to cover.

I forgot to mention earlier, in a separate PR, please feel free to add your name to [docs/about/Authors.rst](https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst)!

PR #3984, opened 2023-11-08
Add search overlay to Locations page

ToxicBananaParty:


plugins/lua/sort/places.lua
```diff
@@ -33,9 +33,45 @@ local language_name_types = {
...
+
+local function stringify_spheres(spheres)
+    local strs = {}
+    for sphere in pairs(spheres) do
+        table.insert(strs, df.sphere_type[sphere])
+    end
+    return table.concat(strs, ' ')
+end
+
+local function get_religion_string(religion_id, religion_type)
```
ToxicBananaParty, line 55: I would've just un-localized the existing `get_religion_string` but want to return slightly different things (namely 'Temple' when `religion_type` is `None (-1)`). Let me know if there's a better way to reuse this code.

ToxicBananaParty:
pre-commit.ci autofix

plugins/lua/sort/places.lua
```diff
@@ -33,9 +34,15 @@ local language_name_types = {
     [df.language_name_type.SymbolFood] = 'Inn Tavern',
     [df.language_name_type.Temple] = 'Temple',
     [df.language_name_type.Hospital] = 'Hospital',
-    [df.language_name_type.Guildhall] = 'Guildhall'
+    [df.language_name_type.Guildhall] = 'Guildhall',
+    [df.language_name_type.Library] = 'Library'
```
myk002, line 38: you can end this line with a comma too to reduce the diff footprint of future edits

plugins/lua/sort/places.lua
```diff
@@ -33,9 +34,15 @@ local language_name_types = {
...
 }
 
+local function get_location_religion(religion_id, religion_type)
+    if religion_type == -1 then return 'Temple'
```
myk002, line 42:
```suggestion
    if religion_type == df.temple_deity_type.None then return 'Temple'
```

myk002: could you possibly fix this in locationselector.lua in `get_religion_string` as well? `0` should be `df.temple_deity_type.Deity` and `1` should be `df.temple_deity_type.Religion`

plugins/lua/sort/places.lua
```diff
@@ -77,6 +85,31 @@ local function get_zone_search_key(zone)
...
+        table.insert(result, language_name_types[building.name.type] or '')
+        if building.name.has_name then
+            table.insert(result, dfhack.TranslateName(building.name, true))
+        end
+
+        -- for temples and guildhalls, get assigned organization
+        if df.abstract_building_templest:is_instance(building) then
+            table.insert(result, get_location_religion(building.deity_data.Deity or building.deity_data.Religion, building.deity_type))
+        elseif df.abstract_building_guildhallst:is_instance(building) then
+            local dwarfified_profession = locationselector.get_profession_string(building.contents.profession):gsub('[Mm][Aa][Nn]', 'dwarf') -- Craftsman becomes Craftsdwarf, etc
```
myk002, line 104: "dwarf"ifying is a good idea, but I'd suggest adding instead of replacing (in case people do actually search for "craftsman"). so the string would become `CRAFTSMAN CRAFTSdwarf` (capitalization doesn't matter). I'd also suggest moving this logic to `locationselector.get_profession_string` so the location selector screens can benefit from it too.

plugins/lua/sort/places.lua
```diff
@@ -77,6 +85,31 @@ local function get_zone_search_key(zone)
     return table.concat(result, ' ')
 end
 
+local function get_location_search_key(zone)
+    local site = df.global.world.world_data.active_site[0]
+    local result = {}
+
+    -- get language_name and type (we dont need user-given zone name because it does not appear on this page)
+    local building, success, _ = utils.binsearch(site.buildings, zone.location_id, 'id')
```
myk002, line 93: you can leave out the final `, _` since Lua doesn't care if you absorb all the return values from a function. the extras are just dropped.

plugins/lua/sort/locationselector.lua
```diff
@@ -50,8 +50,10 @@ local function get_religion_string(religion_id, religion_type)
     return ('%s %s'):format(dfhack.TranslateName(entity.name, true), stringify_spheres(spheres))
 end
 
-local function get_profession_string(profession)
-    return df.profession[profession]:gsub('_', ' ')
+function get_profession_string(profession)
+    local profession_string = df.profession[profession]:gsub('_', ' ')
+    local dwarfified_string = profession_string:gsub('[Mm][Aa][Nn]', 'dwarf')
+    return profession_string .. " " .. dwarfified_string
```
myk002, line 56: `" "` -> `' '` for consistency

myk002 (approved):

myk002 (approved):

Issue #16, opened 2010-03-12
Make a tool that integrates LUA


peterix:
Maybe :)

PR #4019, opened 2023-11-14
Add search to remaining Places pages

ToxicBananaParty:


myk002 (commented):
make sure the Places `sort` additions are represented in `docs/changelog.txt`!

plugins/lua/sort/places.lua
```diff
@@ -111,6 +111,28 @@ local function get_location_search_key(zone)
     return table.concat(result, ' ')
 end
 
+local function get_stockpile_search_key(stockpile)
+    -- I put the stockpile_number in twice so you can search both with and without the octothorpe
+    -- (e.g. if your unnamed stockpile is listed as "Stockpile #16" you can search both "#16" or "16")
```
myk002, line 116: is this necessary? I would have thought that a search key of `Stockpile #16` would get matched by search text of either `16` or `#16`

ToxicBananaParty:
Will do. I also need to add searching by assigned unit to workshops first.

plugins/lua/sort/places.lua
```diff
@@ -111,6 +111,28 @@ local function get_location_search_key(zone)
...
+end
+
+local function get_workshop_search_key(workshop)
+    return ('%s %s %s'):format(workshop.name, df.workshop_type.attrs[workshop.type].name or '', df.workshop_type[workshop.type])
+end
+
+local function get_farmplot_search_key(farmplot)
+    local result = {}
+
+    if #farmplot.name ~= 0 then table.insert(result, farmplot.name) else table.insert(result, 'Farm Plot') end
```
myk002, line 128: I'd think you'd want the word "farm" in the key regardless of whether the farmplot has a specific name, no?

ToxicBananaParty: No, imo. This represents the default name for farms being "Farm Plot" -- there's no point in searching by 'farm' in a Farmplots page, you already know everything in that page is a farm.

plugins/lua/sort/places.lua
ToxicBananaParty: You're right -- I must have forgotten to make an unnamed stockpile first before testing that the first time, or something.

plugins/lua/sort/places.lua
myk002: oh, of course. in my mind I was picturing the workshops page. current code looks good then

plugins/lua/sort/places.lua
```diff
@@ -112,14 +112,22 @@ local function get_location_search_key(zone)
...
-    else return ('Stockpile #%s %s'):format(stockpile.stockpile_number, stockpile.stockpile_number) end
+    else return ('Stockpile #%s'):format(stockpile.stockpile_number) end
 end
 
 local function get_workshop_search_key(workshop)
-    return ('%s %s %s'):format(workshop.name, df.workshop_type.attrs[workshop.type].name or '', df.workshop_type[workshop.type])
+    local result = {}
+    for _, unit_id in ipairs(workshop.profile.permitted_workers) do
+        local unit = df.unit.find(unit_id)
+        if unit then table.insert(result, sortoverlay.get_unit_search_key(unit)) end
```
myk002, line 123: note that this will include the units' professions, which may or may not be desirable here

myk002: meh, it's probably ok. what if you want to see where you assigned your stoneworkers?

ToxicBananaParty: It lists their profession on the page so I figure that's good, in the interest of "can search any text that appears on the page"

myk002 (approved):

PR #4041, opened 2023-11-19
RemoteServer: reset fail counter after success

### LINKED ISSUE #4040: RemoteServer shutdowns after the fifth connection.
"Too many failed accepts, shutting down RemoteServer" after connecting 5 times. Although the fifth connection is still working after that, no new connections will be accepted.

09148698c808d83d9952aff7824b3baf6f52bc3f

In "void ServerMainImpl::threadFn(std::promise<bool> promise, int port)", acceptFail counts all accept calls, not only failed ones.

Even fixing the counter, I'm not sure what is the logic here. Why completely stop the server after five errors?

ab9rf:
should close #4040

### DIFF
--- a/library/RemoteServer.cpp
+++ b/library/RemoteServer.cpp
@@ -80,7 +80,7 @@ namespace {
     struct BlockedException : std::exception {
         const char* what() const noexcept override
         {
-            return "Core has blocked all connection. This should have been catched.";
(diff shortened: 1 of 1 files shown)


myk002 (commented):
Changelog? Otherwise lgtm

myk002 (approved):

PR #4042, opened 2023-11-19
Improve Accept error handling in RemoteServer

### LINKED ISSUE #4040: RemoteServer shutdowns after the fifth connection.

cvuchener:
Also makes sure the socket is blocking to avoid EWOULDBLOCK errors.

Fix #4040, not sure if it fixes #3498 though. I'm still not sure what was the bug.

### DIFF
--- a/library/RemoteServer.cpp
+++ b/library/RemoteServer.cpp
@@ -475,35 +475,28 @@ void ServerMainImpl::threadFn(std::promise<bool> promise, int port)
 {
     ServerMainImpl server{std::move(promise), port};
 
-    CActiveSocket *client = nullptr;
-
+    server.socket.SetBlocking();
     try {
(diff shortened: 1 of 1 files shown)


cvuchener:
Maybe I should always log errors. EINTR should be filtered by clsocket and if there is still a EWOULDBLOCK bug, it will be visible this way.

myk002 (approved):
merging for the obvious improvements. further logging might be useful too, though.

PR #4050, opened 2023-11-21
rename `report_init` to `announcementst` (coordinated change)

ab9rf:
this is the coordinated change to dfhack/df-structures#678


myk002:
could remove or update the ref to df-structures?

myk002:
you probably have to merge develop before you can update the structures ref

PR #4090, opened 2023-12-10
add identity for variant

myk002:
I plan to merge this expediently to unbreak the build, but post-merge review would be welcome

PR #4092, opened 2023-12-11
data identity changes for stl-variant in 50.12

ab9rf:
this is arguably a better approach for the change in #4090

### DIFF
--- a/library/DataIdentity.cpp
+++ b/library/DataIdentity.cpp
@@ -5,6 +5,7 @@
 #include <mutex>
 #include <string>
 #include <vector>
+#include <variant>
 
 #include "DataFuncs.h"
 #include "DataIdentity.h"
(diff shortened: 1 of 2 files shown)


lethosor (approved):
So this is basically letting the preprocessor split arguments by commas, then joining them with `__VA_ARGS__`.

I agree it looks cleaner. Haven't tested.

ab9rf:
> So this is basically letting the preprocessor split arguments by commas, then joining them with `__VA_ARGS__`.

actually think that `__VA_ARGS__` bypasses splitting entirely, but that's a compiler implementation detail that frankly i don't care about here. `__VA_ARGS__` and the `...` meta-argument in macros have been in the standard since C++11, and all of the compilers we care about implement it in the standard way

PR #4100, opened 2023-12-18
[refactoring] Move melting logic from `logistics` to `Items` module and export to Lua

chdoc:
Possible discussion point:: the logic in `logistics` refuses to melt metal bars, even though the game displays the melt icon next to them.

Required to finish https://github.com/DFHack/scripts/pull/896 (see discussion there for details)


lethosor:
Yes, that is not configurable: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review:
> To assign a reviewer to a pull request, you will need write access to the repository.

However, usually we can get to PRs within 24 hours regardless of whether they have an explicitly requested review.

chdoc:
Well I wanted to check with @myk002, because he's the only one has touched `logistics.cpp` before. Originally, he had promised to do this refactoring, but it appears that he has a lot of work on his plate these days.

library/modules/Items.cpp
```diff
@@ -2270,6 +2270,74 @@ bool Items::markForTrade(df::item *item, df::building_tradedepotst *depot) {
...
+        case df::item_type::FISH_RAW:
+        case df::item_type::VERMIN:
+        case df::item_type::PET:
+        case df::item_type::FOOD:
+        case df::item_type::EGG:
+        case df::item_type::BAR:
+            return false;
+        default:
+            break;
+    }
```
lethosor, line 2292: Out of curiosity, where did this section come from? Looks new.

lethosor (commented):
In general I think this looks like a pretty straightforward refactor. No objections from me.

library/modules/Items.cpp
chdoc: This was posted by @ab9rf [on Discord](https://discord.com/channels/793331351645323264/807443277132595211/1173979579409506367). My understanding is that this is the result of reverse engineering. 

That is, except the `BAR` condition, which the game seems to allow but which seems somewhat pointless and was ruled out by the check in `logistics`.

ab9rf: suggest replacing the body of the switch with
```
    using namespace df::enums::item_type;
    case CORPSE:
...
```

imo, the scoped `using namespace` statement makes the switch easier to read

iirc, this is from the reverse-engineering of `itemst::meltable()`, which is non-virtual and thus not available for us to use directly

library/modules/Items.cpp
```diff
@@ -2270,6 +2270,74 @@ bool Items::markForTrade(df::item *item, df::building_tradedepotst *depot) {
     return true;
 }
 
+bool Items::canMelt(df::item *item)
```
ab9rf, line 2273: please note in comment: "this is equivalent to bay12's `itemst::meltable()`

library/modules/Items.cpp
```diff
@@ -2270,6 +2270,74 @@ bool Items::markForTrade(df::item *item, df::building_tradedepotst *depot) {
...
+        }
+    }
+
+    if (item->flags.bits.artifact)
+        return false;
+
+    return true;
+};
+
+bool Items::markForMelting(df::item *item)
```
ab9rf, line 2325: this function should guard against designating an item for melt that is already designated for melt, which could result in the item being the vector twice, which is likely bad

docs/dev/Lua API.rst
```diff
@@ -1834,6 +1834,18 @@ Items module
 
   Marks the given item for trade at the given depot.
 
+* ``dfhack.items.canMelt(item)``
+
+  Returns true if the item can be designated for melting.
+
+* ``dfhack.items.markForMelting(item)``
+
+  Marks the given item for melting.
```
myk002, line 1843: perhaps add something to the effect of: "if it's not already marked for melting"

docs/dev/Lua API.rst
```diff
@@ -1834,6 +1834,18 @@ Items module
...
+
+* ``dfhack.items.cancelMelting(item)``
+
+  Removes an existing melting designation from the given item.
```
myk002, line 1847: "...if it is marked for melting"

library/modules/Items.cpp
myk002: I believe `insert_into_vector` is a noop if the item already exists, so it should be safe. The not-at-all-similarly-named-and-absolutely-not-confusing `vector_insert_at` method has no such protections and would insert blindly.

library/modules/Items.cpp
chdoc: Indeed, there is a global `using namespace df::enums;`, and the predominant form is `item_type::CORPSE`, so I would follow that.

library/modules/Items.cpp
```diff
@@ -2270,6 +2270,74 @@ bool Items::markForTrade(df::item *item, df::building_tradedepotst *depot) {
...
+                for (auto &gg : c->general_refs) {
+                    if (gg->getType() == df::general_ref_type::UNIT_HOLDER)
+                        return false;
+                }
+                break;
+            }
+            default:
+                break;
+        }
+    }
```
chdoc, line 2317: While refactoring, I didn't question this logic too much. But it clearly does not correspond to what DF does.
1. DF allows designating equipped items for melting (at least it displays a melting icon)
2. DF allows melting non-empty metal bins (I tested with bins containing metal bars, the bars end up in the workshop)

myk002: I contend that while this does not accurately mirror DF logic, it is appropriate for how this function is used, which is more along the lines of "should I consider this item as a candidate for melting". I think players would be confused and annoyed if they select all for melting and get their full bins melted.

chdoc: @myk002 Unless you (or someone else) objects, I would move this check back to `logistics`, just like the `melt_masterworks` check.

chdoc: Sorry, I wrote my last message without reloading the page. How about taking a boolean flag (that defaults to true) to toggle those checks that do not strictly mirror the logic from DF? At least this would clearly document the deviations.

myk002: That would be acceptable, though make sure the flag defaults to *false*, otherwise you'll need to write a custom shim function to insert the default value when you expose to Lua.

chdoc:
Anyway, this will have to wait until after Christmas. Happy Holidays :christmas_tree:

library/modules/Items.cpp
lethosor: Right. `insert_into_vector` checks if the object is present, then calls `vector_insert_at`. https://github.com/DFHack/dfhack/blob/bbc79d70c04a993d84da591e8cd0bf18bab15075/library/include/MiscUtils.h#L215-L235

library/modules/Items.cpp
```diff
@@ -2270,6 +2270,83 @@ bool Items::markForTrade(df::item *item, df::building_tradedepotst *depot) {
...
+
+    return true;
+};
+
+bool Items::markForMelting(df::item *item)
+{
+    CHECK_NULL_POINTER(item);
+    if (item->flags.bits.melt) return false;
+    insert_into_vector(world->items.other.ANY_MELT_DESIGNATED, &df::item::id, item);
```
myk002, line 2336: we should either sanity check the input with a call to canMelt(item, true) or document that it is the caller's responsibility to ensure the item passed to this function is meltable

chdoc: I briefly thought about calling `canMelt` and then decided not to. My rationale was that the check is not exactly free, in particular when including the optional checks for objects one might "reasonably want to" melt. 

Calling `canMelt(item,true)` would be sufficient to preserve a consistent gamestate and doesn't crawl over various general refs. Either one is fine with me, make your pick. I strongly doubt this is ever going to be called on tens of thousands of items at once.

myk002: since we don't have a performance problem here, I'd say let's err on the side of safety and call `canMelt(item,true)`

chdoc:
Evidently, this needs a rebase. Is there any preference/policy regarding the preservation of individual commits versus squashing the PR prior to merging.

chdoc:
Sorry for the delay. As anticipated, early January is not a time to do much gaming or hobby coding. I ended up just squashing everything; I don't thing anyone will care about the intermediate stages down the road.

myk002 (approved):

PR #2516, opened 2022-12-21
WIP (pending testing): Remove tinythread as a dependency

paxman101:
Closes https://github.com/DFHack/dfhack/issues/2514

As tinythread was meant to implement the C++11 concurrency library very closely, moving from tinythread to C++11 threads was mostly trivial replacement of ```tthread``` scoped types with ```std``` ones. ```condition_variable```'s ```wait()``` in the std library uses a ```unique_lock``` instead of a mutex pointer so some instances required adding a ```unique_lock```. However, the switch over to using ```unique_lock```s doesn't change anything as a mutex has to be held by the calling thread to a condvar wait anyways.

Some of the changes are in currently unbuilt plugins, so it may be pertinent to test those changes on the v47.05r8 build of DFHack.

PR #4106, opened 2023-12-26
Really excise tinythread

dhthwy:
Takeover #2516

~~There is a deadlock when quitting df. I haven't looked into it much yet.~~

@ab9rf will you test this on Windows when you get a chance? ~~And I'm curious if the hang is reproducible there.~~


dhthwy:
Looks like it's a case of double locking the mutex in PluginManager's Plugin::RefLock wait() line 106. So yeah, should be able to reproduce.

plugins/rendermax/renderer_light.cpp
```diff
@@ -108,7 +105,7 @@ lightingEngineViewscreen::lightingEngineViewscreen(renderer_light* target):light
 {
     reinit();
     defaultSettings();
-    int numTreads=tthread::thread::hardware_concurrency();
+    int numTreads=std::thread::hardware_concurrency();
```
myk002, line 108: you might as well fix the typo while you're here

myk002:
tests show what appears to be a deadlock on Windows

library/PlugLoad-windows.cpp
```diff
@@ -34,7 +34,7 @@ distribution.
 #include "Hooks.h"
 #include <stdio.h>
 
-#include "tinythread.h"
+#include "../plugins/uicommon.h"
```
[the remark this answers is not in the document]
dhthwy: I'll have to look into that. Wasn't me that put that in there, at least on purpose. Could've been a rebase screwup.

dhthwy:
That sucks. It seems to run ok on Linux, including the tests.

library/PluginManager.cpp
```diff
@@ -1019,7 +1012,7 @@ void PluginManager::OnStateChange(color_ostream &out, state_change_event event)
 
 void PluginManager::registerCommands( Plugin * p )
 {
-    cmdlist_mutex->lock();
+    lock_guard<std::mutex> lock{*cmdlist_mutex};
```
quietust, line 1015: Need to remove `cmdlist_mutex->unlock();` at the end of the function, otherwise it double-unlocks (and causes a deadlock later).

library/PluginManager.cpp
```diff
@@ -1037,7 +1030,7 @@ void PluginManager::registerCommands( Plugin * p )
 
 void PluginManager::unregisterCommands( Plugin * p )
 {
-    cmdlist_mutex->lock();
+    lock_guard<std::mutex> lock{*cmdlist_mutex};
```
quietust, line 1033: See above.

plugins/rendermax/renderer_light.hpp
```diff
@@ -263,7 +265,7 @@ class lightThread
     void work(); //main light calculation function
     void combine(); //combine existing canvas into global lightmap
 public:
-    tthread::thread *myThread;
+    std::thread *myThread;
     bool isDone; //no mutex, because bool is atomic
```
ab9rf, line 269: i know it's "widely believed" that "bool is always atomic" but the standard makes no such guarantee

see also https://stackoverflow.com/questions/16320838/when-do-i-really-need-to-use-atomicbool-instead-of-bool

dhthwy: I don't think a mutex is needed for this? isDone can be an std::atomic<bool> from what I see.

I don't even know what rendermax even does lol.

dhthwy: I agree with you. It should be an atomic if the value is shared between threads.

myk002 (commented):
You'll have to test the changes to rendermax in 0.47.05. it hasn't been updated to work with v50 yet. I don't think it's even being built atm

myk002:
FWIW, I'd be happy if you just added a comment to rendermax that improvements could be made and that the threading needs verification and leave it at that. I wouldn't want it holding this PR up indefinitely.

dhthwy:
Whichever is fine. A comment is a good start at any rate.

dhthwy:
I did change the isDone bool to an std::atomic. I'm confident that change won't do any harm except for make it run a tiny bit slower. But that said, I can revert that commit if you'd rather keep it separate and worry about it later.

myk002 (approved):

PR #4128, opened 2024-01-02
Update Maps::getBiomeType and Maps::getBiomeTypeWithRef

Bumber64:
I've updated the functions `getBiomeType` and `getBiomeTypeWithRef` based on reverse-engineering of v50.11.

The only real changes seem to be:
`potential_tropical` (rather than `tropical`) is sufficient for `OCEAN_TROPICAL`
`region.vegetation` isn't used at all, instead replaced by `region.rainfall` checks

`getBiomeType` and `getBiomeTypeWithRef` now do bounds checking on the region coords by calling `getRegionBiome`, then doing a `CHECK_NULL_POINTER` on the region. I'd prefer to return `biome_type::NONE` here instead of throwing an exception, but that doesn't exist so we'd have to add that to structures if appropriate.

I indented the `DFHack` and `Maps` namespaces in `Maps.h`, and inlined `getBiomeType` there.


Bumber64:
Anybody know why `getBiomeTypeWithRef` exists in the first place? Embark assistant uses it to check a region biome using an adjacent tile's tropicality, but I'm not sure of the reason for doing that.

ab9rf:
i'm trying to understand the motivation behind this. is the current implementation incorrect in some way?

for the history, see #1392 and #1394

Bumber64:
> i'm trying to understand the motivation behind this. is the current implementation incorrect in some way?

For one, the old code was reporting a regional ocean tile as `OCEAN_TEMPERATE` while the new code says it's `OCEAN_TROPICAL`. (I used a lua script iterating all region tiles to determine this, but found it too much a hassle to verify what the game says even in a pocket world, since there's no keyboard support for embark screen and I'd basically have to use trial and error to embark on the right tile. Just going to assume the current disassembled function is correct, otherwise there's a Bay12 bug.)

myk002:
do you have a map where I can see the changes in interpretation? I tried to verify on the maps I have, but I'm not sure how to specifically get a biome with `potential_tropical` set.

Bumber64:
> do you have a map where I can see the changes in interpretation? I tried to verify on the maps I have, but I'm not sure how to specifically get a biome with `potential_tropical` set.

Using this branch: https://github.com/Bumber64/dfhack/tree/test-biomes
Generate a `Pocket Region` with `North` pole and seed `test`.
Use this Lua script:
```
function test_biomes()
  local wd = df.global.world.world_data
  for x=0, wd.world_width-1 do
    for y=0, wd.world_height-1 do
      local b_new = df.biome_type[dfhack.maps.getBiomeType(x,y)]
      local b_old = df.biome_type[dfhack.maps.getBiomeTypeOld(x,y)]
      if b_new ~= b_old then
        print(("(%d, %d): %s -> %s"):format(x, y, b_new, b_old))
      end
    end
  end
  print("Done!")
end
```
Output should be:
```
(14, 11): OCEAN_TROPICAL -> OCEAN_TEMPERATE
(15, 11): OCEAN_TROPICAL -> OCEAN_TEMPERATE
(15, 12): OCEAN_TROPICAL -> OCEAN_TEMPERATE
(16, 11): OCEAN_TROPICAL -> OCEAN_TEMPERATE
(16, 12): OCEAN_TROPICAL -> OCEAN_TEMPERATE
Done!
```

myk002 (approved):
nice. thanks!

PR #4130, opened 2024-01-03
Implement Units::getUnitFromHFID

Bumber64:
Added a function for retrieving a unit from a historical figure id. It's `FUN_140fe1570` in v50.11 win64 Steam (though the function is simple enough that it hardly bears mentioning.)

I indented the `DFHack` and `Units` namespaces in `Units.h`. I also deleted a stray comment in `Units.cpp` that didn't represent any nearby function.

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -790,8 +790,6 @@ bool Units::isGreatDanger(df::unit* unit)
     return isDemon(unit) || isTitan(unit) || isMegabeast(unit);
 }
 
-
-
 int32_t Units::getNumUnits()
 {
     return world->units.all.size();
@@ -802,7 +800,15 @@ df::unit *Units::getUnit (const int32_t index)
     return vector_get(world->units.all, index);
 }
 
-// returns index of creature actually read or -1 if no creature can be found
+df::unit *Units::getUnitFromHFID(int32_t hf_id)
+{
+    auto hf = df::historical_figure::find(hf_id);
+    if (hf)
+        return df::unit::find(hf->unit_id);
+    else
+        return vector_get(world->units.all, linear_index(world->units.all, &df::unit::hist_figure_id, hf_id));
+}
+
 bool Units::getUnitsInBox (std::vector<df::unit*> &units,
     int16_t x1, int16_t y1, int16_t z1,
     int16_t x2, int16_t y2, int16_t z2)
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -71,8 +71,10 @@ Template for new versions:
 - UTF-8 text in tool docs is now properly displayed in-game in `gui/launcher` (assuming that it can be converted to cp-437)
 
 ## API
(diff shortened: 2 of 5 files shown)


library/modules/Units.cpp
```diff
@@ -802,7 +800,15 @@ df::unit *Units::getUnit (const int32_t index)
...
 }
 
-// returns index of creature actually read or -1 if no creature can be found
+df::unit *Units::getUnitFromHFID(int32_t hf_id)
+{
+    auto hf = df::historical_figure::find(hf_id);
+    if (hf)
+        return df::unit::find(hf->unit_id);
+    else
+        return vector_get(world->units.all, linear_index(world->units.all, &df::unit::hist_figure_id, hf_id));
```
myk002, line 809: in which situations will the else branch produce a useful result?

Bumber64: If the unit is missing a hf entry, the unit can still be found by its hf id. IDK why the entry wouldn't exist, but the Bay12 code accounts for it.

myk002 (commented):
similar question as the previous review: do you plan on going through the code (both cpp and lua) and refactoring to use these new APIs?

Bumber64:
> similar question as the previous review: do you plan on going through the code (both cpp and lua) and refactoring to use these new APIs?

I'll have to check if its actually used anywhere. I can just make it a utility fn when I need it, otherwise. Is it convenient enough that somebody might want to use it in Lua?

myk002:
If we want to make this our official method of getting a unit from a HF, then it should be consistent. I've never seen anything do anything other than a simple `df::unit::find(hf->unit_id)`, though, in both cpp and Lua, and I'm not aware of any problems we've run into by doing that, so I'm not exactly enthusiastic about this change. Is there really any benefit to converting the code to use this?

Bumber64:
It's actually more niche than finding a unit from its own HF, since you'd already know the unit has one if you're using it. You're likely getting the hf_id from some kind of hf link. Using the function means you don't want to use the unit's HF for anything else, either.

Here are all the places I found that fit those criteria:
https://github.com/DFHack/dfhack/blob/1968cffb13df57850e51fa4bf0f8b32a958449bd/library/modules/Units.cpp#L837
https://github.com/DFHack/dfhack/blob/1968cffb13df57850e51fa4bf0f8b32a958449bd/plugins/lua/sort/diplomacy.lua#L74
https://github.com/DFHack/dfhack/blob/1968cffb13df57850e51fa4bf0f8b32a958449bd/plugins/lua/sort/slab.lua#L90
https://github.com/DFHack/dfhack/blob/1968cffb13df57850e51fa4bf0f8b32a958449bd/plugins/lua/sort/world.lua#L34
https://github.com/DFHack/scripts/blob/f9175740928fa885915226a75b0b9a5ebd1ba400/gui/choose-weapons.lua#L102

While searching, I also stumbled across some code that's missing a NULL check:
https://github.com/DFHack/dfhack/blob/1968cffb13df57850e51fa4bf0f8b32a958449bd/plugins/autolabor/labormanager.cpp#L924

PR #4140, opened 2024-01-05
Update MiscUtils CP437 functions

Bumber64:
Addendum to PR #4137

* Add `toupper_cp437`, `tolower_cp437` for chars
* Use `toupper_cp437` in `capitalize_string_words`
* `toUpper`, `toLower` string fns become `toUpper_cp437`, `toLower_cp437`; add these versions to Lua API
* Update changelog and Lua API docs


library/include/MiscUtils.h
```diff
@@ -417,8 +417,10 @@ inline std::string join_strings(const std::string &separator, T &items) {
     return ss.str();
 }
 
-DFHACK_EXPORT std::string toUpper(const std::string &str);
-DFHACK_EXPORT std::string toLower(const std::string &str);
+DFHACK_EXPORT char cp437_toupper(char c);
+DFHACK_EXPORT char cp437_tolower(char c);
+DFHACK_EXPORT std::string toUpper(const std::string &str, bool use_cp437 = false);
```
lethosor, line 422: I think ideally this should be some other "encoding" parameter, not a boolean. I'll admit that makes the default arguments in the Lua API wrapper harder to implement, though.

Bumber64: CP437 is done with a switch statement. What other encodings do we want, and how would they be implemented?

Bumber64: You want a separate copy of the function just to avoid a bool?

Bumber64:
Were there any more changes that need to be done to this?

library/LuaApi.cpp
```diff
@@ -1318,6 +1318,8 @@ static bool isSiteLoaded() { return World::IsSiteLoaded(); }
 static std::string df2utf(std::string s) { return DF2UTF(s); }
 static std::string utf2df(std::string s) { return UTF2DF(s); }
 static std::string df2console(color_ostream &out, std::string s) { return DF2CONSOLE(out, s); }
+static std::string cp437Upper(std::string s) { return toUpper_cp437(s); }
+static std::string cp437Lower(std::string s) { return toLower_cp437(s); }
```
myk002, line 1322: Could you rename these `upperCp437` (or `toUpperCp437`) and similar for `lower` to match the C++ names and existing Lua functions like `getClipboardTextCp437`?

myk002 (commented):
just a rename request; otherwise looks good to me

myk002 (approved):

PR #4144, opened 2024-01-06
add small letters and on-off textures in top-left corners of 32x32 tiles

TymurGubayev:
These will be used in the `gui/biomes` script (https://github.com/DFHack/scripts/pull/637)


myk002 (commented):
I suggest names with no spaces in them to match with existing filenames. it also looks like we prefer dashes to underscores

TymurGubayev:
is `on-off-top-left.png` ok? It looks like 4 things now... Maybe something like `on-off--top-left.png`?

myk002:
> is `on-off-top-left.png` ok? It looks like 4 things now... Maybe something like `on-off--top-left.png`?

a combination of dashes and underscores might be appropriate here -- `on-off_top-left.png` -- or maybe distinguish it by its tile size: `on-off_map.png`

TymurGubayev:
renamed with underscores (on the off chance we want bottom-right etc versions later)

myk002 (approved):

PR #4155, opened 2024-01-10
Update autoclothing.cpp

LightHardt:
(find_needed_clothing_items) Add check to skip items that are worn or more

(remove_available_clothing) Add check to skip items that are worn or more

(isAvailableItem) Change check on wear level to not include worn (x) items as available

I was running into an issue where autoclothing would not make new clothes and then tailor would have to start its own orders to clothe my dwarves (with clothes I didn't want them to wear). What has appeared to fix this is by having autoclothing not consider worn clothing as available/usable.

Any insight/feedback on this would be appreciated.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -66,6 +66,7 @@ Template for new versions:
 
 ## Misc Improvements
 - `autochop`: better error output when target burrows are not specified on the commandline
+- `autoclothing` : now does not consider worn (x) clothing as usable/available, should help with using with `tailor` at same time
 - wherever units are listed in DFHack tools, properties like "agitated" or (-trained-) are now shown
 - `work-now`: now saves its enabled status with the fort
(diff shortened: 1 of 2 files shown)


myk002 (commented):
@ab9rf what is your opinion on this? how does the wear/damage threshold behavior compare with `tailor`?

Could you add a note to `changelog.txt`? right about here: https://github.com/DFHack/dfhack/blob/develop/docs/changelog.txt#L68

LightHardt:
Updated changelog.

ab9rf:
> what is your opinion on this? how does the wear/damage threshold behavior compare with `tailor`?

that's the same check tailor uses. in game, worn clothing is available to be worn but gives dwarfs a negative thought, so `tailor` does not consider such clothing appropriate to wear. it's reasonable for `autoclothing` to do the same thing

ab9rf (approved):

PR #4173, opened 2024-01-14
add Units::chopTree

ab9rf:
this implements the bay12-provided entry point for chopping down trees

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -2346,3 +2346,15 @@ void Units::setGroupActionTimers(color_ostream &out, df::unit *unit, int32_t amo
         }
     }
 }
+
+void Units::chopTree(df::unit *un, int16_t x, int16_t y, int16_t z)
+{
+    CHECK_NULL_POINTER(un);
+
+    auto fp = df::global::unitst_choptree;
+    CHECK_NULL_POINTER(fp);
+
+    using FT = std::function<void(df::unit *un, int16_t x, int16_t y, int16_t z) > ;
+    auto f = reinterpret_cast<FT*>(fp);
+    (*f)(un, x, y, z);
+}
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -2347,14 +2347,18 @@ void Units::setGroupActionTimers(color_ostream &out, df::unit *unit, int32_t amo
     }
 }
 
-void Units::chopTree(df::unit *un, int16_t x, int16_t y, int16_t z)
+// the only reason this function would ever fail is if the global is not defined,
+// the underlying DF function has no failure mode
+bool Units::chopTree(df::unit* unit, int16_t x, int16_t y, int16_t z)
 {
-    CHECK_NULL_POINTER(un);
+    CHECK_NULL_POINTER(unit);
 
     auto fp = df::global::unitst_choptree;
-    CHECK_NULL_POINTER(fp);
+    if (fp == NULL)
+        return false;
 
-    using FT = std::function<void(df::unit *un, int16_t x, int16_t y, int16_t z) > ;
+    using FT = std::function<void(df::unit *unit, int16_t x, int16_t y, int16_t z) > ;
     auto f = reinterpret_cast<FT*>(fp);
-    (*f)(un, x, y, z);
+    (*f)(unit, x, y, z);
+    return true;
 }
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -1823,6 +1823,7 @@ static const LuaWrapper::FunctionReg dfhack_units_module[] = {
     WRAPM(Units, setActionTimers),
     WRAPM(Units, setGroupActionTimers),
     WRAPM(Units, getUnitByNobleRole),
+    WRAPM(Units, chopTree),
     { NULL, NULL }
 };
 
--- a/library/include/modules/Units.h
+++ b/library/include/modules/Units.h
@@ -60,189 +60,190 @@ namespace df
  */
 namespace DFHack
 {
-namespace Units
-{
+    namespace Units
+    {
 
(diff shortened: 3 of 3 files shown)


library/modules/Units.cpp
```diff
@@ -2346,3 +2346,15 @@ void Units::setGroupActionTimers(color_ostream &out, df::unit *unit, int32_t amo
         }
     }
 }
+
+void Units::chopTree(df::unit *un, int16_t x, int16_t y, int16_t z)
+{
+    CHECK_NULL_POINTER(un);
+
+    auto fp = df::global::unitst_choptree;
+    CHECK_NULL_POINTER(fp);
```
myk002, line 2355: you might not want to throw here. 50.10, which is still supported in the 50.11 binaries, won't have this global defined. maybe check null and return without action if not set?

myk002 (commented):
needs Lua API access (and accompanying documentation) and changelog entry

library/modules/Units.cpp
ab9rf: i'd rather have it throw than fail silently. note also that this code is not currently called from any existing code as it's an extension to the API, not a reimplementation

ab9rf: > For what it's worth, though, exceptions like the ones thrown by `CHECK_NULL_POINTER` will be converted to Lua errors if this function is called from Lua, so only C++ code would be at risk of crashing.

it occurs to me that we might be able to wrap a try block around plugin code and force-unload any plugin that throws instead of crashing DF as a whole, but that is a discussion for another day

for now, i'll have the function return a bool and it'll be up to downstream code to check for "success" that way. not happy about this really but for now it's probably our best available option

ab9rf:
> needs Lua API access (and accompanying documentation)

where does the "accompanying documentation" go? `docs/api` only has a section for `Maps` and that is dreadfully incomplete

myk002:
Lua docs go in the https://github.com/DFHack/dfhack/blob/develop/docs/dev/Lua%20API.rst mega-file

myk002:
the coordinates are supposed to be for a tree, right? I got a segfault when trying this out:
```c
#0  0x0000000001933374 in ?? ()
#1  0x00007ffff79b3ac3 in std::function<void (df::unit*, short, short, short)>::operator()(df::unit*, short, short, short) const (__args#3=<optimized out>, __args#2=<optimized out>, __args#1=<optimized out>, 
    __args#0=<optimized out>, this=<optimized out>)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/std_function.h:591
#2  DFHack::Units::chopTree (unit=<optimized out>, x=<optimized out>, y=<optimized out>, z=<optimized out>)
    at /home/myk/src/dfhack/library/modules/Units.cpp:2363
```

myk002:
hrm, it worked fine the second time. I wonder if the issue was that I chose the "wrong" tile of a multi-tile tree?

myk002:
seems to consistently crash on multi-tile trees.

ab9rf:
> I see the effects of remote chopping on the velocity vector of the logs. They shoot across the map if the unit "responsible" for the chopping isn't nearby.

putnam did warn about this, and rumrusher has confessed to having weaponized this in the past

ab9rf:
> I wonder if the issue was that I chose the "wrong" tile of a multi-tile tree?

the same function that the function pointer points to is also called by organic code, in code that looks like this:

```
    if (iVar121 == 0) {
      uVar117 = (pjVar92->pos).x;
      uVar39 = (pjVar92->pos).y;
      sVar37 = (pjVar92->pos).z;
      tVar45 = world::getTileType(&world,uVar117,uVar39,sVar37);
      uVar108 = FUN_1401ebcf0(tVar45);
      if ((char)uVar108 == '\0') {
        unitst::clearPath(param_1);
        *(undefined4 *)local_1950 = 0;
        pbVar83 = (building_actualst *)CONCAT71((int7)((ulonglong)local_1950 >> 8),1);
        *param_2 = 1;
        game.minimap.mustmake = 1;
      }
      else {
        unitst::chopTree(param_1,uVar117,uVar39,sVar37);
        *(undefined4 *)local_1950 = 0;
        if ((param_1->job).current_job == (job *)0x0) {
          uVar56 = unitst::clearPath(param_1);
          pbVar83 = (building_actualst *)CONCAT71((int7)((ulonglong)uVar56 >> 8),1);
          *param_2 = 1;
          game.minimap.mustmake = 1;
        }
        else {
          pbVar83 = (building_actualst *)((ulonglong)local_1950 & 0xffffffffffffff00);
          *param_2 = 0;
          game.minimap.mustmake = 1;
        }
      }
      break;
    }
```

myk002:
Is `pjVar92` the tree? I'll check what that says for multi-tile trees

ab9rf:
> Is `pjVar92` the tree? I'll check what that says for multi-tile trees

no, `pjVar92` is the chop tree job, so the coordinates should presumably match what you'd expect for what would be in the coordinates of a chop tree job designated in the "normal" way. `getplants` has fairly solid logic for this, iirc

ab9rf:
i just dug through the `chopTree` code and there's a subtle bug that makes this call a bit tricky. basically, after the tree is chopped, down, if the game is in dwarf mode, the game then does various things with respect to the _current_ job of the unit that was chosen to do the tree cutting. if that unit has a job but that job is _not_ a `chopTree` job (to cut down this particular tree, in all likelihood), it'll probably make a hash of that unit's current state. if the unit has no job (`current_job` is `NULL`) then there will be a mess of diagnostics issued, but the game should ultimately be ok with it.

in short, this entry point probably isn't going to be useful to us until Bay12 detangles the "chop tree down" part from the "record that this particular urist chopped down a tree" part

ab9rf:
closing this since the entry point is not actually useful anyway

PR #4217, opened 2024-01-26
Fix dig-now not updating flags when channeling

### LINKED ISSUE #3808: dig-now channels behaving differently than manually dug ones
Normally when channeling down tiles on the surface, newly grown grass may grow in those now-exposed tiles. This doesn't occur with dug tiles on the surface using the dig-now command.

I have a sample save where the top part of the cliff was channeled manually while the right bottom corner was done with dig-now.
Immediately after finishing the top part, new grass started to grow, while on the bottom corner it wouldn't.

The minimap also shows a different color on the flooring where the terrain was dug down with dig-now, appearing with a white hue. While the manually dug tiles appear as the same color as the rest of the natural surface tiles.

Save: https://www.mediafire.com/file/3cs4dta86pe0mbu/Dig-now+test.zip/file

MagistrumT-T:
Fixes #3808
Due to channel designations being handled with recursive calls and exiting early, propagate_vertical_flags() wasn't running on the position of the channel designation itself, but on the tile below. Adding a call to propagate_vertical_flags() before the early exit solved the issue.

### DIFF
--- a/plugins/dig-now.cpp
+++ b/plugins/dig-now.cpp
@@ -489,6 +489,7 @@ static bool dig_tile(color_ostream &out, MapExtras::MapCache &map,
                     if (td_below == df::tile_dig_designation::Default) {
                         dig_tile(out, map, pos_below, td_below, dug_tiles);
                     }
+                    propagate_vertical_flags(map, pos);
                     return true;
                 }
             } else {


myk002 (commented):
looks good! yeah, the ramp/channel code is the most complex case here. I'm not too surprised it had a bug.

Could you add a line to the "Fixes" section in the changelog?

myk002:
While you're there, could you also test the RAMP case? does that propagate flags correctly? Especially when you're digging a ramp and the tile above you is WALL but the tile above that is open to the sky.

plugins/dig-now.cpp
```diff
@@ -531,6 +532,7 @@ static bool dig_tile(color_ostream &out, MapExtras::MapCache &map,
                     map.setTiletypeAt(pos_above,
                             get_target_type(tt, df::tiletype_shape::RAMP_TOP));
                     remove_ramp_top(map, DFCoord(pos.x, pos.y, pos.z+2));
+                    propagate_vertical_flags(map, DFCoord(pos.x, pos.y, pos.z + 1));
```
myk002, line 535: now that we've handled channeling and ramps specifically, is there any reason to keep the `propagate_vertical_flags` call on line 556?

MagistrumT-T: I thought digging a wall with a upstairs below it would reveal a down stairs, but now that I tested and see that is not the case I don't know where I got the idea from.

MagistrumT-T:
Wait, stairs don't let light through so it doesn't matter anyway.
So removing the last call propagate_vertical_flags() now runs twice only in the channel case. Once I manage to understand the dug_tile_info struct I think it can be moved down to post processing like refresh_adjacent_smooth_walls() to avoid this problem, but I don't think it's too big a performance hit right now.

myk002:
No, performance would not likely be an issue. the flags only actually get propagated once per surface break

myk002 (approved):

Issue #4032, opened 2023-11-16
[buildingplan] Make planner UI show the currently selected filter for construction/furniture material


chdoc:
Currently, the planner UI only shows the number of available/missing construction materials/furniture items matching the currently selected material filter, but doesn't show what this filter is. When repeatedly altering between several filters (e.g. when using different stone types for different parts of the fort or different furniture materials for nobles/commoners)

myk002:
The issue here is spacing. I think this might depend on https://github.com/DFHack/dfhack/issues/3300 so the full filter text could show on hover.

chdoc:
Well, I'd be fine with an abridged/incomplete material info for filters that become too long (e.g., "23 rock blocks available" when blocks of different rock types are allowed for construction"). For furniture, I imagine something like: "Statue -> 12 made from green glass available" / "Table -> 8 made from gold needed". 

I suppose the rationale is that people who use permissive filters won't be changing these filters too often. And for people like me, who repeatedly switch between single material filters, this would be good enough. Once #3300 is available, you could add a tooltip to abridged filters showing the full filter definition.

PR #4219, opened 2024-01-27
[buildingplan] show abbreviated filter info in planner overlay

chdoc:
Fixes: https://github.com/DFHack/dfhack/issues/4032

### DIFF
--- a/plugins/lua/buildingplan/planneroverlay.lua
+++ b/plugins/lua/buildingplan/planneroverlay.lua
@@ -264,11 +264,11 @@ local function is_over_options_panel()
 end
 
 local function compress(str, len)
-    if str:len() <= len then
+    if #str <= len then
         return str
     else
         local no_vowels = str:gsub('[aeiou]','')
-        if no_vowels:len() <= len then
+        if #no_vowels <= len then
             return no_vowels
         else
             return no_vowels:sub(1,len-3)..'...'
--- a/plugins/lua/buildingplan/planneroverlay.lua
+++ b/plugins/lua/buildingplan/planneroverlay.lua
@@ -263,9 +263,24 @@ local function is_over_options_panel()
     return v:getMousePos()
 end
 
+local function compress(str, len)
+    if str:len() <= len then
+        return str
+    else
+        local no_vowels = str:gsub('[aeiou]','')
+        if no_vowels:len() <= len then
+            return no_vowels
+        else
+            return no_vowels:sub(1,len-3)..'...'
+        end
+    end
+end
 --------------------------------
 -- ItemLine
 --
+
+-- number of characters for item filter summary (excluding surrounding [ ])
+local item_filter_chars = 17
 
 ItemLine = defclass(ItemLine, widgets.Panel)
 ItemLine.ATTRS{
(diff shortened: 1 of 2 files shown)


plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -368,8 +370,44 @@ function ItemLine:has_filter()
...
+        uibs.building_type, uibs.building_subtype, uibs.custom_type, self.idx-1)
+    local enabled_mat_names = {}
+    local enabled_cat_names = {}
+    for name,props in pairs(mats) do
+        local enabled = props.enabled == 'true' and cats[props.category]
+        if enabled then table.insert(enabled_mat_names, name) end
+    end
+    if #enabled_mat_names == 1 then
+        local mat = enabled_mat_names[1]
+        if mat:len() < 17 then
```
myk002, line 391: 17 should be a constant so it can be shared between the field width+2 and this check

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -368,8 +370,44 @@ function ItemLine:has_filter()
...
+            return '['..mat..']'
+        else
+            return '['..mat:sub(1,14)..'...]'
```
myk002, line 394: DF has a habit of shortening strings by first removing vowels and then truncating. would that help improve information density here?

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -368,8 +370,44 @@ function ItemLine:has_filter()
...
+    elseif #enabled_mat_names > 1 then
+        for cat, _ in pairs(cats) do
+            if cat ~= 'unset' and cats[cat] then
+                table.insert(enabled_cat_names, cat)
+            end
+        end
+        if #enabled_cat_names == 1 then
+            return '[' .. enabled_cat_names[1]:gsub("^%l", string.upper) .. ']'
+        else
+            return '[various materials]'
```
myk002, line 405: maybe use the number of materials instead of "various"? that might be more informative

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -368,8 +370,44 @@ function ItemLine:has_filter()
...
+        end
+    else
+        -- can result from selecting wood and then toggling "fire safe" etc.
+        return '[impossible filter]'
```
myk002, line 409: you can control the pen to turn this red if you change this function to generate the entire `text` attribute table instead of just the text portion of the `text` attribute

plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -302,13 +317,17 @@ function ItemLine:init()
             frame={t=0, l=28},
             text={
                 {text=self:callback('get_filter_text'),
-                 pen=function() return gui.invert_color(COLOR_LIGHTCYAN, self.is_selected_fn()) end},
+                 width=item_filter_chars+2,
+                 rjustify=true,
+                 pen=function() return
+                         self:is_impossible() and COLOR_RED or
+                         gui.invert_color(COLOR_LIGHTCYAN, self.is_selected_fn()) end},
```
myk002, line 324: I thought this would be 
```lua
gui.invert_color(self:is_impossible() and COLOR_RED or COLOR_LIGHTCYAN, self.is_selected_fn()) end
```
I'll have to check in game for which is appropriate

chdoc: I started with the one you suggested. But then only highlighted row show as red and the others show in the same orange as `[clear]`, which also does not change based on selection.

Issue #4043, opened 2023-11-19
Access to geld/butcher/pet options from animal 'unit' info


AedisV:
In 0.44 I believe there was way to designate animal for butchering from animal 'unit' view, allowing to for example get rid of frail and sickly ones first, in 0.5 only way to accomplish it is somewhat tedious sequence of rename from unit screen then butcher from units listing.

PR #4224, opened 2024-01-27
Add Animal Options To Valid Animal Viewsheet

LightHardt:
Continuation of https://github.com/DFHack/scripts/pull/961

This will resolve https://github.com/DFHack/dfhack/issues/4043

Integration of script into zone.lua

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -54,7 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- `zone.animal_actions`: Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
+- ``zone.animal_actions``: Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
 
 ## Fixes
 - `sort`: fix potential crash when switching between info tabs
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -54,7 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- `zone.animal_actions` : Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
+- `zone.animal_actions`: Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
 
 ## Fixes
 - `sort`: fix potential crash when switching between info tabs
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -54,7 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- ``zone.animal_actions``: Add overlay to tamed animals of civ for butchering/gelding/adoption/tame(assign trainer) options in their "Overview" tab
+- `zone`: Add overlay for toggling butchering/gelding/adoption/taming options in animal "Overview" tabs
 
 ## Fixes
 - `sort`: fix potential crash when switching between info tabs
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -54,7 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- ``zone.animal_actions``: Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
+- ``zone.animal_actions``: Add overlay to tamed animals of civ for butchering/gelding/adoption/tame(assign trainer) options in their "Overview" tab
 
 ## Fixes
 - `sort`: fix potential crash when switching between info tabs
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -27,7 +27,6 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- `zone.animal_actions` : Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
 
 ## Fixes
 
@@ -55,6 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `zone.animal_actions` : Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
 
 ## Fixes
 - `sort`: fix potential crash when switching between info tabs
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -27,6 +27,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `zone.animal_actions` : Add overlay to tamed animals of civ for butchering/gelding/adoption options in their "Overview" tab
 
 ## Fixes
 
--- a/plugins/lua/zone.lua
+++ b/plugins/lua/zone.lua
@@ -1285,11 +1285,7 @@ function AnimalActionsWidget:set_geld_flag(option)
 
     if not unit then return end
 
-    if option then
-        unit.flags3.marked_for_gelding = option
-    else
-        unit.flags3.marked_for_gelding = option
-    end
+    unit.flags3.marked_for_gelding = option
 end
 
 -- set available for adoption flag
--- a/plugins/lua/zone.lua
+++ b/plugins/lua/zone.lua
@@ -1232,7 +1232,6 @@ end
 local function is_geldable()
     local unit = dfhack.gui.getSelectedUnit(true)
 
-    -- If the unit is not male or already gelded do nothing
     return dfhack.units.isGeldable(unit)
 end
 
--- a/plugins/lua/zone.lua
+++ b/plugins/lua/zone.lua
@@ -1220,9 +1220,6 @@ end
 -- AnimalActionsWidget
 --
 
-local YES = 'Yes'
-local NO = 'No'
-
 -- Make sure an animal unit of your civ is selected
 local function check_valid_unit()
     local unit = dfhack.gui.getSelectedUnit(true)
@@ -1248,8 +1245,9 @@ end
 local function is_avail_adoption()
     local unit = dfhack.gui.getSelectedUnit(true)
 
+    local creature_id = df.creature_raw.find(unit.race).creature_id
     -- Make sure not already a pet and not a cat as they adopt owner
-    return not dfhack.units.isPet(unit) and unit.race ~= 171
+    return not dfhack.units.isPet(unit) and creature_id ~= 'CAT'
 end
 
 AnimalActionsWidget=defclass(AnimalActionsWidget, overlay.OverlayWidget)
@@ -1269,9 +1267,9 @@ function AnimalActionsWidget:set_slaughter_flag(option)
     if not unit then return end
 
     if option then
-        unit.flags2.slaughter = true
+        unit.flags2.slaughter = option
     else
-        unit.flags2.slaughter = false
+        unit.flags2.slaughter = option
     end
 end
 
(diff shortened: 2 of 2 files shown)


plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,188 @@ function RetireLocationOverlay:confirm_retire()
...
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    return not dfhack.units.isPet(unit)
+end
+
+local function is_avail_adoption()
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    -- Make sure not already a pet and not a cat as they adopt owner
+    return not dfhack.units.isPet(unit) and unit.race ~= 171
```
myk002, line 1252: This won't be true for all worlds. Race IDs will change according to the mods you have loaded.

Is there a race trait we can check against? If not, you might have to compare against the string id (`CAT`)

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,188 @@ function RetireLocationOverlay:confirm_retire()
...
+}
+
+-- The above function already handles checking if valid unit
+-- so just set slaughter flag
+function AnimalActionsWidget:set_slaughter_flag(option)
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    if not unit then return end
+
+    if option then
```
myk002, line 1271: Can just have `unit.flags2.slaughter = option`

myk002: Similar for other callbacks

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,188 @@ function RetireLocationOverlay:confirm_retire()
...
+        unit.flags3.available_for_adoption = true
+    else
+        unit.flags3.available_for_adoption = false
+    end
+end
+
+-- Check current flag status of animal to dynamically set on/off
+function AnimalActionsWidget:get_butcher(unit)
```
myk002, line 1305: These get functions are unused and can be removed

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,188 @@ function RetireLocationOverlay:confirm_retire()
         curry(retire, location))
 end
 
+-- -------------------
+-- AnimalActionsWidget
+--
+
+local YES = 'Yes'
```
myk002, line 1223: These constants are of questionable utility

myk002:
Looking good! Don't forget the changelog entry!

plugins/lua/zone.lua
LightHardt: I found a method that caravan uses and slightly modified it and seems to work.

plugins/lua/zone.lua
LightHardt: made change

plugins/lua/zone.lua
LightHardt: Removed functions

plugins/lua/zone.lua
LightHardt: Removed constants

LightHardt:
Thank you for all the help this process would have been much slower without it!

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,154 @@ function RetireLocationOverlay:confirm_retire()
...
+    return unit
+        and dfhack.units.isFortControlled(unit)
+        and dfhack.units.isAlive(unit)
+        and dfhack.units.isAnimal(unit)
+end
+
+local function is_geldable()
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    return dfhack.units.isGeldable(unit)
```
myk002, line 1235:
```suggestion
    return unit and dfhack.units.isGeldable(unit)
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,154 @@ function RetireLocationOverlay:confirm_retire()
...
+end
+
+local function is_not_pet()
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    return not dfhack.units.isPet(unit)
```
myk002, line 1241:
```suggestion
    return not unit or not dfhack.units.isPet(unit)
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,154 @@ function RetireLocationOverlay:confirm_retire()
...
+end
+
+local function is_avail_adoption()
+    local unit = dfhack.gui.getSelectedUnit(true)
```
myk002, line 1245:
```suggestion
    local unit = dfhack.gui.getSelectedUnit(true)
    if not unit then return false end
```

LightHardt:
Wondering if i should rename the `Tame` label to `Train`. Not sure.

docs/changelog.txt
```diff
@@ -54,6 +54,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- ``zone.animal_actions``: Add overlay to tamed animals of civ for butchering/gelding/adoption/tame(assign trainer) options in their "Overview" tab
```
myk002, line 57:
```suggestion
- `zone`: Add overlay for toggling butchering/gelding/adoption/taming options in animal "Overview" tabs
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,186 @@ function RetireLocationOverlay:confirm_retire()
...
+
+    local creature_id = df.creature_raw.find(unit.race).creature_id
+    -- Make sure not already a pet and not a cat as they adopt owner
+    return not dfhack.units.isPet(unit) and creature_id ~= 'CAT'
```
myk002, line 1249:
```suggestion
    if dfhack.units.isPet(unit) return false end
    
    local raw = df.creature_raw.find(unit.race)
    if not raw then return false end
    
    -- cats adopt owners; owners can't adopt cats
    return raw.creature_id ~= 'CAT'
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,186 @@ function RetireLocationOverlay:confirm_retire()
...
+function AnimalActionsWidget:set_slaughter_flag(option)
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    if not unit then return end
+
+    if option then
+        unit.flags2.slaughter = option
+    else
+        unit.flags2.slaughter = option
+    end
```
myk002, line 1278:
```suggestion
    unit.flags2.slaughter = option
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,186 @@ function RetireLocationOverlay:confirm_retire()
...
+function AnimalActionsWidget:set_geld_flag(option)
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    if not unit then return end
+
+    if option then
+        unit.flags3.marked_for_gelding = option
+    else
+        unit.flags3.marked_for_gelding = option
+    end
```
myk002, line 1291:
```suggestion
    unit.flags3.marked_for_gelding = option
```

plugins/lua/zone.lua
```diff
@@ -1216,10 +1216,186 @@ function RetireLocationOverlay:confirm_retire()
...
+function AnimalActionsWidget:set_adoption_flag(option)
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    if not unit then return end
+
+    if option then
+        unit.flags3.available_for_adoption = option
+    else
+        unit.flags3.available_for_adoption = option
+    end
```
myk002, line 1304:
```suggestion
    unit.flags3.available_for_adoption = option
```

myk002 (commented):
love it. I suggested a few cleanups. The guarding against nil units is good practice, even though the current setup *should* prevent them from ever being nil. If things change in the future, though, it's best not to have an error.

myk002:
> Wondering if i should rename the `Tame` label to `Train`? Not sure.

I'd go with "Tame" since we're not specifically training for war/hunting (though that's a potential future improvement to offer those options)

> Also `isMarkedForTraing` is marking my war Giant Leopard as `Yes` should i also add a `isDomesticated` check for that? Leaning towards no since it is indeed marked for training and it being domesticated already disables the button so really it would be a visual matter of wanting it to say either No or Yes but curious what you think.

I'd go with keeping it as it is -- it does have a trainer assigned, after all.

plugins/lua/zone.lua
LightHardt: Understood

plugins/lua/zone.lua
LightHardt: I now see what you were trying to say my bad

myk002:
> I'd go with "Tame" since we're not specifically training for war/hunting (though that's a potential future improvement to offer those options)

Looking at this in-game..
image

the pig is tame (domesticated) but we say `Tame: No`

Maybe change that to "has trainer"?

myk002:
also looking in-game, maybe line up the labels so it's more like:
```lua
Butcher: No
Geld:    No
Adopt:  Yes
Tame     No
```

myk002:
Finally, it seems that there are some invariants that we have to enforce:

https://github.com/DFHack/dfhack/assets/977482/2f92376b-3302-4c8e-aadb-e7f733b50001

Marking for slaughtering unmarks adoption and gelding
marking for adoption unmarks slaughter
marking for gelding unmarks slaughter

LightHardt:
> Marking for slaughtering unmarks adoption and gelding marking for adoption unmarks slaughter marking for gelding unmarks slaughter

Thats new behavior I tested earlier and that wasn't an issue I wonder which commit it was, ill have to look.

myk002:
> Thats new behavior I tested earlier and that wasn't an issue I wonder which commit it was, ill have to look.

What I was screen capturing was vanilla behavior -- I'm saying we should make sure to emulate it

LightHardt:
> > Thats new behavior I tested earlier and that wasn't an issue I wonder which commit it was, ill have to look.
> 
> What I was screen capturing was vanilla behavior -- I'm saying we should make sure to emulate it

Lol i got it backwards my bad, i see what you're saying i was thinking a shared variable but windering how that would work with multiple units.

myk002:
I think it can just be enforced in the "set" functions -- when we're setting  one of them to true, unconditionally set the ones that should be unmarked to false

myk002:
e.g.: 

```lua
function AnimalActionsWidget:set_slaughter_flag(option)
    local unit = dfhack.gui.getSelectedUnit(true)
    if not unit then return end
    if option then
        self:set_geld_flag(false)
        self:set_adopt_flag(false)
    end
    unit.flags2.slaughter = option
end
```

LightHardt:
Switched to `Has Trainer`
Padded labels so all even
Emulate vanilla method with being only able to set 1 at a time (thank you)

Let me know what you think!

plugins/lua/zone.lua
```diff
@@ -1285,6 +1292,11 @@ function AnimalActionsWidget:set_geld_flag(option)
 
     if not unit then return end
 
+    if option then
+        self:set_slaughter_flag(false)
+        self:set_adoption_flag(false)
```
myk002, line 1297: Setting geld does not unset adoption (which makes sense)

plugins/lua/zone.lua
```diff
@@ -1294,6 +1306,11 @@ function AnimalActionsWidget:set_adoption_flag(option)
 
     if not unit then return end
 
+    if option then
+        self:set_geld_flag(false)
```
myk002, line 1310: likewise, setting adopt does not affect the geld flag

myk002 (approved):

myk002:
This looks so nice. Thank you very much for contributing it!

LightHardt:
> This looks so nice. Thank you very much for contributing it!

Thanks for helping me pretty much all day with it, it's much appreciated!

PR #4227, opened 2024-01-28
Small fixes to show on caged animals

LightHardt:
Noticed that the overlay didn't appear on a crow I caught in a cage trap so made some tweaks to fix that.

Also added additional checks as without it, it lets you set the slaughter flag as well as the others on a caged animal you haven't even tamed yet.

### DIFF
--- a/plugins/lua/zone.lua
+++ b/plugins/lua/zone.lua
@@ -1227,18 +1227,21 @@ local function check_valid_unit()
         and dfhack.units.isFortControlled(unit)
         and dfhack.units.isAlive(unit)
         and dfhack.units.isAnimal(unit)
+        or dfhack.units.isTamable(unit)  -- account for caged animals
+        and unit.flags1.caged
+        and not unit.flags1.tame -- dont show on merchant animals
 end
 
 local function is_geldable()
     local unit = dfhack.gui.getSelectedUnit(true)
 
-    return unit and dfhack.units.isGeldable(unit)
+    return unit and dfhack.units.isGeldable(unit) and dfhack.units.isFortControlled(unit)
 end
 
 local function is_not_pet()
     local unit = dfhack.gui.getSelectedUnit(true)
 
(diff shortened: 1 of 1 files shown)


LightHardt:
After gelding an animal noticed the toggle was still enabled.

Added a flag check to solve this problem

plugins/lua/zone.lua
```diff
@@ -1227,18 +1227,21 @@ local function check_valid_unit()
...
+        or dfhack.units.isTamable(unit)  -- account for caged animals
+        and unit.flags1.caged
+        and not unit.flags1.tame -- dont show on merchant animals
 end
 
 local function is_geldable()
     local unit = dfhack.gui.getSelectedUnit(true)
 
-    return unit and dfhack.units.isGeldable(unit)
+    return unit and dfhack.units.isGeldable(unit) and dfhack.units.isFortControlled(unit) and not unit.flags3.gelded
```
myk002, line 1238:
```suggestion
    return unit and dfhack.units.isGeldable(unit) and dfhack.units.isFortControlled(unit) and not dfhack.units.isGelded(unit)
```

myk002: this is a more complicated check than just the flag -- should use the API function

plugins/lua/zone.lua
```diff
@@ -1227,18 +1227,21 @@ local function check_valid_unit()
         and dfhack.units.isFortControlled(unit)
         and dfhack.units.isAlive(unit)
         and dfhack.units.isAnimal(unit)
+        or dfhack.units.isTamable(unit)  -- account for caged animals
```
myk002, line 1230: sticking an `or` in the middle of a bunch of `and`s is difficult to read. Could you refactor this into sub-functions? e.g.
```lua
return unit and (isFortAnimal(unit) or isCagedWildAnimal(unit))
```

myk002 (approved):

PR #4242, opened 2024-02-01
Make `dig-now` remove ramps for all dig modes.

### LINKED ISSUE #4213: dig-now leaves behind ramps that it shouldn't and removes ramps that it should have left.
dig-now does not remove unusable ramps positioned diagonally from a dig designation, and removes usable diagonal ramps.

MagistrumT-T:
Fixes #4213 by looking for ramps for every designation type and adding the missing diagonal case to clean_ramp() and clean_ramps().

### DIFF
--- a/plugins/dig-now.cpp
+++ b/plugins/dig-now.cpp
@@ -344,7 +344,11 @@ static void clean_ramp(MapExtras::MapCache &map, const DFCoord &pos) {
     if (is_wall(map, DFCoord(pos.x-1, pos.y, pos.z)) ||
             is_wall(map, DFCoord(pos.x+1, pos.y, pos.z)) ||
             is_wall(map, DFCoord(pos.x, pos.y-1, pos.z)) ||
-            is_wall(map, DFCoord(pos.x, pos.y+1, pos.z)))
+            is_wall(map, DFCoord(pos.x, pos.y+1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x-1, pos.y-1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x-1, pos.y+1, pos.z)) ||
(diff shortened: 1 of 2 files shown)


plugins/dig-now.cpp
```diff
@@ -344,7 +344,11 @@ static void clean_ramp(MapExtras::MapCache &map, const DFCoord &pos) {
     if (is_wall(map, DFCoord(pos.x-1, pos.y, pos.z)) ||
             is_wall(map, DFCoord(pos.x+1, pos.y, pos.z)) ||
             is_wall(map, DFCoord(pos.x, pos.y-1, pos.z)) ||
-            is_wall(map, DFCoord(pos.x, pos.y+1, pos.z)))
+            is_wall(map, DFCoord(pos.x, pos.y+1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x-1, pos.y-1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x-1, pos.y+1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x+1, pos.y-1, pos.z)) ||
+            is_wall(map, DFCoord(pos.x+1, pos.y+1, pos.z)))
```
myk002, line 351: I don't think this is correct. This is checking for ramp usability, and diagonal walls do not make the ramp usable

MagistrumT-T: Maybe that was old behavior? DF wasn't marking them as unusable nor removing them when digging, so I included them. Lemme try creating with miners the normal way and test real quick if they get used by dwarves.

MagistrumT-T: Yeah, dwarves will use diagonal ramps without issue. That tracks to my experience in adventure mode using diagonal ramps for as long as I remember, so I'm not sure when this got changed, but evidently it did.

myk002 (approved):

Issue #3817, opened 2023-09-27
Cleanowned without dumping


ae3qe27u:
Hi!

How hard would it be to have a cleanowned option that confiscates dropped items/clothing without also dumping them? I like to have two clothing stockpiles - one directly from the clothier's shop with fresh clothing, and then a second that gathers from anywhere with all the discarded stuff, and so I do want to keep the stuff that's tattered/worn. I currently work around it by going into the stock page and just un-marking the items for deletion, but if it'd be straightforward to add an option that confiscates scattered stuff without also dumping it, that would be fantastic.

myk002:
yes, `cleanowned` could be modified to just clear the ownership and not mark the item for dumping

PR #4244, opened 2024-02-02
Add a "nodump" Option to cleanowned

LightHardt:
Fixes: https://github.com/DFHack/dfhack/issues/3817

Kept changes as small as possible while adding option

Changes

- Update Docs relating to change
- Show correct output for when dumping & confiscating or only confiscating

If need me to change anything just let me know.

### DIFF
--- a/docs/plugins/cleanowned.rst
+++ b/docs/plugins/cleanowned.rst
@@ -31,7 +31,7 @@ You can confiscate additional types of items by adding them to the commandline:
 ``X``
     Confiscate/dump items with wear level 'X' (heavily worn) and more.
 ``nodump``
-    Don't dump items only confiscate
+    Cause dwarves to drop confiscated items, but don't mark them for dumping.
 
 Or you can confiscate all owned items by specifying ``all``.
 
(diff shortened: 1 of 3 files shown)


plugins/cleanowned.cpp
```diff
@@ -160,7 +169,7 @@ command_result df_cleanowned (color_ostream &out, vector <string> & parameters)
             {
                 if (!Items::setOwner(item,NULL))
                     out.print("(unsuccessfully) ");
-                if (dump)
+                if (dump && !nodump)
```
myk002, line 172: would it be possible to not set `dump=true` in the first place instead of canceling it out here? that might make the semantics clearer in the code.

docs/plugins/cleanowned.rst
```diff
@@ -30,6 +30,8 @@ You can confiscate additional types of items by adding them to the commandline:
     Confiscate/dump items with wear level 'x' (lightly worn) and more.
 ``X``
     Confiscate/dump items with wear level 'X' (heavily worn) and more.
+``nodump``
+    Don't dump items only confiscate
```
myk002, line 34:
```suggestion
    Cause dwarves to drop confiscated items, but don't mark them for dumping.
```

plugins/cleanowned.cpp
LightHardt: I can move the `dump = true` into the else statements one second

LightHardt: Moved for better readability

myk002 (approved):

PR #4261, opened 2024-02-09
Add "flask-contents" tweak

quietust:
Makes flasks, vials, and waterskins display the name of their contents, just like other containers do (e.g. barrels, bins, and cages).

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,7 @@ Template for new versions:
 
 ## New Features
 - `cleanowned`: Add a "nodump" option to allow for confiscating items without dumping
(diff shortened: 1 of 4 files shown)


myk002 (commented):
a changelog entry for "New Feature" would be good. also, please add an entry to the control panel registry: https://github.com/DFHack/scripts/blob/master/internal/control-panel/registry.lua
bugfix section. you can copy the formatting of a record from one of the other "tweak" entries.

quietust:
Once the scripts PR is merged, I'll update this one with the changelog entry (and bump the submodule).

myk002 (approved):

PR #4265, opened 2024-02-11
remove tree roots from non-diggable material types

Heneman:


Heneman:
My rationale for this change was that tree root tiles kept interrupting large `dig-now` designations; if I'm already using `dig-now` for something I'd rather not care about removing tree roots :rofl: :sob:

myk002:
I believe the reason tree roots are skipped is because it corrupts the tree data structure. The tree structure is very complex, and dig-now doesn't have the logic to deal with it properly.

from the [docs](https://docs.dfhack.org/en/latest/docs/tools/dig-now.html#dig-now): ```md
Trees and roots are not currently handled by this plugin and will be skipped. Requests for engravings are also skipped since they would depend on the skill and creative choices of individual engravers. Other types of engraving (i.e. smoothing and track carving) are handled.
```

myk002:
Closing since this isn't feasible at the moment.

PR #4267, opened 2024-02-13
Update Units::getVisibleName

Bumber64:
Needs to use `impersonated_hf`, if exists.

Based on reverse engineering of v50.11, and is trivial to confirm just by changing the id of `impersonated_hf` and watching the unit's name change in the UI.

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -1087,7 +1087,7 @@ df::language_name *Units::getVisibleName(df::unit *unit)
     if (identity)
     {
         auto imp_hf = df::historical_figure::find(identity->impersonated_hf);
-        return (imp_hf != NULL && imp_hf->name.has_name) ? &imp_hf->name : &identity->name;
+        return (imp_hf && imp_hf->name.has_name) ? &imp_hf->name : &identity->name;
     }
 
     return &hf->name;
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -1087,7 +1087,7 @@ df::language_name *Units::getVisibleName(df::unit *unit)
     if (identity)
     {
         auto imp_hf = df::historical_figure::find(identity->impersonated_hf);
-        return (imp_hf && imp_hf->name.has_name) ? &imp_hf->name : &identity->name;
+        return (imp_hf != NULL && imp_hf->name.has_name) ? &imp_hf->name : &identity->name;
     }
 
     return &hf->name;
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -1086,7 +1086,7 @@ df::language_name *Units::getVisibleName(df::unit *unit)
     auto identity = getFigureIdentity(hf);
     if (identity)
     {
-        auto imp_hf = df::historical_figure.find(identity->impersonated_hf);
+        auto imp_hf = df::historical_figure::find(identity->impersonated_hf);
         return (imp_hf && imp_hf->has_name) ? &imp_hf->name : &identity->name;
     }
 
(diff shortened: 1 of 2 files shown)


library/modules/Units.cpp
```diff
@@ -1079,11 +1079,18 @@ df::language_name *Units::getVisibleName(df::unit *unit)
...
+    auto hf = df::historical_figure::find(unit->hist_figure_id);
+    if (!hf)
+        return &unit->name;
 
-    return &unit->name;
+    auto identity = getFigureIdentity(hf);
+    if (identity)
+    {
+        auto imp_hf = df::historical_figure::find(identity->impersonated_hf);
+        return (imp_hf && imp_hf->name && imp_hf->name->has_name) ? &imp_hf->name : &identity->name;
```
ab9rf: `historical_figure::name` is a `language_name`, not a `language_name *`, so you can't use `operator->` or attempt an implicit cast to `bool` here as `language_name` has no such conversion

this should probably be `(imp_hf != NULL && imp_hf->name.has_name)`.

myk002 (commented):
Shouldn't this fix happen in `getFigureIdentity`, not in `getVisibleName`?

ab9rf:
this looks to be inspired by `FUN_1409089c0` which is most likely `namest* identityst::get_active_name_ptr()`

however, we don't have a module for dealing with `identityst` objects or even with histfigs, so `Units` is as close as we can get, i suppose

decompile of `FUN_1409089c0`:
```

language_name * FUN_1409089c0(identity *param_1)

{
  int iVar1;
  int iVar2;
  int iVar3;
  int iVar4;
  longlong lVar5;
  historical_figure *phVar6;
  int iVar7;
  int iVar8;
  
  iVar1 = param_1->impersonated_hf;
  if (iVar1 != -1) {
    lVar5 = (longlong)world.history.figures._M_finish - (longlong)world.history.figures._M_start >>
            3;
    if (lVar5 != 0) {
      iVar7 = 0;
      iVar8 = (int)lVar5 + -1;
      if (-1 < iVar8) {
        do {
          iVar4 = iVar8 + iVar7 >> 1;
          phVar6 = world.history.figures._M_start[iVar4];
          iVar2 = phVar6->id;
          if (iVar2 == iVar1) goto LAB_140908a33;
          iVar3 = iVar4 + -1;
          if (iVar2 <= iVar1) {
            iVar3 = iVar8;
          }
          iVar8 = iVar3;
          if (iVar2 <= iVar1) {
            iVar7 = iVar4 + 1;
          }
        } while (iVar7 <= iVar8);
      }
    }
    phVar6 = (historical_figure *)0x0;
LAB_140908a33:
    if ((phVar6 != (historical_figure *)0x0) && ((phVar6->name).has_name != false)) {
      return &phVar6->name;
    }
  }
  return &param_1->name;
}
```

myk002:
Yeah, our current function for this is `Units::getIdentity()` (which calls `Units::getFigureIdentity()`). I'm pretty sure *that* is the function where a fix needs to happen.

Bumber64:
> Yeah, our current function for this is `Units::getIdentity()` (which calls `Units::getFigureIdentity()`). I'm pretty sure _that_ is the function where a fix needs to happen.

Our existing functions return `identity *` instead of `language_name *`, though. Should we just return an identity based on `imp_hf->name.has_name`?

The decompiled code is just these two lines:
```
auto imp_hf = df::historical_figure::find(identity->impersonated_hf);
return (imp_hf && imp_hf->name.has_name) ? &imp_hf->name : &identity->name;
```

myk002:
I see what you're saying. I think what you have is fine.

myk002 (approved):

PR #4273, opened 2024-02-15
Fix Enabled Behavior autoclothing

LightHardt:
Noticed `autoclothing` was not staying enabled for my fort after saving, quitting, and then reloading. Used tailor.cpp as a reference for the fix.

Tested fix and appears to work.


myk002 (commented):
yup, that would do it. thank you!

Please also add a line to the "fixes" section in the changelog

LightHardt:
No problem, glad to help!

myk002 (approved):

PR #4281, opened 2024-02-18
[buildingplan] use closest matching item

chdoc:
I took a first shot at making `buildingplan` attach the closest matching item rather than the newest matching item. Only changing `doVector` roughly as follows:

1. screen the item vector using `itemPassesScreen`.
2. Cycle through the task buckets. For every bucket
    a) try to get a valid building+task
    b) if this is the first valid task, filter available items by the buckets filter
    c) if there are matching items, attach the "closest" item to current task
    d) clean up and continue with (a)

For now, there is no clustering, but I'm not even sure that's necessary. That code doesn't necessarily qualify as "hot" and even in extreme cases (e.g., create 10K blocks and designate a blob of 5K stairs) my 8 year old machine takes only about 1s to do the assignment. The point is that items are de-referenced only in three places: screening, material filtering, and assigning. In particular, not when searching for the closest item, and the vector of matching items will probably be in the CPU cache. 

This version relies on `matchesFilters` (seen as a predicate on items) behaving the same for all tasks in a bucket.

### DIFF
--- a/plugins/buildingplan/buildingplan_cycle.cpp
+++ b/plugins/buildingplan/buildingplan_cycle.cpp
@@ -230,7 +230,7 @@ static df::building * popInvalidTasks(color_ostream &out, Bucket &task_queue,
 
 
 static int distance(df::coord pos1, df::coord pos2) {
-    return std::max(abs(pos1.x - pos2.x), abs(pos1.y - pos2.y) + abs(pos1.z - pos2.z));
+    return std::max(abs(pos1.x - pos2.x), abs(pos1.y - pos2.y)) + abs(pos1.z - pos2.z);
 }
 
 static void doVector(color_ostream &out, df::job_item_vector_id vector_id,
@@ -285,7 +285,7 @@ static void doVector(color_ostream &out, df::job_item_vector_id vector_id,
                                        pb.heat_safety,
                                        pb.item_filters[rev_filter_idx],
                                        pb.specials))
-                        matching.emplace_back(item->pos,item);
+                        matching.emplace_back(Items::getPosition(item),item);
 
                 num_matching = matching.size();
                 first_task = false;
@@ -324,7 +324,8 @@ static void doVector(color_ostream &out, df::job_item_vector_id vector_id,
                 // items so if buildingplan is turned off, the building will
                 // be completed with the correct number of items.
                 --jitems[filter_idx]->quantity;
-                // invaldate attached item
+                // null the item* component in vector of matching items
+                // ensures we don't try to attach this item to another job
                 closest->second = nullptr;
                 --num_matching;
                 // try to finalize building
--- a/plugins/buildingplan/buildingplan_cycle.cpp
+++ b/plugins/buildingplan/buildingplan_cycle.cpp
@@ -228,84 +228,125 @@ static df::building * popInvalidTasks(color_ostream &out, Bucket &task_queue,
     return NULL;
 }
 
+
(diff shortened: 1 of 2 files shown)


plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -228,84 +228,125 @@ static df::building * popInvalidTasks(color_ostream &out, Bucket &task_queue,
...
+            if (first_task) {
+                matching.clear();
+                for(auto item : available)
+                    if (!item->flags.bits.in_job &&
+                        matchesFilters(item,
+                                       jitems[filter_idx],
+                                       pb.heat_safety,
+                                       pb.item_filters[rev_filter_idx],
+                                       pb.specials))
+                        matching.emplace_back(item->pos,item);
```
myk002, line 288: you should probably use `Items::getPosition(item)` instead of `item->pos` since the `pos` field may be garbage if the item is in a container

plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -228,84 +228,125 @@ static df::building * popInvalidTasks(color_ostream &out, Bucket &task_queue,
...
                       bucket_it->first.c_str());
+                // clean up fulfilled task
+                task_queue.pop_front();
                 // keep quantity aligned with the actual number of remaining
                 // items so if buildingplan is turned off, the building will
                 // be completed with the correct number of items.
                 --jitems[filter_idx]->quantity;
-                task_queue.pop_front();
+                // invaldate attached item
+                closest->second = nullptr;
```
myk002, line 328: `closest` goes out of scope in a few lines. moreover, the item it points to is still valid. why invalidate?

myk002 (commented):
looks solid! I'll give this a test run and see how it behaves

plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -228,84 +228,125 @@ static df::building * popInvalidTasks(color_ostream &out, Bucket &task_queue,
     return NULL;
 }
 
+
+static int distance(df::coord pos1, df::coord pos2) {
+    return std::max(abs(pos1.x - pos2.x), abs(pos1.y - pos2.y) + abs(pos1.z - pos2.z));
```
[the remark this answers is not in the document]
chdoc: Umm, yes, the metric I typed is even more arbitrary than the metric I intended. Nice catch.

plugins/buildingplan/buildingplan_cycle.cpp
chdoc: Thanks, will do.

plugins/buildingplan/buildingplan_cycle.cpp
chdoc: Evidently, the comment isn't clear. `closest` is a pointer to a `(coord,item*)` cell in `matching`, so it nulls the second component of that, ensuring that this item gets skipped when looking for the closest item for subsequent tasks. Hence the `if (p.second && ...)`

chdoc:
Disclaimer: I made the changes you suggested; they compile, but I have no time to test again right now. Should be benign though, I guess.

myk002 (approved):

plugins/buildingplan/buildingplan_cycle.cpp
myk002: Ah ok, I missed that. I conflated it with the `in_job` check for the available vector.

PR #4299, opened 2024-02-25
Add `Units::isNobleFromOtherSite`

ong-yinggao98:
Function checks if a unit is a resident of your fort holding a landholder position belonging to another site. To be used for an upcoming Lua script to emigrate such nobility to avoid mandate hell and immigrate rulers that may be elsewhere due to succession.

I'll request a review for merge when that script is ready, in the meantime I don't think people may need this just yet.

TODO: Figure out how to do finding local rulers.

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -1719,6 +1719,7 @@ static const LuaWrapper::FunctionReg dfhack_units_module[] = {
     WRAPM(Units, isFortControlled),
     WRAPM(Units, isOwnCiv),
     WRAPM(Units, isOwnGroup),
+    WRAPM(Units, isNobleFromOtherSite),
     WRAPM(Units, isOwnRace),
     WRAPM(Units, isAlive),
     WRAPM(Units, isDead),
--- a/library/include/modules/Units.h
+++ b/library/include/modules/Units.h
@@ -81,6 +81,7 @@ DFHACK_EXPORT bool isFortControlled(df::unit *unit);
 DFHACK_EXPORT bool isOwnCiv(df::unit* unit);
 DFHACK_EXPORT bool isOwnGroup(df::unit* unit);
 DFHACK_EXPORT bool isOwnRace(df::unit* unit);
+DFHACK_EXPORT bool isNobleFromOtherSite(df::unit* unit);
 
 DFHACK_EXPORT bool isAlive(df::unit *unit);
 DFHACK_EXPORT bool isDead(df::unit *unit);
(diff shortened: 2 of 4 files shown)


library/include/modules/Units.h
```diff
@@ -81,6 +81,7 @@ DFHACK_EXPORT bool isFortControlled(df::unit *unit);
 DFHACK_EXPORT bool isOwnCiv(df::unit* unit);
 DFHACK_EXPORT bool isOwnGroup(df::unit* unit);
 DFHACK_EXPORT bool isOwnRace(df::unit* unit);
+DFHACK_EXPORT bool isNobleFromOtherSite(df::unit* unit, bool ignore_own = true);
```
myk002, line 84: functions that are exposed to Lua via WRAPM must not have parameters that default to `true`. The Lua layer only knows what the type signature is -- it doesn't know anything about C++ default parameters. Calling this function from Lua as `dfhack.units.isNobleFromOtherSite(unit)` will result in `ignore_own == false`.

library/LuaApi.cpp
```diff
@@ -1719,6 +1719,7 @@ static const LuaWrapper::FunctionReg dfhack_units_module[] = {
     WRAPM(Units, isFortControlled),
     WRAPM(Units, isOwnCiv),
     WRAPM(Units, isOwnGroup),
+    WRAPM(Units, isNobleFromOtherSite),
```
myk002, line 1722: needs documentation in `docs/dev/Lua API.rst` and `docs/changelog.txt` entries in the API and Lua sections

myk002 (commented):
What is the motivation for adding this function? Do you have a script in mind (or in progress) that needs it? Are you planning on improving one of DFHack's tools? I'm asking because I have some doubts about the proposed API and I'd like to understand the context more.

ong-yinggao98:
Yeah, I should have mentioned that earlier on. My plan is to use this for a
script to find and emigrate nobles from other sites living in your fortress
to avoid excessive mandates. I put this in C++ because I was unsure if I
should implement such logic in Lua. I'll respond to the other comments
later today.

myk002:
I think for your use case you'd be better off refactoring the code from here: https://github.com/DFHack/scripts/blob/master/internal/caravan/common.lua#L395

Note that IS_LAW_MAKER does *not* necessarily mean that the unit makes mandates. You need to look at the `mandate_max` field.

ong-yinggao98:
Thanks for the heads up, I think I will close this PR for now.

> Note that IS_LAW_MAKER does _not_ necessarily mean that the unit makes mandates. You need to look at the `mandate_max` field.

I was thinking that beyond just mandates, "foreign" nobles also impose room requirements and are a weird narrative thing too. Reasonably you'd expect them to rush home to start ruling proper which is why I was looking for landholders. I suppose that is also something I can look at from the Lua API too!

myk002:
if you take this back up, maybe consider adding it as a feature for `emigration`. It would fit in well there.

PR #4305, opened 2024-02-28
port `suspendmanager` to C++

chdoc:
- new C++ plugin `suspendmanager.cpp`
- new accompanying Lua module (containing the widgets from `scripts/suspendmanager.lua`)
- requires changes to scripts:
  1. remove `suspendmanager.lua`
  2. patch `unsuspend.lua` (suspend marker rendering)


plugins/lua/suspendmanager.lua
```diff
@@ -0,0 +1,130 @@
...
+StatusOverlay.ATTRS{
+    desc='Adds information to suspended building panels about why it is suspended.',
+    default_pos={x=-39,y=16},
+    default_enabled=true,
+    viewscreens='dwarfmode/ViewSheets/BUILDING',
+    frame={w=59, h=3},
+    frame_style=gui.MEDIUM_FRAME,
+    frame_background=gui.CLEAR_PEN,
+    default_enabled=true,
+    overlay_only=true,
```
myk002, line 45: Since this is a repositionable widget, `overlay_only` should not be specified

plugins/lua/suspendmanager.lua
```diff
@@ -0,0 +1,130 @@
...
+ToggleOverlay = defclass(ToggleOverlay, overlay.OverlayWidget)
+ToggleOverlay.ATTRS{
+    desc='Adds a link to suspended building panels for enabling suspendmanager.',
+    default_pos={x=-57,y=23},
+    default_enabled=true,
+    viewscreens='dwarfmode/ViewSheets/BUILDING',
+    frame={w=40, h=1},
+    frame_background=gui.CLEAR_PEN,
+    default_enabled=true,
+    overlay_only=true,
```
myk002, line 84: Same here. This widget is repositionable

plugins/lua/suspendmanager.lua
```diff
@@ -0,0 +1,130 @@
+local _ENV = mkmodule('plugins.suspendmanager')
+
+local utils = require('utils')
+local gui = require('gui')
+local overlay = require('plugins.overlay')
+local widgets = require('gui.widgets')
+
+--- Loop over all the construction jobs
+---@param fn function A function taking a job as argument
+function foreach_construction_job(fn)
```
myk002, line 10: Who uses this function? Can it be moved to the calling module?

myk002 (commented):
Needs suspendmanager docs

plugins/lua/suspendmanager.lua
chdoc: Oh, I think this was an unsuccessful experiment that I forgot to revert. Nice catch.

plugins/lua/suspendmanager.lua
chdoc: Used by `unsuspend` , could be moved there. Although I think that the more reasonable option is to merge `unsuspend` into `suspendmanager`. The widgets can be moved into the Lua module, and the command can be provided by the c++ module. Bonus: the PR to move the rendering of suspension icons to c++ would then become `dfhack` only.

myk002:
oh, hilarious. the docs build is still passing since the plugin docs and the script docs are *merged* before rendering. so it doesn't matter that there are two docs named `suspendmanager.rst` since one overwrites the other as part of the build process.

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+        job->flags.bits.suspend = false;
+    }
+
+    static bool walkable (coord pos) { return Maps::getWalkableGroup(pos) > 0; }
+
+    static bool buildingOnDesignation(df::building *building){
+        CHECK_NULL_POINTER(building);
+        auto z = building->z;
+        for (auto x = building->x1; x <= building->x2; ++x){
+            for (auto y = building->y1; x <= building->y2; ++x){
```
myk002, line 272: ++y?

plugins/lua/suspendmanager.lua
```diff
@@ -0,0 +1,131 @@
...
+        end
+    end
+end
+
+function isKeptSuspended(job)
+    return suspendmanager_isKeptSuspended(job)
+end
+
+-- TODO: check that this is appropriate
+function isBuildingPlanJob(job)
```
myk002, line 23: this code is duplicated in the cpp -- how about exposing that one and removing the Lua version?

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+            return false; // not building a blocking construction, no risk
+
+        coord pos(building->centerx,building->centery,building->z);
+        if (!isSuitableAccess(pos))
+            return false; // construction is on a non-walkable tile, can't block
+
+        auto risk = riskOfStuckConstructionAt(pos);
+        TRACE(cycle,out).print("  risk is %d\n", risk);
+
+        // TOTHINK: on a large grid, this will compute the same risk up to 5 times
```
myk002, line 389: would have to test to see if it matters -- in previous tests (in other plugins), maintaining a hashset to reduce duplication was slower than just doing the work multiple times

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+        if (construction_floor_support[constr_type]){
+            wall_would_support = &neighboursWallSupportsFloor;
+            floor_would_support = &neighboursFloorSupportsFloor;
+            supportbld_would_support = vector<offset>{offset(0,0,-1)};
+        } else if (construction_wall_support[constr_type]){
+            wall_would_support = &neighboursWallSupportsWall;
+            floor_would_support = &neighboursFloorSupportsWall;
+            supportbld_would_support = vector<offset>{offset(0,0,-1), offset(0,0,+1)};
+        } else {
+            return false; // some unknown construction - don't suspend
```
myk002, line 431: worth at least a DEBUG statement

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+
+        return true;
+    }
+
+    void suspendDeadend (color_ostream &out, df::job* job) {
+        auto building = Job::getHolder(job);
+        if (!building) return;
+        coord pos(building->centerx,building->centery,building->z);
+
+        for (size_t count = 0; count < 1000; ++count){
```
myk002, line 449: could you make 1000 a named constant?

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+    // gather some statistics about the last call to do_cycle()
+    string getStatus (color_ostream &out) {
+        std::map<Reason,int> stats;
+
+        for (auto [_ , reason] : suspensions) {
+            stats[reason == Reason::DEADEND ? Reason::RISK_BLOCKING : reason]++;
+        }
+
+        std::stringstream res;
+        res << "maintaining " << suspensions.size() << " suspension reasons\n";
```
myk002, line 531: how about "%zd jobs suspended"?

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+    }
+
+    void refresh(color_ostream &out)
+    {
+        DEBUG(cycle,out).print("starting refresh, prevent blocking is %s\n",
+                               prevent_blocking ? "true" : "false");
+        suspensions.clear();
+        leadsToDeadend.clear();
+
+        for (auto job : df::global::world->jobs.list) {
```
myk002, line 546: I'm still amazed that this exists. I'll have to start using it.

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+DFhackCExport command_result plugin_init(color_ostream &out, std::vector <PluginCommand> &commands) {
+    DEBUG(control,out).print("initializing %s\n", plugin_name);
+
+    suspendmanager_instance = std::make_unique<SuspendManager>();
+    eventhandler_instance = std::make_unique<EventManager::EventHandler>(jobCompletedHandler,1);
+
+    // provide a configuration interface for the plugin
+    commands.push_back(PluginCommand(
+        plugin_name,
+        "Automatically suspend and unsuspend constructions",
```
myk002, line 654:
```suggestion
        "Intelligently suspend and unsuspend jobs.",
```

myk002: should match summary in docs; also needs to end in a period.

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+    } else {
+        return CR_WRONG_USAGE;
+    }
+
+}
+
+static void jobCompletedHandler(color_ostream& out, void* ptr) {
+    DEBUG(cycle,out).print("job completed; updating suspensions\n");
+    df::job* job = static_cast<df::job*>(ptr);
+    if (SuspendManager::isConstructionJob(job))
```
myk002, line 762: or a smooth/carving job

plugins/suspendmanager.cpp
```diff
@@ -0,0 +1,822 @@
...
+    bool tryGetReason (df::job* job, Reason &reason) {
+        auto reason_it = suspensions.find(job->id);
+        if (reason_it != suspensions.end()) {
+                reason = reason_it->second;
+                return true;
+        }
+        return false;
+    }
+
+    std::map<int,Reason> suspensions;
```
plule, line 515: `std::unordered_map`/`std::unordered_set`? Given how often these are accessed, the performance bump could be significant

plugins/suspendmanager.cpp
chdoc: Completed smoothing/carving jobs, unlike new walls, do not risk blocking other jobs. So I think it's fine to not refresh immediately and wait for the next job- or time-triggered cycle.

plugins/suspendmanager.cpp
chdoc: yes, noticed after the push that I accidentally deleted that line just before pushing. Will reinstate that line with the next push. :sweat_smile:

plugins/suspendmanager.cpp
chdoc: I didn't think too hard about these, because they are expected to be small (<100 elements) unless someone is designating a mega-structure in one go. But if the consensus is that even for small maps `unordered_map` outperforms `map`, I'm happy to change that.

plule: Some benchmarks shows it being faster with as low as 5-10 items, I would not expect it to hurt in any case, and poooossibly make a perceptible difference in a mega-project going in the thousands of suspension. TBH the most likely scenario is that there is no perceptible difference either way, so not a big deal either way.

chdoc:
It would appear that MSCV really doesn't like the wrappers I added to `LuaTools.h`. I don't have MSVC. Help on this would be appreciated.

plugins/suspendmanager.cpp
chdoc: Already changed it. :+1:

plugins/lua/suspendmanager.lua
chdoc: Done. :heavy_check_mark:

plugins/suspendmanager.cpp
```diff
@@ -698,11 +714,14 @@ DFhackCExport command_result plugin_load_site_data (color_ostream &out) {
         DEBUG(control,out).print("no config found in this save; initializing\n");
         config = World::AddPersistentSiteData(CONFIG_KEY);
         config.set_bool(CONFIG_IS_ENABLED, is_enabled);
+        config.set_bool(CONFIG_PREVENT_BLOCKING, suspendmanager_instance->prevent_blocking);
```
myk002, line 717: this should initialize the new config data to a default, not whatever happens to be in the suspendmanager_instance from the previously loaded fort

docs/plugins/unsuspend.rst
myk002, line 1: plugin docs are are one file per plugin, not per command. see https://github.com/DFHack/dfhack/blob/develop/docs/plugins/reveal.rst for an example of a plugin doc that documents multiple commands

myk002 (approved):
we're in feature freeze for 50.12-r1 (assuming it will be out very soon). My plan is to merge this after we tag 50.12-r1 and quickly release 50.12-r2rc1 beta will all the new features we have piling up (see https://github.com/orgs/DFHack/projects/42/views/2)

edit: ok, I cleared things up so we can proceed with PR merges. See https://discord.com/channels/793331351645323264/1046628873397358602/1213654410962538536

chdoc:
@myk002 Did you merge this without a changelog entry? I had been meaning to ask how to categorize this?

myk002:
After some thought, I figured a changelog entry wasn't necessary since there was no player-visible change

chdoc:
Except that your FPS should no longer tank if you drop 100-200 build jobs in a busy fort on a low-end machine. I think it is safe to assume, even without proper performance testing, that the C++ implementation outperforms the Lua implementation. So there may be players who left this disabled until now and who may want to enable it now. Maybe add a misc. improvement line?

myk002:
that's fair. Misc improvement SGTM

PR #4319, opened 2024-03-02
Allow detaching loose building items

Bumber64:
Allows detaching workshop clutter, etc., so it can be autodumped.


myk002 (commented):
looks good in theory. Have you identified places in our code that we can simplify with these functions?

Bumber64:
> looks good in theory. Have you identified places in our code that we can simplify with these functions?

``detach_item`` already existed, so this just causes it not to fail for items in buildings. Looks like ``utils.getBuildingName`` does what the other function would've done, so I've reverted that.

myk002:
one place I'm aware of is `build-now`. It has custom logic for detaching clutter. That could be removed, I think.

docs/changelog.txt
```diff
@@ -90,6 +90,7 @@ Template for new versions:
 - ``Gui::makeAnnouncement``, ``Gui::autoDFAnnouncement``: don't display popup for all announcement types
 - ``gui.View:getMouseFramePos``: function now detects the correct coordinates even when the widget is nested within other frames
 - `strangemood`: correctly recognize Stonecutter and Stone Carver as moodable skills, move the Mason's boosted mood chance to the Stone Carver, and select Fell/Macabre based on long-term stress
+- Items module fn ``detach_item`` can now detach items loosely stored in buildings. Impacts ``autodump``, ``Items::moveToGround``, etc.
```
myk002, line 93: needs to get moved up to "Future"

myk002 (commented):
could you remove the now redundant code in `build-now` to take advantage of this?

myk002 (approved):

PR #4374, opened 2024-03-16
Add CONTEXT_SCROLL keys to widgets.lua STANDARDSCROLL

Bumber64:
I noticed that mouse scrolling wasn't being detected in an overlay I'm working on. These are being received as `CONTEXT_SCROLL_*`, which aren't in `widgets.lua`'s `STANDARDSCROLL` table.

I'm using `STANDARDSCROLL` in my overlay's `onInput` rather than a widget. It could be that it's not necessary for the widgets that use it, but it would be good to have them there unless there's a good reason not to.


myk002:
`STANDARDSCROLL` is intended for keyboard navigation. `CONTEXT_SCROLL_*` scrolls the view without changing the cursor position, which is why it isn't in `STANDARDSCROLL`.

`CONTEXT_SCROLL_*` is handled here: https://github.com/DFHack/dfhack/blob/develop/library/lua/gui/widgets.lua#L1072

Bumber64:
> For context, what is it that you need in your `onInput` that a standard widget doesn't handle?

I guess I can use the Scrollbar widget, though it doesn't indicate it's from DFHack since it's not in a panel.

PR #4380, opened 2024-03-17
[WIP] Add graphics support for building-hacks

warmist:
This intends to reenable building-hacks namely machine workshops with animations.

Things to consider:

- [ ] general "is this useful to anyone else"?
- [ ] doc update not done yet
- [ ] redo the api as it sucks. Lua part simplifies the api a little...
- [ ] how do you get tile ids (ala raws) in lua (i.e. how to make the api work?)


plugins/building-hacks.cpp
```diff
@@ -260,9 +263,9 @@ struct work_hook : df::building_workshopst{
         }
         INTERPOSE_NEXT(updateAction)();
     }
-    DEFINE_VMETHOD_INTERPOSE(void, drawBuilding, (df::building_drawbuffer *db, int16_t unk))
+    DEFINE_VMETHOD_INTERPOSE(void, drawBuilding, (int32_t unk1,df::building_drawbuffer *db, int16_t unk2))
```
myk002, line 266: `unk1` is `curtick`, `unk2` is `bz`

plugins/building-hacks.cpp
```diff
@@ -289,15 +292,22 @@ struct work_hook : df::building_workshopst{
...
+                    db->back[tx][ty]= cf.back;
+                    db->bright[tx][ty]= cf.bright;
+                    db->fore[tx][ty]= cf.fore;
                 }
+                if (cf.graphics_tile >= 0)
+                    db->texpos1[tx][ty] = cf.graphics_tile;
+                if (cf.overlay_tile >= 0)
+                    db->texpos2[tx][ty] = cf.overlay_tile;
+                if (cf.unk_tile >= 0)
+                    db->texpos3[tx][ty] = cf.unk_tile;
```
myk002, line 310: let me update the structures so these are named the same as the fields in `enabler`

edit: https://github.com/DFHack/df-structures/pull/737

myk002:
Can the plugin automatically enable itself if it detects that it is used in the raws, the same way that `add-spatter` does? That way the player won't have to manually enable it to get things working.

> * [ ]  general "is this useful to anyone else"?

as useful as `add-spatter`, which I'm also happy to include

> * [ ]  how do you get tile ids (ala raws) in lua (i.e. how to make the api work?)

it depends which tile ids you're getting. will it be a vanilla texpos or one added by DFHack?

warmist:
> Can the plugin automatically enable itself if it detects that it is used in the raws, the same way that `add-spatter` does? That way the player won't have to manually enable it to get things working.

It depends: for basic functionality it should be possible. It can be split in a few different things that this plugin does:

1. Impassable fix - this might be already in df? Need to check. Basically fixes that workshop impassible tiles would allow water through.
2. Generate power - unlikely that it's useful. Usually need some sort of "when fuel exists/etc" and probably used from lua
3. Consume power - probably useful straight from raws, but it needs an argument (i.e. how much power do you need)
4. animation - have no idea how to do it from raws
5. canBeRoomSubset - ugh...

> > * [ ]   how do you get tile ids (ala raws) in lua (i.e. how to make the api work?)
> 
> it depends which tile ids you're getting. will it be a vanilla texpos or one added by DFHack?

It's vanilla texpos but e.g. you do:
```
graphics_dragon_engine

[OBJECT:GRAPHICS]

[TILE_GRAPHICS:DRAGON_ENGINE_TILES:0:0:WORKSHOP_CUSTOM:DRAGON_ENGINE_N:3:0:0]
[TILE_GRAPHICS:DRAGON_ENGINE_TILES:0:1:WORKSHOP_CUSTOM:DRAGON_ENGINE_N:3:0:1]
[TILE_GRAPHICS:DRAGON_ENGINE_TILES:0:2:WORKSHOP_CUSTOM:DRAGON_ENGINE_N:3:0:2]
[TILE_GRAPHICS:DRAGON_ENGINE_TILES:0:3:WORKSHOP_CUSTOM:DRAGON_ENGINE_N:3:0:3]
[TILE_GRAPHICS:DRAGON_ENGINE_TILES:0:4:WORKSHOP_CUSTOM:DRAGON_ENGINE_N:3:0:4]
```
and then you need yet another tile for animated part. Or you need to do something similar in the lua part.

myk002:
`add_spatter` looks for `SPATTER_ADD_` in any reaction. Could this look for a similar token prefix across the loaded building raws?

for looking up the graphics tile from a loaded page, I think you'd use `dfhack.screen.findGraphicsTile(pagename,x,y)`\
that's slow, though, so you'd probably want to cache it and wipe your cache on a map load event.

warmist:
So api is still terrible but less so. The rest of stuff doesn't matter that much, so i think i'll add (or at least try to add) automagical way of this loading stuff from raws and auto-enabling itself. In perfect case having a line `[DFHACK_MACHINE:25]` would make any workshop to have power drain of 25 to work. This ofc is not perfect (e.g. how about interface, what if you want to provide power, etc...) but it would lower cost of entry for a lot of modders.

Another issue that plagues this module for many years: buildings need to built in specific order to connect to machines...

plugins/building-hacks.cpp
```diff
@@ -323,50 +338,71 @@ void clear_mapping()
...
-        lua_pushnil(L);
-        std::vector<graphic_tile> frame;
-        while (lua_next(L, -2) != 0) {
-            graphic_tile t;
-            lua_pushnumber(L,1);
-            lua_gettable(L,-2);
-            if(lua_isnil(L,-1))
+
+    int frame_index = 1;
+
```
myk002, line 346: plz fix lint errors

plugins/lua/building-hacks.lua
```diff
@@ -15,19 +15,30 @@ local _ENV = mkmodule('plugins.building-hacks')
             action -- a table of number (how much ticks to skip) and a function which gets called on shop update
             canBeRoomSubset -- room is considered in to be part of the building defined by chairs etc...
```
myk002, line 16: rooms no longer exist. it's all zones now. the implementation may not need to change, but the docs should be updated

warmist:
This is now relatively done. Needs more testing.

The "make it work from raws" is a separate project and probably lua only.

warmist:
Ok, this is as done as I currently can make it be.

plugins/building-hacks.cpp
```diff
@@ -168,7 +180,8 @@ struct work_hook : df::building_workshopst{
     }
     DEFINE_VMETHOD_INTERPOSE(void, categorize, (bool free))
```
myk002, line 181: can this get run on world load before scripts have a chance to register custom workshop types?

plugins/building-hacks.cpp
```diff
@@ -20,47 +20,56 @@
...
 #include <map>
-
+#include <limits>
 using namespace DFHack;
 using namespace df::enums;
 
 DFHACK_PLUGIN("building-hacks");
 REQUIRE_GLOBAL(world);
 
+constexpr uint32_t invalid_tile = std::numeric_limits<uint32_t>::max();
```
myk002, line 34: other parts of the code (including DF) uses 0 for "no tile". Would that make sense here, or does 0 have another meaning?

plugins/building-hacks.cpp
```diff
@@ -20,47 +20,56 @@
...
 struct graphic_tile //could do just 31x31 and be done, but it's nicer to have flexible imho.
 {
-    int16_t tile; //originally uint8_t but we need to indicate non-animated tiles
+    int16_t tile=-1; //originally uint8_t but we need to indicate non-animated tiles
     int8_t fore;
     int8_t back;
     int8_t bright;
+    //index of texpos
+    uint32_t graphics_tile = invalid_tile;
```
myk002, line 42: the texpos vectors you will be writing these into are `int32_t`. why the type difference?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
 
 Functions
 ---------
 
-``registerBuilding(table)`` where table must contain name, as a workshop raw name, the rest are optional:
+* ``setOwnableBuilding(workshop_type, enable)``
 
-    :name:
-        custom workshop id e.g. ``SOAPMAKER``
+  Set workshop to be included in zones (such as bedroom or inn).
```
myk002, line 5780: Why would you choose to include the workshop in zones? What does the choice affect?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
 
-        .. note:: this is the only mandatory field.
+  :workshop_type:   custom workshop string id e.g. ``SOAPMAKER`` or numeric id
```
myk002, line 5782: it might be worth mentioning, maybe in an intro to the Functions section, that numeric ids can change between worlds and a script should only pass an ID if it has looked it up for this custom workshop type in this world.

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+  :enable:   boolean value to enable or disable this functionality
```
myk002, line 5783: can this be usefully dynamically toggled? what happens if a workshop type that *was* ownable is now not ownable? Does it make sense to have this parameter at all, or should this call just be to flip the value from false (the default) to true?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
-        c. ``{x=<number> y=<number> + 4 numbers like in first case}``,
-           this generates full frame useful for animations that change little (1-2 tiles)
+  Set workshop non walkable tiles to also block liquids (i.e. water and magma).
 
-    :canBeRoomSubset:
-        a flag if this building can be counted in room. 1 means it can, 0 means it can't and -1 default building behaviour
-    :auto_gears:
-        a flag that automatically fills up gears and animations. It looks over the building definition for gear icons and maps them.
+  :workshop_type:   custom workshop string id e.g. ``SOAPMAKER`` or numeric id
+  :enable:   boolean value to enable or disable this functionality
```
myk002, line 5790: should this be enabled per workshop *type* or per workshop type and *state*? If not, and it's just by type, then does it make sense to dynamically toggle? Similar question to setOwnableBuilding above.

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
 
-    Animate table also might contain:
+* ``setMachineInfo(workshop_type, needs_power, power_consumed, power_produced, connection_points)``
```
myk002, line 5792: do we need a separate `needs_power` parameter? Can this be combined with `power_consumed`? e.g. if `power_consumed` is > 0, then the building needs power? Or is there a case where the building still works even if it is without needed power?

maybe what I'm not understanding is: how do the values specified here interact with `setPower`? would it make sense to remove the `power_consumed` and `power_produced` parameters from this function and require that `setPower` be called separately?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
-        a bool that says to try to match to mechanical system (i.e. how gears are turning)
+  Setup and enable machine-like functionality for the workshop. Note: due to implementation limitations
+  workshop only connects to other machines if the other machine is build later than this one.
 
-``getPower(building)`` returns two number - produced and consumed power if building can be modified and returns nothing otherwise
+  :workshop_type:   custom workshop string id e.g. ``SOAPMAKER`` or numeric id
+  :needs_power:   only function if it has sufficient power
+  :power_consumed:   building consumes this amount of power
+  :power_produced:   output this amount of power
+  :connection_points:   a table of ``{x=?,y=?}`` points that can connect to other machines
```
myk002, line 5801: should document that it is zero-based indexing

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
 
-``setPower(building,produced,consumed)`` sets current power production and consumption for a building.
+* ``setMachineInfoAuto(workshop_type, needs_power, power_consumed, power_produced, gear_tiles)``
+
+  Same as ``setMachineInfo`` but fills out the ``connection_points`` table from raws placing connection
+  points on tiles which have the gear tile. ``gear_tiles`` is an optional array of two tiles that are
```
myk002, line 5806: why two tiles? are these the two symbols displayed in the frames of animation? Why are both required to be specified? or does this parameter really mean "any of these symbols on any frame"

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+
+  Animate workshop by replacing displayed tiles (or graphical tiles). There are two ways this works:
+  if ``frame_skip>=0`` then it shows each frame for ``frame_skip`` of frames or if ``frame_skip<0``
+  Frames are synchronized to the machine this building is connected to.
+
+  :workshop_type:   custom workshop string id e.g. ``SOAPMAKER`` or numeric id
+  :frames:          table of frames. Each frame is sparse flat table with ids from ``0`` to ``31*31-1``.
+                    Each frame tile is table of integers from 4 to 8 members long. Tile members are as
+                    follow: tile, foreground color, background color, bright, graphics tile, overlay tile,
+                    signpost tile, item tile.
```
myk002, line 5819: you might need to explain what each graphics tile should be used for

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+  :connection_points:   a table of ``{x=?,y=?}`` points that can connect to other machines
 
-``setPower(building,produced,consumed)`` sets current power production and consumption for a building.
+* ``setMachineInfoAuto(workshop_type, needs_power, power_consumed, power_produced, gear_tiles)``
+
+  Same as ``setMachineInfo`` but fills out the ``connection_points`` table from raws placing connection
+  points on tiles which have the gear tile. ``gear_tiles`` is an optional array of two tiles that are
+  counted as gears in the workshop ascii tile raws.
+
+* ``setAnimationInfo(workshop_type, frames, frame_skip)``
```
myk002, line 5809: optional parameters need to be surrounded in square brackets, e.g. `setAnimationInfo(workshop_type, frames[, frame_skip])`

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+                    signpost tile, item tile.
+  :frame_skip:      How many ticks to display one frame. If set to negative number (or skipped) frames
+                    are synchronized with machine animation.
+
+* ``setAnimationInfoAuto(name, make_graphics_too, frame_skip, gear_tiles)``
+
+  Animate workshop as with function above but generate frames automatically. This works by finding
+  tiles which have gears and animating them with alternating gear tiles.
+
+  :name:            custom workshop string id e.g. ``SOAPMAKER``
```
myk002, line 5828: why the switch from `workshop_type` to `name`?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+  :make_graphics_too:  replace same tiles in graphics mode with tiles from vanilla df mechanism
+  :frame_skip:      How many ticks to display one frame. If set to negative number (or skipped) frames
+                    are synchronized with machine animation.
+  :gear_tiles:      Optional array of 2 or 4 indexes. First two define ascii tiles and next two graphics tiles
```
myk002, line 5832: is this for overriding the default gear graphics? If so, that should be communicated.

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+
+* ``setOnUpdate(name,interval,callback)``
+
+  Setup callback to be called every ``interval`` of ticks for each building of this type. Note: low interval
+  numbers and/or many workshops that use this might reduce DF performance.
+
+  :name:   custom workshop string id e.g. ``SOAPMAKER``
+  :interval:   how many ticks to skip between event triggers
+  :callback:   function to call
```
myk002, line 5841: the callback signature needs to be defined

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,129 @@ The names of the functions are also available as the keys of the
...
+  :building:   specific workshop that produces or consumes power
+  :power_consumed:   set building to consume this amount of power
+  :power_produced:   output this amount of power
+
+
+Events
+------
+
+This module exports two events. However only one is documented here and is intended to be used directly. To use
+``onUpdateAction`` instead call ``setOnUpdate`` function.
```
myk002, line 5863: why does onUpdateAction exist as a Lua event, then?

myk002 (commented):
reviewed docs and lifecycle code; will review actual code once some questions are answered.

- could you add a changelog entry for the update?
- could you update docs/plugins/building-hacks.rst and replace the "unavailable" tag with "fort gameplay buildings". Maybe "adventure fort gameplay buildings" if it's intended to work in adventure mode too.
- we need a full example of the RAWs and scripts for how to make this all work. It could be added to the modding guide as a case study. You could link to the modding guide from both the Lua API docs and buiding-hacks.rst.

docs/dev/Lua API.rst
warmist: The plugin lua part still uses that event (i.e. `setOnUpdate` function)

plugins/building-hacks.cpp
```diff
@@ -20,47 +20,56 @@
...
-    bool machine_timing; //6 frames used in vanilla
+    bool machine_timing=false; //6 frames used in vanilla
     int frame_skip; // e.g. 2 means have to ticks between frames
     //updateCallback:
-    int skip_updates;
-    int room_subset; //0 no, 1 yes, -1 default
+    int skip_updates=0;
+    int room_subset=-1; //0 no, 1 yes, -1 default
 };
 typedef std::map<int32_t,workshop_hack_data> workshops_data_t;
```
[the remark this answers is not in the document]
warmist: Unsure on benefits for int32_t keys. Though map is just because it was like this before fixes.

plugins/building-hacks.cpp
warmist: Ugh this is a difficult issue... Lua script enabling this workshop might not have run yet as they usually run on world load.

plugins/building-hacks.cpp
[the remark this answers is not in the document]
myk002: it would be an O(1) lookup instead of an O(log(n)). however, for small numbers of registered workshop types, it's unlikely to matter much.

docs/dev/Lua API.rst
warmist: It allows for interesting systems: one example i had (before big gui changes etc...) was a personal dwarven hobby workshop that would be built in owned rooms. That way you could query what dwarf owns what buildings (e.g. cabinet, bet, modded-hobby-workshop) and issue jobs that would consume resources, give nothing back, but still level skills. Intended to work as a "do this while dwarf is idle".

docs/dev/Lua API.rst
warmist: yeah probably there is no reason to call this with false

docs/dev/Lua API.rst
warmist: ditto. Will remove the enable

docs/dev/Lua API.rst
warmist: `needs_power` is a bit misnamed argument. It make it function like a vanilla millstone: you can't use it while it's unpowered.

warmist: as for `power_consumed/produced` yeah it probably could be removed and require `setPower` to be called. Will make the api cleaner

warmist: Ah i just tried removing `power_consumed/produced` thingy and remembered: `setPower` sets power for one specific building and the `setMachineInfo` sets for all buildings of that type. Though it's not intended to be called more than once.

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
-    :animate:
-        a table of frames which can be a table of:
+  Animate workshop as with function above but generate frames automatically. This works by finding
+  tiles which have gears and animating them with alternating gear tiles.
 
-        a. tables of 4 numbers ``{tile,fore,back,bright}`` OR
-        b. empty table (tile not modified) OR
-        c. ``{x=<number> y=<number> + 4 numbers like in first case}``,
-           this generates full frame useful for animations that change little (1-2 tiles)
+  :workshop_type:     custom workshop string id e.g. ``SOAPMAKER`` or numeric id
```
myk002, line 5831:
```suggestion
  :workshop_type:     custom workshop string id, e.g. ``SOAPMAKER`` or numeric id
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+  :make_graphics_too:  replace same tiles in graphics mode with tiles from vanilla df mechanism
```
myk002, line 5832: when would you want this set to false?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+  :frame_length:      How many ticks to display one frame. If set to negative number (or skipped) frames
+                      are synchronized with machine animation.
+  :gear_tiles:        Optional array of 2 or 4 indexes. First two define ascii tiles and next two graphics tiles.
+                      This overrides default gear tiles.
```
myk002, line 5836: a map might be more instructive instead of a variable length list, e.g.:
```lua
local bhacks = require('plugins.building-hacks')
bhacks.setAnimationInfoAuto('DRAGON_ENGINE', true, 4, {
    gear_ch=42, -- looks for this in ascii
    gear_ch_alt=15,  -- replaces with this for animated frames
    gear_tile=dfhack.screen.findGraphicsTile('WARMIST_DRAGON_ENGINE', 1, 0),
    gear_tile_alt=dfhack.screen.findGraphicsTile('WARMIST_DRAGON_ENGINE', 4, 0),
})`

File: docs/dev/Lua API.rst [hunk 25]
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+                    First 4 are function same as ascii workshop definition. The latter 4 are graphics
+                    layers. ``signpost tile`` is an optional row that sticks up over the workshop.
 
-    :name:
-        custom workshop id e.g. ``SOAPMAKER``
+  :frame_skip:      How many ticks to display one frame. If set to negative number (or skipped) frames
+                    are synchronized with machine animation.
 
-        .. note:: this is the only mandatory field.
+* ``setAnimationInfoAuto(workshop_type, make_graphics_too, [frame_length], [gear_tiles])``
```
myk002, line 5826:
```suggestion
* ``setAnimationInfoAuto(workshop_type, make_graphics_too[, frame_length[, gear_tiles]])``
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+  :frame_length:      How many ticks to display one frame. If set to negative number (or skipped) frames
+                      are synchronized with machine animation.
+  :gear_tiles:        Optional array of 2 or 4 indexes. First two define ascii tiles and next two graphics tiles.
+                      This overrides default gear tiles.
 
-    :canBeRoomSubset:
-        a flag if this building can be counted in room. 1 means it can, 0 means it can't and -1 default building behaviour
-    :auto_gears:
-        a flag that automatically fills up gears and animations. It looks over the building definition for gear icons and maps them.
+* ``setOnUpdate(workshop_type,interval,callback)``
```
myk002, line 5838:
```suggestion
* ``setOnUpdate(workshop_type, interval, callback)``
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
 
-    Animate table also might contain:
+  Setup callback to be called every ``interval`` of ticks for each building of this type. Note: low interval
+  numbers and/or many workshops that use this might reduce DF performance.
```
myk002, line 5841:
```suggestion
  Register callback to be called every ``interval`` ticks for each building of this
  type. This can be very expensive if the interval is low and/or there are many
  workshops of this type. Keep these callbacks light!
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
 
-    :frameLength:
-        how many ticks does one frame take OR
-    :isMechanical:
-        a bool that says to try to match to mechanical system (i.e. how gears are turning)
+  :workshop_type:   custom workshop string id e.g. ``SOAPMAKER`` or numeric id
```
myk002, line 5843:
```suggestion
  :workshop_type:   custom workshop string id, e.g. ``SOAPMAKER`` or numeric id
```

docs/dev/Lua API.rst
myk002: you may want to explain the difference between frame ticks and world simulation ticks (and note which one this refers to).

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+  :interval:        how many ticks to skip between event triggers
+  :callback:        function to call. Function signature is ``func(workshop)`` where ``workshop`` is of type
+                    ``df.building_workshopst``
 
-``getPower(building)`` returns two number - produced and consumed power if building can be modified and returns nothing otherwise
+* ``getPower(building)``
 
-``setPower(building,produced,consumed)`` sets current power production and consumption for a building.
+  Returns two number - produced and consumed power if building can be modified and returns nothing otherwise.
```
myk002, line 5850:
```suggestion
  If this building is of a type registered with building-hacks, returns values for
  produced and consumed power. Otherwise, returns ``nil``.
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+
+  :building:   specific workshop that produces or consumes power
+
+* ``setPower(building, power_consumed, power_produced)``
+
+  Sets current power production and consumption for a specific workshop building. Can be used to make buildings that
+  dynamically change power consumption and production.
```
myk002, line 5857:
```suggestion
  Dynamically sets current power production and consumption for a specific workshop
  (which must be of a type registered with building-hacks).
```

myk002: I note that the order of consumed and produced power is in the opposite order compared to `setPower`. They should match.

docs/dev/Lua API.rst
myk002: I'm not convinced this is sufficient reasoning. Why do you need a lua event to communicate with your own Lua layer? Why can't you call the Lua functions directly?

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+  :power_produced:   output this amount of power
+
+
+Events
+------
+
+This module exports two events. However only one is documented here and is intended to be used directly. To use
+``onUpdateAction`` instead call ``setOnUpdate`` function.
+
+* ``onSetTriggerState(workshop,state)``
```
myk002, line 5870:
```suggestion
* ``onSetTriggerState(workshop, state)``
```

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
 
-  require('plugins.building-hacks').registerBuilding{name="BONE_GRINDER",
-    consume=15,
-    gears={x=0,y=0}, --connection point
-    animate={
-      isMechanical=true, --animate the same conn. point as vanilla gear
-      frames={
-      {{x=0,y=0,42,7,0,0}}, --first frame, 1 changed tile
-      {{x=0,y=0,15,7,0,0}} -- second frame, same
+  local bld=require('plugins.building-hacks')
```
myk002, line 5882:
```suggestion
  local bhacks = require('plugins.building-hacks')
```

myk002: `bld` is consistently used throughout DFHack to refer to "current building". using the name to refer to the module would be confusing

myk002 (commented):
needs changelog entry

I added some code examples in my comments to show what my understanding is of how these functions are to be used. If the examples are wrong, then I didn't understand something important.

myk002:
how is this coming along? I'm looking forward to using dragon fire spitters in my base defense : )

ab9rf:
What's the status on this? I got an inquiry about `building-hacks` from the kitfox discord and so there is interest in bringing this bird back to life if it's a possibility.

warmist:
> What's the status on this? I got an inquiry about `building-hacks` from the kitfox discord and so there is interest in bringing this bird back to life if it's a possibility.

I'm stuck in factorio land, but it's functionally done but needs to address Myk002's review issues. Also I should upload example project with everything.

There might be something broken by struct changes or something else. I'll try to find some time this week to look into this.

docs/dev/Lua API.rst
```diff
@@ -5766,82 +5766,134 @@ The names of the functions are also available as the keys of the
...
+
+  Same as ``setMachineInfo`` but fills out the ``connection_points`` table from raws placing connection
+  points on tiles which have the gear tile. ``gear_tiles`` is an optional array of two tiles that are
+  counted as gears in the workshop ascii tile raws. The default gear tiles are ``42`` and ``15``.
+
+* ``setAnimationInfo(workshop_type, frames, frame_skip)``
+
+  Animate workshop by replacing displayed tiles (or graphical tiles). There are two ways this works:
+  if ``frame_skip>=0`` then it shows each frame for ``frame_skip`` of frames or if ``frame_skip<0``
+  Frames are synchronized to the machine this building is connected to.
```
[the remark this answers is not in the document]
warmist: Machine in this context is df::machine. I.e. A collection of connected buildings. It has it's own timing (visual_phase). IIRC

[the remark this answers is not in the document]
myk002: that might be too technical for the docs -- maybe "Frames are synchronized with the machines this building is connected to." would be more understandable by players

docs/dev/Lua API.rst
```diff
@@ -6520,87 +6520,138 @@ The names of the functions are also available as the keys of the
...
+  Animate workshop by replacing displayed tiles (or graphical tiles). There are two ways this works:
+  if ``frame_skip>0`` then it shows each frame for ``frame_skip`` of frames or if ``frame_skip<=0``
+  Frames are synchronized with the machines this building is connected to.
+
+  :workshop_type:   custom workshop string id, e.g. ``SOAPMAKER`` or numeric id
+  :frames:          table of frames. Each frame is array of rows of tiles with ids from ``1`` to ``31``.
+                    Each frame tile is table of with following optional members: ``ch``, ``fg``,
+                    ``bg``, ``bold``, ``tile``, ``tile_overlay``, ``tile_signpost``, ``tile_item``.
+                    First 4 are function same as ascii workshop definition. The latter 4 are graphics
+                    layers. ``tile_signpost `` is only valid in first row and it shows up above the workshop.
```
ab9rf, line 6577: remove space before \`\`

PR #4389, opened 2024-03-20
Add mechanism unlinking functionality to buildingplan plugin

Bumber64:
Adds ability to unlink and free mechanisms from buildings.

Replaces DFHack/scripts#1016


plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,678 @@
...
+    end
+
+    self.frame.h = self.num_buttons*3 + 2
+
+    if self.frame.h < 11 then --space between list and "free all" is removed
+        self.frame.h = self.frame.h - 1
+    end
+
+    self.frame.b = 4
+    self.frame.r = 40
```
Bumber64, line 658: I haven't done a lot of testing on various screen resolutions. I used a 1600x900 window with `Interface scaling percentage` at 100%, 120%, 130%, and 140%. That's 4 data points, so I don't know if these values are precise.

Messing with `Maximum interface percentage` breaks horizontal layout. I don't think any of our overlays handle this properly. (Work order conditions overlay, for instance.)

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,678 @@
...
+        widgets.HotkeyLabel
+        {
+            frame = {b=0, r=3},
+            key = "CUSTOM_SHIFT_U",
+            label = "Unlink all",
+            auto_width = true,
+            on_activate = self:callback("ask_unlink_all"),
+            enabled = function() return next(self.links) end,
+        },
+        widgets.Scrollbar --Work around for https://dwarffortressbugtracker.com/view.php?id=12721
```
Bumber64, line 257: There's no indication this is added by DFHack.

plugins/lua/buildingplan/unlink_mechanisms.lua
myk002: > I used a 1600x900 window with Interface scaling percentage at 100%, 120%, 130%, and 140%. That's 4 data points, so I don't know if these values are precise.

In my experience, we handle interface scaling without issues.

> Messing with Maximum interface percentage breaks horizontal layout. I don't think any of our overlays handle this properly.

very very true. this has been a known (to me) deficiency for a long time (since v50 introduced the concept of maximum interface percentage). I think we can solve it centrally, in `overlay.lua`, adjusting the position of the widgets according to the interface percentage.

Bumber64: > > I used a 1600x900 window with Interface scaling percentage at 100%, 120%, 130%, and 140%. That's 4 data points, so I don't know if these values are precise.
> 
> In my experience, we handle interface scaling without issues.

I didn't know the extra space was going to disappear until I ran into it. I don't know that I can test arbitrary values of `parent_height`, so I can't be certain that the space doesn't disappear slightly earlier than `self.frame.h < 11`.

`(self.parent_height - 49) // 3` also fits my data points, but I don't know if that makes a difference.

myk002:
This is what I see, which is a bit more unstable than I'd like:

https://github.com/DFHack/dfhack/assets/977482/4a783d48-d0da-4b76-9f48-3abe29482c74

Bumber64:
> This is what I see, which is a bit more unstable than I'd like:

Should be fixed.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,682 @@
...
+            button.visible = false
+        end
+    end
+
+    local h = self.parent_height - 46 - (has_link_tab(self.building) and 3 or 0)
+    self.frame.h = h + 1 --includes lower border
+    self.num_buttons = h // 3
+
+    self.frame.b = 3
+    self.frame.r = 40
```
Bumber64, line 662: Probably shouldn't need to do this, but for some reason these both end up as 1 otherwise. Not sure why.

Bumber64:
Looks like all `OverlayWidgets` use `default_pos` instead of `frame` t/b/l/r even if they shouldn't be reposition-able? This generates a `pos` value in `overlay.json` which can be edited to put the overlay in the wrong place. More problematic is if we have to change the default in the future (e.g., because UI changes,) it's going to keep using the old value in the json.

We probably shouldn't save positions for `OverlayWidgets` with `overlay_only = true`.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,680 @@
...
+
+function MechItemOverlay:get_button(n, ensure)
+    local button = self.subviews["free_"..n]
+    if not button and ensure then
+        self:addviews
+        {
+            widgets.TextButton
+            {
+                view_id = "free_"..n,
+                frame = {t=0, l=28, w=6, h=1},
```
myk002, line 539: I'd move this to the right two or three tiles to give longer names a little more room:
image

plugins/lua/buildingplan/unlink_mechanisms.lua
myk002: An added scrollbar isn't too worrisome. I added a similar scrollbar to the embark unit list for when `startdwarf` gives you more dwarves than will fit on the screen.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,680 @@
...
+
+    for item in mech_iter(b) do
+        if get_trigger_index(item) then
+            return true
+        end
+    end
+end
+
+local function unlink_building(b, m) --clean up extra stuff
+    if b._type == df.building_trapst then
```
myk002, line 57: we normally write this `df.building_trapst:is_instance(b)`, but this is probably ok (as long as `b` cannot be `nil`)

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -0,0 +1,680 @@
...
+        for k, v in ipairs(b.linked_mechanisms) do
+            if v == m then
+                b.linked_mechanisms:erase(k)
+                break
+            end
+        end
+    elseif b._type == df.building_doorst or b._type == df.building_hatchst then
+        local p = utils.getBuildingCenter(b)
+        local mb = dfhack.maps.getTileBlock(p)
```
myk002, line 66: no need to change for this PR, but the more common approach here is
```lua
local flags, occ = dfhack.maps.getTileFlags(p)
if occ then
    ...
    occ.building = df.tile_building_occ.Dynamic
```

myk002 (approved):
This looks great! Thanks!

plugins/lua/buildingplan/unlink_mechanisms.lua
Bumber64: There can be an extra button on that spot, IIRC.

Bumber64: free
:(

PR #4395, opened 2024-03-21
[tailor] allow turning off automatic confiscation of tattered clothing

chdoc:


ab9rf (approved):

docs/plugins/tailor.rst
```diff
@@ -24,17 +24,23 @@ Usage
...
+used, and in what order. By default, ``tailor`` will "confiscate" (i.e., remove
+ownership and mark for dumping) equipped tattered clothing. This can be changed
+using ``tailor confiscate``. Note that not confiscating tattered clothing may
+cause dwarves to get unhappy thoughts for wearing worn clothing before deciding
+to equip available new clothes.
+
+``tailor`` supports adamantine cloth (using the keyword ``adamantine``) but does
+not use it by default, as most players find adamantine too precious to routinely
+make into cloth. ``tailor`` does not support modded "cloth" types which utilize
+custom reactions to making clothing out of those cloth types.
```
myk002, line 43:
```suggestion
custom reactions for making clothing out of those cloth types.
```

docs/plugins/tailor.rst
```diff
@@ -24,17 +24,23 @@ Usage
...
 
-but you can use the ``tailor materials`` command to restrict which materials
-are used, and in what order. ``tailor`` supports adamantine cloth (using the
-keyword ``adamantine``) but does not use it by default, as most players find
-adamantine too precious to routinely make into cloth. ``tailor`` does not
-support modded "cloth" types which utilize custom reactions to making clothing
-out of those cloth types.
+but you can use the ``tailor materials`` command to restrict which materials are
+used, and in what order. By default, ``tailor`` will "confiscate" (i.e., remove
+ownership and mark for dumping) equipped tattered clothing. This can be changed
```
myk002, line 35: I thought it only confiscated equipped tattered clothing when there are replacements available.

docs/plugins/tailor.rst
```diff
@@ -24,17 +24,23 @@ Usage
...
+using ``tailor confiscate``. Note that not confiscating tattered clothing may
```
myk002, line 36: I don't think this is enough information for the player to decide whether they want to disable confiscation. Why would they choose to disable it?

plugins/lua/tailor.lua
```diff
@@ -37,6 +38,11 @@ function setDebugMode(opt)
     tailor_setDebugFlag(fl)
 end
 
+function setConfiscate(opt)
+    local fl = (opt[1] == "true" or opt[1] == "on")
```
myk002, line 42: please use `argparse.toBool` for standardized behavior here

plugins/tailor.cpp
```diff
@@ -109,6 +110,7 @@ static const MatType
 static const std::list<MatType> all_materials = { M_SILK, M_CLOTH, M_YARN, M_LEATHER, M_ADAMANTINE };
 static const std::list<MatType> default_materials = { M_SILK, M_CLOTH, M_YARN, M_LEATHER }; // adamantine not included by default
 static std::list<MatType> material_order = default_materials;
+static bool confiscate = true;
```
myk002, line 113: I think it's better to not cache the value in a local static and instead pull the value from the source of truth when needed with `config.get_bool(CONFIG_CONFISCATE)`. Then we don't need to worry about the local copy getting out of sync

plugins/tailor.cpp
```diff
@@ -666,9 +668,10 @@ DFhackCExport command_result plugin_load_site_data (color_ostream &out) {
         DEBUG(control,out).print("no config found in this save; initializing\n");
         config = World::AddPersistentSiteData(CONFIG_KEY);
         config.set_bool(CONFIG_IS_ENABLED, is_enabled);
+        config.set_bool(CONFIG_CONFISCATE, true);
     }
-
     is_enabled = config.get_bool(CONFIG_IS_ENABLED);
+    confiscate = config.get_bool(CONFIG_CONFISCATE);
```
myk002, line 674: this will cause all existing saves to disable confiscation. it's much safer to have your boolean state default to false. Maybe rename it `CONFIG_NO_CONFISCATE`

plugins/tailor.cpp
ab9rf: any instance-specific persistent data should be stored as a (regular) member of `class Tailor`. see, for example, the `inventory_sanity_checking` control parameter (which controls an undocumented debugging feature). the only global static in the tool should be the `std::unique_ptr<Tailor> tailor_instance`

as an aside, `static std::list<MatType> material_order = default_materials;` (line 111 or thereabouts) should be made a nonstatic member of `class Tailor` rather than a global static

chdoc:
Great, most of the feedback is about doing the same as the surrounding preexisting code. :man_shrugging: 
I will try to make a pass later today.

plugins/tailor.cpp
```diff
@@ -663,12 +671,20 @@ DFhackCExport command_result plugin_load_site_data (color_ostream &out) {
         DEBUG(control,out).print("no config found in this save; initializing\n");
         config = World::AddPersistentSiteData(CONFIG_KEY);
         config.set_bool(CONFIG_IS_ENABLED, is_enabled);
+        config.set_bool(CONFIG_CONFISCATE, true);
+    }
+    // transition existing saves to CONFIG_CONFISCATE=true
+    if (config.get_int(CONFIG_CONFISCATE) < 0) {
+        INFO(control,out).print("found existing configuration with CONFIG_CONFISCATE unset, initializing to true\n");
```
myk002, line 678:
```suggestion
        DEBUG(control,out).print("found existing configuration with CONFIG_CONFISCATE unset, initializing to true\n");
```

myk002: most users are not likely to care about the migration

myk002 (approved):

PR #4398, opened 2024-03-23
display caged creatures when placing cage

Hiranus:


myk002 (commented):
I think this is needed for wider use so DFHack can have a standard way of formatting item descriptions. Would you be comfortable porting this entire function (including `get_artifact_name`) to C++ and making it available as `dfhack.items.getReadableDesc()`?

myk002:
ported to C++ in https://github.com/DFHack/dfhack/pull/4521

PR #4401, opened 2024-03-24
new Lua module: enable-util

chdoc:
This is obviously not yet ready to be merged. However, I would like to know what you think of the idea, and what would you needed to merge this. Feel free to just edit this as you see fit.

### DIFF
--- a/library/lua/enable-util.lua
+++ b/library/lua/enable-util.lua
@@ -1,0 +1,54 @@
+
+local _ENV = mkmodule("enable-util")
+
+local repeatUtil = require('repeat-util')
+
+function assignPersistentGlobal(global_key, global_store, key, value)
+    if global_store[key] == value then
+        return
+    end
+    global_store[key] = value
+    dfhack.persistent.saveSiteData(global_key, global_store)
+end
+
+function enableEvery(flags, global_store, global_key, time, unit, action)
+
+    local function assign_persistent(key,value)
+        assignPersistentGlobal(global_key, global_store, key, value)
+    end
+
+    dfhack.onStateChange[global_key] = function(sc)
+        if sc == SC_MAP_UNLOADED then
+            global_store.enabled = false
+            -- repeat-util will cancel on its own
+            return
+        end
+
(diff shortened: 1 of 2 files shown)


library/lua/enable-util.lua
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    dfhack.onStateChange[global_key] = function(sc)
+        if sc == SC_MAP_UNLOADED then
+            global_store.enabled = false
+            -- repeat-util will cancel on its own
+            return
+        end
+
+        if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
```
myk002, line 27: especially as adventure mode is expected soon, this will need to become more flexible.

library/lua/enable-util.lua
```diff
@@ -0,0 +1,54 @@
...
+            return
+        end
+
+        -- map was loaded, restore global data and reschedule
+        global_store = dfhack.persistent.getSiteData(global_key, {enabled=false})
```
myk002, line 32: this isn't going to affect the copy of global_store in the client script. maybe `utils.assign`, but you'd also have to manually clear out old contents before `assign`ing to the table.

library/lua/enable-util.lua
```diff
@@ -0,0 +1,54 @@
+
+local _ENV = mkmodule("enable-util")
+
+local repeatUtil = require('repeat-util')
+
+function assignPersistentGlobal(global_key, global_store, key, value)
```
myk002, line 6: I don't think this is flexible enough in practice. scripts may be modifying fields deep in the state table. The pattern I've usually seen is "make a batch of changes to the state, then persist from the top". However, with that pattern, this function reduces to just the call to `dfhack.persistent.saveSiteData` and the value of abstracting that out into an `enable-util` function becomes questionable

myk002 (commented):
reducing boilerplate is good. I'm having trouble seeing exactly how existing use cases would be served by this abstraction, though. It might be useful to prototype conversions of a set of representative scripts to be sure the API is going to serve their needs

myk002:
closing this for now since I think the API isn't quite flexible enough for current script needs

PR #4418, opened 2024-03-31
Add DOWNLOAD_SDL2 CMake variable (and default to true)

numinit:
Since we package this in nixpkgs, we have to resort to horrible hacks to work around the SDL2 downloading. Make this optional and default it to TRUE.

Terrible hack in question: https://github.com/NixOS/nixpkgs/pull/300402/files#diff-0a30597e60bd9d269bd93fdb389f1c898075bf70a4d89fff1a02cb33c570d941R173

Tested:

`cmake -DDOWNLOAD_SDL2=FALSE ..`
`cmake ..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,21 +289,28 @@ endif()
 find_package(ZLIB REQUIRED)
 include_directories(${ZLIB_INCLUDE_DIRS})
 
+set(DOWNLOAD_SDL2 TRUE CACHE BOOL "Download the SDL2 headers and libraries.")
+
 if(BUILD_LIBRARY)
-    # Download SDL release and extract into depends in the build dir
-    # all we need are the header files (including generated headers), so the same release package
-    # will work for all platforms
-    # (the above statement is untested for OSX)
-    set(SDL_VERSION 2.26.2)
-    set(SDL_ZIP_MD5 574daf26d48de753d0b1e19823c9d8bb)
-    set(SDL_ZIP_FILE SDL2-devel-${SDL_VERSION}-VC.zip)
-    set(SDL_ZIP_PATH ${dfhack_SOURCE_DIR}/depends/SDL2/)
-    download_file("https://github.com/libsdl-org/SDL/releases/download/release-${SDL_VERSION}/${SDL_ZIP_FILE}"
-        ${SDL_ZIP_PATH}${SDL_ZIP_FILE}
-        ${SDL_ZIP_MD5})
-    file(ARCHIVE_EXTRACT INPUT ${SDL_ZIP_PATH}${SDL_ZIP_FILE}
-        DESTINATION ${SDL_ZIP_PATH})
-    include_directories(${SDL_ZIP_PATH}/SDL2-${SDL_VERSION}/include)
+    if(DOWNLOAD_SDL2)
+        # Download SDL release and extract into depends in the build dir
+        # all we need are the header files (including generated headers), so the same release package
+        # will work for all platforms
+        # (the above statement is untested for OSX)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -267,14 +267,18 @@ set(Python_FIND_UNVERSIONED_NAMES FIRST)
 
 include(CMake/DownloadFile.cmake)
 
+set(DOWNLOAD_FILES FALSE CACHE BOOL "Allow downloading files during the build")
```
myk002, line 270: Could this be replaced with a narrower scope "USE_SYSTEM_SDL2_HEADERS"? Then, no changes would have to be made to WIN32, DARWIN, or 32-bit code paths

myk002 (commented):
You may find https://github.com/DFHack/dfhack/blob/develop/.github/workflows/build-linux.yml useful for seeing which dependencies are required for building. Our list has slimmed down a bit compared to previous versions.

Also, DF requires SDL2-image at runtime

CMakeLists.txt
numinit: Probably not a bad idea.

numinit: Replaced with DOWNLOAD_SDL2, defaulting to TRUE

CMakeLists.txt
```diff
@@ -289,21 +289,28 @@ endif()
 find_package(ZLIB REQUIRED)
 include_directories(${ZLIB_INCLUDE_DIRS})
 
+set(DOWNLOAD_SDL2 TRUE CACHE BOOL "Download the SDL2 headers and libraries.")
```
myk002, line 292:
```suggestion
set(DOWNLOAD_SDL2 TRUE CACHE BOOL "Download the SDL2 headers for compilation. Set to FALSE to use system headers.")
```

myk002: we don't actually care about the libraries. they don't get installed with the DFHack package.

myk002 (approved):

myk002:
that's true. I thought the issue was that downloading was not allowed by the NixOS package manager as part of the build, but pre-downloading may still be possible. Reverting until we get more clarity on that.

numinit:
Downloading is handled separately by Nix. Since this implementation downloads during the generation of the build system, it doesn't work from the build sandbox which deliberately cannot access the internet to ensure build reproducibility. Treating this like any other dependency allows us to get around this issue.

numinit:
While we could, that would require us to use the exact same version of SDL as the one downloaded by dfhack, all the time, for just headers, versus [the one in nixpkgs accessible to all packages](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=sdl2) that it links against (2.30.1 at the time of writing).

How prescriptive is the exact version being downloaded by the build process? I have successfully been using nixpkgs' SDL2 without problems just by having cmake look for it.

myk002:
It's prescriptive in that it's the same version of SDL2 that DF itself builds against, but unless event-related constants change across SDL2 versions, building against a different SDL2 version is unlikely to really be a problem.

Note that you don't need to have that version of SDL2 **installed**, though. The zip just needs to be placed in the DFHack `CMake/downloads` folder before running the DFHack build. It would be similar to downloading a patch file for the Linux kernel before building that.

numinit:
Gotcha. I think the strange thing with the nixpkgs usecase is that everything is always "installed" pretty much, and we relink DF against the same version of SDL. So it is helpful to override it.

numinit:
Got it; I will resubmit with that change.

numinit:
See #4433

PR #4422, opened 2024-03-31
rename 'lerp' to 'Lerp' in remotefortressreader

numinit:
This ends up shadowing a newer function in `<cmath>` that causes compilation errors.

### DIFF
--- a/plugins/remotefortressreader/remotefortressreader.cpp
+++ b/plugins/remotefortressreader/remotefortressreader.cpp
@@ -1638,7 +1638,8 @@ static command_result GetUnitList(color_ostream &stream, const EmptyMessage *in,
     return GetUnitListInside(stream, NULL, out);
 }
 
-float lerp(float a, float b, float f)
+// NB: Cannot be called 'lerp': https://en.cppreference.com/w/cpp/numeric/lerp
+float Lerp(float a, float b, float f)
 {
     return a + f * (b - a);
 }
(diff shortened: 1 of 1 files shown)


myk002 (approved):

ab9rf:
a better change would just to remove the `using namespace std` that pulls `std::lerp` into the main namespace. `using namespace std` is a code smell anyway

myk002:
> a better change would just to remove whatever invocation of using namespace std

I was going to recommend that too, but when I looked for such a `using namespace std` statement, I didn't find one.

ab9rf:
> > a better change would just to remove whatever invocation of using namespace std
> 
> I was going to recommend that too, but when I looked for such a `using namespace std` statement, I didn't find one.

the reason is that there isn't one in _our_ code; this PR "fixes" a problem that doesn't exist

numinit:
> this PR "fixes" a problem that doesn't exist

It certainly fixed builds with newer libc++ versions where that was shadowed. Perhaps a new PR could be opened to remove the namespace creep in external dependencies?

numinit:
I think the fundamental problem is: https://cplusplus.github.io/LWG/issue3782, BTW:

> `Currently there is implementation divergence: libstdc++ provide ::lerp in <math.h> but not in <cmath>`

So, something that `remotefortressreader.cpp` depends on includes `<math.h>`, and this ends up causing the name to be shadowed.

ab9rf:
given the information about the implementation divergence issue, it's probably best to simply avoid having a function called `lerp`; as such, objection withdrawn

PR #4425, opened 2024-04-01
unlink_mechanisms.lua fix button position based on scrollbar

Bumber64:
Fix an issue in #4389 where the `Free` button can overlap a DF UI icon if a scrollbar is present. Button is now shifted horizontally based on presence of scrollbar.


myk002 (approved):
ah of course. thank you.

PR #4427, opened 2024-04-01
Document enum type next_item function

vallode:
Looks like another silently missing function that is exposed to Lua.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -356,6 +356,10 @@ bi-directional mapping between key strings and values, and
 also map ``_first_item`` and ``_last_item`` to the min and
 max values.
 
+Enum types also support the ``type.next_item(index)`` function,
+which returns the next valid index of the enum (returning the
+first value if index is greater than or equal to the max value).
+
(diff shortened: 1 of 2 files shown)


vallode:
Myk feel free to re-word this, especially the "valid index" part not sure how to word that well ^^

docs/dev/Lua API.rst
```diff
@@ -356,6 +356,10 @@ bi-directional mapping between key strings and values, and
 also map ``_first_item`` and ``_last_item`` to the min and
 max values.
 
+Enum types also support the ``type.next_item(index)`` function,
+which returns the next valid index of the enum (returning the
+first value if index is greater than or equal to the max value).
```
myk002, line 361:
```suggestion
which returns the next valid numeric value of the enum. It
Returns the first enum value if ``index`` is greater than or
equal to the max enum value.
```

myk002: how's that?

myk002 (commented):
this is worth an entry under Documentation in the changelog, I think

myk002 (approved):

PR #4431, opened 2024-04-01
Add type annotations to Lua library files

vallode:
The aim of this PR is to add as much typing to the Lua files as possible, which can then be leveraged in dfhack-lua-annotations directly without needing to manually copy them over.

Right now I covered only the easiest functions to type and only the core dfhack.lua module, but will try to cover more as we get there.

Note these types sometimes rely on annotations that only exist in dfhack-lua-annotations (namely `coord` and `coord_path`) but this has no real impact even if the library is not loaded.

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -475,11 +475,6 @@ end
 
 -- String class extentions
 
--- This fixes LuaLS complaining about field injection.
-
----@type string
-local string
-
 -- prefix is a literal string, not a pattern
 ---@nodiscard
 ---@param self string
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -335,7 +335,7 @@ end
 ---@return T
 function copyall(table)
     local rv = {}
-    for k, v in pairs(table) do rv[k] = v end
+    for k,v in pairs(table) do rv[k] = v end
     return rv
 end
 
@@ -357,11 +357,11 @@ end
 ---@param y number
 ---@param z number
 ---@return coord
-function xyz2pos(x, y, z)
+function xyz2pos(x,y,z)
     if x then
-        return { x = x, y = y, z = z }
+        return {x=x,y=y,z=z}
     else
-        return { x = -30000, y = -30000, z = -30000 }
+        return {x=-30000,y=-30000,z=-30000}
     end
 end
 
@@ -369,7 +369,7 @@ end
 ---@param a coord
 ---@param b coord
 ---@return boolean
-function same_xyz(a, b)
+function same_xyz(a,b)
     return a and b and a.x == b.x and a.y == b.y and a.z == b.z
 end
 
@@ -379,7 +379,7 @@ end
 ---@return number x
 ---@return number y
 ---@return number z
-function get_path_xyz(path, i)
+function get_path_xyz(path,i)
     return path.x[i], path.y[i], path.z[i]
 end
 
@@ -400,11 +400,11 @@ end
 ---@param x number
 ---@param y number
 ---@return coord2d
-function xy2pos(x, y)
+function xy2pos(x,y)
     if x then
-        return { x = x, y = y }
+        return {x=x,y=y}
     else
-        return { x = -30000, y = -30000 }
+        return {x=-30000,y=-30000}
     end
 end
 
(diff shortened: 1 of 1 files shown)


myk002:
the tests appear to be failing because `string` is nil. I'm not sure why that is, given the (supposedly noop) changes in this PR.

Also, could you revert the formatting changes, especially with the added spaces inside of inline table declarations? I'd actually prefer if inline table declarations had no space around the `=`. It differentiates them from regular assignment statements.

library/lua/dfhack.lua
```diff
@@ -367,46 +395,77 @@ function pos2xy(pos)
...
 
-function safe_index(obj,idx,...)
+-- Walks a sequence of dereferences, which may be represented by numbers or
+-- strings. Returns nil if any of obj or indices is nil, or a numeric index is
+-- out of array bounds.
+---@generic T
+---@param obj `T`
+---@param idx number|string
+---@param ... number|string
+---@return T obj
```
myk002, line 434: obj will be some sort of table or vector, but the type of the return value is not related to the type of obj. Shouldn't this be `@return any`?

library/lua/dfhack.lua
```diff
@@ -367,46 +395,77 @@ function pos2xy(pos)
...
+function safe_index(obj, idx, ...)
     if obj == nil or idx == nil then
         return nil
     end
     if type(idx) == 'number' and
-            type(obj) == 'userdata' and -- this check is only relevant for c++
-            (idx < 0 or idx >= #obj) then
+        type(obj) == 'userdata' and     -- this check is only relevant for c++
+        (idx < 0 or idx >= #obj) then
```
myk002, line 441: the convention I've been following here is that if the conditions are multi-line, then the `then` should go on a new line by itself

library/lua/dfhack.lua
```diff
@@ -367,46 +395,77 @@ function pos2xy(pos)
...
     else
         return obj
     end
 end
 
+---@generic T
+---@param t `T`
+---@param key integer|string
+---@param default_value? any
+---@return T
```
myk002, line 456: similar to `safe_index`, I think the return type should be `any`

library/lua/dfhack.lua
```diff
@@ -367,46 +395,77 @@ function pos2xy(pos)
...
         t[key] = (default_value ~= nil) and default_value or {}
     end
     return t[key]
 end
 
+---@generic T
+---@param t `T`
+---@param key integer|string
+---@param ... integer|string
+---@return T
```
myk002, line 468: the return type will always be a table here

library/lua/dfhack.lua
```diff
@@ -417,12 +476,23 @@ end
 
 -- String class extentions
 
+---@class string
+local string
```
myk002, line 480: this is why the tests are failing. this is overriding the builtin string definition

library/lua/dfhack.lua
```diff
@@ -451,6 +526,9 @@ end
 
 -- Removes spaces (i.e. everything that matches '%s') from the start and end of
 -- a string. Spaces between non-space characters are left untouched.
+---@nodiscard
+---@param self string
+---@return self
```
myk002, line 531: for these `string` functions, it doesn't return `self`, it returns a new `string`

library/lua/dfhack.lua
vallode: Figured, the language server complains about the overridden functions otherwise... I'll have to play around with this.

library/lua/dfhack.lua
vallode: Gotcha, is how I formatted it now good?

vallode:
Mentioned on the Discord server but I will send it here for posterity:

> RE: Formatting in the lua files in DFHack. I'm really more of a fire and forget kind of person in terms of formatting, I think it's more trouble than it is worth maintaining a "manual" style but I appreciate preferences. If there was a vote I would say we use whatever linter/formatter there is and adapt to all its rules, seeing as the files right now are not consistent either.

I think seeing as this file alone has inconsistency in whitespacing around functions, arguments, calls, etc. it would benefit us from using any kind of formatting tool instead of trying to manually maintain style.

library/lua/dfhack.lua
```diff
@@ -375,21 +408,38 @@ function xy2pos(x,y)
...
     return path.x[i], path.y[i]
 end
 
+-- Walks a sequence of dereferences, which may be represented by numbers or
+-- strings. Returns nil if any of obj or indices is nil, or a numeric index is
+-- out of array bounds.
+---@param obj table
+---@param idx number|string
+---@param ... number|string
+---@return any obj
```
myk002, line 434: I'm not fully aware of what the syntax is for the annotations, but should this be "any" instead of "any obj"?

library/lua/dfhack.lua
myk002: looks good. part of the problem was just that the formatting changes were making the PR more difficult to grok

library/lua/dfhack.lua
vallode: Return types in LuaLS follow this syntax:
`---@return <type> [<name> [comment] | [name] #<comment>]`

vallode:
Note to self for later: Lua doesn't like string being re-declared and LuaLS doesn't like string not being declared for injecting functions.

myk002:
Can you do something silly like `string = string`?

myk002:
This PR looks mergeable now. Do you plan on making any further changes, or can we unmark it as draft and merge?

vallode:
As mentioned on Discord I think this is safe to merge and we can chip away at more annotations in other PRs. Thanks!

PR #4433, opened 2024-04-02
Add USE_SYSTEM_SDL2 CMake option, defaulting to OFF

numinit:
Since systems like nixpkgs change the search path of Dwarf Fortress, we should be able to use the same SDL2 instead of resorting to horrible hacks to work around the file downloading or using mismatched library versions between Dwarf Fortress and dfhack.


numinit:
No worries. nixpkgs is slightly strange. Everything is installed!

PR #4445, opened 2024-04-05
Use defclass/mkinstance types in dfhack.lua

vallode:
Removes the unnecessary (imo) function declaration in dfhack.lua in favour of an alias and gives the class methods types.

### DIFF
--- a/library/lua/class.lua
+++ b/library/lua/class.lua
@@ -15,6 +15,10 @@ reserved_names = { super = true, ATTRS = true }
 attrs_meta = attrs_meta or {}
 
 -- Create or updates a class; a class has metamethods and thus own metatable.
+---@generic T: table
+---@param class? T
+---@param parent? table
+---@return table|T
 function defclass(class,parent)
     class = class or {}
 
@@ -39,6 +43,10 @@ function defclass(class,parent)
 end
 
 -- An instance uses the class as metatable
+---@generic T: table
+---@param class table
+---@param table? T
+---@return table|T
 function mkinstance(class,table)
     table = table or {}
     setmetatable(table, class)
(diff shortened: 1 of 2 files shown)


library/lua/dfhack.lua
```diff
@@ -166,13 +166,9 @@ end
...
 DEFAULT_NIL = DEFAULT_NIL or {} -- Unique token
 
-function defclass(...)
-    return require('class').defclass(...)
-end
-
-function mkinstance(...)
-    return require('class').mkinstance(...)
-end
+local class = require('class')
```
myk002, line 169: copying my discord message here for posterity:

The reason for these indirections in dfhack.lua is to avoid circular dependencies.  dfhack.lua must not load any other files when it is parsed for the first time.

dfhack.lua is loaded first at Lua context initialization time. If it calls `require` to load another module, that module may try to use symbols in dfhack.lua that are not loaded yet.

library/lua/class.lua
```diff
@@ -15,6 +15,10 @@ reserved_names = { super = true, ATTRS = true }
 attrs_meta = attrs_meta or {}
 
 -- Create or updates a class; a class has metamethods and thus own metatable.
+---@generic T: table
+---@param class? T
+---@param parent? table
+---@return table|T
```
myk002, line 21: is the return type `table|T` because `class` can be nil?

myk002 (approved):

library/lua/class.lua
vallode: Exactly right if class is nill we return a new table. Technically we should define a new class type above this that includes common methods and IT should be the return type but baby steps!

PR #4446, opened 2024-04-05
More dfhack.lua annotations

vallode:
Next round of annotations added to the core dfhack.lua file. Raised questions:

- I'm currently leaning towards abundant usage of `@nodiscard` which will warn users when a function has a return but they have not consumed it. Just something to keep in mind.
- with_finalize and with_onerror use variable arguments when the function they alias/call have known arguments, could be typed better.
- I think we should probably use aliases instead of new functions for the defclass overloads I tried to address this in #4445

### DIFF
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -100,21 +100,19 @@ function qerror(msg, level)
 end
 
 ---@generic T
----@generic S
----@param cleanup_fn fun(...): S
+---@param cleanup_fn function
 ---@param fn fun(...): T
 ---@param ... any
----@return T|S
+---@return T
 function dfhack.with_finalize(cleanup_fn,fn,...)
     return dfhack.call_with_finalizer(0,true,cleanup_fn,fn,...)
 end
 
 ---@generic T
----@generic S
----@param cleanup_fn fun(...): S
+---@param cleanup_fn function
 ---@param fn fun(...): T
 ---@param ... any
----@return T|S
+---@return T
 function dfhack.with_onerror(cleanup_fn,fn,...)
     return dfhack.call_with_finalizer(0,false,cleanup_fn,fn,...)
 end
--- a/library/lua/dfhack.lua
+++ b/library/lua/dfhack.lua
@@ -81,9 +81,10 @@ safecall = dfhack.safecall
 curry = dfhack.curry
 
 ---@generic T
----@param f fun(...):T
+---@param f fun(...): T
 ---@param ... any
----@return T
+---@return boolean success
+---@return T|string ...
 function dfhack.pcall(f, ...)
     return xpcall(f, dfhack.onerror, ...)
 end
@@ -98,16 +99,27 @@ function qerror(msg, level)
     dfhack.error(msg, (level or 1) + 1, false)
 end
 
+---@generic T
+---@generic S
+---@param cleanup_fn fun(...): S
+---@param fn fun(...): T
 ---@param ... any
-function dfhack.with_finalize(...)
-    return dfhack.call_with_finalizer(0,true,...)
+---@return T|S
+function dfhack.with_finalize(cleanup_fn,fn,...)
+    return dfhack.call_with_finalizer(0,true,cleanup_fn,fn,...)
 end
 
+---@generic T
+---@generic S
+---@param cleanup_fn fun(...): S
+---@param fn fun(...): T
 ---@param ... any
-function dfhack.with_onerror(...)
-    return dfhack.call_with_finalizer(0,false,...)
+---@return T|S
+function dfhack.with_onerror(cleanup_fn,fn,...)
+    return dfhack.call_with_finalizer(0,false,cleanup_fn,fn,...)
 end
 
+---@param obj DFObject
 local function call_delete(obj)
     if obj then obj:delete() end
 end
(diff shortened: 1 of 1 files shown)


library/lua/dfhack.lua
```diff
@@ -74,10 +80,16 @@ end
 safecall = dfhack.safecall
 curry = dfhack.curry
 
+---@generic T
+---@param f fun(...):T
+---@param ... any
+---@return T
 function dfhack.pcall(f, ...)
```
myk002, line 87: this one is tricky. On success, this function returns `true, ...` where `...` are the return values from `f` (different from the `@param ...`, which are the arguments to `f`).
On failure, it returns `false, err` where err is the error object, which I think is always string error message, at least whenever I've seen it used.

library/lua/dfhack.lua
```diff
@@ -86,9 +98,12 @@ function qerror(msg, level)
     dfhack.error(msg, (level or 1) + 1, false)
 end
 
+---@param ... any
 function dfhack.with_finalize(...)
     return dfhack.call_with_finalizer(0,true,...)
 end
+
+---@param ... any
 function dfhack.with_onerror(...)
```
myk002, line 107: I agree that these type signatures could be improved by defining the known first N args. The Lua docs define them, why aren't they defined here? https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#locking-and-finalization

library/lua/dfhack.lua
```diff
@@ -97,6 +112,11 @@ local function call_delete(obj)
     if obj then obj:delete() end
 end
 
+---@generic T
+---@param obj DFObject
+---@param fn fun(...):T
+---@param ...? any
+---@return T
```
myk002, line 119: `T` can be multiple values. Is that adequately expressed with a single `T`?

library/lua/dfhack.lua
```diff
@@ -215,10 +240,12 @@ local function print_element(k, v)
     dfhack.println(string.format("%-23s\t = %s", tostring(k), tostring(v)))
 end
 
+---@param table table
```
myk002, line 243: could also be a c++ vector. do we want to draw that disctinction?

library/lua/dfhack.lua
```diff
@@ -636,33 +688,54 @@ function dfhack.gui.getViewscreenByType(scr_type, n)
...
 local function persistent_getData(which, key, default)
     local serialized = dfhack.persistent['get'..which..'DataString'](key)
     if not serialized then return default end
     return require('json').decode(serialized) or default
 end
 
+---@param which string
+---@param key string
+---@param data any
 function persistent_saveData(which, key, data)
```
myk002, line 711: `persistent_saveData` should be `local` like `persistent_getData` -- could you make that fix?

library/lua/dfhack.lua
```diff
@@ -8,6 +8,8 @@
 
 ---@class dfhack
 ---@field BASE_G _G Original Lua global environment
+---@field is_core_context boolean
+---@field is_interactive fun(): boolean
```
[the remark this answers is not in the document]
vallode: For now this is my duct tape method, it works but lacks some of the benefits a proper function decleration gives in LuaLS (there are a few issues open regarding these inline function fields).

I would leave it like this until we have a discussion/brainstorm on finalizing a method of parsing exposed C++ functions. The methods are probably:

- The above, just defining fields on a global `dfhack` object amongst others for the modules.
- Creating a separate definitions file that can be yanked out by the annotations library (the above, but a bit more verbose)
- Add some kind of special tag to the C++ files that can be easily parsed, essentially the above but inlining the definitions with the methods
- Someone whips up a magical bug-free method of parsing the C++ functions perfectly and detecting all exposed ones automatically :magic_wand: 

Again I think this is going to be a slightly more in-depth discussion but I put these in here for now just to have them for the file.

library/lua/dfhack.lua
vallode: Probably a case where we run against the edge of what LuaLS can actually represent. I think in this case we just do:

```
---@generic T
---@param f fun(...): T
---@param ... any -- LuaLS does not like typing variable arguments and especially not inside of a inline function :-(
---@return boolean
---@return T|string ... -- LuaLS will not correctly type variable return types as T but this is not a major drawback.
``

File: library/lua/dfhack.lua (hunk 3 above)
Comment (vallode) [reply library/lua/dfhack.lua:119]: Changed to variable return but as stated above LuaLS doesn't acknowledge that with generics. The rest of the values will be unknown.

File: library/lua/dfhack.lua (hunk 4 above)
Comment (vallode) [reply library/lua/dfhack.lua:243]: If there is no special behaviour there I think it's fine not to include, if it quacks like a table...

File: library/lua/dfhack.lua (hunk 1 above)
Comment (myk002) [reply library/lua/dfhack.lua]: lgtm

File: library/lua/dfhack.lua [hunk 7]
```diff
@@ -86,17 +99,36 @@ function qerror(msg, level)
...
 end
 
-function dfhack.with_finalize(...)
-    return dfhack.call_with_finalizer(0,true,...)
+---@generic T
+---@generic S
+---@param cleanup_fn fun(...): S
+---@param fn fun(...): T
+---@param ... any
+---@return T|S
```
myk002, line 107: The return value from `cleanup_fn` is not propagated here, just the return values from `fun`

same with `with_onerror` below

vallode: I misread the docs!

myk002 (approved):

PR #4454, opened 2024-04-06
[suspendmanager] don't unsuspend constructions blocked by an item

### LINKED ISSUE #4087: `suspendmanager`: handle deadlocks due to obstruction better
the issue:  consider this scenario:
image
the construction in the corner needs to have the boulder cleared before it can proceed, but the game's obstruction-clearing mechanic cannot clear the boulder because none of the four available locations to clear it to are available: two are ineligible open spaces and two contain (suspended) constructions. as a result, the game will suspend the construction with "item blocking site". suspendmanager will then unsuspend the construction, leading to a cycle that generates cancellation spam

this is a deadlock. while in this scenario suspendmanager could allow one of the two adjacent constructions to proceed safely it just leads back to essentially the same scenario. allowing both to proceed would block off the unconstructed corner wall, and the game won't allow the corner wall to be built

the only way to resolve this deadlock is to remove the obstructing boulder. i don't think suspendmanager should be responsible for this; instead, what suspendmanager should do in this situation is detect the deadlock, suspend all involved constructions, notify the player of the deadlock, and wait for the player to do something about it

please note that i have no idea how to go about actually doing this, or even if it's feasible

[related discussion on discord](https://discord.com/channels/793331351645323264/1017471248567124049/1181528853177438258)

plule:
When there is an item on top of a construction that the dwarves can't move, `suspendmanager` would always try to unsuspend it, only for the dwarves to suspend it again. This lead to a cancellation spam and quite a waste of time of dwarves trying to build the same thing again and again.

This change detects items that cannot be moved away, and register it as an "external suspension", so that `suspendmanager` will neither suspend nor unsuspend it while it's in this situation, with a message explaining on the suspended construction.

If an actual deadlock is there, it is not fixed, it will wait suspended for a user action. However, it makes some situations like building a floor on top of its scattered construction material much more efficient, where constructions are unsuspended as soon as they are buildable:

image

Fixes #4087

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -68,6 +68,7 @@ Template for new versions:
 - fix behavior of Linux Steam launcher on systems that don't support the inotify API
 - fix rendering of resize "notch" in lower right corner of resizable windows in ascii mode
 - `quickfort`: stockpiles can now be placed even if there is water covering the tile, as per vanilla behavior
+- `suspendmanager`: fix a cancellation spam when an item prevents a building to be completed
 
 ## Misc Improvements
 - `tailor`: allow turning off automatic confiscation of tattered clothing
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -68,7 +68,7 @@ Template for new versions:
 - fix behavior of Linux Steam launcher on systems that don't support the inotify API
 - fix rendering of resize "notch" in lower right corner of resizable windows in ascii mode
 - `quickfort`: stockpiles can now be placed even if there is water covering the tile, as per vanilla behavior
-- `suspendmanager`: fix a cancellation spam when an item prevents a building to be completed
+- `suspendmanager`: prevent cancellation spam when an item is preventing a building from being completed
 
 ## Misc Improvements
 - `tailor`: allow turning off automatic confiscation of tattered clothing
--- a/plugins/suspendmanager.cpp
+++ b/plugins/suspendmanager.cpp
@@ -317,8 +317,8 @@ private:
         // Check for items on the building, look into all the map blocks overlapping
         // the building
         bool has_item = false;
-        for (size_t block_x = building->x1 >> 4; block_x <= building->x2 >> 4; ++block_x) {
-            for (size_t block_y = building->y1 >> 4; block_y <= building->y2 >> 4; ++block_y) {
+        for (int32_t block_x = building->x1 >> 4; block_x <= building->x2 >> 4; ++block_x) {
+            for (int32_t block_y = building->y1 >> 4; block_y <= building->y2 >> 4; ++block_y) {
                 auto block = Maps::getBlock(block_x,block_y,building->z);
                 if (!block)
                     continue;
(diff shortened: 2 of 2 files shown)


docs/changelog.txt
```diff
@@ -68,6 +68,7 @@ Template for new versions:
 - fix behavior of Linux Steam launcher on systems that don't support the inotify API
 - fix rendering of resize "notch" in lower right corner of resizable windows in ascii mode
 - `quickfort`: stockpiles can now be placed even if there is water covering the tile, as per vanilla behavior
+- `suspendmanager`: fix a cancellation spam when an item prevents a building to be completed
```
myk002, line 71:
```suggestion
- `suspendmanager`: prevent cancellation spam when an item is preventing a building from being completed
```

plugins/suspendmanager.cpp
```diff
@@ -291,6 +298,70 @@ class SuspendManager {
...
+        auto building = Job::getHolder(job);
+        if (!building)
+            return false;
+
+        // Check for items on the building, look into all the map blocks overlapping
+        // the building
+        bool has_item = false;
+        for (int32_t block_x = building->x1 >> 4; block_x <= building->x2 >> 4; ++block_x) {
+            for (int32_t block_y = building->y1 >> 4; block_y <= building->y2 >> 4; ++block_y) {
+                auto block = Maps::getBlock(block_x,block_y,building->z);
```
myk002, line 322: this would be easier to understand, I think, if you didn't bitshift the coordinates and used `Maps::getTileBlock` instead of `Maps::getBlock`

plugins/suspendmanager.cpp
```diff
@@ -291,6 +298,70 @@ class SuspendManager {
...
+                if (suitableToMoveItemTo(coord(x,y,building->z)))
+                    return false;
+            }
+        }
+        for (auto y : {building->y1-1, building->y2+1}) {
+            for (auto x = building->x1; x <= building->x2; ++x) {
+                if (suitableToMoveItemTo(coord(x,y,building->z)))
+                    return false;
+            }
+        }
```
myk002, line 360: are the corners valid positions to move items to? They're not included in this check.

plugins/suspendmanager.cpp
plule: I'm annoyed by that too, but the issue is that the list of items is stored on the map block, and I did not find any direct access from a tile to an item. Due to that, if the iteration is on all the building tile, this would multiply the checks to do (from the tile find the map block, from the map block enumerate the items and find the ones on the tile). A worst case scenario would be a 4x4 building on a single map block, enumerating the items in this block 16 times instead of one :(

plugins/suspendmanager.cpp
plule: Nope, the corners are not valid in my tests, I'm updating the comment.

Here the axle is getting suspended by the dwarves for as long as the wall is not built. Without wall, they will transport the annoying log all the way up though:

image

plugins/suspendmanager.cpp
myk002: ah ok, understood

myk002 (approved):

myk002 (approved):

PR #4468, opened 2024-04-15
suspendmanager: take in account wall plans on the tile below

### LINKED ISSUE #3600: Suspendmanager multi-level ramp issue
Example of incorrect suspendmanager handling I encountered today:

Level 1:
Dwarf_Fortress_JEOhsduxxO

Level 2 (above):
Dwarf_Fortress_TzGtWkj13l

These suspensions cause a central wall to be built before the outer edges and the wall can't be completed. Should have been suspended similar to the wall on the left side in order to work correctly.

After:
Dwarf_Fortress_nm3smzAWw5

### LINKED ISSUE #4466: [Bug] Suspendmanager doesn't suspend jobs which block a second job that some third job allows dwarves to reach
stairs1
stairs2

I was building a pretty simple staircase with side and corner walls, and even with suspendmanager my dwarves built a side wall that blocked a corner wall (second image), because at the time the side wall was reachable but the corner wall was not (because the side wall below this floor was not yet built, layout of completed lower floor shown in first image)

plule:
`suspendmanager` now consider wall being built below. Before, it would sometimes consider that a construction is not at risk of blocking another, since it is not walkable. This would lead it allowing some construction even if they are blocking another, since at the time of the construction the other is unreachable.

It now looks below to consider tiles that are "potentially walkable", to keep access to constructions, even if they are not yet reachable.

Fixes #3600 
Fixes #4466

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -59,7 +59,7 @@ Template for new versions:
 
 ## Misc Improvements
 
-- `suspendmanager`: Take in account walls planned on the z-layer below to determine accessibility to a job
+- `suspendmanager`: Account for walls planned on the z-layer below when determining accessibility to a job
 
 ## Documentation
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -59,6 +59,8 @@ Template for new versions:
 
 ## Misc Improvements
 
+- `suspendmanager`: Take in account walls planned on the z-layer below to determine accessibility to a job
+
 ## Documentation
 
 ## API
--- a/plugins/suspendmanager.cpp
+++ b/plugins/suspendmanager.cpp
@@ -111,6 +111,12 @@ static const std::bitset<64> construction_impassible = std::bitset<64>()
     .set(construction_type::Wall)
     .set(construction_type::Fortification);
 
+// constructions that can be stand on the tile above
+static const std::bitset<64> construction_stand_above = std::bitset<64>()
+    .set(construction_type::Wall)
+    .set(construction_type::UpStair)
+    .set(construction_type::UpDownStair);
+
 // constructions requiring same support as walls
 static const std::bitset<64> construction_wall_support = std::bitset<64>()
     .set(construction_type::Wall)
@@ -378,6 +384,18 @@ private:
         return walkable(pos);
     }
 
+    // check if the tile is suitable to stand on and build, or is planned to become one
+    static bool isPotentialSuitableAccess (coord pos) {
+        if (isSuitableAccess(pos))
+            return true;
+
+        auto below = Buildings::findAtTile(coord(pos.x,pos.y,pos.z-1));
+        if (below && below->getType() == df::building_type::Construction && construction_stand_above[below->getSubtype()])
+            return true;
+
+        return false;
+    }
+
     static bool tileHasSupportWall(coord pos) {
         auto tile_type = Maps::getTileType(pos);
         if (tile_type) {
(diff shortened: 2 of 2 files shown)


docs/changelog.txt
```diff
@@ -59,6 +59,8 @@ Template for new versions:
 
 ## Misc Improvements
 
+- `suspendmanager`: Take in account walls planned on the z-layer below to determine accessibility to a job
```
plule, line 62: Not sure if it is a bug fix or improvement, and the wording is hard here :)

chdoc: Clearly an improvement, there never was the expectation to handle all cases with our heuristics.

plugins/suspendmanager.cpp
```diff
@@ -111,6 +111,12 @@ static const std::bitset<64> construction_impassible = std::bitset<64>()
     .set(construction_type::Wall)
     .set(construction_type::Fortification);
 
+// constructions that can be stand on the tile above
```
chdoc, line 114:
```suggestion
// constructions that allow walking on the tile above
```

plugins/suspendmanager.cpp
```diff
@@ -378,6 +384,18 @@ class SuspendManager {
         return walkable(pos);
     }
 
+    // check if the tile is suitable to stand on and build, or is planned to become one
+    static bool isPotentialSuitableAccess (coord pos) {
+        if (isSuitableAccess(pos))
+            return true;
+
+        auto below = Buildings::findAtTile(coord(pos.x,pos.y,pos.z-1));
+        if (below && below->getType() == df::building_type::Construction && construction_stand_above[below->getSubtype()])
```
chdoc, line 393:
```suggestion
        if (below && below->getType() == df::building_type::Construction &&
            construction_stand_above[below->getSubtype()])
```

plugins/suspendmanager.cpp
```diff
@@ -522,8 +542,9 @@ class SuspendManager {
 
             df::building* exit = nullptr;
             for (auto npos : neighbors | transform(around(pos))) {
-                if (!isSuitableAccess(npos))
-                    continue; // non walkable neighbour, not an exit
+                if (!isPotentialSuitableAccess(npos))
+                    // non walkable neighbour, nor planned to become one, not an exit
+                    continue;
```
chdoc, line 547: If I understand correctly, then this change allows filling multi-z dead ends. However, if the wall built on the level below isn't actually accessible, this may actually cause dead ends to be closed off. But if this allows designating multi-z width-1 walls from one adjacent stairwell for "scaffolding", I'm all for it.

chdoc (commented):
LGTM!

plugins/suspendmanager.cpp
plule: It's intended to adress multi-z dead ends yes (it works better when there is no cat):
multi-level-deadend

> However, if the wall built on the level below isn't actually accessible, this may actually cause dead ends to be closed off

I think if a wall below is inaccessible, this blocks the whole process until the player have restored access to it, which feels okayish

chdoc: "Dead end" doesn't require a wall. So this should also work for free-standing multi-z walls (i.e. to close off part off a cavern), which would be a great plus, certainly making up for some minor inconveniences in rare cases. :clap:

docs/changelog.txt
myk002, line 62:
```suggestion
- `suspendmanager`: Account for walls planned on the z-layer below when determining accessibility to a job
```

myk002 (approved):

PR #4470, opened 2024-04-15
Add hotkey (Shift+R) to call the markdown script

glotov4:
As per https://github.com/DFHack/scripts/pull/901

### DIFF
--- a/data/init/dfhack.keybindings.init
+++ b/data/init/dfhack.keybindings.init
@@ -25,6 +25,9 @@ keybinding add Ctrl-Shift-K gui/cp437-table
 # customizable quick command list
 keybinding add Ctrl-Shift-A gui/quickcmd
 
+# save the description of a selected unit or item to the `/markdown_{YourWorldName}.md` file in the root game directory
+keybinding add Shift-R@dwarfmode/ViewSheets/UNIT|dwarfmode/ViewSheet markdown
+
 
 ###################
 # embark bindings #


data/init/dfhack.keybindings.init
```diff
@@ -25,6 +25,9 @@ keybinding add Ctrl-Shift-K gui/cp437-table
 # customizable quick command list
 keybinding add Ctrl-Shift-A gui/quickcmd
 
+# save the description of a selected unit or item to the `/markdown_{YourWorldName}.md` file in the root game directory
```
myk002, line 28: starting a filename with a slash indicates that it will be saved in the root folder of the entire system, not the current directory. I'd remove that leading slash.

data/init/dfhack.keybindings.init
```diff
@@ -25,6 +25,9 @@ keybinding add Ctrl-Shift-K gui/cp437-table
...
+keybinding add Shift-R@dwarfmode/ViewSheets/UNIT|dwarfmode/ViewSheet markdown
```
myk002, line 29: I'm vary wary about adding a binding to regular text keys. It can cause non-obvious problems with text entry. For example, it would prevent a squad from having a name with a capital R in it:
image

I suggest Alt-R instead

```
dwarfmode/ViewSheet
```
I think this got truncated. This is supposed to be `dwarfmode/ViewSheets/ITEM`, right?

Also, please move this down to the dwarfmode section below.

myk002:
After reviewing https://github.com/DFHack/scripts/pull/901, I think the answer is to use an overlay instead of a keybinding. Individual players can still set up a keybinding if they want custom parameters, though, and that should definitely still be documented in markdown.rst

data/init/dfhack.keybindings.init
myk002, line 28:
```suggestion
# save the description of a selected unit or item to the `markdown_{YourWorldName}.md` file
# in the game root directory
```

data/init/dfhack.keybindings.init
myk002, line 29:
```suggestion
keybinding add Ctrl-T@dwarfmode/ViewSheets/UNIT|dwarfmode/ViewSheets/ITEM markdown
```

PR #4474, opened 2024-04-18
add cron job to watch steam and autodeploy new versions of DFHack

myk002:


myk002:
merging so I can see it execute and see whether the basic logic works

PR #4476, opened 2024-04-18
[cursecheck] don't try to rely on cursor; check for selected unit.

chdoc:
also includes some "mild" cleanup.

### DIFF
--- a/docs/plugins/cursecheck.rst
+++ b/docs/plugins/cursecheck.rst
@@ -5,11 +5,11 @@ cursecheck
     :summary: Check for cursed creatures.
     :tags: fort armok inspection units
 
-This command checks a single map tile (or the whole map/world) for cursed
-creatures (ghosts, vampires, necromancers, werebeasts, zombies, etc.).
+This command checks a single unit or the whole map for curses (ghosts, vampires,
+necromancers, werebeasts, zombies, etc.).
 
-With an active in-game cursor, only the selected tile will be checked. Without a
-cursor, the whole map will be checked.
+If a unit is selected, only the selected unit will be checked. Otherwise, the
+all units on the map will be checked.
 
 By default, you will just see the count of cursed creatures in case you just
 want to find out if you have any of them running around in your fort. Dead and
--- a/docs/plugins/cursecheck.rst
+++ b/docs/plugins/cursecheck.rst
@@ -8,8 +8,8 @@ cursecheck
 This command checks a single unit or the whole map for curses (ghosts, vampires,
 necromancers, werebeasts, zombies, etc.).
 
-If a unit is selected, only the selected unit will be checked. Otherwise, the
-all units on the map will be checked.
+If a unit is selected, only the selected unit will be checked. Otherwise, all
+units on the map will be checked.
 
 By default, you will just see the count of cursed creatures in case you just
 want to find out if you have any of them running around in your fort. Dead and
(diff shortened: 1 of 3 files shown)


docs/plugins/cursecheck.rst
```diff
@@ -5,11 +5,11 @@ cursecheck
...
     :tags: fort armok inspection units
 
-This command checks a single map tile (or the whole map/world) for cursed
-creatures (ghosts, vampires, necromancers, werebeasts, zombies, etc.).
+This command checks a single unit or the whole map for curses (ghosts, vampires,
+necromancers, werebeasts, zombies, etc.).
 
-With an active in-game cursor, only the selected tile will be checked. Without a
-cursor, the whole map will be checked.
+If a unit is selected, only the selected unit will be checked. Otherwise, the
```
myk002, line 11:
```suggestion
If a unit is selected, only the selected unit will be checked. Otherwise,
```

plugins/cursecheck.cpp
```diff
@@ -153,65 +138,36 @@ command_result cursecheck (color_ostream &out, vector <string> & parameters)
...
             giveDetails = true;
             verbose = true;
         }
     }
 
-    // check whole map if no cursor is active
-    bool checkWholeMap = false;
-    if(cursorX == -30000)
+    // check whole map if no unit is selected
+    for(df::unit *unit : selected_unit ? vector{ selected_unit } : world->units.all)
```
myk002, line 162:
```suggestion
    for(df::unit *unit : selected_unit ? vector{ selected_unit } : world->units.active)
```

myk002: I don't think there's any benefit to scanning non-active units

plugins/cursecheck.cpp
```diff
@@ -244,29 +193,23 @@ command_result cursecheck (color_ostream &out, vector <string> & parameters)
                 }
```
myk002, line 193: I think you can replace this entire `if unit->name.has_name ... else` block with `out.print(Units::getReadableName(unit);`

plugins/cursecheck.cpp
```diff
@@ -244,29 +193,23 @@ command_result cursecheck (color_ostream &out, vector <string> & parameters)
...
-
                 out.print("born in %d, cursed in %d to be a %s. (%s%s%s)\n",
                     unit->birth_year,
                     unit->curse_year,
-                    curse_names [static_cast<size_t>(cursetype)].c_str(),
+                    curse_names[cursetype].c_str(),
                     // technically most cursed creatures are undead,
                     // therefore output 'active' if they are not completely dead
                     unit->flags2.bits.killed ? "deceased" : "active",
                     unit->flags3.bits.ghostly ? "-ghostly" : "",
```
myk002, line 203: ghostly will be output as part of getReadableName, if you switch to using that above

plugins/cursecheck.cpp
chdoc: I also don't see the point of doing so. However, it is part of the explicitly advertised functionality. Are you suggesting to remove the `all` option and everything that goes along with it?

plugins/cursecheck.cpp
chdoc: Thanks, I was looking for this functionality but didn't find it! :+1:

plugins/cursecheck.cpp
myk002: you're right -- I didn't see the phrasing "Include dead and passive cursed creatures". that does indeed mean `units.all`

plugins/cursecheck.cpp
myk002: it's relatively new. I got tired of synthesizing a "descriptive unit string" in multiple tools and finally created a new centralized API for it

myk002 (approved):

Issue #3596, opened 2023-07-23
Allow regrass to grow grass in grassless biomes


myk002:
It just looks like it needs to allocate a grass map block event if there isn't one already

Issue #1820, opened 2021-03-30
Script to clean up ashes


Bumber64:
Develop a script that replaces ashes with the appropriate tile type.

lethosor:
Does `regrass` already handle the grass-growing aspect of this?

PR #4479, opened 2024-04-18
Improve regrass plugin

Bumber64:
Improve `regrass` plugin to work on non-floors, ashes, muddy stone, shrubs, and trees. Options to operate on single block or tile. Optionally add new grass events to blocks. Debug logging.

`TileTypes.h`: Recommend passing `nullptr` to `findTileType` instead of `NULL` for proper constructor.
`Gui.cpp`: Replace instance of `binsearch_index() >= 0` with `vector_contains()`.

Fixes https://github.com/DFHack/dfhack/issues/1820
Fixes https://github.com/DFHack/dfhack/issues/3596


docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+Operates on floors, stairs, and ramps. Also works underneath shrubs, saplings,
+and tree trunks. Ignores furrowed soil, beaches, and tiles under buildings.
 
 Usage
 -----
 
 ::
 
-    regrass [max]
+    regrass [options]
```
myk002, line 17:
```suggestion
    regrass [<options>]
```

myk002: `max` was a string literal; `options` is a description and needs angle brackets

plugins/regrass.cpp
```diff
@@ -12,109 +22,423 @@ using namespace DFHack;
...
+            options.new_grass = true;
+        else if (s == "a" || s == "ashes")
+            options.ashes = true;
+        else if (s == "u" || s == "mud")
+            options.mud = true;
+        else if (s == "p" || s == "here")
+            options.here = true;
+        else if (s == "b" || s == "block")
+            options.block = true;
+        else // "h", "help", invalid
```
myk002, line 395: this provides a getopt-like interface, but is not getopt. Could you call out to Lua and use `argparse` instead? See `tailor.cpp` and `tailor.lua` for an example:

```cpp
    bool show_help = false;
    if (!Lua::CallLuaModuleFunction(out, "plugins.tailor", "parse_commandline", parameters, 1,
            [&](lua_State *L) {
                show_help = !lua_toboolean(L, -1);
            })) {
        return CR_FAILURE;
    }

    return show_help ? CR_WRONG_USAGE : CR_OK;
```

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
 
-Specify the ``max`` keyword to pack more grass/moss onto a tile than what the
-game normally allows. This will give your grazers extra chewing time.
+Regrasses the entire map by default.
```
myk002, line 19: I'd clarify here that by default, only areas that had grass to begin with are repopulated

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+
+``debugfilter set Debug regrass log`` can be used to figure out why regrass
+is failing on a tile. (Avoid regrassing the entire map with this enabled, as it
+will make the game unresponsive and flood the console for several minutes!)
+Disable with ``debugfilter set Info regrass log``.
```
myk002, line 24: this should go in a "Troubleshooting" section at the end of the help doc

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+
+Options
+-------
+
+``h``, ``help``
+    Display this help screen.
```
myk002, line 30: we don't usually document tool-specific help options since the canonical way to get help for a command is  of the form: `help regrass`

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+``m``, ``max``
+    Maxes out every grass type in the tile, giving extra grazing time.
+    Not normal DF behavior. Tile will appear to be the first type of grass
+    present in the map block until that is depleted, moving on to the next type.
+    When this option isn't used, a grass type will be selected randomly instead.
```
myk002, line 35: randomly selected *if* there is no grass event already on the map block. otherwise, it will reuse the existing event, right?

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+is failing on a tile. (Avoid regrassing the entire map with this enabled, as it
+will make the game unresponsive and flood the console for several minutes!)
+Disable with ``debugfilter set Info regrass log``.
+
+Options
+-------
+
+``h``, ``help``
+    Display this help screen.
+``m``, ``max``
```
myk002, line 31: short options should be preceded by a single dash, long options with a double dash (getopt style)

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+    Maxes out every grass type in the tile, giving extra grazing time.
+    Not normal DF behavior. Tile will appear to be the first type of grass
+    present in the map block until that is depleted, moving on to the next type.
+    When this option isn't used, a grass type will be selected randomly instead.
+``n``, ``new``
+    Adds biome-compatible grass types that were not originally present in the
+    map block. Allows regrass to work in blocks that never had any grass to
+    begin with. Will still fail in incompatible biomes.
```
myk002, line 39: can there be a `-f`, `--force` option to add grass despite being in an incompatible biome? some players have requested that this tool be able to add grass to, for example, mountain biomes

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+``a``, ``ashes``
+    Regrass tiles that've been burnt to ash.
+``u``, ``mud``
+    Converts non-smoothed, mud-spattered stone into grass. Valid for layer stone,
+    obsidian, and ore.
+``p``, ``here``
+    Only regrass the tile selected by the keyboard cursor.
```
myk002, line 46: will this assume `-ua`? if players are choosing a specific tile, they probably want it specifically regrassed

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,56 @@ regrass
...
+``b``, ``block``
+    Only regrass the map block containing the keyboard cursor.
```
myk002, line 48: maybe add a suggestion to run `devel/block-borders` to see where the map block boundaries are

docs/plugins/regrass.rst
myk002: actually, this could be a positional parameter like `dig-now` uses to define the bounding box: `[<pos> [<pos>]]` where `<pos>` can be the keyword `here` (functionality provided by `argparse.coord`)

plugins/regrass.cpp
Bumber64: Can Lua set the values of the `regrass_options` struct directly or do I need to workaround somehow?

docs/plugins/regrass.rst
Bumber64: What type of grass should it use? User-specified?

plugins/regrass.cpp
myk002: you can either pass the opts to Lua and modify the fields directly (like `prospector` and `blueprint` do) or you can call back into cpp from the commandline parser function according to the values on the commandline like `stockpiles` does.

After working with both styles, I prefer the latter, but both approaches work.

docs/plugins/regrass.rst
myk002: optionally user specified would be a nice to have, but at least random. I'd think the common case here would be "I want some grass because my grazers are starving"

plugins/regrass.cpp
```diff
@@ -1,8 +1,18 @@
+#include "Debug.h"
+#include "modules/Gui.h"
+#include "modules/Maps.h"
 #include "PluginManager.h"
 #include "TileTypes.h"
```
myk002, line 5:
```suggestion
#include "PluginManager.h"
#include "TileTypes.h"

#include "modules/Gui.h"
#include "modules/Maps.h"
```

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,62 @@ regrass
...
+    Regrass tiles that've been burnt to ash.
+``-u``, ``--mud``
+    Converts non-smoothed, mud-spattered stone into grass. Valid for layer stone,
+    obsidian, and ore.
+``-h``, ``--here``
+    Only regrass the tile selected by the keyboard cursor. Automatically enables
+    ``--ashes`` and ``--mud``.
+``-b``, ``--block``
+    Only regrass the map block containing the keyboard cursor.
+    ``devel/block-borders`` can be used to visualize map blocks.
```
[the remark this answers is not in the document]
Bumber64: The highlight just disappears when I try that.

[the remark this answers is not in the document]
myk002: single ticks make a hyperlink

docs/plugins/regrass.rst
```diff
@@ -5,15 +5,101 @@ regrass
...
+    90, refilling existing and depleted grass.
+``regrass 0,0,100 19,19,119 --ashes --mud``
+    Regrass tiles in the 20x20x20 cube defined by the coords, refilling
+    existing and depleted grass, and converting ashes and muddy stone (if
+    respective blocks ever had grass.)
+``regrass 10,10,100 -baudnm``
+    Regrass the block that contains the given coord; converting ashes, muddy
+    stone, and tiles under buildings; adding all compatible grass types, and
+    filling each grass type to max.
+``regrass -f ""``
```
myk002, line 85: I realize this was borne from the lack of optional optargs, but I'm not a fan of this syntax. This is too novel compared to other tools.

How about if `-f`, `--force` takes zero parameters and a separate `--grass-type=<type>` allows the player to (optionally) specify the grass type?

plugins/regrass.cpp
```diff
@@ -12,109 +27,553 @@ using namespace DFHack;
...
+    {   // Iterate all blocks in z-level range
+        for (int32_t x = 0; x < world->map.x_count_block; x++)
+        {
+            for (int32_t y = 0; y < world->map.y_count_block; y++)
+            {
+                auto block = Maps::getBlock(x, y, z);
+                if (block)
+                    count += regrass_block(options, block);
+                else // Probably below HFS
+                    Core::print("No getBlock(%d, %d, %d)! Skipping.\n", x, y, z);
```
myk002, line 433: usage of `Core::print` will cause output to fail to be captured by streams (e.g. `gui/launcher`). Normally, plugins will pass the `color_ostream & out` output sink to all the internal functions for debug output. This goes for the DEBUG statements as well. The `out` param should be passed to those as well so their output can get captured.

myk002 (commented):
This is beautiful. thanks for adding this!

plugins/regrass.cpp
```diff
@@ -12,109 +27,553 @@ using namespace DFHack;
...
+                DEBUG(log).print("Valid tile: Tree\n");
+                return true; // Is main tile
+            }
+        }
+
+        DEBUG(log).print("Invalid tile: Tree\n");
+        return false; // Not main tile
+    }
+    else if (des.bits.flow_size > (des.bits.liquid_type == tile_liquid::Magma ? 0 : 3))
+    {   // Under water/magma (df::plant_raw::shrub_drown_level is usually 4)
```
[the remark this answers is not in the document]
Bumber64: Well, it won't grow there, at least.

plugins/regrass.cpp
```diff
@@ -12,109 +27,553 @@ using namespace DFHack;
...
+        DEBUG(log).print("Invalid tile: Liquid\n");
+        return false;
+    }
+    else if (shape != tiletype_shape::FLOOR &&
+        shape != tiletype_shape::RAMP &&
+        shape != tiletype_shape::STAIR_UP &&
+        shape != tiletype_shape::STAIR_DOWN &&
+        shape != tiletype_shape::STAIR_UPDOWN)
```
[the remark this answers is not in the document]
Bumber64: Saplings and shrubs are handled by `tiletype_material::PLANT` at the top.

Decided to ignore boulders and pebbles for now because they're unique features. You can smooth and build over them first if you want to turn them into grass. I could make it an option. (I'll probably end up reverse engineering at some point to verify how DF actually handles all these various cases, and make that the default.)

plugins/regrass.cpp
```diff
@@ -12,109 +27,553 @@ using namespace DFHack;
...
+        }
+
+        for (auto plant : mbc->plants)
+        {
+            if (plant->pos == p)
+            {
+                DEBUG(log).print("Valid tile: Tree\n");
+                return true; // Is main tile
+            }
+        }
```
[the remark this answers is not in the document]
Bumber64: Yeah. Just realized I'm going to have to check adjacent tiles for thick trees, though, so I'm going to have to use the original version anyway.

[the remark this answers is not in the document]
Bumber64: Doesn't look like `getPlantAtTile` is really used anywhere else. Do you still want the extra option?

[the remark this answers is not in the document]
myk002: Yeah, it should keep the existing behavior by default. It's used by `Gui::getAnyPlant` for getting the plant under the keyboard cursor

plugins/regrass.cpp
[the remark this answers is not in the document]
myk002: got it. SGTM

docs/plugins/regrass.rst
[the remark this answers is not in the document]
myk002: when you say the highlight disappears, where are you seeing that? when Sphinx compiles the docs, it has full knowledge of the doc tree, including scripts. If there is text surrounded by single ticks and the anchor cannot be found, then sphinx will fail to build the docs, so if the docs are succeeding with single ticks, then it's probably working correctly.

plugins/regrass.cpp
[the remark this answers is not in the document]
Bumber64: `getAnyPlant` is going to use the branch version, and `regrass` is going to use the branch version (because of 2x2 trunks.)

There isn't anywhere that would currently use the skip branch version.

Bumber64:
pre-commit.ci autofix

plugins/regrass.cpp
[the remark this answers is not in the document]
Bumber64: Actually, looks like some boulders on my embark have grass under them? I'll do that, then.

docs/plugins/regrass.rst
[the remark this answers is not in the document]
Bumber64: https://github.com/Bumber64/dfhack/blob/Regrass/docs/plugins/regrass.rst

devel/block-borders is in plain text on my browser.

plugins/regrass.cpp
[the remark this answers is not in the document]
myk002: Got it. Yeah if everyone is using the same version, no need to parameterize

docs/plugins/regrass.rst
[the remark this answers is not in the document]
Bumber64: I don't see anything like that. `docs` doesn't seem to produce any artifacts.

myk002 (approved):

docs/plugins/regrass.rst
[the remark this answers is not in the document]
myk002: This one:
image

Issue #898, opened 2016-04-16
createitem and gui/create-item don't set plant growth subtypes properly


lethosor:
Reported by @kane-t at http://www.bay12forums.com/smf/index.php?topic=139553.msg6939738#msg6939738 :

> Noticed a problem with the gui/create-item plugin: it doesn't correctly set the subtype for plant growths.  If you try to create an apple, for instance (plant growth -> plants -> apple tree -> fruit) it'll leave the subtype as -1, which results in a glitched item with no name or properties.  The correct subtype for fruits should be 2.

quietust:
The "createitem" plugin similarly needs special logic to treat the material subtype as a plant growth ID.

lethosor:
Confirmed in 0.47.04-r2. In addition, the wrong tile is used for created growths, so setting `item_plant_growthst.growth_print` may be necessary. 
Relevant forum discussion: http://www.bay12forums.com/smf/index.php?topic=164123.msg8137432#msg8137432 
(https://github.com/DFHack/df-structures/issues/403 is somewhat related)

lethosor:
`createitem` supports this as of #1659, and `ItemTypeInfo` has slightly-improved support with c9609ad5.

Leaving this open until `gui/create-item` and `Items::createItem` also have support.

ab9rf:
#5707 resolved the last open item for this issue

PR #4488, opened 2024-04-21
Make Items::createItem handle multiple output

### LINKED ISSUE #2063: tweak/reaction-gloves interacts poorly with Items::createItem and its callers
If the "reaction-gloves" Tweak is enabled, then _any_ custom reaction which tries to produce a single GLOVES item will actually create two or more (with handedness flags automatically set according to the body plan of whoever's performing the reaction).

However, the `Items::createItem()` function only ever expects to receive *one* item - if it gets more than one, it errors out and leaves the items floating in limbo (rather than moving them to the floor at the feet of the specified unit).

The scripts `gui/create-item` and `modtools/createitem` _also_ expect to get a single item ID back from this function, and they contain extra logic to create matching gloves with appropriate handedness.

Ideally, `Items::createItem` should return a `std::vector<df::item*>` (which Lua can hopefully convert into a native array) rather than just a single item ID (or a list of IDs), since it would save consumers of that function the trouble of having to call `df.items.find(...)` on those IDs (which they currently all do).

quietust:
This changes the function signature to return a list of created items via the first parameter and a boolean "success" code. It also moves the "unit" parameter to the beginning and adds "growth_print" to the end (which is required for PLANT_GROWTH items).

Should fix #2063

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -57,7 +57,6 @@ Template for new versions:
 
 ## Fixes
 - `zone`: fix display of distance from cage for small pets in assignment dialog
-- Fixed misidentification of visitors from your own civ as residents; affects all tools that iterate through citizens/residents
 - `modtools/create-item`: now functions properly when the ``reaction-gloves`` tweak is active
 
 ## Misc Improvements
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -57,17 +57,21 @@ Template for new versions:
 
 ## Fixes
 - `zone`: fix display of distance from cage for small pets in assignment dialog
+- Fixed misidentification of visitors from your own civ as residents; affects all tools that iterate through citizens/residents
+- `modtools/create-item`: now functions properly when the ``reaction-gloves`` tweak is active
 
 ## Misc Improvements
 - `caravan`: display who is in the cages you are selecting for trade and whether they are hostile
 - `regrass`: Can now add grass to stairs, ramps, ashes, buildings, muddy stone, shrubs, and trees
 - `regrass`: Can now restrict regrass effect to specified tile, block, cuboid, or z-levels
 - `regrass`: Can now add grass in map blocks where there hasn't been any, force specific grass type
+- `createitem`: refactored to use ``Items::createItem()`` API function
 
 ## Documentation
 
 ## API
 - ``dfhack.items.getReadableDescription()``: easy API for getting a human-readable item description with useful annotations and information (like tattered markers or who is in a cage)
+- ``Items::createItem``: now returns a list of item pointers rather than a single ID, moved creator parameter to beginning, added growth_print and no_floor parameters at end
 
 ## Lua
 - ``dfhack.internal.setClipboardTextCp437Multiline``: for copying multiline text to the system clipboard
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -83,6 +83,7 @@ distribution.
 #include "df/inorganic_raw.h"
 #include "df/item.h"
 #include "df/itemdef.h"
+#include "df/item_type.h"
 #include "df/job.h"
 #include "df/job_item.h"
 #include "df/job_material_category.h"
@@ -2288,7 +2289,6 @@ static const LuaWrapper::FunctionReg dfhack_items_module[] = {
     WRAPM(Items, getItemBaseValue),
     WRAPM(Items, getValue),
     WRAPM(Items, isRequestedTradeGood),
-    WRAPM(Items, createItem),
     WRAPM(Items, checkMandates),
     WRAPM(Items, canTrade),
     WRAPM(Items, canTradeWithContents),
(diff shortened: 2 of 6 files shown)


library/LuaApi.cpp
```diff
@@ -2358,12 +2358,26 @@ static int items_moveToBuilding(lua_State *state)
     return 1;
 }
 
+static int items_createItem(lua_State *state)
```
myk002, line 2361: needs doc updates in `Lua API.rst`

library/modules/Items.cpp
```diff
@@ -2133,14 +2133,17 @@ int32_t Items::createItem(df::item_type item_type, int16_t item_subtype, int16_t
...
 
     DEBUG(items).print("produced %zd items\n", out_items.size());
-    if ( out_items.size() != 1 )
-        return -1;
 
     for (size_t a = 0; a < out_items.size(); a++ ) {
+        // Plant growths need a valid "growth print", otherwise they behave oddly
+        auto growth = virtual_cast<df::item_plant_growthst>(out_items[a]);
+        if (growth)
+            growth->growth_print = growth_print;
```
myk002, line 2141: does this fix https://github.com/DFHack/dfhack/issues/898 ?

library/modules/Items.cpp
```diff
@@ -2089,7 +2090,7 @@ int Items::getValue(df::item *item, df::caravan_state *caravan)
     return value;
 }
 
-int32_t Items::createItem(df::item_type item_type, int16_t item_subtype, int16_t mat_type, int32_t mat_index, df::unit* unit) {
+bool Items::createItem(std::vector<df::item *> &out_items, df::unit* unit, df::item_type item_type, int16_t item_subtype, int16_t mat_type, int32_t mat_index, int32_t growth_print) {
```
myk002, line 2093: can the `createitem` plugin call this API as well so we don't duplicate code?

library/modules/Items.cpp
```diff
@@ -2133,14 +2133,17 @@ int32_t Items::createItem(df::item_type item_type, int16_t item_subtype, int16_t
     delete prod;
 
     DEBUG(items).print("produced %zd items\n", out_items.size());
```
myk002, line 2135: since we're using the passed-in vector directory, should we clear it before passing it to `produce`?

edit: yes we totally should since we're using the size to determine success for the return value

library/modules/Items.cpp
```diff
@@ -2133,14 +2133,17 @@ int32_t Items::createItem(df::item_type item_type, int16_t item_subtype, int16_t
...
-    if ( out_items.size() != 1 )
-        return -1;
 
     for (size_t a = 0; a < out_items.size(); a++ ) {
+        // Plant growths need a valid "growth print", otherwise they behave oddly
+        auto growth = virtual_cast<df::item_plant_growthst>(out_items[a]);
+        if (growth)
+            growth->growth_print = growth_print;
+
         out_items[a]->moveToGround(unit->pos.x, unit->pos.y, unit->pos.z);
```
myk002, line 2143: should probably use the `pos` variable calculated above, just in case the `unit->pos` values are invalid

library/modules/Items.cpp
quietust: Not by itself - the relevant Lua scripts also need to be updated to pass the correct value here, and I don't know them well enough to do that myself.

library/modules/Items.cpp
quietust: The `createitem` plugin has separate options for placing the items into containers/buildings, while this function always places them on the floor. Unless we want this function to provide *all* of the functionality of the plugin, they should stay separate.

myk002: I think the placement could be kept separate, but can the core creation logic be factored out?

library/modules/Items.cpp
myk002: what is the correct value?

myk002: can this code be refactored to be included in Items::createItem?
```cpp
        for (size_t i = 0; i < world->raws.plants.all.size(); i++)
        {
            string growths = "";
            df::plant_raw *plant = world->raws.plants.all[i];
            if (plant->id != tokens[0])
                continue;

            for (size_t j = 0; j < plant->growths.size(); j++)
            {
                df::plant_growth *growth = plant->growths[j];
                growths += " " + growth->id;
                if (growth->id != tokens[1])
                    continue;

                // Plant growths specify the actual item type/subtype
                // so that certain growths can drop as SEEDS items
                item_type = growth->item_type;
                item_subtype = growth->item_subtype;
                mat_type = growth->mat_type;
                mat_index = growth->mat_index;

                // Try and find a growth print matching the current time
                // (in practice, only tree leaves use this for autumn color changes)
                for (size_t k = 0; k < growth->prints.size(); k++)
                {
                    df::plant_growth_print *print = growth->prints[k];
                    if (print->timing_start <= *cur_year_tick && *cur_year_tick <= print->timing_end)
                    {
                        growth_print = k;
                        break;
                    }
                }
                // If we didn't find one, then pick the first one (if it exists)
                if (growth_print == -1 && growth->prints.size() > 0)
                    growth_print = 0;
                break;
            }
            if (mat_type == -1)
            {
                if (tokens[1].empty())
                    out.printerr("You must also specify a growth ID.\n");
                else
                    out.printerr("The plant you specified has no such growth!\n");
                out.printerr("Valid growths:%s\n", growths.c_str());
                return CR_FAILURE;
            }
        }
```

myk002: like, could the discovery of the value of growth_print be moved into Items::createItem so it doesn't have to be passed in?

quietust: I'd rather not increase the scope of this PR further than necessary.

library/modules/Items.cpp
quietust: Done.

library/modules/Items.cpp
myk002: would this be possible in a future PR? do we have enough information with the other parameters (item_type, item_subtype, mat_type, mat_index) to synthesize the growth print?

docs/changelog.txt
```diff
@@ -57,17 +57,21 @@ Template for new versions:
 
 ## Fixes
 - `zone`: fix display of distance from cage for small pets in assignment dialog
+- Fixed misidentification of visitors from your own civ as residents; affects all tools that iterate through citizens/residents
```
myk002, line 60:
```suggestion
```

myk002: this line was from a previous version

myk002 (approved):

PR #4497, opened 2024-04-21
Add types to argparse and alt_getopt

vallode:
This PR adds type annotations for argparse (and to alt_getopt for further support) and a few odd functions that were beneficial.

- If modifying alt_getopt is not wanted, we can do without the changes there
- alt_getopt has dead code in the `get_opts` function, `optargs` are unused
- `getCursorPos` uses the df-structures global even though it doesn't really fit there, this raises a topic about creating separate types for things like `coords` that can be re-used outside of the context of structures
- I've chosen to namespace classes in modules `---@class argparse.class` and will likely extend that to the structures, as well as other modules, in the future.


library/lua/3rdparty/alt_getopt.lua
```diff
@@ -57,11 +60,16 @@ local function get_opt_map(opts)
     return options
 end
 
+---@param opt string|integer
```
myk002, line 63: opt can only be a string at this point (but I see that the more permissive type might be needed because it can be string|integer in other contexts)

library/lua/3rdparty/alt_getopt.lua
```diff
@@ -45,6 +45,9 @@
 
 local _ENV = mkmodule('3rdparty.alt_getopt')
 
+---@nodiscard
+---@param opts string
+---@return table<string, string|integer>
```
myk002, line 50: table value is always an integer (but I see that the more permissive type might be needed because it can be string|integer in other contexts)

library/lua/3rdparty/alt_getopt.lua
```diff
@@ -80,14 +88,26 @@ local function canonicalize(options, opt)
...
+---@return boolean
 local function has_arg(options, opt)
     return options[canonicalize(options, opt)] == 1
 end
 
 -- returns vectors for opts, optargs, and nonoptions
+---@nodiscard
+---@param args string[] e.g, { ... }
+---@param sh_opts string e.g., 'ak:hv'
+---@param long_opts table<string, string|integer>
```
myk002, line 107: isn't this a table<string, string>?

library/lua/3rdparty/alt_getopt.lua
```diff
@@ -159,6 +179,12 @@ end
 
 -- returns a map of options to their optargs (or 1 if the option doesn't take an
 -- argument), and a vector for nonoptions
+---@nodiscard
+---@param args string[]
+---@param sh_opts string
+---@param long_opts string[]
+---@return table
```
myk002, line 186: specifically, table<string, string|integer>, right?

library/lua/argparse.lua
```diff
@@ -4,6 +4,10 @@ local _ENV = mkmodule('argparse')
 
 local getopt = require('3rdparty.alt_getopt')
 
+---@nodiscard
+---@param args string[] Most commonly `{ ... }`
+---@param validArgs table<string, integer> Use `utils.invert`
```
myk002, line 9: the value of this table just has to be truthy. maybe use boolean instead of integer

library/lua/argparse.lua
```diff
@@ -4,6 +4,10 @@ local _ENV = mkmodule('argparse')
...
+---@return table<string, string[]|nil>
```
myk002, line 10: I think this is a table<string, string|string[]>

library/lua/argparse.lua
```diff
@@ -58,25 +62,34 @@ function processArgs(args, validArgs)
     return result
 end
 
+---@class argparse.OptionAction
+---@field [1] string|nil Short option (eg. "q")
+---@field [2] string|nil Long option (eg. "quiet")
+---@field handler fun(optarg?: string)
+---@field hasArg boolean|nil
```
myk002, line 69: do we have to specify |nil here? nil is just a falsey boolean

library/lua/argparse.lua
```diff
@@ -141,7 +162,13 @@ function stringList(arg, arg_name, list_length)
     return list
 end
 
+---@nodiscard
+---@param arg string
+---@param arg_name? string
+---@param list_length? integer
+---@return integer[]
 function numberList(arg, arg_name, list_length)
+    ---@type integer[]
```
myk002, line 171: ugh, should we just declare a new output table so we don't have to live this lie?

library/lua/argparse.lua
```diff
@@ -153,6 +180,10 @@ function numberList(arg, arg_name, list_length)
     return strings
 end
 
+---@nodiscard
+---@param arg integer
```
myk002, line 184: arg is string|integer (but is likely to be a string)

library/lua/argparse.lua
```diff
@@ -153,6 +180,10 @@ function numberList(arg, arg_name, list_length)
...
+---@param arg_name? string
+---@return number
```
myk002, line 186: specifically an integer, if that matters

library/lua/argparse.lua
```diff
@@ -162,6 +193,10 @@ function positiveInt(arg, arg_name)
     return val
 end
 
+---@nodiscard
+---@param arg integer
```
myk002, line 197: string|integer

library/lua/argparse.lua
```diff
@@ -162,6 +193,10 @@ function positiveInt(arg, arg_name)
...
+---@param arg_name? string
+---@return number
```
myk002, line 199: integer

library/lua/argparse.lua
```diff
@@ -171,6 +206,11 @@ function nonnegativeInt(arg, arg_name)
     return val
 end
 
+---@nodiscard
+---@param arg string|'here'
+---@param arg_name? string
+---@param skip_validation? boolean
+---@return global.T_cursor|coord
```
myk002, line 213: always a coord, never a T_cursor

library/lua/argparse.lua
vallode: Sure can do!

library/lua/argparse.lua
vallode: If we do that the return value of `utils.invert` will be incompatible in LuaLS's eyes. We could do "any" and mention in the comment it has to be truthy?

Just to be clear:

```lua
---@param truthy boolean
function foo(truthy) end

local bar = foo("") -- Shows a warning
local caz = foo(1) -- Shows a warning
```

library/lua/3rdparty/alt_getopt.lua
vallode: `processArgsGetopt` assigns a truthy 1 or 0 to it. So I guess it kind of like the other "just has to be a truthy" fields.

library/lua/3rdparty/alt_getopt.lua
vallode: Yeah this is one of a few "just has to be truthy" fields.

library/lua/3rdparty/alt_getopt.lua
vallode: As above, I believe.

myk002 (approved):

PR #4498, opened 2024-04-21
Fix dump_df_state fullscreen_state

vallode:
Looks like the JSON dump was referring to a non-existent fullscreen_state property in structures.

### DIFF
--- a/ci/test.lua
+++ b/ci/test.lua
@@ -681,7 +681,7 @@ local function dump_df_state()
         enabler = {
             fps = df.global.enabler.fps,
             gfps = df.global.enabler.gfps,
-            fullscreen_state = df.global.enabler.fullscreen_state.whole,
+            fullscreen_state = df.global.enabler.fullscreen_state.fullscreen,
         },
         gps = {
             dimx = df.global.gps.dimx,


ci/test.lua
```diff
@@ -681,7 +681,7 @@ local function dump_df_state()
         enabler = {
             fps = df.global.enabler.fps,
             gfps = df.global.enabler.gfps,
-            fullscreen_state = df.global.enabler.fullscreen_state.whole,
+            fullscreen_state = df.global.enabler.fullscreen_state.fullscreen,
```
myk002, line 684: the `fullscreen_state` field is defined in the xml as:
```xml
        <bitfield name='fullscreen_state' base-type='uint8_t'>
            <flag-bit name='fullscreen'/>
            <flag-bit name='exclusive'/>
        </bitfield>
```
bitfields have a generated "whole" field to represent then entirety of the flag bits as an integer. this proposed change would result in representing only the first bit (as a boolean).

vallode:
Closed, added documentation on `whole` property over at https://github.com/DFHack/dfhack/pull/4500

ci/test.lua
vallode: Thanks, myk! Sorry!

PR #4540, opened 2024-05-01
Fix dangling reference warning in buildingplan plugin

### LINKED ISSUE #4534: Failed to build buildingplan with GCC 14 (compiler bug?)
Compiling 3402d1410 on linux (Fedora 40) with gcc (gcc (GCC) 14.0.1 20240411 (Red Hat 14.0.1-0)), I get:

```
dfhack/plugins/buildingplan/buildingplan.cpp: In function 'int scanAvailableItems(DFHack::color_ostream&, df::enums::building_type::building_type, int16_t, int32_t, int, bool, bool, HeatSafety*, std::vector<int>*, std::map<DFHack::MaterialInfo, int>*)':
dfhack/plugins/buildingplan/buildingplan.cpp:634:11: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  634 |     auto &job_items = get_job_items(out, key);
      |           ^~~~~~~~~
dfhack/plugins/buildingplan/buildingplan.cpp:634:36: note: the temporary was destroyed at the end of the full expression 'get_job_items((* & out), key)'
  634 |     auto &job_items = get_job_items(out, key);
      |                       ~~~~~~~~~~~~~^~~~~~~~~~
dfhack/plugins/buildingplan/buildingplan.cpp: In function 'int countAvailableItems(DFHack::color_ostream&, df::enums::building_type::building_type, int16_t, int32_t, int)':
dfhack/plugins/buildingplan/buildingplan.cpp:737:11: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
  737 |     auto &job_items = get_job_items(out, key);
      |           ^~~~~~~~~
dfhack/plugins/buildingplan/buildingplan.cpp:737:36: note: the temporary was destroyed at the end of the full expression 'get_job_items((* & out), key)'
  737 |     auto &job_items = get_job_items(out, key);
      |                       ~~~~~~~~~~~~~^~~~~~~~~~
```

After a quick look at the code, it seems the reference comes from a cache, it may be a false positive. Can the author (@myk002?) confirm all is right?

cvuchener:
fix #4534

### DIFF
--- a/plugins/buildingplan/buildingplan.cpp
+++ b/plugins/buildingplan/buildingplan.cpp
@@ -15,6 +15,12 @@
 #include "df/organic_mat_category.h"
 #include "df/world.h"
 
+#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 14)
+#define NO_DANGLING_REFERENCE [[gnu::no_dangling]]
+#else
+#define NO_DANGLING_REFERENCE
+#endif
+
 using std::map;
 using std::set;
 using std::string;
@@ -75,6 +81,7 @@ static int get_num_filters(color_ostream &out, BuildingTypeKey key) {
     return num_filters;
 }
 
+NO_DANGLING_REFERENCE
 static const vector<const df::job_item *> & get_job_items(color_ostream &out, BuildingTypeKey key) {
     if (job_item_cache.count(key))
         return job_item_cache[key];


cvuchener:
A version check is needed in the macro definition. :(

cvuchener:
updated

I don't like this compiler/version check, it's too complex. Can it be better? Maybe you want to wait for your tool chain to update to gcc 14 before fixing this issue.

ab9rf:
"[All attributes unknown to an implementation are ignored without causing an error](https://en.cppreference.com/w/cpp/language/attributes)" so if an undefined attribute is causing a compilation error this itself is a compiler bug.

ab9rf:
We are not likely to move to gcc 14 any time soon due to basically nonexistent support for gcc 14's runtime in widely available distributions. Building DFHack with a version of gcc other than the one DF is built with (currently gcc 11) is unsupported.

myk002:
> Yes, forget about this. I'll keep my workaround in my cmake configuration.
I think this can still be a useful addition. we don't *distribute* DFHack built with newer compilers, but we do *build* with newer compilers. We build with gcc-12 in CI, for example, to catch warnings that gcc-10 misses.

ab9rf:
I would use the "pragma" approach to disable the warnings instead of using an attribute, since (I sincerely hope) that no compiler author is doofus enough to issue even so much as a _warning_ for an unrecognized pragma. 🤦‍♀️

plugins/buildingplan/buildingplan.cpp
```diff
@@ -15,6 +15,12 @@
 #include "df/organic_mat_category.h"
 #include "df/world.h"
 
+#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 14)
```
myk002, line 18: could also be `#if __has_cpp_attribute(no_dangling)`, I think

myk002: as per https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/testsuite/g%2B%2B.dg/ext/attr-no-dangling4.C;h=e910723d9852f92914bf84130a80106f56e50ee1;hb=c7607c4cf18986025430ca8626abfe56bfe87106

myk002, line 18:
```suggestion
#if __has_cpp_attribute(no_dangling)
```

PR #4546, opened 2024-05-03
Logistics auto-forbid/claim functionality and stockpile overlay

xBlackTalonX:


plugins/logistics.cpp
```diff
@@ -391,7 +439,7 @@ static const struct BadFlags {
     uint32_t get_bad_flags() {
         df::item_flags flags;
         #define F(x) flags.bits.x = true;
-        F(forbid); F(garbage_collect); F(hostile); F(on_fire);
+        F(garbage_collect); F(hostile); F(on_fire);
```
myk002, line 442: this will cause forbidden items to start being processed by the other processors. I think you need to modify the other `can_designate` functions to compensate

plugins/lua/stockpiles.lua
```diff
@@ -508,6 +529,17 @@ function StockpilesOverlay:init()
                                 {label='Training', value=false}},
                         initial_option=false,
                         on_change=self:callback('toggleLogisticsFeature', 'train'),
+                    },  widgets.CycleHotkeyLabel{
+                        view_id='forbid',
+                        frame={t=2, l=0},
+                        auto_width=true,
```
myk002, line 535: this might end up with too small a size. CycleHotkeyLabels generally need an explicit width to be set in the frame

plugins/lua/stockpiles.lua
```diff
@@ -542,32 +574,51 @@ function StockpilesOverlay:onRenderFrame()
     local sp = dfhack.gui.getSelectedStockpile(true)
     if sp and self.cur_stockpile ~= sp then
         local config = logistics.logistics_getStockpileConfigs(sp.stockpile_number)[1]
+        print("onRenderFrame - logistics_GET StockpileConfigs()")
+        print("melt=",config.melt,"trade=",config.trade,"dump=",config.dump,"forbid=",config.forbid,"train=",config.train)
```
myk002, line 578: leftover debug print?

plugins/lua/stockpiles.lua
```diff
@@ -447,28 +447,49 @@ function StockpilesOverlay:init()
...
+            --     on_activate=do_import,
             -- },
             widgets.Panel{
                 frame={t=0, l=0},
                 subviews={
                     widgets.Label{
                         frame={t=0, l=0, h=1},
                         auto_height=false,
-                        text={'Designate items/animals brought to this stockpile for:'},
+                        text={'Automatically designate items/animals in this stockpile'},
```
myk002, line 490: The grammar of existing options doesn't line up. "Automatically designate items/animals in this stockpile Training" doesn't parse. What is wrong with the original wording?

plugins/lua/stockpiles.lua
```diff
@@ -542,32 +574,51 @@ function StockpilesOverlay:onRenderFrame()
...
         self.subviews.train:setOption(config.train == 1)
         self.cur_stockpile = sp
     end
 end
 
 function StockpilesOverlay:toggleLogisticsFeature(feature)
+    print("entering toggleLogisticsFeature - feature = ", feature)
+    --print("option value - self.subviews.forbid:getOptionValue() = ", self.subviews.forbid:getOptionValue())
+    --print("option value - self.subviews[feature]:getOptionValue() = ", self.subviews[feature]:getOptionValue())
+    --print("option index - self.subviews[feature].option_idx", self.subviews.forbid.option_idx)
```
myk002, line 592: (and below) leftover debug prints?

plugins/logistics.cpp
```diff
@@ -54,6 +54,7 @@ enum StockpileConfigValues {
     STOCKPILE_CONFIG_DUMP = 3,
     STOCKPILE_CONFIG_TRAIN = 4,
     STOCKPILE_CONFIG_MELT_MASTERWORKS = 5,
+    STOCKPILE_CONFIG_FORBID = 6,
```
myk002, line 57: existing savegames will have this variable set to -1, which this PR does not seem to handle gracefully. if the new config were two bools, they'd default to false, but an int needs to be initialized explicitly.

plugins/lua/stockpiles.lua
xBlackTalonX: You're right, I didn't feel strongly about this and will switch it back. I was playing with a few things when doing the layout because I was trying to come up with a way for it to work with some version of forbid/forbidden since I similarly wasn't liking it grammatically for the new option ("Designate items for unforbid/forbidden"). I had thought I was getting somewhere with "Automatically" because I kept thinking of auto-dump, auto-forbid, etc. But I've come up short and will keep the original language.

plugins/logistics.cpp
xBlackTalonX: Thanks - I'll take a closer look, in my testing I saw the negative 1 get cleaned up (and I thought that was due to the initialization in ensure_stockpile config - but thinking about it now I suspect that may not get called until after stockpiles.lua is trying to use the value with an invalid index...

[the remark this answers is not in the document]
xBlackTalonX: Agreed, this was temping to tackle but I'll leave it for another day.

plugins/logistics.cpp
xBlackTalonX: Here's what I have in mind, do you concur?
- forbidden items should **should be marked** for other operations
- forbidden items **should be unmarked** for other operations
- logistics stats should ignore forbidden items

Been slower than expected to get back to this but I still  hope to have this and the rest of the feedback fixed up tomorrow (US-Sunday).

plugins/lua/stockpiles.lua
xBlackTalonX: I expect to do this with `label_width` unless I'm missing some better method within the `frame` specifications? I was interpreting `frame` as the position but I believe I understand the width issue I need to fix.

plugins/lua/stockpiles.lua
xBlackTalonX: These are cleaned up along with some other improvements to readability that were intended to be in the first PR.

I did intentionally leave a couple new debug lines in logistics.cpp that seemed fitting and output there requires being enabled with `debugfilter` (there's no `debugfilter` functionality pattern already in place for lua, right?)

plugins/logistics.cpp
xBlackTalonX: Still reviewing/resolving this one as well.

plugins/logistics.cpp
myk002: I think the general expectation is that forbidden items will be untouched by all tools (other than the ones that specifically deal with forbidding/unforbidding).

stats can maybe call out forbidden items separately

plugins/lua/stockpiles.lua
myk002: label_width only applies to the label part of the widget. you need to set the frame width to be sure it's wide enough to handle all option labels.

plugins/logistics.cpp
xBlackTalonX: I realized I was overthinking this and imagining that toggling off melting would undesignate items formerly designated for melting. So as you suggested I just incorporated the forbid into `can_designate`

plugins/lua/stockpiles.lua
xBlackTalonX: Should be addressed now.

plugins/logistics.cpp
xBlackTalonX: I did this in `validate_stockpile_configs` as that seemed to address it but please let me know if you see a loophole or a better place to shore it up. The original behavior was dodging any negative effects because it was always looking for values of 1 or 2 without care for other values; but I agree that was not great to leave that way.

xBlackTalonX:
I'd welcome more feedback but I believe the comments have been addressed and its ready for another review.

xBlackTalonX:
found in testing that I needed to move the overlay down a bit to avoid overlap with the custom stockpile settings - pushed that tweak and ready for review

myk002:
will review tonight. Thanks!

docs/plugins/stockpiles.rst
```diff
@@ -113,7 +113,8 @@ Overlay
 
 This plugin provides a panel that appears when you select a stockpile via an
 `overlay` widget. You can use it to easily toggle `logistics` plugin features
-like autotrade, automelt, or autotrain. There are also buttons along the top frame for:
+like autotrade, automelt, autodump, autoforbid/unforbid, or autotrain. There
+are also buttons along the top frame for:
```
myk002, line 117:
```suggestion
like autotrade, automelt, or autotrain. There are also buttons along the top frame for:
```

myk002: these are just examples; no need to make the list complete

plugins/logistics.cpp
```diff
@@ -123,10 +129,13 @@ static void validate_stockpile_configs(color_ostream& out,
...
                 !c.get_bool(STOCKPILE_CONFIG_TRADE) &&
                 !c.get_bool(STOCKPILE_CONFIG_DUMP) &&
-                !c.get_bool(STOCKPILE_CONFIG_TRAIN))) {
+                !c.get_bool(STOCKPILE_CONFIG_TRAIN) &&
+                !(c.get_int(STOCKPILE_CONFIG_FORBID) > 0))) {
             to_remove.push_back(stockpile_number);
             continue;
         }
+        if (c.get_int(STOCKPILE_CONFIG_FORBID) == -1)
+            c.set_int(STOCKPILE_CONFIG_FORBID, 0);
```
myk002, line 138: This migration code should be in `plugin_load_site_data`, not here.

plugins/logistics.cpp
```diff
@@ -358,10 +358,11 @@ class TrainStockProcessor : public StockProcessor {
 
     bool can_designate(color_ostream& out, df::item* item) override {
         auto unit = get_caged_unit(item);
-        return unit &&
-            Units::isTamable(unit) &&
-            !Units::isDomesticated(unit) &&
-            !Units::isMarkedForTraining(unit);
+        return !item->flags.bits.forbid &&
```
myk002, line 361: I suggest moving the forbid check above the call to `get_caged_item` so we don't waste time on caged units if cage is forbidden anyway

plugins/logistics.cpp
```diff
@@ -382,6 +383,44 @@ class TrainStockProcessor : public StockProcessor {
...
+public:
+    ForbidStockProcessor(int32_t stockpile_number, bool enabled, ProcessorStats& stats)
+        : StockProcessor("forbid", stockpile_number, enabled, stats) { }
+
+    bool is_designated(color_ostream& out, df::item* item) override {
+        return item->flags.bits.forbid;
+    }
+
+    bool can_designate(color_ostream& out, df::item* item) override {
+        return true;
```
myk002, line 396: you cannot forbid artifacts. this should return `!is_artifact(item)`

myk002: (which the DumpStockProcessor should also be checking for -- could you add that check there too?)

plugins/logistics.cpp
```diff
@@ -181,16 +190,7 @@ DFhackCExport command_result plugin_onupdate(color_ostream &out) {
...
-        do_cycle(out, melt_count, trade_count, dump_count, train_count);
-        if (0 < melt_count)
-            out.print("logistics: marked %d item(s) for melting\n", melt_count);
-        if (0 < trade_count)
-            out.print("logistics: marked %d item(s) for trading\n", trade_count);
-        if (0 < dump_count)
-            out.print("logistics: marked %d item(s) for dumping\n", dump_count);
-        if (0 < train_count)
-            out.print("logistics: marked %d animal(s) for training\n", train_count);
+        logistics_cycle(out);
```
myk002, line 193: the reason this logic is kept separate from `logistics_cycle` is because `logistics_cycle` is called in response to manual triggers. We want this auto-running version to be quiet when it does nothing. we want the manually-called version to be clear and verbose when nothing happens.

plugins/lua/stockpiles.lua
```diff
@@ -545,6 +555,7 @@ function StockpilesOverlay:onRenderFrame()
         self.subviews.melt:setOption(config.melt == 1)
         self.subviews.trade:setOption(config.trade == 1)
         self.subviews.dump:setOption(config.dump == 1)
+        self.subviews.forbid:setOption(config.forbid+1)
```
myk002, line 558: this is a bit opaque. I suggest `self.subviews.forbid:setOption(config.forbid)` instead and changing the options in the widget to:
```lua
                        options={{label='Forbid', value='0'},
                                 {label='Forbid', value='1', pen=COLOR_LIGHTRED},
                                 {label='Unforbid', value='2', pen=COLOR_LIGHTBLUE}},
```

plugins/lua/stockpiles.lua
```diff
@@ -428,10 +428,10 @@ end
 StockpilesOverlay = defclass(StockpilesOverlay, overlay.OverlayWidget)
 StockpilesOverlay.ATTRS{
     desc='Shows a panel when a stockpile is selected for stockpile automation.',
-    default_pos={x=24, y=-6},
+    default_pos={x=5, y=-11},
```
myk002, line 431: This position covers the categories. I think it might be better closer to where it was.
image

however, with the increased height, in its old position, it now completely covers the "add stockpile" button, which is also not good.

what is the least bad option here? should we keep everything on two lines and make the widget even wider?

plugins/lua/logistics.lua
```diff
@@ -49,16 +51,34 @@ local function print_stockpile_data(data)
...
-    print(fmt:format('number', 'name', 'melt', 'melt items', 'trade', 'trade items', 'dump', 'dump items', 'train', 'train items'))
+    --          sp.#  __sp.name____________  __melt__  _trade__  __dump__  _train__  frbd  un/frbd
+    local fmt = '%6s  %-' .. name_len .. 's  %4s %10s  %5s %11s  %4s %10s  %6s %11s  %12s  %12s';
+    print(fmt:format('number', 'name',
+            'melt',  'melt items',
+            'trade', 'trade items',
+            'dump',  'dump items',
+            'train', 'train items',
+            'forbid items',
+            '[un][forbid]'))
```
myk002, line 62: why are these in the opposite order from the other columns?

plugins/lua/logistics.lua
```diff
@@ -91,6 +111,8 @@ local function print_status()
     print(('Total items marked for melting: %5d'):format(global_stats.total_melt))
     print(('Total items marked for trading: %5d'):format(global_stats.total_trade))
     print(('Total items marked for dumping: %5d'):format(global_stats.total_dump))
+    print(('Total items marked forbidden: %7d'):format(global_stats.total_forbid))
+    print(('Total items unforbidden: %12d'):format(global_stats.total_unforbid))
```
myk002, line 115: "unforbidden" sounds like an action. how about `Total items not marked forbidden`?

plugins/lua/logistics.lua
myk002, line 1: something seems off in the output here:
image

with the current code, the number of total dump items should equal the number of total forbid items, right?

plugins/logistics.cpp
xBlackTalonX: Oh! I didn’t know that, thanks!

plugins/logistics.cpp
xBlackTalonX: Ok, I’ll revert that.

It feels like there is a lot of duplicate code in a few places like this one example that could be refactored, I didn’t want to go for a bunch in this PR, but separately, do you believe it worth while to reduce the duplication in this or is there a better longer term spot for these features?

plugins/lua/stockpiles.lua
xBlackTalonX: Oh, I wanted to ask because I saw that other open PR about it, should I call it ‘claim’ instead of unforbid? I take it that’s the new ingame language used?

plugins/lua/logistics.lua
xBlackTalonX: I played with this too and the result came from,
1. Showing un-forbidden items / un-forbidden-able items is basically useless 
2. Showing stockpiles with auto-unforbid enabled is useful, same for auto-forbid
3. Showing forbidden / forbid-able items is semi-valuable
5. Space is limited… 

So to get all the underlined headers on a single line but also add all 4 new columns would’ve been low value and too many characters using up the limited horizontal space. A single [un][forbid] header was space conscious and I believed was appropriate since both can’t be enabled simultaneously. To my eye it looked better having both of those to the right of ‘forbid items’ - are you suggesting it be to the left of ‘forbid items’? I can try that.

Other thoughts when I was doing this, my [setting] concept was born out of trying to have a single heading for two checkboxes, perhaps it would be worth the horizontal space to make other headers follow that format?
[melt] [trade] [dump] [train] [un][forbid]

Or perhaps I’ve misunderstood your concern? I’ll post some comparison screenshots in discord of some options.

xBlackTalonX:
Thank you again for the feedback!! I expect I can tackle this tonight.

xBlackTalonX:
Oh, do you prefer I not force push? Do you squash when you merge? I’m open to your preferences.

plugins/lua/stockpiles.lua
xBlackTalonX: Or maybe it could stay open on the stockpile category settings screen but move locations since the stockpile button is already covered?

Any tips on resolution testing? I had tried 1920x1080 and 1024x768 and didn’t have the overlap shown in your screenshot, which appears to be 1920x1200? More vertical pixels and we have less vertical space! haha - or am I overlooking other UI scaling features?

myk002:
> Oh, do you prefer I not force push? Do you squash when you merge? I’m open to your preferences.

Force pushes do make the review harder to track because past comments get lost. We don't squash commit, but we also don't worry about messy commit histories either. When I do history spelunking, I look at the PR, not the commit.

plugins/lua/logistics.lua
myk002: I was fine with the column contents - I was just curious about the order. Other columns have the checkboxes before the numbers instead of after

xBlackTalonX: Ah, I see. I'll try that.

xBlackTalonX: image
image

So I remember why I changed the order now, but I don't like my reasoning. But since it was your actual question: 'forbid items' causes the horizontal distance between checkboxes and item counts to be larger than any other section and it starts to look like it is more associated with training than forbidden items. At one point I reversed the columns and removed the 'forbid' from 'forbid items' hoping to effectively label both 'forbid' and save on that horizontal space.

Thoughts on something closer to this?
image

xBlackTalonX: some of the spacing still needs tweaking in that last image, but curious for your thoughts on the concept which should be clear enough

myk002: looks good to me

xBlackTalonX: > edit: trade and train are both t -- maybe make the letter match the hotkey in the overlay?
You read my mind! I was going to ask what a good way to grab the user's configured hotkey was so I could do exactly that (in addition to matching the overlay, but I suppose that gets risky?). Any suggestion or example to point me towards? In the meantime I made it 'a' matching the existing overlay.

Meanwhile I haven't been able to stop tweaking it... but I think I'm basically done with the look/layout pending a few bugs to tackle and any further feedback:
image
See anything else you dislike or would prefer improved?

xBlackTalonX: Oh, that example doesn't have the capital [M] for masterworks melting enabled stockpile example, but that's there for both the melt column display and note (which only displays when it exists, per previous code).

xBlackTalonX: This is as additive as the game! Once again I couldn't leave the couple nitpicks I saw alone in the prior screenshot, so here's a refreshed version, 
image
Looking forward to any feedback!

Definitely going to work now! lol

myk002: > This is as additive as the game!
> Definitely going to work now! lol

welcome to my life : )

myk002: > a good way to grab the user's configured hotkey

You certainly *can* get the key that a keybinding resolves to (follow the code that displays hotkey hints to see how), but most DFHack hotkeys use `CUSTOM_<foo>` keybindings, which are not usually changed by the player. E.g. the hotkey for training is `CUSTOM_CTRL_A`, so it's perfectly reasonable for you to just use `a` as the marker

xBlackTalonX: > it's perfectly reasonable for you to just use `a` as the marker
I’ll leave my concept of a hotkey matching on my wishful-thinking backlog and just match letters with the existing vanilla+dfhack experience (i.e. use `[a]` for `autotrain` enabled).

…but just to clarify the idea (maybe this is a case of a bad idea with problems that stimulates better idea)…

Concept: retain DFHack `CUSTOM_CTRL_<letter>` hotkey pattern for Stockpiles Overlay ToggleHotkeyLabels, BUT dynamically change its `<letter>` to correspond to the currently configured vanilla DF hotkey letter for that matching vanilla designation. This applies most directly to melt, dump, and forbid.
Example: So when Vanilla DF hotkey for Dump designations is configured for `[i, d]` then Stockpile Overlay would match the letter `d` for dump and configure its auto-dump toggle hotkey with `[ctrl+d]`. And this same letter (lowercase) would also be what is displayed here in the logistics status output to denote which stockpiles have auto-dump enabled (`[d]`). That way, the mental “muscle memory” of dump = ‘d’ applies in all three experiences:
a.) Vanilla DF dump designation 
b.) Stockpile Overlay auto-dump
c.) Text output denoting dump, such as logistics status output
And this like-for-like letter-to-function experience would hold true even when the user customizes their DF keys (which is a popular DF Workshop item as you likely know and grudgingly deal with). The actual hotkey combination ([i, d] vs [ctrl+d]) would not be the same, just the letter would match.
Obvious Risk: (and presumably deal breaker) A user could configure letters for these functions which in vanilla DF that conflicts with something else in one of these DFHack experiences. This likelihood would then necessitate hotkey fallback logic to a set of unique defaults when a conflict exists. This becomes particularly high maintenance (and even confusing to a user experimenting with changing a particular DF hotkey). Especially if this pattern was extended to additional DFHack experiences. So… dead concept.

plugins/lua/logistics.lua
myk002: I think it's still valuable, at least for completeness sake. It's the complement to the forbidden count

plugins/lua/logistics.lua
myk002: > Obvious Risk
yeah

plugins/lua/stockpiles.lua
myk002: I find the best way to test is to have DF in windowed mode and then dynamically resize the window from very large to the smallest size (where DF starts scaling the graphics instead of reducing the column/row count)

plugins/lua/logistics.lua
xBlackTalonX: I think the trouble is that forbidden items, of which you have 17 in that stockpile, now return false to the `DumpStockpileProcessor`'s `can_designate()` method,
image

When the stockpile items are scanned by each processor those forbidden items don't add to either `can_designate_counts` or `designated_counts`
image

And when those `ProcessorStats` get passed over to logistics.lua it is the sum of those designated and can_designate values for each processor which is printed as the total designatable items in the column,
`local function get_dstat(stats) return ('%d/%d'):format(stats.designated, stats.designated + stats.can_designate) end`

So in the example of the dump column, it hasn't counted the forbidden items as designatable for dump, so that total is 17 short of the total items designatable for forbid, where they were included in the sum due to their designated state.

Therefore, I believe it's working as designed? But I'm not sure if it's too confusing for users. I kinda like the resulting information provided about how many items are in the stockpile and how many of them are forbidden or designated for melt, dump, or trade. If I've understood correctly, I think my two-cents is to keep it this way. But if you have ideas for something clearer to the user I'm open to changing it.

myk002: I think it's working as intended. Forbidden items are intended to be "invisible", and the behavior here is in line with that. They will be highlighted in the "marked forbidden" count of the output, and I think that is correct to do so.

plugins/logistics.cpp
xBlackTalonX: Thank for this feedback, I was both overlooking and not understanding `plugin_load_site_data` 

I've moved the same logic to ~line 180 within `plugin_load_site_data` but I'm still not sure I understand,
```if (c.key() == CONFIG_KEY) continue;```

I put the mitigation after that and before,
```watched_stockpiles.emplace(c.get_int(STOCKPILE_CONFIG_STOCKPILE_NUMBER),  c);```

It seems to work there. But I presume that it is like `migrate_old_keys` in that the logic would be unnecessary if v50.current saves are no longer loadable? Should I include a comment to this effect?

Comment (myk002) [reply plugins/logistics.cpp:138]: > `if (c.key() == CONFIG_KEY) continue;`

that line is there because I was silly with the naming and the plugin-wide config is also matched by CONFIG_KEY_PREFIX. We don't want to treat the plugin-wide config as a stockpile-specific config. They're different schemas.

File: plugins/lua/stockpiles.lua (hunk 12 above)
Comment (xBlackTalonX) [reply plugins/lua/stockpiles.lua:558]: is `:getOptionValue()` expensive? After changing to your suggestion this seems nice and simple,
image

But unless `:getOptionValue()` is expensive then it seems like we'd want to do the same for the other values? Once that was done, would we really even did to getStockpileConfigs (which does hit every watched stockpile when I'd think we just need this one)?

I also couldn't quite figure out how to pass the value and the subviews' `view_id` to the `on_change` callback function. Note that while all my coding abilities are _very very rusty_, this is my first time with lua (and I already love it and hate it! lol). I did see that it will call the callback with (newvalue, oldvalue) but whenever I tried to reference value or view_id I couldn't figure out how.

File: plugins/logistics.cpp (hunk 11 above)
Comment (xBlackTalonX) [reply plugins/logistics.cpp:193]: For instance, what are your thoughts on leaving the call to `logistics_cycle` in `plugin_onupdate` but making the `quiet` aspect an optional input parameter?
image

**Comment by myk002** [comment]
Is this PR still in progress? Please mark it as "Ready for review" when I should take another look at it

**Comment by xBlackTalonX** [comment]
Thanks for the poke - yes, still in progress.

I dabble with it every couple days - I have code for you to review but in
many ways I’ve dabbled for too long (changed too much) and I’m pretty sure
it would be easier to review as separate PRs. I was going to revert some of
my overly aggressive code format changes (that in the end weren’t improving
the readability like I thought they were when I did them), push it, and get
your opinion on breaking it up.

Other outstanding work is the overlay position. Did you have thoughts on
that moving between the two modes or hiding it when configuring stockpile
details?

**Comment by myk002** [comment]
yeah, I think it's ok to only display the overlay on the map screen and not the stockpile customization screen. How about here:
image

**Comment by xBlackTalonX** [comment]
My goodness is that ever atrocious lol - but yes, I was testing this poorly. For now I'm reverting it to what it was as the edge cases for that are far less worse imo.

Mostly looking for some feedback on if I should break this up or maybe what I need to scrap to get a first part merged (such as not exposing it in the UI until we have a better method).

**Comment by xBlackTalonX** [comment]
@myk002 Looking for feedback on breaking this up or getting some aspect of it merged while we work out other aspects like the overlay position. To that end, some notes below - probably missing some things, feel free to ask of course,

logistics.lua
Refactor:
- Removed extra loop through stockpiles to grab longest stockpile name length (moved to `getStockpileData` loop)
- Renamed some variables for consistency within their local scope to avoid multiple definition of 'name' and 'stat'
- Added new function group section-header comments
- Added comments to print_stockpile_data to assist understanding of table spacing and layout
- Precalculate 'designatable' when `getStockpileData` creates the 'data' table for easier accessibility

Style
- Changes usages of 'Forbid' to 'Claim' (command line works with claim or unforbid)
- Changed usages of 'Mark' to 'Designate'
- Grouped and styled 'logistics' plugin feature status output

Fixes:
- ensured data.{type}.enable and data.melt_masterworks get initialize to boolean

Features
- added forbid column for logistics forbid/claim
- added calculation and footer output of total number of designated and designatable items
- enhanced appearance and information conveyed within table header (and new footer)

Not ready to merge imo, but the functionality is there, just not sure about my lua stuff in particular, I don't know any coding norms and welcome feedback so I learn.

**Comment by myk002** [comment]
the list of changes looks good. after some more experimenting, I think the overlay would look good at 
```json
            "x": 5,
            "y": 44
```

I'll review the code itself when I get a chance (hopefully later today, maybe this weekend)

xBlackTalonX:
Did you see enough to know if you’d prefer a separate PR(s) for,
a.) Minor fixes to existing functionality
b.) The visual style changes
c.) The code refactoring
d.) Feature(s)

Once I got in too deep, I figured those being PR’d in combination sure made
it hard to review. But you’re a pro so many you don’t care. :-p

Just let me know, I could reorganize it tonight and make your review(s)
easier.

docs/changelog.txt
```diff
@@ -62,6 +62,7 @@ Template for new versions:
 ## Misc Improvements
 - `blueprint`: capture track carving designations in addition to already-carved tracks
 - `changevein`: affect connected veins even outside of the selected map block
+- `logistics`: new ability to automaticly forbid or claim items brought to a stockpile (or a dump within a stockpile)
```
myk002, line 65:
```suggestion
- `logistics`: new ability to automatically forbid or claim items brought to a stockpile (or a dump within a stockpile)
```

plugins/logistics.cpp
```diff
@@ -170,6 +177,8 @@ DFhackCExport command_result plugin_load_site_data(color_ostream &out) {
         auto& c = loaded_persist_data[idx];
         if (c.key() == CONFIG_KEY)
             continue;
+        if (c.get_int(STOCKPILE_CONFIG_FORBID) == -1) // remove this once saves from 50.08 are no longer compatible
```
myk002, line 180:
```suggestion
        if (c.get_int(STOCKPILE_CONFIG_FORBID) == -1) // remove this once saves from 51.01 are no longer compatible
```

plugins/logistics.cpp
```diff
@@ -391,7 +431,7 @@ static const struct BadFlags {
     uint32_t get_bad_flags() {
         df::item_flags flags;
         #define F(x) flags.bits.x = true;
-        F(forbid); F(garbage_collect); F(hostile); F(on_fire);
+        F(garbage_collect); F(hostile); F(on_fire);
```
myk002, line 434: bah, I didn't see that F(artifact) is already in BadFlags. You don't need to check for it in the processors. sorry!

plugins/logistics.cpp
```diff
@@ -722,9 +806,22 @@ static int logistics_getGlobalCounts(lua_State *L) {
     }
 
     size_t num_dump = 0;
+    size_t num_forbid = 0;
+    size_t num_claim = 0;
     for (auto item : world->items.other.IN_PLAY) {
+        if (item->flags.bits.construction ||
```
myk002, line 812: can you reuse BadFlags here?

plugins/lua/logistics.lua
```diff
@@ -6,94 +6,176 @@ local overlay = require('plugins.overlay')
...
-local function make_stat(name, stockpile_number, stats, configs)
-    return {
-        enabled=configs[stockpile_number] and configs[stockpile_number][name] == 'true',
-        designated=stats[name..'_designated'][stockpile_number] or 0,
-        can_designate=stats[name..'_can_designate'][stockpile_number] or 0,
+-------------------------------------
+-- logistics status console output
+--
+
+local function make_stat(type, sp_number, stats, configs)
```
myk002, line 13: `type` is a reserved keyword in Lua. it's not good practice to override it

plugins/lua/logistics.lua
```diff
@@ -6,94 +6,176 @@ local overlay = require('plugins.overlay')
...
+    local stat = {
+        enabled=(configs[sp_number] ~= nil) and (configs[sp_number][type] == 'true'),
```
myk002, line 15: unless you want to specifically include `false` as a valid value, it's better to leave off the `~= nil`

plugins/lua/logistics.lua
```diff
@@ -6,94 +6,176 @@ local overlay = require('plugins.overlay')
...
-            name = ('Stockpile #%d'):format(bld.stockpile_number)
-            sort_key = ('stockpile #%09d'):format(bld.stockpile_number)
+    local data = {sp_stats = {}, sp_totals = {}, sp_name_max_len = 0}
+    data.sp_totals = {
+        melt = {enabled_count = 0, designated = 0, designatable = 0},
+        trade = {enabled_count = 0, designated = 0, designatable = 0},
+        dump = {enabled_count = 0, designated = 0, designatable = 0},
+        train = {enabled_count = 0, designated = 0, designatable = 0},
+        forbid = {enabled_count = 0, designated = 0, designatable = 0},
+        claim = {enabled_count = 0}
```
myk002, line 33: we generally do not add spaces around `=` inside of inline table declarations. that makes them visually distinct from regular assignment statements.

plugins/lua/stockpiles.lua
```diff
@@ -428,10 +428,10 @@ end
 StockpilesOverlay = defclass(StockpilesOverlay, overlay.OverlayWidget)
 StockpilesOverlay.ATTRS{
     desc='Shows a panel when a stockpile is selected for stockpile automation.',
-    default_pos={x=24, y=-6},
+    default_pos={x=5, y=44},
     default_enabled=true,
-    viewscreens='dwarfmode/Stockpile/Some',
-    frame={w=65, h=4},
+    viewscreens='dwarfmode/Stockpile/Some', -- hmm, how do I close it when 'dwarfmode/CustomStockpile' is also up?
```
myk002, line 433: I don't know why we have it this way right now. let me look into it. Maybe we can get rid of `dwarfmode/CustomStockpile` as a focus string and instead add a suffix to `dwarfmode/Stockpile/Some`. E.g. main map stockpile selected would be `dwarfmode/Stockpile/Some/Default` and customization screen would be `dwarfmode/Stockpile/Some/Customize`. Same goes for `dwarfmode/StockpileLink` and `dwarfmode/StockpileTools`.

plugins/lua/stockpiles.lua
```diff
@@ -508,6 +508,16 @@ function StockpilesOverlay:init()
                                 {label='Training', value=false}},
                         initial_option=false,
                         on_change=self:callback('toggleLogisticsFeature', 'train'),
+                    }, widgets.CycleHotkeyLabel{
+                        view_id='forbid',
+                        frame={t=2, l=0, w=16},
```
myk002, line 513: can you put some other toggles down on `t=2` and make the overlay panel a bit shorter?

myk002:
No need to split this into separate PRs. I think it's fine as it is.

plugins/lua/stockpiles.lua
myk002: Fixed in https://github.com/DFHack/dfhack/pull/4612
You can now use `dwarfmode/Stockpile/Some/Default`

plugins/lua/logistics.lua
xBlackTalonX: Yeah, I'm depending on enabled to become true or false here as a result of nil was messing up my logic in lines like this
```
has_claim_enabled = has_claim_enabled or sp.claim.enabled
```

However, I wasn't sure testing for `nil` like am here is common practice in lua or if there was some other shorthand method.

plugins/lua/stockpiles.lua
xBlackTalonX: Yeah, I was thinking that would be possible - I had some crazier ideas too but for now I'll see if I can get it looking good with options in distributed in two rows.

plugins/logistics.cpp
xBlackTalonX: Missed that as well... thanks again!

xBlackTalonX:
I think these are the changes but I actually didn't mean to push them yet (was trying to test, had to fight with git first, when that was fixed I managed to push before I remembered I was trying to build and test). Anyway - personally I'd prefer to small this up into a different PR but I'll comment once it's tested and get a review on this response to the comments.

plugins/lua/logistics.lua
```diff
@@ -168,8 +168,8 @@ local function print_status()
     print((' - Items designated for trading: %5d'):format(global_sp_stats.total_trade))
     print((' - Items designated for dumping: %5d'):format(global_sp_stats.total_dump))
     print((' - Animals designated for train: %5d'):format(global_sp_stats.total_train))  -- TODO: fix to include all animals designated for training
-    print((' - Forbidden items             : %5d'):format(global_sp_stats.total_forbid)) -- TODO: this should exclude unreachable items and buildings
-    print((' - All (unforbidden) items     : %5d'):format(global_sp_stats.total_claim))  -- TODO: same ^^ implement in logistics.cpp or use item script api? (reqscript('item'))
+    print((' - Forbidden items             : %5d'):format(global_sp_stats.total_forbid)) -- TODO: this should exclude unreachable items
```
myk002, line 171: these numbers only involve items in stockpiles. "reachable" isn't really a concern here, is it?

plugins/lua/logistics.lua
```diff
@@ -168,8 +168,8 @@ local function print_status()
     print((' - Items designated for trading: %5d'):format(global_sp_stats.total_trade))
     print((' - Items designated for dumping: %5d'):format(global_sp_stats.total_dump))
     print((' - Animals designated for train: %5d'):format(global_sp_stats.total_train))  -- TODO: fix to include all animals designated for training
```
myk002, line 170: this is only concerning animals in cages in stockpiles. I don't think we need to show all animals designated for training here

myk002 (commented):
code all looks good. only outstanding item is rearranging the overlay and testing

plugins/lua/stockpiles.lua
xBlackTalonX: This seems to work without changing it from `dwarfmode/Stockpile/Some` - which is great! - I'm just not sure I understand what `dwarfmode/Stockpile/Some/Default` represents.

plugins/lua/logistics.lua
xBlackTalonX: Well, as it was previously written, I'm pretty sure it isn't limited to only items in stockpiles. I created the totals for the items in stockpiles and they don't match. Where you pointed out I could be using get_bad_flags for a bitwise compare, that's looping through all items in play to get total counts of melt designations.

I agree, this plugin isn't about summaries for all items... but it was there and makes sense in the context of, "what do I have designated for melt?" Extending it for, "what do I have designated as forbidden," causes the paradigm to break down because we often have so many items forbidden beyond what we care about. And that's where it's a bit odd for animals being trained as well... "wait, I thought I had more animals being trained! ...oh, I do, just not animals in stockpiles."

The more I thought about it the more I liked the summary's existing function to be a total of everything in the fort, but that's when I noted these as needing improvement. And secondarily, I thought it would be cool to show, "stockpiled" / "total" for each line, which would further clarify why this summary is there rather than just using the new totals in footer of the stockpiles table.

But I dunno... that's basically just my thought process. Maybe I'm overthinking it?

xBlackTalonX: image

xBlackTalonX:
Fixed up in that last commit, interested in feedback on the layout. And yeah, ready for testing, not that I haven't been using it while developing it, my only fears are my more recent changes being something stupid.

myk002 (approved):
looks good to me. I'm happy merging this as-is, but feel free to test/tweak further. There's always the next PR.

PR #4577, opened 2024-05-09
Add typing to widgets and gui

vallode:
This PR aims to add type support to `widgets.lua` and `gui.lua` and as a consequence extend the typing of a few other files as well (namely some globals and `class.lua`).

There are some conventions being established here, feel free to critique anything and everything for the sake of establishing a good foundation.

When defining new classes we should follow this rough pattern for a good experience:

```lua
---@class moduleName.ClassNameAttrs: moduleName.ParentClassAttrs
---@field foo integer This is never nil
---@field bar? integer This can be nil

---@class moduleName.ClassNameInitTable: moduleName.ClassNameAttrs
---@field foo integer This needs to be provided during init
-- This also inherits all attributes from ClassNameAttrs but they are optional.

---@class moduleName.ClassName: moduleName.ParentClass, moduleName.ClassNameAttrs
---@field super moduleName.ParentClass
---@field ATTRS moduleName.ClassNameAttrs|fun(attributes: moduleName.ClassNameAttrs)
---@overload fun(init_table: moduleName.ClassNameInitTable): self
ClassName = defclass(ClassName, Widget)
```

There are a few caveats:

- `defclass(ClassName)` plays weird with LuaLS and `self` parameters in the module file will have to be manually cast as the class, i.e:
```lua
---@param self moduleName.ClassName
---@param init_table moduleName.ClassNameInitTable
function ClassName:init(init_table) end```
- `subviews` indexing has no autocomplete or type support so subviews accessed need to be cast manually, i.e
```lua
---@type widgets.List
local list = self.subviews.list
```


vallode:
pre-commit.ci autofix

vallode:
@myk002 `gui.lua` is probably ready for a first pass to start setting in stone some of the conventions we might use in the future. Namespacing is one issue `dfhack.gui.Painter` vs `gui.Painter`. As well as where to keep things like `dfhack.color` and `dfhack.pen`, inline whereever fits or maybe a more generic `types.lua` file also works. I'll let you look over it before I continue further with typing all the methods in gui and widgets.

library/lua/gui/widgets.lua
```diff
@@ -875,6 +984,11 @@ end
...
 SCROLL_DELAY_MS = 20
 
+---@class widgets.ScrollbarAttrs: widgets.WidgetAttrs
+
+---@class widgets.Scrollbar: widgets.Widget, widgets.ScrollbarAttrs
+---@field super widgets.Widget
+---@field ATTRS fun(attributes: widgets.ScrollbarAttrs)
 Scrollbar = defclass(Scrollbar, Widget)
 
 Scrollbar.ATTRS{
```
myk002, line 994: on_scroll needs definition

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
     end
 end
 
+---@class widgets.LabelToken
+---@field text string|fun(): string
+---@field gap? integer
+---@field tile? any pen
+---@field htile? any pen
```
myk002, line 1441: i believe these can be pens or integers

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
...
+---@field width? integer|fun(): integer
+---@field pad_char? string
+---@field key? string
+---@field key_sep? string
+---@field disabled? boolean|fun(): boolean
+---@field enabled? boolean|fun(): boolean
+---@field pen? pen
+---@field dpen? pen
+---@field hpen? pen
```
myk002, line 1450: these can be dfhack.color or dfhack.Pen

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
...
+---@field on_activiate? function
```
myk002, line 1451: specifically, a function that takes no params and returns no value

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
...
+---@field id? string
+---@field line? any
+---@field x1? any
+---@field x2? any
```
myk002, line 1455: these three are internal; do they need to be declared here?

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
...
+
+---@class widgets.LabelAttrs: widgets.WidgetAttrs
+---@field text string|widgets.LabelToken[]
+---@field text_pen? any
+---@field text_dpen? any
+---@field text_hpen? any
```
myk002, line 1461: colors or Pens

library/lua/gui/widgets.lua
```diff
@@ -1320,6 +1434,43 @@ function check_text_keys(self, keys)
...
+---@field disabled? boolean|fun(): boolean
+---@field enabled? boolean|fun(): boolean
+---@field auto_height? boolean
+---@field auto_width? boolean
+---@field on_click? function
+---@field on_rclick? function
+---@field scroll_keys? widgets.Keys
```
myk002, line 1468: scroll_keys is a map from keybinding strings to either an integer or a string

library/lua/gui/widgets.lua
```diff
@@ -1473,6 +1625,11 @@ end
 -- WrappedLabel --
 ------------------
 
+---@class widgets.WrappedLabelAttrs: widgets.LabelAttrs
```
myk002, line 1628: here and down is missing attr field definitions

myk002 (commented):
looks like a great start! class names look good to me

library/lua/dfhack.lua
```diff
@@ -15,6 +15,16 @@ local dfhack = dfhack
 local base_env = dfhack.BASE_G
 local _ENV = base_env
 
+-- Types
+
+---@alias dfhack.truthy
+---| boolean
+---| string Not an empty string ""
+---| integer Not 0
```
myk002, line 23: in lua, 0 is truthy

only `nil` and `false` are falsey

library/lua/gui.lua
```diff
@@ -8,6 +8,20 @@ local utils = require('utils')
 local dscreen = dfhack.screen
 local getval = utils.getval
 
+---@class dfhack.pen
+---@field ch? string
```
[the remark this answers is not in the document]
vallode: I think this is probably a candidate for an alias/class like `dfhack.pen` or `dfhack.truthy`.

library/lua/gui.lua
```diff
@@ -178,8 +208,31 @@ end
...
+---@field width? integer
+---@field height? integer
+
+---@class gui.ViewRect: dfhack.class, gui.ViewRectAttrs
+---@field ATTRS fun(attributes: gui.ViewRectAttrs)
+---@overload fun(attributes: gui.ViewRectAttrs): self
 ViewRect = defclass(ViewRect, nil)
 
+---@param self gui.ViewRect
+---@param args gui.ViewRectAttrs
```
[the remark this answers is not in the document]
vallode: I wrote up a bit response and I think Github lost it. Essentially the issue here is that since lua language server lacks a generic type `Partial<T>` we are stuck with a choice of making the base class entirely optional or having to write both an instance version of the type and an init version of the type, both having mostly the same fields...

vallode:
@myk002 getting closer to completion, feel free to ignore widgets.lua below `TextButton` I ran out of steam a bit. Also feel free to leave this for a day or two before I get it closer to being 100% done. Cheers!

vallode:
pre-commit.ci autofix

library/lua/gui.lua
```diff
@@ -8,6 +8,20 @@ local utils = require('utils')
 local dscreen = dfhack.screen
 local getval = utils.getval
 
+---@class dfhack.pen
+---@field ch? integer|string
+---@field fg? dfhack.color
+---@field bg? dfhack.color
+---@field bold? boolean
+---@field tile? integer
```
myk002, line 16: can also be a function that returns an int

library/lua/gui/widgets.lua
```diff
@@ -10,12 +10,17 @@ local utils = require('utils')
...
 
+---@param view gui.View
+---@param vis boolean
 local function show_view(view,vis)
     if view then
         view.visible = vis
     end
 end
 
+---@param tab table
```
myk002, line 21: `tab? table<T>`

library/lua/gui/widgets.lua
```diff
@@ -10,12 +10,17 @@ local utils = require('utils')
...
+---@param idx integer
+---@return table|integer
```
myk002, line 23: `T|integer`

library/lua/gui/widgets.lua
```diff
@@ -62,7 +99,26 @@ end
 -- Divider --
 -------------
 
+---@class widgets.Divider.attrs: widgets.Widget.attrs
+---@field frame_style gui.Frame|fun(): gui.Frame
+---@field interior boolean
+---@field frame_style_t? false|gui.Frame|fun(): gui.Frame
+---@field frame_style_b? false|fun(): gui.Frame
+---@field frame_style_l? false|fun(): gui.Frame
+---@field frame_style_r? false|fun(): gui.Frame
```
myk002, line 108: these three can also be a gui.Frame, just like frame_style_t

library/lua/gui/widgets.lua
```diff
@@ -1814,13 +2110,48 @@ ToggleHotkeyLabel.ATTRS{
 -- List --
 ----------
 
+---@class widgets.ListChoice
+---@field text string
```
myk002, line 2114: this can also be a list of text tokens, like Label

library/lua/gui/widgets.lua
vallode: As mentioned on Discord, I think this doesn't work. Full example would be something like:

```lua
---@generic T
---@param tab? table<T>
---@param idx integer
---@return T
local function map_opttab(tab,idx)
    if tab then
        return tab[idx]
    else
        return idx
    end
end

---@type string[]
local foo
local bar = map_opttab(foo, 0)```

Where `bar` remains `unknown`

myk002 (approved):

PR #4592, opened 2024-05-15
Make changevein affect connected veins

### LINKED ISSUE #4519: changevein should follow veins across map blocks
instead of stopping at the edge of the current map block

tatoyoda600:
Fixes #4519
changevein will now convert any connected veins in the same tile and neighboring tiles.

### DIFF
--- a/plugins/changevein.cpp
+++ b/plugins/changevein.cpp
@@ -100,9 +100,8 @@ struct VeinEdgeBitmask
 
 static void ChangeSameBlockVeins(df::map_block* block, df::block_square_event_mineralst* event, VeinEdgeBitmask& mask, int32_t matIndex)
 {
-    for (size_t i = 0; i < block->block_events.size(); i++)
+    for (auto evt : block->block_events)
     {
-        df::block_square_event* evt = block->block_events[i];
         if (evt == event)
             continue;
         if (evt->getType() != block_square_event_type::mineral)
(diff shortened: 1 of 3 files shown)


plugins/changevein.cpp
```diff
@@ -24,6 +24,188 @@ DFHACK_PLUGIN("changevein");
...
+            continue;
+        if (evt->getType() != block_square_event_type::mineral)
+            continue;
+
+        df::block_square_event_mineralst* cur = (df::block_square_event_mineralst*)evt;
+        if (cur->inorganic_mat == mask.mat)
+        {
+            for (uint8_t j = 0; j < 15; j++)
+            {
+                uint16_t bitmask = cur->tile_bitmask.bits[j] | cur->tile_bitmask.bits[std::max(j - 1, 0)] | cur->tile_bitmask.bits[std::min(j + 1, 15)];
```
myk002, line 116: isn't j-1 == 255 since it's a uint8_t and has a starting value of 0?

plugins/changevein.cpp
```diff
@@ -24,6 +24,188 @@ DFHACK_PLUGIN("changevein");
...
+
+        df::coord pos = df::coord(block->map_pos.x + offset.x, block->map_pos.y + offset.y, block->map_pos.z);
+        df::map_block* newBlock = Maps::getTileBlock(pos);
+
+        if (!newBlock)
+            continue;
+
+        for (size_t j = 0; j < newBlock->block_events.size(); j++)
+        {
+            df::block_square_event* evt = newBlock->block_events[j];
```
myk002, line 155:
```suggestion
        for (auto evt : newBlock->block_events)
        {
```

myk002: same for other similar for loops in this file. the original code was written before this syntax was available in C++, but there's no reason not to use the modern syntax

myk002 (commented):
nice work!

please add a line to the changelog under "Misc Improvements" for the "Future" version

also, please document the change in behavior in `docs/changevein.rst`

plugins/changevein.cpp
tatoyoda600: Great catch! Got used to using uint as if it was an int and didn't notice.

myk002 (approved):

PR #4628, opened 2024-05-22
Gui.cpp: Fix skipping index 0 for announcement code

Bumber64:
Fixed some announcement code that wasn't reverse-iterating vectors to index 0. Better to use index-based for loops instead of iterators for this (as erasing from vectors using the proper `std::reverse_iterator` is a pain.)


myk002 (commented):
could you add a line in the changelog? Under Fixes, I believe

myk002 (approved):

Issue #4595, opened 2024-05-16
Cuboid struct/class for df::coord bounds


Bumber64:
Testing if a `df::coord` is inside a given box is an issue that comes up when dealing with map positions. There doesn't seem to be anything handy in the DFHack library, so every tool has to implement its own logic.

Not sure where the struct should be defined. `MiscUtils`?

I created an example while working on an improved `plugins/plants.cpp`:
```
struct cuboid
{
    int16_t x_min = -1;
    int16_t x_max = -1;
    int16_t y_min = -1;
    int16_t y_max = -1;
    int16_t z_min = -1;
    int16_t z_max = -1;

    bool isValid() const
    {   // True if all max >= min >= 0
        return (x_min >= 0 && y_min >= 0 && z_min >= 0 &&
            x_max >= x_min && y_max >= y_min && z_max >= z_min);
    }

    bool addPos(int16_t x, int16_t y, int16_t z)
    {   // Expand cuboid to include point. Return true if bounds changed
        if (x < 0 || y < 0 || z < 0 || (isValid() && containsPos(x, y, z)))
            return false;

        x_min = (x_min < 0 || x < x_min) ? x : x_min;
        x_max = (x_max < 0 || x > x_max) ? x : x_max;

        y_min = (y_min < 0 || y < y_min) ? y : y_min;
        y_max = (y_max < 0 || y > y_max) ? y : y_max;

        z_min = (z_min < 0 || z < z_min) ? z : z_min;
        z_max = (z_max < 0 || z > z_max) ? z : z_max;

        return true;
    }
    inline bool addPos(const df::coord &pos) { return addPos(pos.x, pos.y, pos.z); }

    bool containsPos(int16_t x, int16_t y, int16_t z) const
    {   // Return true if point inside cuboid. Make sure cuboid is valid first!
        return x >= x_min && x <= x_max &&
            y >= y_min && y <= y_max &&
            z >= z_min && z <= z_max;
    }
    inline bool containsPos(const df::coord &pos) const { return containsPos(pos.x, pos.y, pos.z); }
};
```
This could probably be refined. Maybe add a constructor, a `clear()` function, and get/set (so we can keep track of isValid state.) I don't know if there's a use-case for supporting negative values (e.g., relative coordinates,) but this implementation doesn't.

[...]

PR #4632, opened 2024-05-24
Improve plants/plant plugin

Bumber64:
Rename the `plants` plugin to `plant` to match the command, making it easier to find docs.

Add command `plant remove` to remove plants. (Doesn't work on mature trees yet. Will need to reverse-engineer proper tile removal at a later date.) Add `plant list` to list shrub and sapling raw IDs, avoiding use of lengthy `devel/query` command and excluding grasses.

Adds a bunch of options to existing `plant create` and `plant grow`. See: `docs/plugins/plant.rst`. Use proper logic for determining plant `watery` flag, adding to wet/dry vector.

Minor improvement to `regrass` to accept numerical raw ID for grass type. `regrass --list` replaces `regrass --plant ""`. Don't remove mud on regrass (since DF doesn't.)

`plants` doc tag now encompasses tools that affect grass, to include `regrass` plugin.

---
- Requires (#4591) to pass vector ref to Lua
- `plant` uses its own `cuboid` struct. This can be replaced by a more global one: #4595


myk002:
don't mind the compilation failures. they'll clear up once https://github.com/DFHack/dfhack/pull/4631 is merged

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
+// Grow and remove shrubs or trees.
+
+#include <iostream>
+#include <iomanip>
+#include <map>
+#include <algorithm>
+#include <vector>
+#include <string>
```
myk002, line 8: since you're organizing, system headers should go last, after `df/` headers. also you can probably get away with not including most of these since they're already included by other headers

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+using std::vector;
+using std::string;
+using namespace DFHack;
+
+DFHACK_PLUGIN("plant");
+REQUIRE_GLOBAL(world);
+
+namespace DFHack
+{
+    DBG_DECLARE(plant, log, DebugCategory::LINFO);
```
myk002, line 38: haha

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+}
+
+struct cuboid
```
myk002, line 41: I'll keep this in mind for refactoring if we want to make it more widely accessible. seems like it should be useful to many map-related tools

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+{
+    int16_t x_min = -1;
+    int16_t x_max = -1;
+    int16_t y_min = -1;
+    int16_t y_max = -1;
+    int16_t z_min = -1;
+    int16_t z_max = -1;
+
+    bool isValid() const
+    {   // False if any bound is < 0
```
myk002, line 51: shouldn't it also return false if any max is less than the relevant min?

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+        y_max = (y_max < 0 || y > y_max) ? y : y_max;
+
+        z_min = (z_min < 0 || z < z_min) ? z : z_min;
+        z_max = (z_max < 0 || z > z_max) ? z : z_max;
+
+        return true;
+    }
+    inline bool addPos(df::coord pos) { return addPos(pos.x, pos.y, pos.z); }
+
+    bool testPos(int16_t x, int16_t y, int16_t z) const
```
myk002, line 75: maybe "contains" is a clearer name?

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+    {   // Return true if point inside cuboid. Make sure cuboid is valid first!
+        return x >= x_min && x <= x_max &&
+            y >= y_min && y <= y_max &&
+            z >= z_min && z <= z_max;
+    }
+    inline bool testPos(df::coord pos) const { return testPos(pos.x, pos.y, pos.z); }
```
myk002, line 81: `const df::coord & pos`?

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+    { struct_field_info::PRIMITIVE, "filter_ex", offsetof(plant_options, filter_ex), &df::identity_traits<bool>::identity, 0, 0 },
+    { struct_field_info::PRIMITIVE, "zlevel",    offsetof(plant_options, zlevel),    &df::identity_traits<bool>::identity, 0, 0 },
+    { struct_field_info::PRIMITIVE, "dry_run",   offsetof(plant_options, dry_run),   &df::identity_traits<bool>::identity, 0, 0 },
+    { struct_field_info::PRIMITIVE, "plant_idx", offsetof(plant_options, plant_idx), &df::identity_traits<int32_t>::identity, 0, 0 },
+    { struct_field_info::PRIMITIVE, "age",       offsetof(plant_options, age),       &df::identity_traits<int32_t>::identity, 0, 0 },
+    { struct_field_info::END }
+};
+struct_identity plant_options::_identity(sizeof(plant_options), &df::allocator_fn<plant_options>, NULL, "plant_options", NULL, plant_options_fields);
+
+const int32_t sapling_to_tree_threshold = 120 * 28 * 12 * 3 - 1; // 3 years minus 1; let the game handle the actual growing-up
```
myk002, line 120: it's silly that we don't have constants for these already. something we should think about when we address https://github.com/DFHack/dfhack/pull/3907

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+command_result df_createplant(color_ostream &out, df::coord pos, const plant_options &options)
+{
+    auto col = Maps::getBlockColumn((pos.x / 48)*3, (pos.y / 48)*3);
+    if (!Maps::getTileBlock(pos) || !col)
```
myk002, line 125:
```suggestion
    if (!Maps::isValidTilePos(pos) || !col)
```

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+        plant->hitpoints = 400000;
+    else
+    {
+        plant->hitpoints = 100000;
+        plant->flags.bits.is_shrub = true;
+    }
+
+    // This is correct except for RICE, DATE_PALM, and underground plants
+    // near pool/river/brook. These have both WET and DRY flags.
+    // Should more properly detect if near surface water feature
```
myk002, line 174: could you add a github issue for this so we don't forget?

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+    if (!p_raw->flags.is_set(plant_raw_flags::DRY))
+        plant->flags.bits.watery = true;
+
+    plant->material = options.plant_idx;
+    plant->pos = pos;
+    plant->grow_counter = options.age < 0 ? 0 : options.age;
+    plant->update_order = rand() % 10;
+
+    world->plants.all.push_back(plant);
+    switch (plant->flags.whole & 3) // watery, is_shrub
```
myk002, line 184: this is a little too "magic" and could break if the relevant structure changes. maybe just plain old if else if else if else if would be better here

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+        return CR_FAILURE;
+    }
+    else // Check tile mat and building occ
+    {
+        auto mat = tileMaterial(*tt);
+        if (mat != tiletype_material::SOIL &&
+            mat != tiletype_material::GRASS_DARK &&
+            mat != tiletype_material::GRASS_LIGHT)
+        {
+            out.printerr("Plants can only be placed on dirt or grass!\n");
```
myk002, line 144: is this true for saplings? i thought they could grow on muddy natural stone floors

plugins/plant.cpp
```diff
@@ -0,0 +1,627 @@
...
+    else
+        out.print("%d saplings%s set to grow.\n", grown, options.dry_run ? " would be" : "");
+
+    return CR_OK;
+}
+
+static bool uncat_plant(df::plant *plant)
+{   // Remove plant from extra vectors
+    vector<df::plant *> *vec = NULL;
+    switch (plant->flags.whole & 3) // watery, is_shrub
```
myk002, line 267: too magic; suggest if else if...

myk002 (commented):
needs a section in Removed.rst so that old links still work

myk002:
ok, now the build errors are real

docs/plugins/plant.rst
```diff
@@ -0,0 +1,154 @@
+plant
+=====
+
+.. dfhack-tool::
+    :summary: Grow and remove shrubs or trees.
+    :tags: adventure fort armok map plants
```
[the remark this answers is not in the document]
Bumber64: Just went through `Tags.rst` and chose all that applied.

docs/plugins/plant.rst
```diff
@@ -0,0 +1,154 @@
...
+::
+
+    plant create <plant_id> [<pos>] [<options>]
+
+Creates a new plant of the specified type at ``pos`` or the cursor position.
+The target tile must be a dirt or grass floor. ``plant_id`` is not
+case-sensitive, but must be enclosed in quotes if spaces exist. (No unmodded
+shrub or sapling IDs have spaces.) A numerical ID can also be used. Providing
+an empty string with "" will print all available IDs and skip plant creation.
+
```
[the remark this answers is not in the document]
Bumber64: IIRC, the old method lists grasses, which aren't valid.

docs/plugins/regrass.rst
```diff
@@ -46,8 +46,9 @@ Options
     specified.
 ``-p <grass_id>``, ``--plant <grass_id>``
     Specify a grass type for the ``--force`` option. ``grass_id`` is not
-    case-sensitive, but must be enclosed in quotes if spaces exist. Providing
-    an empty string with "" will print all available IDs and skip regrass.
+    case-sensitive, but must be enclosed in quotes if spaces exist. A numerical
+    ID can also be used. Providing an empty string with "" will print all
```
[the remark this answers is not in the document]
Bumber64: Is this not a documentation of how to get the ID? `regrass -p ""` is also listed explicitly in the examples.

plugins/plant.cpp
Bumber64: 🪵

plugins/plant.cpp
Bumber64: Probably, since I didn't define any setter functions to prevent that.

plugins/plant.cpp
Bumber64: It's just a limitation from the previous code. I could update the check like `regrass`.

ab9rf:
> ok, now the build errors are real

this PR depends on #4591 for the use of `std::vector<int32_t>` in a lua call from a plugin

docs/plugins/plant.rst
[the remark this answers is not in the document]
Bumber64: Where are the doc standards located? Couldn't find them.

plugins/plant.cpp
Bumber64: That doesn't seem to check if the block exists. `getTileType` does, so I just moved that up.

Bumber64:
Should regrass be given the `plant` tag? The tag definition doesn't specify grass, but a user might expect grass to be included.

docs/plugins/plant.rst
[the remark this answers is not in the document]
Bumber64: Do I use `:no-command:`, or is that handled by the fact that it's `plant create` instead of `plant-create`?

docs/plugins/plant.rst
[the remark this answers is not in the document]
myk002: oh yeah, that's fine to provide a command to list the appropriate identifiers. I'd suggest `plant list` instead of `plant create ""`, though

docs/plugins/plant.rst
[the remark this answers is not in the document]
myk002: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html#documentation-standards

myk002:
> Should `regrass` be given the `plants` tag? The tag definition doesn't specify grass, but a user might expect grass to be included.

yes, that sounds reasonable

docs/plugins/plant.rst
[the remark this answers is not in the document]
myk002: > Do I use `:no-command:`, or is that handled by the fact that it's `plant create` rather than `plant-create`?

`:no-command:` is for when the command that you run is different from the name of the plugin. You've solved that discrepancy, so you don't need that tag.

docs/Tags.rst
```diff
@@ -42,7 +42,7 @@ for the tag assignment spreadsheet.
 - `labors <labors-tag-index>`: Tools that deal with labor assignment.
 - `map <map-tag-index>`: Tools that interact with the game map.
 - `military <military-tag-index>`: Tools that interact with the military.
-- `plants <plants-tag-index>`: Tools that interact with trees, shrubs, and crops.
+- `plants <plants-tag-index>`: Tools that interact with grass, trees, shrubs, and crops.
```
myk002, line 45: I updated https://docs.google.com/spreadsheets/d/1hiDlo8M_bB_1jE-5HRs2RrrA_VZ4cRu9VXaTctX_nwk/edit#gid=170388995 to match this

docs/changelog.txt
```diff
@@ -52,18 +52,22 @@ Template for new versions:
 # Future
 
 ## New Tools
+- `plant` (formerly `plants`): (reinstated) tool for creating/growing/removing plants
```
myk002, line 55:
```suggestion
- `plant`: (reinstated) tool for creating/growing/removing plants
```

docs/changelog.txt
```diff
@@ -52,18 +52,22 @@ Template for new versions:
...
 
 ## New Features
 - `tweak`: ``named-codices``: display book titles instead of a material description in the stocks/trade screens
+- `plant` (formerly `plants`): can now ``remove`` shrubs and saplings; ``list`` all valid shrub/sapling raw IDs; ``grow`` can make mature trees older; many new command options
```
myk002, line 59:
```suggestion
- `plant`: can now ``remove`` shrubs and saplings; ``list`` all valid shrub/sapling raw IDs; ``grow`` can make mature trees older; many new command options
```

docs/plugins/regrass.rst
```diff
@@ -24,9 +24,45 @@ can be used in place of numeric coordinates to use the position of the keyboard
...
+    ``dog's tooth grass`` if no compatible types exist. Refill existing grass
+    on each tile, else select one of the block's types if depleted or
+    previously non-existent.
+
 Options
 -------
 
+``-l``, ``--list``
+    Print all available grass raw IDs and skip regrass. For use with
+    ``--plant``.
```
myk002, line 65:
```suggestion
    List all available grass raw IDs that you can later pass to the
    ``--plant`` option. The map will not be affected when running
    with this option.
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+::
+
+    plant remove [<pos> [<pos>]] [<options>]
+
+Remove plants from the map (or area defined by ``pos`` arguments.) By default,
+only removes invalid plants that exist on non-plant tiles (due to `Bug 12868
+<https://dwarffortressbugtracker.com/view.php?id=12868>`_.) The ``--shrubs``
+and ``--saplings`` options allow normal plants to be targeted instead. Removal
+of fully-grown trees isn't currently supported.
```
myk002, line 57:
```suggestion
it only removes invalid plants that exist on non-plant tiles (due to
:bug:`12868`). The ``--shrubs`` and ``--saplings`` options allow normal plants
to be targeted instead. Removal of fully-grown trees isn't currently supported.
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+
+``plant list``
+    List all valid shrub and sapling raw IDs.
+``plant create tower_cap``
+    Create a Tower Cap sapling at the cursor.
+``plant create 203 -c -a tree``
+    Create a Willow sapling at the cursor, even away from water features,
+    ready to mature into a tree.
+``plant create single-grain_wheat 70,70,140``
+    Create a Single-grain Wheat shrub at (70, 70, 140.)
```
myk002, line 70:
```suggestion
    Create a Single-grain Wheat shrub at (70, 70, 140).
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+
+``-c``, ``--force``
+    Create plant even on tiles flagged ``no_grow`` and unset the flag. This
+    flag is set on tiles that were originally boulders or pebbles, as well
+    as on some tiles found in deserts, etc. Also allow non-``[DRY]`` plants
+    (e.g., willow) to grow away (3+ tiles) from water features (i.e., pools,
+    brooks, and rivers,) and non-``[WET]`` plants (e.g., prickle berry) to
+    grow near them.
+``-a``, ``--age <value>``
+    Set the created plant to a specific age (in years.) ``value`` can be a
```
myk002, line 99:
```suggestion
    Set the created plant to a specific age (in years). ``value`` can be a
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    non-negative integer, or one of the strings ``tree``/``1x1`` (3 years,)
```
myk002, line 100:
```suggestion
    non-negative integer, or one of the strings ``tree``/``1x1`` (3 years),
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    ``2x2`` (201 years,) or ``3x3`` (401 years.) ``value`` will be capped at
```
myk002, line 101:
```suggestion
    ``2x2`` (201 years), or ``3x3`` (401 years). ``value`` will be capped at
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    1250. Defaults to 0 if option is unused. Only a few tree types grow wider
+    than 1x1, but many may grow taller. (Going directly to higher years will
+    stunt height. It may be more desirable to instead use ``plant grow`` in
+    stages, or just spawn full trees using `gui/sandbox`.)
```
myk002, line 105:
```suggestion
    stages, or just spawn full trees using `gui/sandbox`).
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Grow Options
+------------
+
+``-a``, ``--age <value>``
+    Define the age (in years) to set saplings to. ``value`` can be a
+    non-negative integer, or one of the strings ``tree``/``1x1`` (3 years,)
```
myk002, line 112:
```suggestion
    non-negative integer, or one of the strings ``tree``/``1x1`` (3 years),
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    ``2x2`` (201 years,) or ``3x3`` (401 years.) ``value`` will be capped at
```
myk002, line 113:
```suggestion
    ``2x2`` (201 years), or ``3x3`` (401 years). ``value`` will be capped at
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    types. ``plant_id`` is not case-sensitive, but must be enclosed in quotes
+    if spaces exist. (No unmodded tree raw IDs have spaces.) A numerical ID
+    can be used in place of a string. Use ``plant list`` and check under
+    ``Saplings`` for a list of valid IDs.
+``-e``, ``--exclude <plant_id>[,<plant_id>...]``
+    Same as ``--filter``, but target everything except these. Cannot be used
+    with ``--filter``.
+``-z``, ``--zlevel``
+    Operate on a range of z-levels instead of default targeting. Will do all
+    z-levels between ``pos`` arguments if both are given (instead of cuboid,)
```
myk002, line 130:
```suggestion
    z-levels between ``pos`` arguments if both are given (instead of cuboid),
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    z-level of first ``pos`` if one is given (instead of single tile,) else
```
myk002, line 131:
```suggestion
    z-level of first ``pos`` if one is given (instead of single tile), else
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    z-level of current view if no ``pos`` is given (instead of entire map.)
```
myk002, line 132:
```suggestion
    z-level of current view if no ``pos`` is given (instead of entire map).
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    and can't be used without one of those options. ``plant_id`` is not
+    case-sensitive, but must be enclosed in quotes if spaces exist. (No
+    unmodded shrub or sapling raw IDs have spaces.) A numerical ID can be
+    used in place of a string. Use ``plant list`` for a list of valid IDs.
+``-e``, ``--exclude <plant_id>[,<plant_id>...]``
+    Same as ``--filter``, but target everything except these. Cannot be used
+    with ``--filter``.
+``-z``, ``--zlevel``
+    Operate on a range of z-levels instead of default targeting. Will do all
+    z-levels between ``pos`` arguments if both are given (instead of cuboid,)
```
myk002, line 159:
```suggestion
    z-levels between ``pos`` arguments if both are given (instead of cuboid),
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    z-level of first ``pos`` if one is given (instead of single tile,) else
```
myk002, line 160:
```suggestion
    z-level of first ``pos`` if one is given (instead of single tile), else
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+    z-level of current view if no ``pos`` is given (instead of entire map.)
```
myk002, line 161:
```suggestion
    z-level of current view if no ``pos`` is given (instead of entire map).
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Create Options
+--------------
+
+``-c``, ``--force``
+    Create plant even on tiles flagged ``no_grow`` and unset the flag. This
+    flag is set on tiles that were originally boulders or pebbles, as well
+    as on some tiles found in deserts, etc. Also allow non-``[DRY]`` plants
+    (e.g., willow) to grow away (3+ tiles) from water features (i.e., pools,
+    brooks, and rivers,) and non-``[WET]`` plants (e.g., prickle berry) to
```
myk002, line 96:
```suggestion
    brooks, and rivers), and non-``[WET]`` plants (e.g., prickle berry) to
```

docs/plugins/plant.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Grows saplings (including dead ones) into trees. Will default to all saplings
+on the map if no ``pos`` arguments are used. Saplings will die and fail to grow
+if they are blocked by another tree.
+
+::
+
+    plant remove [<pos> [<pos>]] [<options>]
+
+Remove plants from the map (or area defined by ``pos`` arguments.) By default,
```
myk002, line 53:
```suggestion
Remove plants from the map (or area defined by ``pos`` arguments). By default,
```

myk002 (commented):
waiting on https://github.com/DFHack/dfhack/pull/4591 but otherwise approved

editor's note: final punctuation goes outside of enclosing parentheses unless the entire sentence is inside the parentheses.

myk002 (commented):
comparing with other tools, the parameter is usually named `dry-run` instead of `dryrun`. Could you add a dash?

Bumber64:
> editor's note: final punctuation goes outside of enclosing parentheses unless the entire sentence is inside the parentheses.

U.S. English built different.

myk002 (approved):

myk002:
> > editor's note: final punctuation goes outside of enclosing parentheses unless the entire sentence is inside the parentheses.
> 
> U.S. English built different.

I'm going by https://www.thepunctuationguide.com/parentheses.html (but of course only because that's the way I already write..)

ab9rf:
> > editor's note: final punctuation goes outside of enclosing parentheses unless the entire sentence is inside the parentheses.
> 
> U.S. English built different.

Not according to my legal writing instructor (at a US-based law school).

ab9rf:
i am working on an PR that adds identity support for `std::vector<int32_t>` and `std::vector<int16_t>` in order to make this mergeable

PR #4638, opened 2024-05-26
Add aquifer functions to the Maps module

tatoyoda600:
Adds the following functions:
- `Maps::isTileAquifer`
- `Maps::isTileHeavyAquifer`
- `Maps::setTileAquifer`
- `Maps::removeTileAquifer`
- `dfhack.maps.removeTileAquifer`
- `dfhack.maps.removeTileAquifer`
- `dfhack.maps.removeTileAquifer`
- `dfhack.maps.removeTileAquifer`

Changes `aquifer.cpp` to use these new functions.

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -2513,20 +2513,20 @@ static int maps_setTileAquifer(lua_State* L)
 
     switch (lua_gettop(L))
     {
-    default:
-    case 4:
-        rv = Maps::setTileAquifer(CheckCoordXYZ(L, 1, false), lua_toboolean(L, 4));
-        break;
-    case 3:
-        rv = Maps::setTileAquifer(CheckCoordXYZ(L, 1, false));
-        break;
+    case 1:
+        Lua::CheckDFAssign(L, &p, 1);
+        rv = Maps::setTileAquifer(p);
     case 2:
         Lua::CheckDFAssign(L, &p, 1);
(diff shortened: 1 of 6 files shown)


library/LuaApi.cpp
```diff
@@ -2492,6 +2492,54 @@ static int maps_getBiomeType(lua_State *L)
...
+        rv = Maps::setTileAquifer(CheckCoordXYZ(L, 1, false));
+        break;
+    case 2:
+        Lua::CheckDFAssign(L, &p, 1);
+        rv = Maps::setTileAquifer(p, lua_toboolean(L, 2));
+        break;
+    case 1:
+        Lua::CheckDFAssign(L, &p, 1);
+        rv = Maps::setTileAquifer(p);
+    }
```
ab9rf, line 2530: not fond of the use of "magic numbers" here

myk002 (commented):
The algorithmic changes increase the time it takes to add aquifers by 25% and it increases the time it takes to remove aquifers by a whopping 338%. The real-time impact is on the order of seconds for a large (10x10) map, which isn't completely terrible, but it definitely gives me pause.

Could you possibly add some optional parameters to the C++ API that would allow the algorithm to stay closer to the original? the extra flexibility would not need to be exposed to Lua. Performance isn't as much of an issue when calling from Lua.

myk002 (approved):
performance is comparable to the original code. great work!

PR #4645, opened 2024-05-27
Tiletypes upgrades

tatoyoda600:
Lua interface supports passing in a table with any of the TileTypes class' fields.
Cuboid selections will be processed in batch, instead of each tile individually.
Aquifer option now supports heavy aquifers and uses the new Maps module functions.


docs/dev/Lua API.rst
```diff
@@ -6622,6 +6622,14 @@ tiletypes
   the parameters are enum values from ``df.tiletype_shape``,
   ``df.tiletype_material``, etc. Returns whether the conversion succeeded.
 
+* ``bool tiletypes_setTile(pos, tiletype_options)`` where
+  the ``tiletype_options`` parameter takes in a table, with any fields
+  matching the available tiletypes options:
+  ``shape``, ``material``, ``special``, ``variant``, ``dig``,
```
myk002, line 6628: what does `dig` mean? I don't see it actually being used for anything. If it's intended to set a dig designation, I'd advocate against adding that feature to `tiletypes`. I think it's out of scope. `tiletypes` *does* have a *filter* for dig designated tiles, but that doesn't seem to be relevant in a function called `setTile` (unless it supports filters)

plugins/tiletypes.cpp
```diff
@@ -1097,54 +1222,119 @@ command_result df_tiletypes_here_point (color_ostream &out, vector <string> & pa
...
+            lua_pop(L, 1);
+            return output;
+        };
+
+        target.shape     = (df::tiletype_shape)lua_getintfield("shape", target.shape);
+        target.material  = (df::tiletype_material)lua_getintfield("material", target.material);
+        target.special   = (df::tiletype_special)lua_getintfield("special", target.special);
+        target.variant   = (df::tiletype_variant)lua_getintfield("variant", target.variant);
+        target.dig            = lua_getintfield("dig", target.dig);
+        target.hidden         = lua_getintfield("hidden", target.hidden);
```
myk002, line 1320: these would normally be specified as booleans. does that work correctly with this logic?

docs/dev/Lua API.rst
tatoyoda600: I'd just included it since I saw it in the TileTypes struct definition, probably should've thought it through a bit more though now that you mention it, cause yeah, it doesn't make any actual sense to include it.

plugins/tiletypes.cpp
tatoyoda600: Those were kept as integers to match up with how tiletypes works from the command line, where you specify -1, 0, or 1 for those options. It could be changed to nil, false, or true for Lua, but it didn't seem worth it to introduce such an inconsistency, especially since aquifer has to be an integer now that it accepts -1, 0, 1, or 2.

docs/dev/Lua API.rst
tatoyoda600: I just checked again to make sure, but the `dig` option is available from the command line `tiletypes` function in its current state, should it be removed from there as well?
image

plugins/tiletypes.cpp
myk002: that's fine. I'd document that, though.

docs/dev/Lua API.rst
myk002: Yeah, the code doesn't actually do anything if you paint `designated 1`. I'd remove it.

docs/dev/Lua API.rst
```diff
@@ -6623,12 +6623,20 @@ tiletypes
...
+    - ``shape``: ``df.tiletype_shape``
+    - ``material``: ``df.tiletype_material``
+    - ``special``: ``df.tiletype_special``
+    - ``variant``: ``df.tiletype_variant``
+    - ``hidden``: -1, 0, or 1
+    - ``light``: -1, 0, or 1
+    - ``subterranean``: -1, 0, or 1
+    - ``skyview``: -1, 0, or 1
+    - ``aquifer``: -1, 0, 1, or 2
+    - ``stone_material``: material id
```
myk002, line 6637: integer material id, right, not the token?

docs/dev/Lua API.rst
```diff
@@ -6623,12 +6623,20 @@ tiletypes
...
 
 * ``bool tiletypes_setTile(pos, tiletype_options)`` where
-  the ``tiletype_options`` parameter takes in a table, with any fields
-  matching the available tiletypes options:
-  ``shape``, ``material``, ``special``, ``variant``, ``dig``,
-  ``hidden``, ``light``, ``subterranean``, ``skyview``, ``aquifer``,
-  ``stone_material``, ``vein_type``.
-  Returns whether the conversion succeeded.
+  the ``tiletype_options`` parameter takes in a table, with any fields matching
+  the available tiletypes options. Returns whether the conversion succeeded.
```
myk002, line 6627: I'd also mention that an unset field will result in the property not being changed from the existing tile.

docs/dev/Lua API.rst
tatoyoda600: Yup, it's basically the integer you'd use for in `df::inorganic_raw::find(stone_material)`. I'll add the clarification that it's an integer id in the docs.

myk002 (approved):

PR #4646, opened 2024-05-27
[buildingplan] add option to ignore items from a specified burrow

chdoc:


plugins/buildingplan/buildingplan.cpp
```diff
@@ -44,6 +46,7 @@ const string FILTER_CONFIG_KEY = string(plugin_name) + "/filter";
 const string BLD_CONFIG_KEY = string(plugin_name) + "/building";
 
 static PersistentDataItem config;
+static df::burrow *ignore_burrow = nullptr;
```
myk002, line 49: this seems very dangerous -- what if the burrow is deleted? I think it's better to just keep the id in the config and look it up as necessary -- it could become invalid at any time.

plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -78,13 +80,21 @@ static bool isInWheelbarrow(color_ostream& out, df::item* item) {
     return container->hasToolUse(df::tool_uses::HEAVY_OBJECT_HAULING);
 }
 
+df::burrow* getIgnoreBurrow();
```
myk002, line 83: you can add this to `buildingplan.h`

plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -78,13 +80,21 @@ static bool isInWheelbarrow(color_ostream& out, df::item* item) {
...
+bool isInIgnoreBurrow(df::item *item){
+    CHECK_NULL_POINTER(item)
+    auto ignore_burrow = getIgnoreBurrow();
+    return ignore_burrow && Burrows::isAssignedTile(ignore_burrow, item->pos);
```
myk002, line 87: `item->pos` is unreliable. use `Items::getPosition(item)` instead

myk002 (approved):
added issue for UI implementation: https://github.com/DFHack/dfhack/issues/4648

PR #4656, opened 2024-05-31
export `makeown` to Lua

ab9rf:
oops, that was kinda the whole point now wasn't it


myk002 (commented):
needs docs in Lua API.rst

PR #4657, opened 2024-05-31
add `Buildings::getName`

ab9rf:
convenience method mainly intended for use from Lua, since calling the native DF vmethod from Lua is difficult and nonobvious


myk002 (commented):
needs docs in Lua API.rst

myk002 (approved):

PR #4658, opened 2024-05-31
`autobutcher`: do not butcher pregnant animals

### LINKED ISSUE #1092: autobutcher: ignore units with fertile eggs
Discussed [here](http://www.bay12forums.com/smf/index.php?topic=139553.msg7453545#msg7453545) on the forums.
I wasn't able to find anything in zone.cpp that looked like it dealt with pregnancy, but checking on pregnancy information may be faster than checking all buildings/eggs on the map, and it could apply to other pregnant units as well.

ab9rf:
closes #1092

### DIFF
--- a/plugins/autobutcher.cpp
+++ b/plugins/autobutcher.cpp
@@ -667,6 +667,7 @@ static bool isProtectedUnit(df::unit *unit) {
         // ignore creatures in built cages which are defined as rooms to leave zoos alone
         // (TODO: better solution would be to allow some kind of slaughter cages which you can place near the butcher)
         || (isContainedInItem(unit) && isInBuiltCageRoom(unit))  // !!! see comments in isBuiltCageRoom()
+        || (unit->pregnancy_timer != 0) // do not butcher pregnant animals (which includes brooding female egglayers)
         || Units::isAvailableForAdoption(unit)
         || unit->name.has_name
         || !unit->name.nickname.empty();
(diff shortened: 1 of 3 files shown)


plugins/autobutcher.cpp
```diff
@@ -667,6 +667,7 @@ static bool isProtectedUnit(df::unit *unit) {
         // ignore creatures in built cages which are defined as rooms to leave zoos alone
         // (TODO: better solution would be to allow some kind of slaughter cages which you can place near the butcher)
         || (isContainedInItem(unit) && isInBuiltCageRoom(unit))  // !!! see comments in isBuiltCageRoom()
+        || (unit->pregnancy_timer != 0) // do not butcher pregnant animals (which includes brooding female egglayers)
```
myk002, line 670: do we want this as a hard filter, or just a soft "preference" filter? It could be made soft by balancing pregnancy status against age in the `compareAdults` function, which determines the order in the priority queue

ab9rf: I have no idea. This is a seven year old issue and even after rereading the forum thread referenced in the issue I don't remember what motivated this specifically.

myk002: I presume it was the desire to not miss out on additional butcherable livestock. The question is, is the need for more butcherable livestock more important than keeping the average age low for the unbutchered adults?

I'd say probably yes this is the right move. Could you update the `autobutcher.rst` docs to indicate that pregnant females are exempt from butchering while they are pregnant/brooding?

myk002 (approved):

PR #4659, opened 2024-05-31
remove empty `zone` plugin

ab9rf:
has been code-free since 2022


myk002 (commented):
https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua is the home of several popular and important overlay widgets:
```lua
OVERLAY_WIDGETS = {
    pasturepond=PasturePondOverlay,
    cagechain=CageChainOverlay,
    retirelocation=RetireLocationOverlay,
    animal_actions=AnimalActionsWidget,
}
```
which is why the plugin is currently built and installed. The functionality inside the plugin is also still relevant, but needs to be updated for v50+. I'm also *really* not happy with the API and I think it needs a redesign.

but it provides CLI access to much of the functionality of the pasture/cage/pit/chain assignment dialogs, which is important for scripting, so I don't want to summarily delete it.

quietust:
This is about the C++ Zone plugin, whose entire contents were completely commented out in this commit:
```
commit 2fa4efbb8afc42a948e0535182e42500091b160f
Author: Myk Taylor <myk002@yahoo.com>
Date:   Sat Jul 15 14:31:43 2023 -0700

    initial implementation of pasture assignment screen
```

myk002:
yes, but the Lua module will not be available without a stub plugin to load it. We *could* move the logic to a script, which would reset any custom overlay settings people have set for these overlays, and that's not terrible, but I also don't feel too much pressure to remove `zone` from the build since we're planning on re-populating its cpp logic anyway.

PR #4663, opened 2024-06-02
update to canonicalized field names

myk002:
ref: https://github.com/DFHack/df-structures/pull/772

### DIFF
--- a/plugins/plants.cpp
+++ b/plugins/plants.cpp
@@ -69,7 +69,8 @@ command_result df_grow (color_ostream &out, vector <string> & parameters)
         {
             df::plant *p = world->plants.all[i];
             df::tiletype ttype = map.tiletypeAt(df::coord(p->pos.x,p->pos.y,p->pos.z));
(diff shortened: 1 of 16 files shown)


plugins/plants.cpp
```diff
@@ -148,12 +149,12 @@ command_result df_createplant (color_ostream &out, vector <string> & parameters)
...
         plant->hitpoints = 100000;
-        plant->flags.bits.is_shrub = 1;
+        plant->type = df::plant_type::DRY_PLANT;
     }
-    // for now, always set "watery" for WET-permitted plants, even if they're spawned away from water
+    // for now, always set  to WET_TREE for WET-permitted plants, even if they're spawned away from water
     // the proper method would be to actually look for nearby water features, but it's not clear exactly how that works
     if (plant_raw->flags.is_set(plant_raw_flags::WET))
-        plant->flags.bits.watery = 1;
+        plant->type = df::plant_type::WET_TREE;
```
quietust, line 157: This needs to be either `WET_TREE` or `WET_PLANT`

myk002: thanks for catching that! Fixed in 97153cf09

PR #4667, opened 2024-06-03
use canonicalized structures and field names

myk002:
ref: https://github.com/DFHack/df-structures/pull/774


plugins/manipulator.cpp
```diff
@@ -198,7 +198,7 @@ const SkillColumn columns[] = {
     {12, 4, profession::NONE, unit_labor::PULL_LEVER, job_skill::NONE, "Lv"},
     {12, 4, profession::NONE, unit_labor::BUILD_ROAD, job_skill::NONE, "Ro"},
     {12, 4, profession::NONE, unit_labor::BUILD_CONSTRUCTION, job_skill::NONE, "Co"},
-    {12, 4, profession::NONE, unit_labor::REMOVE_CONSTRUCTION, job_skill::NONE, "CR"},
+    {12, 4, profession::NONE, unit_labor::UNUSED_13, job_skill::NONE, "CR"},
```
quietust, line 201: This line should just go away entirely

myk002: I was worried that it would misalign something else, so I left it in (for now, at least)

PR #4670, opened 2024-06-04
New surroundings option for tiletypes

tatoyoda600:
Adds a new `surroundings` option to `tiletypes`, which attempts to maintain valid tile surroundings, adding floors on top of walls, ramp tops on top of ramps, and telling liquids to flow.
By default it's set to `0` (Off) and will only do anything if set to `1`.


plugins/tiletypes.cpp
```diff
@@ -908,6 +1055,9 @@ static PaintResult paintArea(MapExtras::MapCache& map, const df::coord& pos1, co
                     Maps::removeAreaAquifer(pos1, pos2, filter);
                 else if (target.aquifer > 0)
                     Maps::setAreaAquifer(pos1, pos2, target.aquifer == 2, filter);
+
+                if (target.surroundings > 0 && updateAreaSurroundings(map, pos1, pos2, target))
+                    map.WriteAll();
```
myk002, line 1060: is it safe to do this after modifying the map directly with the `Maps` module calls above? I'd be worried about the `MapCache.WriteAll` sync overwriting changes. It would be safer to do this *before* the calls to the Maps module.

Actually, is this callback happening for each tile? If so, the calls to the Maps module should maybe get buffered until the end and take effect after the last call to MapCache.WriteAll.

(a little history: MapCache exists because DFHack used to be out-of-process and map operations were too slow over IPC. map state was copied into the MapCache, changes happened in the external DFHack process, and then they'd be synced back. Nowadays, MapCache should really be merged into the Maps module, but it has a lot of complex functionality that is difficult to port, such as stone type lookups. Therefore, it still exists and we need to be careful about mixing MapCache calls and Maps module calls)

tatoyoda600: The `.postWrite` callback only runs once, not for each tile, with the batching and all that happening inside the called functions (`Maps::removeAreaAquifer`, `Maps::setAreaAquifer`, `updateAreaSurroundings`).
I haven't run into issues with it overwriting data, but it is a fair point, it should be running before the aquifer stuff just in case.

myk002 (approved):

PR #4675, opened 2024-06-05
Implement df::coord iterator and cuboid class 

### LINKED ISSUE #4595: Cuboid struct/class for df::coord bounds

### LINKED ISSUE #4596: coordRange df::coord iterator for Maps module
Define a `df::coord` iterator `coordRange` for the `Maps` module, similar to `Units` module's `citizensRange`.

This should iterate all `df::coord`s contained in a cuboid defined by two given `df::coord`s. We might decide to not check if the input coords lie in the current loaded map bounds, because the function could be used to iterate block, region, etc., coords. (We can have a version that accepts a `df::coord2d` by setting `z` to 0 and returning the appropriate type.)

We can also accept a cuboid struct (#4595) and make use of it to simplify the iterator (rather than figure out which direction we need to iterate x/y/z.)

Bumber64:
Implement an iterator that operates on all `df::coord`s within a cuboid box. Iterates in the direction of x/y/z provided (rather than always min to max.) By default iterates per z-level, but has an option to iterate in the z direction first.
Closes #4596

Cuboid class for defining a cuboid of tiles. Can test if a coord is inside itself. Can be expanded to contain a given coord. Has a method to call the above iterator using the defined cuboid.
Closes #4595

Also update plugins to use `plant_type` enum attrs, since we're removing the `cuboid` struct from `plugins/plant.cpp` in this PR anyway.

### DIFF
--- a/library/include/modules/Maps.h
+++ b/library/include/modules/Maps.h
@@ -180,6 +180,55 @@ namespace DFHack
         return (p.x & ~15) == 0 && (p.y & ~15) == 0;
     }
 
+    /**
+     * Utility class representing a cuboid of df::coord
+     * \ingroup grp_maps
+     */
+    class cuboid
+    {
+        public:
+        // Bounds
+        int16_t x_min = -1;
+        int16_t x_max = -1;
+        int16_t y_min = -1;
+        int16_t y_max = -1;
+        int16_t z_min = -1;
+        int16_t z_max = -1;
+
+        // Default constructor
+        DFHACK_EXPORT cuboid() {}
+
+        // Construct from two corners
+        DFHACK_EXPORT cuboid(int16_t x1, int16_t y1, int16_t z1, int16_t x2, int16_t y2, int16_t z2);
+        DFHACK_EXPORT cuboid(const df::coord &p1, const df::coord &p2) { cuboid(p1.x, p1.y, p1.z, p2.x, p2.y, p2.z); }
+
+        // Construct as single tile
+        DFHACK_EXPORT cuboid(int16_t x, int16_t y, int16_t z);
+        DFHACK_EXPORT cuboid(const df::coord &p) { cuboid(p.x, p.y, p.z); }
+
+        // Valid cuboid? True if all max >= min >= 0
+        DFHACK_EXPORT bool isValid() const;
+
+        // Clear cuboid dimensions, making it invalid
+        DFHACK_EXPORT void clear() { x_min = x_max = y_min = y_max = z_min = z_max = -1; }
+
+        /// Clamp cuboid within map area and ensure max >= min. Fails if map not loaded or any bound < 0.
+        /// Can optionally treat cuboid as map blocks instead of tiles.
+        /// Note: A point being in map area isn't sufficient to know that a tile block is allocated there!
+        DFHACK_EXPORT bool clamp(bool block = false);
+
+        // Expand cuboid to include point. Returns true if bounds changed.
+        DFHACK_EXPORT bool addPos(int16_t x, int16_t y, int16_t z);
+        DFHACK_EXPORT bool addPos(const df::coord &pos) { return addPos(pos.x, pos.y, pos.z); }
+
+        // Return true if point inside cuboid. Make sure cuboid is valid first!
+        DFHACK_EXPORT bool containsPos(int16_t x, int16_t y, int16_t z) const;
+        DFHACK_EXPORT bool containsPos(const df::coord &pos) const { return containsPos(pos.x, pos.y, pos.z); }
+
+        /// Iterate over every point in the cuboid from top-down, N-S, then W-E. Doesn't guarantee valid map tile!
+        /// "fn" should return true to keep iterating.
+        DFHACK_EXPORT void forCoord(std::function<bool(df::coord)> fn);
+    };
 
     /**
      * The Maps module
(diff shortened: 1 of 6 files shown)


Bumber64:
Need to apply this to existing functions (e.g., `Units::isUnitInBox` and `Units::getUnitsInBox`).

library/include/modules/Maps.h
```diff
@@ -180,6 +180,49 @@ namespace DFHack
...
+
+        // Expand cuboid to include point. Return true if bounds changed
+        DFHACK_EXPORT bool addPos(int16_t x, int16_t y, int16_t z);
+        DFHACK_EXPORT bool addPos(const df::coord &pos) { return addPos(pos.x, pos.y, pos.z); }
+
+        // Return true if point inside cuboid. Make sure cuboid is valid first!
+        DFHACK_EXPORT bool containsPos(int16_t x, int16_t y, int16_t z) const;
+        DFHACK_EXPORT bool containsPos(const df::coord &pos) const { return containsPos(pos.x, pos.y, pos.z); }
+
+        /// Iterate over every point in the cuboid from N-S, W-E, then top-down. Doesn't guarantee valid map tile!
```
myk002, line 222: a `clamp()` function would be useful here to ensure that the cuboid is completely within valid map range.

library/include/modules/Maps.h
```diff
@@ -180,6 +180,49 @@ namespace DFHack
...
+        /// If z_first iterate top-down, N-S, then W-E.
```
myk002, line 223: so far, I've seen code that:
- doesn't care about whether z goes up or down
- requires z to go down (highest z-level to lowest)
so the options you have are likely ok. I don't know if there are any use cases for requiring that z goes from lowest to highest.

library/include/modules/Maps.h
```diff
@@ -180,6 +180,49 @@ namespace DFHack
...
+        DFHACK_EXPORT void forCoord(std::function<void(df::coord)> fn, bool z_first = false);
```
myk002, line 224: how hard would it be to provide a range interface? I added one for `Units::citizensRange`, and I like how it allows you to break out of the for loop early if you need to. `forCoord` would require that you iterate through all points (or you could change the signature of the callback to return a boolean that would indicate whether to exit early). I guess either option would be fine.

library/modules/Maps.cpp
```diff
@@ -105,11 +105,92 @@ const char * DFHack::sa_feature(df::feature_type index)
...
+    if (z_first)
+    {   // Process z-columns first
+        for (int16_t x = x1; x != x2 + dx; x += dx)
+            for (int16_t y = y1; y != y2 + dy; y += dy)
+                for (int16_t z = z1; z != z2 + dz; z += dz)
+                    fn(df::coord(x, y, z));
+    }
+    else
+    {   // One z-level at a time
+        for (int16_t z = z1; z != z2 + dz; z += dz)
```
myk002, line 187: most (all?) existing code does z first. does it make any sort of performance difference? is there any measurable advantage to be acting on the same map block in batches?

library/include/modules/Maps.h
Bumber64: Not quite sure how that works. `Units::citizensRange` filters a vector, but we're generating values from nowhere so I need some kind of example.

Bumber64:
Should a class with only public members more properly be a struct, or is it weirder to have a struct with member functions?

library/include/modules/Maps.h
myk002: I couldn't find a good example, and buffering the whole sequence in a vector just so you can iterate through them seems like a waste. The second option might be better -- return a boolean from the callback to determine whether you should break out of the for loop early.

myk002:
> Should a class with only public members more properly be a struct, or is it weirder to have a struct with member functions?

I don't have an opinion one way or the other here. I'd say do whatever feels natural.

library/include/modules/Maps.h
Bumber64: Tried `break`ing out of the nested for loop. Took me a while to realize the problem with that.

myk002 (approved):

PR #4678, opened 2024-06-06
Fix thread-safety issues around armok mode logic

lethosor:
Followup to feedback in #4674.

I verified that `dfhack.internal.setArmokTools()` and `dfhack.setMortalMode()` can still be called from the `lua` interpreter without deadlocking. I haven't seen meaningful changes occur from calling them, though, so my testing may not be exhaustive.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -2382,10 +2382,12 @@ void Core::setSuppressDuplicateKeyboardEvents(bool suppress) {
 }
 
 void Core::setMortalMode(bool value) {
+    std::lock_guard<std::mutex> lock(HotkeyMutex);
     mortal_mode = value;
 }
 
 void Core::setArmokTools(const std::vector<std::string> &tool_names) {
+    std::lock_guard<std::mutex> lock(HotkeyMutex);
     armok_tools.clear();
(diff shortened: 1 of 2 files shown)


myk002 (approved):

myk002:
quick way to test:
1. enable armok mode
2. go to the main dwarfmode map and hit Ctrl-Shift-T to bring up `gui/teleport` (which should come up)
3. disable armok mode
4. go back to the map and hit Ctrl-Shift-T again. nothing should happen

you can see more info about the change in processing by running this before doing the testing:
```
debugfilter set debug core keybinding
```

myk002:
that shouldn't be affected by this change, and should continue to work.

However, I think I see the issue that you might have been running into -- the tag filters in `gui/launcher` are not reset when you change the mortal mode config setting in `gui/control-panel`. I'll see if I can make that experience a little more smooth.


https://github.com/DFHack/dfhack/assets/977482/b174cbbb-bf78-4b56-b227-9685c482cfaf

myk002:
fixed in https://github.com/DFHack/dfhack/pull/4680 and https://github.com/DFHack/scripts/pull/1153

PR #4691, opened 2024-06-08
(WIP) documenation of type identities 

ab9rf:
WIP

[myk]: available in rendered form here: https://dfhack--4691.org.readthedocs.build/en/4691/docs/dev/data-identity.html

### DIFF
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -11,7 +11,7 @@ held by Dwarf Fortress in a transparent manner, but is also used for several oth
 
 The base class of the identity system is the class ``type_identity``, defined in ``DataDefs.h``. A ``type_identity`` object
 provides information about one _type_ of data object, in either Dwarf Fortress or DFHack, that can be manipulated as a discrete entity in Lua.
-With one specific exception (``global_identity``), there is a one-to-one relationship between C++ classes and ``type_identity`` objects.
+With one specific exception (``global_identity``), there is a one-to-one relationship between C++ types and ``type_identity`` objects.
 In Lua, objects that are being managed via the data identity system are represented as a Lua userdata object. The userdata object
 contains both a pointer to the C++ object itself and a pointer to a ``type_identity`` object that describes the data pointed
 by that pointer. Note that the userdata object does not own the objects pointed to by these pointers, and the Lua engine is
@@ -91,7 +91,7 @@ although this is not enforced at present.
 Having more than one ``struct_identity`` object for the same type might also potentially lead to misoperation.
 
 In general, there should be a one to one correspondence between ``type_identity`` objects and C++ types
-(with the special case that ``global_identity`` has no corresponding class).
+(with the special case that ``global_identity`` has no corresponding type).
 As far as we know, for any type other than ``virtual_identity``,
 violations of this constraint will not lead to misoperation, but this constraint should not be lightly violated.
 The Lua/C++ interface does, in a handful of places, assume that it can compare ``type_identity``
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -9,11 +9,11 @@ DFHack internally has a collection of C++ classes that provide metedata about th
 by various components of DFHack. This metadata is used primarily to enable the Lua scripting system to access data
 held by Dwarf Fortress in a transparent manner, but is also used for several other purposes within DFHack.
 
-The base class of the identity system is the class ``type_identity``, defined in ``DataDefs.h``. A `type_identity` object
+The base class of the identity system is the class ``type_identity``, defined in ``DataDefs.h``. A ``type_identity`` object
 provides information about one _type_ of data object, in either Dwarf Fortress or DFHack, that can be manipulated as a discrete entity in Lua.
-With one specific exception (``global_identity``), there is a one-to-one relationship between C++ classes and `type_identity` objects.
+With one specific exception (``global_identity``), there is a one-to-one relationship between C++ classes and ``type_identity`` objects.
 In Lua, objects that are being managed via the data identity system are represented as a Lua userdata object. The userdata object
-contains both a pointer to the C++ object itself and a pointer to a `type_identity` object that describes the data pointed
+contains both a pointer to the C++ object itself and a pointer to a ``type_identity`` object that describes the data pointed
 by that pointer. Note that the userdata object does not own the objects pointed to by these pointers, and the Lua engine is
 never responsible for managing their lifetimes.
 
@@ -25,7 +25,7 @@ never responsible for managing their lifetimes.
 - ``lua_read``: Used by the Lua engine to "read" the data from a C++ object into the Lua state.
 - ``lua_write``: Used by the Lua engine to "write" a value from the Lua state into a C++ data object.
 - ``build_metatable``: Create a Lua metatable in the specified Lua state corresponding to this type identity.
-- ``is_primitive``: indicates that `lua_read` will store a _copy_ of the object on the Lua stack instead of a non-owning reference to it. Used for types that have direct representations in Lua: numbers, booleans, simple strings
+- ``is_primitive``: indicates that ``lua_read`` will store a _copy_ of the object on the Lua stack instead of a non-owning reference to it. Used for types that have direct representations in Lua: numbers, booleans, simple strings
 - ``is_constructed``: indicates that creating a C++ instance of this type requires the use of a possibly nontrivial constructor. A type identity that is both primitive and constructed cannot be inserted into a container. At the moment the only type identity that is both primitive and constructed is ``stl_string_identity``, which wraps the C++ ``std::string`` type.
 - ``is_container``: indicates that the type is a container and thus implements the methods specific to ``container_identity``
 - ``allocate``: allocate, and construct if necessary, a C++ instance of this type. This may fail if the type does not support construction.
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -148,7 +148,7 @@ those objects must be indistinguishable from one another by anything other than 
 
 The ``type_identity`` object for a given C++ type can be obtained by using the ``get`` method of the ``df::identity_trait``
 trait class.
-More specifically, ``identity_class<T>::get()`` will return a pointer to a ``type_identity`` object for the type ``T``.
+More specifically, ``identity_traits<T>::get()`` will return a pointer to a ``type_identity`` object for the type ``T``.
 Developers who create new type identities must *either* provide an specialization of ``identity_trait`` that implements
 a ``get`` method that returns the correct ``type_identity``
 *or* ensure that a static instance of ``T::_identity`` exists for the type ``T``
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -46,7 +46,7 @@ There are plethora of subclasses of ``type_identity``:
       * ``stl_string_identity`` ``std::string``
       * ``xlsx_file_handle_identity`` special case
       * ``xlsx_sheet_handle_identity`` special case
-    * ``container_identity`` "containers" generally. note that all container types are homogenous. abstract base class
+    * ``container_identity`` "containers" generally. note that all container types are homogeneous (that is, the elements of the container must all be of the same type). abstract base class
       * ``bit_container_identity`` for containers that contain bools stored one element per _bit_ (rather than per byte)
         * ``bit_array_identity`` Dwarf Fortress's ``BitArray`` type
         * ``stl_bit_vector_identity`` ``std::vector<bool>``
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -154,8 +154,8 @@ a ``get`` method that returns the correct ``type_identity``
 *or* ensure that a static instance of ``T::_identity`` exists for the type ``T``
 (which will result in a template in :source:`DataDefs.h <library/include/DataDefs.h>` providing
 an implementation of ``get`` for that type).
-Note that this is only possible for compound types, and is the way that the _vast_ majority of
-compound type's have their identities specified (including all of those defined via codegen).
+Note that this is only possible for compound types, and is the way that the *vast* majority of
+compound types have their identities specified (including all of those defined via codegen).
 
 Because objects in the Lua environment are constructed as a pointer to the data and
 a pointer to the data's ``type_identity`` object, it is necessary for ``type_identity`` objects to have a lifetime
--- a/docs/dev/data-identity.rst
+++ b/docs/dev/data-identity.rst
@@ -101,6 +101,15 @@ that is, two copies having the same values will compare equal in all known such 
 Therefore, if two ``type_identity`` objects do exist (for any reason) for the same underlying C++ type,
 those objects must be indistinguishable from one another by anything other than their address.
 
+The ``type_identity`` object for a given C++ type can be obtained by using the ``get`` method of the ``identity_trait``
+trait class.
+More specifically, ``identity_class<T>::get()`` will return a pointer to a ``type_identity`` object for the type ``T``.
+Developers who create new type identities must _either_ provide an specialization of ``identity_trait`` that implements
+a ``get`` method that returns the correct ``type_identity`` _or_ insure that a static instance of ``T::_identity`` exists
+for the type T (which will result in a template in ``DataDefs.h`` providing one for that type).
+Note that this is only possible for compound types, and is the way that the _vast_ majority of
+compound type's have their identities specified (including all of those defined via codegen).
+
 Because objects in the Lua environment are constructed as a pointer to the data and
 a pointer to the data's ``type_identity`` object, it is necessary for ``type_identity`` objects to have a lifetime
 that exceeds the lifetime in the Lua environment of any object that exists anywhere in the Lua environment.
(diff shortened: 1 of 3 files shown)


docs/dev/data-identity.rst
```diff
@@ -0,0 +1,140 @@
...
+      * ``enum_identity`` C++ ``enum``
+      * ``struct_identity`` C++ ``class`` or ``structure``
+        * ``global_identity`` holds, as a quasiobject, handles for all of the known Dwarf Fortress program-scope static objects as if they were fields of an object called ``global``
+        * ``union_identity`` C++ ``union``
+        * ``other_vectors_identity`` special-case identity for the categorized subvectors of objects that appears in many of Dwarf Fortress's "handler" classes
+        * ``virtual_identity`` polymorphic C++ ``class`` or ``structure`` having a virtual table to handle virtual dispatch
+      * ``stl_string_identity`` ``std::string``
+      * ``xlsx_file_handle_identity`` special case
+      * ``xlsx_sheet_handle_identity`` special case
+    * ``container_identity`` "containers" generally. note that all container types are homogenous. abstract base class
```
myk002, line 49:
```suggestion
    * ``container_identity`` "containers" generally. note that all container types are homogeneous (that is, the elements of the container must all be of the same type). abstract base class
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+        * ``bit_array_identity`` Dwarf Fortress's ``BitArray`` type
+        * ``stl_bit_vector_identity`` ``std::vector<bool>``
+      * ``buffer_container_identity`` C++ static arrays and raw C++ pointers acting as arrays of unspecified bound
+      * ``enum_list_attr_identity`` (template) metaobject with metadata about a C++ enumeration; may also include additional metadata
+      * ``ptr_container_identity`` containers that contain pointers
+        *  ``stl_ptr_container_identity`` containers that are of the form ``std::vector<T*>`` for some ``T``
+      * ``ro_stl_container_identity`` (template) "read only containers"
+        * ``ro_stl_assoc_container_identity`` (template) ``std::map<KT,T>`` and ``std::unordered_map<KT,T>``
+      * ``stl_container_identity`` (template) ``std::vector<T>`` where ``T`` is _not_ a pointer (and not `bool`)
+    * ``opaque_identity`` opaque wrapper around any object all
```
myk002, line 60:
```suggestion
    * ``opaque_identity`` opaque wrapper around any object at all
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+from the core library instead of instantiating them locally. As a result, referencing an instance in a plugin that has
+not been instantiated in the core will result in linkage errors when linking the plugin to the core library.
+
+We could instead _not_ export the templated types, and thus their statically constructed identity objects,
+and instead allow the compiler to instantiate a local copy of these instances while compiling a plugin,
+but this would definitely result in a violation of the current requirement that there be at most one instance of the
+type identity object for a given C++ type across the entire program (including plugins loaded as a shared library).
+
+For primitive and opaque types the static constructors of the identity types
+are generally found in ``DataIdentity.h`` or ``DataIdentity.cpp``.
```
myk002, line 124:
```suggestion
are generally found in :source:`DataIdentity.h <library/include/DataIdentity.h>` or :source:`DataIdentity.cpp <library/DataIdentity.cpp>`.
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+Types defined by Dwarf Fortress are constructed in the header files and the related ``static*.inc`` files created by codegen,
+which are included into DFHack via ``DataStatics.cpp``.
```
myk002, line 126:
```suggestion
which are included into DFHack via :source:`DataStatics.cpp <library/DataStatics.cpp>`.
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Some plugins also define their own type identities. Type identities in plugins should be used with caution,
```
myk002, line 128:
```suggestion
Some plugins (e.g. :source:`blueprint <plugins/blueprint.cpp>`) also define their own type identities. Type identities in plugins should be used with caution,
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+because the DFHack plugin model allows plugins to be unloaded on request.
+Since the ``type_identity`` object is constructed within the the plugin's address space, and Lua objects that reference
+this ``type_identity`` object will hold a (borrowed) pointer to that object,
+unloading the plugin will result in a dangling pointer reference within the Lua environment.
+It is, at present, incumbent on plugin authors to ensure that they do not use plugin defined type identities on objects
+that will persist in the Lua environment beyond the lifetime of the plugin.
+Declaring a ``struct_identity`` in a plugin that is the child of another ``struct_identity`` will also result in
+a potentially dangling reference to that identity in the ``child`` vector of the parent identity, which means this
+must also be approached with caution.
```
myk002, line 137: they aren't automatically unregistered when the plugin unloads? If not, how is the plugin author to ensure safe usage?

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Namespace issues
+================
+
+The type identity system formally lives in the ``DFHack`` namespace.
+However, because the types created by the codegen process live in the ``df`` namespace,
+the identities needed to describe types coming from Dwarf Fortress are imported into the ``df`` namespace
+and at least some types are _only_ defined in that namespace.
```
myk002, line 145: could you give/link to an example here?

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,148 @@
...
+- ``is_primitive``: indicates that ``lua_read`` will store a _copy_ of the object on the Lua stack instead of a non-owning reference to it. Used for types that have direct representations in Lua: numbers, booleans, simple strings
+- ``is_constructed``: indicates that creating a C++ instance of this type requires the use of a possibly nontrivial constructor. A type identity that is both primitive and constructed cannot be inserted into a container. At the moment the only type identity that is both primitive and constructed is ``stl_string_identity``, which wraps the C++ ``std::string`` type.
+- ``is_container``: indicates that the type is a container and thus implements the methods specific to ``container_identity``
+- ``allocate``: allocate, and construct if necessary, a C++ instance of this type. This may fail if the type does not support construction.
+- ``copy``: copy the object at ``src`` onto ``tgt``. This uses ``memmove`` for primitive types, and C++ copy-assignment (when possible) for other types
+
+There are plethora of subclasses of ``type_identity``:
+
+* ``type_identity``
+  * ``constructed_identity`` anything with an internal structure (not primitive)
```
myk002, line 37: try adding blank lines around each segment at the same indentation level to get sphinx to accept this

if that's not enough, make each indent 4 spaces instead of 2

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,165 @@
...
+back to correct behavior as long as the shadow copies are indistinguishable from one another;
+that is, two copies having the same values will compare equal in all known such comparisons.
+Therefore, if two ``type_identity`` objects do exist (for any reason) for the same underlying C++ type,
+those objects must be indistinguishable from one another by anything other than their address.
+
+The ``type_identity`` object for a given C++ type can be obtained by using the ``get`` method of the ``identity_trait``
+trait class.
+More specifically, ``identity_class<T>::get()`` will return a pointer to a ``type_identity`` object for the type ``T``.
+Developers who create new type identities must _either_ provide an specialization of ``identity_trait`` that implements
+a ``get`` method that returns the correct ``type_identity`` _or_ insure that a static instance of ``T::_identity`` exists
```
myk002, line 108:
```suggestion
a ``get`` method that returns the correct ``type_identity`` _or_ ensure that a static instance of ``T::_identity`` exists
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,140 @@
+.. _data_identity:
+
+###########################
+DFHack Data Identity System
+###########################
+
+This article is an attempt to describe DFHack's data identity system.
+DFHack internally has a collection of C++ classes that provide metedata about the data in DF itself as well as data used
+by various components of DFHack. This metadata is used primarily to enable the Lua scripting system to access data
+held by Dwarf Fortress in a transparent manner, but is also used for several other purposes within DFHack.
```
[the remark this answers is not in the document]
ab9rf: i was kinda leaving that open for later. the "other purpose" i was actually thinking of most when i wrote that was vmethod interposes

docs/dev/data-identity.rst
ab9rf: as far as i know they're not, which means that it's unsafe for a plugin to declare a `struct_identity` with a non-NULL `parent`

docs/dev/data-identity.rst
ab9rf: the actual issue is that while the `type_identity` objects are in the `DFHack` namespace, the `identity_traits` trait class is in the `df` namespace

so i am actually mistaken there, and will rewrite this section

docs/dev/data-identity.rst
myk002: I think that's the advice that should be documented, then. For example, `blueprint`'s usage is safe, correct?

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,250 @@
...
+With one specific exception (``global_identity``), there is a one-to-one relationship between C++ types and ``type_identity`` objects.
+In Lua, objects that are being managed via the data identity system are represented as a Lua userdata object. The userdata object
+contains both a pointer to the C++ object itself and a pointer to a ``type_identity`` object that describes the data pointed
+by that pointer. Note that the userdata object does not own the objects pointed to by these pointers, and the Lua engine is
+never responsible for managing their lifetimes.
+
+``type_identity`` defines the following public methods:
+
+- ``byte_size``: returns the size, in bytes, of the object held
+
```
myk002, line 23: you don't need newlines between lines in a list, just before the first and after the last

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,250 @@
...
+        * ``float_identity`` (template) ``double`` and ``float``
+
+      * ``integer_identity_base`` abstract base for integral types
+
+        * ``integer_identity`` (template) ``int8_t``, ``int16_t``, ``int32_t``, ``size_t``, etc. lots of these
+
+    * ``pointer_identity`` any arbitrary C++ pointer (other than ``char*``)
+
+    * ``ptr_string_identity`` C-style (``char *``) string
```
myk002, line 117:
```suggestion
* ``type_identity``

  * ``constructed_identity`` anything with an internal structure (not primitive)
  
    * ``compound_identity`` anything with fields

      * ``bitfield_identity`` a structure defined with fields at bit rather than byte boundaries
      * ``enum_identity`` C++ ``enum``
      * ``struct_identity`` C++ ``class`` or ``structure``

        * ``global_identity`` holds, as a quasiobject, handles for all of the known Dwarf Fortress program-scope static objects as if they were fields of an object called ``global``
        * ``union_identity`` C++ ``union``
        * ``other_vectors_identity`` special-case identity for the categorized subvectors of objects that appears in many of Dwarf Fortress's "handler" classes
        * ``virtual_identity`` polymorphic C++ ``class`` or ``structure`` having a virtual table to handle virtual dispatch

      * ``stl_string_identity`` ``std::string``
      * ``xlsx_file_handle_identity`` special case
      * ``xlsx_sheet_handle_identity`` special case

    * ``container_identity`` "containers" generally. note that all container types are homogeneous (that is, the elements of the container must all be of the same type). abstract base class

      * ``bit_container_identity`` for containers that contain bools stored one element per *bit* (rather than per byte)

        * ``bit_array_identity`` Dwarf Fortress's ``BitArray`` type
        * ``stl_bit_vector_identity`` ``std::vector<bool>``

      * ``buffer_container_identity`` C++ static arrays and raw C++ pointers acting as arrays of unspecified bound
      * ``enum_list_attr_identity`` (template) metaobject with metadata about a C++ enumeration; may also include additional metadata

      * ``ptr_container_identity`` containers that contain pointers

        *  ``stl_ptr_container_identity`` containers that are of the form ``std::vector<T*>`` for some ``T``

      * ``ro_stl_container_identity`` (template) "read only containers"

        * ``ro_stl_assoc_container_identity`` (template) ``std::map<KT,T>`` and ``std::unordered_map<KT,T>``

      * ``stl_container_identity`` (template) ``std::vector<T>`` where ``T`` is *not* a pointer (and not ``bool``)

    * ``opaque_identity`` opaque wrapper around any type, provides no functionality
    * ``stl_string_identity`` ``std::string``

  * ``function_identity_base`` abstract base class for ``function_identity``

    * ``function_identity`` (template) wrapper around a C++ function that can be invoked from Lua

  * ``primitive_identity`` wrapper around a primitive type. primitive types are fixed-length objects with no internal structure

    * ``bool_identity`` ``bool``
    * ``number_identity_base`` abstract base for numeric types

      * ``float_identity_base`` abstract base for floating point types

        * ``float_identity`` (template) ``double`` and ``float``

      * ``integer_identity_base`` abstract base for integral types

        * ``integer_identity`` (template) ``int8_t``, ``int16_t``, ``int32_t``, ``size_t``, etc. lots of these

    * ``pointer_identity`` any arbitrary C++ pointer (other than ``char*``)
    * ``ptr_string_identity`` C-style (``char *``) string
```

myk002: I'm not sure this is actually any better, but you could shave off a few newlines if you wanted to

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,250 @@
...
+
+- ``lua_read``: Used by the Lua engine to "read" the data from a C++ object into the Lua state.
+
+- ``lua_write``: Used by the Lua engine to "write" a value from the Lua state into a C++ data object.
+
+- ``build_metatable``: Create a Lua metatable in the specified Lua state corresponding to this type identity.
+
+- ``is_primitive``: indicates that ``lua_read`` will store a *copy* of the object on the Lua stack instead of a non-owning reference to it. Used for types that have direct representations in Lua: numbers, booleans, simple strings
+
+- ``is_constructed``: indicates that creating a C++ instance of this type requires the use of a possibly nontrivial constructor. A type identity that is both primitive and constructed cannot be inserted into a container. At the moment the only type identity that is both primitive and constructed is ``stl_string_identity``, which wraps the C++ ``std::string`` type.
```
chdoc, line 36: I find this slightly puzzling, because I have definitely seen ` map<string, map<string, string>>` being returned to Lua. https://github.com/DFHack/dfhack/blob/703123e5ec5f2c8d995dbeb9da053b11e37bd5d4/plugins/buildingplan/buildingplan.cpp#L974
Maybe this needs some clarification?

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,250 @@
...
+###########################
+
+This article is an attempt to describe DFHack's data identity system.
+DFHack internally has a collection of C++ classes that provide metedata about the data in DF itself as well as data used
+by various components of DFHack. This metadata is used primarily to enable the Lua scripting system to access data
+held by Dwarf Fortress in a transparent manner, but is also used for several other purposes within DFHack.
+
+The base class of the identity system is the class ``type_identity``, defined in :source:`DataDefs.h <library/include/DataDefs.h>`. A ``type_identity`` object
+provides information about one *type* of data object, in either Dwarf Fortress or DFHack, that can be manipulated as a discrete entity in Lua.
+With one specific exception (``global_identity``), there is a one-to-one relationship between C++ types and ``type_identity`` objects.
```
chdoc, line 14:
```suggestion
With the exception of (``global_identity``), which provides handles for all of the known Dwarf Fortress program-scope static objects, there is a one-to-one relationship between C++ types and ``type_identity`` objects.
```

chdoc (commented):
While this certainly doesn't explain *all* of the identity system, I think it is already pretty helpful. :+1:

docs/dev/data-identity.rst
ab9rf: when an object like this is returned to lua, the object on the lua stack will be a nonowning reference. which suggests that that object might well leak if/when lua GCs the userdata object. we may need to look into that

ab9rf: update: that particular instance is handled _outside_ of the identity trait system; it's handled by `Lua::Push<T>`, which does not always do the same thing as `identity_traits<T>::get->lua_read`. the former handles some special cases that the latter does not. this may or may not be desirable.

chdoc: Yes, I saw the thread on discord. In light of that, this is probably fine.

docs/dev/data-identity.rst
ab9rf: yes, the structure it defines has a `NULL` parent so it should be safe

ab9rf:
i realize this is incomplete but partial docs are better than no docs, and it can always be extended later

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,252 @@
+.. _data_identity:
+
+###########################
+DFHack Data Identity System
+###########################
+
+This article is an attempt to describe DFHack's data identity system.
+DFHack internally has a collection of C++ classes that provide metedata about the data in DF itself as well as data used
```
myk002, line 8:
```suggestion
DFHack internally has a collection of C++ classes that provide metadata about the data in DF itself as well as data used
```

docs/dev/data-identity.rst
```diff
@@ -0,0 +1,252 @@
...
+by various components of DFHack. This metadata is used primarily to enable the Lua scripting system to access data
+held by Dwarf Fortress in a transparent manner, but is also used for several other purposes within DFHack.
```
myk002, line 10:
```suggestion
held by Dwarf Fortress in a transparent manner, but is also used for several other purposes within DFHack, such as rerouting virtual method calls.
```

myk002 (approved):

PR #4702, opened 2024-06-10
constify identities

ab9rf:
This PR marks all of the core data identity classes as `const` to reflect that these types are effectively immutable. 

As noted elsewhere, `compound_identity` has a small degree of technical mutability due to lazy initialization, and `virtual_identity` has mutability due to its dual purpose for implementing vtable interposes; these are both accommodated by marking the related class members as `mutable`. This results in some `const_cast`s, in the `compound_identity` and `struct_identity` constructors, and in a couple of places in code related to virtual identities. There are also several `const_cast`s associated with storing type identity pointers into Lua metatables and lightuserdata objects, which are downcasts into `void*` and thus have to be `const_cast`ed to maintain correctness. Some of these are arguably code smells arguing for a redesign of the code in question, but that's for a future PR.

I also rewrote a couple of methods in `VMethodInterposeLinkBase` to avoid using `map::operator[]` (because the latter breaks `const`ness); this has the side effect that searching for a vmethod by index will no longer insert an empty pointer into a virtual class's vtable interpose map on a lookup for a previously undefined index.

While these changes will enable making a significant fraction of `data_identity` instances `constinit` or `constexpr` (which has advantages in terms of more predictable static initialization, faster startup, and possibly other optimizations), such changes are not included in this PR.

### DIFF
--- a/library/DataDefs.cpp
+++ b/library/DataDefs.cpp
@@ -42,44 +42,44 @@ distribution.
 using namespace DFHack;
 
 
-void *type_identity::do_allocate_pod() {
+void *type_identity::do_allocate_pod() const {
     size_t sz = byte_size();
     void *p = malloc(sz);
     memset(p, 0, sz);
     return p;
 }
 
-void type_identity::do_copy_pod(void *tgt, const void *src) {
+void type_identity::do_copy_pod(void *tgt, const void *src) const {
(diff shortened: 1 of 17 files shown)


myk002 (approved):
that's a lot of const! nice!

PR #4710, opened 2024-06-13
add option to ``unsuspend`` that unsuspends all construction jobs

chdoc:
Note 1: also fixes a missing core suspension in `do_unsuspend_command`
Note 2: running `unsuspend` no longer resets the counter for the next regular cycle. Running `unsuspend` while `suspendmanager` is enabled, doesn't make sense anyway.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -82,7 +82,7 @@ Template for new versions:
 - `autobutcher`: do not butcher pregnant (or brooding) females
 - `quickfort`: support buildable instruments
 - `nestboxes`: increase the scanning frequency for fertile eggs to reduce the chance that they get snarfed by eager dwarves
-- `suspendmanager`: add option to ``unsuspend`` that unsuspends all construction jobs
+- `suspendmanager`: add option to ``unsuspend`` that unsuspends all jobs, regardless of potential issues
 
 ## Documentation
 - `installing`: add instructions for how to use Steam DFHack with non-Steam DF (for DFHack auto-updates and cloud backups)
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -82,6 +82,7 @@ Template for new versions:
 - `autobutcher`: do not butcher pregnant (or brooding) females
 - `quickfort`: support buildable instruments
 - `nestboxes`: increase the scanning frequency for fertile eggs to reduce the chance that they get snarfed by eager dwarves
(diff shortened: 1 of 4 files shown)


docs/plugins/suspendmanager.rst
```diff
@@ -49,6 +49,9 @@ Options
 ``-s``, ``--skipblocking``
     Also resume jobs that may block other jobs.
 
+``-a``, ``--all``
```
myk002, line 52: I'd call this `-f`, `--force` instead of "all" since it's doing something that may be undesirable

docs/plugins/suspendmanager.rst
```diff
@@ -49,6 +49,9 @@ Options
...
+    Resume all construction jobs, even if they are submerged or will collapse on completion.
```
myk002, line 53: does this imply `--skipblocking` as well?

plugins/suspendmanager.cpp
```diff
@@ -957,10 +963,10 @@ static bool suspendmanager_isKeptSuspended(df::job *job) {
     }
 }
 
-static void suspendmanager_runOnce(color_ostream &out, bool blocking) {
+static void suspendmanager_runOnce(color_ostream &out, bool blocking, bool unsuspend_everything) {
     auto save = suspendmanager_instance->prevent_blocking;
     suspendmanager_instance->prevent_blocking = blocking;
-    do_cycle(out);
+    suspendmanager_instance->do_cycle(out, unsuspend_everything);
```
myk002, line 969: the pattern common to other `do_cycle` tools is to update the last run tick counter when manually run like this, but I agree that it's not a big deal to skip the update for a tool that runs this frequently when enabled.

docs/changelog.txt
```diff
@@ -82,6 +82,7 @@ Template for new versions:
 - `autobutcher`: do not butcher pregnant (or brooding) females
 - `quickfort`: support buildable instruments
 - `nestboxes`: increase the scanning frequency for fertile eggs to reduce the chance that they get snarfed by eager dwarves
+- `suspendmanager`: add option to ``unsuspend`` that unsuspends all construction jobs
```
myk002, line 85:
```suggestion
- `suspendmanager`: add option to ``unsuspend`` that unsuspends all jobs, regardless of potential issues
```

myk002 (approved):

Issue #4706, opened 2024-06-11
more numeric formats for formatted numbers


myk002:
From [Reddit](https://www.reddit.com/r/dwarffortress/comments/1ddizn1/comment/l86aef0/?utm_source=share&utm_medium=web2x&context=3):

Another option: 3 significant figures. So (dividing by 10 each time) "1.23M", "123k", "12.3k", "1.23k", "123". If you have a million of something, does it really matter if it's a million and one or a million and two?
Change "3 SF" to however many significant figures you think are, well, significant...

Also from [Reddit](https://www.reddit.com/r/dwarffortress/comments/1ddizn1/comment/l86oz49/?utm_source=share&utm_medium=web2x&context=3):

Honestly seeing uncertain quantities like "1E3" would be kinda neat. Useful? Doubtful. But it would be clear and usable (especially if the user purposefully set it themselves) since you could just compare the E numbers. Doubtful that it's easier than just looking at which string of 0's is longer, but Dwarf Fortress isn't about taking the easy way out is it?

PR #4715, opened 2024-06-14
add SI suffix number formatting

myk002:
Fixes https://github.com/DFHack/dfhack/issues/4706

### DIFF
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -43,10 +43,12 @@ distribution.
 #include <stdarg.h>
 #include <string.h>
 #include <cstdlib>
+#include <cmath>
 
 #include <sstream>
 #include <map>
 #include <array>
+#include <unordered_map>
 
 NumberFormatType preferred_number_format_type = NumberFormatType::DEFAULT;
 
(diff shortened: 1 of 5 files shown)


library/MiscUtils.cpp
```diff
@@ -83,6 +85,45 @@ DFHACK_EXPORT void imbue_with_locale(std::ostringstream &ss, NumberFormatType ty
...
+        return "0";
+
+    std::ostringstream ss;
+    if (sig_figs > sig_fig_db.size()) {
+        ss << std::fixed << num;
+        return ss.str();
+    }
+
+    int sf = (int)sig_figs;
+    int magnitude = std::min((size_t)std::floor(std::log10(std::abs(num))) + 1, sig_fig_db.size() - 1);
```
ab9rf, line 122: this code will behave very strangely for arguments between -1.0 and 1.0 (except for 0.0 exactly): log10 will be negative, which, when cast to the unsigned `size_t`, convert to an extremely large positive number, which at least gets clamped by `std::min` to the highest prefix. as a result, arguments between -1.0 and 1.0 (excluding 0) will result in the `T` prefix being used on the unit... which is almost certainly not at all what is intended or desired

PR #4743, opened 2024-06-22
[tiletypes] reindex pathfinding after altering the map

myk002:
this was done in other parts of tiletypes, but not in the code path called by gui/tiletypes

### DIFF
--- a/plugins/tiletypes.cpp
+++ b/plugins/tiletypes.cpp
@@ -1104,6 +1104,9 @@ static PaintResult paintTile(MapExtras::MapCache &map, const df::coord &pos,
                     Maps::setTileAquifer(pos, target.aquifer == 2);
             }
         };
+
+        // force the game to recompute its walkability cache on the next tick
+        world->reindex_pathfinding = true;
     }
     return PaintResult();
 }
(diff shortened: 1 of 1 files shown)


Bumber64:
This is probably out of the scope of this PR, but you could probably replace stuff like
```
std::string command = commands[loc++];
tolower(command);
```
with
```
std::string command = toLower_cp437(commands[loc++]);
```
and get rid of the plugin's redundant `toupper` and `tolower` functions.

tatoyoda600:
Looks good, just have a question.
Would it maybe be better to actually move the `reindex_pathfinding` code to either the `postWrite`, `paintTile`/`paintArea`, or `paintTileProcessing` section?
That way it's included in the standard processing flow and doesn't depend on it being included in the calling function, since it should always reindex the pathfinding after a map edit

tatoyoda600:
I don't know how to leave a code suggestion thing in GitHub, but in this last commit you moved the line to an inaccessible place, right after the `return` in `paintTile` (Also it's only in the `paintTile` function and not in the `paintArea` one)
It should either be inside the `postWrite`s in both `paintTile` and `paintArea`, or before the `return PaintResult` in both functions.

PR #4787, opened 2024-07-04
Fix changelayer plugin, do geology lookup properly

quietust:
Depending on embark location, it sometimes changed the wrong layers and resulted in nothing appearing to change at all.


myk002 (commented):
could you add a changelog entry under Fixes?

myk002 (approved):

Issue #4318, opened 2024-03-02
`DFLinkedList` does not satisfy `input_iterator`


ab9rf:
Our proxy for iterating the job list does not fully satisfy the C++ STL requirements to be an `input_iterator` which limits the ability to fully use this proxy type with more recent developements in C++. This should be addressed.

![image](https://github.com/DFHack/dfhack/assets/1445859/dac26db3-0278-4550-a3a2-7b955ea00176)

[discord link](https://discord.com/channels/793331351645323264/807443262704058408/1213500700089262110)

ab9rf:
After poking at this for far too long the issue appears to be that `DfLinkedList`'s iterator type is `proxy` while its const iterator type is `I * &`; as these types are not the same the template doesn't comport with the requirements of an `input_iterator`.

Looking through the history, this used to not be the case but @BenLubar changed it (commit 1606483e7e54c455deaacb1f617644e1460ddb16) to the current state of affairs as a "performance improvement". It seems to me that either backing this out _or_ changing the const iterator to also use the proxy might yield compliance with the expectations of `input_iterator`, but I'm still sorting through the specifics.

BenLubar:
Looks like I put `proxy` back in but only partially right after that commit: https://github.com/DFHack/dfhack/commit/b5eb541fd34c44c6268d9464f58d3392ffa8394e

It was so long ago I can't remember anything about what I was doing there.

chdoc:
I'm starting to believe that whatever performance problems the `proxy` was trying to address, has become irrelevant in the meantime. In fact, changing the definition of `struct DfLinkedList` to be the empty struct, causes exactly two compilation errors: the two simple should-be-`const` iterations in `suspendmanager.cpp`, because I am a fan of ranged loops. 

In light of this, I would say we can safely replace this with a naive implementation meeting the STL requirements.

chdoc:
After some more analysis, I have come to the conclusion that the purpose of the `proxy` is to provide "proxy references" that, unlike naked `I*&`, implement an assignment operator that takes care of setting/unsetting the `list_link` fields of the item being placed/replaced:

https://github.com/DFHack/dfhack/commit/b5eb541fd34c44c6268d9464f58d3392ffa8394e#diff-fc625092d138aa24048ed147fb2c381fb05559c45890b82d9eb7ae399db32461R262-R276

[...]

PR #4795, opened 2024-07-06
improve job completion handling performance

dhthwy:
Experimental changes. Though parts have been rewritten, it should be logically equivalent. There may be some DRY issues -- that was intentional as this is a very hot function. 

Assuming I understood correctly, tweaks inspired by ideas from @ab9rf and @chdoc

Of course it is always better handled with hooks, if provided.

### DIFF
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -500,19 +500,20 @@ static void manageJobStartedEvent(color_ostream& out) {
     unordered_set<int32_t> newStartedJobs;
     newStartedJobs.reserve(startedJobs.size());
 
-    for (df::job_list_link* link = &df::global::world->jobs.list; link != nullptr; link = link->next) {
-        if (link->item != nullptr) {
-            df::job& job = *link->item;
-            // To have a worker, there must be at least one general_ref.
+    for (const auto jobPtr : df::global::world->jobs.list) {
+        auto& job = *jobPtr;
+        // To have a worker, there must be at least one general_ref.
+        if (!startedJobs.contains(job.id)) {
             if (job.general_refs.size() > 0 && Job::getWorker(&job)) {
                 newStartedJobs.emplace(job.id);
-                if (!startedJobs.count(job.id)) {
-                    for (auto &[_,handle] : copy) {
-                        DEBUG(log,out).print("calling handler for job started event\n");
-                        run_handler(out, EventType::JOB_STARTED, handle, &job);
-                    }
+                for (auto &[_,handle] : copy) {
+                    DEBUG(log,out).print("calling handler for job started event\n");
+                    run_handler(out, EventType::JOB_STARTED, handle, &job);
                 }
             }
+        } else {
+            // Make sure it stays in startedJobs container for as long as the job is active.
+            newStartedJobs.emplace(job.id);
         }
     }
     startedJobs = std::move(newStartedJobs);
@@ -557,9 +558,11 @@ static void manageJobCompletedEvent(color_ostream& out) {
                     || (seenJob->general_refs.size() != job.general_refs.size())) {
                 seenIt->second = Job::JobUniquePtr(Job::cloneJobStruct(&job, true));
             }
-        } else {
+        } else if (job.completion_timer != -1) {
             seenJobs.emplace(job.id, Job::JobUniquePtr(Job::cloneJobStruct(&job, true)));
         }
+        // We still need to push back all jobs to maintain the invariant of the algorithm used with
+        // prevJobs and nowJobs. See comments below for more details.
         nowJobs.emplace_back(job.id, job.completion_timer, (bool)job.flags.bits.repeat);
     }
 
@@ -638,27 +641,32 @@ static void manageJobCompletedEvent(color_ostream& out) {
          * than any in prevJobs.
          */
         while (prevIt != prevJobs.end()) {
-            if (nowIt == nowJobs.end() || prevIt->id != nowIt->id) { // job ID is in prevJobs. ID does not exist in nowJobs.
-                auto& prevJob = *prevIt;
-                df::job& seenJob = *seenJobs[prevJob.id];
-                // recently finished or cancelled job
-                if (!prevJob.flags_bits_repeat && prevJob.completion_timer == 0) {
-                    for (auto& [_, handle] : copy) {
-                        DEBUG(log, out).print("calling handler for job completed event\n");
-                        run_handler(out, EventType::JOB_COMPLETED, handle, (void*)&seenJob);
+            auto& prevJob = *prevIt;
+            auto seenIt = seenJobs.find(prevJob.id);
+            if (nowIt == nowJobs.end() || prevJob.id != nowIt->id) { // job ID is in prevJobs. ID does not exist in nowJobs.
+                if (seenIt != seenJobs.end()) {
+                    // recently finished or cancelled job
+                    if (!prevJob.flags_bits_repeat && prevJob.completion_timer == 0) {
+                        df::job& seenJob = *seenIt->second;
+                        for (auto& [_, handle] : copy) {
+                            DEBUG(log, out).print("calling handler for job completed event\n");
+                            run_handler(out, EventType::JOB_COMPLETED, handle, (void*)&seenJob);
+                        }
                     }
+                    // note: this is only safe to erase when the job is not in nowJobs.
+                    seenJobs.erase(prevJob.id);
                 }
-                seenJobs.erase(prevJob.id);
             } else { // prevIt job ID and nowIt job ID are equal.
-                // could have just finished if it's a repeat job
-                auto& prevJob = *prevIt;
-                if (prevJob.flags_bits_repeat && prevJob.completion_timer == 0
-                        && nowIt->completion_timer == -1) {
-                    df::job& seenJob = *seenJobs[prevJob.id];
-                    // still false positive if cancelled at EXACTLY the right time, but experiments show this doesn't happen
-                    for (auto& [_, handle] : copy) {
-                        DEBUG(log, out).print("calling handler for repeated job completed event\n");
-                        run_handler(out, EventType::JOB_COMPLETED, handle, (void*)&seenJob);
+                if (seenIt != seenJobs.end()) {
+                    // could have just finished if it's a repeat job
+                    if (prevJob.flags_bits_repeat && prevJob.completion_timer == 0
+                            && nowIt->completion_timer == -1) {
+                        df::job& seenJob = *seenIt->second;
+                        // still false positive if cancelled at EXACTLY the right time, but experiments show this doesn't happen
+                        for (auto& [_, handle] : copy) {
+                            DEBUG(log, out).print("calling handler for repeated job completed event\n");
+                            run_handler(out, EventType::JOB_COMPLETED, handle, (void*)&seenJob);
+                        }
                     }
                 }
                 // prevIt has caught up to nowIt.
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -217,14 +217,8 @@ struct JobCompleteData {
     uint32_t flags_bits_repeat;
 };
 
-struct JobDeleter {
-    void operator()(df::job *ptr) const {
-        if (ptr)
-            Job::deleteJobStruct(ptr, true);
-    }
-};
-using JobUniquePtr = std::unique_ptr<df::job, JobDeleter>;
-static std::unordered_map<int32_t, JobUniquePtr> seenJobs;
+
+static std::unordered_map<int32_t, Job::JobUniquePtr> seenJobs;
 static std::vector<JobCompleteData> prevJobs;
 
 //active units
@@ -561,10 +555,10 @@ static void manageJobCompletedEvent(color_ostream& out) {
             if (seenJob->flags.whole != job.flags.whole
                     || (seenJob->items.size() != job.items.size())
                     || (seenJob->general_refs.size() != job.general_refs.size())) {
-                seenIt->second = JobUniquePtr(Job::cloneJobStruct(&job, true));
+                seenIt->second = Job::JobUniquePtr(Job::cloneJobStruct(&job, true));
             }
         } else {
-            seenJobs.emplace(job.id, JobUniquePtr(Job::cloneJobStruct(&job, true)));
+            seenJobs.emplace(job.id, Job::JobUniquePtr(Job::cloneJobStruct(&job, true)));
         }
         nowJobs.emplace_back(job.id, job.completion_timer, (bool)job.flags.bits.repeat);
     }
@@ -682,7 +676,7 @@ static void manageJobCompletedEvent(color_ostream& out) {
      * these missed jobs.
      */
     if (seenJobs.size() > nowJobs.size() * 2) {
-        std::unordered_map<int32_t, JobUniquePtr> newMap;
+        std::unordered_map<int32_t, Job::JobUniquePtr> newMap;
         newMap.reserve(nowJobs.size());
         for (auto& data : nowJobs) {
             auto it = seenJobs.find(data.id);
--- a/library/modules/EventManager.cpp
+++ b/library/modules/EventManager.cpp
@@ -211,7 +211,21 @@ static int32_t lastJobId = -1;
 static unordered_set<int32_t> startedJobs;
 
 //job completed
-static unordered_map<int32_t, df::job*> prevJobs;
+struct JobCompleteData {
+    int32_t id;
+    int32_t completion_timer;
+    uint32_t flags_bits_repeat;
+};
+
+struct JobDeleter {
+    void operator()(df::job *ptr) const {
+        if (ptr)
+            Job::deleteJobStruct(ptr, true);
+    }
+};
+using JobUniquePtr = std::unique_ptr<df::job, JobDeleter>;
+static std::unordered_map<int32_t, JobUniquePtr> seenJobs;
+static std::vector<JobCompleteData> prevJobs;
 
 //active units
 static unordered_set<int32_t> activeUnits;
@@ -292,9 +306,7 @@ void DFHack::EventManager::onStateChange(color_ostream& out, state_change_event 
     if ( event == DFHack::SC_MAP_UNLOADED ) {
         lastJobId = -1;
         startedJobs.clear();
-        for (auto &prevJob : prevJobs) {
-            Job::deleteJobStruct(prevJob.second, true);
-        }
+        seenJobs.clear();
         prevJobs.clear();
         tickQueue.clear();
         livingUnits.clear();
@@ -492,30 +504,33 @@ static void manageJobStartedEvent(color_ostream& out) {
     multimap<Plugin*, EventHandler> copy(handlers[EventType::JOB_STARTED].begin(), handlers[EventType::JOB_STARTED].end());
 
     unordered_set<int32_t> newStartedJobs;
+    newStartedJobs.reserve(startedJobs.size());
 
-    for (df::job_list_link* link = &df::global::world->jobs.list; link->next != nullptr; link = link->next) {
-        df::job* job = link->next->item;
-        if (!job || !Job::getWorker(job))
-            continue;
-
-        int32_t j_id = job->id;
-        newStartedJobs.emplace(j_id);
-        if (!startedJobs.count(j_id)) {
-            for (auto &[_,handle] : copy) {
-                DEBUG(log,out).print("calling handler for job started event\n");
-                run_handler(out, EventType::JOB_STARTED, handle, job);
+    for (df::job_list_link* link = &df::global::world->jobs.list; link != nullptr; link = link->next) {
+        if (link->item != nullptr) {
+            df::job& job = *link->item;
+            // To have a worker, there must be at least one general_ref.
+            if (job.general_refs.size() > 0 && Job::getWorker(&job)) {
+                newStartedJobs.emplace(job.id);
+                if (!startedJobs.count(job.id)) {
+                    for (auto &[_,handle] : copy) {
+                        DEBUG(log,out).print("calling handler for job started event\n");
+                        run_handler(out, EventType::JOB_STARTED, handle, &job);
+                    }
+                }
             }
         }
     }
-
-    startedJobs = newStartedJobs;
+    startedJobs = std::move(newStartedJobs);
 }
 
 //helper function for manageJobCompletedEvent
-//static int32_t getWorkerID(df::job* job) {
-//    auto ref = findRef(job->general_refs, general_ref_type::UNIT_WORKER);
-//    return ref ? ref->getID() : -1;
-//}
+#if 0
+static int32_t getWorkerID(df::job* job) {
+    auto ref = findRef(job->general_refs, general_ref_type::UNIT_WORKER);
+    return ref ? ref->getID() : -1;
+}
+#endif
 
 /*
 TODO: consider checking item creation / experience gain just in case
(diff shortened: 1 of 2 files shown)


library/modules/EventManager.cpp
```diff
@@ -527,12 +533,18 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
-    for ( df::job_list_link* link = &df::global::world->jobs.list; link != nullptr; link = link->next ) {
-        if ( link->item == nullptr )
-            continue;
-        nowJobs.emplace(link->item->id, link->item);
+    multimap<Plugin*, EventHandler> copy(handlers[EventType::JOB_COMPLETED].begin(), handlers[EventType::JOB_COMPLETED].end());
+    std::unordered_map<int32_t, JobCompleteData> nowJobs;
+    nowJobs.reserve(prevJobs.size());
+    for (df::job_list_link* link = &df::global::world->jobs.list; link != nullptr; link = link->next) {
+        if (link->item != nullptr) {
+            df::job& job = *link->item;
```
chdoc, line 541: This may be a simple as:
```suggestion
for (auto job : df::global::world->jobs.list)
```

library/modules/EventManager.cpp
```diff
@@ -527,12 +533,18 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
+            // Can we pack these into a single integer?
+            nowJobs[job.id] = { job.completion_timer, job.flags.bits.repeat };
+            if (savedJobs.find(job.id) == savedJobs.end()) {
+                savedJobs.emplace(job.id, Job::cloneJobStruct(&job, true));
+            }
```
chdoc, line 546: If I understand correctly, this will clone jobs only once, namely when they are encountered the first time. However, at this point items and workers will, for the most part, not yet be attached to the job.

Maybe it makes more sense to only clone jobs that have a worker attached to them. If a job disappears without ever having been picked up a worker, it most likely was cancelled. And I don't think any jobs can successfully complete the same tick they get picked up by a worker.

myk002: I *think* that's true that jobs can't be completed in the same tick that they get a worker, even with `fastdwarf`, but we should also be able to handle the case where we happen to see the job in the same tick that it gets completed. This can happen if all the registered handlers have a non-zero refresh rate.

dhthwy: Yeah, I wasn't sure about it copying only when first seen. It shouldn't hurt much to re-copy (or only some bits that changed) when a worker is added or items changed.

Tick1 is almost always greater than tick0 from my observations. I figured there must've been a good reason for it not to handle same tick completions so I left that as-is. Will try it with 0 delta.

ab9rf:
Profiling results, using myk's "myk_savepoint_new" fort.

Develop HEAD: 300s clock, DF 245.223s, DFHack 14.225s, Lua 13.328s, for a DFHack cpu share of 10.10%
This PR: 300s clock, DF 254.416s, DFHack 9.368s, Lua 12.707s, for DFHack cpu share of 7.98%

This is a 21% reduction in DFHack's CPU share and brings us under our target of 10%

Profiling procedure was to load the fort, start the profiler, and collect for 300s without any interaction other than unpausing the fort 2s after the profiler starts collection

`manageJobCompletedEvent` went from 15.094s in reference to 6.978s in the PR - this counts both time charged to dfhack as well as time charged to ntdll for heap management so this actually reduces our impact even more than the above numbers indicate. Unfortunately, I can't readily differentiate ntdll time due to calls from dfhack from ntdll time due to calls from DF itself, so the above numbers are a _conservative_ estimate of how much improvement this PR provides

library/modules/EventManager.cpp
```diff
@@ -211,7 +211,12 @@ static int32_t lastJobId = -1;
 static unordered_set<int32_t> startedJobs;
 
 //job completed
-static unordered_map<int32_t, df::job*> prevJobs;
+struct JobCompleteData {
+    int32_t completion_timer;
+    uint32_t flags_bits_repeat;
+};
+static std::unordered_map<int32_t, df::job*> savedJobs;
+static std::unordered_map<int32_t, JobCompleteData> prevJobs;
```
chdoc, line 219: I wonder if `prevJobs` should actually be a vector rather than a map, extending `JobCompleteData` with the job id. 

If I understand correctly, DF keeps the job list ordered by job id, so this would automatically ensure that `nowJobs` and `prevJobs` are ordered as well. Thus, when iterating over one and locating data in the other, lookup can be done in *total* time O(n) without hashing, by keeping the index where we last found (or didn't find) the  associated data.

dhthwy: I'm sure it's doable for some slight gains, and possibly bigger than slight though -- only way to know is to try. 

It's more complicated due to repeat jobs retaining IDs and there are times when there's large holes in the job id from previous generation.

library/modules/EventManager.cpp
dhthwy: That should be a C-style linked list - we don't get to use the fancy range-style-for those types of containers.

chdoc: I literally copied that line from the implementation of `suspendmanager.cpp`: https://github.com/DFHack/dfhack/blob/9e5c4b8ac6f7def545b1d920c1a72e0bb0380cd3/plugins/suspendmanager.cpp#L643

library/modules/EventManager.cpp
chdoc: I don't think large holes actually matter. Sure, when there is a large hole, the index into the second vector may remain stationary for a while. Thus, to be fully precise, I suppose the complexity is O(n1+n2) equality checks. One either advances to the next `nowJob` or to the next `prevJob`.

And repeat jobs need to be checked, just as before. This is only a data structure change, the algorithm remains the same.

library/modules/EventManager.cpp
ab9rf: our currently implementation of `DFLinkedList` does not satisfy the InputIterator concept requirements and thus can't be used with ranges until someone fixes this. See #4318

myk002:
I saw you marked this as draft -- what are the open questions?

myk002:
> job->job_items.elements -- Items required for the job (doesn't mutate? no, not from what I've seen.)

doesn't mutate in most circumstances. `buildingplan` will mutate it as it attaches items to protect against the case where the game is saved while items are partially attached and then loaded by a player that does not have DFHack installed.

library/modules/EventManager.cpp
```diff
@@ -597,57 +630,71 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
-            //still false positive if cancelled at EXACTLY the right time, but experiments show this doesn't happen
-            for (auto &[_,handle] : copy) {
-                DEBUG(log,out).print("calling handler for repeated job completed event\n");
-                run_handler(out, EventType::JOB_COMPLETED, handle, (void*) &job0);
+    // if it happened within a tick, must have been cancelled by the user or a plugin: not completed
+    // note observation: delta 0 tick events happens normally on load, save, paused (whether by user or game)
+    if (tick1 >= tick0) {
+        auto prevIt = prevJobs.begin();
+        auto nowIt = nowJobs.begin();
+        while (prevIt != prevJobs.end() && nowIt != nowJobs.end()) {
```
chdoc, line 638: I think you still need to process the remaining `prevJobs`, even after `nowJobs` has reached the end. Alternatively, you can always iterate to the end of `prevJobs`, and only advance `nowJobs` if it isn't at the end already.

library/modules/EventManager.cpp
```diff
@@ -527,12 +539,33 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
+        if (seenJobs.find(job.id) == seenJobs.end()) {
+            seenJobs.emplace(job.id, Job::cloneJobStruct(&job, true));
+            //out.print("NEWJOB: %d, %ld, %ld, %ld\n", job.id, job.items.size(), job.specific_refs.size(), job.general_refs.size());
+        } else {
+            df::job& seenJob = *seenJobs[job.id];
+
+            //out.print("SEEN0: %d, %ld, %ld, %d, %ld, %d %s\n", seenJob.id, seenJob.items.size(), seenJob.job_items.elements.size(), seenJob.flags.bits.working, seenJob.general_refs.size(), getWorkerID(&seenJob), ENUM_ATTR(job_type, caption, seenJob.job_type));
+            //out.print("SEEN1: %d, %ld, %ld, %d, %ld, %d %s\n", job.id, job.items.size(), job.job_items.elements.size(), job.flags.bits.working, job.general_refs.size(), getWorkerID(&job), ENUM_ATTR(job_type, caption, job.job_type));
+
+            // The key here is to strategically check the most important bits to reduce church.
```
chdoc, line 559: "reduce church." :smiley:

library/modules/EventManager.cpp
```diff
@@ -527,12 +539,33 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
-            continue;
-        nowJobs.emplace(link->item->id, link->item);
+    multimap<Plugin*, EventHandler> copy(handlers[EventType::JOB_COMPLETED].begin(), handlers[EventType::JOB_COMPLETED].end());
+    std::vector<JobCompleteData> nowJobs;
+    // predict the size in advance, this will prevent or reduce rehashing elements.
+    nowJobs.reserve(prevJobs.size());
+    for (const auto jobPtr : df::global::world->jobs.list) {
+        auto& job = *jobPtr;
+        // Can we pack these into a single integer?
+        nowJobs.emplace_back(job.id, job.completion_timer, (bool)job.flags.bits.repeat);
```
chdoc, line 549: Maybe add a constructor that takes the job as argument? In fact, I think you only one constructor either way.

library/modules/EventManager.cpp
```diff
@@ -597,57 +630,71 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
+            for (auto& seenJob : seenJobs) {
+                if (!std::binary_search(nowJobs.begin(), nowJobs.end(), seenJob.first, CompareJobId{})) {
+                    jobIdsGone.push_back(seenJob.first);
+                }
+            }
+            for (auto jobId : jobIdsGone) {
+                //out.print("SEEN0: %d, %ld, %ld, %d, %ld, %d %s\n", jobId, seenJobs[jobId]->items.size(), seenJobs[jobId]->job_items.elements.size(), seenJobs[jobId]->flags.bits.working, seenJobs[jobId]->general_refs.size(), getWorkerID(seenJobs[jobId]), ENUM_ATTR(job_type, caption, seenJobs[jobId]->job_type));
+                Job::deleteJobStruct(seenJobs[jobId], true);
+                seenJobs.erase(jobId);
+            }
```
chdoc, line 694: I would be surprised if this is faster than iterating over `nowJobs` and using `seenJob.find` That's O(n) vs O(n log n). Maybe this is where your comparison time is being spent.

chdoc: My understanding is that we never call any handlers, unless tick1 > tick0. If I understand correctly, registering jobs that are seen while tick0=tick1, only adds detection of jobs that were queued (or modified) while the game was paused, having a `completion_timer` of 0, and that are already gone by the next job handler cycle. I'm having a hard time coming up with scenarios other than loading a save game, where this happens. Maybe initialization after loading can be handled differently. @myk002 What's your view on this?

library/modules/EventManager.cpp
chdoc: Either way, this isn't really a priority. There are bigger fish to fry in this PR.

dhthwy:
These three are the primary hotspots, and got it to include source too!

hotspots1

hotspots2

hotspots3

dhthwy:
When I added a constructor with an initializer list to JobCompleteData, the object lost its POD status and its trivially constructable status, so we're seeing the cost of the added expense of construction being paid there. There is no reason, as it is, for it to not be a POD.

myk002:
> zscreen    42201 ms ( 51.46% of dfhack,   7.30% of elapsed)

What ZScreen (that is, non-overlay windowed tool) do you have active when running these tests? I'm surprised to see this > 0 in these trials.

library/modules/EventManager.cpp
myk002: > My understanding is that we never call any handlers, unless tick1 > tick0

I think this is true, as is the rest of your analysis.

I believe the only other case of jobs appearing between ticks is when DFHack adds them, and we're not likely to delete those same jobs before the next tick.

IOW, I don't think we need sub-tick resolution for job detection.

myk002:
ah ok. as long as that's consistent across tests, that should be ok. I'm kind of surprised that just rendering the launcher takes that much time. something else to look into at some point (obviously, the job handling is far more important)

ab9rf:
Profiling on my windows 10 system with intel vTune identified `EventManager.cpp` line 564 `|| seenJob.general_refs.size() != job.general_refs.size()) {` as the top hotspot. In a 60s (clock time) sample, this line accounted for 273.970 ms (out of 51.364 s total CPU), with 172.481 ms of that being the `mov rax, qword ptr [rsp+0xb8]` at the top of the basic block. I'm pretty sure that `[rsp+0xb8]` is `seenJob`.

No other hotspot is worth mentioning relative to this one: the next highest hotspot in `EventManager.cpp` is line 649, at 10.299ms.

dhthwy:
I don't think there's anything we can do about that unless the job completion handlers don't need that updated information.

ab9rf:
I updated my installation of vTune to 2024.2 and reran a sample in microarchitecture analysis mode; the clear indication is that almost all the stalls are due to waiting on main memory (that is, an L3 cache miss). This is likely because the job objects that are stored in `seenJobs` are naively heap-allocated, so scanning through them blitzes the cache pretty effectively.

library/modules/EventManager.cpp
```diff
@@ -527,12 +544,31 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
+            //out.print("SEEN0: %d, %ld, %ld, %d, %ld, %d %s\n", seenJob.id, seenJob.items.size(), seenJob.job_items.elements.size(), seenJob.flags.bits.working, seenJob.general_refs.size(), getWorkerID(&seenJob), ENUM_ATTR(job_type, caption, seenJob.job_type));
+            //out.print("SEEN1: %d, %ld, %ld, %d, %ld, %d %s\n", job.id, job.items.size(), job.job_items.elements.size(), job.flags.bits.working, job.general_refs.size(), getWorkerID(&job), ENUM_ATTR(job_type, caption, job.job_type));
+
+            // The key here is to strategically check the most important bits to reduce churn.
+            if (seenJob.flags.whole != job.flags.whole
+                    || seenJob.items.size() != job.items.size()
+                    || seenJob.general_refs.size() != job.general_refs.size()) {
+                // Can't use emplace() for preexisting keys.
+                seenJobs[job.id] = JobUniquePtr(Job::cloneJobStruct(&job, true));
+            }
```
dhthwy, line 567: ruh roh. I think that is UB.

seenJob is reference to pointer, but that pointer may be smashed by the assignment from seenJobs[job.id] = JobUniquePtr();  

No warnings, of course.

library/modules/EventManager.cpp
chdoc: I think we need to distinguish between ticks and cycles of the event handler. That being said, if someone registers a handler for `jobCompleted` with a frequency of 150, they should already not rely on being notified for every job. 

As far as I understand, the difference between updating `seenJobs` or not when tick0=tick1is only relevant when there is a registered handler with frequency 0, otherwise the function is not going to be called anyway (except possibly during loading/initialization, not sure). This is ensured by the following guard in `manageEvents`:
```c++
 if ( tick >= eventLastTick[a] && tick - eventLastTick[a] < eventFrequency )
            continue;
```
Without changing documented behavior, I think we actually need option (2). I'm not sure whether there are any tools inside DFHack that actually register a job completed handle with frequency 0. 

I don't know what is calling `manageEvents`, so I don't know whether it is possible that it runs more than once per tick while the game is unpaused. If this can only happen while the game as paused, I suppose we can relax, because this will not impact (simulation) framerate.

dhthwy:
Probably need a custom allocator for this one. As far as the jobs themselves, the job's don't change *that much*, so the re-cloning shouldn't be hurting, but again, maybe needs a custom allocator.

ab9rf:
To be honest I'm not sure it's worth it. 321.865ms out of 51.364s total CPU time is 0.63%. This is already a major improvement over the existing code. Let's not let the perfect become the enemy of the good here.

dhthwy:
I agree. We may be able to get some more gains, but that can be worked on in an enhancement PR. If this one stalls, then the one in current use is a super 10x staller.

dhthwy:
I tried moving the job linked list traversal up the call chain.

I pushed the jobs into a vector with an object that wraps the job pointer and some hot members frequently accessed (partitioning hot and cold data). This should help the caches...

But, simply traversing the linked list is 0.30%
It helped the started event manager a lot, now using the vectored job list, going from 0.20% to 0.05%
It didn't help the job completion manager much. The seenJobs hashmap, despite using the same partitioned object, is still slow. If I take out the bit that updates the job when something changes, that only knocks off 0.05%.

The result is: no material improvement.

dhthwy:
I'm not sure about that last commit. It seems right to put it there so it could benefit other parts of dfhack if needed. the keptEverything flag should be a NOOP if it was cloned with keepEverything = false and then deleted with keptEverything = true since the vectors will be empty.  Unless I'm having a dumb moment.

dhthwy:
Well doh. Given that completion_timer is checked and there are many that have value -1, we don't need to add anything to the hashmap that is -1.

dhthwy:
Alright, assuming I am not having a major brainfart attack. I believe the cache miss issues have been solved, Jim.

dhthwy:
Using myk's save for benchmarks.

After learning more about the perf tool. I was able to get it to give me #s regarding cache misses, stalls, and such. It appears the hash lookup in jobStartedEvent manager is killing the cache.  Since the data set is small and simple, and we only need to perform a lookup for an item for only a subset in the set (job has a worker), we can simply use binary search and utilize the CPU instead as the cache misses are costing us cycles anyway. This should be faster in most situations unless they have like 10000 jobs. This change further reduced event manager from roughly 6100ms to 4900ms

```
> :lua require('script-manager').print_timers()

Summary
-------

Measuring unpaused time only

elapsed   582997 ms (9m 42s)
 dfhack    66759 ms ( 11.45% of elapsed)


DFHack details
--------------

keybinding        1 ms (  0.00% of dfhack,   0.00% of elapsed)
    update    10257 ms ( 15.36% of dfhack,   1.76% of elapsed)
   overlay    20622 ms ( 30.89% of dfhack,   3.54% of elapsed)
   zscreen    35879 ms ( 53.74% of dfhack,   6.15% of elapsed)


Update details
--------------

  event manager     4936 ms ( 48.12% of update,   0.85% of elapsed)
plugin onUpdate     4417 ms ( 43.06% of update,   0.76% of elapsed)
     lua timers      815 ms (  7.95% of update,   0.14% of elapsed)


Event manager per event type
----------------------------

            JOB_COMPLETED     2994 ms ( 60.66% of event manager,   0.51% of elapsed)
              JOB_STARTED     1491 ms ( 30.21% of event manager,   0.26% of elapsed)
            JOB_INITIATED      241 ms (  4.88% of event manager,   0.04% of elapsed)
                 BUILDING       56 ms (  1.13% of event manager,   0.01% of elapsed)
          UNIT_NEW_ACTIVE       50 ms (  1.01% of event manager,   0.01% of elapsed)
                   UNLOAD        8 ms (  0.16% of event manager,   0.00% of elapsed)

                framework       96 ms (  1.94% of event manager,   0.02% of elapsed)
            all subtimers     4840 ms ( 98.06% of event manager,   0.83% of elapsed)
```

library/modules/EventManager.cpp
```diff
@@ -491,31 +497,39 @@ static void manageJobStartedEvent(color_ostream& out) {
...
-                DEBUG(log,out).print("calling handler for job started event\n");
-                run_handler(out, EventType::JOB_STARTED, handle, job);
+    std::vector<int32_t> newStartedJobs;
+    newStartedJobs.reserve(startedJobs.size());
+
+    for (const auto jobPtr : df::global::world->jobs.list) {
+        // posting_index of -1 implies a worker has been assigned to a new job.
+        if (jobPtr->posting_index == -1) {
+            auto jobId = jobPtr->id;
+            newStartedJobs.push_back(jobId);
```
myk002, line 507: are jobs in the list guaranteed to be ordered by id?

my testing shows that they appear to be ordered by id, even with repeat jobs, but I don't know if this is a guarantee. @ab9rf do you know?

library/modules/EventManager.cpp
```diff
@@ -491,31 +497,39 @@ static void manageJobStartedEvent(color_ostream& out) {
...
-//static int32_t getWorkerID(df::job* job) {
-//    auto ref = findRef(job->general_refs, general_ref_type::UNIT_WORKER);
-//    return ref ? ref->getID() : -1;
-//}
+#if 0
+static int32_t getWorkerID(df::job* job) {
+    auto ref = findRef(job->general_refs, general_ref_type::UNIT_WORKER);
+    return ref ? ref->getID() : -1;
+}
+#endif
```
myk002, line 532: can we just get rid of this?

library/modules/EventManager.cpp
```diff
@@ -597,57 +634,77 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
-            //still false positive if cancelled at EXACTLY the right time, but experiments show this doesn't happen
-            for (auto &[_,handle] : copy) {
-                DEBUG(log,out).print("calling handler for repeated job completed event\n");
-                run_handler(out, EventType::JOB_COMPLETED, handle, (void*) &job0);
+    /*
+     * if it happened within a tick, must have been cancelled by the user or a plugin: not completed
+     * note observation: delta 0 tick events happens normally on load, save, paused (whether by user or game).
+     * Does it make sense to check this here? This should be handled by manageEvents().
+     */
+    if (tick1 >= tick0) {
```
myk002, line 642: shouldn't this be `tick1 > tick0`? but you're also right that this should already be gated by the check in `manageEvents`: `if ( tick >= eventLastTick[a] && tick - eventLastTick[a] < eventFrequency )`

so this check can probably be removed

library/modules/EventManager.cpp
ab9rf: the game doesn't intentionally sort the list but the manner in which jobs are created by the game does (currently) guarantees that they will be in order

library/modules/EventManager.cpp
dhthwy: I'd rather remove it. I changed it to tick1 >= tick0 so that it'd work for 0 tick resolution.

library/modules/EventManager.cpp
dhthwy: Yes, it was already there, and used it myself, but there's other functions to use for that. Not needed.

library/modules/EventManager.cpp
dhthwy: that implies he doesn't perform any binary searches on the job list.  We can always sort the vector ourselves, if we want to be sure.

ab9rf: > that implies he doesn't perform any binary searches on the job list. We can always sort the vector ourselves, if we want to be sure.

if he does i've never found evidence of it. i've also never seen any code for inserting elements anywhere other than the end of the list

myk002: I'm not aware of any use case where he'd need to binary search the list. jobs are normally referenced as `df::job *`, not by id

library/modules/EventManager.cpp
```diff
@@ -566,12 +555,40 @@ static void manageJobCompletedEvent(color_ostream& out) {
...
+         * In this case, Job with ID 1 has started after jobs 2, 4, 8, and 10.
+         * But, nowJobs is not greater than or equal to prevJobs because the ID
+         * 1 in nowJobs is less than the smallest ID in prevJobs,
+         * which breaks the algorithm.
          */
         nowJobs.emplace_back(job.id, job.completion_timer, (bool)job.flags.bits.repeat);
     }
 
+    // Do we want this check?
+    //assert(std::is_sorted(nowJobs.begin(), nowJobs.end()));
```
myk002, line 587: I like the idea of this being in a (Lua) unit test that will run when in fortress mode. The check can be in cpp, but callable from Lua via something like `dfhack.internal.checkInvariants()`

myk002 (approved):
unit test can be done in a separate PR. This one looks good to merge so we can start some practical testing

PR #4797, opened 2024-07-07
Add simulate support for mouse middle btn and "*DOWN" mouse btn events

wiktor-obrebski:
Add support for `_MOUSE_L_DOWN`, `_MOUSE_R_DOWN`, `_MOUSE_M_DOWN` and `_MOUSE_M` to `simulateInput` function.

Withot it, it is impossible to test some of gui scripts (e.g. journal) in integration way, as only `_MOUSE_L` and `_MOUSE_R` inputs are working with `gui.simulateInput` right now.

The same table (`MOUSE_KEYS`) is also used on `ZScreen:onInput`. I am not sure if this can affect it negatively somehow.

### DIFF
--- a/docs/about/Authors.rst
+++ b/docs/about/Authors.rst
@@ -171,6 +171,7 @@ PopnROFL                PopnROFL
 potato
 ppaawwll                ppaawwll                🐇🐇🐇🐇
 Priit Laes              plaes
+psychowico              wiktor-obrebski
 Putnam                  Putnam3145
 quarque2                quarque2
 Quietust                quietust                _Q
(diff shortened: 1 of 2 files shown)


myk002:
I forgot to mention this before: please feel free to add yourself to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

myk002 (commented):
I looked through the history to see why I implemented it this way. I found https://github.com/DFHack/dfhack/pull/3841#discussion_r1342308980

```
We don't touch mbut in general so that map scrolling continues to work. The _down fields should not be modified since it can break dragging. Widgets capture the mouse on button click and then react to _down if they handled the click. if we mess with _down, then we can break whatever thinks the mouse belongs to them (including our own window dragging mechanism).
```

Could you test to see if this breaks middle button map dragging when a ZScreen is visible, as well as general window dragging/resizing?

wiktor-obrebski:
> Could you test to see if this breaks middle button map dragging when a ZScreen is visible, as well as general window dragging/resizing?

I tested dragging and resizing, also background map movement (by mouse) for few zscreen tools, including `gui/layout`, `gui/journal`, `open-legens` and `gui/quickfort`. I did not detect any unexpected behaviour.

PR #4812, opened 2024-07-16
test and fix DFLinkedList editing

chdoc:
I wasn't entirely sure about `link->prev = root` in `push_front`, but looking at the job list using `gm-editor`, it appears that the first item does indeed point back to the root.

### DIFF
--- a/library/include/BitArray.h
+++ b/library/include/BitArray.h
@@ -400,8 +400,10 @@ namespace DFHack
         size_t size() const
         {
             size_t n = 0;
-            for (value_type const & i : *this)
-                n++;
+            for (auto It = this->cbegin(); It != this->cend(); ++It)
+            {
+               n++;
+            }
             return n;
         }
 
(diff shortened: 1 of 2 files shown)


library/include/BitArray.h
```diff
@@ -528,7 +530,7 @@ namespace DFHack
         {
             auto root = static_cast<L *>(this);
             auto link = new L();
-            link->prev = nullptr;
+            link->prev = root;
             if (root->next)
             {
                 root->next->prev = link;
```
quietust, line 536: Remark: Toady's own `tlist` class _always_ sets `link->next = root->next;`, though if `new L();` initializes both pointers to NULL it should be logically equivalent.

chdoc: I only looked at the constructor for `job_list_link`, and that does indeed null all components. Other than that, I'm not sure what you mean. This is about the question whether the `prev` pointer of the first (real) list element points back to the static `root` node. From my observations, this seems to be the case when DF adds the first element.

quietust: I wasn't commenting on your fix, just observing another minor difference (which appears to be purely cosmetic) between DFHack's code and Toady's code.

myk002 (approved):

Issue #4826, opened 2024-08-01
Crash occurs when using gui/mass-remove on large sections of furniture


gadict:
https://file.io/UsgBtGwhvJYD is a link to the last two saves that were made prior to the crash being reproduced.

F4-F6 are the hotkeys for the areas where most of the furniture in the fort is located.
Steps to reproduce:
- Use gui/mass-remove to designate a large section of furniture for removal.
- Allow dwarves to pick up those tasks.
- Designate another large section for removal.

It seemed to happen more easily if I selected furniture while unpaused.

I had trouble in the smaller world getting the bug to reproduce, but I was able to force it by selecting the same area for removal multiple times. 

I have no mods installed other than DFHack and have not modified any raw files myself. 

Thanks again for the assistance!

myk002:
discussion on Discord here: https://discord.com/channels/793331351645323264/807444515194798090/1268742053701881936

Was able to reproduce on Linux after a few attempts

PR #4828, opened 2024-08-02
[Buildings] clean up deconstruct

myk002:
as per discord discussion: https://discord.com/channels/793331351645323264/807444515194798090/1268297002676912190

Fixes: https://github.com/DFHack/dfhack/issues/4826

### DIFF
--- a/library/modules/Buildings.cpp
+++ b/library/modules/Buildings.cpp
@@ -1355,6 +1355,7 @@ bool Buildings::deconstruct(df::building *bld)
     bld->removeUses(false, false);
     // Assume: not unit destroy target
     int id = bld->id;
+    vector_erase_at(plotinfo->tax_collection.rooms, linear_index(plotinfo->tax_collection.rooms, id));
     // Assume: not used in punishment
     // Assume: not used in non-own jobs
     // Assume: does not affect pathfinding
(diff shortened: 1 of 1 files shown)


library/modules/Buildings.cpp
```diff
@@ -1376,61 +1307,130 @@ static void on_civzone_delete(df::building_civzonest* civzone)
...
+        btype == df::building_type::Support ||
+        btype == df::building_type::Chain ||
+        btype == df::building_type::Cage ||
+        btype == df::building_type::Weapon ||
+        btype == df::building_type::Hatch ||
+        btype == df::building_type::GrateWall ||
+        btype == df::building_type::GrateFloor ||
+        btype == df::building_type::BarsVertical ||
+        btype == df::building_type::BarsFloor ||
+        btype == df::building_type::GearAssembly;
```
Bumber64, line 1380: `df::building_type::Trap` with `df::trap_type::Lever`, `df::trap_type::PressurePlate`, or `df::trap_type::TrackStop`.

myk002: yeah, I figured just checking for `df::building_type::Trap` would be sufficient here. I don't foresee any false positives or false negatives with the simpler check

PR #4837, opened 2024-08-04
Minor Items module improvements

Bumber64:
Minor improvements to `Items` module code. I tried to separate some of `Items.cpp`'s changes into multiple commits, so hopefully the diffs are more readable.

Improved some of the documentation in `Items.h`, as wells as `Lua API.rst` (where two fns were undocumented).

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -1956,27 +1956,29 @@ Military module
 Items module
 ------------
 
-* ``dfhack.items.getPosition(item)``
+* ``dfhack.items.findType(string)``
 
-  Returns true *x,y,z* of the item, or *nil* if invalid; may be not equal to item.pos
-  if in inventory.
+  Finds an item type by string and returns the ``df.item_type``. String is
+  case-sensitive (e.g., "TOOL").
 
-* ``dfhack.items.getBookTitle(item)``
+* ``dfhack.items.findSubtype(string)``
 
-  Returns the title of the "book" item, or an empty string if the item isn't a "book"
-  or it doesn't have a title. A "book" is a codex or a tool item that has page or
-  writings improvements, such as scrolls and quires.
+  Finds an item subtype by string and returns the subtype or *-1*. String is
+  case-sensitive (e.g., "TOOL:ITEM_TOOL_HIVE").
 
-* ``dfhack.items.getDescription(item, type[, decorate])``
+* ``dfhack.items.isCasteMaterial(item_type)``
 
-  Returns the string description of the item, as produced by the ``getItemDescription``
-  method. If decorate is true, also adds markings for quality and improvements.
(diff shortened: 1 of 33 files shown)


myk002:
CI errors should be fixed now

library/modules/Items.cpp
```diff
@@ -866,38 +703,35 @@ static string get_item_type_str(df::item *item) {
...
 
 static string get_base_desc(df::item *item) {
-    if (auto name = Items::getBookTitle(item); name.size())
+    if (auto name = Items::getBookTitle(item); !name.empty())
         return name;
-    if (auto artifact = get_artifact(item)) {
-        if (artifact->name.has_name)
-            return Translation::TranslateName(&artifact->name, false) + ", " + Translation::TranslateName(&artifact->name) + " (" + get_item_type_str(item) + ")";
-    }
+    if (auto artifact = get_artifact(item); artifact->name.has_name)
```
myk002, line 708: this will crash if `get_artifact(item)` returns NULL. to combine the two if statements, it would have to be
```cpp
if (auto artifact = get_artifact(item); artifact && artifact->name.has_name)
```

library/modules/Items.cpp
```diff
@@ -1111,12 +940,10 @@ bool DFHack::Items::moveToBuilding(df::item *item, df::building_actual *building
 
     auto ref = df::allocate<df::general_ref_building_holderst>();
     if(!ref) {
-        delete ref;
```
myk002, line 1114: ha

library/modules/Items.cpp
```diff
@@ -1241,250 +1064,183 @@ df::proj_itemst *Items::makeProjectile(df::item *item) {
...
+            if (auto craw = Units::getCasteRaw(mat_type, mat_subtype))
+                value = craw->misc.petvalue;
+            return value;
+        case FOOD:
+            return 10;
+        case CORPSE:
+        case CORPSEPIECE:
+        case REMAINS:
+        case ROCK:
+        default:
```
myk002, line 1242: can we remove the `default` case so the compiler warns us if there is an unhandled enum value?

myk002 (commented):
this was a large CL, so I worry I missed something. this is still early in the release cycle, so if there are issues hidden in here, they have time to surface

docs/dev/Lua API.rst
```diff
@@ -2131,7 +2133,7 @@ Items module
 
 * ``dfhack.items.canMelt(item[,game_ui])``
 
-  Returns true if the item can be used for melting. Unless ``game_ui`` is
+  Returns true if the item can be melted (at a forge). Unless ``game_ui`` is
```
myk002, line 2136:
```suggestion
  Returns true if the item can be melted (at a smelter). Unless ``game_ui`` is
```

myk002 (approved):

PR #4857, opened 2024-08-12
[orders] compact and properly hide workers tab overlay

chdoc:


myk002:
in service of https://github.com/DFHack/scripts/pull/1257

plugins/lua/orders.lua
```diff
@@ -280,7 +280,7 @@ local function can_set_skill_level()
             return false
         end
     end
-    return ret
+    return ret and not df.global.game.main_interface.stockpile_link.adding_new_link
```
myk002, line 283: I think this should instead be fixed in the focus string, since the focus string is not accurately representing the UI state. I'll see if I can do that.

myk002: fixed in https://github.com/DFHack/dfhack/pull/4858 -- you should be able to remove this check now

myk002 (approved):

PR #4866, opened 2024-08-17
use consolidated changelog for pre-v50 History

ChrisJohnsen:
While trying out local builds, I noticed that some pre-v50 entries kept showing up the built NEWS files. It looks like they are coming from the scripts and library/xml changelogs (I made a couple other PRs to trim those changelogs down to just v50, too).

But, it looks like the entries from those changelogs are not present in the newly migrated v0.47/v0.44 History (#4856).

This is a draft PR since I'm not sure what the proper procedure is for History.rst, but this commit seemed like I reasonable approach.

This PR re-builds the v0.47/v0.44 entries from (a pre-migration, locally-built) docs/changelogs/news-dev.rst. The format looks close to the older entries, but there are definite differences from the originally migrated v0.47/v0.44 History (besides the entries from the scripts and library/xml changelogs):

- `DFHack ` prefix on version names
  - this matches the older entries in History
- section order is different (order fixed in changelog.py)
- entry order is different (the "features" are sorted in the consolidated news)
- more use of the second-level list entries (though at least one single-second-level-entry was squashed back down to a single-first-level-entry)
- `search` "feature" is renamed to `search-plugin`
  - I didn't research the reason for this. Is this okay for these historical entries?
- `:\`  (used to embed a colon in a feature name?) is translated to `:`
- `@` dev-only flag is removed
- no blank lines between entries

This is hard a hard diff to review. `git diff --color-moved` helped filter out some of the reordering, but the stuff that is moved in/out of second-level lists still causes lots of noise.

### DIFF
--- a/docs/about/History.rst
+++ b/docs/about/History.rst
@@ -15,125 +15,171 @@ generation system <build-changelog>` to handle.
   :local:
   :depth: 1
 
-0.47.05-r8
-==========
+DFHack 0.47.05-r8
+=================
 
 New Plugins
 -----------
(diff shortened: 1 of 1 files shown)


myk002 (approved):
thank you for cleaning this up!

PR #4871, opened 2024-08-25
`sort` - expands places stockpile search filters to include storable item types

jfisher446:
Search results on the Places>Stockpile tab now filter based on: the stockpile's default name & number, customized name, and the storable-category flags that are enabled for the stockpile. 

This fix stemmed from opening the stockpile list, seeing `Armor Stockpile #9`, then having no search results for `Armor`.


jfisher446:
New Behavior
Finds in custom name & default name (default name accomplished via flag checking)
image

Finds stockpile number in default name:
image

Finds hidden stockpile number when custom named:
image

Finds by flag and name:
image

plugins/lua/sort/places.lua
```diff
@@ -112,8 +112,36 @@ local function get_location_search_key(zone)
 end
 
 local function get_stockpile_search_key(stockpile)
-    if #stockpile.name ~= 0 then return stockpile.name
-    else return ('Stockpile #%s'):format(stockpile.stockpile_number) end
+    local result = {}
+    if #stockpile.name ~= 0 then table.insert(result, stockpile.name) end
+
+    local flags = stockpile.settings.flags
+    local flag_names = {
```
myk002, line 119: could you move this table outside of the function? since it's static, there's no reason to reinitialize it every time the function is called.

jfisher446: Great point. Fixed it.

plugins/lua/sort/places.lua
```diff
@@ -200,4 +228,4 @@ function PlacesOverlay:onRenderBody(dc)
     end
 end
 
-return _ENV
+return _ENV
```
myk002, line 231:
```suggestion
return _ENV

```

PR #4872, opened 2024-08-25
fix type confusion when using quick filter slot '0'

chdoc:


plugins/lua/buildingplan/planneroverlay.lua
```diff
@@ -462,12 +462,31 @@ end
...
 -- label: string representation of the filter
 -- mats: list of material names allowed by the filter
 BUILDINGPLAN_FILTERS_KEY = "buildingplan/quick-filters"
 
+-- old saves may use numbers as keys, which we convert to string keys on load
+dfhack.onStateChange[BUILDINGPLAN_FILTERS_KEY] = function(sc)
+    if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
+        return
+    end
+    print('quick filter: checking/updating configuration')
```
myk002, line 476: I don't think the player needs to see these status messages. This can/should be an invisible fix.

myk002 (approved):

PR #4875, opened 2024-08-26
simulateInput mishandles _STRING

ChrisJohnsen:
Note: This is a draft PR due to the second commit probably being unnecessary (and lacking docs).

# Problem

simulateInput is documented as accepting (among other kinds of values) the same
table that is provided to onInput. However, simulateInput seems to mishandle the
`_STRING` field of such tables.

The Git history looks like it shows that simulateInput has had this behavior
since 2012 August (its origin?):
* 38a07a4ca584e2cccc2c2814c35266ba33d692c8 for `_STRING` given to doInput, and
* 601a3a7927d50500b3dbdd2d20c5d7740a5484a1 for simulateInput.

This sure seems like a bug, but its age gives me doubts.

I didn't find any (open or closed) issues on GitHub that looked like they were
related to this, either.

# Proposed Solution

The first commit in this PR changes simulateInput to ignore `_STRING` of any table
arguments it receives. `_STRING` seems to be (informationally) redundant (being
derived from the presence/absence of a `STRING_A…` keycode), so it seemed
reasonable to ignore it.

The second commit offers a possible Lua-side "normalization" helper if such a thing
seems like a good idea.

# Alternate Solutions

Instead of ignoring `_STRING` (the commit with this PR), it could be converted
to a `STRING_A…` keycode (although this risks producing multiple (distinct)
"text" keycodes if the caller made "unaligned" changes to either `_STRING` or
`STRING_A…`). Or simulateInput could integrate something like what
DFHack::Screen::normalize_text_keys does on the C++ side (this gets a bit
complicated due to needing to aggregate effects over multiple arguments).

# Impact

The biggest impact I found is through ZScreen, but I also found:
- scripts/confirm.lua ConfirmOverlay does onInput->simulateInput, but only for a
  very restricted range of inputs that won't *usually* have `_STRING`
- scripts/gui/sandbox.lua Screen (a widgets.Window) using its ScreenView (a
  gui.ZScreen) to forward unhandled input
- several unavailable-tagged scripts that I didn't examine closely

The ZScreen onInput->simulateInput code path happens for defocused ZScreens and
focused ZScreens that have pass_movement_keys=true (and pass_mouse_clicks, but I
haven't investigated that one fully; I have seen single input events with
`_STRING` and mouse fields though).

## demonstration setup

My demonstration setup looks like this:
- a normal fortress game is loaded and running (i.e. ready for designations)
  - (probably paused to avoid the visual noise of other stuff changing on the
    screen)
[...]

PR #4876, opened 2024-08-26
MOVEMENT_KEYS missing Z and diagonal kbd cursor entries

ChrisJohnsen:
I was investigating how the MOVEMENT_KEYS table is used when I found the bug described in #4875.

I have a WIP project that initially used these, but I ended up changing approaches to solve a more general problem.

So I don't know of anything that *needs* these, but these keys are not currently being forwarded by pass_movement_keys ZScreens (since it ends up checking MOVEMENT_KEYS to decide if a key-set should be forwarded).


myk002 (approved):
I *think* this should be ok. this will mean the numpad keys get passed through, which *should* be ok, but keep an eye out for any abnormal behavior in other tools

PR #4879, opened 2024-08-28
Items::createItem - auto-select growth prints

quietust:
instead of requiring the caller to figure it out on their own.

I did some basic tests with oak leaves (to make sure they got the correct colors based on the time of year), but further tests would be welcome.

Conveniently, none of the Lua scripts had to be touched (because they weren't specifying growth print information in the first place).

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -74,12 +74,14 @@ Template for new versions:
 
 ## API
 - ``Units``: new ``isWildlife`` and ``isAgitated`` property checks
+- ``Items::createItem``: removed growth_print parameter, now determined automatically
 
 ## Lua
 - ``dfhack.units``: ``isWildlife`` and ``isAgitated`` property checks
 - ``dfhack.units.isDanger``: no longer returns true for intelligent undead
 - Overlay widgets can now assume their ``active`` and ``visible`` functions will only execute in a context that matches their ``viewscreens`` associations
 - ``gui.simulateInput``: do not generate spurious keycode from ``_STRING``
+- ``dfhack.items.createItem``: removed growth_print parameter to match C++ API
 
 ## Removed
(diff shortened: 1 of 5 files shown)


library/modules/Items.cpp
```diff
@@ -1756,7 +1759,34 @@ bool Items::createItem(vector<df::item *> &out_items, df::unit *unit, df::item_t
     for (auto out_item : out_items)
     {   // Plant growths need a valid "growth print", otherwise they behave oddly
         if (auto growth = virtual_cast<df::item_plant_growthst>(out_item))
+        {
+            int growth_print = -1;
+            // Make sure it's made of a valid plant material, then grab its definition
+            if (growth->mat_type >= 419 && growth->mat_type <= 618 && growth->mat_index >= 0 && (unsigned)growth->mat_index < world->raws.plants.all.size())
```
myk002, line 1765: I am horrified that we don't have constants defined for these, but looking through the code, I see these numbers copied into many places.

myk002 (approved):

library/modules/Items.cpp
quietust: I think we _do_ have them in the Materials module (as `MaterialInfo::PLANT_BASE`), but I couldn't be bothered to look it up at the time. Besides, I'm eventually going to add all of the individual enum values to structures, so I'll clean them up here when that happens..

docs/changelog.txt
```diff
@@ -74,12 +74,14 @@ Template for new versions:
...
 ## API
 - ``Units``: new ``isWildlife`` and ``isAgitated`` property checks
+- ``Items::createItem``: removed growth_print parameter, now determined automatically
 
 ## Lua
 - ``dfhack.units``: ``isWildlife`` and ``isAgitated`` property checks
 - ``dfhack.units.isDanger``: no longer returns true for intelligent undead
 - Overlay widgets can now assume their ``active`` and ``visible`` functions will only execute in a context that matches their ``viewscreens`` associations
 - ``gui.simulateInput``: do not generate spurious keycode from ``_STRING``
+- ``dfhack.items.createItem``: removed growth_print parameter to match C++ API
```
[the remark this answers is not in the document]
myk002: fixed in https://github.com/DFHack/dfhack/pull/4937

Issue #4881, opened 2024-08-28
preserve-rooms v2


myk002:
- [x] implement serialization/persistence (needs to be done ASAP)
- [x] maybe move the role assignment panel to the unit selection screen so it's not always taking up screen space when you click on the zone. that panel can still show if a reservation has been set, though
- [x] evict current owner if they don't match the role assignment
- [x] disable the zone when reserved for a role and no current unit to indicate that it is properly reserved and to prevent it from being snagged by a rando
- [x] better experience for roles that promote into other roles
- [x] quickfort integration/update dreamfort to make use of it

PR #4883, opened 2024-08-29
persist-rooms: implement state persistence and loading

myk002:
and also don't do processing that has been configured to be disabled


myk002:
ref: https://github.com/DFHack/dfhack/issues/4881

PR #4891, opened 2024-08-30
Update regrass plugin; DFHack::cuboid fixes

Bumber64:
Fixed an issue where `cuboid::clampMap` would include map edge tiles for cuboids completely outside of map. Function now returns `cuboid` instead of `bool` to allow for method chaining. Call `cuboid::isValid()` on the result to determine success (e.g., `bool success = some_cuboid.clampMap().isValid();`).

`cuboid::clamp` and `cuboid::clampNew` will now always return invalid cuboids for invalid input, instead of putting the responsibility on the caller.

Updated `regrass` plugin to use `cuboid` class. Improved debug logging. Allowed muddy feature stone (i.e., slade) to be converted to grass, since this can occur in vanilla. Made `--force` no longer imply `--new`, since someone didn't want random grass in their muddy cavern level.

### DIFF
--- a/plugins/regrass.cpp
+++ b/plugins/regrass.cpp
@@ -365,21 +365,6 @@ int regrass_tile(color_ostream &out, const regrass_options &options, df::map_blo
         TRACE(log, out).print("Tiletype no change.\n");
         return 1;
     }
-    /*else if (mat == tiletype_material::STONE || // Note: DF doesn't seem to remove mud.
-        mat == tiletype_material::FEATURE ||
-        mat == tiletype_material::LAVA_STONE ||
-        mat == tiletype_material::MINERAL)
-    {   // Muddy stone.
-        for (auto blev : block->block_events)
-        {   // Remove mud spatter.
-            auto ms_ev = virtual_cast<df::block_square_event_material_spatterst>(blev);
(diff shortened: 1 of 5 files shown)


myk002:
be sure to include a changelog entry for user-visible changes

plugins/regrass.cpp
```diff
@@ -343,134 +361,81 @@ int regrass_tile(color_ostream &out, const regrass_options &options, df::map_blo
...
         mat == tiletype_material::PLANT ||
         mat == tiletype_material::TREE)
-    {   // Already appropriate tile
-        DEBUG(log, out).print("Tiletype no change.\n");
+    {   // Already appropriate tile.
+        TRACE(log, out).print("Tiletype no change.\n");
         return 1;
     }
-    /*else if (mat == tiletype_material::STONE || // DF doesn't seem to remove mud
+    /*else if (mat == tiletype_material::STONE || // Note: DF doesn't seem to remove mud.
```
myk002, line 368: is this dead code? can it be removed?

Bumber64: Assuming we don't want to make it an option.

myk002 (approved):

PR #4896, opened 2024-09-01
Add information about `gui` module requirement for overlay to load

wiktor-obrebski:
Without `gui` required, overlay module will not appear on the overlays list, which is confusing for the dev.

### DIFF
--- a/docs/dev/overlay-dev-guide.rst
+++ b/docs/dev/overlay-dev-guide.rst
@@ -263,6 +263,8 @@ the ``Overlays`` tab.
    (``--@ module = true``) and it does not have side effects when loaded as a
    module (i.e. you check ``dfhack_flags.module`` and return before executing
    any statements if the value is ``true``)
+#. you script import `gui` module, even if it is not use it directly:
+    ``local gui = require('gui')``
 #. your code does not have syntax errors -- run
    ``:lua ~reqscript('myscriptname')`` (if a script) or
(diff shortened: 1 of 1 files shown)


docs/dev/overlay-dev-guide.rst
```diff
@@ -263,6 +263,8 @@ the ``Overlays`` tab.
    (``--@ module = true``) and it does not have side effects when loaded as a
    module (i.e. you check ``dfhack_flags.module`` and return before executing
    any statements if the value is ``true``)
+#. you script import `gui` module, even if it is not use it directly:
+    ``local gui = require('gui')``
```
myk002, line 267: are you sure this is true? what situation did you see this with?

wiktor-obrebski: Without importing `gui` module I was not able to see the `notes` overlay in the control panel overflow list.
Can you try to recreate the issue with empty overlay class?

myk002: `idle-crafting`, for example, has an overlay widget that does not `require('gui')`

wiktor-obrebski: `idle-crafting` require `gui.widgets` which itself require `gui`. I will make some test to confirm if this is the case.

wiktor-obrebski: Can not repeat it, so I closing the PR.
Maybe there was some another factor that I just did not understand earlier.

PR #4897, opened 2024-09-01
[buildingplan] only consider building materials that can be accessed by citizens/residents

chdoc:


plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -45,13 +46,38 @@ struct BadFlags {
...
 
+// cache of walkability groups that have citizens in them
+static std::unordered_set<uint16_t> accessible_walkability_groups;
+
+void update_walkability_groups(){
+    static int32_t update_timestamp = -1;
+
+    // ensure that we update at most once per tick
+    auto frame_counter = df::global::world->frame_counter;
+    if (frame_counter == update_timestamp)
```
myk002, line 57: this will give incorrect results if a world is loaded, exited, and then a different world is loaded (since `frame_counter` is reset to 0 on world load). It would be safer if the `update_timestamp` were declared outside of this function, referenced as an `extern` in `buildingplan.h` (or `buildingplan.cpp`) and reset in `plugin_load_site_data` (where `cycle_timestamp` is reset)

plugins/buildingplan/buildingplan_cycle.cpp
```diff
@@ -45,13 +46,38 @@ struct BadFlags {
...
+        return;
+    else
+        update_timestamp = frame_counter;
+
+    accessible_walkability_groups.clear();
+    auto get_group = [](df::unit *unit){
```
myk002, line 63: I may be wrong, but doesn't this need to be `[&]` to refer to `accessible_walkability_groups` by reference instead of a copy?

plugins/buildingplan/buildingplan_cycle.cpp
chdoc: You mean when you load a world, never unpause (or at least not long enough to reach a cycle), and then load a different world, you have one tick where there is stale data? Nice catch, I certainly didn't test that.

plugins/buildingplan/buildingplan_cycle.cpp
chdoc: GCC started yelling at me when I was trying to capture a variable with static storage duration. This is not being captured by copy, it is not being captured at all. So, I think you are wrong ;)

chdoc:
I noticed that `scanAvailableItems` is not being called on every tick, even if the panel showing the currently available building materials is open. Where does this rate limiting come from, and are the various Lua functions that indirectly call this "synchronized" to do this on the same tick?

plugins/buildingplan/buildingplan_cycle.cpp
myk002: ok, that makes sense.

myk002:
I believe the rate limiting is coming from here:
```lua
function ItemLine:reset()
    self.desc = nil
    self.available = nil
end

...

function ItemLine:get_item_line_text()
    local idx = self.idx
    local filter = get_cur_filters()[idx]
    local quantity = get_quantity(filter, self.is_hollow_fn())

    local buildingplan = require('plugins.buildingplan')
    self.desc = self.desc or buildingplan.get_desc(filter)

    self.available = self.available or buildingplan.countAvailableItems(
        uibs.building_type, uibs.building_subtype, uibs.custom_type, idx - 1)
    if self.available >= quantity then
        self.note_pen = COLOR_GREEN
        self.note = (' %d available now'):format(self.available)
    elseif self.available >= 0 then
        self.note_pen = COLOR_BROWN
        self.note = (' Will link next (need to make %d)'):format(quantity - self.available)
    else
        self.note_pen = COLOR_BROWN
        self.note = (' Will link later (need to make %d)'):format(-self.available + quantity)
    end
    self.note = string.char(192) .. self.note -- character 192 is "└"

    return ('%d %s%s'):format(quantity, self.desc, quantity == 1 and '' or 's')
end
```

so `buildingplan.countAvailableItems()` is only being called when `self.available` is reset (which happens once a cycle or when other UI actions are happening)

myk002 (approved):

PR #4905, opened 2024-09-03
Add icon for coming `notes` feature

wiktor-obrebski:
Add icon for coming `notes` feature.
Needed for https://github.com/DFHack/scripts/pull/1287


data/art/note-green-pin.png
myk002, line 1: could you rename this file `green_pin_map.png`? we're trying to give assets intended for the map layer a `_map.png` suffix.

wiktor-obrebski: I can, but I have doubts about generality of the name.
`green_pin_map.png` suggest that other tool can use it on the map - but in reality other tools would probably need its own version with the pin placed in different relative position to the tile.

myk002: as long as it has the `_map` suffix, it's fine. `note_green_pin_map.png` would be ok too

wiktor-obrebski: done

myk002 (approved):

PR #4913, opened 2024-09-04
rename lua lib from lua to lua53

myk002:
so we don't conflict with other similarly named lua libs


ab9rf (approved):
this is the same change i made in my test builds that worked on windows, except i put it further down the file at line 106. should work regardless

PR #4942, opened 2024-09-16
clean up msvc warning suppression

ab9rf:
removed C4996 (deprecation) the only hit this had was in clsocket and the warning has been suppressed there until we fix it
removed C4800 (narrowing conversion of bool) no longer enabled by `/Wall` since VS2017 (MSVC v14.10)
removed C4341 (doesn't exist)
removed C4018 (signed/unsigned mismatch in compare) with violating code corrected 
removed C4482 (doesn't exist)
removed C4231 (doesn't exist)

for the four remaining, revised editorial comments


library/CMakeLists.txt
```diff
@@ -69,12 +69,10 @@ set(MAIN_HEADERS
     include/Memory.h
     include/MiscUtils.h
     include/Module.h
-    include/Pragma.h
```
myk002, line 72: oh good, I'm glad you got both of them

library/CMakeLists.txt
```diff
@@ -69,12 +69,10 @@ set(MAIN_HEADERS
...
     include/MemAccess.h
     include/ModuleFactory.h
     include/PluginManager.h
     include/PluginStatics.h
-    include/Pragma.h
```
myk002, line 77: oh good, I'm glad you got both of them

myk002 (approved):

library/CMakeLists.txt
ab9rf: i totally didn't notice this; i just had a search for "Pragma.h" in VS code and was just deleting things until there weren't any hits left :)

PR #4948, opened 2024-09-20
Raiding units aren't flagged inactive, update code and docs

Bumber64:
~~Add `Units::isInPlay` fn to check `isActive` then scan `units.active`. Useful when you've got a unit pointer, but you don't know if it's in the active units vector.~~ (DF unsets the `inactive` flag a few ticks after raiding units leave the map.)

Updated some code and docs using this info. I would've changed the `isActive` function, but there may be performance concerns. Would be great if Bay12 could just use a flag for this, but we can't count on raiding bugs being fixed this decade.

Made `getUnitsInBox` rule out inactive units by default, since death and map edge locations aren't where the unit actually is.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -36,7 +36,6 @@ Template for new versions:
 Document caveats about unit/item position fns w/r/t raiding units.
 
 ## API
-``Units::isInPlay``: A fn that scans active unit vector to rule out units off map raiding. Use instead of ``Units::isActive`` when uncertain if unit present in ``world.units.active``.
 
 ## Lua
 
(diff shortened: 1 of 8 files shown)


Bumber64:
Couldn't find any places in existing code that would benefit from a `Units::isInPlay`. `linear_index` can be used if someone needs to scan `world.units.all`.

Bumber64:
`hf.info.whereabouts.army_id` was mentioned as a way to get the army so we could check for raiding. Doing the hf lookup is probably still too much for how `isActive` is used.

Bumber64:
Replaced by #4959

PR #4955, opened 2024-09-22
Refactor/widgets

wiktor-obrebski:
Refactor massive `widgets.lua` into module per component style.

No API or logic has been changed (at least intentionally).

This is a preparation to add `TextArea` (`gui/journal` style input widget) and migrate `TextField` widget to journal style features.


library/lua/gui/widgets.lua
```diff
@@ -2,3338 +2,40 @@
...
+Tab = TabBar.Tab
+makeButtonLabelText = Label.makeButtonLabelText
+parse_label_text = Label.parse_label_text
+render_text = Label.render_text
+check_text_keys = Label.check_text_keys
+
+DOUBLE_CLICK_MS = Panel.DOUBLE_CLICK_MS
+STANDARDSCROLL = Scrollbar.STANDARDSCROLL
+SCROLL_INITIAL_DELAY_MS = Scrollbar.SCROLL_INITIAL_DELAY_MS
+SCROLL_DELAY_MS = Scrollbar.SCROLL_DELAY_MS
```
myk002, line 39: we'll probably need a function-based API for these config elements. As written, if you change these values via the control panel, the relevant widgets won't actually see the changed value and will not change their behavior. See `scripts/internal/control-panel/registry.lua` for how these values are interacted with.

We could change them to a function interface, like we did for `HIDE_ARMOK_TOOLS` in `dfhack.lua`. It is now wrapped with `getMortalMode()` and `setMortalMode()` functions. The code in the control panel registry would then have to be updated to call the new API instead of setting the values directly.

library/lua/gui/widgets/containers/tab_bar.lua
myk002, line 1: I'm not sure TabBar should be classified as a container. It's meant to be used alongside a container, like a Pages widget, but it doesn't contain other widgets itself.

library/lua/gui/widgets.lua
```diff
@@ -2,3338 +2,40 @@
...
+FilteredList = require('gui.widgets.filtered_list')
+TabBar = require('gui.widgets.containers.tab_bar')
+RangeSlider = require('gui.widgets.range_slider')
+DimensionsTooltip = require('gui.widgets.dimensions_tooltip')
+
+Tab = TabBar.Tab
+makeButtonLabelText = Label.makeButtonLabelText
+parse_label_text = Label.parse_label_text
+render_text = Label.render_text
+check_text_keys = Label.check_text_keys
```
myk002, line 34: are there any existing usages of these three functions (`parse_label_text`, `render_text`, and `check_text_keys`)? they aren't really intended for public use. they could probably stay as an internal API and not be exposed in `widgets`

wiktor-obrebski: you are right.
removed from the public API.

library/lua/gui/widgets/containers/tab_bar.lua
wiktor-obrebski: moved to root widgets dir

myk002 (approved):

Issue #4312, opened 2024-03-01
add distance and value info to buildingplan item selection list


myk002:
also make it sortable by those properties

ab9rf:
fixed in #4956

PR #4956, opened 2024-09-22
Add distance and value info to buildingplan item selection list #4312

frogi16:
- add option to sort by value
 - reorder sorting options to match ordering of colums
 - adjust width of existing columns

### DIFF
--- a/plugins/lua/buildingplan/itemselection.lua
+++ b/plugins/lua/buildingplan/itemselection.lua
@@ -60,6 +60,12 @@ local function sort_by_quantity(a, b)
             (ad.quantity == bd.quantity and sort_by_type(a, b))
 end
 
+local function sort_by_value(a, b)
+    local ad, bd = a.data, b.data
+    return ad.value > bd.value or
+            (ad.value == bd.value and sort_by_type(a, b))
+end
+
 ItemSelection = defclass(ItemSelection, widgets.Window)
 ItemSelection.ATTRS{
     frame_title='Choose items',
@@ -151,8 +157,9 @@ function ItemSelection:init()
                     label='Sort by:',
                     options={
                         {label='Recently used', value=sort_by_recency},
-                        {label='Name', value=sort_by_name},
                         {label='Amount', value=sort_by_quantity},
+                        {label='Value', value=sort_by_value},
+                        {label='Name', value=sort_by_name},
                     },
                     on_change=self:callback('on_sort'),
                 },
@@ -270,6 +277,7 @@ function ItemSelection:get_choices(sort_fn)
                     item_subtype=item:getSubtype(),
                     quantity=1,
                     quality=item:getQuality(),
+                    value=dfhack.items.getValue(item),
                     selected=0,
                 },
             }
@@ -281,7 +289,8 @@ function ItemSelection:get_choices(sort_fn)
     for desc,choice in pairs(buckets) do
         local data = choice.data
         choice.text = {
-            {width=10, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
+            {width=8, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
+            {width=8, gap=2, text=function() return ('%d\x0F'):format(data.value) end},
             {gap=2, text=desc},
         }
         table.insert(choices, choice)
--- a/plugins/lua/buildingplan/itemselection.lua
+++ b/plugins/lua/buildingplan/itemselection.lua
@@ -4,6 +4,7 @@ local gui = require('gui')
 local pens = require('plugins.buildingplan.pens')
 local utils = require('utils')
 local widgets = require('gui.widgets')
+local caravan = reqscript('internal/caravan/common')
 
 local uibs = df.global.buildreq
 local to_pen = dfhack.pen.parse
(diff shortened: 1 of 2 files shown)


frogi16:
Would solve issue #4312

frogi16:
FYI: @myk002, @phooenixdassmatte

phooenixdassmatte:
> FYI: @myk002, @phooenixdassmatte
> 
> EDIT: I just thought that some items might have the same name but differ in value because of values of material used to enhance the items (just compare sapphires to agats...). Tomorrow I will test what happens in such situation when my solution is applied, because it's possible the items are lumped together despite having different values.

Hey there! Looks good, and still gives more information than vanilla. 

I have no idea how df code works but an idea to include the encrusted object in the name? 
So: 
+*brass chain*+ enhanced with *polished topaz*

That way it would be a separate line of: 

+*brass chain*+ enhanced with polished topaz.

And would also give more information. We have big screens now so. :D

frogi16:
I managed to ungroup items with different values quite easily
image

@myk002 I would be grateful for code review.

myk002:
Ok, now that the craziness of the 50.14 release is behind us, I can catch up on code reviews. Thank you for your patience.

plugins/lua/buildingplan/itemselection.lua
myk002, line 1: Please add a line to docs/changelog.txt (under Future -> Misc improvements)

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -256,33 +263,38 @@ function ItemSelection:get_choices(sort_fn)
...
         ::continue::
     end
     local choices = {}
-    for desc,choice in pairs(buckets) do
+    for key,choice in pairs(buckets) do
         local data = choice.data
         choice.text = {
-            {width=10, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
-            {gap=2, text=desc},
+            {width=8, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
```
myk002, line 295: this was width=10 to accommodate item counts up to 9999. width=8 only accommodates up to a count of 999. This is probably fine, as such large numbers are an edge case, and the selected count will still be visible for four-digit amounts, even if the total count is pushed out of the field.

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -256,33 +263,38 @@ function ItemSelection:get_choices(sort_fn)
...
+            {width=8, gap=2, text=function() return ('%d\x0F'):format(data.value) end},
```
myk002, line 296: displaying the value directly is a spoiler. Please use the output of `reqscript('internal/caravan/common').obfuscate_value(value)`

myk002, line 296: other tools that display value use a field width of 9. I suggest you do the same here

myk002, line 296: please use a constant instead of a hex value here. There is one pre-defined at `reqscript('internal/caravan/common').CH_MONEY`

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -256,33 +263,38 @@ function ItemSelection:get_choices(sort_fn)
...
+            {gap=2, text=data.desc},
```
myk002, line 297: We might have to make the dialog a little wider to give more space for this string, but I'm not sure. In my testing in current forts, I haven't found any examples where we overflow the default space. maybe for divine metals. however, the dialog *is* resizable, so if the current dialog width handles the majority of item names, then it's probably fine to keep as is.

myk002 (commented):
Sorry the review took so long! Thank you for this addition!

plugins/lua/buildingplan/itemselection.lua
myk002: I agree. And for the purposes of this PR, I don't think the window size needs changing

frogi16:
@myk002  I made post-review changes, please review and merge if everything is fine. I tested it quickly and seems fine.

Force pull mentioned by Github is just me rebasing on top of new changes from official repo, I will try to do it cleaner next time with separate branch.

myk002:
yeah, try not to rebase while the review is active since it causes all the comments to get lost and I lose the ability to see what's changed since the last review. Merging upstream is fine, though!

plugins/lua/buildingplan/itemselection.lua
```diff
@@ -256,33 +264,39 @@ function ItemSelection:get_choices(sort_fn)
...
     local choices = {}
-    for desc,choice in pairs(buckets) do
+    for key,choice in pairs(buckets) do
         local data = choice.data
+        local obfuscated_value = caravan.obfuscate_value(data.value)
         choice.text = {
-            {width=10, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
-            {gap=2, text=desc},
+            {width=8, text=function() return ('%d/%d'):format(data.selected, data.quantity) end},
+            {width=9, gap=2, text=function() return ('%d%s'):format(obfuscated_value, caravan.CH_MONEY) end},
```
myk002, line 298:
```suggestion
            {width=9, gap=2, text=function() return ('%s%s'):format(obfuscated_value, caravan.CH_MONEY) end},
```

myk002: obfuscated value is a string, not a number

myk002:
when you have a minute, could you add your name to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst ?

PR #4959, opened 2024-09-24
Don't include raiding units in Units::isActive; make getPosition return invalid coords

Bumber64:
Raiding units are unflagged as `inactive` shortly after they leave the map, as they are removed from `units.active`. Scan this vector to rule out raiding units. The `move_state` and `can_swap` flags are also unset, which can be used to optimize the check.

`Units::getPosition` now returns invalid coords for inactive units. This impacts `Units::isUnitInBox`, and Units::getUnitsInBox`.

It also impacts `Items::getPosition`, though an item in `items.other.IN_PLAY` has its holder is guaranteed to be active (artifact theft bugs notwithstanding).

### DIFF
--- a/library/RemoteTools.cpp
+++ b/library/RemoteTools.cpp
@@ -584,7 +584,7 @@ static command_result ListUnits(color_ostream &stream,
 
     if (in->scan_all())
     {
-        auto &vec = df::unit::get_vector();
+        auto &vec = df::global::world->units.active;
 
         for (size_t i = 0; i < vec.size(); i++)
         {
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -122,7 +122,11 @@ constexpr uint32_t exclude_flags2 = (
 
 bool Units::isActive(df::unit *unit) {
     CHECK_NULL_POINTER(unit);
-    return !unit->flags1.bits.inactive;
+    if (unit->flags1.bits.inactive)
+        return false;
+    else if (unit->flags1.bits.move_state || unit->flags1.bits.can_swap)
+        return true; // These are always unset when leaving the map
+    return linear_index(world->units.active, &df::unit::id, unit->id) >= 0;
 }
 
 bool Units::isVisible(df::unit *unit) {
@@ -658,7 +662,7 @@ bool Units::isGreatDanger(df::unit *unit) {
 
 bool Units::isUnitInBox(df::unit *u, const cuboid &box) {
     CHECK_NULL_POINTER(u);
-    return box.containsPos(getPosition(u));
+    return isActive(u) ? box.containsPos(getPosition(u)) : false;
 }
 
 bool Units::getUnitsInBox(vector<df::unit *> &units, const cuboid &box, std::function<bool(df::unit *)> filter) {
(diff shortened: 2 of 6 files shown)


library/modules/Units.cpp
```diff
@@ -122,7 +122,11 @@ constexpr uint32_t exclude_flags2 = (
 
 bool Units::isActive(df::unit *unit) {
     CHECK_NULL_POINTER(unit);
-    return !unit->flags1.bits.inactive;
+    if (unit->flags1.bits.inactive)
+        return false;
+    else if (unit->flags1.bits.move_state || unit->flags1.bits.can_swap)
+        return true; // These are always unset when leaving the map
+    return linear_index(world->units.active, &df::unit::id, unit->id) >= 0;
```
myk002, line 129: `Units::isActive` is used in many many places. I'm worried that adding an O(N) search for every call will be a performance killer. I'm also concerned that changing the behavior of this function will break existing callers, both Lua and C++. I'd like to see some research done on those two points before merging this PR.

Bumber64: I did a check on my fort, and all of my units in `units.active` had at least one of `inactive`, `move_state`, or `can_swap` set. There was an `inactive` unit without the other two flags set that had its `pos` directly above my well. It's possible the combination could represent a ghost or climbing unit? Probably not going to be a performance consideration given how infrequently those occur. Then there's the actual raiding units, which you won't hit while iterating `units.active`, and for which the check is mandatory when iterating `units.all`.

I went through all the uses of `isActive`, `getPosition`, `getUnitsInBox`, and `isUnitInBox` I could find, as well as the `inactive` flag. There's no change in behavior to anything iterating the `units.active` vector (or indirectly through `forCitizens`/`getCitizens`). (`isDead` is used instead of `isActive` for the case of actual dead units.) Anything iterating `units.all` wasn't properly considering raiding units, which was just `RemoteTools.cpp`.

I don't fully understand `plugins/preserve-rooms.cpp` but it didn't look like the changes break anything. You'd know better than me.

Bumber64:
pre-commit.ci autofix

plugins/preserve-rooms.cpp
```diff
@@ -446,7 +446,7 @@ static void handle_missing_assignments(color_ostream &out,
         if (!hf || hf->died_year > -1 || was_expelled(hf))
             continue;
         if (auto unit = df::unit::find(hf->unit_id)) {
-            if (Units::isActive(unit) && !Units::isDead(unit) && active_unit_ids.contains(unit->id)) {
+            if (!Units::isDead(unit) && Units::isActive(unit) && active_unit_ids.contains(unit->id)) {
```
Bumber64, line 449: Moved `isActive` after `isDead`, which is a simple flag check. Not sure how `isActive` compares to testing an unordered set. The latter is probably much faster for units that are actually raiding? In that case we should test `isActive` afterwards, or maybe switch to directly testing the `inactive` flag.

library/modules/Units.cpp
ab9rf: That call to `linear_index` is going to (potentially) evict the entire L1D cache and up to half of the L2 cache on the executing core, which is a substantial performance penalty that I do not want in any code that is likely to be run once (or more than once) per tick.

That test is not acceptable in a high-velocity code path. If we need this information in a high-velocity code path, we need to figure out a way to cache it in a manner that doesn't require thousands of scattered reads.

library/RemoteTools.cpp
```diff
@@ -584,7 +584,7 @@ static command_result ListUnits(color_ostream &stream,
 
     if (in->scan_all())
     {
-        auto &vec = df::unit::get_vector();
+        auto &vec = df::global::world->units.active;
```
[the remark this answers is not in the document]
Bumber64: Probably. Everyone would've assumed raiding units didn't pass the "active" test, and the code hasn't been touched in 7-13 years, so I didn't suspect it would break anything (more than Toady already did).

PR #4972, opened 2024-10-01
Migrate widgets constants to use getter/setter style

wiktor-obrebski:
Pair PR: https://github.com/DFHack/scripts/pull/1313

Continuation of https://github.com/DFHack/dfhack/pull/4955, its answer for the comment:
https://github.com/DFHack/dfhack/pull/4955#discussion_r1779366381

Shortly, its migrate widgets constants to use getter/setter style.


docs/dev/Lua API.rst
```diff
@@ -5348,11 +5348,15 @@ If the panel has already been maximized in this fashion, then it will jump to
 its minimum size. Both jumps respect the resizable edges defined by the
 ``resize_anchors`` attribute.
 
-The time duration that a double click can span is defined by the global variable
-``DOUBLE_CLICK_MS``. The default value is ``500`` and can be changed by the end
+The time duration that a double click can span can be controlled by widgets API:
```
myk002, line 5351: Could you update this to indicate that the `control-panel` or `gui/control-panel` interfaces should be used to set these properties? That goes for the others as well. You can remove the `widget.*` function references. These docs were written before the control panel was a thing.

The `gui/control-panel` interface looks like this:
image

and the `control-panel` interface looks like this:
image

the functions should no longer be called by anything other than the control panel interfaces

library/lua/gui/widgets.lua
```diff
@@ -30,9 +30,38 @@ DimensionsTooltip = require('gui.widgets.dimensions_tooltip')
...
+    Panel.DOUBLE_CLICK_MS = value
+end
+
+---@return boolean
+function getStandardScroll()
+    return Scrollbar.STANDARDSCROLL
+end
+function setStandardScroll(value)
+    Scrollbar.STANDARDSCROLL = value
+end
```
myk002, line 47: STANDARDSCROLL is a constant and doesn't need a function wrapper. we *could* keep `getStandardScroll()` if you think it makes the API clearer, but there should not be a `setStandardScroll()`

library/lua/gui/widgets/labels/label.lua
```diff
@@ -277,7 +277,7 @@ Label.ATTRS{
     auto_width = false,
     on_click = DEFAULT_NIL,
     on_rclick = DEFAULT_NIL,
-    scroll_keys = STANDARDSCROLL,
+    scroll_keys = Scrollbar.STANDARDSCROLL,
```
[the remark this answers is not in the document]
myk002: the issue here, I think, is that widgets can't include `gui.widgets` since then it would be a circular dependency

another option is to move all constants (and settable preferences) into a separate, common file

myk002:
> Maybe I'm being a bit grumpy, but I feel like this is a stylistic preference and I'm not sure changing it around benefits us much vs. the potential for broken scripts.
> 
> I do see [#4955 (comment)](https://github.com/DFHack/dfhack/pull/4955#discussion_r1779366381), and I'm not sure how this change fixes the issue - or rather, why the issue was an issue. The things you're changing in https://github.com/DFHack/scripts/pull/1313/files are all resolved at runtime, so changing e.g. `widgets.DOUBLE_CLICK_MS` before your change should have had exactly the same effects as changing what `widgets.getDoubleClickMs()` returns after your change. You could potentially see stale values if those expressions were evaluated at module-load or script-load time, but that's true for both the old and new expressions. So I guess I'm not understanding what you're actually trying to fix.

The issue is not in seeing stale values in `widgets.DOUBLE_CLICK_MS`, but rather in the widgets that now reference `Panel.DOUBLE_CLICK_MS`. This PR fixes the case where:

1. Load DF
2. `widgets.DOUBLE_CLICK_MS` gets the value from `Panel.DOUBLE_CLICK_MS`
3. player starts `gui/control-panel`. it reads and displays `widgets.DOUBLE_CLICK_MS`, and player uses it to change value of `widgets.DOUBLE_CLICK_MS` (does not change `Panel.DOUBLE_CLICK_MS` since `widgets.DOUBLE_CLICK_MS` is a copy, not a reference)
4. a tool starts that includes a Panel and then the player double clicks on the header. The Panel reads `Panel.DOUBLE_CLICK_MS`, and it still has the old, original value.

myk002:
That sounds good to me -- @wiktor-obrebski could you move the settings/constants to a separate file? maybe call it `widgets/wcommon.lua`?

myk002:
right -- that's what I meant. We'd still need setters/getters at the public widgets.lua API level.

docs/dev/Lua API.rst
```diff
@@ -5348,11 +5348,9 @@ If the panel has already been maximized in this fashion, then it will jump to
...
 ``resize_anchors`` attribute.
 
-The time duration that a double click can span is defined by the global variable
-``DOUBLE_CLICK_MS``. The default value is ``500`` and can be changed by the end
-user with a command like::
-
-  :lua require('gui.widgets').DOUBLE_CLICK_MS=1000
+The time duration that a double click can span default value is ``500``.
+It can be controlled via the `control-panel` or `gui/control-panel` interfaces 
+(``Mouse double click speed`` option).
```
myk002, line 5353:
```suggestion
The time duration that a double click can span can be controlled
via the `control-panel` or `gui/control-panel` interfaces (``Mouse
double click speed`` option). It defaults to 500 ms.
```

docs/dev/Lua API.rst
```diff
@@ -5554,16 +5552,14 @@ while scrolling will result in faster movement.
...
-
-:``SCROLL_INITIAL_DELAY_MS``: The delay before the second scroll event.
-:``SCROLL_DELAY_MS``: The delay between further scroll events.
+scroll events when the mouse button is held down can be controlled
+via the `control-panel` or `gui/control-panel` interfaces:
 
-The defaults are 300 and 20, respectively, but they can be overridden by the
-user in their :file:`dfhack-config/init/dfhack.init` file, for example::
+1. The delay before the second scroll event use
+``Mouse initial scroll repeat delay`` option (default is 300)
```
myk002, line 5559:
```suggestion
1. The delay before the second scroll event is the ``Mouse initial scroll repeat
   delay`` setting (default is 300 ms)
```

docs/dev/Lua API.rst
```diff
@@ -5554,16 +5552,14 @@ while scrolling will result in faster movement.
...
 
-  :lua require('gui.widgets').SCROLL_DELAY_MS = 100
+2. The delay between further scroll events use
+``Mouse scroll repeat delay`` option (default is 20)
```
myk002, line 5562:
```suggestion
2. The delay between further scroll events is the ``Mouse scroll repeat delay`` option
   (default is 20 ms)
```

myk002 (approved):

PR #4824, opened 2024-07-29
Remove MapCache dependency for Items module

Bumber64:
Remove `MapCache` dependency for `Items` module.

Moved a bunch of code out of a switch statement in `createitem.cpp` into separate utility fns for readability.

Made improvements to `autodump.cpp` so it's in line with `gui/autodump`. (Convert to projectile, cancel any active job.)

Tidied up `autoclothing.cpp` a bit, made sure to initialize variables in a struct.

Made `dfhack.items.moveToInventory` args `use_mode` and `body_part` optional.

PR #4979, opened 2024-10-04
reenable `forceequip`

ab9rf:
Looked into this based on discord discussion, determined that the code compiles if adjusted to comport with @Bumber64's changes from #4824 so the question at this point is whether it works which I don't have enough information to evaluate


myk002 (commented):
remember also to add a line to New Tools in the changelog and update forceequip.rst (at least the tags, which should no longer be "unavailable")

Tags in the spreadsheet are:

```
adventure fort animals items military units
```

but that classification might need updating. is `animals` actually appropriate?

ab9rf:
yeah i was holding off on the administrivia until i get some intel that this actually works

myk002 (approved):
assuming it works, code looks ok

ab9rf:
> assuming it works, code looks ok

i got assurances on discord from a couple of testers that it worked, which is good enough for me given the nature of the tool

PR #4990, opened 2024-10-07
[nestboxes, timestream] fix handling of eggs

myk002:
testing shows that it is the properties of the egg that matter, not the properties of the mother

this also brings the behavior of `nestboxes` in line with the fertile egg indicator tweak. before this PR, eggs marked "fertile" could be unforbidden by `nestboxes`, collected, and eaten. This was confusing to players.

@Birkow FYI -- logic might need updating in https://github.com/DFHack/dfhack/pull/4947


Birkow:
@myk002 thanks for info, https://github.com/DFHack/dfhack/pull/4947 already checks fertile flag on egg item.

PR #4995, opened 2024-10-08
Widgets/text area

wiktor-obrebski:
This MR introduce `TextArea` widget.
This widgets is already in use in `gui/journal` and coming `gui/notes` tools.
Its provide much better support for text manipultation than TextField do.

Scripts journal has been migrated here:
https://github.com/DFHack/scripts/pull/1327

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -5521,7 +5521,7 @@ text wrapping, mouse control, text selection, clipboard support, history,
 and typical text editor shortcuts.
 
 Cursor Behavior
-===============
+~~~~~~~~~~~~~~~
 
 The cursor in the ``TextArea`` class is index-based, starting from 1,
 consistent with Lua's text indexing conventions.
@@ -5545,25 +5545,25 @@ TextArea Attributes:
 * ``init_text``: The initial text content for the text area.
 
 * ``init_cursor``: The initial cursor position within the text content.
-    If not specified, defaults to end of the text (length of ``init_text`` + 1).
+  If not specified, defaults to end of the text (length of ``init_text`` + 1).
 
 * ``text_pen``: Optional pen used to draw the text. Default is ``COLOR_LIGHTCYAN``.
 
 * ``select_pen``: Optional pen used for text selection. Default is ``COLOR_CYAN``.
 
 * ``ignore_keys``: List of input keys to ignore.
-    Functions similarly to the ``ignore_keys`` attribute in the ``EditField`` class.
+  Functions similarly to the ``ignore_keys`` attribute in the ``EditField`` class.
 
 * ``on_text_change``: Callback function called whenever the text changes.
-    The function signature should be ``on_text_change(new_text, old_text)``.
+  The function signature should be ``on_text_change(new_text, old_text)``.
 
 * ``on_cursor_change``: Callback function called whenever the cursor position changes.
-    Expected function signature is ``on_cursor_change(new_cursor, old_cursor)``.
+  Expected function signature is ``on_cursor_change(new_cursor, old_cursor)``.
 
 * ``one_line_mode``: If set to ``true``, disables multi-line text features.
-    In this mode the :kbd:`Enter` key is not handled by the widget
-    as if it were included in ``ignore_keys``.
-    If multiline text (including ``\n`` chars) is pasted into the widget, newlines are removed.
+  In this mode the :kbd:`Enter` key is not handled by the widget
+  as if it were included in ``ignore_keys``.
+  If multiline text (including ``\n`` chars) is pasted into the widget, newlines are removed.
 
 TextArea Functions:
 
@@ -5598,69 +5598,69 @@ TextArea Functions:
     Clears undo/redo history of the widget.
 
 Functionality
--------------
+~~~~~~~~~~~~~
 
 The TextArea widget provides a familiar and intuitive text editing experience with baseline features such as:
 
 - Text Wrapping: Automatically fits text within the display area.
 - Mouse and Keyboard Support: Standard keys like :kbd:`Home`, :kbd:`End`, :kbd:`Backspace`, and :kbd:`Delete` are supported,
-    along with gestures like double-click to select a word or triple-click to select a line.
+  along with gestures like double-click to select a word or triple-click to select a line.
 - Clipboard Operations: copy, cut, and paste,
-    with intuitive defaults when no text is selected.
+  with intuitive defaults when no text is selected.
 - Undo/Redo: :kbd:`Ctrl` + :kbd:`Z` and :kbd:`Ctrl` + :kbd:`Y` for quick changes.
 - Additional features include advanced navigation, line management,
-    and smooth scrolling for handling long text efficiently.
+  and smooth scrolling for handling long text efficiently.
 
 Detailed list:
 
 - Cursor Control: Navigate through text using arrow keys (Left, Right, Up,
-    and Down) for precise cursor placement.
+  and Down) for precise cursor placement.
 - Mouse Control: Use the mouse to position the cursor within the text,
-    providing an alternative to keyboard navigation.
+  providing an alternative to keyboard navigation.
 - Text Selection: Select text with the mouse, with support for replacing or
-    removing selected text.
+  removing selected text.
 - Select Word/Line: Use double click to select current word, or triple click to
-    select current line.
+  select current line.
 - Move By Word: Use :kbd:`Ctrl` + :kbd:`Left` and :kbd:`Ctrl` + :kbd:`Right` to
-    move the cursor one word back or forward.
+  move the cursor one word back or forward.
 - Line Navigation: :kbd:`Home` moves the cursor to the beginning of the current
-    line, and :kbd:`End` moves it to the end.
+  line, and :kbd:`End` moves it to the end.
 - Jump to Beginning/End: Quickly move the cursor to the beginning or end of the
-    text using :kbd:`Ctrl` + :kbd:`Home` and :kbd:`Ctrl` + :kbd:`End`.
+  text using :kbd:`Ctrl` + :kbd:`Home` and :kbd:`Ctrl` + :kbd:`End`.
 - Longest X Position Memory: The cursor remembers the longest x position when
-    moving up or down, making vertical navigation more intuitive.
+  moving up or down, making vertical navigation more intuitive.
 - New Lines: Easily insert new lines using the :kbd:`Enter` key, supporting
-    multiline text input.
+  multiline text input.
 - Text Wrapping: Text automatically wraps within the editor, ensuring lines fit
-    within the display without manual adjustments.
-- Scrolling behaviour for long text build-in.
+  within the display without manual adjustments.
+- Scrolling for long text entries.
 - Backspace Support: Use the backspace key to delete characters to the left of
-    the cursor.
+  the cursor.
 - Delete Character: :kbd:`Delete` deletes the character under the cursor.
 - Delete Current Line: :kbd:`Ctrl` + :kbd:`U` deletes the entire current line
-    where the cursor is located.
+  where the cursor is located.
 - Delete Rest of Line: :kbd:`Ctrl` + :kbd:`K` deletes text from the cursor to
-    the end of the line.
+  the end of the line.
 - Delete Last Word: :kbd:`Ctrl` + :kbd:`W` removes the word immediately before
-    the cursor.
+  the cursor.
 - Select All: Select entire text by :kbd:`Ctrl` + :kbd:`A`.
 - Undo/Redo: Undo/Redo changes by :kbd:`Ctrl` + :kbd:`Z` / :kbd:`Ctrl` +
-    :kbd:`Y`.
+  :kbd:`Y`.
 - Clipboard Operations: Perform OS clipboard cut, copy, and paste operations on
-    selected text, allowing you to paste the copied content into other
-    applications.
+  selected text, allowing you to paste the copied content into other
+  applications.
 - Copy Text: Use :kbd:`Ctrl` + :kbd:`C` to copy selected text.
-    - copy selected text, if available
-    - if no text is selected it copy the entire current line, including the
-      terminating newline if present
+  - copy selected text, if available
+  - if no text is selected it copy the entire current line, including the
+  terminating newline if present
 - Cut Text: Use :kbd:`Ctrl` + :kbd:`X` to cut selected text.
-    - cut selected text, if available
-    - if no text is selected it will cut the entire current line, including the
-      terminating newline if present
+  - cut selected text, if available
+  - if no text is selected it will cut the entire current line, including the
+  terminating newline if present
 - Paste Text: Use :kbd:`Ctrl` + :kbd:`V` to paste text from the clipboard into
-    the editor.
-    - replace selected text, if available
-    - If no text is selected, paste text in the cursor position
+  the editor.
+  - replace selected text, if available
+  - If no text is selected, paste text in the cursor position
 
 Scrollbar class
 ---------------
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -5589,14 +5589,29 @@ TextArea Functions:
 * ``textarea:scrollToCursor()``
 
     Scrolls the text area view to ensure that the current cursor position is visible.
-    This is useful for automatically scrolling when the user moves the cursor
-    beyond the visible region of the text area.
+    This happens automatically when the user interactively moves the cursor or
+    pastes text into the widget, but may need to be called when ``setCursor`` is
+    called programmatically.
 
 * ``textarea:clearHistory()``
 
     Clear undo/redo history of the widget.
 
-Functionality:
+Functionality
+-------------
+
+The TextArea widget provides a familiar and intuitive text editing experience with baseline features such as:
+
+- Text Wrapping: Automatically fits text within the display area.
+- Mouse and Keyboard Support: Standard keys like :kbd:`Home`, :kbd:`End`, :kbd:`Backspace`, and :kbd:`Delete` are supported,
+    along with gestures like double-click to select a word or triple-click to select a line.
+- Clipboard Operations: copy, cut, and paste,
+    with intuitive defaults when no text is selected.
+- Undo/Redo: :kbd:`Ctrl` + :kbd:`Z` and :kbd:`Ctrl` + :kbd:`Y` for quick changes.
+- Additional features include advanced navigation, line management,
+    and smooth scrolling for handling long text efficiently.
+
+Detailed list:
 
 - Cursor Control: Navigate through text using arrow keys (Left, Right, Up,
     and Down) for precise cursor placement.
@@ -5606,6 +5621,8 @@ Functionality:
     moving up or down, making vertical navigation more intuitive.
 - Mouse Control: Use the mouse to position the cursor within the text,
     providing an alternative to keyboard navigation.
+- Text Selection: Select text with the mouse, with support for replacing or
+    removing selected text.
 - New Lines: Easily insert new lines using the :kbd:`Enter` key, supporting
     multiline text input.
 - Text Wrapping: Text automatically wraps within the editor, ensuring lines fit
@@ -5621,8 +5638,6 @@ Functionality:
     the end of the line.
 - Delete Last Word: :kbd:`Ctrl` + :kbd:`W` removes the word immediately before
     the cursor.
-- Text Selection: Select text with the mouse, with support for replacing or
-    removing selected text.
 - Jump to Beginning/End: Quickly move the cursor to the beginning or end of the
     text using :kbd:`Ctrl` + :kbd:`Home` and :kbd:`Ctrl` + :kbd:`End`.
 - Select Word/Line: Use double click to select current word, or triple click to
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -5513,6 +5513,120 @@ The ``EditField`` class also provides the following functions:
 
   Inserts the given text at the current cursor position.
 
+TextArea class
+--------------
+
+Subclass of Panel; implements a multi-line text field with features such as
+text wrapping, mouse control, text selection, clipboard support, history,
+and typical text editor shortcuts.
+
+Attributes:
+
+* ``init_text``: 
+  The initial text content for the text area.
+
+* ``init_cursor``: 
+  The initial cursor position within the text content.
+
+* ``text_pen``: 
+  Optional pen used to draw the text.
+
+* ``select_pen``: 
+  Optional pen used for text selection.
+
+* ``ignore_keys``: 
+  List of input keys to ignore.
+  Functions similarly to the ``ignore_keys`` attribute in the ``EditField`` class.
+
+* ``on_text_change``: 
+  Callback function called whenever the text changes.
+  The function signature should be ``on_text_change(new_text)``.
+
+* ``on_cursor_change``: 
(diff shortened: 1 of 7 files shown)


library/lua/gui/widgets/text_area.lua
```diff
@@ -0,0 +1,166 @@
...
+    one_line_mode = false,
+    debug = false
+}
+
+function TextArea:init()
+    self.render_start_line_y = 1
+
+    self:addviews{
+        TextAreaContent{
+            view_id='text_area',
```
robob27, line 26: I think other widgets leave out the `view_id` so that the caller can specify their own and then access their elements internally like `self.subviews[element_index]`

library/lua/gui/widgets/text_area.lua
```diff
@@ -0,0 +1,166 @@
...
+            visible=not self.one_line_mode
+        }
+    }
+    self:setFocus(true)
+end
+
+function TextArea:getText()
+    return self.subviews.text_area.text
+end
+
```
robob27, line 57: Could we expose a `setText` method as well?

wiktor-obrebski: done

library/lua/gui/widgets/text_area.lua
```diff
@@ -0,0 +1,166 @@
+-- Multiline text area control
+
+local Panel = require('gui.widgets.containers.panel')
+local Scrollbar = require('gui.widgets.scrollbar')
+local TextAreaContent = require('gui.widgets.text_area.text_area_content')
+
+TextArea = defclass(TextArea, Panel)
+
+TextArea.ATTRS{
+    init_text = '',
```
robob27, line 10: Do you think it makes sense to switch this to `text` instead of `init_text`? It feels like `text` might align more closely with the API of other widgets like `EditField`, `Label`

library/lua/gui/widgets/text_area.lua
wiktor-obrebski: improved

library/lua/gui/widgets/text_area.lua
wiktor-obrebski: I am afraid it will consfuse users and they will treat it as writeable property - where it is not. they need to to use `:setText(...)`

library/lua/gui/widgets/text_area.lua
robob27: Could we make it so that if you call `setText(text)` and then press CTRL-Z, it will revert to the text before the `setText` call?

robob27:
I really appreciate your work on all of this stuff! You really went all out with this widget and the result is fantastic. Journal is sooo good and it's awesome that every other script will have such a great textarea to work with. I have been looking forward to this PR 😄

library/lua/gui/widgets/text_area.lua
wiktor-obrebski: I have mixed feelings about it, can you present a usecase where it would be useful?

robob27: The situation where this came up for me is with this UI in a script I've been working on:

image

I'm using this to be able to quickly add notes to individual units. Once I no longer need the note for the unit, I'll clear it out with ALT+C. I know I could CTRL+A/backspace to clear all but I also wanted to provide a shortcut to clear the notes with a single hotkey press (I had this before while using an EditField here and still find it handy).

If a user ever did ALT+C by accident here, ideally they could CTRL+Z it.

wiktor-obrebski: ok. I would additionally expose `clearHistory` function for cases where it would be a required behaviour.

wiktor-obrebski: done

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+
+Subclass of Panel; implements a multi-line text field with features such as
+text wrapping, mouse control, text selection, clipboard support, history,
+and typical text editor shortcuts.
+
+Attributes:
+
+* ``init_text``: The initial text content for the text area.
+
+* ``init_cursor``: The initial cursor position within the text content.
```
myk002, line 5527: Should document default cursor position (at end of text, right?)

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+* ``select_pen``: Optional pen used for text selection.
+
+* ``ignore_keys``: List of input keys to ignore.
+  Functions similarly to the ``ignore_keys`` attribute in the ``EditField`` class.
+
+* ``on_text_change``: Callback function called whenever the text changes.
+  The function signature should be ``on_text_change(new_text)``.
+
+* ``on_cursor_change``: Callback function called whenever the cursor
+  position changes. The function signature should be ``on_cursor_change(new_cursor_pos)``.
```
myk002, line 5540: I suggest also passing the old text/old cursor position so callbacks that care don't have to keep that state themselves. Callbacks that don't care don't even have to include the parameter in their function signatures.

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+
+* ``one_line_mode``: Boolean attribute that, when set to ``true``,
+  disables multi-line text features and restricts the text area to a single line.
+
+Functions:
+
+* ``textarea:getText()``
+
+    Returns the current text content of the ``TextArea`` widget as a string.
```
myk002, line 5549: Should document that the caller must interpret string.char(10) as newline

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+  Functions similarly to the ``ignore_keys`` attribute in the ``EditField`` class.
+
+* ``on_text_change``: Callback function called whenever the text changes.
+  The function signature should be ``on_text_change(new_text)``.
+
+* ``on_cursor_change``: Callback function called whenever the cursor
+  position changes. The function signature should be ``on_cursor_change(new_cursor_pos)``.
+
+* ``one_line_mode``: Boolean attribute that, when set to ``true``,
+  disables multi-line text features and restricts the text area to a single line.
```
myk002, line 5543: should document whether string.char(10) will be interpreted as the cp437 glyph. It would be nice if it did, but that can be implemented later if it's not already true.

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+
+Functions:
+
+* ``textarea:getText()``
+
+    Returns the current text content of the ``TextArea`` widget as a string.
+
+* ``textarea:setText(text)``
```
myk002, line 5551: this should take an optional second parameter to set the cursor for feature parity with EditField. -1 can mean "end of text" as a shortcut.

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+
+    Sets the content of the ``TextArea`` to the specified string ``text``.
+    The cursor position will not be adjusted, so should be set separately.
+
+* ``textarea:getCursor()``
+
+    Returns the current cursor position within the text content.
+    The position is represented as a single integer, starting from 1.
```
myk002, line 5559: should document whether newlines are included in the offset. since there are several cursor-related functions/parameters, maybe document cursor semantics in the intro text above the attributes.

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+
+* ``textarea:setCursor(cursor)``
+
+    Sets the cursor position within the text content.
+
+* ``textarea:scrollToCursor()``
+
+    Scrolls the text area view to ensure that the current cursor position is visible.
+    This is useful for automatically scrolling when the user moves the cursor
+    beyond the visible region of the text area.
```
myk002, line 5569: Scrolling on cursor movement doesn't happen automatically? I'd expect this to be the default so the parent isn't required to implement on_cursor_change for a good default behavior. What are the use cases for on_cursor_change callback other than to call this function? If we had that "keep cursor visible when moving cursor" default functionality, are there any other use cases for this function?

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+- Line Navigation: :kbd:`Home` moves the cursor to the beginning of the current
+    line, and :kbd:`End` moves it to the end.
+- Delete Current Line: :kbd:`Ctrl` + :kbd:`U` deletes the entire current line
+    where the cursor is located.
+- Delete Rest of Line: :kbd:`Ctrl` + :kbd:`K` deletes text from the cursor to
+    the end of the line.
+- Delete Last Word: :kbd:`Ctrl` + :kbd:`W` removes the word immediately before
+    the cursor.
+- Text Selection: Select text with the mouse, with support for replacing or
+    removing selected text.
```
myk002, line 5601: this should be moved above Ctrl-U so mouse users know they can skip paying attention to the deletion keyboard shortcuts. Actually, on second thought, I'd split this list into two sections:

1st section: baseline functionality (features that players would expect from any text editor: text wrapping, mouse selection, copy/paste, undo/redo). We can probably reduce the amount of documentation dedicated to well-understood features, like Home and End. You can just mention that "the usual keyboard keys are supported, like Home, End, Backspace, Delete, etc. In addition, common mouse gestures are supported, like double click to highlight a word or triple click to highlight a line."

2nd section: keyboard power user shortcuts

docs/dev/Lua API.rst
```diff
@@ -5513,6 +5513,116 @@ The ``EditField`` class also provides the following functions:
...
+- Jump to Beginning/End: Quickly move the cursor to the beginning or end of the
+    text using :kbd:`Ctrl` + :kbd:`Home` and :kbd:`Ctrl` + :kbd:`End`.
+- Select Word/Line: Use double click to select current word, or triple click to
+    select current line
```
myk002, line 5605: missing final period (also true for sentences below)

myk002 (commented):
first pass review done (docs only)

docs/dev/Lua API.rst
wiktor-obrebski: I will do it for the cursor, but I have doubts for it as default behaviour of `on_text_change`.
The text can be potentially long and storing and passing a copy of it on every keypress can be bad for performance and memory, especially that I assume most of the tools will not need it.

wiktor-obrebski: Done for `on_text_change`, also tests add.

docs/dev/Lua API.rst
wiktor-obrebski: added.
`\n` will be removed from the text, not rendered as glyph. 

Its not so obvious for me that they should be rendered as glyph, what do u think that?
maybe precise what do u mean by glyph, so I am sure we are think about the same thing.

docs/dev/Lua API.rst
wiktor-obrebski: I am not convinced for the 2 arguments notation.

Code like this `textarea:setText(text, 5)` is cryptic.
its not a lot of work to write it completly as
```lua
textarea
  :setText(text)
  :setCursor(5)
```

docs/dev/Lua API.rst
wiktor-obrebski: done

docs/dev/Lua API.rst
wiktor-obrebski: > If we had that "keep cursor visible when moving cursor" default functionality, are there any other use cases for this function?

It is default functionality for all default manipulation of TextArea by the user.

But there is `:setCursor` API where we allow the API consumer to manipulate the cursor.
This function do not automatically scroll to the cursor, to allow API consumer to better control.

e.g. imagine someone want to paste a text to end of the textarea for some reason, but do not change user cursor and scroll position.

I had a similar case with `gui/journal` Table of Contents feature, but I do not remember now what it was exactly.

docs/dev/Lua API.rst
wiktor-obrebski: I am not convinced, IMO we should keep all option described.
its quite think line what is "baseline" features and what now, it can be different for every user.

may proposition is to include and intro, like u proposed, but below it, in dedicated section, keep all shortcuts like now.
what do u think?

docs/dev/Lua API.rst
dhthwy: \n is ◙ in cp437 -  It has a graphical representation.

myk002: Yeah, in CP437, single-string multiline text isn't a thing. We're creating a funky hybrid here that needs careful documentation around the handling of `string.char(10)`. You can use ◙ in a name, for example. Currently you cannot use that character at all in this widget (which is a perfectly fine compromise, I think), but if we are to use `one_line_mode` TextArea widgets to replace EditField, we'd probably want to support the whole CP437 character set.

docs/dev/Lua API.rst
myk002: I agree the fluid API style is more readable. I was just looking for feature/API parity with the existing EditField.

docs/dev/Lua API.rst
myk002: Lua strings are immutable (a change to a string causes a new string to be allocated). If you take a reference to the string before modification and pass it to the callback along with the current (modified) string, there is no additional memory being used.

wiktor-obrebski: You are right, as such - I am going to implement the `old_text` param.

wiktor-obrebski: Done and tests added.

library/lua/gui/widgets/text_area/text_area_content.lua
```diff
@@ -0,0 +1,677 @@
...
+
+local CLIPBOARD_MODE = {LOCAL = 1, LINE = 2}
+local HISTORY_ENTRY = HistoryStore.HISTORY_ENTRY
+
+TextAreaContent = defclass(TextAreaContent, Widget)
+
+TextAreaContent.ATTRS{
+    text = '',
+    text_pen = COLOR_LIGHTCYAN,
+    ignore_keys = {'STRING_A096'},
```
myk002, line 15: same comment as above -- I think we can change the default to an empty list due to `INTERCEPT_HANDLED_HOTKEYS`

library/lua/gui/widgets/text_area/text_area_content.lua
```diff
@@ -0,0 +1,677 @@
...
+    pen_selection = COLOR_CYAN,
+    on_text_change = DEFAULT_NIL,
+    on_cursor_change = DEFAULT_NIL,
+    enable_cursor_blink = true,
+    debug = false,
+    one_line_mode = false,
+    history_size = 10,
```
myk002, line 22: the internet does not seem to have a good consensus of what a 'good' default history size is, but 10 seems low to me. How about 20 (or leave the attribute as `DEFAULT_NIL` and use the HistoryStore default of 25)?

library/lua/gui/widgets/text_area/text_area_content.lua
```diff
@@ -0,0 +1,677 @@
...
+    return text
+end
+
+function TextAreaContent:setRenderStartLineY(render_start_line_y)
+    self.render_start_line_y = render_start_line_y
+end
+
+function TextAreaContent:getPreferredFocusState()
+    return true
+end
```
myk002, line 68: it doesn't make sense for both TextArea and TextAreaContent to want focus since TextAreaContent is a child of TextArea. This function should be removed since it is TextArea that really needs the focus.

library/lua/gui/widgets/text_area/wrapped_text.lua
```diff
@@ -0,0 +1,70 @@
...
+    wrap_width = DEFAULT_NIL,
+}
+
+function WrappedText:init()
+    self:update(self.text, self.wrap_width)
+end
+
+function WrappedText:update(text, wrap_width)
+    self.lines = text:wrap(
+        wrap_width,
```
myk002, line 26: if `wrap_width` is `nil`, then the text will be wrapped at 72 characters. Is this appropriate? Should `wrap_width` instead default to `math.huge`?

test/library/gui/widgets.TextArea.lua
```diff
@@ -0,0 +1,3324 @@
+local gui = require('gui')
+local widgets = require('gui.widgets')
+
+config.target = 'core'
+
+local df_major_version = tonumber(dfhack.getCompiledDFVersion():match('%d+'))
```
myk002, line 6: now that 50.14 has been released (which supports the new cursor navigation keybindings), you can take this out and run the tests unconditionally

myk002 (commented):
Please add a line to the Lua section of the changelog.

To allow me to see the diff between reviews, could you keep the commits without force pushes? Force pushing causes GitHub to lose the review context.

docs/dev/Lua API.rst
wiktor-obrebski: Improved, what do u think about current version?

library/lua/gui/widgets/text_area.lua
```diff
@@ -0,0 +1,184 @@
...
+
+function TextArea:setCursor(cursor_offset)
+    return self.text_area:setCursor(cursor_offset)
+end
+
+function TextArea:clearHistory()
+    return self.text_area.history:clear()
+end
+
+function TextArea:onCursorChange(cursor, old_cursor)
```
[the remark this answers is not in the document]
wiktor-obrebski: this make code much harder to read and understand just to hide private API, I do not like it.
u propose a lot of local function that are in reality hided class function.
also, currying, which include another layer of understanding for a new dev.

maybe instead we should incorporate e.g. Python "private" functions by convention?
e.g. `TextArea:__onCursorChange`. This make code still quite simple to read and it will be very hard to have an accidental naming collision.

library/lua/gui/widgets/text_area.lua
```diff
@@ -0,0 +1,184 @@
...
+    if (self.frame_body.height >= lines_count) then
+        render_start_line_y = 1
+    end
+
+    self.render_start_line_y = render_start_line_y
+    self.text_area:setRenderStartLineY(self.render_start_line_y)
+end
+
+function TextArea:renderSubviews(dc)
+    self.text_area.frame_body.y1 = self.frame_body.y1-(self.render_start_line_y - 1)
```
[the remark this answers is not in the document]
wiktor-obrebski: check `TextAreaContent:onRenderBody` I still only render the lines that are visible.

myk002:
I should clarify that force pushing is ok to amend a just-committed commit. The problems happen when force pushing so that already-reviewed commits get a new hash.

library/lua/gui/widgets/text_area.lua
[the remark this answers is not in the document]
myk002: ok, good. that's fine then. now that you mention it, I *do* remember discussing this back then.

maybe a comment in this function would be worthwhile so future me doesn't complain about it again

library/lua/gui/widgets/text_area.lua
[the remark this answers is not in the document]
myk002: I'd be ok with something like that. @lethosor do you have a preference?

docs/dev/Lua API.rst
```diff
@@ -5560,10 +5560,10 @@ TextArea Attributes:
 * ``on_cursor_change``: Callback function called whenever the cursor position changes.
     Expected function signature is ``on_cursor_change(new_cursor, old_cursor)``.
 
-* ``one_line_mode``: Boolean attribute that, when set to ``true``,
-    disables multi-line text features and restricts the text area to a single line.
-    If any "\n" (``string.char(10)``) are available in the text,
-    they will be removed from the text
+* ``one_line_mode``: If set to ``true``, disables multi-line text features.
+    In this mode the :kbd:`Enter` key is not handled by the widget
+    as if it were included in ``ignore_keys``.
```
myk002, line 5565:
```suggestion
    as if it were included in ``ignore_keys``.
```

docs/dev/Lua API.rst
```diff
@@ -5589,44 +5589,60 @@ TextArea Functions:
...
+    called programmatically.
 
 * ``textarea:clearHistory()``
 
-    Clear undo/redo history of the widget.
+    Clears undo/redo history of the widget.
 
-Functionality:
+Functionality
+-------------
```
myk002, line 5601: welcome to the wonderful world of restructured text header underline rules. Currently, this section is interpreted like this:
image
that is, `Functionality` is a peer of `TextArea class` and `Cursor Behavior` is a *higher level* header than `TextArea class`.

You need to scan through the existing headers and discover what underline style is used for what level of header. It's different for every file, and is determined by whatever the file uses first for that particular header depth.

The indentation for list continuation lines is also apparently different from what sphinx expects. You can see how this PR's text is currently rendered here: https://dfhack--4995.org.readthedocs.build/en/4995/docs/dev/Lua%20API.html#textarea-class

You can get to that link by clicking on the "details" link for the ReadTheDocs build action: 
image

docs/dev/Lua API.rst
```diff
@@ -5589,44 +5589,60 @@ TextArea Functions:
...
+    text using :kbd:`Ctrl` + :kbd:`Home` and :kbd:`Ctrl` + :kbd:`End`.
 - Longest X Position Memory: The cursor remembers the longest x position when
     moving up or down, making vertical navigation more intuitive.
-- Mouse Control: Use the mouse to position the cursor within the text,
-    providing an alternative to keyboard navigation.
 - New Lines: Easily insert new lines using the :kbd:`Enter` key, supporting
     multiline text input.
 - Text Wrapping: Text automatically wraps within the editor, ensuring lines fit
     within the display without manual adjustments.
+- Scrolling behaviour for long text build-in.
```
myk002, line 5636:
```suggestion
- Scrolling for long text entries.
```

library/lua/gui/widgets/text_area.lua
```diff
@@ -101,15 +100,15 @@ end
...
             cursor_offset
         )
-        self:updateScrollbar(cursor_liny_y)
+        self:updateScrollbar(cursor_line_y)
     end
 end
 
 function TextArea:getPreferredFocusState()
-    return self.parent_view.focus
+    return self.parent_view and self.parent_view.focus or true
```
myk002, line 111: why not just `true`?

wiktor-obrebski: I read the docs, but I was always confused by this "preferred focus" feature. 
Changed to `true`

library/lua/gui/widgets/text_area.lua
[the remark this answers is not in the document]
wiktor-obrebski: done.

myk002 (approved):

library/lua/gui/widgets/text_area/text_area_content.lua
```diff
@@ -0,0 +1,673 @@
...
+    self.sel_end = nil
+    self.clipboard = nil
+    self.clipboard_mode = CLIPBOARD_MODE.LOCAL
+    self.render_start_line_y = 1
+
+    self.cursor = nil
+
+    self.main_pen = dfhack.pen.parse({
+        fg=self.text_pen,
+        bg=COLOR_RESET,
```
ab9rf, line 35: Question: why is `bg` defaulted to `COLOR_RESET`? This is not a valid value for a background color, and this setting results in an out of bound access when the rendering system tries to retrieve the color triple for color -1, which doesn't exist.

What is the intention here?

PR #4996, opened 2024-10-09
Fix DOUBLE_CLICK_MS reference

robob27:
`Panel.DOUBLE_CLICK_MS` was [moved here](https://github.com/DFHack/dfhack/commit/52ed5162fa1f212d18d73a3ad50723db759efbfb)

It looks like this reference in `List` got missed causing errors like `hack\lua\gui\widgets\list.lua:344: attempt to compare number with nil`. Switched it to use `common.DOUBLE_CLICK_MS`

### DIFF
--- a/library/lua/gui/widgets/list.lua
+++ b/library/lua/gui/widgets/list.lua
@@ -4,7 +4,6 @@ local common = require('gui.widgets.common')
 local Widget = require('gui.widgets.widget')
 local Scrollbar = require('gui.widgets.scrollbar')
 local Label = require('gui.widgets.labels.label')
-local Panel = require('gui.widgets.containers.panel')
 
 local getval = utils.getval
 local to_pen = dfhack.pen.parse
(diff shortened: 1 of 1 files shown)


library/lua/gui/widgets/containers/panel.lua
```diff
@@ -312,7 +313,7 @@ function Panel:onInput(keys)
 
     if self.resizable and y == 0 then
         local now_ms = dfhack.getTickCount()
-        if now_ms - self.last_title_click_ms <= common.DOUBLE_CLICK_MS then
+        if now_ms - self.last_title_click_ms <= widgets.getDoubleClickMs() then
```
robob27, line 316: This one wasn't causing errors but I assume it should also be using the getter

robob27:
Ah I think there are some circular reference issues 🤔

library/lua/gui/widgets/containers/panel.lua
robob27: Nevermind, this causes issues

myk002 (approved):

PR #5022, opened 2024-11-10
Scrolling `TabBar`

robob27:
- Modify `TabBar` to allow tabs to be scrolled instead of wrapping by setting `wrap=false` in the `TabBar` attributes. `wrap` defaults to true so any existing instances of `TabBar` should be unaffected (tested with launcher and control-panel and they're looking good).
- Add tests for Tab/TabBar

I made a little [demo script](https://gist.github.com/robob27/1edf22efff8bedd0956987390612d4a3) with a scrolling TabBar alongside a wrapping one with a resizable window so the behavior can be easily tested (requires this branch). Testers should try changing `autoarrange_subviews` in the script and ensure it doesn't break things, play around with the options, and double check that existing consumers of this widget like launcher and control-panel doesn't get messed up in some way that I missed.

Some notes on the expected behavior:

- If the window is resized while the tabs are scrolled to the right, it will just reset the scroll offset back to 0
- The scroll left/scroll right buttons should only be shown when needed. If resizing the window, the scroll right button should disappear as soon as the last tab is fully in view
- When switching tabs with hotkeys or by clicking, the selected tab should be scrolled into view + a slight (customizable) offset to show part of the next tab. The default offset seems generally appropriate but didn't work well with a `TabBar` with a small width, so I made it possible to customize it.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -6194,9 +6194,16 @@ TabBar class
 ------------
 
 This widget implements a set of one or more tabs to allow navigation between groups
-of content. Tabs automatically wrap on the width of the window and will continue
-rendering on the next line(s) if all tabs cannot fit on a single line.
+of content.
 
+:wrap: If true, tabs automatically wrap on the width of the window and will
+       continue rendering on the next line(s) if all tabs cannot fit on a single line.
+       If false, tabs will be truncated and can be scrolled using ``scroll_key``
+       and ``scroll_key_back``, mouse wheel or by clicking on the scroll labels
+       that will automatically appear on the left and right sides of the tab bar
+       as needed. When clicking on a tab or using ``key`` or ``key_back`` to switch tabs,
+       the selected tab will be scrolled into view if it is not already visible.
+       Defaults to true.
 :key: Specifies a keybinding that can be used to switch to the next tab.
       Defaults to ``CUSTOM_CTRL_T``.
 :key_back: Specifies a keybinding that can be used to switch to the previous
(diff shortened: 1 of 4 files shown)


robob27:
pre-commit.ci autofix

robob27:
I should probably make it possible to customize the colors of the scroll labels as well 🤔

docs/dev/Lua API.rst
```diff
@@ -6222,6 +6229,28 @@ rendering on the next line(s) if all tabs cannot fit on a single line.
...
            situations, returns ``self.active_tab_pens``, if ``self.get_cur_page() == idx``,
            otherwise returns ``self.inactive_tab_pens``.
+:scroll_key: Specifies a keybinding that can be used to scroll the tabs to the right.
+             Defaults to ``CUSTOM_ALT_T``.
+:scroll_key_back: Specifies a keybinding that can be used to scroll the tabs to the left.
+                  Defaults to ``CUSTOM_ALT_Y``.
+:scroll_left_text: The text to display on the left scroll label.
+                   Defaults to "<<<".
+:scroll_right_text: The text to display on the right scroll label.
+                    Defaults to ">>>".
```
myk002, line 6239: As a matter of style consistency, I don't think this should be configurable. I think players would expect this to be constant in all situations where it appears.

I also suggest making the horizontal scroll icon look more like this:
image

perhaps with only one left/right triangle  if two triangles stacked vertically doesn't fit well. The code for that example screen is here: 
- widget: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/sort.lua#L1144
- characters: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/sort.lua#L1015

myk002 (commented):
This looks awesome! Thank you!

Scrolling is very important to get right. Could you add some unit tests in `test/library/gui/widgets.TabBar.lua` (need to create a new file, but can be based on `test/library/gui/widgets.Scrollbar.lua`)

docs/dev/Lua API.rst
robob27: Ah yeah that's a much nicer symbol. Will fix!

robob27:
@myk002 Added a bunch of tests for the new scroll behavior as well as some tests for existing behavior. I think I covered pretty much everything? But let me know if you see something I missed and I'm happy to add it 😄

myk002 (approved):

PR #5027, opened 2024-11-15
Experimental SDL Console

dhthwy:
It's set to build it by default and disables the other ones for now. Still in draft.


ab9rf:
You're inconsistent in your usage of `class` versus `struct` in a lot of this code. They are not interchangeable (as the build log demonstrates) and it might be best if you picked a consistent approach and stuck with it. My personal preference here is to use `struct` for types that are mostly "pure data containers" and `class` for types that have significant behaviors attached to them.

dhthwy:
True my intent aligns with your preference. GCC 14 doesn't warn me about that, it even missed a problem with initialization order that MSVC caught.

Anyway, it needs to use DFSDL, but I'll probably just add a symbol resolver function to it for now. 

This needs some infrastructure work (some basic layout machinery would be nice). And I'm not 100% sure if the design is sound. If anyone happens to see anything awful, feel free to point it out.

dhthwy:
Alright, I think I fixed that inconsistency.  Of course, it doesn't build until I make it bind to sdl at run time.

There are other warts (member variables declared at the start of an object definition vs end), some that are public that should be private, etc. You get used to looking at the same code and these types of warts start to look normal.

Issue #3835, opened 2023-09-30
reenable `infiniteSky`


ab9rf:
i reenabled `infiniteSky` for testing purposes, it compile and ran more or less correctly, but the added z-level is gray on the map instead, so some adjustment is needed. the minimap appears correct

myk002:
see discussion here on potential issues/solutions: https://discord.com/channels/793331351645323264/793331351645323267/1203086314132607006

Issue #2606, opened 2023-01-14
infiniteSky not creating usable Z-levels


myk002:
when trying to build above, says 'blocked at this location' , extra added z levels are all grey

myk002:
dup of https://github.com/DFHack/dfhack/issues/3835

Issue #254, opened 2014-06-01
Bug: infiniteSky resets sky level - also causing cave-ins


GoombaBrother:
When manually adding a few sky levels, they disappear again till the level with the highest construction once after a while, even without saving/reloading inbetween. 

E.g. Having initally 17 sky levels with highest building at level 16. Then adding manually 8 levels via command "InfiniteSky 8" nets to a total of 25. Then building up the contruction to let's say Level 19. Suddenly the additional sky-levels disappear. Total levels drop down to 19 as there is currently the highest construction. (Which should actually not be possible as the game demands one empty sky level between the highest construction and the ceiling of the map). 

Even Weirder: The sudden drop causes cave-ins, when there are constructions going on, even some levels below, e.g. at level 14 of the tower.

I prepared some screenshots. Those cave-ins shouldn't actually happen as the caved in tiles are directly connected to a staircase:

http://i.imgur.com/glLa5tp.png
http://i.imgur.com/ldG85cP.png
http://i.imgur.com/daUaXwW.png

The very same tiles could be built without problems, when increasing the SkyLevels manually again.

Command "InfiniteSky Enable" causes the same problems. Works at the beginning, but after a while (ranging from a few minutes to an hour or maybe more), the sky levels are reseted again the same way as described above.

add: Please note, the level-numbers above are just examples. Happens regularly at any levels.

I am using the Standard Install of LazyNewbPack with the included tools like Dwarf Therapist. No other mods.

lethosor:
Any idea if this is still an issue in 0.43.05?

expwnent:
Nobody's reported anything but I think few people use it.

ab9rf:
Testing seems to confirm that this bug has been remediated, although exactly when, I cannot reliably say.

PR #5029, opened 2024-11-16
Reenable `infiniteSky`

NicksWorld:
This should allow infiniteSky to generate usable tiles in newer DF versions. The bug involving cave-ins removing z-levels appears to still be present, but to my knowledge the old workaround of saving/loading should still work.

As a allocating a new `df::map_block` and adding it to the map seemed to be the solution, there is a possibility of some data being lost for new tiles. For now the following data should be preserved:
- Tiles receive light
- Tiles are marked as outdoors
- Standard and current temperature are inherited from map_block below

Fixes: #2606
Fixes: #3835


plugins/infiniteSky.cpp
```diff
@@ -95,25 +95,68 @@ void doInfiniteSky(color_ostream& out, int32_t howMany) {
...
+
+            df::map_block *last_air_block = blockColumn[z_count_block - 1];
+            for (int32_t count = 0; count < howMany; count++) {
+                df::map_block *air_block = new df::map_block();
+                std::fill(&air_block->tiletype[0][0],
+                          &air_block->tiletype[0][0] + (16 * 16),
+                          df::tiletype::OpenSpace);
+
+                // Set block positions properly (based on prior air layer)
+                air_block->map_pos = blockColumn[z_count_block - 1]->map_pos;
```
quietust, line 119: Any reason for not using `last_air_block` here (and 2 lines down)?

NicksWorld: No reason, I simply forgot to update those lines.

myk002:
Ok, so what you're saying is that #254 is still an issue, but #2606 (and #3835) should be fixed, right?

myk002 (commented):
Could you add a line to the New Tools section in `docs/changelog.txt`? It should look like this: https://github.com/DFHack/dfhack/blob/develop/docs/changelog.txt#L199

Also, please edit `docs/plugins/infiniteSky.rst` and replace the `dfhack-tool` attributes with these lines:
```
    :summary: Automatically allocate new z-levels of sky.
    :tags: fort auto design map
```

Finally, in a separate PR, please feel free to add your name to `docs/about/Authors.rst`!

It looks like infiniteSky needs to be brought up to code standards before it is included in a release. For example, it needs to persist its enabled state and restore it when the fort is reloaded. It also needs to be registered with the control panel so it shows up for players in the tool list. None of this is particularly difficult, but it will take time and may be more than you want to tackle at this point. If you'd like to dig deeper, I can walk you through it. Otherwise, I can do a pass after this PR is merged.

It also looks like some of the existing logic is incorrect. The "new construction scanning" in `plugin_onupdate` doesn't actually scan new constructions. New constructions are inserted sorted, not appended to the end of the vector. Therefore, `infiniteSky` may not add levels when it should. (it would work if the z coordinate were the primary sorting value, but for some reason, the **x** coordinate is the primary sorting value. why? why?)

myk002:
I suspect this plugin would be better off responding to the `EventType::CONSTRUCTION` event rather than trying to do its own detection

plugins/lua/infinite-sky.lua
```diff
@@ -0,0 +1,27 @@
...
+    return argparse.processArgsGetopt(args, {
+            {'h', 'help', handler=function() opts.help = true end},
+        })
+end
+
+function parse_commandline(opts, args)
+    local positionals = process_args(opts, args)
+
+    if opts.help then return end
+end
```
myk002, line 25:
```suggestion
function parse_commandline(opts, args)
    local positionals = argparse.processArgsGetopt(args, {
            {'h', 'help', handler=function() opts.help = true end},
        })

    if opts.help or positionals[1] == 'help' then
        opts.help = true
        return
    end

    if positionals[1] then
        opts.n = argparse.positiveInt(positionals[1])
    end
end
```

myk002: can be simplified, and getopt processing should happen first

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
+
+#include "Core.h"
+#include "DataDefs.h"
+#include "Debug.h"
+#include "Export.h"
+#include "LuaTools.h"
+#include "PluginManager.h"
```
myk002, line 7:
```suggestion
#include "Core.h"
#include "Debug.h"
#include "LuaTools.h"
#include "PluginManager.h"
```

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include "df/map_block.h"
+#include "df/map_block_column.h"
+#include "df/world.h"
+#include "df/z_level_flags.h"
+
+#include <cstring>
+#include <string>
+#include <vector>
+
+using namespace std;
```
myk002, line 22:
```suggestion
using std::string;
using std::vector;
```

myk002: we discourage `using namespace std` and instead encourage `using` specific individual classes

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+                   is_enabled ? "enabled" : "disabled");
+    }
+
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_shutdown(color_ostream &out) {
+    cleanup();
+    return CR_OK;
+}
```
myk002, line 93: listener cleanup on shutdown is managed by the framework and you don't need to handle it here.

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+    config = World::GetPersistentSiteData(CONFIG_KEY);
+
+    if (!config.isValid()) {
+        DEBUG(control, out)
+            .print("no config found in this save; initializing\n");
+        config = World::AddPersistentSiteData(CONFIG_KEY);
+        config.set_bool(CONFIG_IS_ENABLED, is_enabled);
+    }
+
+
```
myk002, line 105:
```suggestion

```

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+    return CR_OK;
+}
+
+DFhackCExport command_result plugin_onstatechange(color_ostream &out,
+                                                  state_change_event event) {
+    if (event == DFHack::SC_WORLD_UNLOADED) {
+        if (is_enabled) {
+            DEBUG(control, out)
+                .print("world unloaded; disabling %s\n", plugin_name);
+            is_enabled = false;
```
myk002, line 120: it would be cleaner to unregister the listener here too

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+static const string CONFIG_KEY = string(plugin_name) + "/config";
+static PersistentDataItem config;
+enum ConfigValues {
+    CONFIG_IS_ENABLED = 0,
+};
+
+command_result infiniteSky (color_ostream &out, std::vector <std::string> & parameters);
+
+static void constructionEventHandler(color_ostream& out, void* ptr);
+EventManager::EventHandler handler(plugin_self, constructionEventHandler,0);
```
myk002, line 48: 0 might be a bit too aggressive since it would cause the EventManager to do a linear scan on every tick. How about 11?

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+void doInfiniteSky(color_ostream& out, int32_t howMany) {
+    CoreSuspender suspend;
+    int32_t x_count_block = world->map.x_count_block;
+    int32_t y_count_block = world->map.y_count_block;
+    int32_t z_count_block = world->map.z_count_block;
+    df::map_block ****block_index = world->map.block_index;
+
+    for (int32_t a = 0; a < x_count_block; a++) {
+        for (int32_t b = 0; b < y_count_block; b++) {
```
myk002, line 143: would the Maps::cuboid class make this a little cleaner? e.g.
```cpp
Maps::cuboid cube(0, 0, world->map.z_count_block,
    world->map.x_count_block - 1, world->map.y_count_block - 1, world->map.z_count_block);
cube.forCoord([&](df::coord bpos) {
            ...
            return true;
        });
```

plugins/infinite-sky.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    infinitesky_options opts;
+    if (!Lua::CallLuaModuleFunction(out, "plugins.infinite-sky",
+                                    "parse_commandline",
+                                    std::make_tuple(&opts, parameters)) ||
+        opts.help)
+        return CR_WRONG_USAGE;
+
+    if (opts.n != 0) {
```
myk002, line 261:
```suggestion
    if (opts.n > 0) {
```

myk002: makes the semantics a bit clearer (even though negative numbers can't happen with current code)

myk002 (commented):
looks great! Thanks!

myk002 (approved):

PR #5040, opened 2024-11-23
REGRESSION: Fix console command completion help

dhthwy:
Console command completion help was inadvertently broken in a prior commit. This PR should resolve the issue. Tested on posix console.

This could probably use a test, if I can figure out a proper way to do it.


quietust:
Looks like this broke way back in 2022 with commit e926e1116eb9490ccb6ce875a86dd33418f303c8.

dhthwy:
if helpdb handled command aliases we could make some nonsense alias to a command and do it that way. Otherwise we'd have to construct a new one - and I don't know how to do that in a test.

There aren't _that_ many commands, the limit of 8 should be much higher. Maybe even no limit provided that the list of possibles isn't the full command set.

PR #5052, opened 2024-11-28
add regression test for size of global_table

myk002:
CI expected to fail for this PR, but will be corrected with https://github.com/DFHack/df-structures/pull/811

### DIFF
--- a/test/core.lua
+++ b/test/core.lua
@@ -38,3 +38,23 @@ function test.getHackPath_chdir()
         expect.eq(clean_path(dfhack.getHackPath()), clean_path(fs.getcwd()))
     end)
 end
+
+-- validates that the size in df.globals.xml is correct
+function test.global_table_size()
+    local elem_size = df.global_table_entry:sizeof()
+    local declared_arr_size = df.global.global_table:sizeof() // elem_size
+    for i=0,declared_arr_size-1 do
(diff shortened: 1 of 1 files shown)


test/core.lua
```diff
@@ -38,3 +38,23 @@ function test.getHackPath_chdir()
...
+
+-- validates that the size in df.globals.xml is correct
+function test.global_table_size()
+    local elem_size = df.global_table_entry:sizeof()
+    local declared_arr_size = df.global.global_table:sizeof() // elem_size
+    for i=0,declared_arr_size-1 do
+        expect.true_(df._displace(df.global.global_table[0], i, elem_size).address,
+            'nil address found in middle of global_table at idx ' .. i)
+    end
+    if df._displace(df.global.global_table[0], declared_arr_size, elem_size).address then
```
ab9rf, line 50: This is UB if `declared_arr_size` is greater than the actual size of the array.

myk002: I think the failed `expect.true_` line would have terminated the test and prevented it from getting here, but it's true the test could be considerably simplified. see version 2

myk002: simplified test. new output:
```txt
Check failed! global_table size mismatch: expected: 366, actual: 379: "366" ~= "379"
  at ...eamapps/common/Dwarf Fortress/hack/scripts/test/core.lua:50
```

ab9rf (approved):

PR #5053, opened 2024-11-29
#5031: Fix accessibility logic for wagons

### LINKED ISSUE #5031: `gui/pathable` says depot is wagon accessible but it is not
Savegame: https://discord.com/channels/793331351645323264/807444467140788254/1307832160165822649

dhthwy:
Fixes #5031

The save for 5031 has down stairs blocking the wagon path. This PR attempts more robust wagon accessibility handling.

Other issues:

There is also an issue with the 'follow mouse' functionality where walkability is being correctly set for top ramps. This causes the UI to erroneously display an 'X' over the ramp tile, incorrectly indicating that the tile (or the z-level below) is inaccessible.

I am unsure about tests for this.

Me, from elsewhere:

"It looks like walkable is computed by df, right? The down ramp itself is not walkable. I suspect we'll have to special case ramps in Maps  and have it tie its walkability group to an adjacent tiles group" AND ensure the tile below the ramp also matches the group of its adjacent tile.

Assumed invariant: adjacent tiles belong to the same walkability group.


myk002:
I see some wonkiness with trees and boulders:
image

test save: https://drive.google.com/file/d/1ID5Tu9XteyEwCpeKuFVjf3ygfWiXe6K0/view?usp=sharing

library/modules/Maps.cpp
myk002, line 1: I see the visual benefits of what you're doing here, but I'm wary about changing the semantics of an established API. Moreover, I'm wary about misrepresenting the contents of the DF walkability cache. I think we should do this PR without ramp tops being shown as walkable and discuss the issue later on Discord.

plugins/pathable.cpp
```diff
@@ -203,24 +208,105 @@ struct FloodCtx {
...
+    }
 
     if (ctx.wgroup == Maps::getWalkableGroup(pos))
         return true;
 
+    // RAMP_TOP is assigned a walkability group if that commit is accepted
+    // so this test, I think, would be useless.
     if (shape == df::tiletype_shape::RAMP_TOP ) {
         df::coord pos_below = pos + df::coord(0, 0, -1);
         if (Maps::getWalkableGroup(pos_below)) {
```
myk002, line 312: assuming the Maps module change does *not* go in quite yet, do we need to add logic here that checks whether the walkability group of `pos_below` matches `ctx.wgroup`?

myk002 (commented):
don't forget a changelog entry!

myk002:
The edge tiles need a separate check (which can be solved in a separate PR). The edge tile map is populated indiscriminately. There should be a wagon pathability check on each tile before it is added to the map

plugins/pathable.cpp
```diff
@@ -248,12 +332,13 @@ static void check_wagon_tile(FloodCtx & ctx, const df::coord & pos) {
     ctx.seen.emplace(pos);
 
     if (ctx.entry_tiles.contains(pos)) {
-        ctx.wagon_path.emplace(pos);
+        ctx.wagon_path.emplace(pos); // Is this needed?
```
myk002, line 335: probably not since the rendering of the edge tile will overwrite the rendering of the wagon path

plugins/pathable.cpp
```diff
@@ -293,7 +378,10 @@ static bool wagon_flood(color_ostream &out, unordered_set<df::coord> * wagon_pat
         TRACE(log,out).print("checking tile: (%d, %d, %d); pathability group: %d\n", pos.x, pos.y, pos.z,
             Maps::getWalkableGroup(pos));
 
-        if (entry_tiles.contains(pos)) {
+        // Ensure our wagon flood end points lands on an edge tile.
+        // When there is no path to the depot entry_tiles will not
+        // contain any edge tiles.
+        if ((pos.x == 0 || pos.y == 0) && entry_tiles.contains(pos)) {
```
myk002, line 384: why are we restricting the entry edge match to the top or left of the map?

plugins/pathable.cpp
```diff
@@ -293,7 +378,10 @@ static bool wagon_flood(color_ostream &out, unordered_set<df::coord> * wagon_pat
         TRACE(log,out).print("checking tile: (%d, %d, %d); pathability group: %d\n", pos.x, pos.y, pos.z,
             Maps::getWalkableGroup(pos));
 
-        if (entry_tiles.contains(pos)) {
+        // Ensure our wagon flood end points lands on an edge tile.
+        // When there is no path to the depot entry_tiles will not
+        // contain any edge tiles.
```
myk002, line 383: this statement is not quite correct. when there is no path to the depot, then the *search_edge* will never contain any edge tiles. `entry_tiles` contains edge tiles regardless of the accessibility of the depot.

plugins/pathable.cpp
```diff
@@ -203,20 +208,99 @@ struct FloodCtx {
...
+
+static bool is_wagon_traversible(FloodCtx & ctx, const df::coord & pos, const df::coord & prev_pos) {
     auto tt = Maps::getTileType(pos);
     if (!tt)
         return false;
 
     auto shape = tileShape(*tt);
-    if (shape == df::tiletype_shape::STAIR_UP || shape == df::tiletype_shape::STAIR_UPDOWN)
-        return false;
+    auto& occ = *Maps::getTileOccupancy(pos);
```
myk002, line 269: while the return value should never ever be NULL if `tt` is not NULL, it's still common practice to check it before dereferencing it

plugins/pathable.cpp
```diff
@@ -203,20 +208,99 @@ struct FloodCtx {
...
+            // workshops (tiles with open space are handled by the tile check)
+            if (is_wagon_tile_traversible(*tt) == false)
+                return false;
+            break;
+        case tile_building_occ::Floored:
+            // depot, lowered bridges or retractable bridges, forbidden hatches
+            break;
+        default:
+            //NOTREACHED
+            break;
```
myk002, line 302: can remove the default case so the compiler complains if another case ever appears

plugins/pathable.cpp
```diff
@@ -203,20 +208,99 @@ struct FloodCtx {
...
+        if (b.gate_flags.bits.closed)
+            return true;
     }
+    // Doors, traps..etc
+    return false;
+}
+
+// NOTE: When i.e. tracks, stairs have a bridge over them, the tile will have
+// an occupancy of floored.
+static bool is_wagon_tile_traversible(df::tiletype& tt) {
```
myk002, line 240: `df::tiletype` is just an int (an enum), so no need to take a reference

plugins/pathable.cpp
```diff
@@ -203,20 +208,99 @@ struct FloodCtx {
...
+    auto shape = tileShape(tt);
+    auto special = tileSpecial(tt);
+    auto material = tileMaterial(tt);
+
+    // Allow murky pool ramps
```
myk002, line 245: maybe explain why you're calling out murky pools in particular

myk002 (commented):
code overally LGTM, just one substantive question and a few nits

plugins/pathable.cpp
myk002: actually, keep this in, since we might want to highlight the overlap (by coloring it green?)

myk002 (approved):

PR #5067, opened 2024-12-01
Fix crashing bugs in infinite-sky

NicksWorld:
Core is now suspended before running any lua functions.
Identity structure now properly includes an end marker.

Tested on Linux and Windows, the command appears to be functioning as it should.


myk002 (approved):
Can now take out the superfluous CoreSuspender in doInfiniteSky, but merging this now for expediency

Issue #5069, opened 2024-12-02
timestream should increment age for all buildings


myk002:
why the heck does age count up one per tick? but it does

PR #5070, opened 2024-12-02
[timestream] update building counters

myk002:
Fixes: https://github.com/DFHack/dfhack/issues/5068
Fixes: https://github.com/DFHack/dfhack/issues/5069

@ab9rf @quietust are there any values of `age` that we need to be sure not to skip?


ab9rf:
i don't know of any "magic values" for `buildingst::age` and will have to research it

PR #5075, opened 2024-12-03
adjustments to core suspender

### LINKED ISSUE #5074: console hang if user makes a typo while lua is busy
The console will hang if the user makes a typo in the console while lua is busy

reproduction:
in the launcher, do `lua "while(true) do end"` (which tosses lua into an infinite loop)
in the console, type any invalid command and hit enter
the console will deadlock

ab9rf:
this is an attempted step to improve semantics around core suspenders. this does not actually resolve #5074 but it is intended as a step toward doing so

### DIFF
--- a/library/include/Core.h
+++ b/library/include/Core.h
@@ -356,19 +356,22 @@ namespace DFHack
 
     class CoreSuspenderBase : protected std::unique_lock< decltype(Core::CoreSuspendMutex) > {
     protected:
-        size_t lock_count = 0;
+        Core& core;
+        bool isLocked = false;
         using mutex_type = decltype(Core::CoreSuspendMutex);
         using parent_t = std::unique_lock< mutex_type >;
         std::thread::id tid;
 
-        CoreSuspenderBase(std::defer_lock_t d) : CoreSuspenderBase{&Core::getInstance(), d} {}
+        CoreSuspenderBase() : CoreSuspenderBase{ Core::getInstance() } {}
 
-        CoreSuspenderBase(Core* core, std::defer_lock_t) :
+        CoreSuspenderBase(Core& core) :
+            core{ core },
             /* Lock the core (jk not really) */
-            parent_t{core->CoreSuspendMutex,std::defer_lock},
+            parent_t{core.CoreSuspendMutex,std::defer_lock},
             /* Mark this thread to be the core owner */
             tid{}
         {}
+
     public:
         void lock()
         {
@@ -385,33 +388,32 @@ namespace DFHack
 
         void unlock()
         {
-            auto& core = Core::getInstance();
+            if (!isLocked)
+                return;
             /* Restore core owner to previous value */
             core.ownerThread.store(tid, std::memory_order_release);
             if (tid == std::thread::id{})
                 Lua::Core::Reset(core.getConsole(), "suspend");
             parent_t::unlock();
-            lock_count--;
+            isLocked = false;
         }
 
         bool owns_lock() const noexcept
         {
-            return lock_count > 0;
-//            return parent_t::owns_lock();
+            return isLocked;
         }
 
         ~CoreSuspenderBase() {
-            while (lock_count > 0)
-                unlock();
+            if (isLocked) unlock();
         }
 
-    private:
+    protected:
         void complete_lock()
         {
             auto& core = Core::getInstance();
             tid = core.ownerThread.exchange(std::this_thread::get_id(),
                 std::memory_order_acquire);
-            lock_count++;
+            isLocked = true;
         }
 
         friend class MainThread;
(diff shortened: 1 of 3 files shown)


ab9rf:
with the last commit this now does resolve #5074

ab9rf:
@dhthwy has confirmed that this works on linux, so i'm releasing it for review

ab9rf:
thanks again to @NicksWorld for their invaluable help on this PR

library/include/Core.h
```diff
@@ -353,26 +354,33 @@ namespace DFHack
         friend struct CoreSuspendReleaseMain;
     };
 
-    class CoreSuspenderBase  : protected std::unique_lock<std::recursive_mutex> {
+    class CoreSuspenderBase : protected std::unique_lock< decltype(Core::CoreSuspendMutex) > {
     protected:
-        using parent_t = std::unique_lock<std::recursive_mutex>;
+        size_t lock_count = 0;
```
NicksWorld, line 359: The `CoreSuspender` in its current state doesn't have a need for keeping a lock count. `unique_lock` will throw an exception if locked more than once, disallowing the CoreSuspender being locked multiple times. In the destructor, unlock can be called if `unique_lock::owns_lock` as it tracks whether that particular instance owns the lock.

dhthwy: It's a recursive mutex though -- it permits multiple locks by definition

NicksWorld: The recursive mutex does support multiple locks, but the unique_lock provides a wrapper around that mutex. The unique_lock prevents ever obtaining the lock twice by throwing an exception if a second attempt is made. Multiple CoreSuspender instances can all have locks at the same time thanks to the recursive mutex, but no single CoreSuspender instance can have multiple locks.

ab9rf: It's a logic error for a `CoreSuspender` to get locked twice. Normally `CoreSuspender`s are locked at construction time, but we provide a deferred-lock variant that has be manually locked, so the `lock` method is public for that specific purpose.

There may be better ways to do this, and I'm totally open to suggestions for improvement.

I'll probably revise this to remove the lock counting; @NicksWorld has convinced me that it's not necessary

I'm thinking about removing the deferred-locking variant of `CoreSuspender` and creating a `ConditionalCoreSuspender` that does a `try_lock` at construction, and if it succeeds returns an object that holds a `CoreSuspender` which will be released when the `ConditionalCoreSuspender` is destroyed, and which evaluates to true if cast to boolean, and if it fails returns an object holding nothing and which evaluates to false if cast to boolean. This would allow something like `if (ConditionalCoreSuspender ccs) { ... }` for things that should not block if a core suspender cannot be acquired, while still providing RAII-like behavior and insuring that the lock is locked at most once and is unlocked when it goes out of scope

ab9rf: i should also add that it's _not_ a logic error for a thread to acquire a `CoreSuspender` when it already has one, as long as it's a _different_ `CoreSuspender`. this will result in the underlying mutex being locked twice, which is harmless because the underlying mutex is recursive

i'm increasingly of the opinion that we should make the `lock` and `unlock` methods on `CoreSuspender` private and only allow `CoreSuspender`s to be locked at creation and unlocked at destruction, instead of relying on consumers to properly manage `CoreSuspender` state themselves, and move the cases that currently require the consumer to manage `CoreSuspender` state themselves to classes that abstract that management for them

ab9rf:
The last batch of changes drastically simplify the `CoreSuspender` API by removing consumer access to `CoreSuspender` state, adding a `ConditionalCoreSuspender` subclass that provides a conditional locking mechanic, fixes various errors around use of an explicit core argument when creating a `CoreSuspender`, which required having `CoreSuspenderBase` hold a reference to the core used to initialize it. I also changed the explicit-core constructors to use a `Core&` instead of a `Core*`

library/include/Core.h
```diff
@@ -419,34 +433,60 @@ namespace DFHack
...
         {
             lock();
         }
-        CoreSuspender(Core* core, std::defer_lock_t) :
-            CoreSuspenderBase{core, std::defer_lock}
-        {}
+
+        ~CoreSuspender() {
+            if (owns_lock()) unlock();
+        }
```
myk002, line 453: are we avoiding making the destructor virtual for performance reasons? also, no need to check owns_lock(), neh?

library/include/Core.h
```diff
@@ -353,47 +354,60 @@ namespace DFHack
...
             if (tid == std::thread::id{})
                 Lua::Core::Reset(core.getConsole(), "suspend");
             parent_t::unlock();
         }
 
-        bool owns_lock() const noexcept
-        {
-            return parent_t::owns_lock();
+        ~CoreSuspenderBase() {
+            if (owns_lock()) unlock();
```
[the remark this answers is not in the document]
ab9rf: belt and suspenders never hurt anyone

library/include/Core.h
ab9rf: this is probably solely a result of the destructor being different before i refactored everything and i simply failed to notice that the implementation was the same

ab9rf: actually looking at it more no, we need the explicit destructor. this is because the `CoreSuspenderBase` destructor will call `CoreSuspenderBase::unlock`, and we need the `CoreSuspender` destructor to call `CoreSuspender::unlock`

remember: none of these methods are virtual

myk002: ah ok -- If would be worth adding a comment to highlight that for future readers

myk002 (approved):

myk002:
I'm not certain of the full effect of this change, but the changes look reasonable to merge for practical testing

myk002:
@lethosor any concerns before we merge this?

library/include/Core.h
```diff
@@ -353,47 +354,60 @@ namespace DFHack
...
+            unlock();
         }
 
-        ~CoreSuspenderBase() {
-            if (owns_lock())
-                unlock();
+    protected:
+        void complete_lock()
+        {
+            auto& core = Core::getInstance();
```
myk002, line 406: why do we need to get another core here (shadowing the class field)?

ab9rf: there were a bunch of these i removed, i just missed that one in the refactor. thanx

myk002 (approved):

myk002:
I'll merge this for now so we can begin practical testing

PR #5076, opened 2024-12-03
isDanger(): don't mark visitors hostile

dhthwy:
This only handles night creatures (the hostile one in the save this patch fixes) and opposed to lifers (necromancers? I don't recall.) 

May want this to apply to them all. It is possible for a demon to visit as a merchant.

I don't see anything obvious where this change will cause unwanted side effects in any of the callers of this function.

### DIFF
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -642,7 +642,8 @@ bool Units::isDanger(df::unit *unit) {
     if (isTame(unit) || isOwnGroup(unit))
         return false;
 
-    return isCrazed(unit)
+    return unit->flags2.bits.visitor_uninvited
+        || isCrazed(unit)
         || isInvader(unit)
         || isAgitated(unit)
         || isSemiMegabeast(unit)
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -646,7 +646,7 @@ bool Units::isDanger(df::unit *unit) {
         || isInvader(unit)
         || isAgitated(unit)
         || isSemiMegabeast(unit)
-        || (!isVisiting(unit) && (isOpposedToLife(unit) || isNightCreature(unit)))
+        || ((isOpposedToLife(unit) || isNightCreature(unit)) && !unit->flags2.bits.visitor)
         || isGreatDanger(unit);
 }
 
--- a/library/modules/Units.cpp
+++ b/library/modules/Units.cpp
@@ -644,10 +644,9 @@ bool Units::isDanger(df::unit *unit) {
 
     return isCrazed(unit)
         || isInvader(unit)
-        || isOpposedToLife(unit)
         || isAgitated(unit)
         || isSemiMegabeast(unit)
-        || isNightCreature(unit)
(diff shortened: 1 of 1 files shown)


library/modules/Units.cpp
```diff
@@ -644,10 +644,9 @@ bool Units::isDanger(df::unit *unit) {
 
     return isCrazed(unit)
         || isInvader(unit)
-        || isOpposedToLife(unit)
         || isAgitated(unit)
         || isSemiMegabeast(unit)
-        || isNightCreature(unit)
+        || (!isVisiting(unit) && (isOpposedToLife(unit) || isNightCreature(unit)))
```
myk002, line 649: isVisiting also returns true for "unwanted" visitors. Is this check enough?

dhthwy: Actually, you're right, forgotton beasts should be marked as visitor_uninvited = true. So they would be dangerous.

I'm confused why the isVisitor() function permits a hostile as a visitor? The tests in that function imply it's a friendly visit. IMO, that should go into the isDanger() or similar function

dhthwy: Sorry, must've been thinking of another function. Thought there were more tests in the isVisitor() function.

I suppose then just check for visitor flag for that then.  But if we want to allow for i.e. friendly demons then I think that visitor flag check would need to be done at the top before the rest of the checks.

library/modules/Units.cpp
```diff
@@ -644,10 +644,9 @@ bool Units::isDanger(df::unit *unit) {
 
     return isCrazed(unit)
         || isInvader(unit)
-        || isOpposedToLife(unit)
         || isAgitated(unit)
         || isSemiMegabeast(unit)
-        || isNightCreature(unit)
+        || ((isOpposedToLife(unit) || isNightCreature(unit)) && !unit->flags2.bits.visitor)
```
myk002, line 649: shouldn't this also check `visitor_uninvited`? If they *are* an unwanted visitor, shouldn't this return true?

maybe that can be an independent check further up as one of the OR clauses.

myk002:
Note to self: when we merge this, revert https://github.com/DFHack/scripts/pull/1343 (and write a more general changelog message about hostile classification in general. it affects several tools)

myk002 (approved):

library/modules/Units.cpp
```diff
@@ -642,13 +642,14 @@ bool Units::isDanger(df::unit *unit) {
...
+    // other demons may visit as a diplomat
     return isCrazed(unit)
         || isInvader(unit)
         || isOpposedToLife(unit)
         || isAgitated(unit)
-        || isSemiMegabeast(unit)
-        || isNightCreature(unit)
-        || isGreatDanger(unit);
+        || unit->flags2.bits.visitor_uninvited;
+        || ((isGreatDanger(unit) || isNightCreature()) && !unit->flags2.bits.visitor)
```
myk002, line 652:
```suggestion
        || unit->flags2.bits.visitor_uninvited
        || ((isGreatDanger(unit) || isNightCreature()) && !unit->flags2.bits.visitor);
```

PR #5077, opened 2024-12-03
test that job ids increase monotonically

dhthwy:


test/modules/job_fortress.lua
```diff
@@ -0,0 +1,20 @@
...
+
+config.target = 'core'
+config.mode = 'fortress'
+
+-- EventManager job completion handling expects sorted order
+function test.jobIDsAreSorted()
+    local is_sorted = true
+    local prev_id = nil
+
+    for _, job in utils.listpairs(df.global.world.jobs.list) do
```
myk002, line 11:
```suggestion
    -- assumes there are at least some "naturally added" jobs currently in the list
    -- but this should always be true for CI test saves
    for _, job in utils.listpairs(df.global.world.jobs.list) do
```

PR #5082, opened 2024-12-06
#5061 fix: add missing pthread_mutexattr_init()

dhthwy:
Initialize attributes before use. I believe this is the source of UB on Linux, which results in stochastic hangs. It is easily reproducable when launching from steam on my system.


ab9rf (approved):
Obviously required by the specification

ab9rf:
additional changelog not required as this is just completing #5061 which hasn't been published yet

PR #5083, opened 2024-12-06
Likely Minor: cleanup the mutex attributes object after mutex pthread_mutex_init().…

dhthwy:
… pthread_mutex_init() temporarily uses the attributes object for configuration but does not take any ownership.

There is a memory allocation performed by pthread_mutexattr_init().

Assuming luai_userstateopen() is only called once or rarely, this resolves a small resource leak.


ab9rf (approved):
i did verify that, per the [spec](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutexattr_init.html):
> After a mutex attributes object has been used to initialize one or more mutexes, any function affecting the attributes object (including destruction) shall not affect any previously initialized mutexes.

so this is safe

myk002 (approved):

PR #5093, opened 2024-12-09
Proper symbol demangling for `doReadClassName`

NicksWorld:
This replaces the flawed manual demangling with proper demanglers capable of supporting additional language features such as namespaces.

For example, the mangled symbol `N6DFHack21dfhack_lua_viewscreenE` was previously demangled as `DFHack21dhack_lua_viewscreenE`. This patch ensures demangling returns the proper `DFHack::dfhack_lua_viewscreen`. This also ensures that class names returned remain consistent between platforms.

Basic functionality has been tested on Linux, Windows, and Wine, with no unexpected results seen so far.

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -330,7 +330,7 @@ elseif(UNIX)
     set(PROJECT_LIBS rt dl dfhack-md5 ${DFHACK_TINYXML})
 else(WIN32)
     # FIXME: do we really need psapi?
-    set(PROJECT_LIBS psapi dfhack-md5 ${DFHACK_TINYXML})
+    set(PROJECT_LIBS psapi dbghelp dfhack-md5 ${DFHACK_TINYXML})
 endif()
 
 set(VERSION_SRCS DFHackVersion.cpp)
--- a/library/Process-darwin.cpp
+++ b/library/Process-darwin.cpp
@@ -37,6 +37,7 @@ distribution.
 #include <set>
 #include <cstdio>
 #include <cstring>
+#include <cxxabi.h>
 using namespace std;
 
 #include <md5wrapper.h>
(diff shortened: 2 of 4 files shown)


NicksWorld:
It is true `doReadClassName` is used mainly for checking viewscreen names. The function is also used sparingly in other places like `virtual_identity` checking to provide debug information.

The main reason for the change was for cross-platform consistency and ensuring the function returns a result that isn't trimmed in unusual manners. This mainly affects classes inside of namespaces or including template parameters. Class names that aren't in the format of `viewscreen_<name>st` are already modified via trimming the first 10 and last 2 chars when creating a focus string, losing more of their symbol.

If this is a change that doesn't seem warranted at this time, I am glad to simply repurpose portions of this experiment into improving `cxx_demangle`.

NicksWorld:
Through further testing I have discovered inconsistencies can occur when running under wine. The affected symbols were discovered via a `devel/scan-vtables` test, and appear to be in only specific highly-nested template parameters. Wine also returns the mangled string in place of the demangled string upon complete failure, which may be undesirable.

An example of an offending symbol:
```
.?AV?$_Func_impl_no_alloc@V<lambda_1>@?1??add_happiness_tex@widgets@@YAXV?$shared_ptr@Vcontainer@widgets@@@std@@PEAX@Z@X$$V@std@@
```
is demangled to
```cpp
std::shared_ptr<widgets::container>::_Func_impl_no_alloc
```

In the simpler symbols with fewer template arguments I am yet to see any unusual results.

NicksWorld:
All strings returned by doClassName are trimmed as described, not just ones not matching the standard pattern. At this point, it looks like adding Windows support to cxxDemangle would be preferable and introduce fewer potential issues.

The proper demangling of the symbol I previously provided would be:
```cpp
std::_Func_impl_no_alloc<`widgets::add_happiness_tex'::`2'::<lambda_1>,void>
```
The demangled result still looks somewhat incorrect, but at least seems more accurate.

Wine contains a flawed implementation of demangling which causes some symbols to fail demangling as well. As such, the use of the `UndecorateSymbolName` should likely only occur on native windows, which may be an acceptable requirement for `cxxDemangle` as it is only used by `devel/scan-vtables`, but likely is unacceptable in a core function used for viewscreens and lua.

For the time being I am closing this PR as the changes are not critical and may cause additional problems in some environments.

Issue #5102, opened 2024-12-14
Make plug-in commands run with suspended core 


myk002:
By default. We can add an option to command registration to have the command run with an unlocked core

ab9rf:
Sorry, I missed one. There's an extra CoreSuspender in there that is wrong and needs to be removed. I'll note it in a comment shortly

PR #5112, opened 2024-12-20
Suspend core by default when calling plugin command functions

myk002:
Fixes: https://github.com/DFHack/dfhack/issues/5102
Should be tested with https://github.com/DFHack/dfhack/pull/5075

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -2416,7 +2416,7 @@ int Core::Shutdown ( void )
     d->hotkeythread.join();
     d->iothread.join();
 
-    CoreSuspendClaimer suspend;
+    CoreSuspender suspend;
     if(plug_mgr)
     {
         delete plug_mgr;
--- a/library/PluginManager.cpp
+++ b/library/PluginManager.cpp
@@ -469,35 +469,32 @@ bool Plugin::reload(color_ostream &out)
 
 command_result Plugin::invoke(color_ostream &out, const std::string & command, std::vector <std::string> & parameters)
 {
-    Core & c = Core::getInstance();
     command_result cr = CR_NOT_IMPLEMENTED;
     access->lock_add();
     if (state == PS_LOADED) {
-        for (auto & cmd : commands) {
-            if (cmd.name != command)
-                continue;
-
+        if (auto cmdIt = std::ranges::find_if(commands, [&](const PluginCommand &cmd) { return cmd.name == command; });
+            commands.end() != cmdIt)
+        {
             // running interactive things from some other source than the console would break it
-            if (!out.is_console() && cmd.interactive)
+            if (!out.is_console() && cmdIt->interactive)
                 cr = CR_NEEDS_CONSOLE;
-            else if (cmd.guard) {
-                CoreSuspender suspend(&c);
-                if (!cmd.guard(c.getTopViewscreen())) {
+            else if (cmdIt->guard) {
+                CoreSuspender suspend;
+                if (!cmdIt->guard(Core::getInstance().getTopViewscreen())) {
                     out.printerr("Could not invoke %s: unsuitable UI state.\n", command.c_str());
                     cr = CR_WRONG_USAGE;
                 }
                 else {
-                    cr = cmd.function(out, parameters);
+                    cr = cmdIt->function(out, parameters);
                 }
             }
-            else if (cmd.unlocked) {
-                cr = cmd.function(out, parameters);
+            else if (cmdIt->unlocked) {
+                cr = cmdIt->function(out, parameters);
             }
             else {
-                CoreSuspender suspend(&c);
-                cr = cmd.function(out, parameters);
+                CoreSuspender suspend;
+                cr = cmdIt->function(out, parameters);
             }
-            break;
         }
     }
     access->lock_sub();
--- a/library/PluginManager.cpp
+++ b/library/PluginManager.cpp
@@ -472,41 +472,32 @@ command_result Plugin::invoke(color_ostream &out, const std::string & command, s
     Core & c = Core::getInstance();
     command_result cr = CR_NOT_IMPLEMENTED;
     access->lock_add();
-    if(state == PS_LOADED)
-    {
-        for (size_t i = 0; i < commands.size();i++)
-        {
-            PluginCommand &cmd = commands[i];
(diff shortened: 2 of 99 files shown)


library/PluginManager.cpp
```diff
@@ -472,41 +472,32 @@ command_result Plugin::invoke(color_ostream &out, const std::string & command, s
...
+    if (state == PS_LOADED) {
+        for (auto & cmd : commands) {
+            if (cmd.name != command)
+                continue;
+
+            // running interactive things from some other source than the console would break it
+            if (!out.is_console() && cmd.interactive)
+                cr = CR_NEEDS_CONSOLE;
+            else if (cmd.guard) {
+                CoreSuspender suspend(&c);
```
ab9rf, line 484: would prefer initializer syntax here `CoreSuspender suspend{&c}`

library/PluginManager.cpp
```diff
@@ -472,41 +472,32 @@ command_result Plugin::invoke(color_ostream &out, const std::string & command, s
...
-                    {
-                        out.printerr("Could not invoke %s: unsuitable UI state.\n", command.c_str());
-                        cr = CR_WRONG_USAGE;
-                    }
-                    else
-                    {
-                        cr = cmd.function(out, parameters);
-                    }
+    if (state == PS_LOADED) {
+        for (auto & cmd : commands) {
```
ab9rf, line 476: maybe use `std:;ranges:find_if` instead of an explicit loop?

library/include/PluginManager.h
```diff
@@ -99,11 +99,12 @@ namespace DFHack
                       const char * _description,
                       command_function function_,
                       bool interactive_ = false,
+                      bool unlocked_ = false,
                       const char * usage_ = ""
                      )
-            : name(_name), description(_description),
-              function(function_), interactive(interactive_),
-              guard(NULL), usage(usage_)
+            : name(_name), description(_description), function(function_),
```
ab9rf, line 105: would again prefer constructor syntax here

library/include/PluginManager.h
```diff
@@ -113,9 +114,9 @@ namespace DFHack
                       command_function function_,
                       command_hotkey_guard guard_,
                       const char * usage_ = "")
-            : name(_name), description(_description),
-              function(function_), interactive(false),
-              guard(guard_), usage(usage_)
+            : name(_name), description(_description), function(function_),
```
ab9rf, line 117: constructor syntax here as well

library/PluginManager.cpp
myk002: done

library/include/PluginManager.h
myk002: done

library/include/PluginManager.h
myk002: done

library/modules/Textures.cpp
```diff
@@ -322,6 +322,7 @@ struct tracking_stage_new_region : df::viewscreen_new_regionst {
 
     DEFINE_VMETHOD_INTERPOSE(void, logic, ()) {
         if (this->m_raw_load_stage != this->raw_load_stage) {
+            CoreSuspender guard;
```
ab9rf, line 325: these methods will never be invoked by anything other than DF's simulation thread
the only reason they would need a core lock is if they are going to manipulate structures in DFHack's core
we don't need to arbitrate access to DF's core in these because there is no plausible scenario in which they are called by anything other than the DF simulation thread

ab9rf (changes requested):
as per comments

i am especially not comfortable with adding a core suspender to viewscreen render or logic methods

library/PluginManager.cpp
myk002: done

myk002:
> as per comments

fixed

> i am especially not comfortable with adding a core suspender to viewscreen render or logic methods

what about overlay.cpp, where we call into Lua from render, logic, and feed?

library/modules/Textures.cpp
myk002: reverted

ab9rf:
> what about overlay.cpp, where we call into Lua from render, logic, and feed?

Might be needed there, since that's going to be using the Lua state in Core, which is shared across threads. I don't know if the lock in the LuaState is sufficient here. @lethosor may be in a better position to offer an opinion here than I am; he is more familiar with that code than I am

ab9rf (approved):
Having looked over the CoreSuspender semantics I think this is probably OK.

We do need to ensure that access to the global Lua state is properly arbitrated, but this isn't the PR to do that.

ab9rf:
Sorry I missed a mistake on review. There's a `CoreSuspender` in `Core::Shutdown` that absolutely should not be there

library/Core.cpp
```diff
@@ -2416,7 +2416,7 @@ int Core::Shutdown ( void )
     d->hotkeythread.join();
     d->iothread.join();
 
-    CoreSuspendClaimer suspend;
```
ab9rf, line 2419: This shouldn't be here at all. This is in `Core::Shutdown`, the core suspension mechanism has already been dismantled at this point and all DFHack threads. as well as the DF simulation thread, have been terminated. The suspender should be removed entirely since there's nothing to protect against here.

myk002:
> Did we remove CoreSuspendClaimer?

CoreSuspendClaimer has been an alias for CoreSuspender since #1304 (merged in 2018). I just removed the alias to make the now-uniform semantics clearer.

library/Core.cpp
myk002: Fixed in https://github.com/DFHack/dfhack/commit/ebac3772cec2f9150f562a50daac63e8dfa9634d

ab9rf:
> I'm guessing from context that it was intended to be used as a stand-in for CoreSuspender on the render thread to prevent deadlocks. 

it was aliased to `CoreSuspender` in a 2018 change. Also, there is simply no safe way to use a `CoreSuspender` from the render thread without risking a deadlock, because the render thread can be blocking the simulation thread. This is why we have the hotkey thread and the hotkey mailslot (to move work units from the render thread, which we cannot block, to the hotkey thread, which we can), so that work units that need a suspended core can be initialized from the render thread.

Issue #4991, opened 2024-10-07
Add PDBs to windows distribution


myk002:
- add `/Zi` to the compile flags
- add `/DEBUG` to the link flags

and package where?

ab9rf:
just store the PDBs as a build artifact, they do not need to be sent to subscribers

myk002:
It looks like ccache does not support the /Zi flag, so we might have to choose between expedient builds and Windows debug symbols.

As it stands, building with a warm cache for Windows takes about 3 minutes. Building without a cache takes about 20 minutes. That's enough of a difference to fuel a careful cost-benefit discussion.

ab9rf:
does ccache support `/Z7`?

myk002:
it looks like it might. I'll get you a link to the artifacts when the test build is done (and it hasn't failed yet at 50%)

PR #5114, opened 2024-12-21
generate pdb files for windows builds

myk002:
so we can retrieve them when debugging player-reported crashes on Windows

Fixes: #4991

This PR also adds a link to the online debug symbols to stderr.log and a new `dfhack-status.dat` file in the savegame directory.

example contents of the status file:

```txt
DF version:  v0.50.15 linux64 STEAM
DFHack version: 50.15-r1.2 (409f60f)
Tagged release: yes
Pre-release:    no
Build url: https://github.com/DFHack/dfhack/actions/runs/12475090213
```

We normally ask the player for their stderr.log when debugging crashes, so the link will be available even if the player does not have a savegame to share.

Example contents of `stderr.log`:

```txt
DFHack build: 50.15-r1.2-65-g409f60f
Build url: https://github.com/DFHack/dfhack/actions/runs/12475090213
Starting with working directory: /home/myk/.local/share/Steam/steamapps/common/Dwarf Fortress
Binding to SDL.
Identifying DF version.
Loading hack/symbols.xml ... OK
...
```

The link takes you to the GitHub workflow run results, where you can download the PDBs:
image


myk002:
The idea I had going into this was to package them as an artifact that could be downloaded from GitHub. I was testing to be sure the build worked before I made changes to the workflow files. However, if we have to choose between ccache and pdb files for Windows builds, there needs to be a more serious discussion about cost/benefit of this change.

myk002:
So the resulting workflow would be:
1. Get player-supplied crash and stderr.log OR the player savegame and repro instructions
2. go to the link in stderr.log or the dfhack-status.dat file in the savegame dir
3. download PDBs at the link
4. debug

myk002:
Merging so we can test PDB generation for Steam deployments. We can address further comments in follow-up PRs

PR #5120, opened 2024-12-22
thread safety checks

ab9rf:
The main thing this PR does is add some code to monitor for attempts to acquire a `CoreSuspender` from the DF render thread, which is impermissible and will lead to deadlock. It currently only uses an `assert` for this (which means it's only operational in a debug build).

It also removes some dead code related to threading and fixes a couple of semantic errors:

* `Core::Shutdown` attempts to unlock the main core lock, but cannot do so because it always runs on the render thread while the main core lock is held by the simulation thread. This is replaced by an atomic flag `shutdown` which `Core::Shutdown` sets and which `Core::Update` checks.
* Plugin unloading would attempt to acquire a `CoreSuspender` to save plugin site and world data. Plugin unload during shutdown cannot acquire a `CoreSuspender` (because shutdown runs on the render thread), so unload now skips saving plugin site and world data during shutdown. This is harmless because site and world data cannot be saved during shutdown anyway.

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -1528,6 +1528,7 @@ df::viewscreen * Core::getTopViewscreen() {
 }
 
 bool Core::InitMainThread() {
+    // this hook is always called from DF's main (render) thread, so capture this thread id
     df_render_thread = std::this_thread::get_id();
 
     Filesystem::init();
@@ -1689,6 +1690,7 @@ bool Core::InitMainThread() {
 
 bool Core::InitSimulationThread()
 {
+    // the update hook is only called from the simulation thread, so capture this thread id
     df_simulation_thread = std::this_thread::get_id();
     if(started)
         return true;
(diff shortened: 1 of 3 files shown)


library/Core.cpp
```diff
@@ -2519,6 +2508,13 @@ bool Core::DFH_SDL_Event(SDL_Event* ev) {
 
 bool Core::doSdlInputEvent(SDL_Event* ev)
 {
+    // DF only calls the sdl input event hook from the render thread
+    // we use this to identify the render thread
+    //
+    // first time this is called, set df_render_thread to calling thread
```
myk002, line 2514: stale comment?

myk002 (approved):

library/Core.cpp
ab9rf: indeed, good catch

PR #5121, opened 2024-12-22
add entry point for dfhooks_sdl_loop

myk002:
FYI: @dhthwy 

Note that DF doesn't actually call this hook yet, and won't until at least 50.16/51.01-beta27

Until then, you can probably synthesize this call by tapping into `dfhooks_sdl_event` and rate limiting with a timer

### DIFF
--- a/library/Hooks.cpp
+++ b/library/Hooks.cpp
@@ -51,7 +51,7 @@ DFhackCExport void dfhooks_update() {
 DFhackCExport void dfhooks_prerender() {
     if (disabled)
         return;
-    // TODO: render overlay widgets that are not attached to a viewscreen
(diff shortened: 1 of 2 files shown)


ab9rf:
we can get a better approximation by interposing `renderer_2d::render`, which is called from nearly the same point that the render hook is expected to be inserted

dhthwy:
cool. Yeah, renderer_2d::render is already being used, and rate limiting is in place. Should be easy to switch it.

20 fps normal and 10 fps while minimized (for preview rendering), no render at all if window isn't shown (except for minimized). iirc.

Reminds me, I need to push some updates.

PR #5132, opened 2024-12-25
test whether an xml header cache speeds up the build

myk002:


myk002:
result: it did not. repurposing this PR for light cleanup

PR #5135, opened 2024-12-25
check result code when unloading a DLL

### LINKED ISSUE #4300: calls to `dlclose`/`FreeLibrary` are unchecked
`dlclose` (on Linux) or `FreeLibrary` (on Windows) can fail under certain circumstances, but our plugin routines (in `PlugLoad-posix.cpp` and `PlugLoad-windows.cpp`) do not check the return value of `dlclose`/`FreeLibrary`.

The known circumstance in which `dlclose` will fail on Linux is if the DLL contains any symbols marked `STB_GNU_UNIQUE`; if a segment contains any such symbol, that segment is marked `NODELETE` by the loader, which prevents the dll from being unloaded. g++ will emit symbols with this flag for, among possibly other things, templated statics and statics inside of inline functions.

In order to prevent user surprise, then, we should detect whether DLL unload fails, and when it does issue an appropriate diagnostic, and mark the plugin as "broken" instead of as "unloaded".

[Relevant Discord discussion](https://discord.com/channels/793331351645323264/807443262704058408/1211379867556511775)

ab9rf:
fixes #4300


library/PlugLoad-posix.cpp
```diff
@@ -38,8 +38,12 @@ namespace DFHack
...
         return (void *) dlsym((void *)plugin, function);
     }
-    void ClosePlugin (DFLibrary * plugin)
+    bool ClosePlugin (DFLibrary * plugin)
     {
-        dlclose((void *) plugin);
+        // dlclose returns 0 on success, nonzero on failure
+        int res = dlclose((void *) plugin);
+        if (res != 0)
+            WARN(plugins).print("dlclose failed with error code %x\n", res);
```
myk002, line 46: needs a `DBG_DECLARE(core, plugins, DebugCategory::LINFO);` in this file

ab9rf: yeah that's part of why i wanted to combine the two into a single file

ab9rf:
i have not personally tested this on linux although the fact that it passes CI testing is a good sign

library/PlugLoad.cpp
```diff
@@ -0,0 +1,124 @@
...
+not claim that you wrote the original software. If you use this
+software in a product, an acknowledgment in the product documentation
+would be appreciated but is not required.
+
+2. Altered source versions must be plainly marked as such, and
+must not be misrepresented as being the original software.
+
+3. This notice may not be removed or altered from any source
+distribution.
+*/
```
myk002, line 23: Since this is a new file, I'd take this opportunity to take this boilerplate header out. we have a published repo-wide license

myk002 (approved):
wow, PlugLoad.cpp is remarkably clean now

ab9rf:
> wow, PlugLoad.cpp is remarkably clean now

i was frankly a bit surprised at how well it cleaned up with some judicious use of `#define`

PR #5138, opened 2024-12-26
reorganize/tidy `Process-*.cpp` source units

ab9rf:
Rearrange all three `Process-*.cpp` files so that the contents are in parallel as much as possible

remove `using namespace std` code smell
reorganize header includes
change minimum supported windows version to 6.0 (Vista)


ab9rf:
> FYI, this will probably conflict with #5093, although likely not too badly since you're not changing logic.

i was aware of that. i'm also not sure what the status of that PR is

myk002 (approved):

PR #5144, opened 2024-12-27
Improved Symbol Demangling

NicksWorld:
This set of changes adds support for symbol demangling via `cxx_demangle` on Windows and uses the improved symbol demangling in reading classnames from RTTI entries on all current targets.

Known problems:
- Wine environments fail on some advanced STL features.

When running a modified `devel/scan-vtables` for Windows under wine, a number of symbols fail demangling, all of which can be filtered out by excluding those belonging to std/Concurency. This can be automatically performed by checking if the end of the mangled name is `std@@` or `Concurrency@@`.

I have not discovered any further issues, but practical testing is likely a requirement.

Thank you to @ab9rf and @lethosor for their help with understanding particular issues that arose.

### DIFF
--- a/library/Process-linux.cpp
+++ b/library/Process-linux.cpp
@@ -40,6 +40,7 @@ distribution.
 #include "Internal.h"
 #include "MemAccess.h"
 #include "Memory.h"
+#include "MiscUtils.h"
 #include "VersionInfo.h"
 #include "VersionInfoFactory.h"
 #include "modules/Filesystem.h"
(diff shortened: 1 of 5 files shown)


library/Process-linux.cpp
```diff
@@ -120,13 +121,18 @@ Process::~Process()
 
 string Process::doReadClassName (void * vptr)
 {
-    //FIXME: BAD!!!!!
-    char * typeinfo = Process::readPtr(((char *)vptr - sizeof(void*)));
-    char * typestring = Process::readPtr(typeinfo + sizeof(void*));
+    char* typeinfo = Process::readPtr(((char *)vptr - sizeof(void*)));
```
myk002, line 124: I think ab9rf is in the process of unifying these files, and uses `diff` to identify the chunks that differ. Changing the formatting at this point would probably cause her pain.

library/Process-linux.cpp
```diff
@@ -120,13 +121,18 @@ Process::~Process()
...
+    char* typestring = Process::readPtr(typeinfo + sizeof(void*));
     string raw = readCString(typestring);
-    size_t start = raw.find_first_of("abcdefghijklmnopqrstuvwxyz");// trim numbers
-    size_t end = raw.length();
-    return raw.substr(start,end-start);
+
+    string status;
+    string demangled = cxx_demangle(raw, &status);
```
myk002, line 129: does `devel/scan-vtables` produce the same output as before (on Linux) with this change?

library/Process-linux.cpp
ab9rf: eh, don't worry about it. i can tell VSCode to ignore whitespace

library/Process-linux.cpp
NicksWorld: scan-vtables doesn't use doReadClassName, and instead uses cxx_demangle directly from lua. I haven't adjusted the linux implementation of that function.

library/Process-linux.cpp
NicksWorld: If desired I can just update styling to be consistent between Linux and Darwin as they are essentially equivalent.

ab9rf: > If desired I can just update styling to be consistent between Linux and Darwin as they are essentially equivalent.

Feel free to do as you see best here.

myk002 (approved):

PR #5152, opened 2024-12-29
Add a brief section explaining how to show stonesense on OBS without flickering

realSquidCoder:


docs/plugins/stonesense.rst
```diff
@@ -56,6 +56,16 @@ views, fog, and rotation. Here's the important section:
...
    :end-before: VALID ACTIONS:
 
+
+Streaming stonesense
+--------------------
+If you wish to stream stonesense thru OBS then you may find that opening
+stonesense causes your main DF window to flicker between DF and stonesense.
+While it is unknown exactly what causes this, a fix does exist. Simply make
+sure that both DF and stonesense are using ``Window Capture``s and NOT
+``Game Capture``s. This will stop the flickering from happening and enable
```
myk002, line 66: if you want text directly after quoted text, you have to escape it. e.g.
```md
``Game Capture``\s
```

However, I think this would read better without the "s"s at all -- ```
"... are using ``Window Capture`` and NOT ``Game Capture``.
```

I think having the period directly after it is ok and doesn't need to be escaped.

Comment (realSquidCoder) [reply docs/plugins/stonesense.rst:66]: ahh ok think for both. i think i shall remove the s

File: docs/plugins/stonesense.rst [hunk 2]
```diff
@@ -56,6 +56,16 @@ views, fog, and rotation. Here's the important section:
    :literal:
    :end-before: VALID ACTIONS:
 
+
+Streaming stonesense
+--------------------
+If you wish to stream stonesense thru OBS then you may find that opening
```
myk002, line 62: You should explain what OBS is and provide a link. It is not a given that readers will know what that is. Also, you should write more about which operating system environment(s) this applies to, and whether it applies to other screen capturing software as well.

realSquidCoder: how's it look now?

docs/plugins/stonesense.rst
```diff
@@ -56,6 +56,17 @@ views, fog, and rotation. Here's the important section:
...
    :end-before: VALID ACTIONS:
 
+
+Streaming stonesense
+--------------------
+If you wish to stream stonesense thru a broadcasting software such as `OBS <https://obsproject.com/>`_
+then you may find that opening stonesense causes your main DF window to flicker
+between DF and stonesense. While it is unknown exactly what causes this, a fix
+does exist. Simply make sure that both DF and stonesense are using ``Window Capture``
+and NOT ``Game Capture``. This will stop the flickering from happening and enable
```
myk002, line 66: Here are the capture options OBS gives me on Linux:
image

This section may be specific to Windows

docs/plugins/stonesense.rst
```diff
@@ -56,6 +56,17 @@ views, fog, and rotation. Here's the important section:
...
+you to stream stonesense for all to enjoy. This has been tested in OBS on Windows 10 but
+should work on all operating systems the OBS supports and `Streamlabs <https://streamlabs.com/>`_.
```
myk002, line 68: Is Streamlabs relevant, or this this purely an OBS capture problem?

docs/plugins/stonesense.rst
realSquidCoder: Do we know anyone who has a Mac to be sure?

docs/plugins/stonesense.rst
realSquidCoder: i know, at least on windows, streamlabs has a game capture option and i assume it has the same issue, have not tested tho

PR #5157, opened 2024-12-30
Remove CoreSuspender from interpose functions

myk002:
since it may end up deadlocking, depending on what's in the call chain


ab9rf:
I am fairly convinced that it is entirely unsafe, and inappropriate, to acquire a `CoreSuspender` from a vmethod interpose. Why are we acquiring a suspender in these situations in the first place?

myk002:
I had thought I found the CoreSuspender to be required in `overlay`, which interposes the viewscreen methods and calls Lua, but when I tested right now, I did not hit any assert.

Let me turn this PR into a "remove all interposed CoreSuspenders" change

ab9rf:
> I had thought I found the CoreSuspender to be required in `overlay`, which interposes the viewscreen methods and calls Lua, but when I tested right now, I did not hit any assert (and I double-checked that this is a non-Release build).

It is likely that I'll add some asserts to monitor for certain thread-safety violations, but this is outside the scope of this PR. We could add an assert-check for a vmethod interpose entered by a thread that doesn't hold the core mutex. This will trigger in the event that such a method is called by a DF threaded worker, but that's intelligence that we could probably stand to learn anyway.

I think we already have an assert for certain Lua operations without holding the core mutex; I have a PR that will move `Lua::Core::State` into `Core` and it would be fairly trivial after that to have the `Core::getLuaState` method that that PR will add include an assert for holding the core mutex as well.

ab9rf (approved):
lgtm - did not confirm that this covers all interposes, however

myk002:
I didn't do stonesense, but this should be everything in the main repo

PR #5159, opened 2024-12-31
Refactor core Lua state and add thread safety assertions

ab9rf:
This moves the global Lua state from a static global in LuaTools.h to a member variable of the singleton `Core`. A public accessor, `getLuaState`, is added to `Core` to provide access. This accessor includes an assertion which guarantees that only threads that hold the core mutex attempt to acquire the core lua state. Passing `true` to `getLuaState` will bypass the assertion; this should only be used in code that is guaranteed to acquire a `CoreSuspender` before trying to use the acquired Lua state.

### DIFF
--- a/library/include/Core.h
+++ b/library/include/Core.h
@@ -25,6 +25,7 @@ distribution.
 #pragma once
 
 #include "Console.h"
+#include "CoreDefs.h"
 #include "Export.h"
 #include "Hooks.h"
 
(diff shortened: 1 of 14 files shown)


library/include/Core.h
```diff
@@ -395,9 +379,11 @@ namespace DFHack
             if (!owns_lock())
                 return;
             /* Restore core owner to previous value */
-            core.ownerThread.store(tid, std::memory_order_release);
             if (tid == std::thread::id{})
                 Lua::Core::Reset(core.getConsole(), "suspend");
+            core.ownerThread.store(tid, std::memory_order_release);
+            //if (tid == std::thread::id{})
+            //    Lua::Core::Reset(core.getConsole(), "suspend");
```
myk002, line 386: leftover debugging?

myk002 (approved):

PR #5161, opened 2024-12-31
split plugin lua macros to separate header

ab9rf:
this will reduce the number of translation units that require the includes from the lua dependency

### DIFF
--- a/plugins/cxxrandom.cpp
+++ b/plugins/cxxrandom.cpp
@@ -25,11 +25,12 @@ Updated: Dec. 21 2017
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "Console.h"
 #include "Error.h"
+#include "Export.h"
 #include "DataFuncs.h"
-#include <Console.h>
-#include <Export.h>
-#include <PluginManager.h>
+#include "PluginManager.h"
+#include "PluginLua.h"
 
 using namespace DFHack;
 DFHACK_PLUGIN("cxxrandom");
--- a/plugins/cxxrandom.cpp
+++ b/plugins/cxxrandom.cpp
@@ -25,9 +25,6 @@ Updated: Dec. 21 2017
 #include <stdio.h>
 #include <stdlib.h>
 
-#include "Console.h"
-#include "Error.h"
-#include "Export.h"
 #include "DataFuncs.h"
 #include "PluginManager.h"
 #include "PluginLua.h"
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -68,6 +68,7 @@ set(MAIN_HEADERS
     include/Module.h
     include/MemAccess.h
     include/ModuleFactory.h
+    include/PluginLua.h
     include/PluginManager.h
     include/PluginStatics.h
     include/RemoteClient.h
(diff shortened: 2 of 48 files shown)


plugins/cxxrandom.cpp
```diff
@@ -25,11 +25,12 @@ Updated: Dec. 21 2017
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "Console.h"
 #include "Error.h"
+#include "Export.h"
```
myk002, line 30: Nothing should need to include Export.h, right? It's injected by the build system

myk002 (commented):
Don't forget to get rid of those lines I just added in `Plugins.cmake`:

```cmake
        # since PluginManager currently uses Lua headers, even when Lua is not used
        get_target_property(lua_INCLUDES lua INTERFACE_INCLUDE_DIRECTORIES)
        target_include_directories(${PLUGIN_NAME} PRIVATE ${lua_INCLUDES})
```

myk002:
Maybe also add a comment to examples/skeleton.cpp about which header files to include if the plugin makes use of Lua

plugins/cxxrandom.cpp
ab9rf: it was there before, i just reordered the headers

ab9rf: I tested it without `Export.h` and it compiled fine. On a lark I also took out `Console.h` (which is also unusual for plugins) and it still compiled, and on inspection it's not needed either so I'm removing both.

ab9rf:
> Don't forget to get rid of those lines I just added in `Plugins.cmake`:
> 
> ```cmake
>         # since PluginManager currently uses Lua headers, even when Lua is not used
>         get_target_property(lua_INCLUDES lua INTERFACE_INCLUDE_DIRECTORIES)
>         target_include_directories(${PLUGIN_NAME} PRIVATE ${lua_INCLUDES})
> ```

my base branch for this PR is before that was merged, so i either have to rebase, or cherry-pick that commit so i can reverse it. it'll be a bit

myk002:
you can also merge develop into this branch

ab9rf:
> you can also merge develop into this branch

i tried that but it made a mess of the PR

ab9rf:
need to add lua include to `add-spatter` because it uses interposes, which have a dependency on DataFuncs that probably isn't absolutely necessary but will need more work to unwind

`eventExample` also had problems, but there we can just remove the unnecessary headers, which included `VTableInterpose.h`, `Export.h`, and `Console.h`. removed

i can't figure out how to get `add-spatter` to compile with removing the changes above, so i'm backing that out for now, we can catch that subsequently

myk002:
> i can't figure out how to get `add-spatter` to compile with backing out the change in `Plugins.cmake`, so i'm backing that out for now, we can catch that subsequently

If it's just `add-spatter`, we can include Lua headers for just that build target in cmake by adding this to `plugins/CMakeLists.txt`:

```cmake
    if(BUILD_PLUGINS)
        # remove when VTableInterpose.h no longer has a dependency on DataFuncs.h
        get_target_property(lua_INCLUDES lua INTERFACE_INCLUDE_DIRECTORIES)
        target_include_directories(add-spatter PRIVATE ${lua_INCLUDES})
    endif()
```

can be added just after `dfhack_plugin(add-spatter add-spatter.cpp)`

myk002 (approved):

PR #5169, opened 2025-01-06
fix mysterious new errors in CI

myk002:
I'm not sure where these came from. the runner hasn't changed, and no errors are seen in my local builds


myk002:
This could be related to https://github.com/pygments/pygments/issues/2818

Issue #5166, opened 2025-01-03
Add dfhack.GenerateName API


myk002:
re-reverse engineer the name generation routine in v50 and refactor the name generation code from `strangemood` into a core library API function

myk002:
Once this is ready, the "Generate name" button in `gui/rename` can be uncommented and made available. The call to `dfhack.GenerateName` will also likely need updating.

PR #5175, opened 2025-01-08
add Translation::GenerateName()

myk002:
Also available from Lua as "dfhack.generateName(...)"

Fixes: #5166

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -953,6 +953,10 @@ can be omitted.
 
   Convert a ``df.language_name`` (or only the last name part) to string.
 
+* ``dfhack.GenerateName(name,language,type,major_selector,minor_selector)``
+
+  Dynamically generate a name using the same logic the game itself uses.
+
 * ``dfhack.df2utf(string)``
 
   Convert a string from DF's CP437 encoding to UTF-8.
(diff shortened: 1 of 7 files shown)


docs/dev/Lua API.rst
```diff
@@ -953,6 +953,10 @@ can be omitted.
 
   Convert a ``df.language_name`` (or only the last name part) to string.
 
+* ``dfhack.GenerateName(name,language,type,major_selector,minor_selector)``
```
myk002, line 956: Would it be feasible to provide an additional signature of `dfhack.GenerateName(name, language, type)` that chooses appropriate selectors based on the `language_name_type`? I can write the LuaApi glue if you write the cpp.

quietust: There is no single set of rules for how the major/minor selectors are chosen - sometimes they're taken from the global list, and sometimes some are taken from the entity definition.

myk002:
Could you resolve the changelist conflicts? I can't seem to push updates to your branch

myk002:
> > I've felt for a long time that we ought to move this stuff to a dfhack.translation module, to match C++
> > we could alias the old `dfhack.TranslateName = dfhack.translation.translateName` or something like that, and make the case in the new API consistent with the other APIs

captured in #5180

quietust:
Note that while I haven't tested the Lua bindings for this, I *have* confirmed that the function works correctly from C++ - the StrangeMood plugin successfully generated the artifact name "Mosusthilség", "Roomtarnished".

PR #5187, opened 2025-01-10
Update Stonesense docs (see description)

realSquidCoder:
Just a ton of various fixes to improve the docs since there's been a lot of changes and new issues found etc.


docs/plugins/stonesense.rst
```diff
@@ -78,6 +78,20 @@ line containing that folder, or :dffd:`use these smaller sprites <6096>`.
 Stonesense requires working graphics acceleration, and we recommend
 at least a dual core CPU to avoid slowing down your game of DF.
 
+If you are on Windows, you may find that using OpenGL runs better and more stable.
+Simply open :file:`dfhack-config/stonesense/init.txt` and set the renderer to ``[RENDERER:OPENGL]``
```
myk002, line 82: this should no longer matter

realSquidCoder: yep i was removing that right now

docs/plugins/stonesense.rst
```diff
@@ -78,10 +78,24 @@ line containing that folder, or :dffd:`use these smaller sprites <6096>`.
...
+
+.. figure:: ../images/stonesense-yellowcubes.png
+   :align: center
+
+   An example of the yellow cubes.
+
+If you would like to help us in fixing this, there are 2 things you can do:
+- Make an issue on Github with what item is missing and pictures of what it looks like in DF
+- Create the art yourself. For help with this, please see the wiki page on :wiki:`adding content <Utility:Stonesense/Adding_Content>`
+
```
myk002, line 93: can be removed, since this section is already included in #5207

realSquidCoder: done

docs/images/stonesense-yellowcubes.png
myk002, line 1: can also remove this file from this PR since it is included in #5207

docs/plugins/stonesense.rst
```diff
@@ -19,7 +19,7 @@ or be moved independently. Configuration for stonesense can be set in the
 ``dfhack-config/stonesense/init.txt`` file in your DF game directory. If the window refresh
 rate is too low, change ``SEGMENTSIZE_Z`` to ``2`` in this file, and if you are
 unable to see the edges of the map with the overlay active, try decreasing the
```
myk002, line 21: The overlay isn't a thing right now, and the reference to it might better be removed

docs/plugins/stonesense.rst
```diff
@@ -19,7 +19,7 @@ or be moved independently. Configuration for stonesense can be set in the
...
-value for ``SEGMENTSIZE_XY`` -- normal values are ``50`` to ``80``, depending
+value for ``SEGMENTSIZE_X`` and ``SEGMENTSIZE_Y`` -- normal values are ``50`` to ``80``, depending
```
myk002, line 22: would it be better to promote setting them to AUTO?

realSquidCoder: it would, but how should it be worded?

myk002: maybe we don't need to discuss the segment size here at all. Once we change the config file format, it will default to AUTO and nobody will need to care about the segment size

docs/plugins/stonesense.rst
realSquidCoder: done

docs/plugins/stonesense.rst
realSquidCoder: done

realSquidCoder:
@myk002 Can you look this over again, i dont think this needs to wait anymore for the segment size things and we can just use this as is. the segment stuff was all for the overlay anyways and now theres more useful things in here.

docs/plugins/stonesense.rst
```diff
@@ -75,6 +70,13 @@ If Stonesense gives an error saying that it can't load
 detail sprites used. Either open :file:`creatures/init.txt` and remove the
 line containing that folder, or :dffd:`use these smaller sprites <6096>`.
 
+Sometimes if you have opened Stonesense and then resize the DF window, DF will appear to be
+unresponsive. This bug is graphical only and if you hit :kbd:`Ctrl`:kbd:`Alt`:kbd:`S` and wait
+a minute or so (since you can't see when the game finishes saving) the game should quicksave.
+
+If you have Stonesense open in a fort and want to load a new fort, you MUST close Stonesense before
+loading the new fort or the game will crash.
```
myk002, line 78: this seems like something we should fix. why does it crash?

myk002 (approved):

docs/plugins/stonesense.rst
realSquidCoder: Because stonesense _**still**_ doesn't properly unload

PR #5198, opened 2025-01-13
Prevent deadlocks when unlocking a CoreSuspender

NicksWorld:
`dec_tool_count` was being called on already unlocked CoreSuspenders, causing the tool count to go negative. As the core only resumes when the tool count is 0, a deadlock occured. By returning early if the underlying lock isn't held, this issue is avoided.

### DIFF
--- a/library/include/Core.h
+++ b/library/include/Core.h
@@ -376,8 +376,6 @@ namespace DFHack
 
         void unlock()
         {
-            if (!owns_lock())
-                return;
             /* Restore core owner to previous value */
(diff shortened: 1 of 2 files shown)


library/include/Core.h
```diff
@@ -465,6 +465,8 @@ namespace DFHack
 
         void unlock()
         {
+            if (!owns_lock())
+                return;
```
ab9rf, line 469: i would put this in the destructor instead of here

ab9rf (approved):

PR #5201, opened 2025-01-15
Widgets/edit field to text area

wiktor-obrebski:
This PR migrate commonly used `EditField` widget to be based on `TextArea` component.
Its inherits all `TextArea` features in one line mode, keeping original interface of `EditField` widget.

Also, I polished one line mode of `TextArea` to display cp437 character (`◙`) instead of remove them.

screen

### DIFF
--- a/test/library/gui/widgets.EditField.lua
+++ b/test/library/gui/widgets.EditField.lua
@@ -4,6 +4,9 @@ local widgets = require('gui.widgets')
 
 function test.editfield_cursor()
     local e = widgets.EditField{}
+
+    -- cursor is normally set in `postUpdateLayout`, hard to test in unit tests
+    e:setCursor(1)
     e:setFocus(true)
     expect.eq(1, e.cursor, 'cursor should be after the empty string')
 
@@ -64,6 +67,8 @@ end
 
 function test.editfield_ignore_keys()
     local e = widgets.EditField{ignore_keys={'CUSTOM_B', 'CUSTOM_C'}}
+    -- cursor is normally set in `postUpdateLayout`, hard to test in unit tests
+    e:setCursor(1)
     e:setFocus(true)
 
     e:onInput{_STRING=string.byte('a'), CUSTOM_A=true}
--- a/test/library/gui/widgets.EditField.lua
+++ b/test/library/gui/widgets.EditField.lua
@@ -29,11 +29,11 @@ function test.editfield_cursor()
     e:onInput{CUSTOM_HOME=true}
     expect.eq(1, e.cursor, 'cursor should be at beginning of string')
     e:onInput{CUSTOM_CTRL_RIGHT=true}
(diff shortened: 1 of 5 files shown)


test/library/gui/widgets.EditField.lua
```diff
@@ -29,36 +32,43 @@ function test.editfield_cursor()
     e:onInput{CUSTOM_HOME=true}
     expect.eq(1, e.cursor, 'cursor should be at beginning of string')
     e:onInput{CUSTOM_CTRL_RIGHT=true}
-    expect.eq(6, e.cursor, 'goto beginning of next word')
+    expect.eq(5, e.cursor, 'goto end of current word')
```
myk002, line 35: I think I was mimicking readline behavior before, but I agree with this change since it matches browser edit field behavior.

myk002 (dismissed):
approved, but holding off on merge until after 50.15-r2 is tagged since we're so close to release

myk002:
From testing, there is an issue with focus setting in `gui/launcher`. If you do a history search (via Alt-s or by clicking on the search widget) and then right click or hit esc to go back to editing the commandline, no widget has focus and key input is lost. Are the `onunfocus` handlers not being run correctly?

Also, for activatable edit fields that you can click to set focus and start editing, like the blueprint name in `gui/blueprint`, the existing text will become highlighted from the end of the word to the place where you clicked. It would make more sense to highlight the entire string on activation or not highlight anything and ignore the first click, leaving the cursor at the end of the string.

However, if we highlight everything on focus, we should fix another (existing) behavior I noticed when I double clicked to highlight a word, then hit an arrow key. Currently, the cursor gets positioned relative to the *beginning* of the word when you do this. If a word is highlighted and I hit the left arrow, I'd expect the cursor to be at the beginning of the word, not one position to the left of the word. If a word is highlighted and I hit the right arrow key, I expect the cursor to be just after the last letter of the word, not one position to the right from the first letter of the word.

myk002:
Focus is fixed for `gui/launcher` and the partial-highlight-on-activate issue in `gui/blueprint` is fixed, but now I'm seeing a few new issues:

- in `gui/launcher` you can no longer iteratively search the history. Repro steps:
  - type something that will match multiple elements in your history
  - hit Alt-s to search. first match shows on the commandline
  - hit Alt-s again. in HEAD, the next matched commandline in the history will show. with this PR, the matched commandline doesn't change
- in `gui/blueprint`, the edit field contents can revert to an incorrect string. Repro steps:
  - start `gui/blueprint`
  - edit the name field to something unique and hit enter to commit
  - re-select the name field and cancel out
  - expected: the string does not change; actual: the string is cleared to the default initial value
  - I If you commit a new string again, then re-activate and cancel, you'll get the first string you entered up at step 2. I suspect the onfocus/onunfocus behavior is different from the original EditField such that `gui/blueprint` is not getting the events at the same time as before.

myk002 (approved):

PR #5203, opened 2025-01-15
Add the API code to know when the last save was.

realSquidCoder:
This code is supposed to note down what the millisecond time is on save or load and be able to return the difference from that time and now to give to `notifications.lua`

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -771,6 +771,10 @@ arbitrary Lua tables.
   Same semantics as for the ``Site`` functions, but will associated the data
   with the global world context.
 
+* ``dfhack.persistent.getUnsavedSecs()``
+
+  Returns the number of seconds since last save or load of a save
+
 The data is kept in memory, so no I/O occurs when getting or saving keys. It is
 all written to a json file in the game save directory when the game is saved.
 
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -771,9 +771,9 @@ arbitrary Lua tables.
   Same semantics as for the ``Site`` functions, but will associated the data
   with the global world context.
 
-* ``dfhack.persistent.getUnsavedSecs()``
+* ``dfhack.persistent.getUnsavedSeconds()``
 
-  Returns the number of seconds since last save or load of a save
+  Returns the number of seconds since last save or load of a save.
 
 The data is kept in memory, so no I/O occurs when getting or saving keys. It is
 all written to a json file in the game save directory when the game is saved.
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -365,7 +365,7 @@ static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getWorldDataString", dfhack_persistent_get_world_data_string },
     { "saveWorldDataString", dfhack_persistent_save_world_data_string },
     { "deleteWorldData", dfhack_persistent_delete_world_data },
-    { "getUnsavedSecs", dfhack_persistent_get_last_save_time },
+    { "getUnsavedSeconds", dfhack_persistent_get_last_save_time },
     { NULL, NULL }
 };
 
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -353,7 +353,7 @@ static int dfhack_persistent_delete_world_data(lua_State *L) {
     return delete_site_data(L, get_world_data);
 }
 
-static int dfhack_persistent_get_last_save_time(lua_State *L) {
+static int dfhack_persistent_get_unsaved_seconds(lua_State *L) {
     lua_pushinteger(L, Persistence::getUnsavedSeconds());
     return 1;
 }
@@ -365,7 +365,7 @@ static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getWorldDataString", dfhack_persistent_get_world_data_string },
     { "saveWorldDataString", dfhack_persistent_save_world_data_string },
     { "deleteWorldData", dfhack_persistent_delete_world_data },
-    { "getUnsavedSeconds", dfhack_persistent_get_last_save_time },
+    { "getUnsavedSeconds", dfhack_persistent_get_unsaved_seconds },
     { NULL, NULL }
 };
 
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -353,6 +353,11 @@ static int dfhack_persistent_delete_world_data(lua_State *L) {
     return delete_site_data(L, get_world_data);
 }
 
+static int dfhack_persistent_get_last_save_time(lua_State *L) {
+    lua_pushinteger(L, Persistence::getUnsavedSeconds());
+    return 1;
+}
+
 static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getSiteDataString", dfhack_persistent_get_site_data_string },
     { "saveSiteDataString", dfhack_persistent_save_site_data_string },
@@ -360,6 +365,7 @@ static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getWorldDataString", dfhack_persistent_get_world_data_string },
     { "saveWorldDataString", dfhack_persistent_save_world_data_string },
     { "deleteWorldData", dfhack_persistent_delete_world_data },
+    { "getUnsavedSecs", dfhack_persistent_get_last_save_time },
     { NULL, NULL }
 };
 
@@ -1343,7 +1349,7 @@ static string getArchitectureName()
 
 static string getDFVersion() { return Core::getInstance().vinfo->getVersion(); }
 static uint32_t getTickCount() { return Core::getInstance().p->getTickCount(); }
-static uint32_t getCurSaveDur() { return Persistence::getSaveDur(); }
+static uint32_t getCurSaveDur() { return Persistence::getUnsavedSeconds(); }
 
 static string getDFPath() { return Core::getInstance().p->getPath(); }
 static string getHackPath() { return Core::getInstance().getHackPath(); }
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -1349,7 +1349,6 @@ static string getArchitectureName()
 
 static string getDFVersion() { return Core::getInstance().vinfo->getVersion(); }
 static uint32_t getTickCount() { return Core::getInstance().p->getTickCount(); }
-static uint32_t getCurSaveDur() { return Persistence::getUnsavedSeconds(); }
 
 static string getDFPath() { return Core::getInstance().p->getPath(); }
 static string getHackPath() { return Core::getInstance().getHackPath(); }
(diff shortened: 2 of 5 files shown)


library/LuaApi.cpp
```diff
@@ -1370,6 +1371,7 @@ static const LuaWrapper::FunctionReg dfhack_module[] = {
     WRAP(getDFVersion),
     WRAP(getDFPath),
     WRAP(getTickCount),
+    WRAP(getCurSaveDur),
```
myk002, line 1374: Please add this to the `persistent` API (apparently we called it "persistent" instead of "persistence"). Code is on line 356 of this file (search for `dfhack_persistent_funcs`).

docs go here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#persistent-configuration-storage

library/include/modules/Persistence.h
```diff
@@ -213,5 +213,7 @@ namespace DFHack
         // Fills the vector with references to each persistent item with a key that is
         // equal to the given key.
         DFHACK_EXPORT void getAllByKey(std::vector<PersistentDataItem> &vec, int entity_id, const std::string &key);
+        // Returns the tickcount of the most recent save.
+        DFHACK_EXPORT uint32_t getSaveDur();
```
myk002, line 217: I don't think `getSaveDur` is an appropriate name. ~~How about `getLastLoadSaveTickCount()`? It's long, but it's also much more specific. The name needs to communicate that it is updated on load or save, and that it talks in the same units as getTickCount().~~ see comment in Persistence.cpp

This information should also be in the documentation in `Lua API.rst`.

library/modules/Persistence.cpp
```diff
@@ -49,6 +49,8 @@ using namespace DFHack;
 static std::unordered_map<int, std::multimap<std::string, std::shared_ptr<Persistence::DataEntry>>> store;
 static std::unordered_map<size_t, std::shared_ptr<Persistence::DataEntry>> entry_cache;
 
+static uint32_t lastTickCount = 0;
```
myk002, line 52: `lastTickCount` is not specific enough. How about `lastLoadSaveTickCount`?

library/modules/Persistence.cpp
```diff
@@ -198,6 +200,7 @@ void Persistence::Internal::save(color_ostream& out) {
     CoreSuspender suspend;
 
     // write status
+    lastTickCount = Core::getInstance().p->getTickCount();
```
myk002, line 203: i'd put this at the *end* of the function so that as much as possible of the save logic itself is not included

library/modules/Persistence.cpp
```diff
@@ -307,6 +310,7 @@ void Persistence::Internal::load(color_ostream& out) {
     }
 
     if (found)
+        lastTickCount = Core::getInstance().p->getTickCount();
```
myk002, line 313: you want to update this regardless of whether DFHack state is found

library/modules/Persistence.cpp
```diff
@@ -417,3 +421,8 @@ void Persistence::getAllByKey(std::vector<PersistentDataItem> &vec, int entity_i
     for (auto it = range.first; it != range.second; ++it)
         vec.emplace_back(it->second);
 }
+
+uint32_t Persistence::getSaveDur() {
```
myk002, line 425: Oh, I see, this returns the number of seconds since the last load or save. That was not clear from the name. How about `getUnsavedSeconds`?

realSquidCoder:
unsure if i missed anything besides docs

library/modules/Persistence.cpp
realSquidCoder: done

library/modules/Persistence.cpp
realSquidCoder: fixed

library/modules/Persistence.cpp
realSquidCoder: makes sense, done.

library/modules/Persistence.cpp
realSquidCoder: updated the name, i couldnt think of a good one

library/LuaApi.cpp
```diff
@@ -353,13 +353,19 @@ static int dfhack_persistent_delete_world_data(lua_State *L) {
...
+}
+
 static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getSiteDataString", dfhack_persistent_get_site_data_string },
     { "saveSiteDataString", dfhack_persistent_save_site_data_string },
     { "deleteSiteData", dfhack_persistent_delete_site_data },
     { "getWorldDataString", dfhack_persistent_get_world_data_string },
     { "saveWorldDataString", dfhack_persistent_save_world_data_string },
     { "deleteWorldData", dfhack_persistent_delete_world_data },
+    { "getUnsavedSecs", dfhack_persistent_get_last_save_time },
```
myk002, line 368:
```suggestion
    { "getUnsavedSeconds", dfhack_persistent_get_last_save_time },
```

myk002: the Lua API should match the C++ API where possible

library/LuaApi.cpp
```diff
@@ -1343,6 +1349,7 @@ static string getArchitectureName()
 
 static string getDFVersion() { return Core::getInstance().vinfo->getVersion(); }
 static uint32_t getTickCount() { return Core::getInstance().p->getTickCount(); }
+static uint32_t getCurSaveDur() { return Persistence::getUnsavedSeconds(); }
```
myk002, line 1352: can remove this

library/LuaApi.cpp
myk002: you can remove this line now

library/include/modules/Persistence.h
```diff
@@ -213,5 +213,7 @@ namespace DFHack
         // Fills the vector with references to each persistent item with a key that is
         // equal to the given key.
         DFHACK_EXPORT void getAllByKey(std::vector<PersistentDataItem> &vec, int entity_id, const std::string &key);
+        // Returns the tickcount of the most recent save.
```
myk002, line 216:
```suggestion
        // Returns the number of seconds since the current savegame was saved or loaded.
```

library/modules/Persistence.cpp
```diff
@@ -237,6 +239,7 @@ void Persistence::Internal::save(color_ostream& out) {
         color_ostream_wrapper wrapper(file);
         Lua::CallLuaModuleFunction(wrapper, "script-manager", "print_timers");
     }
+    lastLoadSaveTickCount = Core::getInstance().p->getTickCount();
```
myk002, line 242:
```suggestion

    lastLoadSaveTickCount = Core::getInstance().p->getTickCount();
```

myk002: since this is not part of the preceding logical block

library/LuaApi.cpp
realSquidCoder: added the docs

library/modules/Persistence.cpp
```diff
@@ -306,6 +309,7 @@ void Persistence::Internal::load(color_ostream& out) {
             out.printerr("Cannot load data from: '%s'\n", path.c_str());
     }
 
+    lastLoadSaveTickCount = Core::getInstance().p->getTickCount();
```
myk002, line 312: There are three points of return in this function, and you'll want to update the state regardless of how you return from this function.

I think what you want here is a wrapper class that updates the state on destruction.

```
struct LastLoadSaveTickCountUpdater {
    ~LastLoadSaveTickCountUpdater() {
        lastLoadSaveTickCount = Core::getInstance().p->getTickCount();
    }
}
```

Then at the top of this function, near `CoreSuspender suspend;`, add a `LastLoadSaveTickCountUpdater tickCountUpdater;`

You can use this for `save` as well to cut down on code duplication.

library/modules/Persistence.cpp
```diff
@@ -417,3 +421,8 @@ void Persistence::getAllByKey(std::vector<PersistentDataItem> &vec, int entity_i
     for (auto it = range.first; it != range.second; ++it)
         vec.emplace_back(it->second);
 }
+
+uint32_t Persistence::getUnsavedSeconds() {
+    uint32_t durMS =  Core::getInstance().p->getTickCount() - lastLoadSaveTickCount;
+    return durMS / (1000);
```
myk002, line 427:
```suggestion
    return durMS / 1000;
```

myk002 (commented):
docs go in Lua API.rst

library/modules/Persistence.cpp
realSquidCoder: where do i put this wrapper? this is a new concept for me

docs/dev/Lua API.rst
```diff
@@ -771,6 +771,10 @@ arbitrary Lua tables.
   Same semantics as for the ``Site`` functions, but will associated the data
   with the global world context.
 
+* ``dfhack.persistent.getUnsavedSecs()``
```
myk002, line 774:
```suggestion
* ``dfhack.persistent.getUnsavedSeconds()``
```

docs/dev/Lua API.rst
```diff
@@ -771,6 +771,10 @@ arbitrary Lua tables.
...
+
+  Returns the number of seconds since last save or load of a save
```
myk002, line 776:
```suggestion
  Returns the number of seconds since last save or load of a save.
```

library/LuaApi.cpp
```diff
@@ -353,13 +353,19 @@ static int dfhack_persistent_delete_world_data(lua_State *L) {
     return delete_site_data(L, get_world_data);
 }
 
+static int dfhack_persistent_get_last_save_time(lua_State *L) {
```
myk002, line 356:
```suggestion
static int dfhack_persistent_get_unsaved_seconds(lua_State *L) {
```

library/LuaApi.cpp
```diff
@@ -353,13 +353,19 @@ static int dfhack_persistent_delete_world_data(lua_State *L) {
...
+}
+
 static const luaL_Reg dfhack_persistent_funcs[] = {
     { "getSiteDataString", dfhack_persistent_get_site_data_string },
     { "saveSiteDataString", dfhack_persistent_save_site_data_string },
     { "deleteSiteData", dfhack_persistent_delete_site_data },
     { "getWorldDataString", dfhack_persistent_get_world_data_string },
     { "saveWorldDataString", dfhack_persistent_save_world_data_string },
     { "deleteWorldData", dfhack_persistent_delete_world_data },
+    { "getUnsavedSeconds", dfhack_persistent_get_last_save_time },
```
myk002, line 368:
```suggestion
    { "getUnsavedSeconds", dfhack_persistent_get_unsaved_seconds },
```

library/LuaApi.cpp
```diff
@@ -1370,6 +1377,7 @@ static const LuaWrapper::FunctionReg dfhack_module[] = {
     WRAP(getDFVersion),
     WRAP(getDFPath),
     WRAP(getTickCount),
+    WRAP(getCurSaveDur),
```
myk002, line 1380: can remove this

myk002 (commented):
code looks good! all it needs now is changelog entries (API and Lua sections)

myk002 (approved):

PR #5208, opened 2025-01-20
New Widget: Slider

realSquidCoder:
This is just like Range_Slider except that it has only one slider bar.

### DIFF
--- a/library/lua/gui/widgets/slider.lua
+++ b/library/lua/gui/widgets/slider.lua
@@ -1,0 +1,132 @@
+local Widget = require('gui.widgets.widget')
+
+local to_pen = dfhack.pen.parse
+
+--------------------------------
+-- Slider
+--------------------------------
+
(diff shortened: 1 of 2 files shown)


library/lua/gui/widgets/slider.lua
myk002, line 1: This shares a lot of data and logic with RangeSlider. Please refactor to avoid duplication. You can make shared logic non-`local` and refer to it from both files.

myk002 (commented):
don't forget a changelog entry (in the Lua section)!

myk002 (commented):
this is cleaner, but not quite clean enough. The RangeSlider attributes are still visible in Slider, but are useless, which leads to a confusing API.

I think both RangeSlider and Slider should inherit from a base class that holds the common logic, keeping the externally visible ATTRS clean for both classes.

ab9rf (approved):
I don't see any obvious problems and this code doesn't appear to break existing functionality in testing

Issue #4994, opened 2024-10-08
Question: Is Quickfort / Blueprint copy zones expected anytime soon?


adiamaq:
I was wondering if we could copy and paste zones (especially for filling pond orders, as setting the whole pond with a single zone will create only one task for one dwarf, which makes the water dry before the second bucket). I saw that the old version allowed copying zones, but this is not implemented yet in version 50. Is there any estimated time for including zones with blueprints/quickfort? Thanks in advance!

PR #5222, opened 2025-01-27
Bring back zones in `gui/blueprint` (WIP)

realSquidCoder:
ref: https://github.com/DFHack/dfhack/issues/4994


plugins/blueprint.cpp
```diff
@@ -1107,78 +1104,48 @@ static const char * get_tile_place(const df::coord &pos,
...
-    return a.max_thread == b.max_thread &&
-            a.max_cloth == b.max_cloth &&
-            a.max_splints == b.max_splints &&
-            a.max_crutches == b.max_crutches &&
-            a.max_plaster == b.max_plaster &&
-            a.max_buckets == b.max_buckets &&
-            a.max_soap == b.max_soap;
-}
-
 static const char * get_zone_keys(const df::building_civzonest *zone) {
```
myk002, line 1107: This function needs to be rewritten to support the blueprint format for zones and locations. The docs are here: https://docs.dfhack.org/en/latest/docs/guides/quickfort-user-guide.html#zone-mode and here: https://docs.dfhack.org/en/latest/docs/guides/quickfort-user-guide.html#zone-mode-reference

myk002 (commented):
a good way to test this is to set up zones of each type next to each other, take a blueprint of them, and apply the blueprint with `gui/quickfort`. Do the `quickfort`-created zones look the same as the originals?

Also, create zones and add taverns/temples/guildhalls/hospitals. When you blueprint a zone with an attached location, do you successfully create a new location with the same configuration as the original?

Create two meeting zones and attach the same tavern hall to both of them. When you blueprint the pair, does `quickfort` correctly create two identical zones, with both attached to the same new tavern (which should have the same configuration as the original tavern).

myk002:
This isn't quite correct, but I have a follow-up PR that will fill in the holes

PR #5230, opened 2025-01-31
Support adventure mode look cursor

Bumber64:
Make the fns in `gui/dwarfmode.lua` modify the adv cursor if in adv mode.
Do the same for `Gui::getCursorPos`, `Gui::getCursorCoords`, and `Gui::setCursorCoords`.
Clean up some direct uses of the cursor.


Bumber64:
A list of scripts that should be potentially be using the `guidm` fns instead:
```
advfort.lua
blueprint.lua
colonies.lua
companion-order.lua
create-unit.lua
hfs-pit.lua
launch.lua
light.lua
position.lua
putontable.lua
query.lua
source.lua
stripcaged.lua
teleport.lua
tiletypes.lua
tree-info.lua
```

`toggle-kbd-cursor.lua` can use `guidm.clearCursorPos()` instead of `guidm.setCursorPos()` in one spot.

library/modules/Gui.cpp
```diff
@@ -2911,56 +2923,91 @@ bool Gui::inRenameBuilding()
...
+            auto &look = game->main_interface.adventure.look;
+            if (look.open)
+                p = look.cursor;
+        }
+    }
+    else if (cursor)
+        p = df::coord(cursor->x, cursor->y, cursor->z);
+
+    pos = p;
+    return p.isValid();
```
myk002, line 2980:
```suggestion
    int32_t x, y, z;
    bool ret = getCursorCoords(x, y, z);
    pos.x = x;
    pos.y = y;
    pos.z = z;
    return ret;
```

myk002: to avoid code duplication?

Bumber64: I think `getCursorPos` works better, since the `int32_t` is lost anyway.

library/modules/Gui.cpp
```diff
@@ -2912,56 +2924,77 @@ bool Gui::inRenameBuilding()
...
+            auto &look = game->main_interface.adventure.look;
+            if (look.open)
+                p = look.cursor;
+        }
+        x = p.x; y = p.y; z = p.z;
+        return p.isValid();
+    }
+
+    x = cursor->x; y = cursor->y; z = cursor->z;
     return has_cursor();
```
Bumber64, line 2961: `has_cursor()` is equivalent to `getCursorPos().isValid()`, which means it's grabbing the cursor a second time. Is it sensible to do `x >= 0` instead? Or is better to make sure we're relying on `isValid()` since there's not likely a performance concern?

docs/changelog.txt
```diff
@@ -58,6 +57,7 @@ Template for new versions:
 
 ## Fixes
 - `gui/launcher`: ensure commandline is fully visible when searching through history and switching from a very long command to a short command
+- `createitem`: output items will now end up at look cursor if active
```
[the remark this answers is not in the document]
Bumber64: Sure. I'll handle it with the other scripts.

library/modules/Gui.cpp
myk002: I'd say rely on isValid so there's only one place to change the condition if it changes (we have already changed the condition once, since we used to compare against -30000 for validity, but that value is no longer stable)

docs/changelog.txt
```diff
@@ -62,6 +59,8 @@ Template for new versions:
 - `spectate`: new global keybinding for toggling spectate mode: Ctrl-Shift-S
 
 ## Fixes
+- `gui/launcher`: ensure commandline is fully visible when searching through history and switching from a very long command to a short command
```
myk002, line 62:
```suggestion
```

library/lua/gui/dwarfmode.lua
myk002, line 1: someday, we should migrate the contents of this file to `gui.lua`. It lost most of its functionality in the move to v50, and the remaining logic seems as applicable to adventure mode as it is to dwarfmode

myk002 (commented):
I think more changelog entries are warranted in the Lua section about adventure mode cursor support

library/lua/gui/dwarfmode.lua
Bumber64: Now might be a good time, given all the structure changes.

myk002 (approved):

PR #5256, opened 2025-02-07
Add function to remove unit from squad

ong-yinggao98:
Originally intended to be used in a now-cancelled Lua script feature, but figured it may be useful for those dabbling in military matters.

### DIFF
--- a/library/modules/Military.cpp
+++ b/library/modules/Military.cpp
@@ -296,7 +296,7 @@ void Military::updateRoomAssignments(int32_t squad_id, int32_t civzone_id, df::s
     }
 }
 
-static bool remove_soldier_entity_link(df::historical_figure* hf, df::squad* squad)
+static void remove_soldier_entity_link(df::historical_figure* hf, df::squad* squad)
 {
     int32_t start_year = -1;
     for (size_t i = 0; i < hf->entity_links.size(); i++)
@@ -317,7 +317,7 @@ static bool remove_soldier_entity_link(df::historical_figure* hf, df::squad* squ
     }
 
     if (start_year == -1)
-        return false;
+        return;
 
     auto former_squad = df::allocate<df::histfig_entity_link_former_squadst>();
     former_squad->squad_id = squad->id;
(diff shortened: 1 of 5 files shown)


myk002 (commented):
Could you expose this function to Lua (in `LuaAPI.cpp`) and add documentation in `Lua API.rst`?

library/modules/Military.cpp
```diff
@@ -289,3 +295,116 @@ void Military::updateRoomAssignments(int32_t squad_id, int32_t civzone_id, df::s
...
+        return false;
+
+    int32_t squad_id = unit->military.squad_id;
+    df::squad* squad = df::squad::find(squad_id);
+    if (squad == nullptr)
+        return false;
+
+    // remove from squad information
+    int32_t squad_pos = unit->military.squad_position;
+    df::squad_position* pos = squad->positions.at(squad_pos);
```
myk002, line 396: `at` can crash if given bad data. use `vector_get`, which is a safe wrapper. Be sure to test the return for null

library/modules/Military.cpp
```diff
@@ -289,3 +295,116 @@ void Military::updateRoomAssignments(int32_t squad_id, int32_t civzone_id, df::s
         }
     }
 }
+
+static bool remove_soldier_entity_link(df::historical_figure* hf, df::squad* squad)
+{
+    int32_t start_year = -1;
+    for (size_t i = 0; i < hf->entity_links.size(); i++)
+    {
+        auto link = strict_virtual_cast<df::histfig_entity_link_squadst>(hf->entity_links[i]);
```
myk002, line 304: `strict_virtual_cast` can be slow. it's good to guard it with a check of  `link->getType()`

myk002 (commented):
LGTM -- should I remove the "draft" PR status?

ong-yinggao98:
> LGTM -- should I remove the "draft" PR status?

Sure, thank you!

myk002:
> Do you think that the function should return a value?

That's a good point. I think that it should return false if the unit is not in a squad or other preconditions fail. I think it should return true if it has updated all the state that is feasible to update. If some dependent data structure cannot be found, I think it can skip it and still return true.

myk002 (approved):

PR #5269, opened 2025-02-11
`Military::removeFromSquad`: Add missing history event when unassigning officer

ong-yinggao98:
Minor and virtually cosmetic change for consistency's sake.

### DIFF
--- a/library/modules/Military.cpp
+++ b/library/modules/Military.cpp
@@ -14,6 +14,7 @@
 #include "df/histfig_entity_link_squadst.h"
 #include "df/historical_figure.h"
 #include "df/historical_entity.h"
+#include "df/history_event_remove_hf_entity_linkst.h"
 #include "df/entity_position.h"
 #include "df/entity_position_assignment.h"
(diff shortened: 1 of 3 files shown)


library/modules/Military.cpp
```diff
@@ -389,6 +394,19 @@ static void remove_officer_entity_link(df::historical_figure* hf, df::squad* squ
     former_pos->link_strength = 100;
 
     hf->entity_links.push_back(former_pos);
+
+    int32_t event_id = *df::global::hist_event_next_id;
```
myk002, line 398:
```suggestion
    int32_t event_id = (*df::global::hist_event_next_id)++;
```

myk002: this should work, allowing you to remove the explicit increment below

ong-yinggao98: Yep, I tested that earlier before publishing this but could not tell which would be more readable. if you prefer it I will do just that.

myk002 (approved):

PR #5277, opened 2025-02-16
Binpatch to disable vanilla dimensions

NicksWorld:
This adds a new binpatch for all current versions to disable the new dimension tooltip.

The patch should be safe as it replaces only the relevant call to `graphicst::addst` from libdfgraphics that is implemented [here](https://github.com/Putnam3145/Dwarf-Fortress--libgraphics--/blob/c295a18d89c50b765b25b6c8d6f2cef44dc081aa/g_src/graphics.cpp#L340) with nop opcodes. I have applied each patch on their relevant executable and have found them to be effective.


docs/changelog.txt
```diff
@@ -58,6 +58,7 @@ Template for new versions:
 - `spectate`: new "cinematic-action" mode that dynamically speeds up perspective switches based on intensity of conflict
 - `spectate`: new global keybinding for toggling spectate mode: Ctrl-Shift-S
 - `spectate`: when spectate mode is enabled, left/right arrow will cycle through following next/prevous units
+- disable-vanilla-dimensions: new binpatch to disable the vanilla dimensions tooltip
```
myk002, line 61:
```suggestion
- ``disable-vanilla-dimensions``: new binpatch to disable the vanilla dimensions tooltip
```

myk002 (approved):
/patches or /data/patches? @lethosor do you have an opinion? otherwise LGTM

myk002:
Sorry, I was asking about the source path, not the install path

PR #5293, opened 2025-02-21
spectate.lua: make config public

TymurGubayev:
Also improve error messages when missing command arguments

### DIFF
--- a/plugins/lua/spectate.lua
+++ b/plugins/lua/spectate.lua
@@ -84,7 +84,8 @@ local function load_state()
            function() config_file:write() end
 end
 
-local config, save_state = load_state()
+-- global variables to make the UI script simpler
+_ENV.config, _ENV.save_state = load_state()
 
 function refresh_cpp_config()
     for name,value in pairs(config) do
@@ -158,11 +159,15 @@ local function do_toggle()
 end
 
 local function set_setting(args)
+    local n = #args
+    if n == 0 then
+        qerror('missing key')
+    end
     local key = table.remove(args, 1)
     if config[key] == nil then
         qerror('unknown setting: ' .. key)
     end
-    local n = #args
+    n = #args
     if n == 0 then
         qerror('missing value')
     end
@@ -221,6 +226,7 @@ function parse_commandline(args)
     elseif command == 'set' then
         set_setting(args)
     elseif command == 'overlay' then
+        if #args == 0 then qerror('missing option') end
         set_overlay(args[1])
     else
         return false


TymurGubayev:
`argparse.boolean(nil)` should probably still do something more user-friendly than `bad argument #1 to 'lower' (string expected, got nil)`

plugins/lua/spectate.lua
```diff
@@ -84,7 +84,8 @@ local function load_state()
            function() config_file:write() end
 end
 
-local config, save_state = load_state()
+-- global variables to make the UI script simpler
+_ENV.config, _ENV.save_state = load_state()
```
myk002, line 88: No need for the _ENV prefix. Just non local is equivalent.

TymurGubayev: I know, but this way I'm signaling this is intended as a public variable.
I'd also get a "Global variable in lowercase initial" diagnostic:

image

(which could be resolved by renaming, but the signaling issue remains)

TymurGubayev: (and using `local M = {}` for the module table would make public declarations just two characters - `M.` - instead of 5 - `_ENV.`)

TymurGubayev: (but that would depend on the exact version of Lua and what `mkmodule` does, because we definitely do want to have a local environment table)

myk002: Alternately, could add public functions that return the local vars, if that's cleaner

myk002, line 88:
```suggestion
local config, save_state = load_state()

-- called by gui/spectate
function get_config_elem(name, key)
    local elem = config[name]
    if not elem then return end
    if type(elem) == 'table' then
        return elem[key]
    end
    return elem
end

```

myk002 (approved):

myk002:
> `argparse.boolean(nil)` should probably still do something more user-friendly than `bad argument #1 to 'lower' (string expected, got nil)`

done in: https://github.com/DFHack/dfhack/pull/5297

plugins/lua/spectate.lua
TymurGubayev: both `config` and `save_state` are required though :)

how about leaving them local here, but at the end of the file making them public explicitly via `_ENV.config, _ENV.save_state = config, save_state`?

TymurGubayev: also, now a `set_config_elem` is required :)

PR #5295, opened 2025-02-21
flush cache after mempatch

ab9rf:
This adds code to flush cache after applying a hotpatch. While it's unlikely that a patched location will be in cache at the time we apply a patch, it is better to be safe than sorry

Also fixed the spelling error in `setPermissions` while I was here

No changelog; this is a fix to existing functionality to avoid a potential future bug instead of fixing a known existing one

### DIFF
--- a/library/Process-linux.cpp
+++ b/library/Process-linux.cpp
@@ -32,7 +32,6 @@ distribution.
 
 #include <dirent.h>
 #include <errno.h>
-#include <sys/cachectl.h>
 #include <sys/mman.h>
 #include <sys/time.h>
 #include <unistd.h>
@@ -220,7 +219,7 @@ bool Process::setPermissions(const t_memrange & range,const t_memrange &trgrange
 
 bool Process::flushCache(const void* target, size_t count)
 {
-    return cacheflush(target, count, BCACHE);
+    __builtin___clear_cache((char*)target, (char*)target + count - 1);
 }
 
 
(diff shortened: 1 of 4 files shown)


ab9rf:
side comment: i did not implement this for MacOS

myk002 (approved):

myk002:
sys/cachectl.h: No such file or directory
   35 | #include <sys/cachectl.h>
      |          ^~~~~~~~~~~~~~~~

ab9rf:
> sys/cachectl.h: No such file or directory 35 | #include <sys/cachectl.h> | ^~~~~~~~~~~~~~~~

any idea what it should be? that's what the documentation for `cacheflush` indicates is the correct header

library/Process-linux.cpp
```diff
@@ -220,7 +219,8 @@ bool Process::setPermissions(const t_memrange & range,const t_memrange &trgrange
 
 bool Process::flushCache(const void* target, size_t count)
 {
-    return cacheflush(target, count, BCACHE);
+    __builtin___clear_cache((char*)target, (char*)target + count - 1);
+    return false; /* assume always succeeds, as the builtin has no return type */
```
myk002, line 223: Just verifying, you return false for success?

ab9rf: yeah, following the unix standard of 0 (false) for success and non-zero for failure. this is used elsewhere in DFHack for OS API calls (see, e.g. `PlugLoad`)

ab9rf: let me be clear that i don't have a problem with returning true for success :)

myk002: It seems to me that returning 0 for success makes sense when the return value is an int, but if it's a bool, then it should return true for success. If this matches the semantics of the surrounding functions and is consistent across the different implementations in Process-linux and Process-windows, though, I'm fine with it.

ab9rf: switched, since tbh i prefer returning `true` on success as well

myk002 (approved):

PR #5298, opened 2025-02-22
spectate.lua: alternative way to make config public

TymurGubayev:


TymurGubayev:
another alternative: public `get_config()` and `get_save_state()`

myk002:
I refactored it to do read-only lookups of plugin state and to use the public API for setting the values. That way the cpp part of the plugin stays synced with the Lua

TymurGubayev:
I see you are now doing `dfhack.run_command('spectate', 'set', cfg_elem, text)`. But `parse_commandline` is already public and would have the same effect(*). Using it would avoid one layer of indirection.

Also, to be really really sure, we should read the config state on the UI side after `dfhack.run_command/parse_commandline` is executed, in case it is rejected for whatever reason, and update the UI accordingly.

That said, this is all not important. Please feel free to close this PR.

(*) if there are exceptions depends on what `dfhack.run_command` does, but that could be solved by adding a `pcall` on the UI script side.

PR #5300, opened 2025-02-22
Fix spectate.lua

realSquidCoder:
fixed the hitbox of the follow panel overlay, as it was still the old `enable/disabled` hitbox


plugins/lua/spectate.lua
```diff
@@ -590,7 +590,7 @@ function FollowPanelOverlay:init()
             on_click=spectate_followNext,
         },
         widgets.Label{
-            frame={l=10, t=0, w=25},
+            frame={l=10, t=0, w=15},
```
myk002:
```suggestion
            frame={l=10, t=0, w=14},
```

myk002 (approved):

PR #5308, opened 2025-02-26
spectate.lua: minor simplification of `render_unit_banners()`

TymurGubayev:
Assigning `writer` back after using its methods isn't necessary.

### DIFF
--- a/plugins/lua/spectate.lua
+++ b/plugins/lua/spectate.lua
@@ -504,7 +504,6 @@ function TooltipOverlay:render_unit_banners(dc)
 
             ix = ix + len
         end
-        writer:pen(pen) -- just in case a different dc isn't resetting it
 
         -- mark coordinates as used
         for j = 0, #str - 1 do
--- a/plugins/lua/spectate.lua
+++ b/plugins/lua/spectate.lua
@@ -482,10 +482,10 @@ function TooltipOverlay:render_unit_banners(dc)
         for _, tok in ipairs(info) do
             local s
             if type(tok) == "string" then
-                writer = writer:pen(pen)
+                writer:pen(pen)
                 s = tok
             else
-                writer = writer:pen(tok.pen)
+                writer:pen(tok.pen)
                 s = tok.text
             end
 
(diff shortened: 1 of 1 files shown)


ChrisJohnsen:
The background color bleeding comes from intra-frame Painter re-use in `plugins/lua/overlay.lua` `render_viewscreen_widgets` and `_render_viewscreen_widgets`.

myk mentioned that this [might be addressed by resetting the Painter][r] between widgets, so spectate might not need any changes.

[r]: https://discord.com/channels/793331351645323264/1017471248567124049/1342514184180596827

DFHack does not seem to include any other fullscreen overlay widgets that render in `dwarfmode/Default` (like spectate's TooltipOverlay) so this problem is not readily observable in typical installations. I have a local fullscreen overlay that renders in `dwarfmode/Default`, so I was able to observe its last background color (and "map state") bleed into spectate's unit banners.

Complicating the situation, overlay widget rendering order is effectively randomized (for each DF launch) since it is based on `pairs`. So, even when there is another widget active for the same context, the banners may still be drawn normally unless that other widget is rendered right before the spectate banner widget.

The following overlay can demonstrate the problem if conditions are right:
``` lua
--@ module = true

OverlayPainterColors = defclass(OverlayPainterColors, require('plugins.overlay').OverlayWidget)
OverlayPainterColors.ATTRS{
    desc = 'Leaves the overlay Painter configured with random colors.',
    default_enabled = true,
    fullscreen = true,
    viewscreens = 'dwarfmode',
}

local rng = dfhack.random.new()

function OverlayPainterColors:render(painter)
    OverlayPainterColors.super.render(self, painter)

    -- Leave the painter configured with random colors; the Painter is shared
    -- among overlays of the same class (fullscreen or scaled), so these random
    -- colors might "bleed" into another overlay if it doesn't fully establish
    -- its own colors.
    painter:pen(rng:random(16), rng:random(16))
end

OVERLAY_WIDGETS = { fullscreen = OverlayPainterColors }

-- configure many instances to increase the chances one will render immediately
-- before spectate's TooltipOverlay

-- for k in ('abcdefhijk'):gmatch('.') do
--     OVERLAY_WIDGETS[k] = OverlayPainterColors
-- end
```

As noted, the reproduction is non-deterministic since it is subject to `pairs`-based rendering order. Uncommenting the last bit can help with increasing the odds that the problem can be observed without having to restart DF.

myk002:
the critical part is the call to `painter:pen`, which none of our "production" overlays do. So the color bleeding is a bug that should be fixed, certainly, but it's not super critical

myk002 (approved):

PR #5323, opened 2025-03-02
spectate.lua: add "hold-to-show" option (f.e. `ctrl`)

TymurGubayev:
Also, add a `toggle <setting>` command line for boolean settings. This allows keybindings for `spectate toggle tooltip-follow`.

### DIFF
--- a/docs/plugins/spectate.rst
+++ b/docs/plugins/spectate.rst
@@ -35,7 +35,7 @@ Usage
 
     enable spectate
     spectate [status]
-    spectate toggle <setting>
+    spectate toggle [<setting>]
     spectate set <setting> <value> [<subvalue>]
     spectate overlay enable|disable
 
--- a/docs/plugins/spectate.rst
+++ b/docs/plugins/spectate.rst
@@ -37,6 +37,7 @@ Usage
     spectate [status]
     spectate toggle
     spectate set <setting> <value> [<subvalue>]
+    spectate toggle <setting>
     spectate overlay enable|disable
 
 Examples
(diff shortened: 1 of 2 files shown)


realSquidCoder:
could we make it so one could make it "ctrl to hide" to?

TymurGubayev:
> could we make it so one could make it "ctrl to hide" to?

sure, but I don't understand the use case?

If it should still supersede the blinking, just remove the `not` in line 422:

before:
```lua
if not dfhack.internal.getModifiers()[hold_to_show] then
```
after:
```lua
if dfhack.internal.getModifiers()[hold_to_show] then
```

realSquidCoder:
my use case for this is we have ctrl to show for a quick heads up, or i could leave it on at all times and use ctrl to hide when theres a bunch and i need to see (like my animal pens) i dont like the blinking personally

TymurGubayev:
But then you might keep it pressed while doing other things, which would trigger some undesired action. It's still a problem with "hold to show", but IMO less so because it shows additional stuff instead of hiding. I'm just not sure the added complexity is worth it.

How about the `toggle` functionality bound to some hotkey on the default screen (not sure if it's possible to bind just `ctrl`, but something like `ctrl+t` should work)

docs/plugins/spectate.rst
```diff
@@ -37,6 +37,7 @@ Usage
     spectate [status]
     spectate toggle
     spectate set <setting> <value> [<subvalue>]
+    spectate toggle <setting>
```
myk002, line 40: could you combine this with the `spectate toggle` above as `spectate toggle [<setting>]`?

TymurGubayev: sure, I just thought it's distinct enough

myk002 (approved):

PR #5356, opened 2025-03-17
use `std::filesystem::path` for filesystem operations

ab9rf:
Recode most uses of `std::string` to represent a path name to use `std::filesystem::path` instead. Replace most of old crunchy filesystem module with `std::filesystem` equivalents

Some modules have also cleaned to address `using filesystem std` and/or missing or unnecessary `#include`s

The `atime`, `ctime`, and `filetype` methods of the `Filesystem` module have been removed because they're POSIX-specific, not supported by `std::filesystem`, and unused in DFHack

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -544,7 +544,6 @@ std::filesystem::path Core::findScript(std::string name)
     for (auto it = paths.begin(); it != paths.end(); ++it)
     {
         std::filesystem::path path = *it / name;
-        std::cerr << "findScript " << name << " | " << path << " | " << Filesystem::isfile(path) << std::endl;
         if (Filesystem::isfile(path))
             return path;
     }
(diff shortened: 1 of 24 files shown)


ab9rf:
it would appear that i need to rewrite the tests for `Filesystem` :)

library/Core.cpp
```diff
@@ -2272,16 +2265,16 @@ void Core::handleLoadAndUnloadScripts(color_ostream& out, state_change_event eve
     if (!df::global::world)
         return;
 
-    std::string rawFolder = !isWorldLoaded() ? "" : "save/" + World::ReadWorldFolder() + "/init";
+    std::filesystem::path rawFolder = !isWorldLoaded() ? std::filesystem::path{} : std::filesystem::path{} / "save" / World::ReadWorldFolder() / "init";
```
myk002, line 2268: I believe this path is deprecated, but we can address that in another PR

myk002 (approved):

PR #5361, opened 2025-03-23
more path/string normalization

ab9rf:
added `Filesystem::as_string` to provide namespace recanonicalization and change use in LuaApi to use this exported API

use `as_string` in `orders`

further revision of #5360

### DIFF
--- a/plugins/orders.cpp
+++ b/plugins/orders.cpp
@@ -153,7 +153,7 @@ static void list_library(color_ostream &out) {
         if (name.extension() != ".json")
             continue; // skip non-.json files
         auto sname = name.stem();
-        out << "library/" << sname << std::endl;
+        out << Filesystem::as_string("library" / sname) << std::endl;
     }
 }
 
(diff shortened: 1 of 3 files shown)


plugins/orders.cpp
```diff
@@ -153,7 +153,7 @@ static void list_library(color_ostream &out) {
         if (name.extension() != ".json")
             continue; // skip non-.json files
         auto sname = name.stem();
-        out << "library/" << sname << std::endl;
+        out << Filesystem::as_string("library" / sname) << std::endl;
```
myk002, line 156: The documented names of the library orders sets use the forward slash. Will this preserve that? Same for stockpiles, blueprint, quickfort, and other tools that come with libraries

myk002 (approved):

plugins/orders.cpp
ab9rf: this only transforms names that are transformed from pathnames enumerated by `listdir_recursive`. input from other sources is completely unaffected

PR #5363, opened 2025-03-26
add `Military::addToSquad`

ab9rf:
does not handle setting a squad leader yet so will just fail if you try to add to position 0


myk002 (commented):
could you expose this to Lua and add some docs in Lua API.rst?

myk002 (approved):

PR #5371, opened 2025-04-01
[gui/sitemap] make toolbar icon fit in with others in the group

myk002:
add black line border like adjacent icons


realSquidCoder:
agreed

PR #5376, opened 2025-04-03
Export DF world map for use in GIS tools

chdoc:
The goal of this plugin is to export the world map at the region-tile (i.e., embark tile or midmap-tile) level for use with GIS programs like ArcGIS or QGIS. For a medium-sized map, this amounts to (129*16)^2 = 4260096 region tiles with biome information for every region tile fetched from one of the surrounding world tiles. 

In order to keep the output down to a manageable size (17 MB of CSV for a medium-sized world), region tiles are clustered by the world tile they take their information from, and every cluster is exported as a single feature with its geometry described as a WKT (MULTI)POLYGON, resulting in an export of only 129^2 = 16641 features.

- [x] documentation and change log


plugins/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ if(BUILD_SUPPORTED)
     #dfhack_plugin(dwarfmonitor dwarfmonitor.cpp LINK_LIBRARIES lua)
     #add_subdirectory(embark-assistant)
     dfhack_plugin(eventful eventful.cpp LINK_LIBRARIES lua)
+    dfhack_plugin(export-map export-map.cpp COMPILE_FLAGS_GCC -fno-gnu-unique)
```
myk002, line 74: what is requiring `-fno-gnu-unique`?

ab9rf: probably a STL template; it can be hard to figure out which one because you have to pore over the libc++ implementation and that's very timeconsuming. 

`-fno-gnu-unique` should always be safe for DFHack plugins, as long as they don't crossload another another plug-in module

chdoc: Nothing is actually requiring this. Is what a (failed) attempt to make the plugin unload. I will have to check whether it unloads, now that I have removed the GDAL dependency. Keeping this there and the conversation open in the meantime.

plugins/export-map.cpp
```diff
@@ -0,0 +1,475 @@
...
+        // hashing is easy, if values are smaller than the hash
+        return ((std::size_t)pos.x << 16) | (std::size_t)pos.y;
+    }
+};
+
+// were only dealing with 2D coordinates in this file
+using coord = df::coord2d;
+
+// static int wdim = 768; // dimension of a world tile
+static int rdim = 48;  // dimension of a region tile
```
ab9rf, line 58: should this be `constexpr` instead of `static`?

plugins/CMakeLists.txt
chdoc: I removed the `-fno-gnu-unique`.

library/include/df/custom/coord2d.methods.inc
```diff
@@ -1,4 +1,4 @@
-coord2d(uint16_t _x, uint16_t _y) : x(_x), y(_y) {}
+coord2d(int16_t _x, int16_t _y) : x(_x), y(_y) {}
```
ab9rf, line 1: I've been reluctant to make this change myself because I've been uncertain of the consequences - how well tested is this behaviorally?

library/include/MiscUtils.h
```diff
@@ -413,6 +413,42 @@ inline bool static_add_to_map(CT *pmap, const typename CT::key_type key, const t
...
  * MISC
  */
+template<
+    std::ranges::input_range Range,
+    std::invocable<std::ostream&, std::ranges::range_reference_t<Range>> Callable>
+void print_range(
+    std::ostream &out,
+    const Range &elements,
+    Callable&& print_element,
+    const std::string &prefix = "[",
```
ab9rf: `const std::string &` is usually an indication that `std::string_view` should be used instead

plugins/export-world-map.cpp
```diff
@@ -0,0 +1,920 @@
...
+
+    // ensure that we have an output file
+    auto out_file = open_output_file("sites.csv");
+    if (!out_file) {
+        return CR_FAILURE;
+    }
+
+    // If you change anything in this vector, don't forget to change the
+    // corresponding comments and arguments in the call to print_csv below
+    vector<std::string> headings = {
```
ab9rf: consider using `const char*` or `std::string_view` here, to avoid potentially unneeded heap allocations/deallocations

plugins/export-world-map.cpp
```diff
@@ -0,0 +1,920 @@
...
+
+constexpr int wdim = 768; // dimension of a world tile
+constexpr int rdim = 48;  // dimension of a region tile
+
+/**
+ * Takes a vector of coordinates interpreted as global region tile coordinates
+ * (i.e. 16 region tiles per world tile) and emits a WKT path in GIS-compatible
+ * local tile coordinates (negative y-coordinates, 48 stepts per region tile)
+ */
+static void print_path(std::ostream &out, const std::vector<coord> &path) {
```
ab9rf: `std::span` instead of reference to `vector`

library/include/MiscUtils.h
```diff
@@ -413,6 +413,42 @@ inline bool static_add_to_map(CT *pmap, const typename CT::key_type key, const t
...
+        }
+    }
+    out << suffix;
+}
+
+template<std::ranges::input_range Range>
+void print_range(
+    std::ostream &out,
+    const Range& elements,
+    const std::string &prefix = "[",
```
ab9rf: these should all be `std::string_view`

plugins/export-world-map.cpp
```diff
@@ -0,0 +1,920 @@
...
+
+    // ensure that we have an output file
+    auto out_file = open_output_file("regions.csv");
+    if (!out_file) {
+        return CR_FAILURE;
+    }
+
+    // If you change anything in this vector, don't forget to change the
+    // corresponding comments and arguments in the call to print_csv below
+    vector<std::string> headings = {
```
ab9rf: consider `std::string_view` or `const char *` here to avoid heap allocations

ab9rf (changes requested):

plugins/export-world-map.cpp
ab9rf: Also this code appear repetitive of code elsewhere, should there be an abstraction here?

library/include/df/custom/coord2d.methods.inc
chdoc: Apart from the fact that it now allows using brace initialization using the correct type for the members, I have not noticed any adverse effects. I haven't played a whole lot on this branch, but I did play a few hours in total and I didn't notice anything off.

plugins/export-world-map.cpp
chdoc: I went with `constexpr std::array<std::string_view, _>` and I did factor out the `print_csv` definitions. Beyond that, one would actually have to use a real class of a CSV writer. However, that wouldn't really be a good fit, because the current implementation uses unfinished lines to then add the geometry as a separate step.

ab9rf (approved):

Issue #5275, opened 2025-02-16
Check stockpiles library validity


myk002:
As reported on discord: https://discord.com/channels/793331351645323264/1017471248567124049/1340449047948820572

xBlackTalonX:
I quickly throw up a new for to test this again and the stockpile configs match their names. I'm not sure how but I have to assume user error on my part somehow because I can't imagine something changing these later down the line. Sorry for the false report and hope you didn't waste time on it yet.

myk002:
There is another report that somehow things are going wrong. There might be something to find yet

xBlackTalonX:
#5362 or something else as well? I didn’t see anything in my quick review of Discord but if there are any more specifics I’d be happy to attempt to reproduce and get more details this weekend.

PR #5377, opened 2025-04-04
[stockpiles] fix conversion of enum token to value

myk002:
was not correctly handling enums that did not start at 0, leading to errors in furniture stockpiles

Fixes: https://github.com/DFHack/dfhack/issues/5275


ab9rf:
This introduced a bug: see #5557

PR #5385, opened 2025-04-07
New `gui.dflayout` module to provide position information for DF UI elements; starting with fort mode toolbars

ChrisJohnsen:
Here is the library code from DFHack/scripts#1426 moved to a top-level DFHack Lua module.

This spin has tests, and two hunks of docs: one for the library itself, and another in the overlay dev guide that describes a generalized "UI-relative" positioning technique (the warmdamp technique, but using the new library to encapsulate the magic numbers, and extending to right- and top-anchored overlay positions).

### DIFF
--- a/docs/dev/overlay-dev-guide.rst
+++ b/docs/dev/overlay-dev-guide.rst
@@ -459,3 +459,188 @@ screen (by default, but the player can move it wherever).
             },
         }
     end
+
+Widget example 4: positioning relative to DF UI
+-----------------------------------------------
+
+The player-controllable positioning provided by the overlay framework is
+relative to some corner of the interface area (technically, the positioning is
+edge-relative, but the anchors are always two adjacent edges, so it is
+equivalent to corner-relative positioning). This corner-relative positioning
+works well to adapt overlay positions for interface areas of varying sizes (due
+to DF window size changes, or because the DF interface percentage has been
+reconfigured).
+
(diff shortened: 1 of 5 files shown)


docs/dev/overlay-dev-guide.rst
```diff
@@ -459,3 +459,99 @@ screen (by default, but the player can move it wherever).
...
+Instead, with some clever manipulation, an overlay's corner-relative positioning
+can be "translated" to be relative to some other UI element.
+
+The ``gui.dflayout`` module provides the ``getOverlayPlacementInfo`` helper
+function that provides helper values that can automatically adapt an overlay's
+corner-relative, player-customizable position to be relative to a UI element
+while still being player-customizable::
+
+    local overlay = require('plugins.overlay')
+    local Label = require('gui.widgets.labels.label')
```
ChrisJohnsen, line 495: I have a local commit that changes this to the normal `local widgets = require('gui.widgets')`.
I can push it whenever, but thought I might hold on to it (avoiding a possible Windows-build CI hiccup) until I also have something substantive from my attempts to further generalize the overlay placement code.

lethosor:
Just putting a thought down (haven't looked thoroughly): I feel like this is conceptually similar to `gui/dwarfmode.lua`, which is party obsolete - maybe the two could be merged.

PR #5388, opened 2025-04-07
add `army-controller-sanity` plugin

ab9rf:
this add a development tool that monitors, on a tick by tick basis, for inconsistencies in the game's army controller data, pausing if an inconsistency is found

the tool has a modest performance impact so it's not enabled by default


plugins/army-controller-sanity.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+DFhackCExport DFHack::command_result plugin_onupdate(DFHack::color_ostream& out)
+{
+    if (is_enabled)
+    {
+        bool ok = checkArmyControllerSanity();
+        if (!ok) {
+            ERR(log).print("Army controller sanity check failed! Game pause forced.\n");
+            *df::global::pause_state = true;
```
myk002, line 128: Can simplify to `*pause_state = true` since `REQUIRE_GLOBAL` adds a `using` statement.

You can also forego the REQUIRE_GLOBAL entirely by using `World::SetPauseState`

plugins/army-controller-sanity.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+DFhackCExport DFHack::command_result plugin_enable(DFHack::color_ostream& out, bool enable)
+{
+    is_enabled = enable;
+    return DFHack::CR_OK;
+}
+
+DFhackCExport DFHack::command_result plugin_onupdate(DFHack::color_ostream& out)
+{
+    if (is_enabled)
```
myk002, line 123: This isn't required. `plugin_onupdate` only gets called if `is_enabled` is true

myk002: however, it might be a good idea to only run once per tick.

plugins/army-controller-sanity.cpp
```diff
@@ -0,0 +1,133 @@
...
+        // army controllers are found:
+        //   in viewscreen_worldst (army_controller (list), last_hover_ac) (not checked)
+        //   in entitst (army_controller (list))
+        //   in armyst (controller)
+        //   in unitst (army_controller)
+        //
+        // master list is in army_controller_handlerst
+
+        static size_t last_ac_vec_size = 0;
+        static int last_army_controller_next_id = 0;
```
myk002, line 36: clear state on world unload? what about disable/enable?

plugins/army-controller-sanity.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        if (last_army_controller_next_id == *df::global::army_controller_next_id &&
```
myk002, line 38: explicit `df::global` usage in plugins is a warning sign that REQUIRE_GLOBAL was forgotten. this plugin adds REQUIRE_GLOBAL correctly, so the df::global prefix should be elided

plugins/army-controller-sanity.cpp
ab9rf: the cost to test if it's running only once a tick is basically as expensive as the test that checks if the army controller list length has changed, so i don't see any benefit to it

plugins/army-controller-sanity.cpp
ab9rf: i thought about that but it doesn't really matter, as the only point of these is to gatekeep when the test is run

myk002 (approved):

PR #5391, opened 2025-04-11
code modernization

ab9rf:
this commit changes several templates to have more precisely constrained match patterns to yield more clear compilation errors when a mistake is made

certain `Lua::Push` overloads have been made either more or less general; a side effect is that more "string-like" objects (notably, `std::string_view`) will be converted to lua strings on push

 a chunk of ancient dead code bracketed by `#if 0` was removed

a declaration for the undefined and unused `Lua::Core::Init` was removed

it also replaces several uses of `enable_if` with `if constexpr` or templates constrained by appropriate types, to make that code more readable

some other minor cleanups, mostly around namespaces blocks,, are also included

### DIFF
--- a/library/include/LuaTools.h
+++ b/library/include/LuaTools.h
@@ -32,6 +32,7 @@ distribution.
 #include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
+#include <concepts>
 
 #include "Core.h"
 #include "ColorText.h"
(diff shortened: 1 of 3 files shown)


library/include/LuaTools.h
```diff
@@ -60,7 +61,7 @@ namespace DFHack {
     };
 }
 
-namespace DFHack {namespace Lua {
+namespace DFHack::Lua {
```
myk002, line 64: I didn't know we could do this : )

myk002 (approved):

library/include/LuaTools.h
ab9rf: C++17

PR #5406, opened 2025-04-18
Fix Items::getDescription

### LINKED ISSUE #5404: Items::getDescription does not display quality correctly
The logic used by the Items module function "getDescription" to display quality levels does not match the game, causing inconsistencies with certain items types.

The game itself does the following:
```
  base_quality = item->getQuality();
  imp_quality = item->getImprovementQuality()
  show_imps = item->isImproved();
  if ((!show_imps) && (base_quality < imp_quality)) {
    base_quality = imp_quality;
  }
```
The result is that "non-displayed" Improvements cause the item's displayed base quality level to be boosted. Such improvements include Thread (on dyed cloth), Cloth (on clothing items), and Prepared Meal Ingredients.

Additionally, the `SHOW_IMP_QUALITY` init flag is not being respected - when not set, this should result in the "outer" quality level being hidden (though it does *not* cause the inner quality level to be boosted, as it was in very old versions before that option was added).

quietust:
* Display correct quality level for items with "non-displayed" improvements
* Add ALL item designations here (and remove "XXwearXX" from other spot)
* Obey SHOW_IMP_QUALITY setting

Fixes #5404

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -95,6 +95,7 @@ Template for new versions:
 - ``Buildings`` module: add ``getOwner`` (using the ``Units::get_cached_unit_by_global_id`` mechanic) to reflect changes in 51.11
 - ``Units::teleport``: projectile information is now cleared for teleported units
 - ``Buildings::setOwner``: updated for changes in 51.11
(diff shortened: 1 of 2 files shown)


quietust:
I don't have a fortress with which to confirm that all of these changes work as expected - please test before merging

quietust:
See PR #5434 (created from a proper branch rather than "develop")

Issue #5239, opened 2025-02-03
Ctrl-a on empty EditField makes cursor disappear


myk002:
@wiktor-obrebski could you look at this?

repro steps:

1. open `gui/launcher`
2. hit Ctrl-A to select all on the empty text
3. observe that the cursor has disappeared (it comes back if you type something or click on the field)

myk002:
Fixed in https://github.com/DFHack/dfhack/pull/5412

PR #5412, opened 2025-04-21
Fix/empty text select all

wiktor-obrebski:
For TextArea and EditField, visual cursor dissapeared when "Select All" (`Ctrl+A`) has been requested.
This fixes it (https://github.com/DFHack/dfhack/issues/5239)


myk002 (approved):
Thanks! I'll add a changelog when I do the release notes for the release later today.

PR #5422, opened 2025-04-23
Fix for mods not being properly reloaded on Windows from saved game

dvantwisk:
Steam mods with lua scripts in the "scripts_modactive" directory of a mod were not being properly reloaded after the initial load. That is, the script-manager purges these scripts from `dfhack.internal.scripts` before reloading them (more specifically to register their onStateChange). This purge wasn't being done since the file path wasn't being recognized on windows, thus dfhack mods weren't being loaded in games from resumed saves (if DF wasn't shut down before then). This change simply adds the correct windows file path as a key to access the script. It may be better to standardize all OS's paths used as keys in `dfhack.internal.scripts` but it isn't clear to me the locations where all these files are being added, so this is the simple solution.

### DIFF
--- a/docs/about/Authors.rst
+++ b/docs/about/Authors.rst
@@ -63,6 +63,7 @@ DoctorVanGogh           DoctorVanGogh
 Donald Ruegsegger       hashaash
 doomchild               doomchild
 Droseran                Droseran
+dvantwisk               dvantwisk
 DwarvenM                DwarvenM
 Eamon Bode              eamondo2                Baron Von Munchhausen
 EarthPulseAcademy       EarthPulseAcademy
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -72,6 +72,7 @@ Template for new versions:
 ## Fixes
 - `preserve-rooms`: will no longer crash when a civzone is assigned to a unit that does not exist
 - `gui/design`: fix misaligned shape icons
+- `scripts/script-manager`: fix lua scripts in mods not being reloaded properly on Windows.
 
 # 51.11-r1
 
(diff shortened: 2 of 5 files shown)


myk002 (approved):
I moved the fix into the core function that is returning the incorrect string. Thank you for helping us track this down!

dvantwisk:
@myk002
I'm trying to run a script on 51.11-r1.2 and it doesn't seem to conserve its state on Windows. This version should have the change you added. It does work if I use my fix. Can you double check to see if it works? It may be easier to just use the simple solution I originally had.

myk002:
That commit was merged after 51.11-r1.2 was tagged. Try the `testing` branch to see if the issue persists there

dvantwisk:
@myk002
Okay, just looked at it. The problem is that the keys are now all forwardslashed, but the script_paths are still backslashed. It can be fixed with this at line 55 after full path is initially declared:
           ` full_path = string.gsub(full_path, "\\", "/")`

myk002:
Ah, are you saying that the paths returned from `dfhack.filesystem.listdir_recursive` have backslashes in them? The fix should be in that function, then. The intent of @ab9rf 's change is that backslashes should not make their way back to Lua.

dvantwisk:
That seems to be what is happening. I guess I don't know exactly how you guys want to plug up all the window's backslash holes.

ab9rf:
> Ah, are you saying that the paths returned from `dfhack.filesystem.listdir_recursive` have backslashes in them? The fix should be in that function, then. The intent of @ab9rf 's change is that backslashes should not make their way back to Lua.

No, listdir_recursive generates a list of pathname objects, not a list of strings. The fix implemented in this PR is the correct fix: the transition should occur when a pathname is to be intepreted as a string, and not before.

myk002:
> `filesystem_listdir` does _not_ have the `as_string` conversion

bollocks, let me fix that then.

I'll get that change pushed to the `testing` branch.

PR #5428, opened 2025-04-25
Fix getplants "ripe" and "picked" logic

### LINKED ISSUE #5426: getplants - "ripeness" formula no longer correct, "picked" logic broken
The fix for #1479 added a formula (https://github.com/DFHack/dfhack/issues/1479#issuecomment-649822893) for determining the random variance in "ripeness" time for plant growths. That formula appears to have changed in version 50.01 and is now as follows:

1. Calculate `x_coord`+`y_coord`+`z_coord` to be the embark-local X/Y/Z coordinates plus `(region_x*48, region_y*48, region_z)`
2. Calculate a 64-bit "seed" value of `x_coord + (y_coord * 10000) + (z_coord * 100000000)`
3. Feed the "seed" value into a [SplitMix64](https://rosettacode.org/wiki/Pseudo-random_numbers/Splitmix64) RNG instance
4. Take the upper 32 bits of the RNG's "next" value, modulo 2000, then add that to `cur_year_tick` and modulo 403200 (as before)

quietust:
Also adds Splitmix64 RNG to Random module (not yet visible from Lua).

Should fix #5426


quietust:
This will need to be checked against a fortress containing shrubs whose growths are in the process of ripening.

myk002 (approved):
code looks good. will merge once once of us has a chance to test

PR #5434, opened 2025-04-27
Fix Items::getDescription

### LINKED ISSUE #5404: Items::getDescription does not display quality correctly

quietust:
* Display correct quality level for items with "non-displayed" improvements
* Add ALL item designations here (and remove "XXwearXX" from other spot)
* Obey SHOW_IMP_QUALITY setting

Fixes #5404

Replaces PR #5406 (which was merging from the wrong branch)

### DIFF
--- a/library/modules/Items.cpp
+++ b/library/modules/Items.cpp
@@ -49,6 +49,7 @@ distribution.
 #include "df/caravan_state.h"
 #include "df/creature_raw.h"
 #include "df/dfhack_material_category.h"
+#include "df/d_init.h"
 #include "df/entity_buy_prices.h"
 #include "df/entity_buy_requests.h"
 #include "df/entity_raw.h"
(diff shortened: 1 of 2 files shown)


library/modules/Items.cpp
```diff
@@ -678,14 +679,54 @@ string Items::getDescription(df::item *item, int type, bool decorate) {
...
+
+        // Next, magic enchants
+        if (item->getMagic() != NULL)
+            tmp = '\x11' + tmp + '\x10'; // <| |>
+
+        // Next, improvements
+        if (has_displayed_item_improvements) {
             tmp = '\xAE' + tmp + '\xAF'; // («) + tmp + (»)
-            addQuality(tmp, item->getImprovementQuality());
+            if (df::global::d_init->display.flags.is_set(d_init_flags1::SHOW_IMP_QUALITY))
```
myk002, line 702: although we've traditionally tested globals before using them in the core library, I think this function is well out of the way of the code path that might be hit when globals aren't defined, so this is fine.

myk002 (approved):

PR #5441, opened 2025-05-03
merge `Process.cpp` into a single source file

ab9rf:
lots of `#ifdef`

i tried to keep as much commonality as possible. it gets a bit difficult in places


library/Process.cpp
```diff
@@ -0,0 +1,801 @@
...
+
+3. This notice may not be removed or altered from any source
+distribution.
+*/
+
+#ifndef WIN32
+#ifndef _DARWIN
+#endif /* ! WIN32 */
+#include <cstdlib>
+#ifndef WIN32
```
myk002, line 29: maybe adding spaces would make the nesting easier to read here.

myk002 (commented):
ugh, that's grotty, but arguably better than having diverging implementations in different files

library/Process.cpp
ab9rf: oh that's just a mess, lemme make it better

ab9rf:
> ugh, that's grotty, but arguably better than having diverging implementations in different files

yeah that's where i am too. i had aligned the separate files as best i could (using side by side edit) a year or so ago, and even with that they had managed to drift a bit in that relatively limited time

myk002 (approved):

PR #5445, opened 2025-05-04
More code modernizations

ab9rf:
Most of these changes address removing unnecessary type erasures

Several related to the change to `std::thread`; `std::thread` is polymorphic while our preceding threading library was not. When we converted, we never removed the type erasure the prior library required

The other major change is to switch `RunQueryCoreLoop` to take a `std::function` instead of a bare function pointer and a type-erased argument list. This makes the functors for `RunQueryCoreLoop` more transparent and easier to understand. Hopefully gcc 10 will accept `std::placeholders`. :)


myk002:
> Hopefully gcc 10 will accept std::placeholders

Yeah I used those in `stockpiles` without issues

myk002 (approved):

PR #5451, opened 2025-05-07
adjust `df_object` constraint

### LINKED ISSUE #5450: Bug: Vector passed to Lua as bool
Invoking `plant remove -pf alder` causes a Lua error. The code
```c++
vector<int32_t> filter; // Unsorted

if (!Lua::CallLuaModuleFunction(out, "plugins.plant", "parse_commandline",
    std::make_tuple(&options, &pos_1, &pos_2, &filter, parameters)))
```
in `plant.cpp` is resulting in the `vector<int32_t>` being passed into Lua as a `bool`, which then causes an error when the script tries to get the length. The other args of the tuple are passed as expected.

ab9rf:
some (most, actually) `get` methods return a pointer to a `const` instance of a subclass of `type_identity`  which isn't convertible to a pointer of non-`const` `type_identity`. adding `const` fixes this. a pointer to non-`const` is always convertible to a pointer of `const` of the same type or of a compatible type

fixes #5450

the reason why the pointer to vector glossed as a boolean is that pointers have an implicit conversion to `bool`, and so when the type inference for the template failed, it fell back to the next best overload, which was the one for `bool`

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -62,6 +62,8 @@ Template for new versions:
 - `gui/teleport`: adapt to new behavior in DF 51.11 to avoid a crash when teleporting items into mid-air
 - `script-manager`: fix lua scripts in mods not being reloaded properly upon entering a saved world on Windows
 - `preserve-rooms`: don't warn when a room is assigned to a non-existent unit.  this is now common behavior for DF when it keeps a room for an unloaded unit
+- fixed an overly restrictive type constraint that resulted in some object types being glossed as a boolean when passed as an argument from C++ to Lua
+- `plants`: will no longer generate a traceback when a filter is used
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


ab9rf:
> Is this something we could catch in a unit test? I guess it would require a C++ function that takes an affected type...

I should probably add some `static_assert`s, since that's basically how I found it (along with single stepping).

This particular one is insuring that `Lua::Push(lua_State,bool)` doesn't get "grabby", since that overload will capture any type that has an implicit conversion to `bool` (and lots of things have implicit conversions to `bool`) and doesn't get matched without a conversion chain.

myk002:
> C++ unit testing

We do have cpp unit tests set up. There's an example unit test for miscutils

https://github.com/DFHack/dfhack/blob/develop/library/MiscUtils.test.cpp

myk002 (approved):

PR #5462, opened 2025-05-15
add research info to spectate-overlay

brmley:
Some changes to allow research-info (https://github.com/DFHack/scripts/pull/1454) works with spectate-overlay


chdoc:
In any case, this depends on https://github.com/DFHack/scripts/pull/1454 being merged first. The way I understand it, it adds spectate hooks for something that hasn't been merged yet. That being said, it also requires a changelog entry and possibly a mention in the documentation for `spectate`.

PR #5464, opened 2025-05-18
Fix potential vulnerable cloned function

npt-1707:
Hi Development Team,

I identified a potential vulnerability in a clone function luaG_runerror() in `depends/lua/src/ldebug.c` sourced from [lua/lua](https://github.com/lua/lua). This issue, originally reported in [CVE-2022-33099](https://nvd.nist.gov/vuln/detail/CVE-2022-33099), was resolved in the repository via this commit https://github.com/lua/lua/commit/42d40581dd919fb134c07027ca1ce0844c670daf.

This PR applies the corresponding patch to fix the vulnerabilities in this codebase.

Please review at your convenience. Thank you!

### DIFF
--- a/depends/lua/src/ldebug.c
+++ b/depends/lua/src/ldebug.c
@@ -658,8 +658,11 @@ l_noret luaG_runerror (lua_State *L, const char *fmt, ...) {
   va_start(argp, fmt);
   msg = luaO_pushvfstring(L, fmt, argp);  /* format message */
   va_end(argp);
-  if (isLua(ci))  /* if Lua function, add source:line information */
+  if (isLua(ci)) {  /* if Lua function, add source:line information */
     luaG_addinfo(L, msg, ci_func(ci)->p->source, currentline(ci));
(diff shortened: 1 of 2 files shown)


silverwing235:
Duplicate of #5465

ab9rf:
This isn't a duplicate of #5465; however, the patch is based on a change in 5.4 and as we're on 5.3 I am not going to blithely apply it without more information.  Given the low importance of this change, I'm not inclined to put out a lot of effort doing this.

Let me also point out that this is a buffer overrun, which in our context would mean that a Lua script author could potentially use this to run local code. As a Lua script author already has the unlimited ability to run local code _anyway_, all this amounts to is that there is a small risk that extremely stupid Lua code could cause an application crash. Obviously, this is not something we care all that terribly much about, and so the effort that we'll go to to investigate whether this change is worth applying is very limited.

I'm therefore not recommend this be merged until someone has confirmed that the change proposed is appropriate for Lua 5.3, either by examining the differential between 5.3 and 5.4, or by testing it on our platform. I very much doubt that the OP has done either of these things.

PR #5475, opened 2025-06-14
Adventure Mode Site Helpers: `SC_NEW_MAP_AVAILABLE` state change and `World::GetCurrentSiteIdsWithExtraRange`

UnknowableCoder:
After doing some research on the ways to properly get access to site IDs in Adventure Mode, I realized the `SC_MAP_LOADED` state change only triggers on transitioning from no map to having a map loaded. The logic can easily be expanded to also consider the cases when there is a map and the map changes. For this, I propose a new state change, which I have tentatively called `SC_NEW_MAP_AVAILABLE`.

In the same vein, while `World::GetCurrentSiteId` should correctly return the current site ID (or -1 if no site), some practical use cases may benefit from being able to get all sites in a caller-provided range around the adventurer. For this reason, I propose the introduction of `World::GetCurrentSiteIdsWithExtraRange`, which extends the logic of the former function to take an `std::vector` as an in-parameter and return all applicable sites, optionally with extra range in x and y and a maximum number of sites to be return (which would be ordered by ID and not distance...).

### DIFF
--- a/library/include/modules/World.h
+++ b/library/include/modules/World.h
@@ -92,6 +92,7 @@ namespace DFHack
         DFHACK_EXPORT df::unit * getAdventurer();
 
         DFHACK_EXPORT int32_t GetCurrentSiteId();
+        DFHACK_EXPORT void GetCurrentSiteIdsWithExtraRange(std::vector<int32_t>& ids, const int32_t x_range = 0, const int32_t y_range = 0, const uint32_t max = UINT_MAX);
         DFHACK_EXPORT bool IsSiteLoaded();
 
         // Store DFHack tool data in the game save directory.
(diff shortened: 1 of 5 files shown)


library/include/modules/World.h
```diff
@@ -92,6 +92,7 @@ namespace DFHack
         DFHACK_EXPORT df::unit * getAdventurer();
 
         DFHACK_EXPORT int32_t GetCurrentSiteId();
+        DFHACK_EXPORT void GetCurrentSiteIdsWithExtraRange(std::vector<int32_t>& ids, const int32_t x_range = 0, const int32_t y_range = 0, const uint32_t max = UINT_MAX);
```
ab9rf, line 95: Needs to be documented in `docs/dev/Lua API.rst`

ab9rf (changes requested):

library/include/modules/World.h
UnknowableCoder: My apologies for the delay in getting back to this. I had not added this function to the Lua API; would adding it be desirable, then?
I see that `GetCurrentSiteID()` is also not documented, despite being exported (and used to implement the actual documented `dfhack.world.getCurrentSite()`), so I am not sure what the criterion should be here.
I am willing to implement any changes as desired to reach the best code possible, my apologies for any confusion.

ab9rf: The criterion is very simple: every visibly exported function should be documented. The fact that some things have slipped under the radar in the past is not a justification for doing it going forward.

PR #5476, opened 2025-06-15
Extending Persistency to Allow Associating a File to Entity and Key Pair

UnknowableCoder:
I propose a set of functions to `DFHack::Persistency` to allow managing whole files, following the same entity ID and key logic as the pre-existing functions.

The API was designed around managing the possible paths for these files rather than the files themselves; it would be up to the callers to interact with them according to the demands of their particular use case. Every (entity ID, key) pair will be associated with a unique file ID, corresponding to a file with a name of the form `dfhack-pf-<ID>.dat`. The association between (entity ID, key) pairs and file IDs is saved as JSON in `dfhack-extra-files.dat`.

Files that are requested will be copied over to the `current` folder unless specifically requested as read only through the API (in which case modifying their contents leads to these changes not being reflected if the user does not save to the same folder).

No attempt is made at this stage to offer any particular form of synchronization or memory management of the files themselves. Operations that add to or query the list of available files are guarded by `CoreSuspender`, but this ultimately only protects the list of file IDs.

No provisions are made for associating more than one file to an (entity ID, key) pair.

Note: I surmised from perusing previous discussions that it was not yet consensual among the main developers whether DFHack should offer the ability to request an entire file for persistency. However, for my current project of overhauling the dwarf logic system, serialization/deserialization could be important for performance if the user configures many objects to be handled by my system, so a more optimized way of handling that seems desirable. My apologies if I overstep any boundaries by proposing this.

### DIFF
--- a/library/modules/Persistence.cpp
+++ b/library/modules/Persistence.cpp
@@ -22,6 +22,8 @@ must not be misrepresented as being the original software.
 distribution.
 */
 
+#include <filesystem>
+
 #include "Core.h"
 #include "DFHackVersion.h"
 #include "Debug.h"
@@ -197,6 +199,57 @@ struct LastLoadSaveTickCountUpdater {
     }
 };
 
+struct FileEntry
+{
+    size_t id;
+    bool in_current;
+    std::filesystem::path to_current_path() const
+    {
+        return getSaveFilePath("current", "pf-" + std::to_string(id));
+    }
+    std::filesystem::path to_world_path() const
+    {
+        return getSaveFilePath(World::ReadWorldFolder(), "pf-" + std::to_string(id));
+    }
+    std::filesystem::path to_path() const
+    {
+        if (in_current)
+        {
+            return to_current_path();
+        }
+        else
+        {
+            return to_world_path();
+        }
+    }
+    bool create_file()
+    {
+        std::ofstream f(to_path());
+        return f.is_open();
+    }
+    bool delete_file()
+    {
+        bool ret = false;
+        if (in_current)
+        {
+            ret |= std::filesystem::remove(to_current_path());
+        }
+        ret |= std::filesystem::remove(to_world_path());
+        return ret;
+    }
+    void move_to_current()
+    {
+        if (!in_current)
+        {
+            std::filesystem::copy_file(to_world_path(), to_current_path());
+            in_current = true;
+        }
+    }
+};
+
+static std::unordered_map<int, std::map<std::string, FileEntry>> file_storage;
+static size_t num_stored_files;
+
 void Persistence::Internal::save(color_ostream& out) {
     Core &core = Core::getInstance();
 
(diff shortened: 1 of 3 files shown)


library/modules/Persistence.cpp
```diff
@@ -197,6 +199,57 @@ struct LastLoadSaveTickCountUpdater {
     }
 };
 
+struct FileEntry
+{
+    size_t id;
+    bool in_current;
+    std::filesystem::path to_current_path() const
+    {
+        return getSaveFilePath("current", "pf-" + std::to_string(id));
```
ab9rf, line 211: I would prefer that the filename prefix start with `dfhack` so as to clearly distinguish it as part of DFHack's cosave mechanics

library/modules/Persistence.cpp
```diff
@@ -291,6 +374,54 @@ static bool load_file(const std::filesystem::path & path, int entity_id) {
...
+{
+    Json::Value json;
+    try {
+        std::ifstream file(path);
+        file >> json;
+    }
+    catch (std::exception&) {
+        // empty file?
+        return false;
+    }
```
ab9rf, line 390: At least log the exception reason as a diagnostic; failing silently is unkind

library/modules/Persistence.cpp
```diff
@@ -291,6 +374,54 @@ static bool load_file(const std::filesystem::path & path, int entity_id) {
...
+
+                                pair.first->second[k_v["k"].asString()] = FileEntry{this_idx, false};
+                                //file_storage[e][k] = v
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
ab9rf: i don't like the deep nesting here and would suggest using `std::ranges::filter` to prefilter the iterators instead

library/modules/Persistence.cpp
```diff
@@ -307,14 +438,23 @@ void Persistence::Internal::load(color_ostream& out) {
 
     bool found = false;
     for (auto & fname : files) {
-        int entity_id = Persistence::WORLD_ENTITY_ID;
-        if (fname != "dfhack-world.dat" && !get_entity_id(fname.string(), entity_id))
-            continue;
+        if (fname == "dfhack-extra-files")
```
ab9rf: using file names with no extension is rude to windows platform users

library/modules/Persistence.cpp
```diff
@@ -433,3 +573,154 @@ uint32_t Persistence::getUnsavedSeconds() {
...
+    }
+
+    CoreSuspender suspend;
+
+    return create_or_get_file(entity_id, key).to_path();
+}
+
+std::filesystem::path Persistence::getFile(int entity_id, const std::string& key, bool* added, bool just_for_reading)
+{
+    if (!is_good_entity_id(entity_id) || key.empty() || !Core::getInstance().isWorldLoaded())
```
ab9rf: this exact test appears at least three times on lines 580, 600, 610, and in part in 645 and 673 (and possibly other places).  should this be abstracted into a utility function to avoid repeating yourself?

library/modules/Persistence.cpp
UnknowableCoder: If I understand correctly, `getSaveFilePath` should already prefix the string provided as the second argument with `dfhack`.

library/modules/Persistence.cpp
UnknowableCoder: I followed the pre-existing `load_file` function above.
When the function returns `false`, this ultimately results in a printout from `Persistence::Internal::load`: `"Cannot load extra persistence files from: <path>"`.
I can add the more specific diagnostic here if you think that is important; the same should probably be added to `load_file` too.

library/modules/Persistence.cpp
UnknowableCoder: Done. Thank you for the suggestion.

library/modules/Persistence.cpp
UnknowableCoder: Sorry, this was actually a typo on my end and for some reason I did not push the commit that fixed this!
The file is indeed `dfhack-extra-files.dat`.

library/modules/Persistence.cpp
UnknowableCoder: Done. Thank you for the suggestion.

PR #5482, opened 2025-06-21
Add "count" parameter to Items::createItem

### LINKED ISSUE #5481: createitem no longer creates stacked items
When the `createitem` plugin was updated to use `DFHack::Items::createItem`, it lost the ability to create items in stacks - it originally specified the stack size directly in the `reaction_product_itemst` it created, but that got replaced with a loop which caused it to create individual items instead.

quietust:
and update the "createitem" plugin to specify it instead of creating a bunch of individual items in a loop.

Fixes #5481.


quietust:
I did not modify `modtools/create-item`, since it already has its own workaround code for handling item stacks (and it should continue to work as-is).

Issue #5366, opened 2025-03-30
autoclothing: add explicit clear command


myk002:
Right now, you clear a config element by setting it to 0. We could use a proper "clear" command with accompanying documentation to make this clearer.

https://discord.com/channels/793331351645323264/807444467140788254/1355897691946881349

ab9rf:
closed in #5510

PR #5510, opened 2025-07-24
Add clear option to autoclothing that will unset a previosly set order

LightHardt:
These changes aim to handle #5366. Updated documentation to alert users of new `clear` option that will allow them to unset a previously set order. This probably could have been a little cleaner if I added the clear at the end however I think putting it as the first parameter makes it read quite well. Compiled and tested on windows. Please let me know if there is anything I need ti change/do!

### DIFF
--- a/plugins/autoclothing.cpp
+++ b/plugins/autoclothing.cpp
@@ -136,6 +136,10 @@ struct ClothingRequirement {
                 out << "Unrecognized item name or token: " << parameters[2] << endl;
                 return false;
             }
+            else if (!validateMaterialCategory(this)) {
+                out << parameters[1] << " is not a valid material category for " << parameters[2] << endl;
+                return false;
+            }
             return true;
         }
         if (!set_bitfield_field(&material_category, parameters[0], 1))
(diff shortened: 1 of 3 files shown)


plugins/autoclothing.cpp
```diff
@@ -459,7 +474,10 @@ command_result autoclothing(color_ostream &out, vector<string> &parameters)
         if (settingSize) {
             if (newRequirement.needed_per_citizen == 0) {
                 clothingOrders.erase(clothingOrders.begin() + i);
-                out << "Unset " << parameters[0] << " " << parameters[1] << endl;
+                    if (parameters[0] == "clear")
```
ab9rf, line 481: I don't like this sort of downstream dependency on the specific command issued is bad design, but this is an architectural problem with the way this plugin was implemented in the first place and isn't a problem with this PR so I'm not going to block this PR on it

ab9rf (approved):

ab9rf:
Needs changelog, otherwise lgtm

Issue #5519, opened 2025-07-26
gui/mod-manager not loading: attempt to index a nil value (local 'path')


MalikMAlna:
**OS:** Windows 10
**DF Version:** 52.02-r1
**DFHack Version:** 52.02-r1

**Stack Traceback when I try to run `gui/mod-manager` from the launcher:**

```
> gui/mod-manager

...amapps\common\Dwarf Fortress\hack\lua\script-manager.lua:224: attempt to index a nil value (local 'path')
stack traceback:
 ...amapps\common\Dwarf Fortress\hack\lua\script-manager.lua:224: in function 'script-manager.get_active_mods'
 ...s/common/Dwarf Fortress/hack/scripts/gui/mod-manager.lua:423: in upvalue 'get_num_vanilla_mods'
 ...s/common/Dwarf Fortress/hack/scripts/gui/mod-manager.lua:508: in local 'fun'
 ...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:124: in upvalue 'invoke_after_rec'
 ...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:153: in global 'ModlistWindow'
 ...s/common/Dwarf Fortress/hack/scripts/gui/mod-manager.lua:558: in local 'fun'
 ...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:124: in upvalue 'invoke_after_rec'
 ...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:153: in global 'ModlistScreen'
 ...s/common/Dwarf Fortress/hack/scripts/gui/mod-manager.lua:666: in local 'script_code'
 ...team\steamapps\common\Dwarf Fortress\hack\lua\dfhack.lua:1107: in function 'dfhack.run_script_with_env'
 (...tail calls...)
 [C]: in field 'runCommand'
 ...team\steamapps\common\Dwarf Fortress\hack\lua\dfhack.lua:1150: in upvalue '_run_command'
 ...team\steamapps\common\Dwarf Fortress\hack\lua\dfhack.lua:1154: in function 'dfhack.run_command_silent'
 [C]: in field 'hideGuard'
 ...apps/common/Dwarf Fortress/hack/scripts/gui/launcher.lua:1148: in function <...apps/common/Dwarf Fortress/hack/scripts/gui/launcher.lua:1126>
 [C]: in field 'on_submit'
[...]

PR #5524, opened 2025-07-26
Fixing gui/mod-manager not loading

MalikMAlna:
Not sure if this mod makes an externally visible change in this case, but it should fix the issue I encountered in #5519 

Also, I ended up adding my name to Authors.rst for the fix. I hope that's fine.

### DIFF
--- a/library/lua/script-manager.lua
+++ b/library/lua/script-manager.lua
@@ -214,14 +214,44 @@ function get_active_mods()
     local ol = df.global.world.object_loader
 
     for idx=0,#ol.object_load_order_id-1 do
-        local path = ol.object_load_order_src_dir[idx]
+        local path = nil
+        if ol.object_load_order_src_dir[idx] then
+            path = ol.object_load_order_src_dir[idx].value
+        end
+        -- If path is still nil, try to construct a fallback path
+        if not path then
+            local mod_id = ol.object_load_order_id[idx].value
+            if mod_id then
+                -- Try to find the mod path using existing functions
+                path = getModSourcePath(mod_id)
+                if not path then
+                    -- If we can't find the path, assume it might be vanilla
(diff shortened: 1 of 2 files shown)


library/lua/script-manager.lua
```diff
@@ -214,14 +214,44 @@ function get_active_mods()
     local ol = df.global.world.object_loader
 
     for idx=0,#ol.object_load_order_id-1 do
-        local path = ol.object_load_order_src_dir[idx]
+        local path = nil
+        if ol.object_load_order_src_dir[idx] then
+            path = ol.object_load_order_src_dir[idx].value
```
ab9rf, line 219: `src_dir` is a `vector<path>` in 52.02 and so its elements do not have the `.value` metamethod. This code will not work with 52.02, or against develop

ab9rf (changes requested):

ab9rf:
Please confirm that the bug you're trying to fix here hasn't already been fixed in dev.

MalikMAlna:
Will do, if it's fixed in dev then I'll close out the PR here.

MalikMAlna:
Alright, it's resolved in dev, gonna close out this PR and the Issue. 

May I ask when you all plan to drop the change in as a release?

SilasD:
rome said this in Discord # planning
rome of oxtrot — 1:46 PM
at this point i'm planning on releasing 52.02-r2 with the fixes already committed. i have adequate UAT on everything that's been touched since 52.02-r1 (as far as i can tell) so i feel confident that it's safe to release. gonna wait till morning so as not to hit the saturday evening peak

MalikMAlna:
Cool, thank you both for your time and best of luck with the release to you all! 👋 😄

PR #5526, opened 2025-07-27
extend implementations of `addToSquad` and `removeFromSquad`

chdoc:
I have run the in-development `autotraining` for the better part of a season using this implementation, and I have not seen a single unit get stuck in a generic training session.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -66,6 +66,7 @@ Template for new versions:
 
 ## API
 - ``Filesystem::getBaseDir`` and ``Filesystem::getInstallDir`` added (and made available in Lua)
+- expand the partial implementations of ``Military::addToSquad`` and ``Military::removeFromSquad``
 
 ## Lua
 - inserting values into STL containers containing nonprimitive types is now supported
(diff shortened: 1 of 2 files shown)


ab9rf (approved):

quietust:
The original code is actually slightly different - rather than fetching the squad's global activity and removing the unit from ALL of its events, it should be fetching the unit's **position** within the squad and then removing the unit from each of the 3 activity/event IDs specified inside:

```
    auto squad = binsearch_in_vector(df::global::world->squads.all,&df::squad::id,unit->military.squad_id);
    auto position = squad->positions[unit->military.squad_position];
    for (int i = 0; i < 3; i++) { // iterate across possible values of enum "squad_event_type"
        auto activity = binsearch_in_vector(df::global::world->activities.all,&df::activity_entry::id,position->activities[i]);
        if (activity) {
            auto event = binsearch_in_vector(activity->events,&df::activity_event::id,position->events[i]);
            event->removeParticipant(unit->hist_figure_id, unit->id, false);
        }
    }
```

PR #5527, opened 2025-07-27
Update removeFromSquad to match in-game logic

quietust:
aside from the link removal stuff at the end, which might still be slightly different

Amends #5526

### DIFF
--- a/library/modules/Military.cpp
+++ b/library/modules/Military.cpp
@@ -477,42 +477,53 @@ bool Military::addToSquad(int32_t unit_id, int32_t squad_id, int32_t squad_pos)
 
 bool Military::removeFromSquad(int32_t unit_id)
 {
+    // based on unitst::remove_squad_info
     df::unit *unit = df::unit::find(unit_id);
     if (unit == nullptr || unit->military.squad_id == -1 || unit->military.squad_position == -1)
         return false;
 
+    // abort individual training
+    if (unit->individual_drills.size())
+        unit->flags3.bits.verify_personal_training = true;
+
     int32_t squad_id = unit->military.squad_id;
-    df::squad* squad = df::squad::find(squad_id);
-    if (squad == nullptr)
-        return false;
-
     int32_t squad_pos = unit->military.squad_position;
-    df::squad_position* pos = vector_get(squad->positions, squad_pos);
-    if (pos == nullptr)
-        return false;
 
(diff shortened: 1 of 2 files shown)


quietust:
Note: I don't have any fortresses with which to test this change - please verify before merging.

library/modules/Military.cpp
```diff
@@ -477,42 +477,53 @@ bool Military::addToSquad(int32_t unit_id, int32_t squad_id, int32_t squad_pos)
...
 
-    // remove from squad information
-    pos->occupant = -1;
+            // remove from squad position
+            pos->occupant = -1;
+        }
+    }
     // remove from unit information
     unit->military.squad_id = -1;
     unit->military.squad_position = -1;
```
chdoc, line 521: How about moving this before the `if (squad)` and turning the test into an early return if the squad is not defined, reducing the nesting level by one.

library/modules/Military.cpp
```diff
@@ -477,42 +477,53 @@ bool Military::addToSquad(int32_t unit_id, int32_t squad_id, int32_t squad_pos)
...
-    // remove unit from squad activities
-    auto activity_entry = binsearch_in_vector(df::global::world->activities.all,&df::activity_entry::id,squad->activity);
-    if (activity_entry) {
-        for (auto const activity_event : activity_entry->events) {
-            activity_event->removeParticipant(unit->hist_figure_id, unit->id, false);
-        }
-    }
+    // remove entity squad link
+    df::historical_figure* hf = df::historical_figure::find(unit->hist_figure_id);
+    if (hf == nullptr || squad == nullptr)
```
chdoc, line 525:
```suggestion
    if (hf == nullptr)
```
assuming the test for the squad is turned into an early return

quietust: Because the in-game logic will still clear `user->military.squad_id` + `user->military.squad_position` even if the squad doesn't actually exist anymore.

I suppose the case could be made for clearing those variables earlier, but I'm also replicating the **order** in which the game's own logic does it.

chdoc: Yes, I wasn't suggesting to change the behavior. As it comes to the order of things, I don't see any volatile variables. So the real order is whatever the compiler makes of it, both for our code and the game code. I would prefer to see less nesting and not fetching the `historical_figure` if there is no squad. But then, this is all well and truly in the eye of the beholder. What I truly care about is that `autotraining` works, which it seems to do with his implementation.

PR #5535, opened 2025-08-03
helper functions for creating and assigning jobs

chdoc:
I finally got around to following up on [this discussion](https://discord.com/channels/793331351645323264/873659717174587403/1289305706314596352) concerning `unitst::have_unbailable_sp_activities`

The motivation for this is that we now have several tools that create jobs and immediately assign them to dwarves (`idle-crafting`, `autocheese`, `immortal-cravings`). This has led to a bunch of code duplication and cross-importing between those tools. I would like to clean this up a bit by putting some of the frequently used functionality on a more solid footing and moving it into the C++ libraries.

@ab9rf It would be nice if you could take a look at `unbailableSocialActivity` and see if it needs any corrections or extensions for our purposes.

### DIFF
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -2337,7 +2337,7 @@ int32_t units_getFocusPenalty(lua_State *L) {
     for (int i = 2; i <= top; ++i) {
       try {
         needs.set(luaL_checkint(L, i));
-      } catch (const std::out_of_range &e) {
+      } catch ([[maybe_unused]] const std::out_of_range &e) {
         luaL_argerror(L, i, "Expected a need type");
       }
     }
--- a/library/LuaApi.cpp
+++ b/library/LuaApi.cpp
@@ -69,6 +69,7 @@ distribution.
 #include "df/building_civzonest.h"
 #include "df/building_stockpilest.h"
 #include "df/building_tradedepotst.h"
+#include "df/building_workshopst.h"
 #include "df/burrow.h"
 #include "df/caravan_state.h"
 #include "df/construction.h"
@@ -116,6 +117,7 @@ distribution.
 #include <string>
 #include <vector>
 #include <filesystem>
+#include <stdexcept>
 
 namespace DFHack {
     DBG_DECLARE(core, luaapi, DebugCategory::LINFO);
@@ -1902,6 +1904,8 @@ static const LuaWrapper::FunctionReg dfhack_job_module[] = {
     WRAPM(Job,disconnectJobItem),
     WRAPM(Job,disconnectJobGeneralRef),
     WRAPM(Job,removeJob),
+    WRAPM(Job,createLinked),
+    WRAPM(Job,assignToWorkshop),
     WRAPN(is_equal, jobEqual),
     WRAPN(is_item_equal, jobItemEqual),
     { NULL, NULL }
@@ -2128,6 +2132,8 @@ static const LuaWrapper::FunctionReg dfhack_units_module[] = {
     WRAPM(Units, setGroupActionTimers),
     WRAPM(Units, getUnitByNobleRole),
     WRAPM(Units, unassignTrainer),
+    WRAPM(Units, hasUnbailableSocialActivity),
+    WRAPM(Units, isJobAvailable),
     { NULL, NULL }
 };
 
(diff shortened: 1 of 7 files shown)


SilasD:
Wrong Discord link?  I don't see anything relevant near that post.

chdoc:
> Edit: I don't think I'd use the proposed job creation code. I want fine-grained control, and I want to add the job to the world as the last step, in case something goes wrong during my job creation.

Usually, I tend to check all the conditions that are necessary to create a job before attempting the creation. For tools that immediately attach workers, that includes looking whether there is an available worker. So it makes a lot of sense to keep the test separate from `addWorker`.

There isn't really much that can "go wrong" in setting job parameters and adding general references. In fact, if memory serves me right, DF links jobs into the world as part of the constructor. That is of course very restrictive because it doesn't allow you to store and pass around (unlinked) copies of jobs.

> I might like to have a findBestAvailableWorker(job), or simply addBestWorker(job). This would require some analysis of the job. 

If the goal is to get work done, then the preferred way is probably you just post the job and let the game handle the rest. `addWorker` is mainly for the case where you want a unit to do something that will satisfy one of its needs.

library/modules/Units.cpp
```diff
@@ -2017,6 +2021,92 @@ df::activity_event *Units::getMainSocialEvent(df::unit *unit) {
...
+
+    static const need_types_set socialize_needs = need_types_set()
+        .set(df::need_type::Socialize)
+        .set(df::need_type::BeCreative)
+        .set(df::need_type::Excitement)
+        .set(df::need_type::AdmireArt);
+
+    static const need_types_set read_needs = need_types_set()
+        .set(df::need_type::ThinkAbstractly)
+        .set(df::need_type::LearnSomething);
```
ab9rf: I would prefer that these be `constexpr` but `std::bitset`'s operators aren't `constexpr` until C++23, so I'm going to put this here as a vague reminder to reconsider if/when we move to C++23 compliance

ab9rf:
> Same for `unbailableSocialActivity()`, I don't like 'bail'. `hasUninteruptableSocialActivity()` ?

"bailable" is literally Tarn's term for this: this function is a (fairly faithful) reimplementation of a function in Bay12 code called `have_unbailable_sp_activities` and I'm fairly certain the reason @chdoc called it that is because I mentioned that that was Tarn's name for the function during the reverse engineering process that led to this implementation. I can see value in preferring Bay12 names for concepts, and thus I think we can live with using "bailable" instead of "interruptible". Not to mention people seem to have a lot of trouble spelling "interruptible" as it is.

> In fact, if memory serves me right, DF links jobs into the world as part of the constructor.

This is correct. DF just doesn't have a concept of a job that isn't on the master job list. While it might be nice to have a job that isn't linked in some situations, this doesn't agree with how DF conceptualizes jobs. While I don't think the game will actively misoperate if a job that isn't linked gets referred to, I wouldn't count on this and it's generally best that we not blithely violate the assumptions of the core game engine without _carefully_ considering what the consequence of such violations are. As such, I'm not in favor of having a utility function that creates jobs but doesn't link them, without a _specific_ demonstration of a meritorious use case for such a utility function. If the only situation you can come up with is "something could go wrong during job construction", the solution to _that_ is "verify that all of your preconditions are satisfied before you actually create the job".

library/modules/Units.cpp
chdoc: I share your sentiment. This is why I left a comment a few lines above the lines you selected.

library/LuaApi.cpp
```diff
@@ -2321,6 +2327,25 @@ static int units_getProfessionName(lua_State *L) {
...
+        needs.set(luaL_checkint(L, i));
+      } catch (const std::out_of_range &e) {
+        luaL_argerror(L, i, "Expected a need type");
+      }
+    }
+    Lua::Push(L, Units::getFocusPenalty(unit, needs));
+  }
+  return 1;
+}
+
```
ab9rf, line 2348: above and beyond, well done

library/LuaApi.cpp
```diff
@@ -2321,6 +2327,25 @@ static int units_getProfessionName(lua_State *L) {
...
+  auto unit = Lua::GetDFObject<df::unit>(L, 1);
+  Units::need_type_set needs;
+  auto top = lua_gettop(L);
+  if (top < 2) {
+    luaL_argerror(L, 2, "Expected at least one need type");
+  } else {
+    for (int i = 2; i <= top; ++i) {
+      try {
+        needs.set(luaL_checkint(L, i));
+      } catch (const std::out_of_range &e) {
```
ab9rf: omit the `e` here, since you don't use it

PR #5560, opened 2025-08-17
[nestboxes] allow limiting egg protection to a designated burrow

chdoc:
"Minor" change that I found useful while domesticating elk birds. The change is larger than I thought it would be, because `nestboxes` did not have any command-line parsing beforehand.


ab9rf:
I hate to be annoying but shouldn't the burrow setting be persisted?

ab9rf (approved):

Issue #5580, opened 2025-09-10
DFhack assign function on Pen/Pastures not working


turniprodeo:
<img width="409" height="112" alt="Image" src="https://github.com/user-attachments/assets/b8e9ea4d-70b2-42a4-8098-07bf2af28dd9" />

This button and hotkey to assign animals to a pen or pasture does not activate/work.

ab9rf:
relevant log file extract

```
...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:992: attempt to index a nil value (local 'civzone')
stack traceback:
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:992: in field 'get_status'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:642: in method 'cache_choices'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:709: in method 'get_choices'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:456: in local 'fun'
	...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:124: in upvalue 'invoke_after_rec'
	...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:153: in global 'AssignAnimal'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:889: in local 'fun'
	...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:124: in upvalue 'invoke_after_rec'
	...Steam\steamapps\common\Dwarf Fortress\hack\lua\class.lua:153: in global 'AssignAnimalScreen'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:1030: in upvalue 'show_pasture_pond_screen'
	...teamapps\common\Dwarf Fortress\hack\lua\plugins\zone.lua:1049: in field 'on_activate'
	...rf Fortress\hack\lua\gui\widgets\labels\hotkey_label.lua:65: in function 'gui.widgets.labels.hotkey_label.onInput'
	...)\Steam\steamapps\common\Dwarf Fortress\hack\lua\gui.lua:869: in function <...)\Steam\steamapps\common\Dwarf Fortress\hack\lua\gui.lua:856>
	(...tail calls...)
[...]

PR #5581, opened 2025-09-11
zone overlay -- tolerate corrupt animal-to-pasture links

SilasD:
This is a mitigation for #5580 "DFhack assign function on Pen/Pastures not working".  There is further discussion in that issue.

Based on the stacktrace, the issue was apparently caused by an animal with a corrupt link to a pasture zone or to a pit/pond zone.

The animal apparently had a general reference of type `BUILDING_CIVZONE_ASSIGNED` which had a `building_id` field pointing to a building that either did not exist or was not a civzone.

This patch tests for those conditions.  When detected, the animal is assigned a status of `NONE`, which causes the animal to be entirely hidden from the assign-to-pasture and assign-to-cage interfaces.

The changes made by this patch are minimal; they load and run; they mitigate the two ways I found to trigger the issue; but they **are not well-tested**.

I would like to have **a code review** with a focus on whether there are any other ways that line 992 could have triggered an error().

As I understand it: this is purely a bug workaround, so it doesn't need a changelog entry.  There's no visible change except for corrupt animal -> pasture links, which now don't abort the overlay.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -56,6 +56,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- `zone`: animal assignment dialog now tolerates corrupt animal-to-pasture links.
 
 ## Misc Improvements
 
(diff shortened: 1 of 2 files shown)


ab9rf:
I was going to suggest that we need to validate that the pointer is actually a pointer to a building but we're getting the pointer from the `building.find`, which is the only way we have to validate if a pointer is a building, so....

ab9rf (approved):

PR #5584, opened 2025-09-25
use noexcept std::filesystem overloads

dhthwy:
The overloads that take ec are the noexcept variants.

It now prints an error instead of crashing when an exception is thrown in Core::findScript() -  if the error msg is wanted.

```Error loading ''/home/dh/.local/share/Steam/steamapps/common/Dwarf Fortress/dfhack-config/scripts/lllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll.lua' (File name too long)```

Crash confirmed by Quietust.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -57,6 +57,7 @@ Template for new versions:
 
 ## Fixes
 - `zone`: animal assignment dialog now tolerates corrupt animal-to-pasture links.
+- improved file system handling: gracefully handle errors from operations, preventing crashes.
 
 ## Misc Improvements
 

### CODE REVIEW

File: library/modules/Filesystem.cpp [hunk 1]
```diff
@@ -154,17 +154,20 @@ bool Filesystem::stat (std::filesystem::path path, std::filesystem::file_status
 
 bool Filesystem::exists (std::filesystem::path path) noexcept
 {
-    return std::filesystem::exists(path);
+    std::error_code ec;
+    return std::filesystem::exists(path, ec);
```
ab9rf: As far as I can tell, the return value of `exists` is unspecified if an error is detected. I think I'd prefer for this function to return `false` if an error is detected, instead of returning an unspecified value.

library/modules/Filesystem.cpp
```diff
@@ -154,17 +154,20 @@ bool Filesystem::stat (std::filesystem::path path, std::filesystem::file_status
...
 }
 
 bool Filesystem::isfile(std::filesystem::path path) noexcept
 {
-    return std::filesystem::exists(path) && std::filesystem::is_regular_file(path);
+    std::error_code ec;
+    return std::filesystem::exists(path, ec) && std::filesystem::is_regular_file(path, ec);
```
ab9rf: Same comment as above.

library/modules/Filesystem.cpp
```diff
@@ -154,17 +154,20 @@ bool Filesystem::stat (std::filesystem::path path, std::filesystem::file_status
...
 }
 
 bool Filesystem::isdir (std::filesystem::path path) noexcept
 {
-    return std::filesystem::exists(path) && std::filesystem::is_directory(path);
+    std::error_code ec;
+    return std::filesystem::exists(path, ec) && std::filesystem::is_directory(path, ec);
```
ab9rf: Same comment as above.

ab9rf (changes requested):

dhthwy:
oops, forgot about the changelog

ab9rf (approved):

Issue #5591, opened 2025-10-04
manually invoking scripts in a scripts_modactive directory is failing


RoyalGreen32:
https://discord.com/channels/793331351645323264/807443277132595211/1424106396450033664

SilasD:
The issue is that you can't explicitly invoke a mod's script from the console or the launcher.  Autoload of modules does work.
(Modules are scripts with `--@module=true`, per the docs.)

This applies to scripts in all mods' `scripts_modactive` subdirectories, but *not* to scripts in all mods' `scripts_modinstalled` subdirectories.

SilasD:
diagnostic: load a game that has at least one mod that has a `scripts_modactive` subdirectory:
run this command at the console:
`:lua @ dfhack.internal.getScriptPaths()`
check if there are any lines ending in `scripts_modactive`

failing in 52.05
failing in 52.04
working in 51.08
failing in 52.03
~~failing *worse* in 52.02-r2~~ AppData vs portable.txt issue
working in 52.01
failing in 52.02-r2

SilasD:
Maybe traced to [`script-manager.lua` line 156](https://github.com/DFHack/dfhack/blob/b6872194f9f1ca4643064d8defd9137f8d981959/library/lua/script-manager.lua#L156C13-L156C79)
```Lua
if not path:startswith(INSTALLED_MODS_PATH) then goto continue end
```
which is comparing a relative path with an absolute path, so is always failing.

e.g. for the `vanilla_text` mod:
`data/vanilla/vanilla_text/` is compared with `C:/GAMES/DWARF FORTRESS/Z5205/data/installed_mods/`

(Note: this particular installation of Dwarf Fortress is in `C:\Games\Dwarf Fortress\z5205`.)

(tested in portable mode, which probably matters.)

PR #5592, opened 2025-10-05
Mitigate #5591

SilasD:
This is an attempt to mitigate issue #5591 .

The `scripts_modactive` and `scripts_modinstalled` directories of installed DFHack-enabled mods were not being searched for either auto-executing modules or manually invoked scripts.

This issue was traced to a relative path being compared with an absolute path.  PR #5514 changed mod-detection path processing to use absolute paths, but the list of active mods was still being retrieved from DF's list of active mods, which contains relative paths.

This was fixed by prepending the base directory (normally a subdirectory of AppData, but in Portable Mode, DF's own directory) to these paths.

This patch has been tested in both AppData Mode and Portable Mode.

**IMPORTANT!**  This patch has only been tested with DF Classic Windows.
It has not been tested with the Steam or Itch.io versions.
It has not been tested with any Linux versions.

It needs to be tested for all of these cases before being merged.

Testing: the [DFHack Example scripts](https://steamcommunity.com/sharedfiles/filedetails/?id=2945575775) mod was used as a sample case.  It was manually downloaded, as I don't have the Steam version of DF.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -59,6 +59,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- modding: the ``scripts_modactive`` and ``scripts_modinstalled`` subdirectories of mods will be searched for scripts as per the modding-guide documentation
 
 ## Misc Improvements
 
--- a/library/lua/script-manager.lua
+++ b/library/lua/script-manager.lua
@@ -152,6 +152,7 @@ function get_mod_paths(installed_subdir, active_subdir)
     -- if a world is loaded, process active mods first, and lock to active version
     if dfhack.isWorldLoaded() then
         for _,path in ipairs(df.global.world.object_loader.object_load_order_src_dir) do
+            path = dfhack.filesystem.getBaseDir() .. path
             -- skip vanilla "mods"
             if not path:startswith(INSTALLED_MODS_PATH) then goto continue end
             local id = get_mod_id_and_version(path)


docs/changelog.txt
```diff
@@ -59,6 +59,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- modding: the ``scripts_modactive`` and ``scripts_modinstalled`` subdirectories of mods will be searched for scripts as per the modding-guide documentation
```
ab9rf: suggest changing this to 
> script-manager: the `scripts_modactive` folder of a script-enabled mod will be properly added to the script path search list

library/lua/script-manager.lua
```diff
@@ -152,6 +152,7 @@ function get_mod_paths(installed_subdir, active_subdir)
     -- if a world is loaded, process active mods first, and lock to active version
     if dfhack.isWorldLoaded() then
         for _,path in ipairs(df.global.world.object_loader.object_load_order_src_dir) do
+            path = dfhack.filesystem.getBaseDir() .. path
```
ab9rf, line 155: btw, i confirmed with the SDL2 documentation that the trailing directory separator is guaranteed here so inserting another one is unnecessary

docs/changelog.txt
SilasD: I changed it as suggested, but kept the `scripts_modinstalled` -- that part is correct.

The reason that mods with a `scripts_modinstalled` were getting added to the search path is that there is [fallback code](https://github.com/DFHack/dfhack/blob/b6872194f9f1ca4643064d8defd9137f8d981959/library/lua/script-manager.lua#L168-L196) that hunts through the pre-installation mods in the `mods/` directory.

It adds any `scripts_modinstalled` subdirectories to the search paths if there isn't an *installed* version of that mod (with the same id number) in the `data/installed_mods/` directory.

If there is an exact installed version, it takes precedence.

But of course, the bug meant that no installed versions were found, so the fallback code kicked in.

PR #5602, opened 2025-10-24
Add stockpiles plugin support for dyes

### LINKED ISSUE #5556: DFhack stockpile tools like qui/quantum and dreamfort stockpile blueprints have limited functionality after stockpile updates in df version 52.03
New dye and color related updates to stockpiles have broken intended function of dfhack stockpile tools. Dreamfort stockpiles for trade goods, the clothing industry, and armor meltables are not compatible with 52.03. Gui/quantum does not appear to work at all when used with a feeder stockpile that has the color, dyed, undyed settings for the stockpile items.

### LINKED ISSUE #5586: stockpile import/export needs to be extended to support colors
The stockpile selection filters for color added in 52.03/04 are not supported by stockpile import/export

NicksWorld:
Enables support for "dyed", "undyed", and "color" fields for armor, cloth, leather, and finished goods.

The added protobuf fields are all `optional` or `repeated`, and as such old dfstock saved settings will remain usable. The only change in functionality for some of these old dfstock settings is that any categories set to `all` will now have `dyed`, `undyed` and all `color` selected.

Fixes #5586 

I have tested `library/cat_leather` and `gui/quantum`, both now set dyed, undyed, and all colors. I have limited understanding of quantum stockpiles, but I believe this fixes #5556.

### DIFF
--- a/plugins/stockpiles/StockpileSerializer.cpp
+++ b/plugins/stockpiles/StockpileSerializer.cpp
@@ -499,7 +499,7 @@ static void unserialize_list_organic_mat(color_ostream& out, const char* subcat,
 }
 
 static bool serialize_list_color(color_ostream& out, FuncWriteExport add_value, const vector<char>* colors) {
-    bool all = true;
+    bool all = world->raws.descriptors.colors.size() == colors->size();
     if (!colors) {
         DEBUG(log, out).print("serialize_list_color: list null\n");
         return all;
@@ -1508,14 +1508,14 @@ static bool finished_goods_mat_is_allowed(const MaterialInfo& mi) {
 }
 
 bool StockpileSettingsSerializer::write_finished_goods(color_ostream& out, StockpileSettings::FinishedGoodsSet* finished_goods) {
-    bool all = mSettings->leather.dyed && mSettings->leather.undyed;
+    bool all = mSettings->finished_goods.dyed && mSettings->finished_goods.undyed;
     finished_goods->set_dyed(mSettings->finished_goods.dyed);
     finished_goods->set_undyed(mSettings->finished_goods.undyed);
 
     all = serialize_list_item_type(out,
         finished_goods_type_is_allowed,
         [&](const string& token) { finished_goods->add_type(token); },
-        mSettings->finished_goods.type);
+        mSettings->finished_goods.type) && all;
 
     all = serialize_list_material(out,
         finished_goods_mat_is_allowed,


plugins/stockpiles/StockpileSerializer.cpp
```diff
@@ -497,6 +498,57 @@ static void unserialize_list_organic_mat(color_ostream& out, const char* subcat,
     }
 }
 
+static bool serialize_list_color(color_ostream& out, FuncWriteExport add_value, const vector<char>* colors) {
+    bool all = true;
+    if (!colors) {
+        DEBUG(log, out).print("serialize_list_color: list null\n");
+        return all;
+    }
+    for (size_t i = 0; i < colors->size(); ++i) {
```
ab9rf, line 507: consider ranged for here; you can use iterator difference to get the index

plugins/stockpiles/StockpileSerializer.cpp
```diff
@@ -497,6 +498,57 @@ static void unserialize_list_organic_mat(color_ostream& out, const char* subcat,
...
+        }
+        add_value(world->raws.descriptors.colors[i]->id);
+    }
+    return all;
+}
+
+// Find the index of a named color, returning SIZE_MAX on unknown.
+static size_t find_color_by_token(const string& token) {
+    auto& colors = world->raws.descriptors.colors;
+    size_t num_colors = colors.size();
```
ab9rf, line 520: this should definitely be ranged for

plugins/stockpiles/StockpileSerializer.cpp
```diff
@@ -497,6 +498,57 @@ static void unserialize_list_organic_mat(color_ostream& out, const char* subcat,
...
+    }
+    return SIZE_MAX;
+}
+
+static void unserialize_list_color(color_ostream& out, const char* subcat, bool all, char val, const vector<string>& filters,
+        FuncReadImport read_value, int32_t list_size, vector<char>& pile_list) {
+    size_t num_elems = world->raws.descriptors.colors.size();
+    pile_list.resize(num_elems, '\0');
+    if (all) {
+        for (size_t idx = 0; idx < num_elems; ++idx) {
```
ab9rf, line 534: i'd prefer ranged for here as well

plugins/stockpiles/StockpileSerializer.cpp
```diff
@@ -497,6 +498,57 @@ static void unserialize_list_organic_mat(color_ostream& out, const char* subcat,
...
+    for (size_t idx = 0; idx < num_colors; ++ idx) {
+        if (colors[idx]->id == token) {
+            return idx;
+        }
+    }
+    return SIZE_MAX;
+}
+
+static void unserialize_list_color(color_ostream& out, const char* subcat, bool all, char val, const vector<string>& filters,
+        FuncReadImport read_value, int32_t list_size, vector<char>& pile_list) {
```
ab9rf, line 530: we should consider if these should be `int8_t` instead of `char`, although this is a structures issue rather than anything directly relevant to this PR (not a blocker)

ab9rf (changes requested):
I haven't moved on this myself because I wasn't sure how to ensure that older export files would be loadable, and would definitely like confirmation that it'll work

plugins/stockpiles/StockpileSerializer.cpp
```diff
@@ -1425,7 +1508,11 @@ static bool finished_goods_mat_is_allowed(const MaterialInfo& mi) {
...
 bool StockpileSettingsSerializer::write_finished_goods(color_ostream& out, StockpileSettings::FinishedGoodsSet* finished_goods) {
-    bool all = serialize_list_item_type(out,
+    bool all = mSettings->leather.dyed && mSettings->leather.undyed;
+    finished_goods->set_dyed(mSettings->finished_goods.dyed);
+    finished_goods->set_undyed(mSettings->finished_goods.undyed);
+
+    all = serialize_list_item_type(out,
         finished_goods_type_is_allowed,
         [&](const string& token) { finished_goods->add_type(token); },
         mSettings->finished_goods.type);
```
SilasD: To my non-expert eye, it looks like the `all` value set in line 1511 is clobbered by the `all` value set in 1515-1518.
The latter statement is missing the `&& all` that is used in the statements below.

PR #5625, opened 2025-11-11
Try to make Core be more modern per c++20

dhthwy:
There are definitely parts that need some refactoring, I think.

The intent was not to change any behavior or risk changing behavior, so refactoring was kept to minimum.


dhthwy:
I tested most of this. Ought to have tests for all the commands. Manual checking sucks.

The keybinding commands work as expected with std::views::drop and std::views::reverse.

ab9rf:
I'm not ignoring this PR, I just haven't gotten to it yet. You needn't worry about messing up the PR queue.

library/Core.cpp
```diff
@@ -315,29 +316,31 @@ static std::string dfhack_version_desc()
...
     lua_remove(state, -2);
     lua_pushstring(state, pcmd.c_str());
-    for (auto& arg : pargs)
+    for (const auto& arg : pargs)
         lua_pushstring(state, arg.c_str());
     return true;
 }
 
-static command_result runLuaScript(color_ostream &out, std::string name, std::vector<std::string> &args)
+static command_result runLuaScript(color_ostream &out, std::string name, const std::span<const std::string> args)
```
ab9rf: could name here be a `string_view` instead?

library/Core.cpp
```diff
@@ -315,29 +316,31 @@ static std::string dfhack_version_desc()
...
+        return init_run_script(out, state, n, args);
+    };
+
     bool ok = Lua::RunCoreQueryLoop(out, DFHack::Core::getInstance().getLuaState(true), init_fn);
 
     return ok ? CR_OK : CR_FAILURE;
 }
 
-static bool init_enable_script(color_ostream &out, lua_State *state, std::string& name, bool enable)
+static bool init_enable_script(color_ostream &out, lua_State *state, const std::string& name, bool enable)
```
ab9rf: consider using `string_view` here instead, and change `lua_pushstring(state, name.c_str())` in line 350 to `lua_pushlstring(state, name.data(), name.size())` 

in general we should use `lua_pushlstring` in favor of `lua_pushstring` whenever possible

library/Core.cpp
```diff
@@ -668,14 +673,14 @@ void tags_helper(color_ostream &con, const std::string &tag) {
     }
 }
 
-void ls_helper(color_ostream &con, const std::vector<std::string> &params) {
+static void ls_helper(color_ostream &con, const std::span<const std::string> params) {
     std::vector<std::string> filter;
     bool skip_tags = false;
     bool show_dev_commands = false;
-    std::string exclude_strs = "";
+    std::string exclude_strs;
```
ab9rf: can be  `string_view`; no need to make another copy of the string here

library/Core.cpp
```diff
@@ -2151,43 +2151,44 @@ void Core::onUpdate(color_ostream &out)
...
-        typedef std::vector<Entry> EntryVector;
-        typedef std::map<Key,Val> InitVariationTable;
+        using Key = state_change_event;
+        using Val = std::vector<std::string>;
+        using Entry = std::pair<Key,Val>;
+        using EntryVector =  std::vector<Entry>;
+        using InitVariationTable = std::map<Key,Val>;
 
-        EntryVector computeInitVariationTable(void* none, ...) {
+        static EntryVector computeInitVariationTable(void* none, ...) {
```
ab9rf: I would love to see this use of C-style variadics go away

dhthwy: It should have the same behavior. This is from a test print:

```
placing key 0: onLoad,onLoadWorld,onWorldLoaded
placing key 1: onUnload,onUnloadWorld,onWorldUnloaded
placing key 2: onMapLoad,onLoadMap
placing key 3: onMapUnload,onUnloadMap
```

I removed the trailing empty strings. I don't think that's needed anymore.

i.e. removed ``""`` at the end.  
```
SC_WORLD_LOADED, "onLoad", "onLoadWorld", "onWorldLoaded",
```

That was the easiest way I could think of to do it.

PR #5627, opened 2025-11-12
Add copy building overlay

robob27:
TODO: fill this out before taking out of draft

Add a "Copy" button to buildings using an overlay that allows quickly getting into placement mode for a building of the same type.


ab9rf:
Concerned that this seems highly dependent on vanilla raws and on Bay12 not making changes to the interface

ab9rf:
This has been dormant for nearly a year. Is there still interest in this PR?

PR #5628, opened 2025-11-12
remove lua.h dependency from vtableinterpose.h

### LINKED ISSUE #5517: including `VTableInterpose.h` forces a dependency on lua for no good reason
There's no reason why including `VTableInterpose.h` should end up requiring `lua.h`, but our includes are a tangled mess and right now it does.

This issue is opened as a reminder that this should someday be fixed.

dhthwy:
may resolve #5517

I believe this does the trick.  Just added some wrappers so that ``DataIdentity.h`` doesn't have to include ``LuaWrapper.h`` which in turn includes ``lua.h``.

sadly ``ninja -d stats`` didn't report any measurable difference in build times.

### DIFF
--- a/library/DataIdentity.cpp
+++ b/library/DataIdentity.cpp
@@ -9,6 +9,7 @@
 
 #include "DataFuncs.h"
 #include "DataIdentity.h"
+#include "LuaWrapper.h"
 
 // the space after the uses of "type" in OPAQUE_IDENTITY_TRAITS_NAME is _required_
 // without it the macro generates a syntax error when type is a template specification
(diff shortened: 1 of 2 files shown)


library/DataIdentity.cpp
```diff
@@ -62,4 +63,19 @@ namespace df {
...
+
+    void* get_object_internal_(lua_State* state, type_identity* id, int val_index, bool val)
+    {
+        return DFHack::LuaWrapper::get_object_internal(state, id, val_index, val);
+    }
+
+    void field_error_(lua_State *state, int index, const char* err, const char* mode)
+    {
+        DFHack::LuaWrapper::field_error(state, index, err, mode);
+    }
```
ab9rf, line 80: These introduce potential pessimizations by preventing the compiler from inlining these at their call sites. Since at least two of these functions are used in potentially high-velocity code paths, I'm reluctant to approve this without runtime profiling

that said, i just tested dfhack compiled with LTCG and it seems to work on Windows, which would obviate this concern if we can confirm that using LTCG doesn't break dfhack, and someone can figure out how to do the same thing for linux users

library/include/DataIdentity.h
```diff
@@ -393,6 +391,10 @@ namespace df
     };
 #endif
 
+    type_identity* lua_touserdata_(lua_State* state);
```
lethosor, line 394: are the trailing underscores necessary? can we not just fully-qualify the names when needed (or would that require changing too many places)?

dhthwy: These are forwarding functions (intended for ``DataIdentity`` only) used by ``DataIdentity.c`` so that we can include  ``LuaWrapper.h`` which in turn includes ``lua.h``  in the DataIdentity TU instead of the header.

the real ``lua_touserdata()`` takes another argument too.

``_wrapper`` or something similar is another alternative. I wasn't sure what name to give it, hence the underscores.

I don't think they're necessary?

dhthwy:
LTCG breaks dfhack on Linux ( tested on gcc 15.2 ).

Going to close this PR for now.

PR #5629, opened 2025-11-12
try to cleanup unused includes from headers

dhthwy:
We'll see how this goes.

gcc/clang has been lying to me. Seems standard types like std::string don't require an #include in my version.


dhthwy:
It was actually the linters I was referring to (namely clangd), and I think it was the templates it was struggling with.

Issue #5652, opened 2025-11-19
BitArray class uses wrong memory allocation functions


quietust:
DFHack's `BitArray<>` class is a reimplementation of Toady's `flagarrayst` with a few additional features. However, it is *not* strictly compatible with it because it uses `realloc()` and `free()` to manage the flag data buffer, while Toady's version uses `new unsigned char[]` and `delete[]`.

In practice, this "works" on all of our supported platforms, but tools such as valgrind rightfully complain that we're mixing memory management functions incorrectly.

PR #5653, opened 2025-11-20
rewrite `BitArray` to be a little less crusty

ab9rf:
and no malloc/free/realloc neither

### DIFF
--- a/library/include/BitArray.h
+++ b/library/include/BitArray.h
@@ -79,7 +79,7 @@ namespace DFHack
 
         void extend(T index)
         {
-            size_type newsize = (index / 8) + 1;
+            size_type newsize = (index + 7 ) / 8;
             if (newsize > _size)
                 resize(newsize);
         }


library/include/BitArray.h
```diff
@@ -162,23 +206,44 @@ namespace DFHack
...
+        void resize(unsigned new_size, const DfArray<T>* replacement)
+        {
+            if (new_size == m_size)
+                return;
+
+            T* old_data = m_data;
+
+            m_data = (T*) new T[new_size];
+
+            T* copysrc = replacement ? replacement->m_data : m_data;
```
NicksWorld: Shouldn't this copy from `old_data`, not our newly created `m_data`?

library/include/BitArray.h
```diff
@@ -162,23 +206,44 @@ namespace DFHack
...
+            unsigned short copysize = replacement ? replacement->m_size : m_size;
+
+            if (copysrc)
+                std::memcpy(m_data, copysrc, sizeof(T) * copysize);
```
NicksWorld: `copysize` is calculated from our old size, which means `m_size` > `new_size` will cause out-of-bounds writes.

library/include/BitArray.h
```diff
@@ -27,132 +27,176 @@ distribution.
...
+                std::memset(_bits + _size, 0, newsize - _size);
+
+            delete[] old_data;
+
+            _size = newsize;
+        }
+
+        void extend(T index)
+        {
+            size_type newsize = (index / 8) + 1;
```
quietust: An index of `0` will allocate 1 byte, and an index of `8` will allocate 2 bytes, which we probably don't want - `(index + 7) / 8` might give more sensible results.

quietust:
This will fix #5652

PR #5659, opened 2025-11-22
refactor `Core`

ab9rf:
Core is altogether too large and contains too many only vaguely related parts - this PR is an attempt to make it more tractable


ab9rf:
I "inadvertently" recovered the ability to use `fpause` to pause worldgen while doing this refactor, so this change does have user-visible effects, hence the changelog

dhthwy:
lgtm

PR #5661, opened 2025-11-23
add fmtlib

ab9rf:
adds fmtlib as a dependency
adds `format` and `format_err` template functions to `color_ostream`


CMakeLists.txt
```diff
@@ -347,6 +347,14 @@ if(BUILD_LIBRARY)
     endif()
 endif()
 
+INCLUDE(FetchContent)
+FetchContent_Declare(
+    fmt
+    GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+    GIT_TAG master
```
lethosor: This is not a stable ref.

Is it possible to vendor a specific version like we do in `depends/` for some other libs?

ab9rf: We can select a specific tag, yes - 12.1.0 is the most recent release and if we selected that tag (instead of "master") it would use that as a stable pin

ab9rf: well, except 12.1.0. throws a warning while compiling in msvc, so either i have to suppress that warning while building fmtlib or find a reference that does _not_ do that

CMakeLists.txt
```diff
@@ -347,6 +347,17 @@ if(BUILD_LIBRARY)
     endif()
 endif()
 
+# this can be made conditional once we get to better platform support for std::format
+INCLUDE(FetchContent)
```
lethosor, line 351: FWIW, it looks like we previously ran into issues with including libzip/xlsxio via `ExternalProject_Add`, and that's why they are vendored like they are now. Maybe those issues were largely on the Buildmaster runner? I don't quite remember.

This is mainly a heads-up that there could be unforeseen issues with pulling in libraries in other ways.

Also, it may be good to document any necessary extra steps to compile DFHack offline in `docs/dev/compile/Compile.rst`

ab9rf: I'd be curious to know more about that. `FetchContent` is relatively new to CMake (3.11, released 2018) and I can't find any past mention of it in our github. (We didn't even start requiring CMake 3.11 until 2023; prior to moving to 3.18 in May 2023, our minimum requirement was 3.6, which would not have supported `FetchContent`.) This PR does not use `ExternalProject`. I was able to find some discussion in our history around using `ExternalProject`, and the problems that arose there are a large part of why `FetchContent` has largely replaced `ExternalProject` as the preferred means for pulling in external dependencies in CMake these days. 

[This article](https://discourse.cmake.org/t/relationship-between-fetchcontent-and-externalproject/10162), on the differences between `FetchContent` and `ExternalProject`, may be informative reading.

FetchContent ordinarily requires a connection during the first build, but after that subsequent builds (once the dependency has been fetched) can be disconnected. The `FETCHCONTENT_FULLY_DISCONNECTED` and `FETCHCONTENT_UPDATES_DISCONNECTED` CMake defines can be used to control behavior in this instance; see [Cmake documentation](https://cmake.org/cmake/help/latest/module/FetchContent.html). First-time offline builds will require a [CMake dependency provider](https://cmake.org/cmake/help/latest/command/cmake_language.html#dependency-providers) pointed at a local repository. (A simple external provider is available at that documentation page.) I suspect such scenarios are increasingly rare today, and I'd argue that in today's day and age, offline building is _far_ less important than it used to be.

PR #5667, opened 2025-11-26
fix debug plugin use-after-free on shutdown (LTO)

dhthwy:
This only popped when compiling with LTO (dfhack library, as well as dfhack library + plugins).

For some reason DebugManager's instance was being destroyed before the debug plugin's FilterManager, which lead to a use-after-free in Signal.

A 'proper fix' might be to fix the Signal library, but since DebugManager is a singleton, I think it's ok to just leak for these types.


dhthwy:
pre-commit.ci autofix

PR #5670, opened 2025-11-27
Add new plugin edgescroll to automate panning of game/region maps

NicksWorld:
This adds a new plugin that when enabled can automatically pan the game world and various region maps throughout the game.
The particular supported screens are:
- Fortress/Adventure Game World
- Worldgen Preview
- Embark (overview, and zoomed)
- World screen

In action:

https://github.com/user-attachments/assets/0dd48b26-6cee-4126-909e-62b867dd2c2f

https://github.com/user-attachments/assets/0034534e-c45c-4069-9e9c-8f66a655a849

### DIFF
--- a/plugins/edgescroll.cpp
+++ b/plugins/edgescroll.cpp
@@ -56,15 +56,15 @@ DFhackCExport command_result plugin_shutdown([[maybe_unused]] color_ostream &out
 static std::atomic_bool callback_queued = false;
 
 struct scroll_state {
-    int8_t xdiff;
-    int8_t ydiff;
+    int8_t xdiff = 0;
+    int8_t ydiff = 0;
 };
 
 static scroll_state state;
 static scroll_state queued;
 
 static void render_thread_cb() {
-    queued = {0};
+    queued = {};
     // Ignore the mouse if outside the window
     if (!enabler->mouse_focus) {
         callback_queued.store(false);
@@ -107,7 +107,7 @@ static bool update_mouse_pos() {
 
     // Queued callback complete, save the results and enqueue again
     state = queued;
-    queued = {0};
+    queued = {};
     DFHack::runOnRenderThread(render_thread_cb);
     callback_queued.store(true);
     return true;


plugins/edgescroll.cpp
```diff
@@ -0,0 +1,223 @@
...
+    queued = state_default;
+    // Ignore the mouse if outside the window
+    if (!enabler->mouse_focus) {
+        request_queued.store(false);
+        return;
+    }
+
+    // Determine window border location in window coordinates
+    auto* renderer = virtual_cast<df::renderer_2d>(enabler->renderer);
+    int origin_x, origin_y = 0;
```
dhthwy:
```suggestion
    int origin_x, origin_y;
```

For consistency.

plugins/edgescroll.cpp
```diff
@@ -0,0 +1,232 @@
...
+    callback_queued.store(false);
+}
+
+static bool update_mouse_pos() {
+    if (callback_queued.load())
+        return false; // Queued callback not complete, check back later
+
+    // Queued callback complete, save the results and enqueue again
+    state = queued;
+    queued = {0};
```
ab9rf: why `{0}` here? either use `{}` for a default-initialized temporary, or `{0,0}` for a fully initialized object. using `{0}` results in an incompletely initialized temporary, which means accessing `.ydiff` in that temporary would lead to UB, and since the (default) copy-assignment operator _immediately_ accesses that member, this code has UB

note that using `{}` also yields UB, again because these member variables have no specified default values and the class's constructor is also defaulted, which means these variables are default-initialized and (being as they are fundamental types) will thus be uninitialized and so any attempt to access their values will yield UB. thus, if you intend to use `{}` you need to add either a defined constructor to `scroll_state` or add member variable defaults to `.xdiff` and `.ydiff`. i obviously recommend the latter

plugins/edgescroll.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+DFhackCExport command_result plugin_shutdown([[maybe_unused]] color_ostream &out) {
+    return CR_OK;
+}
+
+static std::atomic_bool callback_queued = false;
+
+struct scroll_state {
+    int8_t xdiff;
+    int8_t ydiff;
```
ab9rf: these should probably have explicit default values:
```
int8_t xdiff = 0;
int8_t ydiff = 0;
```

PR #5672, opened 2025-11-30
Exclude signal handler from linux crashlogs

NicksWorld:
Currently all crashlogs produced on Linux end with a opaque offset within `libdfhack.so`, which can cause some confusion.
```
Signal SIGSEGV
0> hack/libdfhack.so(+0x936f90) [0x7b040d336f90]
```

This PR causes the crashhandler to skip printing the first backtrace entry, which will *always* be `dfhack_crashlog_handle_signal` or `dfhack_crashlog_handle_terminate`, where the backtrace is retrieved. To ensure exactly one crashhandler-related entry is added to the stack frames, `handle_signal_internal` is marked with `[[gnu::always_inline]]` to ensure it doesn't create a stack frame.

Alternatively, we can clear up the opaque entry by enabling symbol visibility on `dfhack_crashlog_handle_(signal|terminate)` by marking them with `DFHACK_EXPORT`, with the following result:

```
Signal SIGSEGV
0> hack/libdfhack.so(dfhack_crashlog_handle_signal+0x62) [0x7fffe3955052]
```

### DIFF
--- a/library/Crashlog.cpp
+++ b/library/Crashlog.cpp
@@ -111,10 +111,6 @@ std::filesystem::path get_crashlog_path() {
 
 void dfhack_save_crashlog() {
     char** backtrace_strings = backtrace_symbols(crash_info.backtrace, crash_info.backtrace_entries);
-    if (!backtrace_strings) {
-        // Allocation failed, give up
-        return;
-    }
     try {
         std::filesystem::path crashlog_path = get_crashlog_path();
         std::ofstream crashlog(crashlog_path);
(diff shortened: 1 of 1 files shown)


library/Crashlog.cpp
```diff
@@ -122,8 +128,9 @@ void dfhack_save_crashlog() {
             crashlog << "Signal " << signal << "\n";
         }
 
-        for (int i = 0; i < crash_info.backtrace_entries; i++) {
-            crashlog << i << "> " << backtrace_strings[i] << "\n";
+        // Skip the first backtrace entry as it will always be dfhack_crashlog_handle_(signal|terminate)
+        for (int i = 1; i < crash_info.backtrace_entries; i++) {
+            crashlog << i - 1 << "> " << backtrace_strings[i] << "\n";
```
ab9rf: this code will print nothing if -- somehow -- the backtrace only has one entry. this can happen if the stack is corrupt at the time of the crash. i would advise printing a "omg disaster happened" message in the event that `crash_info.backtrace_entries` has fewer than two entries, rather than printing nothing as this is potentially slightly less confusing

the none at all case is handled above, and probably should also emit some sort of omg message, again to reduce user confusion. maybe something like "stack trace unavailable". if the API used to acquire the stack trace has any error indicators, these should probably be emitted when they obtain

ab9rf (approved):

Issue #3346, opened 2023-05-03
[suggestion] add search and further sort functionality to the Work Order interface


Totherphoenix:
The current sort function will send any non-repeatable work orders to the top of the list - this is amazing and exploitable to set up my priorities without excess clicking. However, once my work orders hit the 2-3 pages mark, we revert back into the territory of it being the worst UI in the game. Any chance DFHack could add a search function to this screen to allow us to search for existing work orders, as well as further sort functionality (by job, by alphabet, by material, etc.). Also, and finally, an error-check for duplicate work orders would be a great QoL addon - giving me an alert if I've added a work order that already exists in order to keep the work order list de-duped.

ab9rf:
search filtering added by #5677 

some of the other feature suggestions here remain open

Issue #1910, opened 2021-07-25
Feature request: searchable Work Orders


TimurKelman:
Search functionality is in so many other views, but unfortunately is missing from Work Orders.

lethosor:
The main reason this one is trickier to implement properly is because the screen doesn't actually maintain its own list of manager orders - they are _only_ stored in `world.manager_orders`. The screen just maintains its cursor position, mode, etc.

There are some other search implementations that search over global data that isn't part of the current screen (e.g. the `k` search, aka `look_menu_search`), so it's possible, but tricky to do right in this case because the data persists well beyond the lifetime of the screen (in contrast, the data used by the `k` view, while global, is only really valid for a short duration). It's further complicated by the fact that orders can refer to each other, so if you filter out a dependency of another order and open it with `Enter`, you would get something that looks like this:

![image](https://user-images.githubusercontent.com/3719547/126925712-2f058d1c-46ea-456d-adf4-4a515d82f148.png)

There should be an order name above "must". There might be other cases where DF crashes due to entries being removed (even temporarily) from `world.manager_orders` to enable searching.

Short version: it might be possible, but it's more complicated to do safely.

ab9rf:
At a minimum, you'd have to ensure to not remove any order that is dependent on the orders that match the search.

Also, you'd have to restore the full list when you go into the detail screen because the user might choose to create an order dependency while in the detail screen, and that needs the full list available. This may cause problems since the detail screen uses an index into the global order list. I suppose you could restore the full list only if the user asks to create a order dependency, but... ugh.

myk002:
an alternative might be to just *highlight* orders that match a search, and provide a hotkey (or just override the behavior of the arrow keys) to jump the cursor to the next highlighted match. this would be a different look-and-feel from the other search interfaces, but it would avoid all the rather scary pitfalls of dynamically modifying the global list.

ab9rf:
closed in #5677

PR #5677, opened 2025-12-02
orders: add search overlay for manager orders

pajawojciech:
# Add search to manager orders

Adds search box to find orders by job name + arrow indicators to show which one you're looking at.

## What it does

- Search box appears on orders screen (Alt+S to focus)
- Type keywords to filter (e.g. "iron picks" finds "forge iron picks")
- Multi-word search works in any order ("picks iron" and "pi ir" also matches)
- Alt+P/N to jump between matches. Also Enter and Shift+Enter works when focused
- Shows "3 of 12" style counter
- Arrow point at current match (highlighted) and at all matches

## Implementation

- Uses df::interface_button_building_new_jobst to get order name
- New overlay: `OrdersSearchOverlay` enabled by default 
- Moved `importexport` overlay position to make room for search (and make new version of config)

## What didn't work

- Tried to clear search input when exiting orders window - couldn't find reliable way to detect window exit

## Related issues

- https://github.com/DFHack/dfhack/issues/1910
- https://github.com/DFHack/dfhack/issues/3346

https://github.com/user-attachments/assets/d4e5f219-2ecf-403b-8fdc-47896f5934c1

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -116,7 +116,7 @@ Template for new versions:
 ## Documentation
 
 ## API
-- `dfhack.job.getManagerOrderName()`: New function to get the display name of a manager order
+- ``dfhack.job.getManagerOrderName``: New function to get the display name of a manager order
 
 ## Lua
 


plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+
+    return string.format('%d:%d:%d:%d:%d:%s:%s:%s',
+        order.job_type,
+        order.item_type,
+        order.item_subtype,
+        order.mat_type,
+        order.mat_index,
+        order.reaction_name or '',
+        mat_cat_str,
+        encrust_str)
```
ab9rf: this results in lexicographic ordering by the text representation of job type's number, which means 0, 1, 10, 11, 12, ..., 19, 2, 20, 21...? while this is a stable sort order, it is not an order the player will likely anticipate

plugins/lua/orders.lua
```diff
@@ -74,7 +79,7 @@ local mi = df.global.game.main_interface
 OrdersOverlay = defclass(OrdersOverlay, overlay.OverlayWidget)
 OrdersOverlay.ATTRS{
     desc='Adds import, export, and other functions to the manager orders screen.',
-    default_pos={x=53,y=-6},
+    default_pos={x=41,y=-6},
```
ChrisJohnsen, line 77: The new default position can be forced by setting/changing the overlay's `version`, but that will also re-enable this overlay for anyone that has previously disabled it, so it isn't always appropriate. But it may be better than players having to seek out how to adjust/reset the overlay's position.

plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    return true
+end
+
+OrdersSearchOverlay = defclass(OrdersSearchOverlay, overlay.OverlayWidget)
+OrdersSearchOverlay.ATTRS{
+    desc='Adds a search box to find and navigate to matching manager orders.',
+    default_pos={x=85, y=-6},
+    default_enabled=false,
+    viewscreens='dwarfmode/Info/WORK_ORDERS/Default',
+    frame={w=34, h=4},
```
ChrisJohnsen: This combination of position and size does not properly fit in a minimum-size DF interface (114x46 UI cells, 912x552px with the default 8x12 UI font). The x-offset of 84 plus width of 34 puts the right edge at 118 columns.

This isn't a problem in "wide" DF windows, but if the the interface ends up narrower than 118 (due to window size and/or DF interface scaling percentage), then the `overlay.search` window will be pushed to the left. In a minimum-width interface this will cause the `orders.search` and `orders.importexport` windows (at their default positions established in this commit) to overlap by three columns (which one is on top varies across startups due to the Lua `pairs` iteration done for overlay rendering).

The leftward push experienced while the DF interface is narrower than 118 columns will remain until the overlay is reloaded (e.g., the player reloads the fort (via overlay's `onStateChange` handler), or `:lua require'plugins.overlay'.rescan()`). Only the widget's frame table is affected, not its persisted overlay configuration.

---

I don't know if it is standard procedure for DFHack (I'm just a contributor here), but I always try out overlay stuff in a variety of window sizes (especially a minimum-size window, which seems to heavily influences DF's own layout, and thus DFHack's DF-relative overlays) to make sure things will fit.

For this overlay, it looks like it could be narrowed by four columns without affecting its utility too much. Actually, I would recommend narrowing by six columns to keep the right edge away from the right edge of the (minimum-width) interface area (which will cause issues if `gui/overlay` is used to move the overlay window in a minimum-width interface: it will automatically swap to a right-anchored position which (in wider interfaces) will cause the overlay to "float" to the right instead of "sticking with" the (default position) of the `orders.importexport` overlay window).

plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+-- -------------------
+-- OrderHighlightOverlay
+-- -------------------
+
+OrderHighlightOverlay = defclass(OrderHighlightOverlay, overlay.OverlayWidget)
+OrderHighlightOverlay.ATTRS{
+    desc='Shows arrows next to the work order found by orders.search',
+    default_enabled=false,
+    viewscreens='dwarfmode/Info/WORK_ORDERS/Default',
+    frame={w=80, h=3},
```
ChrisJohnsen: This should probably be `full_interface=true` instead of using a frame. This prevents an "phantom window" from being configurable in `gui/overlay`.

plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+
+function OrderHighlightOverlay:getViewportSize()
+    local rect = gui.get_interface_rect()
+    local list_start_y = self:getListStartY()
+
+    local available_height = rect.height - list_start_y - self.BOTTOM_MARGIN
+    return math.floor(available_height / self.ORDER_HEIGHT)
+end
+
+function OrderHighlightOverlay:calculateSelectedOrderY()
```
ChrisJohnsen: There is a bug here when deleting an order above the highlighted order. Here is a way to reproduce the problem:

1. Have more than one screen-full (page) of orders (i.e., the scrollbar is active).
1. Have a searchable order at or near the bottom of the the last page of orders.
1. Search for the order.
1. Move highlight to the order that is on the last page.

    The list should now be scrolled to the bottom of the list and there should be other orders above the highlighted order.

1. Delete an order above the highlighted order.

    The highlight is now drawn on the wrong row (or not drawn if it was "pushed off the bottom").

----

You might fix this by canceling the highlight (and invalidating the search -- or re-doing the search to update the match count) if the number of orders changes.

plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+
+    for _, reaction in ipairs(reactions) do
+        local key = make_order_key(reaction.order)
+        map[key] = reaction.name:lower()
+    end
+
+    return map
+end
+
+local reaction_map_cache = nil
```
ChrisJohnsen: I kept noticing a stutter when doing my first search of a bootup (I would type "plant" and end up with "pnt"). I think it is this cache being populated with ~3500 entries. Not sure if anything can be done to smooth that out. Maybe I am just running on a potato?

plugins/lua/orders.lua
```diff
@@ -709,11 +714,401 @@ function QuantityRightClickOverlay:onInput(keys)
...
+        if search_last_scroll_position ~= -1 and current_scroll ~= search_last_scroll_position then
+            search_cursor_visible = false
+        end
+        search_last_scroll_position = current_scroll
+    end
+
+    OrderHighlightOverlay.super.render(self, dc)
+end
+
+function OrderHighlightOverlay:onRenderFrame(dc, rect)
```
ChrisJohnsen: It probably isn't a big deal, but there doesn't seem to be a reason to have both `render` and `onRenderFrame`? Putting everything in one (probably `render`, since this is rightfully a `full_interface` overlay widget without any frame) worked okay when I tried it briefly.

plugins/lua/orders.lua
ab9rf: 3500 entries in a lua map is not something we want to be doing lightly

lethosor: I would expect building the 3500 hash entries is probably the more expensive part, rather than inserting them into the hash. But it would be interesting to profile. I know you've been looking at lua table performance issues recently, but building a single hash in an application like this does not concern me much, at least from a runtime perspective, based partly on some simple "populate hash with 10000 elements"-type tests.

ab9rf: i have concerns about lua hash tables, but in this context i don't actually think it's a serious issue as long as the hash table is "initialize-once". while lua's hashing strategy leads to a relatively high collision rate, the fact that lua hash tables are contiguously allocated means that collisions will usually resolve without requiring more than a L1 cache stall, and thus won't be very expensive, so the performance impact of lua's hypothetically terrible hashing approach isn't actually that bad on modern architectures. mostly i want to ensure that the table isn't getting copied or modified after it's originally constructed, since _those_ would be expensive operations

plugins/lua/orders.lua
pajawojciech: Thanks, I bumped the version on importexport to force the new position.

plugins/lua/orders.lua
pajawojciech: I narrowed the overlay by 8 columns (now w=26) so it fits in minimum-size windows and still has room for 16 chars in the search field.

plugins/lua/orders.lua
pajawojciech: I changed it to full_interface=true and removed the frame since it's not needed.

plugins/lua/orders.lua
pajawojciech: I added a check to hide the highlight when the order count changes

plugins/lua/orders.lua
pajawojciech: @ChrisJohnsen Good catch! I run the cache building on init so it happens on overlay load instead of during first keystroke, which eliminates that stutter

I'll also look into the performance and timing of building the entries and creating the hashtable.

plugins/lua/orders.lua
pajawojciech: You're right, I consolidated everything into the render method since there's no need for both.

pajawojciech:
>     * Would utils.search_text be useful here? The main SortOverlay uses it for searching.
I switched to using utils.search_text instead of the custom search logic
 
>     * Could the first match automatically be highlighted on Enter?
Added Enter/Shift+Enter to cycle through matches using the default submit/submit2 methods.

>     * The highlight should probably be cleared when the search text changes (especially when the highlighted order does not match the new search input).
Fixed, now the highlight gets cleared whenever the search text changes.

>     * It might just be my color vision being flaky, but I completely did not notice the highlight arrow at first. Now that I know what to look for, it isn't hard to spot. But my first cycling through the matches was confusing because it wasn't obvious which order was being indicated.
I reshaped the arrow, moved it to the right side of icon and used more contrasting colors (black on white) to make it more visible.
obraz

---

@ChrisJohnsen Thanks for all the detailed feedback! I made each change in a separate commit to make the review easier. Let me know if there's anything else that needs adjusting.

plugins/lua/orders.lua
```diff
@@ -6,6 +6,12 @@ local overlay = require('plugins.overlay')
 local textures = require('gui.textures')
 local utils = require('utils')
 local widgets = require('gui.widgets')
+local stockflow = reqscript('internal/quickfort/stockflow')
+
+-- Shared state for search cursor visibility
+local search_cursor_visible = false
+local search_last_scroll_position = -1
+local order_count_at_highlight = 0
```
ChrisJohnsen: The first two of these are shadowed later by the locals declared at the top of the OrdersSearchOverlay section. Probably all these search/highlight locals should just be consolidated in that section.

plugins/lua/orders.lua
pajawojciech:
```
stockflow.collect_reactions() - 39 ms (runs once)
Building reaction map (3656 entries) - 141 ms (runs once)
get_order_search_key - under 1 ms (runs once per order, per search)
update_filter - 0-16 ms for 170 orders (runs once per search edit)
```

Based on these timings, the performance is good. We've got a ~180ms startup cost that only happens once. The actual searching is fast - under 16ms even with 170 orders. I can also cache the order keys, but I don't think it's worth it - search is already fast enough

plugins/lua/orders.lua
pajawojciech: The string format is just used to create a unique identifier for matching orders to their reaction names. It's never used for sorting anything - it's purely for hash map lookup.

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    if text == '' then
+        self.subviews.main_panel.frame_title = 'Search'
+        return
+    end
+
+    local orders = df.global.world.manager_orders.all
+    for i = 0, #orders - 1 do
+        local order = orders[i]
+        local search_key = get_order_search_key(order)
+        if search_key and utils.search_text(search_key, text) then
```
ChrisJohnsen: As it currently is, `get_order_search_key` has the following returns:
- some non-falsy value from the reaction map
- an empty string
(it can also throw an error if the reaction map couldn't be loaded)

All of the return values are currently "truthy", so they won't affect the if condition here.

Searching an empty `search_key` will only "succeed" when it is searched for an empty string, which won't happen here because of the guard before the loop. So it all _works_, but it indicates a misalignment with the definition of `get_order_search_key`.

---

Probably it makes the most sense to keep this check and change `get_order_search_key` to return nil if the reaction map isn't available (guarding against a potential "attempt to index a nil value" error there) or if the map didn't have an entry for the order.

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    self.LIST_START_Y_TWO_TABS_ROWS = 10
+    self.BOTTOM_MARGIN = 9
+    self.ARROW_X = 10
+    self.ARROW_FG = COLOR_BLACK
+    self.ARROW_BG = COLOR_WHITE
+
+    self.cached_list_start_y = nil
+    self.cached_viewport_size = nil
+    self.cached_screen_width = nil
+    self.cached_screen_height = nil
```
ChrisJohnsen: None of these appear to be used anywhere?

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    full_interface=true,
+}
+
+function OrderHighlightOverlay:init()
+    self.ORDER_HEIGHT = 3
+    self.TABS_WIDTH_THRESHOLD = 155
+    self.LIST_START_Y_ONE_TABS_ROW = 8
+    self.LIST_START_Y_TWO_TABS_ROWS = 10
+    self.BOTTOM_MARGIN = 9
+    self.ARROW_X = 10
```
ChrisJohnsen: These constants don't need to be fields of the overlay. Probably just a collection of module-locals at the top of the OrderHighlightOverlay section would be fine.

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    self.ARROW_FG = COLOR_BLACK
+    self.ARROW_BG = COLOR_WHITE
```
ChrisJohnsen: These values can probably just be inlined into the `dfhack.pen.parse` call. The table key names there will establish the fg/bg intentions.

plugins/lua/orders.lua
ChrisJohnsen: Hiding the highlight makes it less obvious, but the bug is still there. The same reproduction still applies, just add a Alt-n/Alt-p (cycle highlight) to re-reveal the misaligned highlight (be sure to cycle back to the matches that were after the deleted order).

The matched indices either need to be recalculated or cleared. Probably this is better done in the other overlay (to avoid needing another shared local to ask for search invalidation).

A related bug is where the "Search: N of M" text doesn't update when a matching order is deleted.

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+
+    local selected_y = list_start_y + (pos_in_viewport * self.ORDER_HEIGHT)
+
+    return selected_y
+end
+
+function OrderHighlightOverlay:render(dc)
+    if mi.job_details.open then return end
+
+    if search_cursor_visible then
```
ChrisJohnsen: The indentation isn't getting too bad here, but this might be better as a early-returning guard instead of having the rest of the body of the function tucked into the if block. It makes it clear that nothing else is going to happen for the guarded condition.

E.g.
```lua
if not search_cursor_visible then return end
```

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+    end
+
+    local pos_in_viewport = scroll_pos - viewport_start
+
+    local selected_y = list_start_y + (pos_in_viewport * self.ORDER_HEIGHT)
+
+    return selected_y
+end
+
+function OrderHighlightOverlay:render(dc)
```
ChrisJohnsen: It is a good idea bring back an unconditional call to the super render method here. It isn't technically necessary since there is nothing else that needs to be drawn in this overlay (e.g., a frame, sub-widgets, or onRenderFrame/onRenderBody methods), but having it can prevent hard-to-guess-at "why isn't this stuff isn't being drawn?" problems if any of those are added in the future.

plugins/lua/orders.lua
```diff
@@ -709,11 +716,397 @@ function QuantityRightClickOverlay:onInput(keys)
...
+        order.item_type,
+        order.item_subtype,
+        order.mat_type,
+        order.mat_index,
+        order.reaction_name or '',
+        mat_cat_str,
+        encrust_str)
+end
+
+local function build_reaction_map()
```
ChrisJohnsen: `build_reaction_map` could be inlined into `get_cached_reaction_map` (the only caller). I don't think there is any need to have the non-caching version available. An early return (if the map is already cached) can be used to limit the nesting in `get_cached_reaction_map`.

ChrisJohnsen (commented):
Good spotting of the need to hide when `job_details.open`. The other functionality changes all seem good (with a new note about order deletion handling).

I've put some more thoughts on the specifics of the code in this review.

ChrisJohnsen:
So, I hadn't looked into the details of `collect_reactions` from `internal/quickfort/stockflow` until now... I had assumed (yeah, my mistake) that its `order` fields were Lua tables that mimicked `df::manager_order` objects enough that (e.g.) `make_order_key` could use either a real order or a "mocked up" one.

But those `order` fields are actual `df::manager_order` objects (~3500 of them) that are allocated and filled out in (functions called by) `collect_reactions` . So if callers don't free them, they end up leaked. Since these objects aren't needed by this code after keys are derived from them, they should probably just be freed right away.

The other caller is `internal/quickfort/orders`; it accepts the leak and only calls `collect_reactions` once (storing it in a module global; disclaiming any need for world-varying or modded reactions).

plugins/lua/orders.lua
pajawojciech: Fixed

plugins/lua/orders.lua
pajawojciech: Good catch, fixed

plugins/lua/orders.lua
pajawojciech: Changed to locals at the top of overlay section

plugins/lua/orders.lua
pajawojciech: fixed

plugins/lua/orders.lua
pajawojciech: Fixed. Also added returns to cursor hiding

plugins/lua/orders.lua
pajawojciech: Added super render on top of method

plugins/lua/orders.lua
pajawojciech: Inlined

plugins/lua/orders.lua
```diff
@@ -709,11 +711,415 @@ function QuantityRightClickOverlay:onInput(keys)
...
+        order.item_type,
+        order.item_subtype,
+        order.mat_type,
+        order.mat_index,
+        order.reaction_name or '',
+        mat_cat_str,
+        encrust_str)
+end
+
+local reaction_map_cache = nil
```
ChrisJohnsen: When I tried loading different worlds I found that the cache wasn't being flushed like I expected. I mistakenly thought that a local cache variable would suffice. (I blame working too much on just overlays in script modules where a rescan will reload the module — when the script has been modified.)

The reaction cache needs to be flushed when a (new) world is loaded so that the world- or civilization-specific (e.g., `clothing_reactions`) orders can be cached.

For example, I have two forts: "A" knows how to make dresses but not shirts, "B" knows how to make shirts but not dresses. If I load "A" first, unload it, then load "B", searching "shirt" will not find a "make yarn shirt" order (though it does find still find a "forge iron mail shirt" order). This is because the "shirt" armor subtype was not cached since it was not available in fort "A".

Additionally, the cache should probably be a global variable. Leaving it as a local would force a rebuild after a module reload (even if the world hasn't changed).

---

You can watch for world loads with `dfhack.onStateChange`. The typical usage works like this:

At the top (after the `require` imports):

    local GLOBAL_KEY = 'orders'

The local can be replaced with a global initialization that copies any existing value instead of starting off as nil:

    reaction_map_cache = reaction_map_cache or nil

At the end (between `OVERLAY_WIDGETS` and `return _ENV`):

    dfhack.onStateChange[GLOBAL_KEY] = function(sc)
        if sc ~= SC_WORLD_LOADED then return end

        reaction_map_cache = nil
        get_cached_reaction_map()
    end

plugins/lua/orders.lua
```diff
@@ -709,11 +711,415 @@ function QuantityRightClickOverlay:onInput(keys)
...
+        map[key] = reaction.name:lower()
+        df.delete(reaction.order)
+    end
+    reactions = nil
+
+    reaction_map_cache = map
+    return reaction_map_cache
+end
+
+local function get_order_search_key(order)
```
ChrisJohnsen: If I take a basic "make wooden earring" order and use the DF UI to specify a wood type, the order becomes unsearchable. (using the shorter key strings described in another section of this review)
- make wooden earring: `235:-1:-1:-1:-1::2:`
- make willow earring: `235:-1:-1:420:203::2:`

Only the generic "wooden earring" order is generated by `collect_reactions`, so the key for the order with willow specified is not found in the cache.

Rock pots works similarly, but the generated order is only generic on the mat_index:
- make rock pot: `216:-1:12:0:-1::0:`
- make microcline pot: `216:-1:12:0:233::0:`

I experimented (in `get_order_search_key`) with checking for "genericized" orders (added material value "override" arguments to `make_order_key`) when the direct order is not found (first only with mat_index == -1, then with also with mat_type == -1 if there was still no match). When such a genericized match was found I extended the "search_key" with a material description fetched with

    dfhack.matinfo.decode(order.mat_type, order.mat_index):toString()

This made the order searchable by its generic wording and its material description. The "search_key" ends up being "make wooden earring willow" but searching for "willow ear" (or "ow ear" with full text searching enabled) works due to `utils.search_text` tokenizing the sought text.

I have never needed to directly use material values before, so I am not sure if this "genericizing" is correct for all materials/items/jobs. It seems okay from how `collect_reactions` works (`mat_type` and `mat_index` default to -1, and sometimes only `mat_type` is specified), but I don't know the extent of what DF allows. Hopefully someone with more experience there can weight in here.

plugins/lua/orders.lua
```diff
@@ -709,11 +711,415 @@ function QuantityRightClickOverlay:onInput(keys)
...
 
+--
+-- OrdersSearchOverlay
+--
+
+local search_cursor_visible = false
+local search_last_scroll_position = -1
+local order_count_at_highlight = 0
+
+local function make_order_key(order)
```
ChrisJohnsen: While I was diagnosing the inability to search for orders with specified materials I ended up looking at a bunch of these order keys strings.

The last two components of the key are basically just expanded bitfields that list the flag names (and values for the material categories). It seems like these could just be represented by their (masked) integer values instead of the verbose "flag name" version.

The (current) "flag name" version of "Make wooden earring" looks like this (182 bytes for the key):

      235:-1:-1:-1:-1::bone=false;cloth=false;horn=false;leather=false;pearl=false;plant=false;shell=false;silk=false;soap=false;strand=false;tooth=false;wood=true;wood2=false;yarn=false;:   = make wooden earring

Using the integers for the flags gives this (19 bytes for the key):

    235:-1:-1:-1:-1::2:     = make wooden earring

In one of my forts, the total length of the "flag name" keys for the 3533 generated orders was 653,367 bytes. With the mostly-integer keys it was only 66,632.

I extracted the integer values like this:

    local mat_cat = 0
    for flag_name, flag_value in pairs(order.material_category) do
        if type(flag_name) == 'string' and flag_value == true then
            mat_cat = mat_cat | 1 << df.job_material_category[flag_name]
        end
    end
    local encrust_str = ''
    if order.job_type == df.job_type.EncrustWithGems
        or order.job_type == df.job_type.EncrustWithGlass
        or order.job_type == df.job_type.EncrustWithStones
    then
        encrust_str = tostring(order.specflag.encrust_flags.whole
            & (1 << df.stockpile_group_set.finished_goods
                | 1 << df.stockpile_group_set.furniture
                | 1 << df.stockpile_group_set.ammo))
    end

Not only does this give shorter keys, but it is faster (250-400ms; "long keys" was 550-650ms).
It can be even faster by only computing the appropriate masks once (140-250ms).

There doesn't appear to be much support for "multi-field" manipulation of bitfields from the Lua side (a couple of functions in utils, but none that generates masks). It isn't too bad here because all the flags in question all single bits, but that isn't true for all bitfields.

The masking is required for the encrust flags (since we only want to pick a few of them), but it is probably a good idea for the material category flags (even though we want to know about all of them) to avoid picking up "stray bits" from the undefined portion of the bitfield (though this probably shouldn't happen...).

---

Whether or not the shorter keys are adopted, the falsity checks for `material_category`, `specflag` and `specflag.encrust_flags` are not needed since they are not behind pointers. If those fields are actually missing (renamed, moved, ...) even trying to check them for nil will raise an error.

Also, `specflag` is a union, so it isn't really appropriate to access `specflags.encrust_flags` without checking whether that member is the active variant. In the code above I checked the `job_type`, but I don't really know if that is the right "tag" to check before using the `encrust_flags` variant (it does seem to be the main thing `collect_reactions` sets when also setting `encrust_flags`).

ChrisJohnsen (commented):
I found couple of functionality things that I didn't notice before:
- cache invalidation between world loads,
- searchability after setting material for an order.

While looking at the details of customized orders I wondered if the overall size of the cache can be reduced.

Some functions can be made local (they don't access `self`, or won't once the methods the call are themselves made local):
- `perform_search`
- `getListStartY`
- `getViewportSize`
- `calculateSelectedOrderY`

When adding/removing an order there is still a small UI bug: the "Search: X of Y" title remains unchanged even if a matching order is added or removed. You could refresh the search when the number of orders changes, or just revert to the plain "Search" title so that the (possibly) stale count isn't displayed.

plugins/lua/orders.lua
pajawojciech: Consolidated

plugins/lua/orders.lua
pajawojciech: I want it to stay like this, header text will update on next search overlay user action

plugins/lua/orders.lua
pajawojciech: I changed my approach to name generation, so the comment is outdated. There is no cache now

plugins/lua/orders.lua
pajawojciech: I changed my approach to name generation, so the comment is outdated. There is no problems with materials, colors, instrumets, etc.

plugins/lua/orders.lua
pajawojciech: Thank you for this proposition. 
I changed my approach to name generation, so now there is no keys

library/modules/Job.cpp
```diff
@@ -686,3 +687,29 @@ std::string Job::getName(df::job *job)
...
+    button->specdata.hist_figure_id = order->specdata.hist_figure_id;
+    button->jobtype = order->job_type;
+    button->itemtype = order->item_type;
+    button->subtype = order->item_subtype;
+    button->material = order->mat_type;
+    button->matgloss = order->mat_index;
+    button->specflag = order->specflag;
+    button->job_item_flag = order->material_category;
+    button->specdata = order->specdata;
+    button->art_specifier = order->art_spec.type;
```
ChrisJohnsen: This isn't in `getName`. Was is needed for a particular kind of order?

If so, maybe `getName` should adopt it, too.

I didn't notice any orders that seemed to incorporate `job_art_specifier_type` details in the UI text, but I could easily have missed something.

library/modules/Job.cpp
```diff
@@ -686,3 +687,29 @@ std::string Job::getName(df::job *job)
...
 }
+
+std::string Job::getManagerOrderName(df::manager_order *order)
+{
+    CHECK_NULL_POINTER(order);
+
+    std::string desc;
+    auto button = df::allocate<df::interface_button_building_new_jobst>();
+    button->mstring = order->reaction_name;
+    button->specdata.hist_figure_id = order->specdata.hist_figure_id;
```
ChrisJohnsen: For confirmation by someone that works more on the C++ side:

This may not be necessary since `specdata` is copied as a whole later?

It seemed to work for me when I leaving this one off (copying one of them was definitely needed for some orders (e.g. differently-sized clothes)).

ChrisJohnsen (commented):
Here is a review to tick the requested-review box. My previous comment also mentions a couple of remaining UI bugs.

ChrisJohnsen:
Found a leftover while reviewing my local notes:

If `getManagerOrderName` ends up in the C++-exported Lua API (whether it ends up in Jobs or elsewhere), it should probably get an entry in `docs/dev/Lua API.rst` (similar to the one for `dfhack.job.getName`).

library/modules/Job.cpp
pajawojciech: It is not needed. Removed

library/modules/Job.cpp
pajawojciech: It is a struct and there is no need to copy hist_figure_id. Removed

pajawojciech:
> Here is a review to tick the requested-review box.

Sorry, but I don't understand what this is about.


> My previous comment also mentions a couple of remaining UI bugs.

UI bugs fixed and some more changes:
- render method now contains periodically check for orders changes
- search automatically refreshes when orders are added, removed, or modified
- all matched orders are highlighted, not just the selected one
- matching orders are highlighted in real-time as you type
- highlights remain visible while scrolling for selected and matching orders

I love how it feels to use now that the order highlights stay visible. It's much more user-friendly. Thanks @ChrisJohnsen for the review!

obraz

ChrisJohnsen:
> > Here is a review to tick the requested-review box.
> 
> Sorry, but I don't understand what this is about.

Yeah, that was just GitHub noise. My [first comment][a] ("Interesting. …") was just a "comment" (not a "review"), so it did not clear the "review requested" state. I posted the [second one][b] ("Here is a review…") as a actual "review" to have GitHub mark the PR as "reviewed".

[a]: https://github.com/DFHack/dfhack/pull/5677#issuecomment-3703699680
[b]: https://github.com/DFHack/dfhack/pull/5677#pullrequestreview-3621870537

ChrisJohnsen:
Since all matches are now being highlighted `search_cursor_visible` is now synonymous with `#search_matched_indices > 0`, so a separate variable isn't really necessary.

---

You might want to use time-based updating instead of frame-based so that the update time doesn't vary (as much) with configured (or effective) frame rate.

The overlay system provides periodic calls to the `overlay_onupdate` method when it is defined. The period can be controlled by the `overlay_onupdate_max_freq_seconds` attribute/field. The default period is 5 seconds, but since the game is paused while these overlays are active it should be okay to request more frequent updates. A 1 second period would approximate your 50 frame interval (for the default 50 FPS setting), but float values between 0 and 1 also work here and could make it more responsive.

Hmm, this is another case where a "post input" callback  would be handy. I've come across other uses for an overlay callback that occurs _after_ DF has handled an input. Pretty much all player-originated order changes will need to done through some kind of input, so this overlay could (if a "post input" callback was available) just check whether the orders have changed after each input instead of periodically... I've added this use case to my notes.

---

Since the inter-overlay coordination has been awkward a couple of times (and just changed to be bi-directional with `search_overlay_instance`), it might be worth considering consolidating the overlays. I have shied away from suggesting it since it involves a bit of non-standard `render` usage, but it might end up being less awkward than the variable-mediated inter-overlay coordination.

The variable-mediated coordination relies to the pseudo-singleton nature of overlays (as managed by the overlay system) which isn't documented at all as far as I have seen.

Drawing the highlights outside the main widget's frame is kind of non-standard behavior and relies on the painter being clipped to the full interface area.

The `view:render` documentation does state that the parent's painter is supplied to `render`, but it is an assumption that an overlay widget's parent's painter is clipped to the full interface area (though not much of an assumption, in my opinion, since overlay widgets can normally be placed anywhere in the interface area). `render` is already present to cut off drawing when `mi.job_details.open`, so it would be easy to call a "draw highlights" function from there.

Neither is a perfect, but combining the overlays would get rid of all the non-constant variables.

---

A minor thing about "checksum": To me, the "checksum" strongly implies a small (or constant-sized) value. The concatenation of order names is serving a similar purpose (change detection), but it is basically just a copy of the data it is checking. Something like `last_order_names` might be better?

I experimented with a "checksum"-free version that just periodically, unconditionally called a modified `update_filter` that checked whether the new `search_matched_indices` was different from the old value and only reset `search_current_match_idx` when there was a change. It worked okay, but it did pull the frame rate down a bit (this was with a 0.1 second period, so fairly aggressive). The idea was that `perform_search` needed to get all the order names anyway, maybe that work could also serve to detect changes. But it looks like all the `utils.search_text` usage is more expensive than just collecting the order names an extra time.

docs/dev/Lua API.rst
```diff
@@ -1430,6 +1430,10 @@ Job module
 
   Returns the job's description, as seen in the Units and Jobs screens.
 
+* ``dfhack.job.getManagerOrderName(manager_order)``
```
ab9rf, line 1433: This API change requires a changelog entry (in the API section).

pajawojciech: Added

docs/changelog.txt
```diff
@@ -115,6 +116,7 @@ Template for new versions:
 ## Documentation
 
 ## API
+- `dfhack.job.getManagerOrderName()`: New function to get the display name of a manager order
```
ab9rf: need doubled backticks here

pajawojciech: Done

PR #5678, opened 2025-12-03
Convert build-windows workflow to run on windows natively

NicksWorld:
This converts the windows ci build process to using a native windows runner. The runner appears to complete builds more reliably, as in testing I have yet to encounter any spurious failures or hangs whilst compiling.

The build times appear mostly the same to our previous runner, being 25minutes for a cleanbuild, and around 6minutes with a healthy cache.

Sample runs (from a messy testing branch):
https://github.com/NicksWorld/dfhack/actions/runs/19900868400
https://github.com/NicksWorld/dfhack/actions/runs/19902814344

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -115,11 +115,13 @@ jobs:
         run: |
           mkdir output
           mkdir pdb
+      - name: Get sccache path
+        run: echo ("SCCACHE_DIR=" + $env:LOCALAPPDATA + "\Mozilla\sccache\cache") >> $env:GITHUB_ENV
       - name: Fetch ccache
         if: inputs.platform-files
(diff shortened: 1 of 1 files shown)


.github/workflows/build-windows.yml
```diff
@@ -108,49 +111,80 @@ jobs:
...
+      - name: Prepare output directories
+        run: |
+          mkdir output
+          mkdir pdb
       - name: Fetch ccache
         if: inputs.platform-files
         uses: actions/cache/restore@v4
         with:
-          path: build/win64-cross/ccache
+          path: C:\\Users\\runneradmin\\AppData\\Local\\Mozilla\\sccache\\cache
```
ab9rf: this feels very magical to me

PR #5682, opened 2025-12-08
initial implementation of `Files` subsystem

ab9rf:
move methods related to file system  operations out of core

begin to implement mirrored implementation of DF's "portable mode" behavior with respect to dfhack configuration files


ab9rf:
Not happy with the direction this started off; will revisit with a new PR

PR #5684, opened 2025-12-11
change `Core`'s lifecycle

ab9rf:
instead of having `Core` be a static singleton with indeterminate construction and destruction sequencing, make Core's instance lifecycle explicit by tying it to `dfhooks_init` and` dfhooks_shutdown`


dhthwy:
Good improvement

ab9rf:
I hate git.

PR #5695, opened 2026-01-05
Fix the suspendmanager overlay appearing for followed units

### LINKED ISSUE #5439: `suspendmanager` UI overlaps workshops when following builder
When following a unit that's scheduled to construct a building using the vanilla DF follow feature, the `suspendmanager` UI for the unit's job will always render on top of other UIs (workshops, buildings, etc.).

In the following screenshots you'll see me following a dwarf of mine that was assigned to build a door, and the `suspendmanager` UI for that construction job overlaps other UIs that it shouldn't.

Image

Image

Image

Image

### LINKED ISSUE #5681: `SuspendManager`(?) info for queued workshop may show up in identically typed existing workshops
IF you have already built workshop(s) of a type, and queue another workshop of that type, then the dfhack suspendmanager(?) info may "leak" into the display of the already built ones.

Example - New Kitchen under construction:

Image

Examining a **different* already built kitchen, which just happens to have no running jobs (happens to every kitchen who is not currently running a job):

Image

Examining a different already built kitchen which *has* a job running:

Image

So it *appears* that for same type workshops where current jobs do not block the place the suspendmanager info would take, the (last?) info from a queued building is erroneously displayed.

plule:
The suspendmanager overlay was relying on `dfhack.gui.getSelectedJob()`, which looks for a job not only in buildings but also selected or followed units.

Relying on the selected building instead removes these edge cases.

getSelectedJob also prioritize the followed unit over the selected building, but reversing this is not sufficient for cases where the building does not have any job.

Fixes #5681 
Fixes #5439

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -71,6 +71,7 @@ jobs:
       - name: Install build dependencies
         run: |
           choco install sccache
+          pip install Jinja2
       - name: Install doc dependencies
         if: inputs.docs
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       args: ['--fix=lf']
     - id: trailing-whitespace
 - repo: https://github.com/python-jsonschema/check-jsonschema
-  rev: 0.35.0
+  rev: 0.36.0
   hooks:
     - id: check-github-workflows
 - repo: https://github.com/Lucas-C/pre-commit-hooks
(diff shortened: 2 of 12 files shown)


ab9rf:
I get the feeling that we need to rethink some aspect of the API here, but I am not going to delay implementing this while we do that.

plule:
> I get the feeling that we need to rethink some aspect of the API here, but I am not going to delay implementing this while we do that.

@ab9rf yes, in a way "getSelectedJob()" being too smart makes it hard to rely on. Maybe more spécialized versions of it would be better ("getSelectedBuildingJob()"/"getSelectedUnitJob()"). I'll give it a shot someday

Issue #5034, opened 2024-11-18
add filter for miners/woodcutters/hunters to squad selection screen


myk002:

myk002:
I had originally *not* included this since they already have an icon that clearly identifies them as miners/woodcutters/hunters. Maybe this is not necessary and instead we could add a notification for this situation?

see: https://github.com/DFHack/dfhack/issues/5028

Bumber64:
This can be closed due to #5700.

ab9rf:
closed in #5700

PR #5700, opened 2026-01-07
 Add Uniformed filter to squad selection screen

pajawojciech:
New "Uniformed:" filter in squad selection screen:
  - Filters dwarves with Mining / Woodcutting / Hunting labors
  - Hotkey: Shift+U
  
Issue: #5034 

obraz

### DIFF
--- a/docs/about/Authors.rst
+++ b/docs/about/Authors.rst
@@ -163,6 +163,7 @@ Omniclasm
 Ong Ying Gao            ong-yinggao98
 oorzkws                 oorzkws
 OwnageIsMagic           OwnageIsMagic
+pajawojciech            pajawojciech
 palenerd                dlmarquis
 PassionateAngler        PassionateAngler
 Patrik Lundell          PatrikLundell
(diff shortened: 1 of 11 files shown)


plugins/lua/sort.lua
```diff
@@ -1029,12 +1029,29 @@ function SquadFilterOverlay:init()
 
     local left_panel = widgets.Panel{
         view_id='left_panel',
-        frame={t=1, b=0, l=0, w=NARROW_WIDTH-4},
+        frame={t=0, b=0, l=0, w=NARROW_WIDTH-4},
         visible=true,
         subviews={
+            widgets.HotkeyLabel{
```
ab9rf, line 1035: Why was this moved from the right panel to the left panel?

pajawojciech: @ab9rf 
Currently, 'toggle all' is in the first row on the right side, with the left side empty. Since I added a new element, I moved 'toggle all' to the left and placed the new button on the right to avoid leaving empty space while keeping 'toggle all' in the first position.
Alternative option: I can keep 'toggle all' where it was and add the new button in a new row at bottom.

ab9rf: I was just trying to understand why it was changed. It's fine as is, I think.

Issue #3694, opened 2023-08-23
announcement monitor


ab9rf:
offer an option to filter sparring events because otherwise they overwhelm the monitor

myk002:
can we do that -- just eject the sparring messages from the queue before they push other messages out?

Issue #3397, opened 2023-05-19
Feature Request: Fort Combat Tracker


Dragoon209:
Something I would love to see is a 'combat tracker' that provides a better snapshot of what combat looks like.  This would be a pop-up or hovering window (hotkey summonable overlay?) that would show the combat reports for all combatants simultaneously, but not block the ability to unpause, or advance step of combat.

While I think this will help any player get a better idea of the flow of combat and who is involved, I think this is would also be a super helpful tool for a streamer.  Navigating through multiple combat reports and trying to explain what is going on is difficult and confusing.

The specific features that I think are important:

- Ability to enable/disable sparring reports
- Ability to filter what dwarves it shows the combat reports for (follow a specific unit like a hero, or just show combat for a squad vs. a forgotten beast for example)
- Ability to toggle the FPS down to 1 (or user set value) to allow a live combat feed, while still being readable in realtime.
- Wishlist: Ability to add / remove all the other alert types as well
- Wishlist: Ability to mark entries or range, and export them to a text file.  Bonus points for maintaining the color formatting, or other alert type indicator.
- Wishlist: Ability to stream the reports to an external file or other mechanism to allow the combat reports to be integrated to streaming platforms/twitch bots/discord chat

Thanks for the time and awesome work!

PR #5704, opened 2026-01-10
Create plugin for clearing of combat, sparring, and hunting reports with configurable filtering and overlay UI

pajawojciech:
## Summary
Adds a new plugin to automatically clear combat/sparring/hunting reports from buffer. Features configurable filtering by type, overlay UI settings, and control panel integration.

This does **NOT** hide the left-side notifications on screen:
obraz

## Features
- **Time-triggered clearing**: Runs every 100 ticks
- **Configurable filtering**: Separate toggles for combat, sparring, and hunting logs
- **Sparring enabled by default**: Targets the most common spam source
- **Command interface**: Configure via console commands; auto-enables when filters specified
- **Overlay UI**: Settings panel with enable toggle and filter toggles 
obraz

- **Control panel integration**: Plugin in Gameplay tab with autostart options
obraz

## Usage
```
logcleaner              - show status
logcleaner enable       - enable plugin
logcleaner disable      - disable plugin
logcleaner sparring     - enable sparring filter (auto-enables plugin)
logcleaner all          - enable all filters (auto-enables plugin)
logcleaner combat,hunting - enable specific filters
gui/logcleaner          - open settings overlay
```

## Tests done
- 50 dwarves training generate ~100-200 reports per 100 ticks
- Cleaning time < 2 ms (minimal performance impact)
- Plugin state persists across save/load                                                        
- Autostart works for new embark
- Old fort ran over 1 year, and all wanted reports preserved (my calculations show in this case ~3 years before buffer overflow)

### Three tabs are empty
obraz

### Now you can finally keep your reports clean enough to actually spot when Urist McSad is depressed or when a tantrum spiral starts
obraz

## Related issues
https://github.com/DFHack/dfhack/issues/3694
https://github.com/DFHack/dfhack/issues/3397

## Related PR in scripts repo (gui and autostart)
https://github.com/DFHack/scripts/pull/1536

### DIFF
--- a/plugins/logcleaner/logcleaner.cpp
+++ b/plugins/logcleaner/logcleaner.cpp
@@ -31,7 +31,7 @@ static bool clear_combat = false;
 static bool clear_sparring = true;
 static bool clear_hunting = false;
 
-static const int32_t CLEANUP_TICK_INTERVAL = 97;
+static constexpr int32_t CLEANUP_TICK_INTERVAL = 97;
 
 static void cleanupLogs();
 static command_result do_command(color_ostream& out, std::vector<std::string>& params);
@@ -130,6 +130,9 @@ static void cleanupLogs() {
     if (!is_enabled || !world)
         return;
 
+    if (!clear_combat && !clear_sparring && !clear_hunting)
+        return;
+
     // Collect all report IDs from unit combat/sparring/hunting logs
     std::unordered_set<int32_t> report_ids_to_remove;
     bool log_types[] = {clear_combat, clear_sparring, clear_hunting};
@@ -152,17 +155,12 @@ static void cleanupLogs() {
     // Remove collected reports from global buffers
     auto& reports = world->status.reports;
 
-    for (auto report_id : report_ids_to_remove) {
-        df::report* report = df::report::find(report_id);
-        if (!report)
-            continue;
-
-        auto it = std::find(reports.begin(), reports.end(), report);
-        if (it != reports.end()) {
-            delete report;
-            reports.erase(it);
-        }
-    }
+    std::erase_if(reports, [&](df::report* report) {
+        if (!report || !report_ids_to_remove.contains(report->id))
+            return false;
+        delete report;
+        return true;
+    });
 }
 
 DFhackCExport command_result plugin_onupdate(color_ostream& out, state_change_event event) {


pajawojciech:
> The frequency should be a prime number

Fixed

---

> You need something like this during report deletion:

world->status.announcements does not contain sparring, combat and hunting reports

--- 

> Myk has a similar unfinished plugin. He ran into some kind of issue, but I don't remember specifics. Maybe it was the performance impact of unit.reports.log, which is less an issue at the frequency you're running?

I set frequency to 1 (every tick) and still can't see any performance impact. Performance check says that one run takes less than 2 ms.
Myk proposition uses buckets and reinserts reports. I think my version is simpler: foreach unit, get sparring reports (or combat and hunting reports if enabled) and remove them from units and from world.status.reports

---

> You could probably delete the unit report logs as you're iterating them

I accomplish this using log.clear();

ab9rf:
> The frequency should be a prime number (e.g., 97) for `timestream` reasons. This should be defined as a variable near the top, not a magic number in an if statement.

While I agree that the magic number should be defined as a symbolic constant, using a prime number doesn't really have anything directly to do with timestream. The idea behind using prime numbers is to avoid phasing with other tools that also run at periodic tick intervals so that two "expensive" tools don't both always, or frequently, run on the same tick, which might lead to visible stuttering for the user.

We've talked about providing a system where tools that need to run periodically request an "approximate frequency" and Core then alllocates them in a uniform manner across all tools so that everyone gets about the frequency they want while avoiding phasing, but this hasn't been implemented yet.

As to 2 ms per tick: this may not seem by much, but please remember that at 100 fps the _total_ per tick budget is only 10 milliseconds and the target time budget for _all of DFHack_ is 5% of that, or 0.5 ms. If you're using 2 ms per tick, that's 4x over budget and thus not remotely acceptable.

Finally, running more often than once per ten ticks is almost certainly overkill. A lot of things in DF happen once every ten, once every 50, or once every 100 ticks. Given this, 97 is probably an especially good number to use, as it falls just below the "once every 100" frequency.

plugins/logcleaner/logcleaner.cpp
```diff
@@ -0,0 +1,191 @@
...
+                }
+                log.clear();
+            }
+        }
+    }
+
+    if (report_ids_to_remove.empty())
+        return;
+
+    // Remove collected reports from global buffers
```
ab9rf, line 155: This is inefficient, requiring repeated scans of the `reports` vector for each removal and may have to move elements repeatedly. Consider using `std::erase_if` instead, which will only make _one_ scan of the vector, and intelligently collapses the gaps.

plugins/logcleaner/logcleaner.cpp
```diff
@@ -0,0 +1,191 @@
...
+    CONFIG_CLEAR_SPARING = 2,
+    CONFIG_CLEAR_HUNTING = 3,
+};
+
+static bool clear_combat = false;
+static bool clear_sparring = true;
+static bool clear_hunting = false;
+
+static const int32_t CLEANUP_TICK_INTERVAL = 97;
+
```
ab9rf, line 35: these should all be `constexpr`

plugins/logcleaner/logcleaner.cpp
pajawojciech: Changed to std::erase_if. In my case, there was a performance improvement from 440 to 340 microseconds.

plugins/logcleaner/logcleaner.cpp
pajawojciech: Fixed

Bumber64:
> world->status.announcements does not contain sparring, combat and hunting reports

I missed this before the PR merged. The above assertion is not correct. It's user configurable, and is the default for citizens being struck down.

Proposed fix #5726

Issue #3938, opened 2023-11-01
Add a button to the details window of buildings linked to levers


chdoc:
Adding a button to the details window of buildings linked to levers (e.g. a bridge) that would queue the "pull lever" job on the linked lever would be convenient.

myk002:
Note to imlementor (self?) -- code for creating a pull lever job is already in `lever`. we'd just need a simple UI that links up the functionality

PR #5725, opened 2026-02-26
Add overlay to queue lever pull jobs from linked building view

pajawojciech:
## How it works
- New overlay display button next to lever in 'Show linked buildings' tab. 
- Button show lever status indicator: slash or backslash, representing current position of lever. 
- Button adds / removes pull lever jobs 
- Works fine even with more than one linked levers. 
obraz

### Click button
After click text changes to "Pulling" with yellow color, and job is queued. Now you can click again if you change your mind - it removes queued job.
obraz

obraz

### Job assigned 
When job is assigned text changes color to green:
obraz
obraz

### Pulled

After pull text, color and indicator changes
obraz

### When Unlink is enabled button moves to the left
obraz

### What I tried
Instead of pull text i tried to set Open / opening / close / closing labels. For standard bridges it works fine, but for example retracting bridges or cages should have reversed operation naming, so I use 'Pull' text and lever position indicator

### Issue
https://github.com/DFHack/dfhack/issues/3938

### DIFF
--- a/plugins/lua/buildingplan/unlink_mechanisms.lua
+++ b/plugins/lua/buildingplan/unlink_mechanisms.lua
@@ -486,7 +486,7 @@ function MechLeverPullOverlay:get_button(button_idx, label, text_pen)
             widgets.TextButton
             {
                 view_id = "pull_"..button_idx,
-                frame = {t=0, r=17, w=9, h=1},
+                frame = {t=0, r=17, w=11, h=1},
                 label = label,
                 text_pen = text_pen or COLOR_WHITE,
                 on_activate = function() self:activate_button(button_idx) end,
@@ -551,13 +551,14 @@ function MechLeverPullOverlay:update_buttons()
             if show_pull then
                 local job = self:get_lever_pull_job(lever)
                 local label, text_pen
+                local state_char = lever.state == 0 and "/" or "\\"
 
                 if job then
-                    label = lever.state == 0 and "Closing" or "Opening"
+                    label = "Pulling " .. state_char
                     local worker = dfhack.job.getWorker(job)
                     text_pen = worker and COLOR_GREEN or COLOR_YELLOW
                 else
-                    label = lever.state == 0 and "Close" or "Open"
+                    label = "Pull    " .. state_char
                     text_pen = COLOR_WHITE
                 end
 
--- a/plugins/lua/buildingplan/unlink_mechanisms.lua
+++ b/plugins/lua/buildingplan/unlink_mechanisms.lua
@@ -414,22 +414,22 @@ function MechLeverPullOverlay:init()
     self.links = {}
 end
 
-function MechLeverPullOverlay:is_lever(b)
-    return b._type == df.building_trapst and b.trap_type == df.trap_type.Lever
+function MechLeverPullOverlay:is_lever(building)
+    return building._type == df.building_trapst and building.trap_type == df.trap_type.Lever
 end
 
-function MechLeverPullOverlay:has_lever_pull_job(lever)
-    for _, j in ipairs(lever.jobs) do
-        if j.job_type == df.job_type.PullLever then
-            return true
+function MechLeverPullOverlay:get_lever_pull_job(lever)
+    for _, job in ipairs(lever.jobs) do
+        if job.job_type == df.job_type.PullLever then
+            return job
         end
     end
-    return false
+    return nil
 end
 
 function MechLeverPullOverlay:queue_lever_pull_job(lever)
-    local ref = df.general_ref_building_holderst:new()
-    ref.building_id = lever.id
+    local building_ref = df.general_ref_building_holderst:new()
+    building_ref.building_id = lever.id
 
     local job = df.job:new()
     job.job_type = df.job_type.PullLever
@@ -439,7 +439,7 @@ function MechLeverPullOverlay:queue_lever_pull_job(lever)
         z = lever.z
     }
     job.flags.do_now = false
-    job.general_refs:insert("#", ref)
+    job.general_refs:insert("#", building_ref)
     lever.jobs:insert("#", job)
 
     dfhack.job.linkIntoWorld(job, true)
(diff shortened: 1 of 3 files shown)


chdoc:
I like the idea. However, I don't see any reason why this should be extending the `buildingplan` tool. In fact, I don't see why this couldn't be in `scripts`, for instance as part of the `lever` tool.

pajawojciech:
> I like the idea. However, I don't see any reason why this should be extending the `buildingplan` tool. In fact, I don't see why this couldn't be in `scripts`, for instance as part of the `lever` tool.

Thanks for review. 
I made it this way because overlay reuses code from unlink_mechanisms.lua file (for example mech_iter, get_trigger_index and more). Also one new and two old overlays share the same screen space (and pull overlay checks whether unlink is enabled for good positioning), so splitting them would create duplicated code and fragille coordination. 
Can scripts register overlay widgets?

chdoc:
I still maintain that an overlay that exclusively deals with pulling levers does not belong in a file that is called `unlink_mechanisms`. So, at the very least, the existing module should be modified to export the functionality that needs to be reused and the new overlay should be put into a separate module. Just because things are not in the same file doesn't mean you can't reuse functionality. And once the relevant functionality is exported, there really isn't a whole lot of a reason anymore for this to be part of `buildingplan`. There are plenty of scripts that define overlays, and quite a few of them need to be positioned not to conflict with other overlays: https://github.com/DFHack/scripts/blob/af457f9b5c86fc041a064ef5bdfbcab38f38a50f/idle-crafting.lua#L611-L613

The only slightly annoying part of moving this to `scripts` is that merging the pull request to scripts will be contingent on merging the pull request for exporting the functionality. But that can be dealt with.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+end
+
+function MechLeverPullOverlay:get_lever_pull_job(lever)
+    for _, job in ipairs(lever.jobs) do
+        if job.job_type == df.job_type.PullLever then
+            return job
+        end
+    end
+    return nil
+end
```
chdoc, line 428: These neither read nor write member variables, so there is no need for them to be member functions.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+
+function MechLeverPullOverlay:queue_lever_pull_job(lever)
+    local building_ref = df.general_ref_building_holderst:new()
+    building_ref.building_id = lever.id
+
+    local job = df.job:new()
```
chdoc, line 434: Please use `dfhack.job.createLinked()` and remove the call to `linkIntoWorld` (or better reuse the existing functionality from the `lever` script),

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+    job.job_type = df.job_type.PullLever
+    job.pos = {
+        x = lever.centerx,
+        y = lever.centery,
+        z = lever.z
+    }
+    job.flags.do_now = false
```
chdoc, line 441:
```suggestion
    job.flags.do_now = true
```
I think one can argue that the default for pulling levers should be with urgency.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+
+    dfhack.job.linkIntoWorld(job, true)
+    dfhack.job.checkBuildingsNow()
+end
+
+function MechLeverPullOverlay:remove_lever_pull_job(lever)
+    for i = #lever.jobs, 1, -1 do
+        local job = lever.jobs[i-1]
+        if job.job_type == df.job_type.PullLever then
+            lever.jobs:erase(i-1)
```
chdoc, line 453: Is it really necessary to call `erase`? `removeJob` should already remove all references to the job,

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+
+    local height = parent_rect.height - 49
+    self.frame.h = height + 1 --includes lower border
+    self.num_buttons = height // 3
+end
+
+function MechLeverPullOverlay:onRenderFrame(dc, rect)
+    if self.bld_id ~= sheet.viewing_bldid then
+        self.bld_id = sheet.viewing_bldid
+        self.building = df.building.find(self.bld_id)
```
chdoc, line 594: Since buildings can be destroyed at any time, storing buildings is generally not safe. This may be fine, because none of this is called unless there is a viewscreen open and therefore the game is paused, but please double check.

plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -391,6 +391,214 @@ function MechLinkOverlay:onRenderFrame(dc, rect)
...
+            else
+                self:queue_lever_pull_job(lever)
+            end
+        else
+            dfhack.printerr("MechLeverPullOverlay: Mechanism is not linked to a lever")
+        end
+    end
+end
+
+function MechLeverPullOverlay:update_buttons()
```
chdoc, line 527: I find it really hard to understand what is going on in these methods. I would really appreciate 1-3 lines of comments about the general idea here.

chdoc (changes requested):
Finally got around to providing an in-depth review.

pajawojciech:
Closed because there is similiar PR: https://github.com/DFHack/dfhack/pull/5824

PR #5726, opened 2026-02-28
Logcleaner: Properly delete announcements

Bumber64:
Remove deleted reports from the announcement vector to prevent dangling pointers.

I haven't tested this code. It's an emergency fix before the update goes out.


Bumber64:
Seems to be removing announcements properly.

I noticed that the plugin removes _every single unit report_ the instant it's enabled, even while paused. I think the "ticks" its counting are actually frames. You'll never read a single sparring report as long as `logcleaner` is enabled.

ab9rf (approved):

PR #5727, opened 2026-03-02
Fix `logcleaner` to use ticks

Bumber64:
Fixed `logcleaner` to use ticks instead of frames, using `autolabor` as reference. Also corrected docs that it runs "approximately" every 100 ticks.

### DIFF
--- a/plugins/autolabor/autolabor.cpp
+++ b/plugins/autolabor/autolabor.cpp
@@ -740,7 +740,7 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
         return CR_OK;
     }
 
-    if (world->frame_counter - cycle_timestamp <= CYCLE_TICKS)
+    if (world->frame_counter - cycle_timestamp < CYCLE_TICKS)
         return CR_OK;
 
     cycle_timestamp = world->frame_counter;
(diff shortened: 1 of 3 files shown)


Bumber64:
Looks like it works.

I copied the `<=` cycle condition from `autolabor` which causes it to run every 98 ticks. Feels like that should be changed? Someone probably forgot to remove the `=` after inverting it.

plugins/autolabor/autolabor.cpp
```diff
@@ -740,7 +740,7 @@ DFhackCExport command_result plugin_onupdate ( color_ostream &out )
         return CR_OK;
     }
 
-    if (world->frame_counter - cycle_timestamp <= CYCLE_TICKS)
+    if (world->frame_counter - cycle_timestamp < CYCLE_TICKS)
```
ab9rf: while there might be merit to this change, it's unnecessary to dealing with a defect in a pending release. feel free to repropose in a separate PR

even when not in a pre-release feature freeze it's best to keep PRs to a single issue, and a PR that is intended to fix a defect in a imminently pending release should _absolutely_ only address the specific defect and nothing else

ab9rf (changes requested):

ab9rf (approved):

Issue #5733, opened 2026-03-04
Some filters in squad assignement screen behave strangely


Halavus:
Other Squad: works fine
Officials: works fine
Nobility: works fine

Uniformed: doesn't work (see screenshot, miners and woodcutters not excluded, clicking on the filter-> nothing happens)
With infants: works fine
Hates combat: modifies the list, but I can't figure out what it does. Doesn't work in any cases
Maimed: doesn't work (see screenshot, the dwarf named "crutched Miner" has "ability to stand lost", clicking on the filter-> nothing happens)

<img width="1372" height="477" alt="Image" src="https://github.com/user-attachments/assets/f01c298d-b9d2-4e66-97dd-ee76f096333f" />

Halavus:
Addendum: it seems, that used separately, the filters kind of work (not entirely sure about maimed), but break when combined.

Halavus:
Console threw this error as well, which I did not see:

```
Failed Lua call to 'plugins.sort.do_squad_filter'
...teamapps\common\Dwarf Fortress\hack\lua\plugins\sort.lua:296: Cannot read field uint16_t[].PERSEVERENCE: not found.
stack traceback:
        [C]: in metamethod '__index'
        ...teamapps\common\Dwarf Fortress\hack\lua\plugins\sort.lua:296: in upvalue 'get_mental_stability'
        ...teamapps\common\Dwarf Fortress\hack\lua\plugins\sort.lua:1260: in upvalue 'is_unstable'
```

ab9rf:
That error message explains (nearly) everything: it's the "hates combat" filter

Halavus:
Indeed, everything works, when hates combat not set.

PR #5734, opened 2026-03-04
Perseverancing

ab9rf:
missed while closing out https://github.com/DFHack/df-structures/pull/872

closes #5733

PR #5735, opened 2026-03-04
Fix spelling of 'perseverance' in sort.lua

Halavus:
changed e to a

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -33,6 +33,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


ab9rf:
Sniped by #5734

Halavus:
I'll never be a DFHack contributor then... ^^
Maybe... someday...

ab9rf:
Closes #5733

This change was necessitated by DFHack/df-structures#572 and was noted there as required, but was missed and I missed that it was missed while finalizing those changes

ab9rf (approved):

PR #5739, opened 2026-03-06
fix windows console (#5737)

### LINKED ISSUE #1474: dfhack console on windows does not handle cp 65001 correctly

### LINKED ISSUE #5737: console (windows) not printing strings with extended ascii
This started in one of the recent (53.10-r2 or 53.11-r1) versions.
in the console, strings that contain extended ascii either don't print at all, or are truncated before the extended ascii character (including removing the newline)
examples:
with lua interpretter, and a unit with a name like Rigòth selected.
~unit.name.first_name -- nothing prints
~unit.name  -- truncated before the ò, no newline

this effects functions like dfhack.units.getReadableName()

It does not effect output to gui windows, or to the launcher. However, the same output that displays correctly in the launcher does not print in the console (tested with bprobe and a building with a name with a diacritical in it)

Not including a save because this bug can be reproduced without even loading a world or mods. from lua interpretter when first loading the game: 
print(string.char(128))  
in the current version, no output. with dfhack 53.10-r1 it prints Ç (and a newline)

ab9rf:
always use utf-8 when writing to the Windows console

fixes #5737
fixes #1474

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -61,6 +61,7 @@ Template for new versions:
 ## Fixes
 
 ## Misc Improvements
+- General: DFHack will unconditionally use UTF-8 for the console on Windows, now that DF forces the process effective system code page to 65001 during startup
 
 ## Documentation
 
--- a/library/MiscUtils.cpp
+++ b/library/MiscUtils.cpp
@@ -653,20 +653,31 @@ std::string UTF2DF(const std::string &in)
         out.resize(pos);
     return out;
 }
-
-DFHACK_EXPORT std::string DF2CONSOLE(const std::string &in)
+static bool console_is_utf8()
 {
-    bool is_utf = false;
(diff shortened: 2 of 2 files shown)


SilasD:
no action needed.

just commenting to add search keywords, because I searched on most of these before finally tracking down this PR in the big list.

codepage
utf8
unicode
terminal

PR #5748, opened 2026-03-13
Issue #5736 - Fix buildingplan planneroverlay

### LINKED ISSUE #5736: Bug: `buildingplanner` prevents toggling pressure plate triggers
I can't seem to enable plates to trigger on minecarts with the `buildingplanner` overlay active. I have to minimize planner back to the default DF building mode, click the minecart trigger checkbox, then I can switch back to the planner and mess with the weight ranges. (Planner remembers which triggers you had active when you're placing the next plate. That feature might have something to do with it, or it might be related to intercepting input.)

I'm too unfamiliar with the plugin to track down the issue, but I did notice that `get_quantity()` in `planneroverlay.lua` can return greater than 1 if multiple plate triggers are active, which is probably incorrect.

Halavus:
Resolves [Issue#5736](https://github.com/DFHack/dfhack/issues/5736)


Bumber64:
Seems to fix the issue for me.

ab9rf (approved):

PR #5749, opened 2026-03-13
Fix sliders allowing values outside their number of stops

Halavus:
Clicking left and right only recorded -1 or +1 based on mouse position without any safeguard against potential larger frame than the slider itself. EDIT: Dragging was affected too.

Capture d&#39;écran 2026-03-13 233008


library/lua/gui/widgets/slider.lua
```diff
@@ -19,7 +19,7 @@ local core = require('gui.widgets.slide_core')
 ---@field super widgets.Widget
 ---@field ATTRS widgets.Slider.attrs|fun(attributes: widgets.Slider.attrs.partial)
 ---@overload fun(init_table: widgets.Slider.initTable): self
-Slider = defclass(Slider, slide_core)
+Slider = defclass(Slider, core)
```
ab9rf, line 22: why is `slide_core` changed to `core` here?

Halavus: Because otherwise `local core` in line 1 is pointless.

ab9rf (approved):

library/lua/gui/widgets/slider.lua
SilasD: This is the wrong way to do it.  revert the change here, and change line 1 to
`local slide_core = require('gui.widgets.slide_core')`
or things just get too confusing.

the class's *name* is `slide_core`, it should be named exactly that by all code that uses it.  this is realSquidCoder's mistake, not yours, but this is a good time to fix it.

this is especially important because every script has a global variable named `slide_core` that is a reference to the `slide_core` class, so the existing code does work as written, despite that unused local variable `core`.

per our conventions, `slide_core` should *also* be named using CamelCase.  `_SliderCore` maybe?  i.e. starting the class name with a single underscore to hopefully flag that it's not intended to be instantiated.  that's out of scope of the intended fix, but it's a good time to make the change.

ab9rf: I'm holding this PR until the aforementioned issues are addressed

ab9rf:
I've revised this PR based on the discussion above and also applied the same bug fix to `RangeSlider`

PR #5754, opened 2026-03-17
fix more instances of `hack` hardcoding

ab9rf:
also in some places switch to using pathnames instead of strings for paths


library/modules/DFSteam.cpp
```diff
@@ -208,9 +208,10 @@ static bool launchDFHack(color_ostream& out) {
         return false;
     } else if (pid == 0) {
         // child process
-        static const char * command = "hack/launchdf";
+        auto procpath = Core::getInstance().getHackPath() / "launchdf.exe";
```
quietust: This is the Linux code path, so there shouldn't be a ".exe" here

PR #5781, opened 2026-03-31
Refactor button widgets; add `widgets.RadioButton`

Bumber64:
Refactor `widgets.ConfigureButton` and make `widgets.HelpButton` a subclass of it. Add `widgets.RadioButton`, which looks and functions like the buttons in the `gui/control-panel` list.

I don't understand the `---@field` stuff very well, so I based it on examples as best I could. Needs to be verified.


library/lua/gui/widgets/buttons/configure_button.lua
```diff
@@ -6,17 +8,20 @@ local to_pen = dfhack.pen.parse
...
 local button_pen_left = to_pen{fg=COLOR_CYAN,
     tile=curry(textures.tp_control_panel, 7) or nil, ch=string.byte('[')}
+local button_pen_center = to_pen{
+    tile=curry(textures.tp_control_panel, 10) or nil, ch=15} -- gear/masterwork symbol
 local button_pen_right = to_pen{fg=COLOR_CYAN,
     tile=curry(textures.tp_control_panel, 8) or nil, ch=string.byte(']')}
-local configure_pen_center = to_pen{
-    tile=curry(textures.tp_control_panel, 10) or nil, ch=15} -- gear/masterwork symbol
 
 ---------------------
```
SilasD, line 16: this is not Bumber's work here, but I think this code is more than a bit strange.

first, `curry()` cannot return falsey; it either throws an error or returns a function.  accordingly, the `or nil` clauses don't ever execute.

second, that `curry()` is used *at all*.  if I understand this construction, `tp.control_panel()` will be called to get the relevant texpos each time the pen is drawn instead of adding the texture *once* at module load time with e.g. `tile=tp.control_panel(10)`.  I do not know very much about textures, but I would be extremely surprised if the texpos for a control panel texture changes dynamically.

~~and third, that three individual pens are used instead of a three-tile table.~~  I now see that a three-tile Label is created in the postinit stage.

myk checked this in https://github.com/DFHack/dfhack/commit/a651f6461b28941a25b2135ecc11b5cd8af6b207 and he certainly knows more about textures than I do.

anyway, not directly relevant to the PR.

library/lua/gui/widgets/buttons/help_button.lua
```diff
@@ -1,53 +1,35 @@
...
 ---@class widgets.HelpButton.attrs.partial: widgets.HelpButton.attrs
 
----@class widgets.HelpButton: widgets.Panel, widgets.HelpButton.attrs
----@field super widgets.Panel
+---@class widgets.HelpButton: widgets.ConfigureButton, widgets.HelpButton.attrs
+---@field super widgets.ConfigureButton
 ---@field ATTRS widgets.HelpButton.attrs|fun(attributes: widgets.HelpButton.attrs.partial)
 ---@overload fun(init_table: widgets.HelpButton.attrs.partial): self
-HelpButton = defclass(HelpButton, Panel)
+HelpButton = defclass(HelpButton, ConfigureButton)
```
SilasD, line 22: subclassing `ConfigureButton` is reasonable I suppose.  better than copy-paste for sure.

there *ought* to be an abstract class used as the base of `ConfigureButton`, `HelpButton`, and `RadioButton`, but the class system doesn't really support abstract classes meant to only be subclassed.

library/lua/gui/widgets/buttons/configure_button.lua
```diff
@@ -27,27 +32,40 @@ local configure_pen_center = to_pen{
 ConfigureButton = defclass(ConfigureButton, Panel)
 
 ConfigureButton.ATTRS{
+    frame={t=0, l=0, w=3, h=1},
     on_click=DEFAULT_NIL,
+    pen_left=button_pen_left,
+    pen_center=button_pen_center,
+    pen_right=button_pen_right,
```
SilasD, line 39: I like moving the pens into `self` via `ATTRS`.  that's where it belongs.  that `preinit(init_table)` was weird and unneccessary.

I'm not so sure about allowing instances, even subclassed instances, to override `frame` size.

it's fine.

library/lua/gui/widgets/buttons/help_button.lua
```diff
@@ -1,53 +1,35 @@
...
 
 HelpButton.ATTRS{
     frame={t=0, r=1, w=3, h=1},
     command=DEFAULT_NIL,
+    pen_center=help_pen_center,
```
SilasD, line 27: the `frame` and `command` are redundant, aren't they?  they'll be picked up from `ConfigureButton`, the same way that `pen_left` and `pen_right` will be.

overriding `pen_center` shows the benefit of moving the pens into ATTRS, good.

not worth holding up the merge over a quibble.

library/lua/gui/widgets/buttons/help_button.lua
```diff
@@ -1,53 +1,35 @@
...
-            frame={t=0, l=0, w=3, h=1},
-            text={
-                {tile=button_pen_left},
-                {tile=help_pen_center},
-                {tile=button_pen_right},
-            },
-            on_click=function() dfhack.run_command('gui/launcher', command) end,
-        },
-    }
+    self.on_click = function() dfhack.run_command('gui/launcher', command) end
```
SilasD, line 32: if line 26 is redundant, then this is as well.  but harmless.

library/lua/gui/widgets/buttons/radio_button.lua
```diff
@@ -0,0 +1,49 @@
+-- A 3x1 tile button that toggles state when clicked
+
+local textures = require('gui.textures')
+local ConfigureButton = require('gui.widgets.buttons.configure_button')
+
+local to_pen = dfhack.pen.parse
+
+local enabled_pen_left = to_pen{fg=COLOR_CYAN,
+    tile=curry(textures.tp_control_panel, 1), ch=string.byte('[')}
```
SilasD, line 9: I do not believe the `or nil` clauses are necessary, but I will pint out this tile doesn't have one.

library/lua/gui/widgets/buttons/radio_button.lua
```diff
@@ -0,0 +1,49 @@
...
+
+function RadioButton:init()
+    self.toggle_state = self.initial_state
+
+    self.on_click = function() self.toggle_state = not self.toggle_state end
+    self.pen_left = function() return self.toggle_state and enabled_pen_left or disabled_pen_left end
+    self.pen_center = function() return self.toggle_state and enabled_pen_center or disabled_pen_center end
+    self.pen_right = function() return self.toggle_state and enabled_pen_right or disabled_pen_right end
+end
+
```
SilasD, line 48: no getter function for the toggle state?  are users supposed to probe an instance's `toggle_state` field ?

yeah, that's what the new documenation says.  hm.

I don't like that.

are there other widgets that similarly require looking inside them for state info?

also, no callbacks?  `on_enable`, `on_disable`, `on_toggle` ?`

I *guess* a user could catch the `on_click` that the ConfigureButton base class sets up, and then probe the `toggle_state`.

SilasD (commented):
I'm okay with this, I guess.

RadioButton looks a little bit half-baked, but it's in a usable state.

ab9rf:
For the record: Pure code-quality changes that don't have user-facing effects do not require a changelog.

library/lua/gui/widgets/buttons/configure_button.lua
Bumber64: Worth bothering Myk to ask if he remembers why it's used?

library/lua/gui/widgets/buttons/help_button.lua
Bumber64: I had a `GraphicButton` class before I removed it. It just made `ConfigureButton` a small subclass full of redundant code, but it could be added back for organization.

library/lua/gui/widgets/buttons/configure_button.lua
Bumber64: The size doesn't need to be overwritten, but help buttons must align to the right side of their parent panel.

library/lua/gui/widgets/buttons/help_button.lua
Bumber64: There's an `r=1` in the frame. Help buttons always default to the top right, and all existing uses have them left at default.

`ConfigureButton` doesn't have `command`. Our existing help buttons take a string, which needs to converted to an `on_click` function.

library/lua/gui/widgets/buttons/radio_button.lua
Bumber64: Closest I can think of is checking if a widget is `hidden`. `CycleHotkeyLabel` and `EditField` let you define an `on_change` callback that receives the new value.

library/lua/gui/widgets/buttons/help_button.lua
SilasD: right, no `command` in `ConfigureButton`.  I thought I saw one in it.  I wonder where I picked that up.  the whole gui system only has the one occurance of `command` in the `HelpButton`.  my bad there.

library/lua/gui/widgets/buttons/help_button.lua
SilasD: would it be useful by itself, or or is it (the equipvalent of) an abstract class?

if it could be used as is, with some configuration obviously, then maybe so.

library/lua/gui/widgets/buttons/configure_button.lua
SilasD: I don't think so... from what I've seen rome only bothers him if it's necessary.  details about three year old code, probably not necessary.

anyway, from the docs, emphasis added:

> `getTexposByHandle(handle)`

> Get the current `texpos` for the given `TexposHandle`. Always use this method to get the `texpos` for your texture. **`texpos` can change when game textures are reset**, but the handle will be the same.

so it's documented as best practive.  since this is all likely myk's code, I think just go with what is known to work.

not sure if this applies to textures under DFHack control, but I suppose it's reasonable.  the LuaApi.cpp uses modules/Textures.cpp `Textures::loadTexture` which has a fast path for the most common cases, so not much slower than consulting a Lua table or such.  Edit: *faster* than consulting two or three Lua tables probably.

library/lua/gui/widgets/buttons/configure_button.lua
SilasD: ~~I don't see how those `frame` settings lead to right-align.~~

~~the docs imply that a `r=0` or `r=1` would set right-align or right-align with a 1 character gap.~~

forget that, wrong file.

library/lua/gui/widgets/buttons/help_button.lua
Bumber64: It was just `ConfigureButton` by another name, which is already conceptually a 3x1 button than runs a callback. The only defining trait of config is that it has a gear and the name is already in use.

Bumber64:
Did we confirm that the `---@field` stuff is all okay? It wasn't mentioned in review.

IIRC, it's used for a Lua language server, which I don't use.

SilasD:
> Did we confirm that the `---@field` stuff is all okay? It wasn't mentioned in review.
> 
> IIRC, it's used for a Lua language server, which I don't use.

It all looked good.  the `widgets.RadioButton.attrs` class probably wants
`---@field toggle_state? boolean`
but it's no more broken than all the other definitions.

when I get my local fork of vallode's LuaLS whipped into better shape I'll spend some time with all of these classes' annotations.

library/lua/gui/widgets/buttons/radio_button.lua
Bumber64: Created #5850 to address this.

PR #5809, opened 2026-05-19
stocks: add expand_all to stocks interface

### LINKED ISSUE #5338: Stocks screen option to "expand all" categories
The stocks overlay lets you collapse all categories, but there isn't a way to undo that by expanding all. You have to exit and re-enter stocks if you don't want to reopen categories by hand.

emreduzenli:
Fixes https://github.com/DFHack/dfhack/issues/5338


plugins/lua/stocks.lua
```diff
@@ -14,6 +14,14 @@ local function collapse_all()
     stocks.i_height = num_sections * 3
 end
 
+local function expand_all()
+    local num_sections = #stocks.current_type_a_expanded
+    for idx=0,num_sections-1 do
+        stocks.current_type_a_expanded[idx] = true
+    end
+    stocks.i_height = num_sections * 3
```
ab9rf: I'm uncertain that this is the correct thing to do here, but I also don't understand why `collapse_all` does the same thing. So if someone could explain this to me, it would help me feel better about approving this change.

chdoc: I think you were right to be suspicious. `stocks.i_height` appears to be the total height of the right hand side (in interface lines, 3 per entry). When expanding all entries, the total height of the right inside could be computed as simply as:
```suggestion
    num_items = #stocks.current_type_i_list
    stocks.i_height = (num_items + num_sections) * 3
```

emreduzenli: thank you for the clarification, I've updated the branch to reflect your suggested way of computing it

ab9rf (approved):

PR #5822, opened 2026-06-01
move infinite sky's actual work code to `Maps`

ab9rf:
the infinite sky plugin will still handle the "automatic" extension, but the code to actually extend the map block columns moves to the Maps module

this is due to a request from a modder to be able to extend map block columns from scripts without having to rely on a plugin API call


library/modules/Maps.cpp
```diff
@@ -1522,3 +1531,129 @@ int Maps::removeAreaAquifer(df::coord pos1, df::coord pos2, std::function<bool(d
...
+        // Allocate a new block column and copy over data from the old
+        df::map_block** blockColumn =
+            new df::map_block * [z_count_block + quantity];
+        std::memcpy(blockColumn, block_index[bpos.x][bpos.y],
+            z_count_block * sizeof(df::map_block*));
+        delete[] block_index[bpos.x][bpos.y];
+        block_index[bpos.x][bpos.y] = blockColumn;
+
+        df::map_block* last_air_block = blockColumn[bpos.z];
+        for (int32_t count = 0; count < quantity; count++)
```
ctrlaltmilk, line 1559: Is there a reason you're using a count rather than looping over Z position? (i.e. `for (int32_t z_pos = last_air_block->map_pos.z + 1; z_pos < z_count_block + quantity; z_pos++)`)
Might just be a me thing but the 
```
air_block->map_pos = last_air_block->map_pos;
air_block->map_pos.z += count + 1;
```
later is a little sketchy IMHO.

library/modules/Maps.cpp
```diff
@@ -1522,3 +1531,129 @@ int Maps::removeAreaAquifer(df::coord pos1, df::coord pos2, std::function<bool(d
...
+
+    cuboid last_air_layer(
+        0, 0, world->map.z_count_block - 1,
+        world->map.x_count_block - 1, world->map.y_count_block - 1, world->map.z_count_block - 1);
+
+    last_air_layer.forCoord([&] (df::coord bpos) {
+        // Allocate a new block column and copy over data from the old
+        df::map_block** blockColumn =
+            new df::map_block * [z_count_block + quantity];
+        std::memcpy(blockColumn, block_index[bpos.x][bpos.y],
```
ctrlaltmilk, line 1553: Is `std::memcpy` worth replacing with `std::copy` for type-safety and not having to mess with sizeof?

ctrlaltmilk (commented):
Review mentioned on Discord, mostly minor questions

library/modules/Maps.cpp
ab9rf: Yes, but this is being uncooperative for me right now so I am going to defer this for later.

library/modules/Maps.cpp
ab9rf: replaced with `air_block->map_pos = last_air_block->map_pos + df::coord{0, 0, uint16_t(count + 1)};`

going to leave using the `count` iterator for now, though; may revisit later (or not)

PR #5823, opened 2026-06-01
make `getConfigPath` a public `Core` API

ab9rf:
also export to Lua as `dfhack.getConfigPath`

mainly so that we can move `dfhack-config` without having to update hundreds of source locations

### DIFF
--- a/library/Core.cpp
+++ b/library/Core.cpp
@@ -135,16 +135,6 @@ namespace DFHack {
     DBG_DECLARE(core, keybinding, DebugCategory::LINFO);
     DBG_DECLARE(core, script, DebugCategory::LINFO);
 
-    static const std::filesystem::path getConfigPath()
-    {
-        return Filesystem::getInstallDir() / "dfhack-config";
-    };
-
-    static const std::filesystem::path getConfigDefaultsPath()
-    {
-        return Core::getInstance().getHackPath() / "data" / "dfhack-config-defaults";
-    };
-
 class MainThread {
 public:
     //! MainThread::suspend keeps the main DF thread suspended from Core::Init to
@@ -538,9 +528,9 @@ std::filesystem::path Core::findScript(std::string name)
     return {};
 }
 
-bool loadScriptPaths(color_ostream &out, bool silent = false)
+bool loadScriptPathsCore(Core& core, color_ostream &out, bool silent = false)
 {
-    std::filesystem::path filename{ getConfigPath() / "script-paths.txt" };
+    std::filesystem::path filename{ core.getConfigPath() / "script-paths.txt" };
     std::ifstream file(filename);
     if (!file)
     {
@@ -563,7 +553,7 @@ bool loadScriptPaths(color_ostream &out, bool silent = false)
         getline(ss, path);
         if (ch == '+' || ch == '-')
         {
-            if (!Core::getInstance().addScriptPath(path, ch == '+') && !silent)
+            if (!core.addScriptPath(path, ch == '+') && !silent)
                 out.printerr("{}:{}: Failed to add path: {}\n", filename, line, path);
         }
         else if (!silent)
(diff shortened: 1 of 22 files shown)


library/include/LuaTools.h
```diff
@@ -43,6 +33,17 @@ distribution.
...
+#include <functional>
+#include <map>
+#include <set>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
```
chdoc, line 37: I think that, if you're actively reordering includes, it is time to adapt the [contributing guidelines](https://docs.dfhack.org/en/stable/docs/dev/Contributing.html#id9) to the current practice.

library/Core.cpp
```diff
@@ -538,9 +528,9 @@ std::filesystem::path Core::findScript(std::string name)
     return {};
 }
 
-bool loadScriptPaths(color_ostream &out, bool silent = false)
+bool loadScriptPathsCore(Core& core, color_ostream &out, bool silent = false)
```
chdoc: I do not understand the rationale for this signature change, but I assume that you have your reasons.

library/LuaApi.cpp
```diff
@@ -1384,6 +1385,7 @@ static const LuaWrapper::FunctionReg dfhack_module[] = {
     WRAP(getDFPath),
     WRAP(getTickCount),
     WRAP(getHackPath),
+    WRAP(getConfigPath),
```
chdoc, line 1388: The new export needs to be documented in the LuaAPI documentation.

plugins/orders.cpp
```diff
@@ -44,8 +44,15 @@ DFHACK_PLUGIN("orders");
...
-static std::filesystem::path ORDERS_LIBRARY_DIR = Core::getInstance().getHackPath() / "data" / "orders";
+static const std::filesystem::path get_orders_dir()
+{
+    return Core::getInstance().getConfigPath() / "orders";
+}
+
+static std::filesystem::path get_orders_library_dir()
+{
+    return Core::getInstance().getHackPath() / "data" / "orders";
+}
```
chdoc, line 55: Are you expecting problems with the initialization order of static variables, if the configuration location is determined at startup? If it is unsafe to store the result of `getConfigPath` and `getHackPath` in a static variable, then I think this deserves a comment in the header file where these functions are declared.

chdoc (changes requested):
This (obviously) needs a changelog entry as well as changes to the documentation of the LuaAPI. Other than that, all changes look reasonable to me, even though I do not fully understand the motivation for all of the changes.

library/Core.cpp
ab9rf: I don't remember why I did that either, and I'm not sure there is a good reason for it. I'll take another look.

ab9rf: Mostly this is because I still want to [change the lifecycle of `Core`](https://github.com/DFHack/dfhack/pull/5684) and so static initialization based on the static core instance would be wrong because the core instance simply wouldn't exist at initial DLL load.

Also, since #5784, DFHack doesn't know what `hack_path` _is_ until `dfhooks_init` is called, which won't happen until _after_ static initialization is complete. Thus, no global static variable can depend on `hack_path`, directly or indirectly, because it won't have a meaningful value during at-start static initialization.

All that said, this is suboptimal and it would make more sense to make this a `Core` method instead of a free function.

plugins/orders.cpp
ab9rf: The issue that I don't think we want to guarantee that these will never change for the life of the `Core` instance. (Plugins are loaded during `Core` startup and unloaded during `Core` shutdown, so this shouldn't actually be an issue, but note that when a plugin is loaded, unloaded, and loaded again, there is not a guarantee that at-start static initialization takes place a second time.) Currently they don't, but I can't rule that out in the future.

It is probably reasonable to document this.

library/include/LuaTools.h
ab9rf: I didn't follow the guidelines here, which is my fault, Will fix.

chdoc (approved):
Looks good to me.

PR #5824, opened 2026-06-02
buildingplan: add Pull button for linked levers 

gwilymtv:
Add a Pull/Queued button next to linked levers on the "Show linked buildings" tab so a pull-lever job can be queued (do-now priority) without navigating to the lever. Safety/appropriateness of the job is left up to the player (e.g. no check for lever being linked to other buildings).

Feature-wise maybe not the best fit for `buildingplan` overall? Doesn't seem like a big tangent from the existing `unlink_mechanism` module though. Open to alternatives.

Uses `lever.leverPullJob` from the existing lever script. Happy to inline that though if that's risky.

image

image
(Edit: "Queued" is now clickable to cancel the job)

image

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -75,6 +75,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `buildingplan`: add a ``Pull`` button next to linked levers on a building's "Show linked buildings" tab so you can queue a pull-lever job without navigating to the lever
 
 ## Fixes
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -57,6 +57,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `buildingplan`: add a ``Pull`` button next to linked levers on a building's "Show linked buildings" tab so you can queue a pull-lever job without navigating to the lever
 
 ## Fixes
 
@@ -75,7 +76,6 @@ Template for new versions:
 ## New Tools
 
 ## New Features
-- `buildingplan`: add a ``Pull`` button next to linked levers on a building's "Show linked buildings" tab so you can queue a pull-lever job without navigating to the lever
 
 ## Fixes
 
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -57,7 +57,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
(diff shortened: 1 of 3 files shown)


gwilymtv:
Didn't realise there was a very similar PR already open. When I get a chance to circle back to this I'll take a look at the feedback that was on that other PR already because at a glance it might be repeated here.

ab9rf:
> Didn't realise there was a very similar PR already open. When I get a chance to circle back to this I'll take a look at the feedback that was on that other PR already because at a glance it might be repeated here.

It looks like the other PR has been withdrawn. I'd appreciate if someone reviewed this one relative to the withdrawn one

gwilymtv:
@chdoc compared to #5725 (separate overlay, duplicated layout machinery), this adds the Pull button to the existing `MechLinkOverlay`, reusing both it and `lever.leverPullJob` though at the cost of perhaps still exacerbating an existing issue.

What's left seems to be this list-row-decoration mechanism, which (as best as I can as a new contributor) I can't spot an existing framework for, and already looks to be duplicated between this and `MechItem`.

Is that what you had in mind for extraction?

Not to say that it isn't worth it, but it's of course more than I had in mind. I need to check directions here before deciding what to do next.

chdoc:
Without going too much into the technical details, I prefer the approach here. Putting both functionalities into the same overlay avoids the need to communicate between two overlays. With this design, I am fine with this being part of the existing tool.

Regarding the visual information, I prefer what was shown in #5725. In particular, I like that it shows the state of the connected lever and whether the lever pull job has already been taken up by a citizen. Maybe a combination of the two would be even clearer: `Pull /` (white), `Queued /` (yellow), `Pulling /` (green), `Pull \` (white).

> Uses `lever.leverPullJob` from the existing lever script. Happy to inline that though if that's risky.

I'm not a fan of the dependency of this tool on something that is in the `scripts` repository. However, `lever` depending on `buildingplan` it doesn't sound like a clean solution either. At this point, it will probably be best to add this to our Lua API. However, that doesn't have to be done in this pull request. 

I will try to look at the details of the implementation at a later point.

gwilymtv:
@chdoc agreed on the Lua API note. I've updated this to show both lever and job state (Pull/Pulling/Queued, with colours), with one extra accommodation for ASCII mode, because (to me at least) the two modes contradict in their reading of lever "direction," so I thought I'd adopt the ASCII glyph in ASCII mode:

Premium shows:
`[Pulling /]` (off/zero lever state)

ASCII shows:
`[Pulling ò]` (off/zero lever state)

PR #5832, opened 2026-06-19
Draft: dig-now plugin: remove dependencies on MapCache module.

SilasD:
I request **code review** on this draft PR, with a focus on `dig-now.cpp` lines 60 to 380 or so, which contain new code, and the `dig_type()` function in lines 624-682, which is a complete reimplementation.

I would like this code to be up to modern C++ standards, and I know that my C++ knowledge and skills are not the best.

The remainder of the file is old code updated with different types, e.g. `df::coord` instead of `DFCoord`, and removal of all references to the `MapExtras::MapCache &map` object.

There are also a couple of minor bugfixes, in particular in the `dig_tile()` function `Channel` case.

I don't think that part of the plugin needs full review yet, but I do welcome comments.

LuaLS annotations were added to `dig-now.lua`.  No code changes.

General notes:

This PR is intended to be a naive-but-correct reimplementation of dig-now using the Maps module instead of the MapCache module.  Note that it does not have the caching that MapCache provided.  This results in a lot of unnecessary info lookups.

This PR is **known to have bugs**.  These bugs already existed in the old version.  They will be fixed in the near future.

This code has been verified to be *almost* bug-for-bug compatible with the old version.  It results in the same map state with these exceptions:
* Blueprint-mode designations for carving fortifications and tracks: the old version improperly carves them even though they are set as blueprints, and it also doesn't clear `map_block.flags.designated`.  This new version replicates the buggy behavior, but it properly clears that flag.  I can't be bothered to track down the exact line and state to replicate that tiny bug, when I expect to fix the larger bug soon.
* Undiggable ramps are no longer improperly removed by the `clean_ramp()` function.  This has been seen on tower-cap cap ramps, but has not been verified for all undiggable ramps.

Intended future directions, roughly in this order:
* Extend the Lua API calls or create new parallel calls to allow Lua scripts to process a list of tiles instead of processing one tile per invocation.
* The Lua API calls must be able to request dig/smooth operations instead of depending on map tile designations.  This is because one of the expected uses is for adventure-mode mods to alter the map, and adventure mode uses the dig and smooth bits for other purposes.  The current code does not support this use case.
* Fix many, many ways that dig-now doesn't leave the map in the same state as actual mining does.  See bug-report issue #TODO.
* Make the core loop map-block-centric, perhaps using `Maps::cuboid::forBlock()`.  Currently it is a naive
layer-by-layer, column-by-column, row-by-row scan.
* Re-add caching of region data, biome and geo-biome data, and material data.  This reimplementation does a lot of unnecessary info lookups.


plugins/dig-now.cpp
```diff
@@ -48,8 +57,336 @@ namespace DFHack {
...
+    using namespace df::enums::builtin_mats;
+    auto geo_biome = geo_biome_at(block, p);
+    if (!geo_biome)
+        return t_matpair(INORGANIC, -1);    // can't happen, generic "rock"
+    // gcc-11 complained that 0 was of different signed-ness than size_t!  trying to cast it to unsigned.
+    layer = clip_range(layer, static_cast<size_t>(0), geo_biome->layers.size() - 1);
+    return t_matpair(INORGANIC, geo_biome->layers[layer]->mat_index);
+}
+
+#if false   // gcc warns about unused functions, and warnings are errors.
```
ab9rf, line 117: use `[[maybe_unused]]` here to suppress the warning - the compiler will not warn about the function being unused but will still validate its syntax which doesn't happen with the `#if`

long term, if this is actually dead code it should simply be removed, there's no need to preserve it for reference or legacy purposes as it'll be forever in the repo change history

SilasD: Thank you for the compiler directive.

I'm uncertain if it is dead code or not; I originally wrote these functions wiuth the code combined into one function each, then decided to split out the `map_block` and local `pos` cases.

I will maintain it for the moment, but my long-term intent is to move all the code to block-centric algorithms, so this will likely be permanently dead code and I will remove it then.

plugins/dig-now.cpp
```diff
@@ -48,8 +57,336 @@ namespace DFHack {
...
+
+static const df::inclusion_type veinTypeAt(df::coord pos)
+{
+    auto block = Maps::getTileBlock(pos);
+    if (!block)
+        return df::enums::inclusion_type::TOTAL;
+    auto blockref = const_cast<const df::map_block &>(static_cast<df::map_block &>(*block));
+    return veinTypeAt(blockref, df::coord2d(pos) & 15);
+}
+
```
ab9rf, line 389: Should we be adding some or all of these as custom methods to the relevant codegen types? This is something we can do.

PR #5834, opened 2026-06-24
Add dfhack.filesystem.rmdir_recursive to proxy std::filesystem::remove_all

sizzlins:
Exposes \std::filesystem::remove_all\ to Lua as \dfhack.filesystem.rmdir_recursive\.

Required for PR DFHack/scripts#1583 (\manual-save\) to implement the \--cleanup\ snapshot pruning feature natively.

### DIFF
--- a/docs/dev/Lua API.rst
+++ b/docs/dev/Lua API.rst
@@ -3291,6 +3291,10 @@ unless otherwise noted.
 
   Removes a directory. Only works if the directory is already empty.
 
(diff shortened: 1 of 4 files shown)


ab9rf (approved):

ab9rf:
This looks good except for one thing: it needs a changelog

ab9rf:
After some consideration, I've decided that this is too risky to add without some safety measures - it would be far too easy for someone to use this to do significant damage. Some sort of protection to limit what files can be removed seems necessary.

PR #5840, opened 2026-07-06
Autoclothing fix

### LINKED ISSUE #5837: autoclothing order parsing: material not matching
autoclothing commands such as `autoclothing leather vest 1` are not accepted in 53.15.  (they are accepted in 53.14.)
it fails with the response `leather is not a valid material category for vest`.

reported on Discord: [what did i mess up here?](https://discord.com/channels/793331351645323264/873014631315148840/1520451068910239894).

this may be more general:
reported on Discord: [hm is material subtype matching just completely busted currently](https://discord.com/channels/793331351645323264/1321200750000341034/1520380159499173929).

LightHardt:
When printing out the values of variables along the path for validating the material type I saw that `requirement.material_category.whole` was 0. So in the `armorFlagsMatch` function it was trying to check bits that weren't set. From a quick look over it seemed the purpose is just to check that the user specified a material that is valid for the armor wanted. With this in mind I passed in the material_category that the user wants and in game was able to get the desired behavior of creating orders of cloth gloves. I also saw the desired behavior of invalid material type when i tried ordering wood gloves.

This tries to fix #5837, please let me know if any issues

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -33,6 +33,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- `autoclothing`: Fix incorrect invalid material type error
 
 ## Misc Improvements
 
--- a/plugins/autoclothing.cpp
+++ b/plugins/autoclothing.cpp
@@ -65,7 +65,7 @@ enum ConfigValues {
 struct ClothingRequirement;
 command_result autoclothing(color_ostream &out, vector<string> &parameters);
 static void do_autoclothing();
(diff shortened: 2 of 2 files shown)


ab9rf:
There is a `material_requirement` variable inside of a `ClothingRequest` which should be used here instead of passing it separately. If the `material_requirement` inside the request isn't being set correctly, the fix is to correct the constructor, not to pass that information along separately

LightHardt:
@ab9rf Understood. See you put a fix up will close this one.

PR #5841, opened 2026-07-06
add protective code against vanilla misbehavior

ab9rf:
add protective code to avoid misoperation (as much as possible) when the game's plant list contains nonsense

### DIFF
--- a/plugins/getplants.cpp
+++ b/plugins/getplants.cpp
@@ -478,41 +478,52 @@ command_result df_getplants(color_ostream& out, vector <string>& parameters) {
     }
 
     count = 0;
-    for (size_t i = 0; i < world->plants.all.size(); i++) {
-        const df::plant* plant = world->plants.all[i];
+    for (auto* plant : world->plants.all)
+    {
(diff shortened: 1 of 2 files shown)


plugins/getplants.cpp
```diff
@@ -478,41 +478,52 @@ command_result df_getplants(color_ostream& out, vector <string>& parameters) {
...
     count = 0;
-    for (size_t i = 0; i < world->plants.all.size(); i++) {
-        const df::plant* plant = world->plants.all[i];
+    for (auto* plant : world->plants.all)
+    {
         df::map_block* cur = Maps::getTileBlock(plant->pos);
 
-        TRACE(log, out).print("Examining {} at ({}, {}, {}) [index={}]\n", world->raws.plants.all[plant->material]->id, plant->pos.x, plant->pos.y, plant->pos.z, (int)i);
+        auto mat = plant->material;
+        if (mat < 0 || mat > int16_t(world->raws.plants.all.size()))
```
quietust: `>=`?

Side note, checking vector index validity sounds like something we should have a helper function for.

PR #5847, opened 2026-07-10
changes required as part of changing codegen to use `std::array`

ab9rf:
Note: no separate changelog, the changelog on structures is enough to cover


ab9rf:
Note: merge DFHack/df-structures#889 when merging this

PR #5850, opened 2026-07-13
`RadioButton` use `on_change` callback

Bumber64:
Have `RadioButton` use a callback when it changes state instead of needing to check manually.


Bumber64:
Not sure if `toggle_state` should have a getter function. Maybe there's a reason to check the state outside the callback, but you can just read the value if you want it enough.

SilasD (commented):
I think this covers all of my quibbles.  I'm good with this, asuming it's had a bit of testing.

Edit: in response to the above, I think a `getState()` to match the `setState()` would be good -- if I were just trying to use this, I would expect the matching function to exist.

PR #5852, opened 2026-07-17
planner: adjust "upper" bottom-anchored widget positions (hollow constructions, stair type, engraved slabs, empty cages, weapon count)

ChrisJohnsen:
As reported in [Discord][d], the "Hollow area" UI element in buildingplan's planneroverlay disappeared in 53.15-r2.

[d]: https://discord.com/channels/793331351645323264/793331351645323267/1527090699491737722

It is still there (its hotkey still works, and it is still actually clickable—if you already knew its hotkey and/or where to blind-click), but the divider is being drawn over it.

When 39cc838655f3a97c74cb6f4cc57d17c9f59cec8d added an extra UI row for the "queue order" button, the bottom-anchored UI elements in the upper area of the main_panel were not adjusted to account for the extra line.

The "stair type", "engraved slabs only", "empty cages only", and "weapon count" planner UI elements (all similarly conditional) are affected in the same way.

The commit bumps all the bottom-anchored UI elements in the "upper" portion of the planner overlay up by one UI row.

### DIFF
--- a/docs/changelog.txt
+++ b/docs/changelog.txt
@@ -59,6 +59,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
+- `buildingplan`: restore planner UI elements: hollow constructions, only engraved slabs, only empty cages, weapon count
 
 ## Misc Improvements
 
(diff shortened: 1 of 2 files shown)


ChrisJohnsen:
Here is a visual overview of the widget positions in 53.15-r1, 53.15-r2, and this PR:

ab9rf (approved):

PR #5855, opened 2026-07-19
Update Lua API documentation: Add parameters description table to dfhack.translation.generateName()

agnieszka-dev:
Adds a table containing parameter descriptions and types to dfhack.translation.generateName() documentation. This table format is inspired by ArcPy documentation ([example](https://doc.esri.com/en/arcgis-pro/latest/tool-reference/analysis/buffer.html?tabs=python#parameters)) and is meant to be especially friendly for scripting beginners looking up how to use Lua API functions. 

Ideally, this should serve as an example for expanding the remainder of Lua API documentation in the future.

Preview in VSCode using Esbonio:
obraz

I also added a small modification of the general description to explicitly inform the user that the name that will be generated is randomized.

### DIFF
--- a/docs/styles/dfhack.css
+++ b/docs/styles/dfhack.css
@@ -80,11 +80,11 @@ aside.dfhack-tool-summary p:last-child {
    margin-bottom: 0;
 }
 
-table {
+table.dfhack-param-table {
     table-layout: fixed;
     width: 110%;
 }
 
-table span.pre {
+table.dfhack-param-table span.pre {
     white-space: wrap;
 }
--- a/docs/styles/dfhack.css
+++ b/docs/styles/dfhack.css
@@ -80,7 +80,6 @@ aside.dfhack-tool-summary p:last-child {
    margin-bottom: 0;
 }
 
-
 table {
     table-layout: fixed;
     width: 110%;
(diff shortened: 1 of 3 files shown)


chdoc:
Certainly adds valuable information and I am definitely not opposed to tables in general. Looking at the [output](https://dfhack--5855.org.readthedocs.build/en/5855/docs/dev/Lua%20API.html#translation-module), this tabtle considerably overruns the (admittedly quite limited) width of our current documentation. Increasing the margins of the documentation is a possibility, but of course bears the risk of widespread typesetting issues all over the place. Would it be possible to limit the width of the table to something that doesn't look quite as much out of place as the current table?

Not your fault, but most of our Lua API actually returns objects instead of using output parameters. I think it would be worthwhile to explicitly state that `name` is the output parameter (e.g. "the name object being modified (output parameter)")

agnieszka-dev:
@chdoc I managed to make the table narrower. It did require enabling wrapping for inline code though. What do you think? 

obraz

ab9rf (approved):

docs/styles/dfhack.css
```diff
@@ -79,3 +79,12 @@ div.dfhack-tool-summary p:last-child,
 aside.dfhack-tool-summary p:last-child {
    margin-bottom: 0;
 }
+
+table {
```
lethosor: These rules should be made more specific if they are necessary, e.g. with a custom CSS class on the `table`

lethosor (dismissed):

PR #5857, opened 2026-07-23
Remove MaterialInfo consts

quietust:
in favor of using df::builtin_mats enum entries

Also adjust some comparisons to use more easily understandable values (e.g. <= CREATURE_200 instead of < HIST_FIG_1)


quietust:
I'm open to leaving some of those constants behind, such as `NUM_BUILTIN`, since some of them might not be immediately obvious what you should use instead. `GROUP_SIZE` is something that needs to be removed, though, since while the 3 groups are currently the same size, there's no guarantee that this will continue to be the case in the future.

ab9rf (approved):

Issue #5865, opened 2026-08-05
std::array assertion in doSetTile_char for non-optimized builds


ChrisJohnsen:
(This is more detail on a cause of the assertion that prompted #5864.)

The combination of a local, non-optimized build of DFHack (*not production
builds*) made with GCC 15 and the recent change to using `std::array`
highlighted a potential out-of-bounds indexing in `doSetTile_char`. Just
starting `gui/launcher` is sufficient to trigger a hard crash that does not
apparently produce any logs (not on the DFHack console, not in the `stderr.log`,
nor in the `errorlog.txt`). This is not the only trigger for this assertion.

# Stack Trace

Asking GDB for a backtrace shows these as the uppermost frames (excluding the
internals above the assertion and below `do_render`):

    std::__glibcxx_assert_fail
    std::array<std::array<unsigned char, 3ul>, 16ul>::operator[](unsigned long)
    doSetTile_char(DFHack::Screen::Pen const&, int, int, bool)
    doSetTile_default(DFHack::Screen::Pen const&, int, int, bool, int* df::graphic_viewportst::*)
    doSetTile(DFHack::Screen::Pen const&, int, int, bool, int* df::graphic_viewportst::*)
    DFHack::Screen::paintTile(DFHack::Screen::Pen const&, int, int, bool, int* df::graphic_viewportst::*)
    screen_paintTile(lua_State*)
    luaD_precall(lua_State*, lua_TValue*, int)
    luaV_execute(lua_State*)
    luaD_call(lua_State*, lua_TValue*, int)
    luaD_callnoyield(lua_State*, lua_TValue*, int)
    lua_callk(lua_State*, int, int, long, int (*)(lua_State*, int, long))
    DFHack::dfhack_lua_viewscreen::do_render(lua_State*)

The `std::array<std::array<unsigned char, 3>, 16>::operator[]` looked
suspiciously like the indexing of `gps->uccolor`.

Dropping a debug print into `doSetTile_char` just before `uccolor` indexing
shows that `bg` was 255 in my crash scenario (this happens even in optimized
GCC 12 builds).

# Source of This Particular Crash

The TextAreaComponent used in the launcher's input field sets a default `bg` of
COLOR_RESET (-1). It isn't overridden in the case of the launcher, so it makes
it into the Pen instance as a `bg` data member with values -1. When used as
a `uint8_t` (in `doSetTile_char`) the value is 255 and is OOB for the
16-element `gps->uccolor` array.

# Possible Fixes

[...]

PR #5866, opened 2026-08-05
Screen::Pen: handle COLOR_RESET color values

ChrisJohnsen:
Here is some mitigation code that I put together for #5865.

It largely focuses on the Lua-originated color value handling. Perfect color value protection seems much less tractable on the C++ side since the relevant data members are all public.

Note: this will prevent the logging in #5864 from triggering for Lua-originated sources of out-of-range color values. Should similar logging be done in the places where this change does defaulting and masking?

If COLOR_RESET handling and automatic masking is thought to be good idea, it might be worth un-static-ing the function that does the "reset handling and masking" and use it in the few places (ctors, etc.) that the "C++ code" in this PR does simple masking. Overall there seems to be much less drawing code in C++, and what I saw (from asking clangd for Pen ctor uses) seemed okay (all using "proper" color values).

The color "masking" behavior is not covered in the unit tests, since I was not sure if it was the overall right direction.


ChrisJohnsen:
I've tried a few more variations locally:
- raise Lua error on out of range value
- mask color to 4 bits, show warning and traceback
- just show warning with traceback

The traceback from raising an error can be helpful in tracking down the source of the erroneous color, but actually raising an error can be pretty intrusive (due to minimal `pcall`/`safecall` in the rendering code paths).
Each Lua viewscreen gets a safecall, but overlays are not as isolated. An (uncaught) error while rendering a Screen or ZScreen will prevent "later" parts of that same viewscreen from being rendered. An uncaught error while rendering an overlay will prevent other overlays from rendering (*NB*: overlay render order changes from launch to launch due to the indeterminacy of `pairs` iteration order).

Capturing a trackback without raising an actual error provides most of the benefit of raising an error (except the big, obvious render breakage, which is of possibly dubious value…).

Masking the value might not be terribly useful since there is now code to check for out-of-bounds colors before they are actually used (#5864).

I'm leaning towards "warn with traceback" (no masking) and letting `doSetTile_char` handle filtering out the bad values.

ChrisJohnsen:
#5910 covers part of what this PR addressed. This one can be closed, unless you want any of its other bits recast along the lines of what 5910 does.

PR #5875, opened 2026-08-13
Add smooth-movement plugin

notliad:
## Summary

Adds the `smooth-movement` plugin to DFHack. It provides smooth visual interpolation for creatures, items, vehicles, and related overlays in the fortress viewport.

## Notes

I closed the previous PR in favor of this version, which preserves the original code structure that was already reviewed by humans in the standalone repository.

Standalone repository: https://github.com/notliad/df-smooth-movement

## Testing

- Built the `smooth-movement` plugin through the DFHack build system.
- Built and ran `smooth-movement-test`.
- All regression assertions passed.


plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,1314 @@
+// SPDX-License-Identifier: MIT
```
ab9rf: From our license statement
> The core, plugins, scripts, and other DFHack code all use the ZLib license unless noted otherwise. By contributing to DFHack, authors release the contributed work under this license.

Please relicense under ZLib so I don't have to add an additional line to our licenses statement for this plugin

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,1314 @@
...
+decltype(&SDL_RenderFillRect) render_fill_rect = nullptr;
+decltype(&SDL_RenderSetClipRect) render_set_clip_rect = nullptr;
+decltype(&SDL_GetRenderDrawColor) get_render_draw_color = nullptr;
+decltype(&SDL_SetRenderDrawColor) set_render_draw_color = nullptr;
+
+visual_animation_managerst animation_manager;
+std::set<std::pair<int32_t, int32_t>> previous_coverage;
+uint64_t visual_context_revision = 0;
+const void *previous_viewport = nullptr;
+std::array<int32_t, 12> previous_view_signature{};
```
ab9rf: why is this an array? looking elsewhere, it looks like each of these 12 elements has a fixed, established meaning, which means this should be a proper compound of some sort

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,1314 @@
...
+// Runtime harness for the engine-owned visual state; gameplay data is never read.
+decltype(&SDL_RenderCopyF) render_copy_f = nullptr;
+decltype(&SDL_RenderCopyExF) render_copy_ex_f = nullptr;
+decltype(&SDL_RenderFillRect) render_fill_rect = nullptr;
+decltype(&SDL_RenderSetClipRect) render_set_clip_rect = nullptr;
+decltype(&SDL_GetRenderDrawColor) get_render_draw_color = nullptr;
+decltype(&SDL_SetRenderDrawColor) set_render_draw_color = nullptr;
+
+visual_animation_managerst animation_manager;
+std::set<std::pair<int32_t, int32_t>> previous_coverage;
```
ab9rf: why is this not a `tile_coveragest`?

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,1314 @@
...
+    float source_x;
+    float source_y;
+    int32_t target_x;
+    int32_t target_y;
+    int32_t texpos;
+    float progress;
+    SDL_Texture *texture;
+    bool mirrored = false;
+    int32_t mirror_shift = 0;
+    std::set<std::pair<int32_t, int32_t>> coverage;
```
ab9rf: this should also be a `tile_coveragest`

plugins/smooth-movement/visual_animation.h
```diff
@@ -0,0 +1,793 @@
...
+    viewport_visual_layer layer;
+    visual_render_groupst render_group;
+    bool moves_independently;
+    bool matches_any_previous;
+    uint8_t draw_order;
+    int8_t center_x;
+    int8_t center_y;
+};
+
+constexpr std::array visual_layer_descriptors = {
```
ab9rf: where are the magic numbers here coming from?

ab9rf:
A general comment as this is a pervasive issue and I do not wish to pepper the review with dozens if not hundreds of comments: there are a lot of operations in here that work on (x,y) coordinate pairs that involve effectively duplicated code. DFHack has a 2d coordinate type `df::coord2d` that can be used to store a coordinate pair as a unit and operate on both components in parallel. Consider using `df::coord2d` in lieu of separate x and y variables wherever possible as this provides clearer indication of intention as well as reducing the risk of updating one but not the other resulting in inconsistent behavior.

I've made some other specific comments but this should not be considered a full review.

notliad:
> A general comment as this is a pervasive issue and I do not wish to pepper the review with dozens if not hundreds of comments: there are a lot of operations in here that work on (x,y) coordinate pairs that involve effectively duplicated code. DFHack has a 2d coordinate type `df::coord2d` that can be used to store a coordinate pair as a unit and operate on both components in parallel. Consider using `df::coord2d` in lieu of separate x and y variables wherever possible as this provides clearer indication of intention as well as reducing the risk of updating one but not the other resulting in inconsistent behavior.
> 
> I've made some other specific comments but this should not be considered a full review.

This will be a substantial refactor. I will look into it and the others comments as well.
I'm bit worried making lots of changes cause I can't reproduce (on both of my machines) the most common bug and need someone to test it for me, this can take a day or two.

notliad:
Camera feature was not working properly and introduced a mouse click bug so I opted to remove it.

Edit: forgot to tag @ab9rf

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -33,339 +33,83 @@ using namespace DFHack;
...
+    df::coord2d viewport_dimensions;
+    df::coord2d clip_min;
+    df::coord2d clip_max;
+    int32_t zoom;
+    df::coord2d origin;
+    df::coord2d screen_dimensions;
+
+    bool operator==(const visual_contextst &) const = default;
+};
+visual_contextst previous_visual_context{};
```
ab9rf, line 72: blank line here please

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -33,339 +33,83 @@ using namespace DFHack;
...
-        transient_x *= k;
-        transient_y *= k;
-        if (std::abs(transient_x) < 0.5 && std::abs(transient_y) < 0.5) {
-            transient_x = 0.0;
-            transient_y = 0.0;
-        }
-    }
-}
-
 constexpr uint32_t fire_bits = 0x70000000U;
```
ab9rf: this is a magic number. please construct this out of published constants

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -33,339 +33,83 @@ using namespace DFHack;
...
-    has_view_signature = true;
+    previous_visual_context = context;
+    has_visual_context = true;
 
-    // On a pure pan the reset signature is unchanged, but last frame's blackout coverage is in the
-    // old viewport frame, so discard it (the engine repaints the whole scrolled viewport anyway).
+    // On a pure pan the reset signature is unchanged, but last frame's blackout
+    // coverage is in the old viewport frame, so discard it (the engine repaints
+    // the whole scrolled viewport anyway).
     const int32_t pan_x = window_x ? *window_x : 0;
```
ab9rf: shouldn't this be `const coord2d pan` with an appropriate initializer?

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -33,339 +33,83 @@ using namespace DFHack;
...
     const int32_t pan_y = window_y ? *window_y : 0;
-    if (!has_pan_context || previous_pan_x != pan_x || previous_pan_y != pan_y)
+    if (!has_pan_context || previous_pan.x != pan_x || previous_pan.y != pan_y)
         previous_coverage.clear();
-    previous_pan_x = pan_x;
-    previous_pan_y = pan_y;
+    previous_pan.x = pan_x;
```
ab9rf: `previous_pan = pan` with `pan` initialized as mentioned above

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -485,20 +227,16 @@ void with_zeroed_values(const Callback &callback, T &value, Values &...values) {
 
 struct render_proxyst {
     viewport_visual_layer layer;
-    float source_x;
-    float source_y;
-    int32_t target_x;
-    int32_t target_y;
+    point2dst<float> source;
```
ab9rf: toady likes the silly nonsense of putting `st` after structure names, but imo it detracts from readability here. consider renaming the `point2d` template class without toady's `st` suffix on types, which DFHack has no love for and is absolutely not required for DFHack-defined types

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -686,11 +427,11 @@ void render_copy_maybe_mirrored(SDL_Renderer *renderer, SDL_Texture *texture,
 
 void draw_proxy(df::renderer_2d_base *renderer, const render_proxyst &proxy) {
     const int32_t zoom = renderer->viewport_zoom_factor;
-    const int32_t target_x = tile_pixel(proxy.target_x, renderer->origin_x, zoom);
-    const int32_t target_y = tile_pixel(proxy.target_y, renderer->origin_y, zoom);
+    const int32_t target_x = tile_pixel(proxy.target.x, renderer->origin_x, zoom);
```
ab9rf: why not a `coord2d` here

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -686,11 +427,11 @@ void render_copy_maybe_mirrored(SDL_Renderer *renderer, SDL_Texture *texture,
...
+    const int32_t target_y = tile_pixel(proxy.target.y, renderer->origin_y, zoom);
     const float tile_size = float(zoom == 128 ? 32 : std::max(1, zoom * 32 / 128));
-    const float source_x = target_x + (proxy.source_x - proxy.target_x) * tile_size;
-    const float source_y = target_y + (proxy.source_y - proxy.target_y) * tile_size;
+    const float source_x = target_x + (proxy.source.x - proxy.target.x) * tile_size;
```
ab9rf: we can extend `coord2d` as a template that provides `coord2d` methods for base types other than `int16_t`

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -720,20 +460,20 @@ std::vector<render_proxyst> collect_proxies(df::renderer_2d_base *renderer,
                 if (!movement.active)
                     continue;
                 const int32_t inherited_source_x =
```
ab9rf, line 477: this is still repetitive

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -720,20 +460,20 @@ std::vector<render_proxyst> collect_proxies(df::renderer_2d_base *renderer,
...
                     bool anchored = false;
                     for (const render_proxyst &anchor : proxies) {
                         if (anchor.layer == viewport_visual_layer::center &&
-                            std::abs(anchor.target_x - x) <= 1 &&
-                            std::abs(anchor.target_y - y) <= 1 &&
-                            anchor.source_x - anchor.target_x == movement.source_x - x &&
-                            anchor.source_y - anchor.target_y == movement.source_y - y &&
+                            std::abs(anchor.target.x - x) <= 1 &&
+                            std::abs(anchor.target.y - y) <= 1 &&
+                            anchor.source.x - anchor.target.x == movement.source.x - x &&
```
ab9rf: why not `anchor.source - anchor.target == movement.source - coord2d{x,y}` here?

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -782,32 +522,31 @@ std::vector<render_proxyst> collect_proxies(df::renderer_2d_base *renderer,
                 const bool mirror_eligible =
                     flip_enabled &&
                     (group == visual_render_groupst::main || group == visual_render_groupst::upper);
-                // Facing is read from the anchor tile so every fragment of one creature agrees.
+                // Facing is read from the anchor tile so every fragment of one creature
+                // agrees.
                 const bool mirrored =
                     mirror_eligible && animation_manager.get_facing(
```
ab9rf: second argument to `get_facing` should be a `df::coord`, don't split it into x/y

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -782,32 +522,31 @@ std::vector<render_proxyst> collect_proxies(df::renderer_2d_base *renderer,
...
                                         texpos,
                                         movement.progress,
                                         nullptr,
                                         mirrored,
                                         mirror_shift,
                                         {}};
                 bool blocked = false;
-                for (int32_t coverage_x = int32_t(std::floor(std::min(proxy.source_x, float(x))));
-                     coverage_x <= int32_t(std::ceil(std::max(proxy.source_x, float(x))));
+                for (int32_t coverage_x = int32_t(std::floor(std::min(proxy.source.x, float(x))));
```
ab9rf: this is a bit too magical for me, and that computation should be broken ouot before the loop so that it only has to do the `CVTSI2SS` once instead of on each iteration - for loops are prone to evaluate _everything_ in the _condition_ every iteration, even if it's apparently constant

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -853,24 +592,23 @@ std::vector<render_proxyst> collect_proxies(df::renderer_2d_base *renderer,
...
 
-    // A creature that has stopped still needs its mirrored sprite painted each frame.
-    // Otherwise the engine repaints it natively and the two orientations alternate between steps.
-    // A fragment's tile is its anchor minus the layer's centre offset, inverting the moving path.
+    // A creature that has stopped still needs its mirrored sprite painted each
+    // frame. Otherwise the engine repaints it natively and the two orientations
+    // alternate between steps. A fragment's tile is its anchor minus the layer's
+    // centre offset, inverting the moving path.
     if (flip_enabled) {
         for (int32_t anchor_x = 0; anchor_x < vp->dim_x; ++anchor_x) {
```
ab9rf, line 619: this argues for having an iterator for `coord2d` regions the same way we have for `coord` cuboids.

in any case i'd definitely add a `auto anchor = coord2d{anchor_x, anchor_y}` in here until we get a smart iterator for 2d regions

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -929,7 +667,7 @@ render_coveragest collect_coverage(const std::vector<render_proxyst> &proxies, i
     render_coveragest coverage;
     for (const render_proxyst &proxy : proxies) {
         coverage.all.insert(proxy.coverage.begin(), proxy.coverage.end());
-        coverage.selected[proxy.target_x * dim_y + proxy.target_y] |= visual_layer_bit(proxy.layer);
+        coverage.selected[proxy.target.x * dim_y + proxy.target.y] |= visual_layer_bit(proxy.layer);
```
ab9rf, line 691: this should be marked with `// TODO: use std::mdspan when C++23 available`

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -1000,14 +738,16 @@ void draw_viewport_interpolation_stages(df::renderer_2d_base *renderer,
...
         if (index > 0)
             redraw_viewport_tiles(renderer, viewports[index], coverage);
         const viewport_renderst &viewport = viewports[index];
         draw_interpolation_stages(renderer, viewport.viewport, viewport.proxies, viewport.coverage);
-        // A viewport shades everything drawn beneath it, so this covers every staged tile.
-        // Restricting it to the tiles this viewport has sprites on would not deepen with distance.
+        // A viewport shades everything drawn beneath it, so this covers every
+        // staged tile. Restricting it to the tiles this viewport has sprites on
+        // would not deepen with distance.
         for (const auto &[x, y] : coverage) {
```
ab9rf: change `inside_clip` and `draw_interface_only` to use a `coord2d` and bypass the structured binding here

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -1062,43 +789,6 @@ void render_interpolated_world(df::renderer_2d_base *renderer) {
     const int32_t zoom = renderer->viewport_zoom_factor;
     const int32_t tile_size = zoom == 128 ? 32 : std::max(1, zoom * 32 / 128);
```
ab9rf: uncomfortable with magic numbers here

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -1169,78 +857,31 @@ void reset_state() {
     previous_coverage.clear();
     visual_context_revision = 0;
     previous_viewport = nullptr;
-    previous_view_signature = {};
-    has_view_signature = false;
-    previous_pan_x = 0;
-    previous_pan_y = 0;
+    previous_visual_context = {};
+    has_visual_context = false;
+    previous_pan.x = 0;
```
ab9rf: `previous_pan = {};`

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -71,11 +70,12 @@ int main() {
     assert(mirrored_tile_x(5, 5) == 5); // anchor reflects to itself
```
ab9rf: magic numbers

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -247,9 +251,10 @@ int main() {
...
-    // Facing is keyed by SCREEN tile, so a scroll moves the creatures out from under it.
-    // The two outcomes differ fundamentally: a landed shift has a known delta and is translated.
-    // An abandoned shift never identifies a delta, so the grid can only be dropped.
+    // Facing is keyed by SCREEN tile, so a scroll moves the creatures out from
+    // under it. The two outcomes differ fundamentally: a landed shift has a known
+    // delta and is translated. An abandoned shift never identifies a delta, so
+    // the grid can only be dropped.
     {
         constexpr int32_t pan_dim = 4;
         int32_t pan_empty[pan_dim * pan_dim] = {};
```
ab9rf: consider a c++ `std::array` instead of a C array here

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -247,9 +251,10 @@ int main() {
...
 
-    // Facing is keyed by SCREEN tile, so a scroll moves the creatures out from under it.
-    // The two outcomes differ fundamentally: a landed shift has a known delta and is translated.
-    // An abandoned shift never identifies a delta, so the grid can only be dropped.
+    // Facing is keyed by SCREEN tile, so a scroll moves the creatures out from
+    // under it. The two outcomes differ fundamentally: a landed shift has a known
+    // delta and is translated. An abandoned shift never identifies a delta, so
+    // the grid can only be dropped.
     {
         constexpr int32_t pan_dim = 4;
```
ab9rf: 4 seems like a magic number here, especially in light of below

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -277,22 +282,24 @@ int main() {
             assert(landed.has_mirrored_facing(pan_viewport));
 
             // Frame A: the scroll is announced but the buffers have not moved yet.
-            input.pan_x = 1;
+            input.pan.x = 1;
             set_layer(input, viewport_visual_layer::center, at_two, at_two);
             run_frame(landed, input, 1032);
```
ab9rf: 1032 - and indeed all the `run_frame` numbers - are magic numbers all through here

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -322,9 +329,10 @@ int main() {
...
 
-    // Regression: A and B move in the same frame, chained -- A's target tile is B's source tile.
-    // A's write must not corrupt B's read of its own pre-frame facing.
-    // A is sent east first, so its facing differs from the default B carries and a swap shows.
+    // Regression: A and B move in the same frame, chained -- A's target tile is
+    // B's source tile. A's write must not corrupt B's read of its own pre-frame
+    // facing. A is sent east first, so its facing differs from the default B
+    // carries and a swap shows.
     {
         constexpr int32_t chase_dim = 5;
```
ab9rf, line 392: another magic number

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -339,8 +347,8 @@ int main() {
         frame_b[2 * chase_dim + 1] = 77; // A steps east: (1,1) -> (2,1), picks up an east facing
```
ab9rf: magic number

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -356,14 +364,15 @@ int main() {
...
 
-    // Regression: a vacated source tile is reoccupied the same frame by an UNTRACKED creature.
-    // Nothing targets that tile, so no target write clears it, and it is not empty either.
-    // Only an explicit, order-independent source clear restores the default there.
+    // Regression: a vacated source tile is reoccupied the same frame by an
+    // UNTRACKED creature. Nothing targets that tile, so no target write clears
+    // it, and it is not empty either. Only an explicit, order-independent source
+    // clear restores the default there.
     {
         constexpr int32_t gap_dim = 5;
```
ab9rf, line 434: another magic number

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -408,14 +420,14 @@ int main() {
...
+    assert(visual_layer_descriptor(viewport_visual_layer::right).anchor_offset.x == -1);
+    assert(visual_layer_descriptor(viewport_visual_layer::left).anchor_offset.x == 1);
+    assert(visual_layer_descriptor(viewport_visual_layer::upright).anchor_offset.x == -1 &&
+           visual_layer_descriptor(viewport_visual_layer::upright).anchor_offset.y == 1);
+    assert(visual_layer_descriptor(viewport_visual_layer::up).anchor_offset.x == 0 &&
+           visual_layer_descriptor(viewport_visual_layer::up).anchor_offset.y == 1);
+    assert(visual_layer_descriptor(viewport_visual_layer::upleft).anchor_offset.x == 1 &&
+           visual_layer_descriptor(viewport_visual_layer::upleft).anchor_offset.y == 1);
 
     std::array<int32_t, 9> empty{};
```
ab9rf: 9 is a magic number here, and this should probably be a proper compound instead of a bare array

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -504,23 +517,25 @@ int main() {
...
     assert(!context.get_movement(viewport, viewport_visual_layer::center, 1, 1).active);
 
-    // Camera-pan handling. window_x/window_y change at input time but the buffers shift on a later
-    // render frame, so the manager must (a) NOT create movements from the buffer shift itself (the
-    // floating-sprite bug), and (b) translate in-flight movements on the frame the shift lands.
+    // Camera-pan handling. window_x/window_y change at input time but the buffers
+    // shift on a later render frame, so the manager must (a) NOT create movements
+    // from the buffer shift itself (the floating-sprite bug), and (b) translate
+    // in-flight movements on the frame the shift lands.
     std::array<int32_t, 9> pan_current{};
```
ab9rf: `std::array<int32_t, 9>` absolutely needs to be a proper compound here

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -504,23 +517,25 @@ int main() {
...
     std::array<int32_t, 9> pan_previous{};
     std::array<int32_t, 9> pan_empty{};
     auto pan_input = make_input(viewport, 3, pan_empty.data());
     set_layer(pan_input, viewport_visual_layer::center, pan_current.data(), pan_previous.data());
```
ab9rf: why are we creating a `std::array` and then stripping it down to its data pointer, in code we control? just pass the array - this is C++, not C

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -504,23 +517,25 @@ int main() {
...
+    // FLOAT REGRESSION: a stationary creature, pan announced at frame A, buffers
+    // shift at frame B. Frame B's buffers look exactly like a real move
+    // ((1,1)->(0,1) with a unique source) — the manager must recognize it as the
+    // pending pan and create NO movement.
     visual_animation_managerst floaty;
     pan_previous[1 * 3 + 1] = 42;
     pan_current[1 * 3 + 1] = 42;
     run_frame(floaty, pan_input, 4990);
-    pan_input.pan_x = 1; // frame A: window scrolled, buffers unchanged
+    pan_input.pan.x = 1; // frame A: window scrolled, buffers unchanged
```
ab9rf: why is this assigning x but not y? suspicious to change one element of a coordinate but not the other, is a variable being used for a covert purpose here?

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -529,57 +544,59 @@ int main() {
...
+    // FOLLOW: an in-flight movement survives the announce frame untouched and is
+    // translated on the frame the buffers shift, so the sprite tracks the
+    // scrolled world.
     visual_animation_managerst panner;
     pan_current.fill(0);
     pan_previous.fill(0);
-    pan_input.pan_x = 0;
+    pan_input.pan.x = 0;
     run_frame(panner, pan_input, 5990);
     pan_previous[0 * 3 + 1] = 42; // creature steps (0,1) -> (1,1)
```
ab9rf: 42 is a magic number

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -590,7 +607,8 @@ int main() {
     run_frame(reset_on_zoom, pan_input, 8020);
     assert(!reset_on_zoom.get_movement(viewport, viewport_visual_layer::center, 1, 1).active);
 
-    // Status fragments inherit nearby center motion even while their texture flashes.
+    // Status fragments inherit nearby center motion even while their texture
+    // flashes.
     std::array<int32_t, 9> status_current{};
```
ab9rf: same as previous comment about `std::array<int32_t, 9>`

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -651,9 +670,10 @@ int main() {
...
     assert(!status.inherited);
-    assert(status.source_x == 1 && status.source_y == 0);
+    assert(status.source.x == 1 && status.source.y == 0);
 
-    // Wheelbarrows use the item layer and the same independent adjacent-movement detection.
+    // Wheelbarrows use the item layer and the same independent adjacent-movement
+    // detection.
     current.fill(0);
     previous.fill(0);
     input.current.fill(empty.data());
```
ab9rf: what even is this doing

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -689,6 +710,6 @@ int main() {
     current[2 * 3 + 1] = 80;
```
ab9rf: way way way way too many magic numbers

plugins/smooth-movement/visual_animation.h
```diff
@@ -10,6 +10,13 @@
 #include <cstdlib>
 #include <vector>
 
+#include "DataDefs.h"
+
+template <typename T> struct point2dst {
```
ab9rf: calling this `point2dst` is a nice homage to toady's weird behavior, but it's confusing and not helpful. also, this loses all the overloads that are on `df::coord2d`, which are highly useful

i can add a `coord2d` utility template to dfhack and define `df::coord2d` as a specialization of that template if it would be helpful

plugins/smooth-movement/visual_animation.h
```diff
@@ -103,18 +144,17 @@ constexpr bool visual_layer_matches(viewport_visual_layer layer, int32_t current
 
 struct viewport_visual_animation_inputst {
     const void *viewport = nullptr;
```
ab9rf: is there no better type that can be used here?

plugins/smooth-movement/visual_animation.h
```diff
@@ -298,10 +344,9 @@ class visual_animation_managerst {
         return double(matches) / double(considered);
     }
 
-    static std::array<int32_t, 2> shared_movement_delta(const int32_t *current,
-                                                        const int32_t *previous, int32_t dim_x,
-                                                        int32_t dim_y) {
-        std::array<int32_t, 2> best{};
+    static df::coord2d shared_movement_delta(const int32_t *current, const int32_t *previous,
```
ab9rf: not liking the use of decayed arrays here, would probably prefer a `std::span` or a `std::array` by reference

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
-        // Only a replaced view leaves a previous buffer belonging somewhere else; a first
-        // sighting does not.
+        // Only a replaced view leaves a previous buffer belonging somewhere else; a
+        // first sighting does not.
         const bool view_switched =
-            state.has_context && (state.context_revision != input.context_revision ||
-                                  state.dim_x != input.dim_x || state.dim_y != input.dim_y);
+            state.has_context &&
+            (state.context_revision != input.context_revision ||
+             state.dimensions.x != input.dimensions.x || state.dimensions.y != input.dimensions.y);
```
ab9rf: `state.dimensions == input.dimensions`

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
         const bool context_changed = !state.has_context || view_switched;
-        // The scroll delta is queued here as a hint; the buffers are hypothesis-tested each
-        // frame to find where it lands. Detection stays suppressed until then: a shifted
-        // buffer makes every panned creature look like a real move.
-        if (state.has_pan && (state.pan_x != input.pan_x || state.pan_y != input.pan_y)) {
+        // The scroll delta is queued here as a hint; the buffers are
+        // hypothesis-tested each frame to find where it lands. Detection stays
+        // suppressed until then: a shifted buffer makes every panned creature look
+        // like a real move.
+        if (state.has_pan && (state.pan.x != input.pan.x || state.pan.y != input.pan.y)) {
```
ab9rf: `state.pan == input.pan`

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
             if (state.pending.size() >= max_pending_shifts) {
-                // Same give-up as the other two sites: the owed shifts are unknowable now.
+                // Same give-up as the other two sites: the owed shifts are unknowable
+                // now.
                 abandon_pending(state);
                 reset_facing(state);
             }
-            state.pending.push_back({input.pan_x - state.pan_x, input.pan_y - state.pan_y});
+            state.pending.emplace_back(input.pan.x - state.pan.x, input.pan.y - state.pan.y);
```
ab9rf: should probably be `emplace_back(input.pan - state.pan)` but might need to be `emplace` instead, in any case use the operator overloads, that's what they're for

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
                                 int8_t(native_sprite_facing));
             state.has_mirrored = false;
         }
-        // This hook runs per frame; the viewport is recomputed only when it changes, and while
-        // paused hardly at all. Re-reading a landed scroll steps every sprite by a tile.
+        // This hook runs per frame; the viewport is recomputed only when it
+        // changes, and while paused hardly at all. Re-reading a landed scroll steps
+        // every sprite by a tile.
         const uint64_t signature = compute_buffer_signature(input);
         const bool buffers_advanced =
```
ab9rf, line 459: this to me advertise that `buffer_signature` should be a `std::optional`

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
             bool any_data = false;
-            std::array<int32_t, 2> shift{};
+            df::coord2d shift(0, 0);
             for (size_t count = 1; count <= state.pending.size(); ++count) {
-                shift[0] += state.pending[count - 1][0];
-                shift[1] += state.pending[count - 1][1];
-                // A prefix netting to zero is indistinguishable from "nothing landed yet".
-                // Accepting it would retire shifts the buffers have still to apply.
-                if (shift[0] == 0 && shift[1] == 0)
+                shift.x += state.pending[count - 1].x;
```
ab9rf: use coord2d operator overload here

plugins/smooth-movement/visual_animation.h
```diff
@@ -378,85 +423,94 @@ class visual_animation_managerst {
...
+                shift.y += state.pending[count - 1].y;
+                // A prefix netting to zero is indistinguishable from "nothing landed
+                // yet". Accepting it would retire shifts the buffers have still to
+                // apply.
+                if (shift.x == 0 && shift.y == 0)
```
ab9rf: `if (shift == {0,0})`

ab9rf (changes requested):
note: this is not a complete review - i'll make another pass after these comments are addressed

ab9rf:
> Edit: forgot to tag @ab9rf

not necessary to tag me, i'm notified on all activity on dfhack repositories

plugins/smooth-movement/visual_animation.h
notliad: Yeah, that’d be useful if another plugin needs it too. For now I kept it local so we don’t grow the shared API for one use case. It has the same operators as df::coord2d now; we can promote it later if it gets reused.

ab9rf:
I'm not ignoring this PR; instead, reviewing this PR has illuminated for me a few places where our infrastructure is insufficient and so I am working on improving those. I should have some more comments in another day or two.

notliad:
> I'm not ignoring this PR; instead, reviewing this PR has illuminated for me a few places where our infrastructure is insufficient and so I am working on improving those. I should have some more comments in another day or two.

No worries, just let me know what I can do on my side.

ab9rf:
> No worries, just let me know what I can do on my side.

There are two things I'm working on that will impact this PR.

* I am working on making a generalized template for coordinate pairs and coordinate triples. This is now mostly done but the process has illuminated a good deal tech debt which I am trying to clean up as best I can. Once this is finished, I will be expecting use to use these templates in place of the copy you've included in this PR.
* Your code is using some constants that are known to us (they're in `g_src` which is published) but which we don't include in `df-structures` because, I guess, we didn't think anyone would need them. I'm going to be adding those to `df-structures` in the near future, and I will expect this PR to use them once they've been added.

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,981 @@
...
+// Map scroll (window_x/window_y) is tracked separately from the reset
+// signature: a pure pan is followed (movements are translated) instead of
+// triggering a full reset, so it must NOT bump the context revision. It only
+// invalidates the viewport-space blackout coverage from the prior frame.
+df::coord2d previous_pan = df::coord2d(0, 0);
+bool has_pan_context = false;
+bool flip_enabled = false;
+bool zlevel_enabled = false;
+
+// DF stores fire's animation frame in bits 28--31 of screentexpos_spatter_flag.
```
ab9rf: These appear to be incorrect. the encoding is actually 
frame1 = 001 (1)
frame2 = 010 (2)
frame3 = 100 (4)
frame4 = 011 (3)
in bits 28-30. Bit 31 is used for something else.

Why Toady did it this way, I have no idea, but this is what's in the headers so...

Also this type is being added to `df_structures` as `viewport_spatter_flag`; please use the codegen-provided constants here instead of hardcoding them - see DFHack/df-structures#892

notliad: Done

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,975 @@
...
+        render_copy_ex_f(renderer, texture, nullptr, &destination, 0.0, nullptr,
+                         SDL_FLIP_HORIZONTAL);
+        return;
+    }
+    render_copy_f(renderer, texture, nullptr, &destination);
+}
+
+void draw_proxy(df::renderer_2d_base *renderer, const render_proxyst &proxy) {
+    const int32_t zoom = renderer->viewport_zoom_factor;
+    const coord2dst<int32_t> target{
```
ab9rf: `coord2dst` should be replaced throughout with `DFHack::Coord2d`

plugins/smooth-movement/test_visual_animation_manager.cpp
```diff
@@ -0,0 +1,778 @@
+// SPDX-License-Identifier: Zlib
+
+#include <array>
```
ab9rf: move this below the NDEBUG logic

plugins/smooth-movement/visual_animation.h
```diff
@@ -0,0 +1,896 @@
...
+
+#include "DataDefs.h"
+#include "df/coord2d.h"
+
+namespace df {
+struct graphic_viewportst;
+}
+
+template <typename T>
+struct coord2dst {
```
ab9rf: as noted elsewhere, replace this template with `DFHack::Coord2d`

plugins/smooth-movement/visual_animation.h
```diff
@@ -0,0 +1,896 @@
...
+                                        source = candidate;
+                                        ++candidate_count;
+                                    }
+                                }
+                            }
+                        }
+                        if (candidate_count != 1)
+                            continue;
+
+                        claimed_sources[source] = 1;
```
ab9rf, line 693: consider using `Coord2d<float>` here

plugins/smooth-movement/visual_animation.h
```diff
@@ -0,0 +1,896 @@
...
+                        float visual_source_x = float(source / input.dimensions.y);
+                        float visual_source_y = float(source % input.dimensions.y);
+                        for (size_t i = 0; i < existing_movement_count; ++i) {
+                            const movementst &movement = state.movements[i];
+                            if (movement.layer != static_cast<viewport_visual_layer>(layer) ||
+                                movement.target.x != visual_source_x ||
+                                movement.target.y != visual_source_y)
+                                continue;
+                            const float progress = movement_progress(movement.start_time_ms);
+                            visual_source_x = movement.source.x +
```
ab9rf: this is a linear interpolaton for which `Coord2d` has a handy method

ab9rf (changes requested):

ab9rf:
As far as I'm concerned, this PR would be mergeable except that it fails CI.

ab9rf:
I don't know why the test is failing - you'll need to figure that one out. 

The doc build failure is because, in a docs-only build, `dfhack_plugin` does not create a build target, but your CMake scripting assumed that `dfhack_plugin` did, in fact, create a build target. The solution here is to either check to see if the build target was created before attempting to modify its include directories, either by bracketing that stanza with something like something like `if(TARGET smooth-movement)` ... `endif()`, or by modifying `Plugins.cmake` to provide support for an `INCLUDE_DIRECTORIES` argument to the `dfhack_plugin` macro that will conditionally set `target_include_directories` if and only if the conditions in the macro for creating the build target obtain.

notliad:
> I don't know why the test is failing - you'll need to figure that one out.
> 
> The doc build failure is because, in a docs-only build, `dfhack_plugin` does not create a build target, but your CMake scripting assumed that `dfhack_plugin` did, in fact, create a build target. The solution here is to either check to see if the build target was created before attempting to modify its include directories, either by bracketing that stanza with something like something like `if(TARGET smooth-movement)` ... `endif()`, or by modifying `Plugins.cmake` to provide support for an `INCLUDE_DIRECTORIES` argument to the `dfhack_plugin` macro that will conditionally set `target_include_directories` if and only if the conditions in the macro for creating the build target obtain.

Found: It was the new coord lib typos. Only linux build calls smooth-movement-test that's why windows was passing. I wasn't getting this error because along the way codex fixed the library without me realizing it.

ab9rf:
> because along the way codex fixed the library without me realizing it.

This is, indeed, one of the main problems I have with vibecoding.

notliad:
Just added a fix for multi-tile creatures.
They were being rendered on different passes. So they would break apart.

plugins/smooth-movement/smooth-movement.cpp
```diff
@@ -0,0 +1,978 @@
...
+DFHACK_PLUGIN_IS_ENABLED(is_enabled);
+
+REQUIRE_GLOBAL(gps);
+REQUIRE_GLOBAL(window_x);
+REQUIRE_GLOBAL(window_y);
+REQUIRE_GLOBAL(window_z);
+
+namespace {
+
+using tile_coveragest = std::set<df::coord2d>;
```
quietust: Remark: type names ending with `st` are a Toady-ism, and I don't know that we really want to replicate that in code that is unique to DFHack.

Granted, there might be other instances of that already in DFHack, but I'm of the opinion that we should remove those rather than add more.

notliad: Done, removed all 'st' from local types.

ab9rf (approved):

ab9rf:
I pulled this down locally and tested it and it seems to work as advertised. I personally would not care to use it, but freedom is about choices. :)

PR #5891, opened 2026-09-01
orders: add position editing overlay

Robuske:
## Summary

- Add the `orders.position` overlay for viewing and changing manager-order positions directly on the fort-wide Work Orders screen.
- Display one-based position fields beside visible orders and move the existing manager-order pointer when a destination is submitted.
- Add a bottom-panel toggle for showing and hiding the position fields.
- Keep editing state synchronized across scrolling and reordering, and coordinate focus with the existing orders search field.
- Show invalid input in an error dialog while keeping successful and no-op moves silent.
- Add focused tests, documentation, and a Future changelog entry.

This PR is stacked on the work-order list geometry refactor in #5900.

## Testing

- Ran the targeted `orders` tests in a loaded fortress:

  ```text
  test -d C:/Users/rodbu/Developer/Dwarf/dfhack/test -t orders
  ```

  Result: 14/14 tests passed, 78/78 checks passed.
- Manually tested showing and hiding positions, valid and invalid moves, no-op moves, cancellation, focus changes, scrolling, search interaction, and reordering while a field is active.

## Documentation

Includes `orders.position` overlay documentation and the required `docs/changelog.txt` entry.

## Demo

https://github.com/user-attachments/assets/8136443b-ccdb-4e0b-8406-ad5fb78396b9


chdoc:
Before I spend a lot of time reviewing the code in detail, let me first ask a few high-level questions and make a few high-level comments: 

1. What is the motivation for this change? I have never actually felt the need to reorder work orders. The only reason I can see to reorder work orders is to avoid low frequency work orders being starved by high frequency work orders. That is the purpose of `orders sort`.
2. Assuming I can be convinced that this has general utility, the described workflow of selecting a work order and then entering a number where it should be placed it does not sound appealing. If I wanted to sort orders, I would appreciate label buttons to move the selected order (we can select orders) one up, one down, to the top, or to the bottom. Has this been considered? 
3. I cannot see a small UI/UX improvement like this requiring a change of roughly 900 additional lines plus another roughly 400 lines of tests. This seems over-engineered and gives me strong AI code vibes. Please give me a description in your own words (that is without the use of AI) of the general idea behind the implementation and why you think that this cannot be achieved using a smaller, more focused change.
4. In fact, some of the changes look like you are refactoring the tool at the same time as you are changing features. Please keep refactoring pull requests and feature pull requests separate. Is just because it caught my eye: Replacing `self.subviews.filter` with `self.subviews.[SEARCH_FILTER_VIEW_ID]` is an absolute no go, because it reduces readability for no gain whatsoever.

Robuske:
@chdoc 
> 1. What is the motivation for this change? I have never actually felt the need to reorder work orders. The only reason I can see to reorder work orders is to avoid low frequency work orders being starved by high frequency work orders. That is the purpose of `orders sort`.

Really? Maybe there is something wrong with my understanding or how I play, but can't repeating orders starve each other?
If I already had a repeating order for making steel armor, and add a new one for steel swords, won't the steel armor have priority? If I want the swords to have priority, I will have to bump its order up one by one in the list

> 2. Assuming I can be convinced that this has general utility, the described workflow of selecting a work order and then entering a number where it should be placed it does not sound appealing. If I wanted to sort orders, I would appreciate label buttons to move the selected order (we can select orders) one up, one down, to the top, or to the bottom. Has this been considered?

I am a little confused by this question, the game already has an option for moving one up or down, what I added is a way to move a arbitrary number, including to the top or bottom if you want, you just use the 1 for the top and whatever the last order is for bottom. Maybe what I have implemented isn't very clear?

> 3. I cannot see a small UI/UX improvement like this requiring a change of roughly 900 additional lines plus another roughly 400 lines of tests. This seems over-engineered and gives me strong AI code vibes. Please give me a description in your own words (that is without the use of AI) of the general idea behind the implementation and why you think that this cannot be achieved using a smaller, more focused change.

Oh yeah, the "core" of this change is basically a few lines, removing an order from the array and moving it to another position, but:
1. It includes "cli" commands instead of just the overlay, it seemed to me that DFHack usually includes both
2. Giving a good UX to text fields is a pain in the ass, it deselects if you right click, press esc, click a random place, click another field, enter some shortcuts, etc...

To be clear I have used AI, I am an experienced iOS developer, but I only have a basic knowledge of C++ and none of Lua so it was basically vibe coded, that said I did check if the code "seemed reasonable" and tested the experience a lot, if it is overengineered it is more my own fault than the AI's (with exception of the tests, I admit that I didn't do much besides running them)

> 4. In fact, some of the changes look like you are refactoring the tool at the same time as you are changing features. Please keep refactoring pull requests and feature pull requests separate. Is just because it caught my eye: Replacing `self.subviews.filter` with `self.subviews.[SEARCH_FILTER_VIEW_ID]` is an absolute no go, because it reduces readability for no gain whatsoever.

One of the text fields UX issues I have run into is selecting/deselecting the correct field, by default multiple fields would stay selected. The search field is the only other text field in that screen, so instead of searching the subviews "manually" for controlling focus, it was easier to define an ID to access it.
The refactoring is just an extraction of what both overlays have in common, mostly the code for following the orders that the search highlight and this new position overlay use. I can move it to another PR, but this refactor doesn't really make sense without the position overlay introduction, so I didn't really see the reason to do that.

I am just explaining why I did the things I did, like I said, I am an absolute noob when it comes to Lua or DFHack itself, it is very likely that there are better ways to do parts of what I am doing (or maybe there is something that you just prefer done another way), just let me know and I can fix it.

chdoc:
> Really? Maybe there is something wrong with my understanding or how I play, but can't repeating orders starve each other? If I already had a repeating order for making steel armor, and add a new one for steel swords, won't the steel armor have priority? If I want the swords to have priority, I will have to bump its order up one by one in the list

My understanding is that every work order will only generate at most one job per applicable workshop. That is, unless you are very short on steel or have an extremely large number of forges, one order starving another is actually pretty rare. 

> I am a little confused by this question, the game already has an option for moving one up or down, what I added is a way to move a arbitrary number, including to the top or bottom if you want, you just use the 1 for the top and whatever the last order is for bottom. Maybe what I have implemented isn't very clear?

Apologies, that question is merely an indication that in hundreds of hours of playing DwarfFortress I never felt the need to reorder work orders so I never used those buttons or knew they existed. Still, I would expect a button to toggle the overlay (with the position indicators, this screen feels cluttered to me), and buttons to move things to the top or the bottom. If this positioning can be computed reliably, then maybe putting them to the left or the right of the central column might make sense?

> Oh yeah, the "core" of this change is basically a few lines, removing an order from the array and moving it to another position, but:
> 
>     1. It includes "cli" commands instead of just the overlay, it seemed to me that DFHack usually includes both
> 
>     2. Giving a good UX to text fields is a pain in the ass, it deselects if you right click, press esc, click a random place, click another field, enter some shortcuts, etc...

One of the main reasons of CLI-tools existing in addition to GUI-tools is the CLI-tool predating the GUI-tool. Another is a command you might reasonably want to run automatically (e.g. scripting). The third is performance considerations. I'm not sure any of that applies here. I mean, you basically have to have the window open to get the numbers. I don't see any point in listing the orders by number on the command line and then fiddling with them there. That sounds extremely cumbersome, I'm not sure why anyone would want to do this. Don't add CLI-tools just for the sake of adding CLI-tools.
 
> To be clear I have used AI, I am an experienced iOS developer, but I only have a basic knowledge of C++ and none of Lua so it was basically vibe coded, that said I did check if the code "seemed reasonable" and tested the experience a lot, if it is overengineered it is more my own fault than the AI's (with exception of the tests, I admit that I didn't do much besides running them)

I don't have anything against using AI in general. However, I am very cautious about even starting to do an in-depth review of code that was written by people who, by their own admission, do not really know the language the code is written in. It is a requirement to be willing and able to engage in a meaningful discussion about the design of the change. You have engaged in a discussion instead of adopting a "take it or leave it" attitude. Not everyone does. So we are good here.

> One of the text fields UX issues I have run into is selecting/deselecting the correct field, by default multiple fields would stay selected. The search field is the only other text field in that screen, so instead of searching the subviews "manually" for controlling focus, it was easier to define an ID to access it. The refactoring is just an extraction of what both overlays have in common, mostly the code for following the orders that the search highlight and this new position overlay use. I can move it to another PR, but this refactor doesn't really make sense without the position overlay introduction, so I didn't really see the reason to do that.

Please try to stay away from refactoring as much as possible, and please try to minimize the amount of lines changed (every line changed is a line I need to review). In particular because this is your first contribution.

For this manual tinkering with order positions, I think you can safely drop the command-line part. I don't see any reason for this to touch the C++ part at all.

Robuske:
@chdoc 
> That is, unless you are very short on steel...

Yeah, that was my recurrent issue 😢 

>  Still, I would expect a button to toggle the overlay (with the position indicators, this screen feels cluttered to me), and buttons to move things to the top or the bottom.

I can add a button to toggle the overlay, maybe as a header for whichever column they end up in, or part of the bottom overlay that orders already adds.

> If this positioning can be computed reliably, then maybe putting them to the left or the right of the central column might make sense?

I will push back on this a bit due to the relation to the other UI elements. I also don't love how it looks, but moving it to the left of the central column would make it look associated with the amount controls, and to the right it would look associated with the order details or workshops.
That said, I don't think that it looks horrible in the right, I will do some tests for other positions.

> Don't add CLI-tools just for the sake of adding CLI-tools.

Fair enough.

> However, I am very cautious about even starting to do an in-depth review of code that was written by people who, by their own admission, do not really know the language the code is written in. It is a requirement to be willing and able to engage in a meaningful discussion about the design of the change.

Also fair, I did try to understand the basics while reviewing it myself, but I certainly wouldn't argue that I really understand Lua at this point, that said, I think that at this point most of the necessary changes should me more about DFHack itself than Lua (like what we already are discussing).
 
> Please try to stay away from refactoring as much as possible, and please try to minimize the amount of lines changed (every line changed is a line I need to review). In particular because this is your first contribution.
> 
> For this manual tinkering with order positions, I think you can safely drop the command-line part. I don't see any reason for this to touch the C++ part at all.

All right, I will rewrite it to be more focused even if it leads to some duplication for now, I will also review the tests to make sure they are really important at this point.

ab9rf:
I'm going to concur with @chdoc on the importance of not mixing a refactor with a feature addition. If a feature addition needs a refactor in order to work, _propose the refactor first_, and then propose the feature addition contingent on the refactor. Either way, I'd like to see them as separate changes, since that makes reviewing both changes, and understanding their consequences, much easier. I'd rather review two smaller PRs than one massive one.

You might have noticed that my recent change to use templates for `df::coord` and `df::coord2d` ultimately required 14 pull requests to implement. while a few of them were required to clean up something i had missed, most of them were simply because i wanted to leave behind a clear, documented history of the series of changes being made, in well-defined units, instead of a monolith that addressed too many thing all in a piece. don't be afraid to break a change into multiple PRs, especially in a situation where a change requires or benefits from a refactor or requires a collateral feature or adjustment in order to be implemented

Robuske:
@ab9rf @chdoc I have removed a lot of stuff, and moved the refactor to this [PR](https://github.com/DFHack/dfhack/pull/5900) (basically just an extraction)
This branch is stacked on that other one, so some stuff will appear here as well until it is merged.

I have also updated the description for this PR accordingly.

PR #5893, opened 2026-09-03
v0.47.05: Fix errors from newer versions of CMake and GCC

cizra:
Bump the minimum CMake version, fixes build.
Fix GCC warnings-as-errors, fixes build.
Bump version number.


ab9rf:
We're not likely to entertain changes to legacy versions since we can't validate them; our CI isn't set up to build or test for this version of DF. As such, I don't imagine we'll be entertaining this PR.

cizra:
@ab9rf I understand, and I won't press the issue. Feel free to close this PR, if it has no way forward.

My motivation for this change was that v0.47.05 is the last version of DF preserving ncurses UI and familiar keybindings. The newer versions introduce larger UI changes, which may not be preferable for some people. Like me.

This branch, as it currently is, does not build - making it mostly useless. Does it really matter if CI is not set up here, or if we merge a commit breaking it further?

I have built it myself, and I confirm that it runs fine (tested by launching DF through `dfhack`, and sending `dfhack-run quicksave`). If you tell me which other tests to run, I can do that.

PR #5900, opened 2026-09-11
orders: extract work order list geometry

Robuske:
## Summary

- Extract the Work Orders list geometry calculations into `plugins.orders.work_order_list` so Lua overlays can share the same row positions and visible range.
- Update the existing orders search overlay to use the extracted module without changing its behavior.
- Add focused coverage for list placement, viewport sizing, and end-of-list clamping.
- Add the required changelog and contributor entries.

## Testing

- Ran the targeted `orders` tests in a loaded fortress:

  ```text
  test -d C:/Users/rodbu/Developer/Dwarf/dfhack/test -t orders
  ```

  Result on this branch: 10/10 tests passed, 48/48 checks passed.
- The stacked position-overlay branch also passed 14/14 tests and 78/78 checks.


chdoc:
While we are far from consistent about this, I would prefer if new libraries were using camel case instead of snake case. This would bring `local_module.someThing` more in line with `dfhack.module.someThing`. The global API consistently uses camel case. In particular, given that most of the local imports use single words or acronyms as names.

Robuske:
@chdoc PR has been updated to use camelCase

PR #5904, opened 2026-09-15
autodump: add 'undestroy' option; fix container/unit-item destruction

### LINKED ISSUE #5769: Bug: `autodump destroy` on containers leaves items in limbo
Using `autodump destroy` on a container but not its contents leaves items in limbo. I also got a crash upon saving one time when doing it to an equipped quiver.

Undoing destruction by running the command again before unpausing isn't working for whatever reason. (It still works for the gui tool.)

### LINKED ISSUE #5896: `autodump destroy` doesn't revert pending item destruction
The docs for `autodump destroy` claim it can cancel pending item destruction when used before unpausing. As far as I can tell, this has never been true? Only `autodump-destroy-item` has this feature.

The functionality could be added, though it should be a separate option to prevent constantly reverting destruction while unpaused.

Alistair-Afton:
## Summary

Adds an `autodump undestroy` command that reverts items pending destruction, the same undo that `gui/autodump` already had (fixes #5896).

While in there I fixed two related issues with `autodump destroy` (fixes #5769):

- It now recursively marks the contents of destroyed containers instead of leaving them in limbo
- It skips items held in unit inventories — destroying an equipped container (e.g. a quiver) could previously strand its contents and crash

Pending-destruction flags are also restored from saved state on undo rather than blindly cleared, so items that were already marked before the destroy don't get unmarked.

## Testing

- Live fort: container + contents both get marked on destroy; `autodump undestroy` restores both
- Equipped pick correctly skipped
- `ci/lint.py` clean; plugin builds cleanly


plugins/autodump.cpp
```diff
@@ -105,6 +141,27 @@ static command_result autodump_main(color_ostream &out, vector<string> &paramete
...
 
+    reset_pending_destroy();
+
+    if (undestroy)
+    {   // Revert items marked for destruction while the game is still paused.
+        int restored_total = 0;
+        for (auto &entry : pending_destroy)
+        {
+            if (auto itm = df::item::find(entry.first))
+            {   // Item could have already been garbage collected
```
quietust, line 152: If another plugin forcibly deleted the item while the game was still paused, `itm` could be NULL and result in a crash

Alistair-Afton: Isn't that case already covered? `itm` is only dereferenced inside `if (auto itm = df::item::find(entry.first))`, so if the item was deleted while paused, `find` returns NULL and the entry is skipped. Unless you mean a case where `find` can still return a pointer to a dead item?

plugins/autodump.cpp
```diff
@@ -215,18 +277,14 @@ command_result df_autodump_destroy_item(color_ostream &out, vector<string> &para
...
 
     // Allow undoing the destroy.
-    if (world->frame_counter != last_frame)
-    {
-        last_frame = world->frame_counter;
-        pending_destroy.clear();
-    }
+    reset_pending_destroy();
 
     if (pending_destroy.count(item->id))
```
ab9rf, line 282: use `contains` here - you don't actually care what the count is and there is no reason to force the library to count them when all you care about is whether the key is present

plugins/autodump.cpp
```diff
@@ -72,9 +72,43 @@ static constexpr uint32_t proj_flags = (
...
+        last_frame = world->frame_counter;
+        pending_destroy.clear();
+    }
+}
+
+// mark an item for destruction, along with anything it contains, so that
+// contained items are not left pointing at a deleted container
+static void mark_for_destroy(df::item *itm)
+{
+    if (!pending_destroy.count(itm->id))
```
ab9rf, line 94: use `contains` here as well

plugins/autodump.cpp
ab9rf: the comment here is confusing - it implies that this block handles the case where the item _has_ been deleted, which is not actually what is going on

quietust: Disregard - I misread the code, not noticing that this was an assignment inside an `if` statement (which, to be honest, is probably something to be avoided _specifically_ because it leads to confusion like this).

plugins/autodump.cpp
Alistair-Afton: Done.

plugins/autodump.cpp
Alistair-Afton: Done.

plugins/autodump.cpp
Alistair-Afton: @ab9rf Reworded it, hopefully clearer now.

Alistair-Afton: @quietust Fair point, I split the declaration out of the if.

ab9rf (approved):

PR #5905, opened 2026-09-15
stocks: fix blank item list after collapsing while scrolled down

### LINKED ISSUE #5772: Bug: `stocks.overlay` doesn't enforce scroll bounds when collapsing categories
Choosing to collapse all items on the stocks screen while scrolled downwards can leave the view with no items on screen and no way to scroll back up. The user is forced to switch stock categories and back again to fix it, re-expanding the item list. The tool should enforce the proper scroll bounds when collapsing.

Alistair-Afton:
## Summary

`collapse_all` shrank `i_height` but left `scroll_position_item` stale, so collapsing a list while scrolled down left the viewport past the end of the list and rendered a blank pane (fixes #5772).

Applied the same scroll clamp to `remove_empty`, which had the identical hazard on the type list, and fixed a leaked global (`num_items` in `expand_all` was missing `local`).

## Testing

- Live fort: collapse resets scroll to 0; flags continue toggling correctly
- `ci/lint.py` clean


SilasD (changes requested):
I am very uncomfortable making `collapse_all`, `expand_all`, and `remove_empty` global.

these functions directly manipulate the UI; the UI is generally fragile and tends to leave dangling pointers when a viewsheet is closed.

making these functions global means that any other script can call them at any time.  if the stocks viewsheet is not open, I think there is a significant risk of crashing the game.

to emphasize: these functions **should (must) *not* be called** if the stocks interface is not open.

I would either like these changed back to `local` or have a safety-test added, something like
`if stocks.open == false then return; end`
at the start of each function.

or both.  both would be good.

---

Other than that issue, the code generally looks fine, though I didn't test it.

Alistair-Afton:
Done both: the functions are `local` again, and each now starts with `if not stocks.open then return end` so a stray activation as the page closes can't touch the interface state.

Also added `test/plugins/stocks.lua`, which drives the overlay hotkeys through the widget's real `onInput` path in a running game and verifies:
- collapse resets `scroll_position_item`
- expand restores all sections
- remove-empties keeps `scroll_position_type` in bounds
- all three hotkeys are no-ops while the stocks page is closed (exercises the `stocks.open` guard directly)

Verified in-game: 4/4 tests pass, full suite green (271/271).

SilasD:
Looks good now, I'm happy.

Reminder to ab9rf: PR #5920 superseeds this.

SilasD (approved):
Looks good.

someday I'll figure out the differences between commenting and reviewing; today is not that day.

Alistair-Afton:
Closing: this fix already merged via #5920, which shared the same commit chain (`99288ddbf`). After rebasing on current develop the remaining diff is empty.

PR #5907, opened 2026-09-15
ci/test.lua: reinstate loading of a specified save dir

Alistair-Afton:
## Summary

Reinstates the `load-save` call in `ensure_fortress` so the documented `-s`/`--save_dir` test option actually selects which save to load, instead of always loading the first save in the list. The `load_first_save`/`click_top_title_button` helpers it replaced are no longer needed.

Depends on DFHack/scripts#1619 (the `load-save` fix) merging and the scripts submodule pin updating. Until then the first `run_script('load-save', ...)` call hits the `unavailable` warning and qerrors, which would abort fortress-mode tests.

## Testing

- Locally: `dfhack.run_script('load-save', 'region1')` from the title screen loads the region1 fort
- CI here will exercise `ensure_fortress` itself once the scripts pin includes the fix


SilasD (approved):
Looks good.  The deleted code has been moved into https://github.com/DFHack/scripts/pull/1619 , with updates for current DF.

this should not be merged until https://github.com/DFHack/scripts/pull/1619 is merged.

PR #5908, opened 2026-09-15
hotkeys: don't query lua viewscreen selections on the render thread

### LINKED ISSUE #5835: autodump hotkeys can run without acquiring a coresuspender under some conditions
Hitting `Alt-K` while a designation is being selected _and_ while a DFHack overlay window is open but not selected (e.g. `gui/reveal`) will cause an assertion failure in `dfhack_lua_viewscreen::getSelected` for attempting to acquire the shared Lua state without holding the core suspender.

Replication process: Open `gui/reveal`. Click outside the window to make it nonactive; do not close it. Press `x` to begin a "remove designation" operation. Click somewhere to start designating a rectangle to remove designations. Hit `Ctrl-K` to trigger the hotkey for `autodump-destroy-item`. A production instance will hang (presumably due to deadlock); a developer build will assertion-fail on the assertion in `DFHack::Core::getInstance().getLuaState()` that requires `isSuspended()` to be true.

See [Discord](https://discord.com/channels/793331351645323264/873014631315148840/1519801647700902018) for more context.

Alistair-Afton:
## Summary

Fixes #5835.

When a keypress matches a bound hotkey, `Plugin::can_invoke_hotkey` runs the command's guard on DF's render thread before the command is queued. Guards that inspect the UI selection (`Gui::any_item_hotkey`, `workshop_job_hotkey`, etc.) call `getSelected*()` on the topmost viewscreen, and when that screen is a `dfhack_lua_viewscreen` the virtual enters Lua through `CoreSuspender`, which can never be acquired on the render thread. The result was an assertion failure in debug builds and a hang in production, e.g. with `gui/reveal` open but unfocused.

This change routes the selection query through a helper that checks the calling thread first:

- On the render thread, a focused lua screen owns the selection, so we report none.
- A defocused lua screen is transparent to the UI below, so the native checks still apply, which keeps hotkeys working against the underlying screen (e.g. during a designation with an unfocused overlay).
- Off the render thread (command handlers running under the suspender) and for C++ plugin screens, the `getSelected*()` virtual is called as before.

Also adds `Core::isRenderThread()` so the check has a clear name.

Guards that never inspect the selection (`anywhere_hotkey`, `default_hotkey`, focus-based guards) are unaffected, as is `Gui::getFocusStrings`/`matchFocusString` (`isFocused`/`getFocusString` are pure C++ on lua screens).

## Testing

- Built and installed locally (win64, DF 53.16).
- Reproduced the issue's setup in a live fort: opened `gui/reveal`, defocused it (focus resolves to `dwarfmode/...` through the screen), and pressed `Ctrl-K` (`autodump-destroy-item`). No assert, no hang; the key fell through to the screen as expected.
- Confirmed a test hotkey bound to `@dwarfmode` still dispatched through the defocused lua screen.
- Previously, the same sequence tripped the `CoreSuspenderBase` assertion (or hung, waiting on `CoreSuspendMutex` from the render thread).


library/modules/Gui.cpp
```diff
@@ -1065,6 +1066,33 @@ bool Gui::any_job_hotkey(df::viewscreen *top)
...
+// to the native UI checks.
+//
+// getSelected*() on a lua-backed screen enters Lua through CoreSuspender,
+// which asserts or deadlocks on the render thread, where hotkey guards run.
+// A focused lua screen owns the selection, so we must report none; a
+// defocused lua screen is transparent to the UI below, so the native checks
+// still apply.
+template<typename T>
+static std::optional<T *> dfscreen_selection(df::viewscreen *top,
+        T * (dfhack_viewscreen::*method)())
```
ab9rf: Is there a reason why this template is restricted to returning a pointer?

Alistair-Afton: @ab9rf No good reason. It only needed the pointer return so the focused case could yield nullptr as the no selection answer, but deducing R directly does that just as well. Generalized it.

ab9rf (approved):

PR #5910, opened 2026-09-15
Normalize out-of-range pen colors from Lua

### LINKED ISSUE #5865: std::array assertion in doSetTile_char for non-optimized builds

Alistair-Afton:
## Summary
- `COLOR_RESET` (-1) in a Lua pen table was stored raw into `Pen.fg`/`bg`/`tile_fg`/`tile_bg`, then reinterpreted as `uint8_t` (255) in `doSetTile_char` and used to index the 16-entry `uccolor` array. In an optimized build this is an out-of-bounds read; the launcher triggers it through `TextAreaContent` pens that use `bg=COLOR_RESET`.
- `decode_pen` now substitutes the Pen defaults (grey fg, black bg) for any out-of-range color before the `bold`-flag extraction, so `fg=COLOR_RESET` no longer implies `bold`.
- `doSetTile_char` applies the same defaults to pens from other sources before indexing `uccolor`, keeping the existing out-of-range warning as a last resort.

Fixes #5865

## Testing
- Built Release and exercised `gui/launcher` (the reported trigger) plus direct `gui.Painter` calls with `bg=COLOR_RESET`, `fg=200`, `tile_fg=-1`, `tile_bg=100`; no `out of range` warnings and no crash.


library/modules/Screen.cpp
```diff
@@ -179,17 +179,19 @@ static bool doSetTile_char(const Pen &pen, int x, int y, bool use_graphics)
         *flag &= flag_mask;
     }
 
-    uint8_t fg = pen.fg | (pen.bold << 3);
-    uint8_t bg = pen.bg;
+    // COLOR_RESET only has meaning on text consoles; use the Pen defaults
+    // (grey text on black) so out-of-range colors can't reach the buffers
+    uint8_t fg = (pen.fg < 0 ? COLOR_GREY : pen.fg) | (pen.bold << 3);
+    uint8_t bg = pen.bg < 0 ? COLOR_BLACK : pen.bg;
```
ab9rf: if we're going to do this, we should probably also ensure that neither `fg` nor `bg` exceeds the maximum color

Alistair-Afton: @ab9rf Good catch, done. Both ends are clamped to the Pen defaults now, in decode_pen and in doSetTile_char for pens that bypass it.

ab9rf (approved):

PR #5916, opened 2026-09-16
buildingplan: fix Unlink all skipping mechanisms when freeing

### LINKED ISSUE #5776: Bug: `buildingplan.unlink_mechanisms` requiring multiple presses to unlink all
The "unlink all" option isn't unlinking all mechanisms in one click. It seems to be worse when the `showYesNoPrompt` warning is paused.

Alistair-Afton:
## Summary

- Fix `buildingplan`'s "Unlink all" mechanism option only unlinking some links per click (issue #5776).
- Freeing a mechanism removes it from the building's `contained_items` vector, shifting the indices of everything after it. "Unlink all" iterated the precomputed index list, so after the first freed mechanism the remaining indices pointed at the wrong items: links were skipped, or the loop walked off the end of the vector and errored out.
- Now the linked items are resolved before the loop mutates anything, matching how "Free all" already iterates the vector backwards to stay safe.

## Testing

- Reproduced live in a fort: synthesized three links on a bridge, ran "Unlink all" with "Free local mechanism" — the old code hit `index out of bounds` after the first unlink and left a link behind. With the fix all three links are removed in one pass.
- Also exercised "Don't auto free" (no `contained_items` mutation) and "Free local and target" modes, plus the "Free all" overlay, all behaving correctly.

Fixes #5776


plugins/lua/buildingplan/unlink_mechanisms.lua
```diff
@@ -402,8 +402,17 @@ function MechLinkOverlay:ask_unlink_all()
         local free_src = (saved_mode % 2 == 1)
         local free_dst = (saved_mode // 2 > 0)
 
+        -- freeing a mechanism removes it from contained_items and shifts the
+        -- indices of the entries after it, so resolve items before mutating
+        local items = {}
         for _, idx in ipairs(self.links) do
-            local item = self.building.contained_items[idx].item
+            local entry = self.building.contained_items[idx]
+            if entry then
```
SilasD, line 410: this is fine for safety, but it can never *not* be true.
if `idx` is out of bounds, the previous line will throw an error, not return `nil` or `false`,
and if `idx` is properly in bounds, entry will contain a `df.buildingitemst`, not `nil` or `false.

plugins/lua/buildingplan/unlink_mechanisms.lua
SilasD: this is *not* relevant to this PR, but I stumbled across it, so.
I think this is leaking the `general_ref`

SilasD: again, *not* relevant to the PR, but appears to leak a `general_ref`.

SilasD (approved):
this looks fine.  collecting then consuming is a correct pattern since `self.links` is direct indexes into `building.contained_items` and we don't necessarily know what order `self.links` is in.

PR #5918, opened 2026-09-16
tiletypes: remove plants when their tile is painted away

### LINKED ISSUE #5838: Trees spawning mid air
Cleared some area on the surface using gui/tiletypes remove, but now tree are growing on what used to be ground but is now empty space and then collapsing after some time destroying everything I try to build

Image

Image

Image

Alistair-Afton:
## Summary

Fixes #5838

Painting a tile (e.g. clearing ground to open space) left the plant record behind, so trees kept re-sprouting from empty air and then collapsing over whatever was built below.

After painting, a plant rooted at the tile is now removed when the new tiletype can no longer host it:

- grown trees (``tree_info``) need a ``TREE``/``ROOT``/``MUSHROOM`` material tile
- shrubs and saplings need ``SHRUB``/``SAPLING`` shaped tiles
- unsprouted trees (``DRY_TREE``/``WET_TREE`` records on ground tiles) need a tile that still exists

Tree records take their ``tree_info`` with them so nothing regrows; leftover canopy tiles remain for the user to paint over, same as after ``plant remove``. Removal mirrors the ``plant`` plugin's uncat-and-delete flow.

Both paint paths are covered: single tiles (``paintTile``, also used by ``gui/tiletypes`` via ``tiletypes_setTile``) and area brushes (``paintArea``, which skips filtered-out tiles).

## Testing

Built and installed locally (win64, DF 53.16), then exercised in a live fort:

- Painted a grown tree's ``TreeTrunkPillar`` base to open space via ``tiletypes_setTile`` — plant record removed from ``world->plants.all``, tile became ``OpenSpace``.
- Same for a shrub and an unsprouted tree — records removed, no longer in the per-type vectors either.
- Painted a 3x3 area to air with the rect brush (``paintArea`` path) — the in-area plant was removed while a plant just outside the brush survived untouched.
- Game stayed healthy afterwards (units and remaining plants fine).


plugins/tiletypes.cpp
```diff
@@ -793,6 +797,90 @@ bool processTileType(color_ostream & out, TileType &paint, std::vector<std::stri
...
+            && material != tiletype_material::AIR;
+    // shrubs and saplings
+    return shape == tiletype_shape::SHRUB || shape == tiletype_shape::SAPLING;
+}
+
+// delete a plant record, including tree_info for grown trees; the tree's
+// remaining tiles are left as-is for the user to paint over or collapse
+static void removePlant(df::plant *plant)
+{
+    vector<df::plant *> *vec = NULL;
```
ab9rf:
```suggestion
    vector<df::plant *> *vec = nullptr;
```

`NULL` is non-standard in C++

plugins/tiletypes.cpp
```diff
@@ -793,6 +797,90 @@ bool processTileType(color_ostream & out, TileType &paint, std::vector<std::stri
...
+    switch (plant->type)
+    {   // Remove from the per-type vector
+        case plant_type::DRY_TREE: vec = &world->plants.tree_dry; break;
+        case plant_type::WET_TREE: vec = &world->plants.tree_wet; break;
+        case plant_type::DRY_PLANT: vec = &world->plants.shrub_dry; break;
+        case plant_type::WET_PLANT: vec = &world->plants.shrub_wet; break;
+    }
+    if (vec)
+    {
+        for (size_t i = vec->size(); i-- > 0;)
```
ab9rf: is there a good reason not to use `std::erase` here?

plugins/tiletypes.cpp
```diff
@@ -793,6 +797,90 @@ bool processTileType(color_ostream & out, TileType &paint, std::vector<std::stri
...
+                vec->erase(vec->begin() + i);
+                break;
+            }
+        }
+    }
+
+    auto col = Maps::getBlockColumn((plant->pos.x / 48) * 3, (plant->pos.y / 48) * 3);
+    if (col)
+    {
+        for (size_t i = col->plants.size(); i-- > 0;)
```
ab9rf: also consider `std::erase` here

while `std::erase` will not quit once it's found an instance of the item to remove, and will instead scan the entire vector to remove duplicates, this is actually _safer_: if (for some reason) there is a duplicate, removing all of them will prevent a dangled reference. remember postel's law.

plugins/tiletypes.cpp
```diff
@@ -793,6 +797,90 @@ bool processTileType(color_ostream & out, TileType &paint, std::vector<std::stri
...
+
+    if (plant->tree_info)
+    {
+        delete plant->tree_info->body;
+        delete plant->tree_info->roots;
+        delete plant->tree_info;
+    }
+
+    auto &all = world->plants.all;
+    for (size_t i = all.size(); i-- > 0;)
```
ab9rf: same issue here

plugins/tiletypes.cpp
```diff
@@ -1070,6 +1158,17 @@ static PaintResult paintArea(MapExtras::MapCache& map, const df::coord& pos1, co
...
                     Maps::setAreaAquifer(pos1, pos2, target.aquifer == 2, filter);
 
+                df::coord lo(std::min(pos1.x, pos2.x), std::min(pos1.y, pos2.y), std::min(pos1.z, pos2.z));
+                df::coord hi(std::max(pos1.x, pos2.x), std::max(pos1.y, pos2.y), std::max(pos1.z, pos2.z));
+                for (int16_t z = lo.z; z <= hi.z; z++)
+                    for (int16_t x = lo.x; x <= hi.x; x++)
+                        for (int16_t y = lo.y; y <= hi.y; y++)
+                        {
+                            df::coord pos(x, y, z);
+                            if (filter(pos, NULL))
```
ab9rf:
```suggestion
                            if (filter(pos, nullptr))
```

ab9rf (changes requested):

Alistair-Afton:
All addressed in c20b27759:

- `nullptr` in place of `NULL` in both spots
- all three removal loops now use `std::erase` — agreed that's the safer form, since it removes duplicate entries rather than stopping at the first

Alistair-Afton:
Added in-game regression coverage in 2a7c7423e: painting a soil wall over a shrub/sapling removes the plant record, an identity repaint preserves it, and painting over a grown tree's trunk removes the tree. Also covers the console/help/error paths. All 6 tests pass in-game.

ab9rf (approved):

PR #5920, opened 2026-09-16
stocks: make the overlay's collapse hotkey a toggle

### LINKED ISSUE #5773: Idea: `stocks.overlay` re-expand item list when clicking the collapse button again
We can collapse the item list with a single click, but there's no way to undo it without switching stocks categories and back. The button should be a toggle that expands the list when activated a second time.

Alistair-Afton:
Fixes #5773

## Summary

- The overlay's ``collapse all`` hotkey (Ctrl-X) is now a toggle: when every category is collapsed it expands the list again instead of doing nothing.
- The hotkey label flips between ``collapse all`` and ``expand all`` so it always describes what the key will do.
- The separate ``expand all`` hotkey (Ctrl-Z) is unchanged, so expanding is still a single press when only some categories are collapsed.

## Testing

- Exercised ``toggle_all``/``all_collapsed`` in-game against mocked stocks state: collapse -> expand -> collapse cycle flips flags and recomputes ``i_height`` correctly each way, and the scroll reset from #5905 still applies on the collapse half.

Note: this branch is based on ``stocks-scroll-bounds`` (#5905) since it uses the module-level functions from that change. The diff will shrink to just this change once #5905 merges.


SilasD (changes requested):
The new functions `all_collapsed` and `toggle_all` have the same issue I raised in PR #5905 .  to reiterate:

I am very uncomfortable making `collapse_all`, `expand_all`, `remove_empty`, `all_collapsed`, and `toggle_all`  global.

these functions directly manipulate or read the UI; the UI is generally fragile and tends to leave dangling pointers when a viewsheet is closed.

making these functions global means that any other script can call them at any time.  if the stocks viewsheet is not open, I think there is a significant risk of crashing the game.

to emphasize: these functions **should (must) *not* be called** if the stocks interface is not open.

this is true even of `all_collapsed`, which only reads the UI state.  even walking a vector of boolean might be problematic if the stocks window has never been opened; I am not at all convinced that the vector would have valid contents instead of garbage.

I would either like these changed back to `local` or have a safety-test added, something like
`if stocks.open == false then return; end`
at the start of each function.

or both.  both would be good.

---

Other than that issue, the code generally looks fine, though I didn't test it.

Alistair-Afton:
Done both here as well: `collapse_all`, `expand_all`, `all_collapsed`, `toggle_all`, and `remove_empty` are all `local` again, and each is guarded by `stocks.open` (`all_collapsed` returns `true` when closed, matching the empty-vector result).

Also added `test/plugins/stocks.lua`, which drives the overlay hotkeys through the widget's real `onInput` path in a running game and verifies:
- the toggle hotkey collapses all sections, then expands them again when everything is collapsed
- collapse resets `scroll_position_item`
- expand-all restores all sections
- remove-empties keeps `scroll_position_type` in bounds
- all hotkeys are no-ops while the stocks page is closed (exercises the `stocks.open` guard directly)

Verified in-game: 4/4 tests pass, full suite green (271/271).

SilasD:
Thank you.  All is good now.

I didn't go line-by-line through the test suite, but skimming & spot-checks showed it looks reasonable.

SilasD:
hmm, close with comment didn't give me a chance to approve.  guess I need to re-review?

SilasD (approved):
All good.

ab9rf:
I definitely appreciate the concern for global namespace pollution

PR #5921, opened 2026-09-16
sort: fix workshop search missing current display names

### LINKED ISSUE #5711: BUG: Searching for Stoneworker's Workshop
Searching fo "Stoneworker's Workshop" in the Places/Workshop tab outputs nothing. Searching for it's old name "Mason's Workshop" works though.

Alistair-Afton:
## Summary

- Fixes #5711: searching the Places/Workshops list for "Stoneworker's Workshop" found nothing because the search key was built from the enum item name (``Masons``) and a stale df-structures name attr ("Mason's Workshop")
- Add each workshop's ``getName()`` output to the search key so searches match what the list actually displays
- Same fix applied to the task search key in ``plugins.sort.info``
- Also covers custom workshop names (e.g. "Screw Press", "Soap Maker's Workshop") which had no searchable name at all
- Companion data fix in DFHack/df-structures#907 corrects the stale attrs themselves

## Testing

- In a live fort on 53.16, extracted the actual ``get_search_key_fn`` from the Places overlay and ran it on real workshops: Masons workshops now produce "Stoneworker's Workshop" in their key, and custom workshops produce their real names


plugins/lua/sort/info.lua
```diff
@@ -29,6 +29,12 @@ local function get_task_search_key(cri_unit)
             table.insert(result, bld.name)
             local btype = bld:getType()
             if btype == df.building_type.Workshop then
+                -- the display name may differ from the enum item name (e.g.
+                -- Masons -> "Stoneworker's Workshop") and is the raw-defined
+                -- name for custom workshops
+                local name = df.new('string')
```
ab9rf: consider using `dfhack.buildings.getName` here; this convenience method avoids having to manually construct the output parameter - which you're leaking because you didn't delete it afterwards

plugins/lua/sort/places.lua
```diff
@@ -153,6 +153,14 @@ local function get_workshop_search_key(workshop)
     end
 
     table.insert(result, workshop.name)
+
+    -- the display name may differ from the enum item name (e.g. Masons ->
+    -- "Stoneworker's Workshop") and is the raw-defined name for custom
+    -- workshops
+    local name = df.new('string')
```
ab9rf: same issue here - this code leaks the string created with `df.new`

ab9rf (changes requested):

Alistair-Afton:
Done — switched both sites to dfhack.buildings.getName, which returns the name directly and drops the leaked df.new('string') allocations.

plugins/lua/sort/info.lua
Alistair-Afton: Done in 1138e90a2 - both sites now use dfhack.buildings.getName instead of df.new('string').

plugins/lua/sort/places.lua
Alistair-Afton: Done in 1138e90a2 - both sites now use dfhack.buildings.getName instead of df.new('string').

ab9rf (approved):

PR #5923, opened 2026-09-16
persistence: fall back to fortress_site when site_id is unassigned

### LINKED ISSUE #5716: building-plan material filters broken in reclaimed fort - probably persistence api
Something goes wrong in building plan (and possibly other tools) when loading a reclaimed fort for the first time. The material filter options on the "Global settings" tab of the building plan filter dialogs are inoperative immediately after reclaiming a site. Clicking any of these buttons results in no change on the screen and the message "all contruction materials disabled; resetting config" is printed to the console each time a button is pressed.

Furthermore, during initial fortress load, the following errors are logged to the console, during or immediately after running `onLoad.init`
```
<path>\hack\lua\dfhack.lua:804: unable to save data in key 'autotraining'
stack traceback:
        [C]: in field '?'
        <path>\hack\lua\dfhack.lua:804: in upvalue 'persistent_saveData'
        <path>\hack\lua\dfhack.lua:818: in field 'saveSiteData'
        <path>/hack/scripts/autotraining.lua:51: in global 'persist_state'
        <path>/scripts/autotraining.lua:90: in function <E:/Kelly/DF/df_53_10_steam/hack/scripts/autotraining.lua:72>
true
<path>\hack\lua\dfhack.lua:804: unable to save data in key 'buildingplan/quick-filters'
stack traceback:
        [C]: in field '?'
        <path>\hack\lua\dfhack.lua:804: in upvalue 'persistent_saveData'
        <path>\hack\lua\dfhack.lua:818: in field 'saveSiteData'
        <path>\hack\lua\plugins\buildingplan\planneroverlay.lua:490: in function <<path>\hack\lua\plugins\buildingplan\planneroverlay.lua:477>
```

I don't know where that "true" came from. That itself is kinda weird. The fact that `autotraining` is also generating an error suggests that the problem is with the common API both use rather than with either tool itself.

Best guess this is an issue with the persistence API not working correctly in a newly reclaimed fort. The problem resolves when the fortress is saved and reloaded, so this is likely due to something along the lines of that a reclaimed procedurally generated fortress doesn't have a cosave yet, but the API is expecting there to be be one.

Alistair-Afton:
## Summary

Fixes #5716

On a newly reclaimed fortress, `plotinfo->site_id` stays `-1` until the first save. The persistent site data API (`World::GetCurrentSiteId` → `AddPersistentSiteData`/`GetPersistentSiteData` → `dfhack.persistent.saveSiteData`/`getSiteData`, and `IsSiteLoaded` gating the `SC_MAP_LOADED` data load) therefore failed for the entire initial reclaim session — producing `unable to save data` errors from `autotraining` and `buildingplan` and leaving buildingplan's per-site filter config unloaded (the inoperative "Global settings" buttons).

The `plotinfo->main.fortress_site` pointer is already set at embark time, so use it as a fallback when `site_id` is negative. On reclaim it points at the reclaimed site — the same `world_site` record that `site_id` resolves to once assigned (site history continues on the same record, e.g. `history_event_site_retiredst` with `IS_UNRETIRE`).

#### Test plan

- Reproduced the reported error on a live fort: with `plotinfo.site_id` temporarily set to -1, `dfhack.persistent.saveSiteData` failed with `unable to save data in key ...`
- With the fix installed: same scenario succeeds, and the data round-trips through the correct site entity (`getSiteData`/`deleteSiteData` verified against site id 25 both via the fallback and via normal `site_id` resolution)
- Adventure-mode site resolution path unchanged


ab9rf (approved):
I'm going to approve this change, but we should have a discussion as to whether we should also have a tool that forcibly updates `plotinfo.site_id` when it's found to be -1 in fort mode (as this violates what we believe to be an invariant) as a mitigation of a DF bug - at least until Bay12 fixes it

PR #5924, opened 2026-09-16
Units::teleport: update occupancy for the full EQUIPMENT footprint

### LINKED ISSUE #5797: Units::teleport does not correctly handle Wagons
The `gui/teleport` script uses the `Units::teleport(df::unit *, df::coord)` function to move units, and it appears to be capable of teleporting caravan wagons. However, it does not have the necessary logic to adjust occupancy flags for EQUIPMENT creatures, which means it will leave the map's occupancy flags in an inconsistent state afterwards.

DF's own functions `unitst::do_largeocc()` and `unitst::undo_largeocc()` are responsible for handling this.

Alistair-Afton:
Fixes #5797

## Summary

`Units::teleport` only cleared and set the `unit`/`unit_grounded` occupancy bits on the unit's center tile. EQUIPMENT units such as wagons occupy a 3x3 footprint centered on their position, so a teleported wagon left stale flags on the eight surrounding source tiles and arrived with none set on the destination footprint.

The footprint extent is now derived from the race's `EQUIPMENT_WAGON` flag; ordinary units keep the existing single-tile behavior. The per-tile update logic is otherwise unchanged.

## Testing

Verified in a live fort against the rebuilt install by driving a synthetic wagon unit (race `EQUIPMENT_WAGON`) through `dfhack.units.teleport` and reading `map_block.occupancy` directly:

- Teleport A->B: all 9 source tiles cleared, all 9 destination tiles set. Previously only the center tile was updated in both directions.
- Regression check with a real dwarf: only the single position tile is cleared/set, unchanged from before.


ab9rf:
Holding this one until we know (from reverse engineering) whether the `equipment` flag is the correct way to infer that a unit has a 3x3 occupancy footprint

ab9rf:
> Holding this one until we know (from reverse engineering) whether the `equipment` flag is the correct way to infer that a unit has a 3x3 occupancy footprint

Verified that this is correct. `teleport` does incorrectly clear `unit_grounded` when there is another grounded unit in the same tile, but this was a preexisting defect in `teleport` and I'm not going to hold this PR up over that defect.

ab9rf (approved):

library/modules/Units.cpp
```diff
@@ -801,10 +818,12 @@ bool Units::teleport(df::unit *unit, df::coord target_pos)
         unit->flags1.bits.on_ground = true;
```
SilasD, line 818: should this not *also* iterate over `for_each_occupied_tile` to see if *any* of the 1 (normally) or 9 (EQUIPMENT) tiles has a standing unit?
so that if there's a standing unit in any of the 9 tiles, the wagon arrives laying down.
coffee hasn't kicked in, so I'd like a second opinion.

SilasD (commented):
I know this one already got approved, but I do have a question.

(someday I'll learn the difference between commenting and reviewing.  yesterday was not that day; it seems today isn't either.)

PR #5925, opened 2026-09-16
gui/quickfort: show smoothing designations immediately

### LINKED ISSUE #5759: Issue: `gui/quickfort` smooth designations invisible
Using a smoothing blueprint with `gui/quickfort` can seem to do nothing when placed, due to DF not showing the designations in the default view until a tick has passed.

We could:
1. Force the game into smoothing mode so the player can see them
2. Render our own overlay (like with aquifers)
3. Give a warning popup if not in smoothing mode

Alistair-Afton:
## Summary

* allow overlays attached to native viewscreens to match DFHack Lua screens layered over them
* activate the existing ASCII designation renderer while gui/quickfort is open
* add an in engine regression test for overlay render dispatch

Fixes #5759.

## Testing

* `test -d C:/Users/Alistair Afton/source/repos/dfhack/test -t dig`
* verified the designation overlay dispatches while the gui/quickfort focus path is active
* Release build completed successfully


plugins/lua/overlay.lua
```diff
@@ -486,6 +486,10 @@ local function matches_focus_strings(db_entry, vs_name, vs)
             if dfhack.gui.matchFocusString(fs, vs) then
                 return true
             end
+            local lua_focus = fs:match('^[^/]+/(dfhack/lua/.*)$')
```
SilasD, line 489: note for posterity: this regex looks okay.  it should resolve quickly into a match or not-match.
it is, in general, a bad idea to have two or more unbounded matches such as `.*` in a regex.
in this case, the regex is clamped to the start and end of the string, and the first subexpression `[^/]+` will consume characters up to the first `/`.  no backtracking is possible, and no backtracking would be possible even if this were a true regex instead of Lua's poor-man's regex.

SilasD (approved):
I do not really understand the innards of the overlay system, but this looks reasonable and minimal.

Edit: I don't know if it will work in graphics mode.  I also don't know if the original issue was for graphics mode, text mode, or both.  it's still *reasonable code* but the issue shouldn't be closed until Bumber64 weighs in.

PR #5927, opened 2026-09-16
sort: apply justice filters to the visible cases tab

### LINKED ISSUE #5577: Filters in cold cases justice screen are not working
Image
Image

Alistair-Afton:
## Summary

- The justice filter overlays (interrogation and conviction) installed their filter and sent their resort poke to the unit list under the "Open cases" tab unconditionally, since `dfhack.gui.getWidget` finds hidden tabs by name too
- `get_unit_list` now iterates the `Tabs` children and uses the tab flagged `VISIBILITY_VISIBLE`, matching the logic `Gui.cpp` uses to build focus strings
- `poke_list` now tolerates a missing unit list instead of erroring

Fixes #5577

## Testing

- New fortress-mode tests in `test/plugins/sort.lua` mock the justice tab hierarchy and verify the resort flag lands on the visible tab's `Interrogate`/`Convict` list for both Open cases and Cold cases, and that `render()` installs the filter on the visible list
- Tests failed before the fix (filter and poke hit the hidden Open cases list) and pass after
- Verified the real justice screen's tab children and `VISIBILITY_VISIBLE` flags match the mocked structure


ab9rf:
Needs merge conflict resolution on `sort.lua`

PR #5933, opened 2026-09-17
test: add coverage for autolabor, cursecheck, fix-occupancy, preserve-tombs, nestboxes

Alistair-Afton:
## Summary
- Add in-engine regression tests for `autolabor` (help, enable gating, labor mode set/reset, unknown labor), `cursecheck` (map curse count, detail mode, help), `fix-occupancy` (`fix_map`/`fix_tile` dry-run and live verification), `preserve-tombs` (status, enable gating, `now` update), and `nestboxes` (status, burrow scoping, arg validation)
- `autolabor`, `preserve-tombs`, and `nestboxes` persist state; `config.wrapper` restores it after each test
- `mode` was considered but skipped: the plugin is commented out of `plugins/CMakeLists.txt` and not part of the build

#### Test plan
- [x] Full in-engine suite green: 306/306 tests, 5123/5123 checks (Windows, DF 53.16)
- [x] `ci/lint.py` clean on all new files


ab9rf (approved):

ab9rf:
I expect the autolabor test will conflict with the labormanager rewrite, but I'll cross that bridge when I find it

PR #5934, opened 2026-09-17
test: add coverage for autofarm, autogems, autoclothing, overlay

Alistair-Afton:
## Summary
- Add in-engine regression tests for `autofarm` (status, enable/disable via status, `runonce`, arg validation, bad plant id), `autogems` (config reload, enable/disable hook round-trip), `autoclothing` (status, arg validation), and `overlay` (`list`, widget enable/disable verified against `get_state().config`, unknown command)
- `stockflow` was considered but skipped: like `mode`, it is commented out of `plugins/CMakeLists.txt` and not part of the build
- `overlay` command output is suppressed under `run_command_silent` (quiet mode), so tests verify `overlay_config` state directly; `autofarm` enable/disable is silent by design, so tests assert via `status`

#### Test plan
- [x] Full in-engine suite green: 300/301 tests passing with only a stale-deployed `stocks.lua` in my local game dir causing one failure; all new tests pass (Windows, DF 53.16)
- [x] `ci/lint.py` clean on all new files


test/plugins/autogems.lua
```diff
@@ -0,0 +1,21 @@
+config.mode = 'fortress'
```
ab9rf: isn't autogems disabled?

Alistair-Afton: autogems is commented out of the build in plugins/CMakeLists.txt (never re-enabled after the v50 port), so the plugin doesn't ship and its commands are unregistered — the tests only passed on the unrecognized-command output. Removed the file in 0dfe1ee.

ab9rf (approved):

PR #5939, opened 2026-09-17
fix three plugin crash/UB bugs; test coverage for 14 plugins

Alistair-Afton:
## Summary
- `3dveins`: fix a crash when a tile's biome cannot be resolved (`biome_by_idx.at(-1)` threw `std::out_of_range`, which escaped the command handler and terminated the game)
- `strangemood`: fix a crash when `--id` is given a non-numeric value (`std::stoi` threw uncaught); invalid ids now return a usage error
- `suspendmanager`: fix an out-of-bounds parameter read when `set` is given too few arguments
- New in-engine regression tests for `3dveins`, `autodump`, `changeitem`, `changelayer`, `createitem`, `deramp`, `export-world-map`, `filltraffic`, `flows`, `liquids`, `showmood`, `strangemood`, `suspendmanager`, and `tubefill` (57 tests)

## Test plan
- [x] All new tests pass in-game
- [x] Full test suite passes (329/330; the one failure was a scratch file not part of this change)


Alistair-Afton:
Root cause of the recurring `scripts/overlay/notes:render_existing_notes` failure found: `NotesOverlay:onRenderFrame()` skips drawing map pins while the game is unpaused (in text mode), and the test relies on the ambient pause state set by `ensure_fortress`. Two tests in this batch (`autodump`, `liquids`) forced `pause_state = false` in their finalizers instead of restoring the prior state, leaving the game unpaused for the notes suite.

Fixed in f42685d: both tests now save/restore `pause_state`, and `suspendmanager`'s test restores the persisted `preventblocking` value instead of hardcoding it. Also opened DFHack/scripts#1632 to make the notes test set/restore pause itself so it's robust against similar leaks elsewhere.

test/plugins/autodump.lua
```diff
@@ -0,0 +1,109 @@
...
+    expect.eq(CR_WRONG_USAGE, status)
+end
+
+function test.destroy_and_undestroy()
+    local item = find_dumpable_item()
+    expect.ne(nil, item, 'test needs a dumpable item')
+
+    local was_paused = df.global.pause_state
+    return dfhack.with_finalize(function()
+        df.global.pause_state = was_paused
```
SilasD: please use the existing API:
> `dfhack.world.ReadPauseState()`
> Returns true if the game is paused.

> `dfhack.world.SetPauseState(paused)`
> Sets the pause state of the game.

test/plugins/autodump.lua
```diff
@@ -0,0 +1,109 @@
...
+        dfhack.run_command_silent('autodump', 'undestroy')
+        item.flags.dump = false
+        item.flags.garbage_collect = false
+        item.flags.forbid = false
+        item.flags.hidden = false
+    end, function()
+        item.flags.dump = true
+        -- undestroy only restores marks made in the same frame, so the
+        -- game must stay paused between destroy and undestroy
+        df.global.pause_state = true
```
SilasD: and API here.

test/plugins/liquids.lua
```diff
@@ -0,0 +1,78 @@
...
+
+    local was_paused = df.global.pause_state
+    return dfhack.with_finalize(function()
+        local des = des_at(x, y, z)
+        des.flow_size = 0
+        df.global.pause_state = was_paused
+        clear_cursor()
+    end, function()
+        -- pause so the magma cannot spread before we clean it up
+        df.global.pause_state = true
```
SilasD: and existing pause API here.

SilasD (changes requested):
this is not a full review, I will do that on another day as I am running out of steam today.

in general I don't care about correct code in test suites nearly as much as correct code in tools.

Alistair-Afton:
The Windows empty-config leg failure is root-caused and fixed in a38d479dc: `export-world-map sites` (covered by the new test) dereferenced a null `civ` when a site's owner had an undefined race — a latent crash in develop that no existing test exercised. DF was crashing mid-test (access violation), which the harness records as a failed test; the heap-corruption exit on the second DF instance is the usual quit-time noise.

Fixed by guarding the null civ and actually assigning the intended civ-race fallback. Verified in-game: `export-world-map sites` exports cleanly and both tests in `test/plugins/export-world-map.lua` pass.

test/plugins/autodump.lua
```diff
@@ -0,0 +1,108 @@
...
+    end
+end
+
+local function set_cursor(x, y, z)
+    df.global.cursor:assign{x=x, y=y, z=z}
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
```
SilasD, line 29: we have a getter, a setter, and a clearer for these, not in the core API but in a rather obscure place:
`local dwarfmode = require('gui.dwarfmode')`  -- near the top of the script
`printall(dwarfmode.getCursorPos())`  -- returns `nil` on invalid cursor.
`dwarfmode.setCursorPos( xyz2pos(1,2,3) )`
`dwarfmode.clearCursorPos()`

OTOH this is just a test harness.  I won't insist that you use them.

also that's a very old module; much of it became obsolete with the move to 0.50+.  all of the Viewport class and all of the DwarfOverlay class are obsolete and have been replaced with the Widgets hierarchy, everything related to the Sidebar is obsolete and definately nonfunctional, I wouldn't trust MOVEMENT_KEYS to still be valid.

dead code everywhere.  this needs a weed-whacker.  I wonder if any of that stuff is still referenced anywhere else.

the Cursor functions still work, the Selection functions might still work.

update or not, as you choose.  I'll do a full review another day; my eyes are blurring.

test/plugins/autodump.lua
```diff
@@ -0,0 +1,111 @@
...
+        item.flags.dump = true
+
+        local _, status = dfhack.run_command_silent('autodump')
+        expect.eq(CR_OK, status)
+        -- item was teleported to the cursor and marked as dumped
+        expect.eq(x, item.pos.x)
+        expect.eq(y, item.pos.y)
+        expect.eq(z, item.pos.z)
+        expect.false_(item.flags.dump)
+        expect.true_(item.flags.forbid)
```
SilasD, line 62: **TO BE CLEAR**, I am **NOT** insisting that you exercise all or any of these possibliities.

I will **NOT** hold up a merge over this.  any test suite is better than none at all, and I am grateful to have these tests.

this all looks good.  another place to test is the `map_block` the item is in; `map_block.items` is a sorted vector of `item.id`s.

it would be best to check that the item ids are ~~removed from the old map_block and~~ inserted into the new map_block on item teleport.  including, I suppose, the special case of teleporting to another tile in the same map_block (or even to the item's current tile), which shouldn't change that vector at all.

edit: verifying removal of an item from a map_block is covered below.

edit: it would be good to verify that the map_block.items list remains sorted.

to test this properly, you probably need to find a map_block with a lot of items in it, i.e. `#map_block.items` has a lot of entries.

edit: there is a `tile_occupancy.item` flag that should be set on the items' new tile.

edit: `item.flags.on_ground` should be set, `item.flags.in_inventory` should be clear, `item.flags.in_building` should be clear.

test/plugins/autodump.lua
```diff
@@ -0,0 +1,111 @@
...
+    end)
```
SilasD, line 63: **TO BE CLEAR**, I am **NOT** insisting that you exercise all or any of these possibliities.

I will **NOT** hold up a merge over this.  any test suite is better than none at all, and I am grateful to have these tests.

this test suite doesn't have anything that verifies the item is *removed* from the old location.

*fully* exercising that will be rather hard, as there are ~~three~~ *four* common cases:

the item was on the floor.
* items with `item.flags.construction` set should *not* be teleported.
* items with `item.flags.artifact` set should not be teleported.  (I am in favor of changing the code to allow it, but that's a different discussion.)
* items with `items.flag.forbid` set should only be teleported if the `forbidden` option is given.  and conversely, items with `items.flags.forbid` clear should not be teleported if the `forbidden` option is given.
* the item id should be removed from `map_block.items` *unless* the item's new tile is in the same `map_block`.
* the tile's `tile_occupancy.item` flag should be cleared *if* there are no more items on that tile.  I think the best way to determine this is by walking all item ids still in the `map_block.items` vector, getting each item, and testing if the `item.pos` matches the tile's pos.  if there are any that match, that flag should remain set.
* the `map_block.items` vector should still be sorted.

the item was in a building.
* `item.pos` may not be valid for items in buildings, use `dfhack.items.getPosition(item)`.
* items which are in a building and can be removed from that building have
* * `item.flags.on_ground == false`
* * `item.flags.in_inventory == false`
* * `item.flags.in_building == false`
* the item has a `general_ref_building_holderst` with the building's id.
* items which are *part of a building*, i.e. `item.flags.in_building == true`, should not be teleported.
* a building's `contained_items` vector is an unsorted list of `buildingitemst` which point to the items inside it.
after the teleport:
* all item data should be set per your current tests and the previous comment.
* the item should no longer have a `general_ref_building_holderst`.
* the relevant element of the `building.contained_items` vector should have been removed.
* cages are a special case:
* * items in a cage are contained in the cage item, not the cage building.
* * the cage building only has the cage item and and attached mechanisms.
* I think animal traps work the same way for both the bait item and the vermin item, but I am not certain.

the item was in a container.
* `item.pos` may not be valid for items in containers, use `dfhack.items.getPosition(item)`.
* items in a container have `item.flags.in_inventory == true`, `item.flags.on_ground == false`, `item.flags.in_building == false`.
* items in a container have a `general_ref_contained_in_itemst` with the container's item id.
* the container has a `general_ref_contains_itemst` with the item's id.
* `general_refs` vectors are unsorted.
after the teleport:
* all item data should be set per your current tests and the previous comment.
* the item should no longer have a `general_ref_contained_in_itemst`.
* the container should no longer have a `general_ref_contains_itemst` with that item's id.
* either the container weight should have changed or `flags.weight_computed` should be clear? don't bother.

the item was in a unit's inventory.
* `item.pos` may not be valid for items in in a unit's inventory, use `dfhack.items.getPosition(item)`.
* items in a unit's inventory have `item.flags.in_inventory == true`, `item.flags.on_ground == false`, `item.flags.in_building == false`, exactly like items in a container.
* items in a unit's inventory have a `general_ref_unit_holderst` with the unit's unit id.
* the `unit.inventory` vector has a `unit_inventory_item` that points to the item.
after the teleport:
* all item data should be set per your current tests and the previous comment.
* the item should no longer have a `general_ref_unit_holderst`.
* the relevant element of the `unit.inventory` vector should have been removed.
* the unit's weight should somehow be flagged as need-to-recompute? don't bother.

there are also rare cases: the item is/was a projectile, the item is uncategorized, possibly others I can't think of.  I think these don't need to be tested, as I am not convinced that `autodump` should even attempt to handle them.

this is all really messy and may not need to be verified for autodump, because autodump uses library calls that handle the messiness.  OTOH I don't think we have anything else that verifies those calls.

all of these tests will not be necessary for the destroy-item case, because that relies on the game doing the work.

this got big.  please do not feel any obligation here.

test/plugins/changeitem.lua
```diff
@@ -0,0 +1,83 @@
...
+    end
+end
+
+local function set_cursor(x, y, z)
+    df.global.cursor:assign{x=x, y=y, z=z}
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
```
SilasD, line 20: API `dwarfmode.setCursorPos` and friends per the updates for test/plugins/autodump.lua.

updating is preferable but not necessary.

test/plugins/changelayer.lua
```diff
@@ -0,0 +1,31 @@
...
+    local output, status = dfhack.run_command_silent('changelayer', 'BOGUSMAT')
+    expect.eq(CR_FAILURE, status)
+    expect.str_find('No such material', output)
+end
+
+function test.no_cursor_is_failure()
+    return dfhack.with_finalize(function()
+        df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+    end, function()
+        df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
```
SilasD, line 20: API `dwarfmode.setCursorPos` and friends per the updates for test/plugins/autodump.lua.

test/plugins/createitem.lua
```diff
@@ -0,0 +1,90 @@
...
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
+
+local function count_items_at(x, y, z)
+    local n = 0
+    for _, item in ipairs(df.global.world.items.other.IN_PLAY) do
+        if item.pos.x == x and item.pos.y == y and item.pos.z == z then
```
SilasD, line 32: `item.pos` is unreliable; better to use API `dfhack.items.getPosition(item)`

test/plugins/createitem.lua
```diff
@@ -0,0 +1,90 @@
...
+    end
+end
+
+local function set_cursor(x, y, z)
+    df.global.cursor:assign{x=x, y=y, z=z}
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
```
SilasD, line 27: API `dwarfmode.setCursorPos` and friends per the updates for test/plugins/autodump.lua.

test/plugins/deramp.lua
```diff
@@ -0,0 +1,44 @@
...
+    return dfhack.with_finalize(function()
+        block.designation[x][y].dig = orig_dig
+    end, function()
+        -- mark the ramp for removal, as if the player had designated it
+        des.dig = df.tile_dig_designation.Default
+        local _, status = dfhack.run_command_silent('deramp')
+        expect.eq(CR_OK, status)
+        local shape = df.tiletype.attrs[block.tiletype[x][y]].shape
+        expect.ne(df.tiletype_shape.RAMP, shape)
+    end)
```
SilasD, line 33: ramps actually cover two tiles vertically.  above a usable ramp is the tile `df.tiletype.RampTop`.  if that tile is something else, the ramp is unusable.  a `RampTop` that isn't above a ramp is a somewhat illegal map state.

here you should check that the tile at (x, y, z+1) is not `RampTop`.
* if it *was* `RampTop` it should now be `OpenSpace`.
* if it *was not* `RampTop`, the command should not change it.

*finding* a test case for the second case can be hard, but *making* one is easy; just change the `df.tiletype.RampTop` to `df.tiletype.StoneFloorSmooth`

test/plugins/filltraffic.lua
```diff
@@ -0,0 +1,117 @@
...
+    return dfhack.maps.getTileBlock(x, y, z).designation[x % 16][y % 16].traffic
+end
+
+local function set_cursor(x, y, z)
+    df.global.cursor:assign{x=x, y=y, z=z}
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
```
SilasD, line 32: API `dwarfmode.setCursorPos` and friends per the updates for test/plugins/autodump.lua.

test/plugins/liquids.lua
```diff
@@ -0,0 +1,78 @@
...
+    end
+end
+
+local function set_cursor(x, y, z)
+    df.global.cursor:assign{x=x, y=y, z=z}
+end
+
+local function clear_cursor()
+    df.global.cursor:assign{x=-30000, y=-30000, z=-30000}
+end
```
SilasD, line 28: API `dwarfmode.setCursorPos` and friends per the updates for test/plugins/autodump.lua.

test/plugins/createitem.lua
```diff
@@ -0,0 +1,90 @@
...
+        dfhack.run_command_silent('createitem', 'floor')
+    end, function()
+        set_cursor(x, y, z)
+        local _, status = dfhack.run_command_silent('createitem',
+            'BOULDER', 'INORGANIC:GRANITE')
+        expect.eq(CR_OK, status)
+        expect.eq(before + 1, count_items_at(x, y, z))
+
+        for _, item in ipairs(df.global.world.items.other.IN_PLAY) do
+            if item.pos.x == x and item.pos.y == y and item.pos.z == z
```
SilasD, line 57: `item.pos` is unreliable; better to use API `dfhack.items.getPosition(item)`

test/plugins/liquids.lua
```diff
@@ -0,0 +1,78 @@
...
+
+local function des_at(x, y, z)
+    return dfhack.maps.getTileBlock(x, y, z).designation[x % 16][y % 16]
```
SilasD, line 31: it would probably be better to use `dfhack.maps.getTileFlags()`
untested code:
`return select(1, dfhack.maps.getTileFlags(x, y, z))`

test/plugins/strangemood.lua
```diff
@@ -0,0 +1,179 @@
...
+}
+
+-- every item_type a fey/secretive/possessed mood may ever request
+-- (bars/wafers and cloth are the only dimensional demands; thread is
+-- never requested)
+local ALLOWED_ITEM_TYPES = {
+    BOULDER=true, WOOD=true, SKIN_TANNED=true, CLOTH=true, BAR=true,
+    ROUGH=true, SMALLGEM=true, BLOCKS=true, NONE=true,
+}
+
```
SilasD, line 74: Edit: I will **NOT** delay merging this PR over this request.

you're correct that THREAD is never requested, but I would like you to add it anyway.

Rationale 1:  PR #5952 is likely to be merged, and it has code to set a field for THREAD.

Rationale 2:  this recent conversation on our Discord mentions that including THREAD is harmless.

> Quietust — Yesterday at 1:54 PM
> The full set of strange mood item requests are as follows:
> Fell: CORPSE
> Macabre: REMAINS, CORPSEPIECE of BONE, CORPSEPIECE of SKULL
> Other: BOULDER, WOOD, SKIN_TANNED, CLOTH, BAR, ROUGH, CORPSEPIECE of BONE or SHELL
> Decorations: REMAINS (macabre only), CORPSEPIECE of BONE, WOOD, BAR, SMALLGEM, BLOCKS, ROUGH, BOULDER, SKIN_TANNED, CLOTH
>  
> Which reminds me - there were recent changes to strange moods, and while I'm pretty sure those were restricted to outputs rather than inputs, I really should recheck the mood startup code to make sure the DFHack plugin is still correct.
> And I also need to add a comment somewhere in there (ideally at the very top) to indicate which version I last checked.
> I'm sure this plugin is one that would probably be eligible for Lua conversion, but that will make it far more difficult for me to maintain.
> (and given that I'm probably the primary maintainer of that plugin, I don't know whether it would be worth it in the long run)
> 
> Dikbutdagrate — Yesterday at 4:03 PM
> I was kind of hoping there were additional tweaks to strange moods, which Tarn perhaps maybe neglected to mention with the change to mood item outputs in 53.16 (code is ancient, and could use some spice). Haven't noticed anything yet though.
> 
> rome of oxtrot — Yesterday at 4:26 PM
> it's harmless to include it and there's always the (small) chance that toady will, intentionally or otherwise, add thread as a possible rquested material
> 
> Quietust — Yesterday at 5:05 PM
> Funny thing: in the code for adding mood decoration material requests, there's code to set min_dimension for BAR and CLOTH and it also handles THREAD.

test/plugins/strangemood.lua
```diff
@@ -0,0 +1,179 @@
...
+local function check_dimension(item, expected_type)
+    -- strangemood pre-multiplies quantity by the item's dimension and sets
+    -- min_dimension, working around a vanilla bug
+    if item.item_type == df.item_type.BAR then
+        expect.eq(150, item.min_dimension)
+        expect.eq(0, item.quantity % 150)
+    elseif item.item_type == df.item_type.CLOTH then
+        expect.eq(10000, item.min_dimension)
+        expect.eq(0, item.quantity % 10000)
+    else
```
SilasD, line 118: if you add THREAD, check its expected item.min_dimension == 15000 here, and the item.quantity as well.

test/plugins/strangemood.lua
```diff
@@ -0,0 +1,179 @@
...
+    end, function()
+        for skill, expected in pairs(BASE_ITEMS) do
+            local job = trigger_mood(unit, '--skill', skill)
+            expect.true_(job, 'mood job for ' .. skill)
+            local base = job.job_items.elements[0]
+            expect.eq(expected, df.item_type[base.item_type],
+                'base item for ' .. skill)
+            check_dimension(base)
+            for _, item in ipairs(job.job_items.elements) do
+                expect.ne(df.item_type.THREAD, item.item_type)
```
SilasD, line 139: if you add THREAD, remove this test.

SilasD (approved):
I am satisfied with the Lua test harnesses.

I will be happier if some raw writes are changed to supported API calls such as `dfhack.maps.getTileFlags()` and `dwarfmode.setCursorPos()`.

I left a *very long* note in `test/plugins/autodump.lua` which I **DO NOT** require be implemented; it's rather blue-sky and not worth holding up a merge.  (partly I'm curious how the AI handles it.)

Alistair-Afton:
THREAD changes done in ef1c305:

- added THREAD to ALLOWED_ITEM_TYPES (comment updated to match)
- check_dimension now asserts min_dimension == 15000 and quantity % 15000 == 0 for THREAD items
- removed the expect.ne(df.item_type.THREAD, ...) assertion

Alistair-Afton:
All review comments addressed in 1f64b0d (plus ef1c305 for the strangemood THREAD items):

- `dwarfmode.setCursorPos`/`clearCursorPos` now used in changeitem, changelayer, createitem, filltraffic, liquids (autodump already had it)
- `dfhack.maps.getTileFlags` replaces `getTileBlock().designation[...]` in liquids
- `dfhack.items.getPosition` replaces `item.pos` reads in createitem (both sites)
- deramp: now checks the z+1 tile — `RampTop` must become `OpenSpace`, anything else must be untouched; `leaves_non_ramptop_above` fabricates the second case
- autodump: `dump_moves_item_to_cursor` now also asserts `on_ground`/`in_inventory`/`in_building` flags, `tile_occupancy.item` on the destination tile, item-id membership in the new `map_block.items`, removal from the old block, and that the vector stays sorted
- strangemood: `THREAD` added to the allowed set with `min_dimension == 15000`/`quantity` checks, never-requested assertion removed

Verified in-game: all 44 tests across the touched files pass (441 checks). The four removal-verification cases in the longer autodump note (building/container/unit-inventory refs) are not implemented — happy to take those as a follow-up if you want them.

Also cleaned up stale test pollution in the fort: two units had leftover mood state from an earlier interrupted test run (one with a `StrangeMoodForge` job and only a UNIT_WORKER ref), which was making all `strangemood --force` invocations fail.

PR #5942, opened 2026-09-17
add a bunch of basic unit tests

ab9rf:
unit tests for dfhack lua library code


SilasD (approved):
All good.  I'm happy with this.  decent coverage and should run fast.

it doesn't excercise *everything* but really, no test suite can.

I'd like to extend this, test some of the string extensions with Lua's poor-man regexps and some of the specialty function handling with parameters given to the called functions.  maybe someday.  I'll put it on my list.

PR #5936, opened 2026-09-17
test: add in-game regression tests for five more plugins

Alistair-Afton:
## Summary
- `reveal`: reveal/unreveal/revtoggle hidden-flag round-trips, already-revealed and disabled-demon guards
- `dwarfvet`: status output reflects enable state, `now` checkup, usage errors
- `cleaners`: `clean map|units|items|plants|all` run clean, usage errors
- `cleanconst`: full item scan for bogus construction flags
- `cleanowned`: `dryrun` scans with all filters, usage errors

#### Test plan
- [x] All 20 new tests pass in-game (42/42 checks)
- [x] `ci/lint.py` clean

PR #5943, opened 2026-09-17
ci: raise lua test retry timeout to 3 minutes

Alistair-Afton:
## Summary
- The `Run lua tests` step retries via `nick-fields/retry` with `timeout_minutes: 1`.
- Each attempt reruns the **entire** suite: `ci/run-tests.py` deletes `test_status.json` on startup, so `--resume` only helps DF restarts *within* an attempt, not across outer retries.
- The suite has grown enough that a full Windows run (DF boot + all tests + quit) now exceeds 60s. PR #5936 hit this: all 324 tests passed, but every attempt died at the 60s mark during shutdown, failing the job.
- Raises the per-attempt limit to 3 minutes. Worst-case wall time for a genuinely hung suite goes from 3 to 9 minutes, which is still bounded.

#### Test plan
- [x] YAML parses
- [x] Verified #5936's artifact shows the suite completing all tests before the timeout killed DF during quit


SilasD (approved):
this is a necessary fix now that we're actually getting some test suite coverage.

the change is minimal and looks both reasonable and correct.

PR #5944, opened 2026-09-17
showmood: report bar/cloth/thread requirements in items, not raw units

magnus-ISU:
The wiki says:

> Once a workshop is claimed, the dwarf will begin collecting materials. Each artifact will require 1-3 "base items" and up to 7 additional items for decorations. The dwarf may well need several items of one material!

However, strange moods actually only require one unit of bars. The reason is that DF doesn't multiply the requested amount by 150 (the size of a bar). Therefore, it requests 1-3 (probably) bars and gets 150 from the first bar, and then is done.

The old code checked if more than the divisor was being requested, and if not didn't use it. This code always uses the divisor, but rounds up to 1.

Cloth moods request 10000 to 30000 cloth, which correctly shows up as 1 to 3 cloth.

This PR makes showmood accurately report that 3 bars requested means 1 bar actually.

It also adds a size for thread, which isn't necessary but would be if thread was added to moods, and rome told me I shouldn't remove that.


ab9rf:
I'm not convinced that this will correctly report on the remaining requirement in the situation where some of the item stacks already fetched for a cloth (or, in principle, thread) requirement are partial stacks. There was a bug in the base game related to this which, to the best of my knowledge, was fixed by Bay12 fairly recently.

PR #5946, opened 2026-09-17
Units::teleport: keep occupancy flags while other units remain on the tile

### LINKED ISSUE #5938: `teleport` does not properly adjust `unit_grounded` flags when teleporting a unit
`teleport` incorrectly clears `unit_grounded` when there is another grounded unit in the tile from which a unit was teleported away. DF's own logic only clears the `unit_grounded` flag if there are no other grounded units in the same tile.

The current code acknowledges that this is incorrect in a comment and suggests that it's acceptable (`// This is potentially wrong, but the game will recompute this as needed`) but I would prefer for the tool to do the right thing here.

_Originally posted by @ab9rf in https://github.com/DFHack/dfhack/issues/5924#issuecomment-5716719044_

Alistair-Afton:
## Summary

Fixes #5938.

`Units::teleport` cleared the source tile's `unit_grounded` flag unconditionally, so teleporting one grounded unit away from a tile left the flag unset even when another grounded unit still occupied it. The game itself only clears the flag when the last grounded unit leaves the tile.

teleport now records which tiles still have a grounded or standing unit on them (honoring the 3x3 EQUIPMENT footprint) and only clears `unit_grounded`/`unit` when the departing unit was the last of its kind there.

## Testing

- New `test/modules/units_fortress.lua` covers both directions: teleporting one of two grounded units off a tile keeps `unit_grounded`, teleporting one of two standing units keeps `unit`, and removing the last unit still clears the flag.
- 2/2 tests, 19/19 checks pass in-game on 53.16, plus a map-wide audit afterward found zero stale occupancy flags.
- All modules tests pass (5/5; the 2 skipped are title-mode-only).


library/modules/Units.cpp
```diff
@@ -780,25 +781,45 @@ bool Units::teleport(df::unit *unit, df::coord target_pos)
 
     // EQUIPMENT units (e.g. wagons) occupy a 3x3 footprint centered on their
     // position; all other units occupy just their position tile
-    int extent = 0;
-    if (auto craw = df::creature_raw::find(unit->race); craw &&
-        craw->flags.is_set(df::creature_raw_flags::EQUIPMENT_WAGON))
-        extent = 1;
+    auto unit_extent = [](df::unit *u) {
```
SilasD, line 787: terminology question for ab9rf or whoever:
in this closure, no variables are captured from the outer function, in contrast to `for_each_occupied_tile` immediately below, which captures `extent`.
is it still called a closure, or is it a nested function or similar term?

in Lua, closures are functions that capture upvalues, so this would be a nested function.
(although the Lua language spec basically says that new closures are made from all functions on each execution.)

Alistair-Afton: @SilasD In C++ this is still a lambda expression producing a closure object, just with an empty capture set. Captureless lambdas are even implicitly convertible to plain function pointers, which is the practical consequence of capturing nothing. Nested function is not standard C++ terminology since the language has no named nested functions, only lambdas and std::function tricks.

library/modules/Units.cpp
```diff
@@ -780,25 +781,45 @@ bool Units::teleport(df::unit *unit, df::coord target_pos)
...
     };
 
+    // Occupancy flags are per-tile, not per-unit: only clear a flag when no
+    // other unit of the same kind still occupies the tile, matching how the
+    // game itself updates the flags when a unit leaves a tile
+    std::unordered_set<df::coord> grounded_tiles, standing_tiles;
+    for (auto other : world->units.active) {
+        if (other == unit)
+            continue;
+        auto &tiles = other->flags1.bits.on_ground ? grounded_tiles : standing_tiles;
```
SilasD, line 814: special cases that need consideration:
* are units that are flying standing, grounded, or neither?
* * (I think they follow the normal rule of only one standing per tile, but I don't *know* that to be true.)
* are units that are swimming standing, grounded, or neither?
* * (again I think the standard rule, but I don't know for certain.)
* are units that are projectiles standing, grounded, or neither?
* * I'm guessing neither, but that's only a guess.
* are units that are in a built cage standing, grounded, or neither?
* are units that are in a loose cage standing, grounded, or neither?
* * I think in both cage cases they're grounded or neither, because units can move through the cage's tile without crawling.
* Edit: what about ghostly units?  standing, grounded, or neither?

library/modules/Units.cpp
SilasD, line 852: standing vs. grounded is treated as a binary here, but it's probably correct to do so.
we don't teleport units *into* cages, for example.

test/modules/units_fortress.lua
```diff
@@ -0,0 +1,136 @@
+config.mode = 'fortress'
+config.target = 'core'
+
+local function tile_occupancy(pos)
+    local block = dfhack.maps.getTileBlock(pos)
+    return block and block.occupancy[pos.x % 16][pos.y % 16]
+end
```
SilasD, line 6: we have a getter in the API for this, it's 
`select(2, dfhack.maps.getTileFlags(pos))`
or
`local designation, occupancy = dfhack.maps.getTileFlags(pos)`

test/modules/units_fortress.lua
```diff
@@ -0,0 +1,136 @@
...
+
+-- find an allocated tile with no unit occupancy near pos
+local function free_tile_near(pos)
+    for dx = -4, 4 do for dy = -4, 4 do
+        if dx ~= 0 or dy ~= 0 then
+            local other = xyz2pos(pos.x + dx, pos.y + dy, pos.z)
+            local occ = tile_occupancy(other)
+            if occ and not occ.unit and not occ.unit_grounded
+                    and occ.building == df.tile_building_occ.None
+                    and dfhack.maps.getTileType(other) then
```
SilasD: this checks that the target tile is not `nil`.
~~it might be important to the test suite to also check that it is not empty space.~~
~~i.e. check that it is not equal to `df.tiletype.RampTop` or `df.tiletype.OpenSpace`.~~
~~might as well check not equal to the illegal `df.tiletype.Void`.~~
cancel that, you want to check that the `other` tile can be walked on, so add this test:
`and df.tiletype_shape.attrs[ df.tiletype.attrs[dfhack.maps.getTileType(other)].shape ].walkable`
I think I got that right.

~~Edit: to be clear, testing that the tiletype is not `nil` is still necessary.~~
no it isn't, because of the clause that tests that `occ` is not `nil`, so the tile exists, so `getTileType` will succeed.

test/modules/units_fortress.lua
```diff
@@ -0,0 +1,136 @@
...
+                return other
+            end
+        end
+    end end
+end
```
SilasD, line 39: this returns `nil` as a failure code.  the code below detects failure with the string `'no free tile near the shared tile'`.

test/modules/units_fortress.lua
```diff
@@ -0,0 +1,136 @@
...
+    end
+    for _, unit in ipairs(units) do
+        local occ = tile_occupancy(unit.pos)
+        if occ then
+            if unit.flags1.on_ground then
+                occ.unit_grounded = true
+            else
+                occ.unit = true
+            end
+        end
```
SilasD, line 59: standing vs. grounded is treated as a binary here, which is probably fine in a test suite.

Edit: presumably also considered a binary in the actual test code below.  again, fine for a test suite.

test/modules/units_fortress.lua
```diff
@@ -0,0 +1,136 @@
...
+    end
+    for _, unit in ipairs(units) do
+        local occ = tile_occupancy(unit.pos)
+        if occ then
+            if unit.flags1.on_ground then
+                occ.unit_grounded = true
+            else
+                occ.unit = true
+            end
```
SilasD, line 58: I don't think this works.  `occ` is not a pointer to the tile occupancy, it is a copy of it.
we don't have a setter yet, so you need to use this after line 47 and line 56:
`dfhack.maps.getTileBlock(pos).occupancy[pos.x%16][pos.y%16] = occ`
(untested code)
substitute `unit.pos` for the line 56 copy.
you can skip a `nil` test, the tile is known to exist, so the block also exists.

library/modules/Units.cpp
SilasD: can a caged creature be valid for teleportation?  if so, this needs to use `Units::getPosition()`.

the teleport UI can absolutely choose to disallow this, but I envision players wanting to free a friendly unit from a cage, and it's non-obvious how to do that using only the game's built-in mechanics.

anyway, it's best practice to use `Units::getPosition()`.

Edit: teleporting a unit out of a cage would not be as simple as changing their `unit.pos`, there are general refs that would need to be removed, and (from memory) there's a `cage.occupants` vector.

SilasD (dismissed):
~~this one is NOT ready for prime-time.~~

I think this will work 99% of the time, but there are some interesting special cases that should be considered and possibly resolved.

~~this one probably needs attention from ab9rf or Quietust.~~

library/modules/Units.cpp
SilasD: makes sense, thank you.

library/modules/Units.cpp
Alistair-Afton: Checked these in a live fort:

- flying: there is no flying unit flag; fliers are ordinary units on the map and set occupancy normally
- swimming: verified `on_ground=false` with `occ.unit=true` on their tile -- swimming units count as standing
- projectiles: now skipped in the recompute scan (`flags1.projectile`)
- caged (built and loose cages both use `flags1.caged`): verified two caged units have `on_ground=false` and their pos tile has neither occupancy flag set -- caged units do not set occupancy, so they are now skipped in the scan
- ghostly: none in the test fort, so I could not verify; they are left counted since I have no evidence they differ from normal units

test/modules/units_fortress.lua
Alistair-Afton: done -- tile_occupancy() now uses getTileFlags.

test/modules/units_fortress.lua
Alistair-Afton: done -- the walkable check is in, and the nil getTileType guard is dropped since a non-nil occ already implies the tile exists.

test/modules/units_fortress.lua
Alistair-Afton: fixed -- free_tile_near now returns nil plus the failure string, and the callers forward it to expect.ne.

test/modules/units_fortress.lua
Alistair-Afton: I checked this in-game: getTileFlags pushes the pointer returned by Maps::getTileOccupancy, which is &block->occupancy[x&15][y&15], so the Lua object is a live ref into the block (same address as getTileBlock(pos).occupancy[x%16][y%16]), and direct field writes do propagate. The mutation sites work as written; the only change needed was reading via getTileFlags, which I have now adopted.

library/modules/Units.cpp
Alistair-Afton: done -- the source tile now resolves through getPosition(unit), which is correct even for caged units. You are right that actually freeing a caged unit needs more than a pos write (general refs, cage occupants), so teleport still only moves unit->pos; callers that want to exclude caged units can gate on flags1.caged.

Alistair-Afton:
For context on the "needs attention from ab9rf or Quietust" point: ab9rf already verified the underlying behavior on #5924 — from reverse engineering, `equipment` on the tile is the correct way to infer a 3x3 occupancy footprint, and he explicitly confirmed there that `teleport` incorrectly clearing `unit_grounded` while another grounded unit remains is a preexisting defect. This PR is that fix.

All of SilasD's special-case items are addressed in 86463ce1df plus the test-side fixes: caged and projectile units are skipped in the recompute scan (verified in-game that they set no occupancy), the source tile resolves through `Units::getPosition`, the test uses `getTileFlags` plus a walkable-tiletype check, and `free_tile_near` returns nil+message. The only unverifiable case is ghostly units (none in the test fort); they are left counted.

test/modules/units_fortress.lua
SilasD: ~~Hmmm I'll experiment.~~  verified it is a live pointer.

I have always treated it as a copy, so I've been doing extra work.  (I know the C++ version is a pointer.)

```C++
static int maps_getTileFlags(lua_State *L)
{
    auto pos = CheckCoordXYZ(L, 1, true);
    Lua::PushDFObject(L, Maps::getTileDesignation(pos));
    Lua::PushDFObject(L, Maps::getTileOccupancy(pos));
    return 2;
}
```
so that *doesn't* copy, it is live data, interesting.

in contrast, `dfhack.maps.getTileType()` returns an integer, not a pointer to a uint16_t.  it doesn't have a setter either.

it's been on my TODO list to add setters for these, but clearly `getTileFlags` doesn't need a separate setter.

SilasD (approved):
I think this covers my issues.  **All good to go.**

if some user is teleporting a ghost around, or teleporting a unit onto a ghost's tile, they have bigger problems than occupancy flags.

as an aside, there is a flying unit flag but it has nothing to do with block occupancy.
from memory, it's unit.enemy.caste_flags.FLIER, and clearing it will cause a flying unit to plummet to the ground. this group of flags are a copy of the master flags in the unit's creature_raw's caste_raw.flags, and they are not "sticky" -- they are recreated during the map load.

Issue #5937, opened 2026-09-17
`plotinfo.site_id` is -1 in a reclaimed fortress - can we safely mitigate?


ab9rf:
should we have a tool that forcibly updates `plotinfo.site_id` when it's found to be -1 in fort mode (as this violates what we believe to be an invariant) as a mitigation of a DF bug - at least until Bay12 fixes it

_Originally posted by @ab9rf in https://github.com/DFHack/dfhack/pull/5923#pullrequestreview-5237414073_

Alistair-Afton:
Implemented in #5948: `quickfix.repair_site_id()` runs from `onMapLoad.default.init` and restores `site_id` from `plotinfo.main.fortress_site` when DF leaves it at -1. It only writes when the invariant is violated and prints a note when it repairs something.

ab9rf:
closed by #5948

PR #5948, opened 2026-09-17
quickfix: repair unassigned site_id on reclaimed fortresses

Alistair-Afton:
## Summary

Implements the mitigation discussed in #5937: on map load, if `plotinfo.site_id` is still -1 (a vanilla defect on reclaimed fortresses until the first save), it is restored from `plotinfo.main.fortress_site`, which is set at embark.

Lives in `quickfix` alongside the existing entity race repair, called from `onMapLoad.default.init`. Silent unless it actually repairs something.

## Testing

- New `test/modules/quickfix_fortress.lua`: fabricates the defect (site_id = -1), verifies the repair restores `fortress_site.id`, and verifies a no-op when site_id is already assigned.
- 2/2 tests, 3/3 checks pass in-game on 53.16.


docs/changelog.txt
```diff
@@ -76,6 +76,7 @@ Template for new versions:
 - `timestream`: deal properly with units who have breathing difficulties
 - `stocks`: overlay now resets scroll position when collapsing categories so the item list is no longer left blank and unscrollable
 - Fixed persistent site data API (``dfhack.persistent.saveSiteData``/``getSiteData``) failing on newly reclaimed fortresses until the first save
+- ``quickfix``: repair ``plotinfo.site_id`` on load when DF leaves it unassigned on reclaimed fortresses
```
ab9rf: This isn't a "fix"; fixes are changes that fix a defect in _DFHack_. Because it provides a mitigation for a defect in DF, this is a _new feature_ and thus should be placed in that category.

ab9rf (approved):

PR #5953, opened 2026-09-18
sort: clickable stress icons in the top bar

Alistair-Afton:
## Summary
- Adds an invisible hotspot overlay widget (`sort.stress_icons`) over the seven stress/happiness icons next to the "Pop" label in the fortress-mode top bar.
- Clicking an icon opens a searchable `ListBox` of the citizens in that stress category (e.g. "Citizens: Unhappy").
- Selecting a citizen dismisses the list and centers the map on them via `dfhack.gui.revealInDwarfmodeMap`.
- Empty categories show a brief message instead of an empty list.
- Icon positions are discovered each second by scanning the top bar for the `Pop` label, so the hotspot tracks window resizes and fort-name length changes. Only active on `dwarfmode/Default`.
- Includes a new in-engine test file (`test/plugins/sort.lua`) covering rect detection, click handling, list filtering, and map centering.

## Test plan
- [x] `test -t plugins.sort` in-game on a fortress save: 5/5 tests, 38/38 checks pass
- [x] Clicked all seven icons in-game: populated categories open the filtered list; empty categories show the "no citizens" message
- [x] Selecting a dwarf recenters the map viewport on their position (verified across z-levels)
- [x] Clicks outside the icons pass through to the map; clicks on the modal list are consumed


Alistair-Afton:
Windows test-leg failures are the known DF startup crash (`DF exited with 3221226356` at launch, before plugins load) — same signature appears in passing runs' logs, e.g. the fix-workshop-search-names run. Rebased onto develop (which now includes the #5943 retry fixes) to retrigger.

Alistair-Afton:
Correction to the above: the Windows failures were not the startup crash — the tests ran (310/314) and errored on `icon_rects` being nil. On Windows CI the window is hidden, so the vanilla top bar never renders and the `Pop` text the scan anchors on is absent. Fixed in 7d9af7f: the wrapper waits ~20 frames for the top bar, marks scanning unavailable when absent, and the click/list/reveal tests inject synthetic rects so the logic is still exercised. CI is now green on all legs.

ab9rf (approved):

PR #5967, opened 2026-09-22
make stderr unconditionally unbuffered

ab9rf:
windows `freopen` opens the file in _buffered_ mode, which is undesirable for a log file


ab9rf:
This one is a real annoyance when testing on Windows so I'm not going to delay merging it

PR #5968, opened 2026-09-22
fix dfhack-run on windows

### LINKED ISSUE #3827: `dfhack-run` on windows murders any console it is run from
`dfhack-run` will forcibly close the Windows console it was run from, if run from either Powershell or CMD

this is because the process already has a console when it calls `AllocConsole` so this call fails, and then `GetConsoleWindow` returns the already-existing console (which is _shared_ with the parent PowerShell or CMD process, this is how Windows works), and then when `dfhack-run` exits it frees this console which causes it to close which is undesired

probably the best solution here is to check the result of `AllocConsole` for failure, and if it failed mark the console as "borrowed" and not shut it down on exit

ab9rf:
fixes #3827


ab9rf:
Also a real annoyance when testing on Windows.

